PR #53, opened 2015-02-17
[WIP] Add configbridge script

arif-ali:
* Allows to add linuxbridge
* Allows to add OVS bridges
* taken from xHRM script and modified

### DIFF
--- a/xCAT/postscripts/configbridge
+++ b/xCAT/postscripts/configbridge
@@ -1,0 +1,594 @@
+#!/bin/bash
+
+# load library for network caculation
+if [ "$(uname -s|tr 'A-Z' 'a-z')" = "linux" ];then
+   str_dir_name=`dirname $0`
+   . $str_dir_name/xcatlib.sh
+fi
+
+function add_br() {
+  BNAME=$1
+
+  if [[ $BRIDGE == "ovs" ]]; then
+    ovs-vsctl add-br $BNAME
+  elif [[ $BRIDGE == "linux" ]]; then
+    brctl addbr $BNAME
+  fi
+}
+function add_if() {
+  BNAME=$1
+  PORT=$2
+
+  if [[ $BRIDGE == "ovs" ]]; then
+    ovs-vsctl add-port $BNAME $PORT
+  elif [[ $BRIDGE == "linux" ]]; then
+    brctl addif $BNAME $PORT
+  fi
+}
+
+#written in bash for fewest prerequisites
+function get_def_interface {
+    #gethostbyname on `hostname` would be helpful, but since we
+    #are in bash, the best alternative is to use ping to get at it
+    #don't want to grep in /etc/hosts or presume DNS
+    #we are, however, presuming ipv4 for the moment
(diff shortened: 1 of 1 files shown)


topstarnetwork:
Hello,
I would like to know what the difference is between xcat https://sourceforge.net/projects/xcat/ and opennebula?

Are those both same products or can I use opennebla on top of xcat???

I had asked this question to nebula support and here is thiere answer;



xCat is somehow a different problem solving toolkit, oriented towards bare metal preparation, configuration and management - in this sense is more similar to Foreman, Cobbler or Crowbar (an overview with some descriptions is here: http://aucouranton.com/2013/07/31/puppet-chef-orchestration-and-devops/1 ). OpenNebula takes from there, exactly like OpenStack. There is an initial opennebula-foreman integration (here, from the great people of Netways: http://opennebula.org/foreman-integration/ ) to provide initial bare metal support to opennebula as well.


So my replay to this ticket is as follow,if someone could guide me please it would be nice;



So if I understand well xcat is anything that all the all configuration and management softwares that you sent me in your list,
as I am not a system admin but just a project owner if I am not mistaking xcat is superiour and has more features then foremen and all other softwares mentioned in your list am I right?

If si should I use xcat plugged into opennebula or opennebula plugged into foreman and puppet????

A far as I read xcat has clustering features that other softwares do not have am I right????
Regards, Heather


Thanks, Heather

arif-ali:
re-opened, as at the time PR's weren't being used to get commits

daniceexi:
@arif-ali 
Could you give more information of the usage of this script. And could you give out the reason why you want to rewrite from xHRM instead of change it?

@zet809 will review this code.

arif-ali:
I am using xCAT to deploy xCAT with saltstack, and want to be able to create the external bridge interfaces using openvswitch. As the creation of the external bridge is external to the OpenStack, i'd like to be able to do it through xCAT, and manage all my interfaces through this.

For example, currently I have the following in my postscripts table 

```
configbridge ovs enp32s0f0@enp32s0f1%bond0:br-bridge0
```

This allows me to create an openvswitch bridge, create a bond, and assign the 2 interfaces to the bond. I am happy to change `xHRM`, but that uses linuxbridge with `brctl`, and not openvswitch with `ovs-vsctl`. However by changing `xHRM` that could change other people's configs, and therefore cause people problems. Plus configbridge makes more sense than `xHRM`.

xCAT/postscripts/configbridge
```diff
@@ -0,0 +1,594 @@
...
+  fi
+else 
+  nwdir="/etc/sysconfig/network-scripts"
+  isRedhat=1
+fi 
+
+#Type of bridge, ovs, brctl
+
+if [ "ovs" = "$1" ]; then
+
```
zet809: We need to check whether openvswitch related pkgs have been installed before using them.

zet809: We have configbond to config bond for a group of NICs, and have configvlan to config vlan. Why don't configbridge focus on config bridge without config bond or vlan? How to position configbridge? It have function overlap with configbond and configvlan.

arif-ali:
#469 will actually do this now, so closing pull request

looking forward to using the new scripts in 2.12

PR #56, opened 2015-07-31
Fix configefi script

arif-ali:
* use df instead of mount, and check for the disk that /boot is on, rather than /
* use grub2-install if that exists, as that is the default on el7

This is something I have been testing on a system at the moment, and these changes were able to resolve some of the issues

### DIFF
--- a/xCAT/postscripts/configefi
+++ b/xCAT/postscripts/configefi
@@ -21,7 +21,7 @@ if [ "$arch" =  "x86_64" ]; then
                 elif [ -f "/boot/efi/EFI/redhat/grubx64.efi" ];then
                         efibootmgr -c -l \\EFI\\redhat\\grubx64.efi -L syscloneLinux
                         
-                        boot_root=`mount | grep -E ' on\s+/ type ' | awk '{print $1}'`
+                        boot_root=`df /boot | awk '{print $1}' | tail -n 1`
                         sed -i 's| root=\S*| root='$boot_root' net.ifnames=0|' /boot/efi/EFI/redhat/grub.cfg 
       
                         blkid -c /dev/null |grep UUID|while read str_line
@@ -52,10 +52,10 @@ if [ "$arch" =  "x86_64" ]; then
 		boot_device=''
 		if [ -f "/etc/systemconfig/systemconfig.conf" ];then
 			#boot_root=`cat /etc/systemconfig/systemconfig.conf | grep ROOTDEV | awk '{print $3}'`
-                        boot_root=`mount | grep -E ' on\s+/ type ' | awk '{print $1}'`
+                        boot_root=`df /boot | awk '{print $1}' | tail -n 1`
 			boot_device=`cat /etc/systemconfig/systemconfig.conf | grep BOOTDEV | awk '{print $3}'`
 		else
-			boot_root=`mount | grep -E ' on\s+/ type ' | awk '{print $1}'`
+			boot_root=`df /boot | awk '{print $1}' | tail -n 1`
 			boot_device=`echo $boot_root | sed -e 's/[0-9]*$//'`
 
 			#str_temp=`mount | awk '{print $1","$3}'`
@@ -85,8 +85,13 @@ if [ "$arch" =  "x86_64" ]; then
 			# note: the error about grub-set-default not existing is harmless, because we want the default to be 0 anyway
 		else
 			#for sles10, should run grub-install with parameters
-			echo "grub-install --no-floppy --recheck $boot_device"
-			grub-install --no-floppy --recheck $boot_device
+			if [ -x /usr/sbin/grub2-install ]; then
+				echo "grub2-install -s --recheck $boot_device"
+				grub2-install -s --recheck $boot_device
+			else
+				echo "grub-install --no-floppy --recheck $boot_device"
+				grub-install --no-floppy --recheck $boot_device
+			fi
 		fi
 
 		# change the entries in the grub conf file to use the correct boot root device


daniceexi:
The change looks OK. But could you give more information of what was the issue so that we can figure out what was this change for?

arif-ali:
1) The partition layout for these machines had a seperate /boot partition. The current method of the script checks for /, and therefore grub-install was installing to the LVM partition rather than the main root device. Hopefully this would fix the problem

2) el7 OS's do not have grub-install by default, and therefore will not be able to install the grub to the device, so need to check for grub2-install, so that we can use that to configure and install grub onto the device

I hope that makes sense

hu-weihua:
Hi arif, thank you very much for your contribution.  The second change looks fine.  but for the first one, I have some question.  In my environment, the boot partition is also a separate partition. but grub-install was not installed to the LVM partition. cause when you capture image from golden client, the file /etc/systemconfig/systemconfig.conf was generated, you can see the  boot_device come from this file by below code:
boot_device=`cat /etc/systemconfig/systemconfig.conf | grep BOOTDEV | awk '{print $3}'`

below code is how systemconfig.conf get BOOTDEV in golden client:
-------------------
open(IN, 'mount |') or
        die("FATAL: couldn't execute mount!\n");
    chomp(@_ = <IN>);
    close(IN);
.......
    my @boots = grep(/ \/boot /, reverse @_);
    my $boot_dev = (split(/ /, $boots[0]))[0] if (@boots);
    unless ($boot_dev) {
        # Unable to detect boot device (using the root device)
        $boot_dev = $root_dev;
    }
    if ($boot_dev =~ /^\/dev\/([hs]|ps3)d/) {
        # Standard disk naming (hd*, sd*, ps3d*).
        $boot_dev =~ s/[0-9]+$//;
    } else {
        # Hardware RAID device.
        $boot_dev =~ s/p[0-9]+$//;
    }
-----------------
it seems $boot_dev do come from the information of /boot 

I am not sure why it failed in your environment, why  systemconfig.conf  wasn't generated or  systemconfig.conf  got wrong information. Maybe not wrong information, just because you have especial environment. could you give me more information about your environment?  thanks

arif-ali:
In my environment the `/etc/systemconfig/systemconfig.conf` doesn't get created, so therefore it uses the mount or df command to grab the details it needs.

I first installed a node, using the following osimage 

```
Object name: compute
    groups=all
    imagetype=linux
    osarch=x86_64
    osname=Linux
    osvers=rhels7.1
    otherpkgdir=<..snip..>
    otherpkglist=/install/templates/compute/otherpkgs.pkglist
    partitionfile=/install/templates/compute/partfile
    pkgdir=/install/rhels7.1/x86_64
    pkglist=/install/templates/compute/pkglist
    profile=compute
    provmethod=install
    synclists=/install/templates/compute/synclists
    template=/opt/xcat/share/xcat/install/rh/compute.rhels7.tmpl
```

where the pkglist is

```
#INCLUDE:/opt/xcat/share/xcat/install/rh/compute.rhels7.pkglist#
tcl
tk
```

partitionfile is

```
part /boot --fstype=xfs --size=500
part /boot/efi --fstype=vfat --size=100
part pv.1 --grow --size=1

volgroup system --pesize=4096 pv.1

logvol / --fstype=xfs --name=root --vgname=system --size=20480
logvol swap --name=swap --vgname=system --size=4096
logvol /var --fstype=xfs --name=var --vgname=system --size=20480
```

otherpkgs.pkglist

```
#INCLUDE:/opt/xcat/share/xcat/install/rh/sysclone.rhels7.x86_64.otherpkgs.pkglist#
ganglia/ganglia-gmond
nagios/nrpe
nagios/nagios-plugins
```

I then installed the machine, and added a few drivers for Intel IB, and then used the Howto on https://sourceforge.net/p/xcat/wiki/Using_Clone_to_Deploy_Server/ to create the image. Even then, I had to modify the `/install/sysclone/scripts/<imagename>.master` as the partitioning wasn't being created correctly

I hope that makes sense

hu-weihua:
Hi Arif,  it seems your golden client is x86_64, rhels7.1 and using UEFI.  then I have 3 questions.

1. you said you have to modify <imagename>.master as the partitioning wasn't being created correctly. if so, why did you need to change the code of configefi.

2. You partition involved efi. I guess your x system is UEFI mode.  so while the code went into configefi, is should go below part:
------------------------------
if [ "$arch" =  "x86_64" ]; then
	if [ -d /sys/firmware/efi ]; then
		echo "Setting Boot Manager for the next boot."
		echo "delete all sysclone boot list"
		str_bootnums=`efibootmgr | grep 'syscloneLinux' | awk '{print $1}' | sed 's/boot//i' | sed 's/*//'`
		for str_num in $str_bootnums
		do
			efibootmgr -b $str_num -B -q
		done

		if [ -f "/boot/efi/EFI/redhat/grub.efi" ];then
			efibootmgr -c -l \\EFI\\redhat\\grub.efi -L syscloneLinux
		elif [ -f "/boot/efi/efi/SuSE/elilo.efi" ];then
			efibootmgr -c -l \\efi\\SuSE\\elilo.efi -L syscloneLinux
		else
			echo "Can not find the boot loader."
			exit 1
		fi
--------------------------------
so how did you find "grub-install was installing to the LVM partition rather than the main root device".  this is the code logic of another scenario that x system is legacy mode

3. if your <imagename>.master didn't create the partition correctly, cloud you please offer me the partition details of <imagename>.master?

arif-ali:
1) Because the script didn't detect where the /boot, and grub-install wasn't there

2) `/sys/firmware/efi` was never created in genesis, as this is the first thing I checked, when I was diagnosing

3) 

```
--- cloned-node.master	2015-08-05 09:57:24.651288393 +0100
+++ cloned-node.master.bak.arif	2015-08-04 11:50:08.061254459 +0100
@@ -259,10 +259,6 @@
 ### END LVM groups creation commands ###
 
 ### BEGIN LVM volumes creation commands -AR- ###
-logmsg "lvcreate  -L20971520K -y -n var system || shellout"
-lvcreate  -L20971520K -y -n var system || shellout
-logmsg "lvscan > /dev/null; lvchange -a y /dev/system/var || shellout"
-lvscan > /dev/null; lvchange -a y /dev/system/var || shellout
 logmsg "lvcreate  -L4194304K -y -n swap system || shellout"
 lvcreate  -L4194304K -y -n swap system || shellout
 logmsg "lvscan > /dev/null; lvchange -a y /dev/system/swap || shellout"
@@ -271,6 +267,10 @@
 lvcreate  -L20971520K -y -n root system || shellout
 logmsg "lvscan > /dev/null; lvchange -a y /dev/system/root || shellout"
 lvscan > /dev/null; lvchange -a y /dev/system/root || shellout
+logmsg "lvcreate  -L20971520K -y -n var system || shellout"
+lvcreate  -L20971520K -y -n var system || shellout
+logmsg "lvscan > /dev/null; lvchange -a y /dev/system/var || shellout"
+lvscan > /dev/null; lvchange -a y /dev/system/var || shellout
 ### END LVM volumes creation commands ###
 
 logmsg "Load additional filesystem drivers."
@@ -306,7 +306,7 @@
 mount ${DISK0}1 /a/boot -t xfs -o defaults || shellout
 
 logmsg "mkdosfs -F 32 -v ${DISK0}2 || shellout"
-mkdosfs -F 32 -v ${DISK0}2 || shellout
+mkdosfs -F 32 -i 79D701AA -v ${DISK0}2 || shellout
 logmsg "mkdir -p /a/boot/efi || shellout"
 mkdir -p /a/boot/efi || shellout
 logmsg "mount ${DISK0}2 /a/boot/efi -t vfat -o defaults,uid=0,gid=0,umask=0077,shortname=winnt || shellout"
```

But when the .master file does the mkfs, and the mounting of the filesystems, it doesn't use the LVM names, rather, it uses the dm-X names, but that can be in any order when they are created. So I had to move the LV creation to match the order correctly.

Also the mkdosfs wasn't adding the correct UUID, that was mentioned in the fstab

hu-weihua:
Hi Arif, you are right, thanks for your code contribution

PR #57, opened 2015-08-02
enable syslog in the stateless initrd, add logs in the xcat dracut module

immarvin:
enable syslog module in the stateless initrd
add logs in the xcat dracut module
add kernel options for syslog module when xcatdebugmode=1


daniceexi:
Good implementation.

My major concern for this change is the code source of the patches for syslog module. Where did you get the pacthes? Current you just overwrite the original files with patch files instead of patch them. I am afraid that this way does not work for the future version of dracut.

Another comment is you did not use switcher to control the enable/disable the debug mode in dracut.

immarvin:
added the test of xcatdebugmode in the xcatroot

all the files need to be patched in the syslog dracut module should be included in the netboot/<osver>/dracut_033/patch/syslog, the relevant files in the os shipped syslog module  will be overwritten.

all the dracut stuff should be placed in the netboot/<osver>/dracut_033/patch/<module name> directory

daniceexi:
Two comments for your latest checkin.

1. You should check the genimage too since the file path of patch have been changed.

2. I just realized that you changed the initqueue.sh in the 99base module. Since this is a common module, I am concerning your change that just overwirte the whole file instead of make patch. Could you give the diff of your change in this file?

immarvin:
hi,
the genimage has been checked in 

the diff of initqueue :

Redhat 7.1:

[root@c910f02c02p08 rootimg]# diff usr/lib/dracut/modules.d/99base/initqueue.sh /usr/lib/dracut/modules.d/99base/initqueue.sh
24,25d23
<         --online)
<             qname="/online";;

Sles12:
c910f03c04k04:~ # diff /opt/xcat/share/xcat/netboot/sles/dracut_033/patch/base/initqueue.sh  /usr/lib/dracut/modules.d/99base/initqueue.sh 
24,25d23
<         --online)
<             qname="/online";;

immarvin:
since the syslog module and logger binary are added into the diskless initrd, the size of initrd is increased from 22044908 B to 22514864 B.

daniceexi:
Good. You have moved the 'start' control part for syslogd to the syslog module from the base module. Then I am comfortable with the change since all the change in this pull request only affect the syslog service at all.

Please merge the changes to our master branch. Thanks.

PR #74, opened 2015-08-11
update bmcdiscover related rest api

bybai:
1, set default scan_method nmap
2, update bmcdiscover related rest api


samveen:
@bybai If you look at the xcat code you will notice that the convention for blocks is:
```
if (condition) {
    ...
}
```
instead of  
```
if (condition)
{
    ...
}

It would be great if you would stay with that for consistency purposes.

PR #68, opened 2015-08-10
bmcdiscover -s -z -w enhancements.

bybai:
1, bmcdiscover -z -w suppport
2, --help changes
3, command attributes enhancements

PR #75, opened 2015-08-12
add -m to void breaking backwards compatibility

bybai:
bmcdiscover command accepts both -s and -m for methods;
We document -s, but -m should also work for backward compatibility;


whowutwut:
This is in response to the discussion started in Issue #68 

According to Christopher's comment in the mailing list, should we start the process for deprecating this option?  We need to print a message saying -m is no longer supported, user should use -s instead.  Then we can plan to deprecate the -m option in a future release.

daniceexi:
@whowutwut Since we announced this bmcdiscover command as experimental, I think we can change the interface without consider the customer too much. 
Now Baiyuan's fix is to take back the '-m' option for backward compatibility. It should be OK. What do you think?

whowutwut:
@daniceexi , ok sounds good

daniceexi:
@bybai We agree to merge in the code. Thanks.

Issue #78, opened 2015-08-13
nodediscoverls output truncates serial


samveen:
xCAT version:
```
[root@inframgtlivexcatmn1 xCAT_plugin]# rpm -q xCAT-server
xCAT-server-2.9.1-snap201503190325.noarch
```

Serial length on node:
```
[root@inframgtlivexcatmn1 xCAT-scripts]# ssh spare35-b6 "awk '{printf \$1}' /sys/class/dmi/id/product_serial  |wc -c"
10
```
`nodediscoverls` output:
```
[root@inframgtlivexcatmn1 xCAT_plugin]# nodediscoverls |grep spare35-b6
  31313636-3938-4753-4835-3137584B3557    spare35-b6          sequential               XXXXXXXX  
```

Offending line:
```
[root@inframgtlivexcatmn1 xCAT_plugin]# nl -ba /opt/xcat/lib/perl/xCAT_plugin/seqdiscovery.pm  |tail -n +891 |head -n 1
   891	            push @discoverednodes, sprintf("  %-40s%-20s%-15s%-10s%-10s", $ent->{'uuid'}, $ent->{'node'}, $ent->{'method'}, $ent->{'mtm'}, substr($ent->{'serial'},0,8));
```

jjohnson42:
Note for future, originally the concept was that presentation and the data format would not be  mixed.  Having sizes in that sprintf on the server side is a recipe for ugliness versus having the fields be arbitrary length.

I understand some of the factors driving this today (though I'm wondering why this field is fixed length as opposed to letting length vary), but in confluent the emphasis is on thicker client code to handle presentation issues more granular fashion.

jjohnson42:
Sorry clicked wrong button

daniceexi:
@jjohnson42 A better way is to calculate the max length for each column and give a proper space base on this max length.

samveen:
Please take into consideration that serial number width will keep increasing as time goes by :).

daniceexi:
Right. When that happens, we'll consider to calculate the max length of all entries to determine how many chars we need to use to display a column.

PR #79, opened 2015-08-13
Fix for #78 - nodediscoverls: Reduce UUID width to 37 and increase serial width to 13

samveen:
Fix for #78 - nodediscoverls: Reduce UUID width to 37 and increase serial width to 13

UUIDs [by definition](https://en.wikipedia.org/wiki/UUID) are 36 characters wide in their printable form:
```
[root@xcatmn01 ~]# awk '{printf $1}' < /proc/sys/kernel/random/uuid |wc -c
36
```
Thus for printing them, a width of 36 + 1 separator is enough, which is what I did, moving the extra 3 characters gained from it's reduction into the serial field.


daniceexi:
Thanks for the patch.

Extend the length of serial no to 13 is good for me, but I don't want to reduce the length of uuid to 37. It will be ugly to display.  I don't see the problem that extend the output length with 3 chars.

samveen:
Reverted reduction.

PR #80, opened 2015-08-14
Enhance docker.pm to communicate with docker host

bybai:
1, enhance genreq and send_req using ssl ca/cert/key.
2, add usage example for genreq and send_req to call docker rest api to communicate with docker host.


daniceexi:
Since this is just some experimental code that we will refine them for the real application, I just review it roughly.

PR #81, opened 2015-08-14
Trouble shooting--Record command output and add trace information

hu-weihua:
These new code covers 2 independent feature

1  record every command output into log file /var/log/xcat/command.log . This is new feature about xcatd. from now on, xcatd will record all commands received from client and their output into a log file, it's convenience for developer to find out client operation history.

the format of command.log likes below:

```
	====================================================
	[Date]       2015-7-13 23:4:55
	[ClientType] cli 
	[Request]    tabrestore 
	[Response]
	====================================================
	[Date]       2015-7-13 23:14:45
	[ClientType] cli 
	[Request]    nodeset c910f02c02p30 osimage=rhels6.5-ppc64-install-compute 
	[Response]
	c910f02c02p30: install rhels6.5-ppc64-compute
	====================================================

```

2 provide a trace()  for developer to display message in syslog.

This subroutine is implemented in MsgUtiles.pm. 
please refer below description to use trace subroutine

```
    Arguments:
        $verbose: indicate whether current command is with -V option. 1 is yes and 0 is no.  
        $level: the level of trace message, can be one of "I","W","E","D","i","w","e","d".
            "I" or "i": means information level.
            "W" or "w": means warning level.
            "E" or "e": means error level.
            "D" or "d": means debug level.
        
            As long as the trace subroutine is called, "information", "warning" and "error" level message will be displayed in syslog.
            The "debug" level message is displayed in syslog only when any one of the below two conditions is true
            1. The current command with -V option. i.e. $verbose=1.
            2. The xcatdebugmode, which is an attribute of site table, equals 1. 
        
            If $level is not any one of "I","W","E","D","i","w","e","d", the trace subroutine do nothing.
        
        $logcontent: the specific message wanted to be displayed in syslog
    Note:
        Label "xcat" and trace message level, such as "INFO", "ERR", "WARNING" AND "DEBUG", will be added before real trace message automatically. It's convenient to filter in syslog.
    Example:
        xCAT::MsgUtils->trace(1,"D","this is debug trace message");
        xCAT::MsgUtils->trace(0,"i","this is information trace message");
        xCAT::MsgUtils->trace(0,"E","this is error trace message");   

```

I have used trace() for nodeset and makedhcp command, this can be a sample for how to ues trace(), it involved those files:destiny.pm  petitboot.pm  anaconda.pm  yaboot.pm pxe.pm grub2.pm debian.pm sles.pm prescripts.pm  dhcp.pm  xnba.pm.

### DIFF
--- a/perl-xCAT/xCAT/MsgUtils.pm
+++ b/perl-xCAT/xCAT/MsgUtils.pm
@@ -756,7 +756,7 @@ sub verbose_message
             "D" or "d": means debug level.
         
             As long as the trace subroutine is called, "information", "warning" and "error" level message will be displayed in syslog.
-            The "debug" level message is displayed in syslog noly when any one of the below two conditions is true
+            The "debug" level message is displayed in syslog only when any one of the below two conditions is true
             1. The current command with -V option. i.e. $verbose=1.
             2. The xcatdebugmode, which is an attribute of site table, equals 1. 
         
--- a/perl-xCAT/xCAT/MsgUtils.pm
+++ b/perl-xCAT/xCAT/MsgUtils.pm
@@ -741,45 +741,80 @@ sub verbose_message
     }
 }
 
+#--------------------------------------------------------------------------------
+
+=head3 trace
+
+    Display different level trace message in systemd.
+    
+    Arguments:
+        $verbose: indicate whether current command is with -V option. 1 is yes and 0 is no.  
+        $level: the level of trace message, can be one of "I","W","E","D","i","w","e","d".
+            "I" or "i": means information level.
+            "W" or "w": means warning level.
+            "E" or "e": means error level.
+            "D" or "d": means debug level.
+        
+            As long as the trace subroutine is called, "information", "warning" and "error" level message will be displayed in systemd.
+            The "debug" level message is displayed in systemd noly when any one of the below two conditions is true
+            1. The current command with -V option. i.e. $verbose=1.
+            2. The xcatdebugmode, which is an attribute of site table, equals 1. 
+        
+            If $level is not any one of "I","W","E","D","i","w","e","d", the trace subroutine do nothing.
+        
+        $logcontent: the specific message wanted to be displayed in systemd
+    
+    Returns:
+        none
+		
+    Error:
+        none
+		
+    Note:
+        Label "xcat" and trace message level, such as "INFO", "ERR", "WARNING" AND "DEBUG", will be added before real trace message automatically. It's convenient to filter in systemd.
+    
+    Example:
+        xCAT::MsgUtils->trace(1,"D","this is debug trace message");
+        xCAT::MsgUtils->trace(0,"i","this is information trace message");
+        xCAT::MsgUtils->trace(0,"E","this is error trace message");
+=cut
+
+#--------------------------------------------------------------------------------
 sub trace(){
     shift;
     my $verbose = shift;
     my $level = shift;
-	my $logcontent = shift;
-
-	my $prefix = "";
-	if(($level eq "E")||($level eq "e")){$prefix="ERR";}
-	if(($level eq "W")||($level eq "w")){$prefix="WARNING";}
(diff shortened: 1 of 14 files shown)


daniceexi:
I will review this pull request.

daniceexi:
Great Implementation!!!

See my comments for each commits.

whowutwut:
@hu-weihua , Is it possible to add a sample in the initial issue post showing how one would use the new enhancements?

hu-weihua:
@daniceexi, the main modifications I did is listed below:
1. Add description for the trace subroutine in MsgUtils.pm and removed the useless lines
2. Add description for the cmdlog_submitlog and cmdlog_collectlog subroutine in xcatd
3. change 'xcatcport' to 'xcatlport'
4. handle the 'INT' or 'TERM' in 'Command log writer' process
5. Change some implementation of cmdlog_submitlog and make it just try 3 times when it connect with 'Command log writer' process
6. fix indent problem
7. combine the message to reduce the time to call trace
8. use the join() subroutine

hu-weihua:
@whowutwut  I have changed the initial issue post, is it helpful? any comment are welcome

whowutwut:
@hu-weihua  ,

*This subroutine is implemented in MsgUtiles.pm. I have add trace for nodeset and makedhcp command, this will be a sample for how to ues trace subroutine, it involved those files:destiny.pm petitboot.pm anaconda.pm yaboot.pm pxe.pm grub2.pm debian.pm sles.pm prescripts.pm dhcp.pm xnba.pm.*

So if I understand this correctly, the changes to nodeset and makdhcp are only samples. If the user or developer wants to utilize the trace function, they will need to modify the code and add new entries to call trace().... ?    

It looks like there's no way to turn off the trace function if the code is added, right? 
* Information/Warning/Error is always printed 
* Debug is printed with -V option and xcatdebug=1 



Should we have a way to turn on trace and not affect normal behavior?
As an example: 
* nothing set, normal behavior --> no msg are printed
* xcatdebug =1 -->  Info/Warning/Errors printed 
* add -V   --> Debug printed 

Just thinking that doing this we could slowly add more debug/trace msg to the code without adding unnecessary output messages that the end user would see...

daniceexi:
@hu-weihua Did you updated the Schema.pm for the new added site attribute xcatlport?

daniceexi:
@whowutwut We have discussed how to control the trace message base on the confgiuration. The basic idea was the 'Error/Warning/Info' messages should be written to log file (currently is syslog) always since xcat lack of messages for user to know what happens in the background, many customer complain that xCAT should add more messages in syslog.

If someday we want to control which log level can write to syslog, it will be easy to achive by chaning the trace subroutine.

hu-weihua:
@whowutwut 
the thing is, most developers hope there is a simple function to replace original 'message' function, which is complex, to display some different level message in syslog. that's trace().  During the future development, people will use trace() in everywhere they hope to display some message in syslog. we will also add trace in all commands to make trouble shooting easier. so I implemented the trace() in MsgUtiles.pm and used trace function in the code of nodeset and makedhcp. why are nodeset and makedhcp, is for 2 reasons, one is developer bothered with debugging nodeset and makedhcp for a long time, these tow commands have higher priority. another is people need an example to know how to use trace function in their own code, we chose nodeset and makedhcp as an example. so people need not to modify the code of trace(), or trace invoking already existed in nodeset and makedhcp. they just need utilize trace() in the same way as I used it in nodeset and makedhcp.

The logic of trace() depends on the discussion results in CDL meeting.  so people need to think carefully about the message level when they use trace() in their own code, if the message isn't very necessary, recommend debug level. Maybe you have noticed, almost all trace I added in nodeset and makedhcp are debug level. these trace can be turn off easily.

Just as @daniceexi said, if the logic of trace() isn't suitable for future requirement, we can change the code of trace() easily.

whowutwut:
@daniceexi @hu-weihua  :+1:  Thanks for the explanation!

daniceexi:
I reviewed all the changes, and I agree to check in.

PR #84, opened 2015-08-17
Look for Packages or Packages.gz in the otherpkgdir

mattaezell:
Allow the otherpkgdir to use a compressed Packages file so it can also be served over http


bybai:
Agree with these changes, Packages and Packages.gz are both  Packages index files for ubuntu.

daniceexi:
yes, the change looks good.

Issue #86, opened 2015-08-18
[DEV] "Error: either "dev" is duplicate, or "virbr0" is a garbage." message on the 1st reboot after kvmhost installation


immarvin:
while trying to provision a  kvmhost  on a redhat 6 x86_64 node with the following node and osimage definition:

```

[root@c910f04x18 tmp]# lsdef c910f04x20
Object name: c910f04x20
    arch=x86_64
    bmc=10.4.20.254
    cons=ipmi
    conserver=10.4.18.1
    currchain=boot
    currstate=install rhels6.7-x86_64-kvm
    getmac=ipmi
    groups=compute
    hcp=10.4.20.254
    initrd=xcat/osimage/rhels6.7-x86_64-install-kvm/initrd.img
    installnic=mac
    ip=10.4.20.1
    kcmdline=quiet repo=http://c910f04x18:80/install/rhels6.7/x86_64 ks=http://c910f04x18:80/install/autoinst/c910f04x20 ksdevice=40:f2:e9:0a:8c:64  cmdline  console=tty0 console=ttyS0,115200
    kernel=xcat/osimage/rhels6.7-x86_64-install-kvm/vmlinuz
    mac=40:f2:e9:0a:8c:64
    mgt=ipmi
    monserver=c910f04x18
    netboot=xnba
    nfsserver=c910f04x18
    os=rhels6.7
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles,xHRM bridgeprereq virbr0
    primarynic=mac
    profile=kvm
    provmethod=rhels6.7-x86_64-install-kvm
    serialport=0
    serialspeed=115200
    status=installing
    statustime=08-17-2015 03:22:13
    tftpserver=c910f04x18
    updatestatus=synced
    updatestatustime=08-17-2015 02:47:55
    xcatmaster=c910f04x18

[root@c910f04x18 tmp]# lsdef -t osimage -o rhels6.7-x86_64-install-kvm
Object name: rhels6.7-x86_64-install-kvm
    imagetype=linux
    osarch=x86_64
    osdistroname=rhels6.7-x86_64
    osname=Linux
    osvers=rhels6.7
    otherpkgdir=/install/post/otherpkgs/rhels6.7/x86_64
    pkgdir=/install/rhels6.7/x86_64
    pkglist=/opt/xcat/share/xcat/install/rh/kvm.rhels6.pkglist
    profile=kvm
    provmethod=install
[...]

PR #87, opened 2015-08-18
fix defect [DEV] Error: either "dev" is duplicate, or "virbr0" is a garbage." message on the 1st reboot after kvmhost installation #86

immarvin:
see https://github.com/xcat2/xcat-core/issues/86 for details


immarvin:
corrected some code logic and indent problem

daniceexi:
I am OK with the change.

PR #88, opened 2015-08-18
support rflash for firestone machine, need to verify with Firestone m…

zet809:
…achine

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -31,6 +31,7 @@ use xCAT::Usage;
 use Thread qw(yield);
 use LWP 5.64;
 use HTTP::Request::Common;
+use Time::HiRes qw/time/;
 my $iem_support;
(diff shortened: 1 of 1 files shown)


whowutwut:
Should this change wait until it's verified with the Firestone machine before it's merged into the xcat2?

daniceexi:
@whowutwut 
1. The code is independent with other features, so it's safe to check in even without real test.
2. We will try to wait for the ready of JMET machine, but if it still cannot ready before we make all the code review done, we will check in the code otherwise this task will hang.

Generally for this kind of case that task depends on other stuff, I'd like to check in code first to make a milestone so that people could put effort on other task. Of course that check in should not affect ohter functionalities.

daniceexi:
Has done the first round of review with Er Tao.

whowutwut:
@daniceexi, ok sure. Sounds good

zet809:
For the comments, I have checked in the modifications.

daniceexi:
I have reviewed pull request and agree to check in.

PR #89, opened 2015-08-18
update buildcore.sh and builddep.sh to upload the build to xcat.org

tingtli:


buildcore.sh
```diff
@@ -128,7 +129,8 @@ if [ "$REL" = "xcat-core" ]; then    # using git
 fi
 
 YUMDIR=$FRS
-YUMREPOURL="https://sourceforge.net/projects/xcat/files/yum"
+#YUMREPOURL="https://sourceforge.net/projects/xcat/files/yum"
+YUMREPOURL="https://xcat.org/files/yum"
```
arif-ali: I can't get to the https, should this be http

Is there an easy way to mirror this site at all?

daniceexi: Arif, What's the issue that you cannot access xcat.org with https? To satisfy the security requirement from different user, we must use https as default.

For the creating mirror, is that enough that xcat put all files in /var/www/xcat.org/files and you can download them.

arif-ali: I get the following, when accessing the https page

```
The page you were looking for doesn't exist.

You may have mistyped the address or the page may have moved.
```

I tend to have a local mirror, so that I can do a quicker xCAT install in-house. At the moment, as I have contributer access to the sourceforge, I am able to use rsync to do that. I was just wondering how we would go about that once the migration to xcat.org is done

whowutwut: Before we make this switch, need to address this invalid certificate. 

image

The certificate being used for xcat.org is:
* for domain www.autoappmaker.com
* and has expired... 

image

whowutwut:
I think there are issues with xcat.org that we need to address by means of a trial period or "opt-in"  period where we have both SF and xcat.org running at the same time.  Then we should ask users to give xcat.org a try first before we flip off the switch. 

The issues that Arif posted is exactly to avoid causing headaches to existing users.  Also with the invalid SSL certificate, it just looks bad.  

Ill add some comments inline for the code changes regarding this..

buildcore.sh
```diff
@@ -41,7 +41,8 @@
 
 # you can change this if you need to
 UPLOADUSER=litingt
-FRS=/home/frs/project/x/xc/xcat
+#FRS=/home/frs/project/x/xc/xcat
```
whowutwut: I really don't like the commenting out of the code pieces to leave in the old code. This causes bloat and hard to read.  I would have much preferred to see the line modified. 

However, I started a discussion in the pull request about having both SF and xcat.org running at the same time.  In order to do this we can't just comment out....

I would suggest something a new environment variable to control the build... 
SOURCEFORGE=1  <== default to SF

Then have a function that will set the common variables based on SOURCEFORGE=1 or =0

buildcore.sh
```diff
@@ -128,7 +129,8 @@ if [ "$REL" = "xcat-core" ]; then    # using git
 fi
 
 YUMDIR=$FRS
-YUMREPOURL="https://sourceforge.net/projects/xcat/files/yum"
+#YUMREPOURL="https://sourceforge.net/projects/xcat/files/yum"
```
whowutwut: For this, could use: 
```
TARGET_MACHINE="https://xcat.org" 
TARGET_PATH="/files/yum"

if [[ SOURCEFORGE = "1" ]]; then
    TARGET_MACHINE="https://sourceforge.net"
    TARGET_PATH="/project/xcat/files/yum"
fi
```

buildcore.sh
```diff
@@ -547,14 +549,16 @@ fi
 if [ "$REL" = "devel" -o "$PREGA" != 1 ]; then
     i=0
     echo "Uploading RPMs from $CORE to $YUMDIR/$YUM/$REL$EMBEDDIR/ ..."
-    while [ $((i+=1)) -le 5 ] && ! rsync -urLv --delete $CORE $UPLOADUSER,xcat@web.sourceforge.net:$YUMDIR/$YUM/$REL$EMBEDDIR/
+#    while [ $((i+=1)) -le 5 ] && ! rsync -urLv --delete $CORE $UPLOADUSER,xcat@web.sourceforge.net:$YUMDIR/$YUM/$REL$EMBEDDIR/
+     while [ $((i+=1)) -le 5 ] && ! rsync -urLv --delete $CORE xcat@xcat.org:$YUMDIR/$YUM/$REL$EMBEDDIR/
```
whowutwut: Same idea here  ```web.sourceforge.net``` and ``xcat.org`` should be a variable so we could control the upload destination

same comment applies below to the similar code

buildcore.sh
neo954: The web site hosted on tcp port 443 is another apache virtual host. It is no related with xcat.org.

whowutwut: Right, that's because this server that is hosting xcat.org is Shared Web Hosting.  We would need to apply for a "veri-signed" SSL certificate (non self signed) and configure the certificate in the xcat.org httpd config file.  

Right now the certificate that's being used when we hit the https port on xcat.org is the certificate for autoappmaker.com.  Then we get a domain mismatch.

build-ubunturepo
```diff
@@ -11,18 +11,18 @@
 #    uploaded to in sourceforge (e.g. devel, or 2.3).
 #  - You probably also want to put root's pub key from the build machine onto sourceforge for
 #    the upload user listed below, so you don't have to keep entering pw's.  You can do this
-#    at https://sourceforge.net/account/ssh
+#    at https://xcat.org/setting/ssh
```
whowutwut: This set of instructions does not apply anymore for xcat.org. 
There's no option to upload the ssh keys

buildcore.sh
```diff
@@ -6,10 +6,10 @@
 # Getting Started:
 #  - Check out the xcat-core svn repository (either the trunk or a branch) into
```
whowutwut: Are these steps still valid?  Seems like we are no longer use svn repository.  Also do we need to check out into a dir <rel>/src/xcat-core?

buildcore.sh
```diff
@@ -6,10 +6,10 @@
...
 #    a dir called <rel>/src/xcat-core, where <rel> is the same as the release dir it will be
-#    uploaded to in sourceforge (e.g. devel, or 2.3).
-#  - You probably also want to put root's pub key from the build machine onto sourceforge for
+#    uploaded to in xcat.org (e.g. devel, or 2.3).
+#  - You probably also want to put root's pub key from the build machine onto xcat.org for
```
whowutwut: We should delete line 10-12, does not apply anymore.

buildcore.sh
```diff
@@ -22,12 +22,12 @@
 # Usage:  buildcore.sh [attr=value attr=value ...]
 #    Before running buildcore.sh, you must change the local git repo to the branch you want built, using: git checkout <branch>
 #        PROMOTE=1 - if the attribute "PROMOTE" is specified, means an official dot release.  This does not actually build
-#                    xcat, just uploads the most recent snap build to https://sourceforge.net/projects/xcat/files/xcat/ .
-#                    If not specified, a snap build is assumed, which uploads to https://sourceforge.net/projects/xcat/files/yum/
-#                    or https://sourceforge.net/projects/xcat/files/aix/.
+#                    xcat, just uploads the most recent snap build to http://xcat.org/files/xcat/ .
+#                    If not specified, a snap build is assumed, which uploads to http://xcat.org/files/yum/
+#                    or https//xcat.org/files/aix/.
```
whowutwut: should not be https

buildcore.sh
```diff
@@ -547,51 +550,57 @@ fi
...
+     while [ $((i+=1)) -le 5 ] && ! rsync -urLv --delete $CORE $USER@$TARGET_MACHINE:$YUMDIR/$YUM/$REL$EMBEDDIR/
     do : ; done
 fi
 
-# Upload the individual source RPMs to sourceforge
+# Upload the individual source RPMs to xcat.org 
 i=0
 echo "Uploading src RPMs from $SRCD to $YUMDIR/$YUM/$REL$EMBEDDIR/ ..."
-while [ $((i+=1)) -le 5 ] && ! rsync -urLv --delete $SRCD $UPLOADUSER,xcat@web.sourceforge.net:$YUMDIR/$YUM/$REL$EMBEDDIR/
+ while [ $((i+=1)) -le 5 ] && ! rsync -urLv --delete $SRCD $USER@$TARGET_MACHINE:$YUMDIR/$YUM/$REL$EMBEDDIR/
```
whowutwut: I think extra space here in front of "while"

builddep.sh
```diff
@@ -8,7 +8,7 @@
 #  - You need to install gsa-client on the build machine.
 #  - You probably want to put root's pub key from the build machine onto sourceforge for
```
whowutwut: Delete line 8-11, does not apply anymore

builddep.sh
```diff
@@ -20,8 +20,10 @@
 
 # you can change this if you need to
 UPLOADUSER=litingt
```
whowutwut: This variable UPLOADUSER is no longer being used.. can delete

builddep.sh
```diff
@@ -43,8 +45,10 @@ else
 fi
 
 # this is needed only when we are transitioning the yum over to frs
-YUMREPOURL1="http://xcat.sourceforge.net/yum"
-YUMREPOURL2="https://sourceforge.net/projects/xcat/files/yum"
+#YUMREPOURL1="http://xcat.sourceforge.net/yum"
```
whowutwut: Should we delete the old references instead of comment out.

builddep.sh
```diff
@@ -252,13 +256,17 @@ else
 	links="-l"
 fi
 echo "Uploading RPMs from xcat-dep to $YUMDIR/$YUM/ ..."
-while [ $((i+=1)) -le 5 ] && ! rsync $links -ruv --delete xcat-dep $UPLOADUSER,xcat@web.sourceforge.net:$YUMDIR/$YUM/
+#while [ $((i+=1)) -le 5 ] && ! rsync $links -ruv --delete xcat-dep $UPLOADUSER,xcat@web.sourceforge.net:$YUMDIR/$YUM/
```
whowutwut: Delete old code instead of comment out?

builddep.sh
```diff
@@ -252,13 +256,17 @@ else
...
+#while [ $((i+=1)) -le 5 ] && ! rsync $links -ruv --delete xcat-dep xcat@xcat.org:$YUMDIR/$YUM/
+while [ $((i+=1)) -le 5 ] && ! rsync $links -ruv --delete xcat-dep $USER@$TARGET_MACHINE:$YUMDIR/$YUM/
 do : ; done
 
 # Upload the tarball to the SF FRS Area
 i=0
 echo "Uploading $DFNAME to $FRS/xcat-dep/$FRSDIR/ ..."
-while [ $((i+=1)) -le 5 ] && ! rsync -v $DFNAME $UPLOADUSER,xcat@web.sourceforge.net:$FRS/xcat-dep/$FRSDIR/
+#while [ $((i+=1)) -le 5 ] && ! rsync -v $DFNAME $UPLOADUSER,xcat@web.sourceforge.net:$FRS/xcat-dep/$FRSDIR/
```
whowutwut: Delete old code instead of comment out?

whowutwut:
@tingtli, Thanks! :+1:   Merging the pull request..

PR #91, opened 2015-08-18
allow multiple valid access tokens to exist at the same time for a user

bybai:
Currently, xcat do not keep several tokens for the same user in token table.If the user is the same, when you get a new token, the old one is deleted from token table.
After code is changed, it allows multiple valid access tokens to exist at the same time for a user, at the same time it removes all expire token.


daniceexi:
I have reviewed the code and agree to check in.

PR #105, opened 2015-08-20
fix : updatestatus is not correct when -F run with -P and -S

bybai:
The issue:
Updatastatus is not correct happens when there is no syncfiles.
When "updatenode CN -F" and there is no syncfile, updatestatus is set to "synced" in code directly. 
When "updatenode CN" or updatenode using -P,-F,-S at the same time, -F is first ran, its hard code change updatastatus as "synced", then, the initial state for -P or -S becomes "synced" but not "syncing",  all -P or -S procedure keeping synced updatestatus until it get the final updatestatus.
 
My fix:
1, Do not use hard code state "synced", when -F and there is no syncfile, set $::NOSYNCFILE=1;
2, When there is no -P -S and no syncfile ($::NOSYNCFILE=1), final state for updatestatus is synced.
3, If there is -P -S -F, run as before, updatestatus get final state as before;


daniceexi:
I reviewed the code and agree to merge in.

Issue #109, opened 2015-08-20
setupntp has problem to run in sh -x


daniceexi:
Today we found an issue that when run setupntp postscript with '-x' option, the code in following part would fail. The root cause was when run with '-x', the verbose message like '+++/usr/sbin/rcntp ntptimeset .....' will be output to 'error output', that means get into $output. Then pmatch in the line 124 failed. To fix this problem, we can add wide char * at beginning of "Time could not*". Any thought?
```
122 output=eval $cmd 2>&1
123 rc=$?
124 if [ "$rc" != "0" ] || (pmatch "$output" "Time could not*");then
125 echo "$cmd failed, NTP maynot be synchronized, please wait for NTP synchronized then run updatenode nodename -P setupntp"
126 logger -t xcat "$cmd failed"
127 exit 1
```

PR #111, opened 2015-08-20
Add wide char * todo the pmatch of output

cxhong:
this is the fix for #109


daniceexi:
@cxhong 

Did you verified this fix?
And could you talk a look is there any samilar issue in the other postscripts?

cxhong:
yeh, I verified the fix.
I took a look in the other postscripts, didn't see any obvious issues.

daniceexi:
Ok, I reviewed this change and agree to check in.

PR #112, opened 2015-08-20
Moving the xcat documentation into the xcat-core project

whowutwut:
The xcat-doc was initially a proof of concept to try and figure out new documentation and to utilitze ReadTheDocs integration with GitHub.  

We did not think that xcat-core would be moved from SF to GitHub in the near future so we didn't put the documentation as a part of xcat-core.  Now that xcat-core is no longer being mirrored, moving the docs under xcat-core makes sense.  

The xcat-doc project would eventually be removed.


whowutwut:
Do we still need to move over the doc_outline file?

whowutwut:
@ligc  @daniceexi  
Although we talked about this at last nights meeting, I want to give a chance for comments before this is merged.  I'll need to update the pointer for the readTheDocs once we merge this to xcat-core.

whowutwut:
Added the RTD webhook and removed the xcat-doc Repository from the 3 defined Teams.

daniceexi:
@whowutwut 

I think the action will be:

1. move all the docs directory from xcat-doc to xcat-core
2. point readTheDcos to xcat-core/docs

I am good with the merge. Could you make them done in sametime so that other members can start the doc change next week.

BTW, I am ok that don't move doc_outline file.

whowutwut:
Ok, pointed the RTD to xcat-core

image

When viewing the docs at http://xcat-docs.readthedocs.org/en/latest/, clicking on "Edit on GitHub" redirects to https://github.com/xcat2/xcat-core/blob/master/docs/source/index.rst.  

Goes to xcat-core project, looks good.

PR #116, opened 2015-08-24
xCAT provision Sles11.2 will hang-on when excuting remoteshell script

SStar1314:
xCAT provision Sles11.2 will hang-on and wait for input passphrase when executing "ssh-keygen -y -f /etc/ssh/ssh_host_ecdsa_key > /etc/ssh/ssh_host_ecdsa_key.pub" command in remoteshell script.

Root Cause: Sles11.2 install openssh-5.1p1-41.57.1 build-in package, and this version openssh don't support ecdsa key type.
So there needs a openssh support check before ecdsa key generation. In remoteshell script, line 283, we will add "ssh-keygen -t ecdsa -y -f /etc/ssh/ssh_host_ecdsa_key -P "" " command and check the result to judge support or not.

### DIFF
--- a/xCAT/postscripts/remoteshell
+++ b/xCAT/postscripts/remoteshell
@@ -281,8 +281,7 @@ if [ -f /etc/ssh/ssh_host_ecdsa_key ]; then
 	if ! grep "PRIVATE KEY" /etc/ssh/ssh_host_ecdsa_key > /dev/null 2>&1 ; then
    		rm /etc/ssh/ssh_host_ecdsa_key
 	else
-		ssh-keygen -t ecdsa -y -f /etc/ssh/ssh_host_ecdsa_key -P ""
-		if [ "x$?" = "x0" ]; then
+		if ssh-keygen -t ecdsa -y -f /etc/ssh/ssh_host_ecdsa_key -P "" &>/dev/null ; then
 			ssh-keygen -y -f /etc/ssh/ssh_host_ecdsa_key > /etc/ssh/ssh_host_ecdsa_key.pub
 			chmod 644 /etc/ssh/ssh_host_ecdsa_key.pub
 			chown root /etc/ssh/ssh_host_ecdsa_key.pub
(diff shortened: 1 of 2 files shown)


xCAT/postscripts/remoteshell
```diff
@@ -281,9 +281,14 @@ if [ -f /etc/ssh/ssh_host_ecdsa_key ]; then
 	if ! grep "PRIVATE KEY" /etc/ssh/ssh_host_ecdsa_key > /dev/null 2>&1 ; then
    		rm /etc/ssh/ssh_host_ecdsa_key
 	else
-		ssh-keygen -y -f /etc/ssh/ssh_host_ecdsa_key > /etc/ssh/ssh_host_ecdsa_key.pub
-		chmod 644 /etc/ssh/ssh_host_ecdsa_key.pub
-		chown root /etc/ssh/ssh_host_ecdsa_key.pub
+		ssh-keygen -t ecdsa -y -f /etc/ssh/ssh_host_ecdsa_key -P ""
+		if [ "x$?" = "x0" ]; then
```
samveen: I suggest that you collapse the `ssh-keygen` and `if` into a single line like this:
```
if ssh-keygen -t ecdsa -y -f /etc/ssh/ssh_host_ecdsa_key -P "" &>/dev/null ; then
```
Interesting read: http://www.pixelbeat.org/programming/shell_script_mistakes.html

SStar1314: Thanks for your suggestion. The interesting read link is very useful, it covers many common shortage usages and offers corresponding solutions. And it will help me a lot, I appreciate it.
By the way, I will re-pull a merge request without the redundantly "[ ]". Need your check again. 
I am a new git-user, if there is some inappropriate behaviors, please let me know and I will correct myself.

SStar1314:
Imagine MN have system A and install version AA OpenSSH, AA support ecdsa key encryption. While CN will install system B with version BB OpenSSH installed.
When

daniceexi:
@SStar1314 thanks for the detail information. Could add some comments in the code to describe this change?

daniceexi:
I reveiwed the change and agree to check in.

PR #117, opened 2015-08-24
Remove the license CLA from the docs, using xcat-core/CONTRIBUTORS.md

whowutwut:
The license files that were under Developers in the documentation is no longer needed since we have placed the contributor process under the root xcat-core directory.


daniceexi:
@whowutwut 

Where is the entry to manage the agreement for a new user after removing the pages from RTD->Developer?

whowutwut:
The page is here: https://github.com/xcat2/xcat-core/blob/master/CONTRIBUTING.md
I just committed a change to README -> README.md and added a link to the Contributors page. 

If we want, we can add a pointer back to the CONTRIBUTORS.md page from ReadTheDocs.  We should just have this in one place

Did you like it better in RTD?

daniceexi:
@whowutwut 

I sent out an email to discuss the scope of xcat.org, readme, wiki and RTD.

I think we don't put any essential content in readme.  
From my point of view, the pages tell a user how to be a contributor of xCAT is a user guide, so should put it in RTD. Stay it in RTD->Develpers part is good for me.

whowutwut:
@daniceexi  Fair point.  I am going to close this pull request and I'm generating another pull request that will remove the CLAs from the README files and only have them in the docs/source/developers/license area

PR #120, opened 2015-08-25
Update README.md

arif-ali:
To make it correct with markdown, The 4 ~ were for the sourceforge version of documentation


whowutwut:
Thanks. Looks good

PR #122, opened 2015-08-26
The CLAs should stay in the xCAT documentation on RTD

whowutwut:
The CLAs should stay in the xCAT documentation instead of being put in to the xcat-core source pages. 

Removing those files and cleaning up the agreements under the docs directory. In reference to discussions in pull request #117


daniceexi:
The change is good to me!

daniceexi:
@whowutwut 
One question here that the agreement for contributor is a web page, bu the agreement for mainainter is a pdf file for download. Why they are different?

whowutwut:
@daniceexi , I don't know, I just copied what was there on SF. :-)   We can generate a PDF and have them download as PDF, I did notice that from RTD, doing a print of the page then saving as PDF file was pretty clean (no table of contents on the side)

daniceexi:
@whowutwut I did not try the print of the CLA, if the print page is not clear for user to sign the agreement, we should make it to be a downloadable pdf file.

whowutwut:
Just checked....
* chrome prints very clean from RTD, no headers/footers
* firefox has the URL and timestamps in the header/footers, but is fairly clean too
* IE prints out very ugly with the RTD table of contents. 

If we agree, I could leave the .rst files there but just hidden so that we can modify them if needed and re-generate PDF files. 

Then have the CLA and CCLA be downloaded PDF files.

daniceexi:
@whowutwut 

Keep .rst file for CLA is OK for me! And thanks for the downloadable pdf file.

PR #123, opened 2015-08-27
[feature-req][WIP]update configbond script to us ethe nicextraparams

arif-ali:
As in feature request number [191](https://sourceforge.net/p/xcat/feature-requests/191/ ) on sourceforge.

> Arif, do you see a need to honor the extra parameters for configbond? Can you give us an example of such parameters? Thanks.

We have used it significantly in at least 2 of our customer sites, where we have multiple VLANs on the bonds, and need xCAT to configure them, rather than having to do them manually. I am happy to work on this script further to get it fine tuned

/cc @linggao @ligc


daniceexi:
@arif-ali Thanks for your contribution, @bybai will review your change.

bybai:
@arif-ali I found your another pull request for configbridge,  in configbridge, you do something about bond, how do you define the differentiate between configbridge, configbond and configeth. 
I think something is overlap in them.

whowutwut:
We are building Release candidate for 2.11 and only working on bug fixes to preserve stability.. moving this to 2.12 release.

arif-ali:
#469 will do this instead

PR #125, opened 2015-08-27
Create pdf files of the agreements for the contributor license agreem…

whowutwut:
…ents.

To be consistent with the maintainer agreement, provide downloadble pdf
files for contributors to print and sign.  This makes the form that is
returned to us more consistent as well.

Discussed with @daniceexi in pull request #122


daniceexi:
@whowutwut The change looks good, thanks.

PR #128, opened 2015-08-28
use pigz to improve the performance of genimage/packimage

immarvin:
1. correct some indent problem in ubuntu/genimage 
2. when "pigz" is available on the management node, "genimage"  on ubuntu and "packimage" will use "pigz" instead of "gzip" as the compress tool  
3. when "pigz" is available in the chrooted rootimg generated by "genimage", "genimage" will use "pigz" as the dracut compress tool. The process number of pigz will be the number of online cpus on the server which performs the "genimage" operation.
4. for “genimage” on redhat and sles, the "pigz" feature is only supported on the os releases which use dracut to generated the initrd, i.e, redhat 6 or above, sles12 or above
5. there will be a document on how to enable the "pigz" in genimage and packimage

### DIFF
--- a/xCAT-server/lib/xcat/plugins/packimage.pm
+++ b/xCAT-server/lib/xcat/plugins/packimage.pm
@@ -366,17 +366,27 @@ sub process_request {
     $callback->({data=>["$verb contents of $rootimg_dir"]});
     unlink("$destdir/rootimg.gz");
     unlink("$destdir/rootimg.sfs");
+
+    my $compress="gzip";
+    #use "pigz" as the compress tool instead of gzip if "pigz" exist
+    my $ispigz=system("bash -c 'type -p pigz'");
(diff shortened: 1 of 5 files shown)


chrissamuel:
Hi there,

Thanks for this work, I notice that this code also needs to update the statelite.pm file to get "liteimg" to use pigz as well - we use statelite an awful lot. :-)

thanks!
Chris

immarvin:
@chrissamuel  the code for liteimg is ready, see https://github.com/immarvin/xcat-core/commit/81409e1899b3440b4af89157a563b2eb11ba055a

immarvin:
@daniceexi fixed the problems according to your comments

daniceexi:
I have reviewed the code change and agree to check in.

chrissamuel:
@immarvin ahh thanks!

Issue #114, opened 2015-08-24
[fvt]2.11:copycds failed using the latest build


junxiawang:
How to reproduce:
1.down load the latest build
2.c910f03c05k06:/ # lsdef -v
lsdef - Version 2.11 (git commit 4058f6b60af47a849ffaf36eda730ba3ca4bb949, built Mon Aug 24 04:34:29 EDT 2015)
3.c910f03c05k06:/ # copycds SLE-12-Server-DVD-ppc64le-GM-DVD1.iso 
Copying media to /install/sles12/ppc64le/1
Error: A fatal error was encountered, the following information may help identify a bug: Can't use string ("Copying media to /install/sles12"...) as an ARRAY ref while "strict refs" in use at /opt/xcat/sbin/xcatd line 3022.
ERROR/WARNING: communication with the xCAT server seems to have been ended prematurely

hu-weihua:
This issue was caused by parsing copycds response format error.

Issue #127, opened 2015-08-28
[For Task 52219] record response part into /var/log/xcat/commands.log failed when $rsponse isn't ARRAY


hu-weihua:
For function cmdlog_collectlog() in xcatd. when attribute $rsponse isn't ARRAY,  such kind of response will be missed in /var/log/xcat/commands.log

hu-weihua:
fixed in 2.11

PR #129, opened 2015-08-28
Make command log process support restartxcatd

hu-weihua:
The changing code belongs to story "Provide a way to make the xCAT offline trouble shooting be easier".

The code checked in this time covers below things:
1. make command log process support restartxcatd command
2. fix bug #114 and #127 
3. refine the format of /vat/log/xcat/commands.log


daniceexi:
I have reviewed the code change and agreed to check in.

PR #133, opened 2015-08-29
add doc for accelerating genimage and packimage with pigz

immarvin:


immarvin:
polished the doc according to the comments

docs/source/guides/admin-guides/manage_clusters/common/deployment/acc_initrd_rootimg_gen.rst
```diff
@@ -0,0 +1,56 @@
+Accelerating the diskless initrd and rootimg generating
+========================================================
+
+Generating diskless initrd with ``genimage`` and compressed rootimg with ``packimage`` and ``liteimg`` is a time-comsuming process, it can be accelerated by enabling paralell compression tool ``pigz`` on the management node with multiple processors and cores.
```
whowutwut: I think here we are saying that optionally, the user can choose to use a newer compression tool that currently not yet provided by the OS?  Once pigz is provided by the OS, there isn't any steps needed to install it right?   Should we describe the current process? 

...time-consuming process which currently uses gzip.  By using parallel compression tools such as ``pigz``, the time required to generate the images can be reduced.

docs/source/guides/admin-guides/manage_clusters/common/deployment/acc_initrd_rootimg_gen.rst
```diff
@@ -0,0 +1,56 @@
...
+
+The paralell compression tool ``pigz`` can be enabled by installing ``pigz`` package on the management server or diskless rootimg. Depending on the method of generating the initrd and compressed rootimg, the steps differ in different Linux distributions. 
+
+
+Enabling the ``pigz`` on :ref:`Ubuntu Server LTS<ubuntu-os-support-label>`
+--------------------------------------------------------------------------
+
+Make sure the ``pigz`` is installed on the management node with the following command::
+
+   # dpkg -l|grep pigz
```
whowutwut: The # sign is being interpreted as a comment.  So it's grayed out.  Suggest removing the # in the code blocks for all the code blocks below in this file

whowutwut:
@immarvin  Overall i think this document is great. Its more describing how to install pigz so that we can use the compression.   Is the implementation seamless where the user does not know the compression, if pigz is installed, we will automatically use the parallel compression? 

Could we show some examples maybe run through ``time`` to show the performance enhancement of using the parallel compression over the std compression?

whowutwut:
@immarvin 

I think this latest commit that you made in this pull request: 5e30e55 is due to working in your immarvin:master branch.  I was waiting for a response from the last comment I made before I merged the pull requst, sorry...Since you created this pull request from the 'master' branch, when you refresh our master to work on something else, the merge automatically will go into this pull request. 

This is why we need to work on branches that we create under our own fork of xcat-core so we can leave the changes alone until the pull request is reviewed and merged. 

I'm going to close , can you create a new pull request with the 3 original files?

PR #135, opened 2015-08-31
add parallel commands doc and software firmware inventory doc 

bybai:
Discussed with team:
1, firmware_inventory.rst should be in advanced topics;
2, "parallel commands" page is null, so parallel commands link is empty.

### DIFF
--- a/docs/source/advanced/firmware_inventory.rst
+++ b/docs/source/advanced/firmware_inventory.rst
@@ -1,7 +1,7 @@
 Software and Firmware Inventory
 ===============================
 
-xCAT provides a command '**sinv'** that checks the software and firmware configuration in this cluster.
+xCAT provides a command **sinv** that checks the software and firmware inventory in this cluster.
 
 The command creates an inventory of the input software/firmware check, comparing to other machines in the cluster and produces an output of node that are installed the same and those that are not.
 
--- a/docs/source/advanced/firmware_inventory.rst
+++ b/docs/source/advanced/firmware_inventory.rst
@@ -1,7 +1,7 @@
 Software and Firmware Inventory
 ===============================
 
-xCAT provides a command **sinv** that checks the software and firmware inventory in this cluster.
+xCAT provides a command ``sinv`` that checks the software and firmware inventory in this cluster.
 
 The command creates an inventory of the input software/firmware check, comparing to other machines in the cluster and produces an output of node that are installed the same and those that are not.
 
--- a/docs/source/advanced/firmware_inventory.rst
+++ b/docs/source/advanced/firmware_inventory.rst
@@ -1,0 +1,41 @@
+Software and Firmware Inventory
+===============================
+
+xCAT provides a command '**sinv'** that checks the software and firmware configuration in this cluster.
+
+The command creates an inventory of the input software/firmware check, comparing to other machines in the cluster and produces an output of node that are installed the same and those that are not.
+
+This command uses the xdsh parallel command, so it is in itself a parallel command, and thus can be run on multiple cluster nodes at one time and is hierarchical.
+
+The sinv command is designed to check the configuration of the nodes in a cluster. The command takes as input command line flags, and one or more templates which will be compared against the output of the xdsh command, designated to be run on the nodes in the noderange.
+
+The nodes will then be grouped according to the template they match and a report returned to the administrator in the output file designated or to stdout.
+
+sinv supports checking the output from the rinv or xdsh command.
+
+For example, if you wanted to check the ssh level on all the nodes and make sure they were the same as on the service node, you would first generate a template from the "good" service node (sn1) by running the following: ::
+
+    xdsh sn1 "rpm -qa | grep ssh " | xdshcoll > /tmp/sinv/sinv.template
+
+To execute sinv using the sinv.template generated above on the nodegroup, testnodes ,writing output report to /tmp/sinv.output, enter: ::
+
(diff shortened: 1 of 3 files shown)


bybai:
parallel commands :http://10.5.106.1/install/pll/xcat-core/docs/build/html/guides/admin-guides/manage_clusters/ppc64le/parallel_cmd.html
firmware_inventory: http://10.5.106.1/install/pll/xcat-core/docs/build/html/advanced/firmware_inventory.html

whowutwut:
@bybai , the 10. IP addresses are internal IPs for our dev machines.   It's not really something that the public can access.  I think the command line instructions should allow developers to pull the code from the pull request into a separate branch and then do a 'make html" to view the changes.

docs/source/advanced/firmware_inventory.rst
```diff
@@ -0,0 +1,41 @@
+Software and Firmware Inventory
+===============================
+
+xCAT provides a command '**sinv'** that checks the software and firmware configuration in this cluster.
```
whowutwut: suggest using two back ticks around sinv to indicate command

whowutwut: Change to "...checks the software and firmware inventory of the cluster" since the command is sinv - for inventory.

docs/source/advanced/firmware_inventory.rst
```diff
@@ -0,0 +1,41 @@
...
+
+The command creates an inventory of the input software/firmware check, comparing to other machines in the cluster and produces an output of node that are installed the same and those that are not.
```
whowutwut: This documentation is very hard to understand, it was taken from http://sourceforge.net/p/xcat/wiki/Parallel_Commands_and_Inventory/#software-and-firmware-inventory and I think should be re-written to clarify the content..

docs/source/advanced/firmware_inventory.rst
```diff
@@ -0,0 +1,41 @@
...
+
+    xdsh sn1 "rpm -qa | grep ssh " | xdshcoll > /tmp/sinv/sinv.template
+
+To execute sinv using the sinv.template generated above on the nodegroup, testnodes ,writing output report to /tmp/sinv.output, enter: ::
+
+    sinv  -c "xdsh testnodes rpm -qa | grep ssh" -p /tmp/sinv/sinv.template -o /tmp/sinv.output
+
+The report will look something like this, if every node matches:
+
+   Command started with following input: ::
```
whowutwut: I'm not sure what this following example is showing us.  The report doesn't seem to show much useful information.. does it?

docs/source/advanced/firmware_inventory.rst
daniceexi: @whowutwut This part of doc describes something that is useful in big cluster.  And I did not see anyone talk of using this feature. So I don't want us to put effort to write this doc. We can leave it as is or do not add it in our new doc.

samveen:
I suggest that the documentation that is unclear or obscure be added to the wiki section of the repository instead of the docs. Being in the wiki here will allow for users to contribute cleanups and improvements inline, instead of needing to follow the Github contribution model. Once a page has reached a certain level of improvement, it can be moved from the wiki to the docs folder.

daniceexi:
@samveen 
We are good with the content, just discuss the format/structure. Discussion will help to make the doc to be better at the first post.

whowutwut:
@samveen.... interesting point you bring up. If I understand correctly, you are suggesting that we should use the GitHub wiki as a staging area where multiple people can edit and once it's ready, pulled into the official docs? 

I do like this idea... but there are some small issues i see: 
* The GitHub wiki is an enhanced flavor of markdown, we didn't want to tie documentation to wiki pages again after this difficulty of moving docs from SF wiki to GitHub.  
* our new doc is written in restructured text using sphinx to build and generate pages, so the markdown in wiki may or may not render correctly as .rst 

I guess the problem is that we have lots of documentation to cover and it's quickest to divide and conquer. However, the pull request  process does add a lot of overhead because it is easier to fix small text directly instead of making a comment.  But I think this will all get better as the team familiarizes with Sphinx syntax. 

I am firm on the point that we should not directly copy the content from SF to GitHub just in a different ordering of the docs, there should be at least a minimum effort to see whether the documentation can be created more clearly with the Sphinx and at least is correct and flows with the rest of the documentation being added.  .....   Perhaps order does matter, meaning not to add pages without previous pages already there??

bybai:
@whowutwut  
1, firmware_inventory.rst  is for useful in big cluster. I think we can leave it as it was in advanced topics just as @daniceexi  described. It is no need to pay more effort to it. 
2, I used 10.xx ip as an example link just for more team members to view it. We can set tunnel to 10.xx ip range. It is more convenient for someone does not want to git clone the branch. 
3. "parallel commands" is a simple overview introduction page, here, I list the examples of  the most common function for parallel commands xdsh and xdcp.  They can be as a quick start for user. Other parallel commands can be found in "xcat commands" page.

samveen:
@whowutwut Yes, Victor, that was exactly what i was suggesting, as a wiki is so much easier to contribute to, as compared to a Github project. 
* I completely agree that it is a headache to work with different formats, but one possible solution is to use `pandoc` to convert the github markdown to reST and then move it into docs
* As above, followed by a quick crosscheck.

However, I suggest this only for that documentation if the team starts getting bogged down with documentation related tasks instead of xCAT itself :)

daniceexi:

PR #139, opened 2015-09-01
Add links to the developers guide in the README

samveen:
Added direct links for to developers guide for new developers and perspective contributors (given that the people who visit Github are looking to contribute)

### DIFF
--- a/README
+++ b/README
@@ -8,7 +8,7 @@ xCAT is made available as open source software under the EPL license:
 http://www.opensource.org/licenses/eclipse-1.0.php
 
 
-
-
-
-
+Developers and prospective contributors are encouraged to read the developers guide:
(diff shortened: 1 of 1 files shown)


whowutwut:
Thanks for this, on a previous pull request that we decided not to merge, I had made a change to make this README be a markdown file...  (ie. rename README => README.md)  .  The purpose was that the URLs could then be a clickable link.  

@daniceexi   Should we make this file type change? 
@samveen   If we agree, do you want to make that change with this pull request, if not, I can do it, no big deal.

samveen:
@whowutwut From what I can make out, links in project READMEs are parsed for display anyway, so I see no difference either way. Eg. https://github.com/samveen/xcat-core/tree/modify_readme

whowutwut:
Ok yea I see that. I'll just merge as is

PR #142, opened 2015-09-01
Fix postscripts

samveen:
The postscripts for debian, ubuntu and esx missed the fix for using the parameterizing interface name in the dhclient invocation (using a hardcoded eth0 instead of the introspected/passed parameter).

These correspond to the change made for the RH postscript in 0dba62260c995c824bb26bad9f8686ee711e963f


samveen:
Closing this as it contains changes from another pull request as well (and reopening another from correct base)

PR #145, opened 2015-09-02
fix bug#140 offline trouble shooting time stamp wrong

hu-weihua:
fix bug#140 offline trouble shooting time stamp wrong


daniceexi:

PR #148, opened 2015-09-02
Add hierarchy doc

chenglch:
Add hierarchy documents including introduction, database
management,defining node, service node setup, test service
node and appendix.

TODO: resolve the fake_todo link reference in the following patch

### DIFF
--- a/docs/source/advanced/hierarchy/appendix_a_setup_backup_service_nodes.rst
+++ b/docs/source/advanced/hierarchy/appendix_a_setup_backup_service_nodes.rst
@@ -83,7 +83,7 @@ rhsn. lsdef cn4 | grep servicenode. ::
 
   servicenode=service1,rhsn
 
-f a service node is offline ( e.g. service1), then you will see errors on
+If a service node is offline ( e.g. service1), then you will see errors on
 your xdcp command, and yet if rhsn is online then the xdcp will actually
 work. This may be a little confusing. For example, here service1 is offline,
 but we are able to use rhsn to complete the xdcp. ::
--- a/docs/source/advanced/hierarchy/appendix_a_setup_backup_service_nodes.rst
+++ b/docs/source/advanced/hierarchy/appendix_a_setup_backup_service_nodes.rst
@@ -1,0 +1,167 @@
+Appendix A: Setup backup Service Nodes
+======================================
+
+For reliability, availability, and serviceability purposes you may wish to
+designate backup service nodes in your hierarchical cluster. The backup
+service node will be another active service node that is set up to easily
+take over from the original service node if a problem occurs. This is not an
+automatic failover feature. You will have to initiate the switch from the
+primary service node to the backup manually. The xCAT support will handle most
+of the setup and transfer of the nodes to the new service node. This
+procedure can also be used to simply switch some compute nodes to a new
+service node, for example, for planned maintenance.
+
+Abbreviations used below:
+
+* MN - management node.
+* SN - service node.
+* CN - compute node.
+
+Initial deployment
+------------------
+
+Integrate the following steps into the hierarchical deployment process
+described above.
+
+
+#. Make sure both the primary and backup service nodes are installed,
+   configured, and can access the MN database.
+#. When defining the CNs add the necessary service node values to the
+   "servicenode" and "xcatmaster" attributes of the `node definitions
+   <http://localhost/fake_todo>`_.
+#. (Optional) Create an xCAT group for the nodes that are assigned to each SN.
+   This will be useful when setting node attributes as well as providing an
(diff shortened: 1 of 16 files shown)


chenglch:
see detail: http://10.4.40.23:8085/advanced/hierarchy/

whowutwut:
@cxhong & @pdlun92  Can you take some time to review the content in the pull request for the hierarchy doc ..

docs/source/advanced/hierarchy/appendix_a_setup_backup_service_nodes.rst
```diff
@@ -0,0 +1,167 @@
...
+the service node that is it's master. The files are placed in
+``/var/xcat/syncfiles`` directory by default, or what is set in site table
+SNsyncfiledir attribute. If the node has multiple service nodes assigned,
+then xdcp will copy the file to each of the service nodes assigned to the
+node. For example, here the files will be copied (scp) to both service1 and
+rhsn. lsdef cn4 | grep servicenode. ::
+
+  servicenode=service1,rhsn
+
+f a service node is offline ( e.g. service1), then you will see errors on
```
cxhong: I think on like 86, you missed "I" before "f",  should be "If"  :)

docs/source/advanced/hierarchy/define_service_node.rst
```diff
@@ -0,0 +1,345 @@
...
+code will automatically substitute the IP address of the node's xcatmaster for 
+the keyword. Here is an example:
+::
+
+  mkdef -t network net1 net=10.5.1.0 mask=255.255.255.224 gateway=<xcatmaster>
+
+The ipforward attribute should be enabled on all the xcatmaster nodes that 
+will be acting as default gateways. You can set ipforward to 1 in the 
+servicenode table or add the line "net.ipv4.ip_forward = 1" in file 
+``/etc/sysctl``.conf and then run "sysctl -p /etc/sysctl.conf" manually to
```
cxhong: Line 308,  I think this should be: ``/etc/sysctl.conf`` and then run ``sysctl -p /etc/sysctl.conf`` manually to

docs/source/advanced/hierarchy/define_service_node.rst
```diff
@@ -0,0 +1,345 @@
...
+executable.
+
+Note: For system p servers, if you plan to have your service node perform the 
+hardware control functions for its compute nodes, it is necessary that the SN 
+ethernet network adapters connected to the HW service VLAN be configured.
+
+Configuring Secondary Adapters
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+o configure secondary adapters, see `Configuring_Secondary_Adapters
```
cxhong: Line 342.   Miss "T" before "o",  should be "To configure......."

cxhong:
Just general question,  Should we use rhel7 as example not rhel6 or lower version of OS?

chenglch:
Thanks cxhong, As the original document is a little old, I'm trying to prepare a environment to learn how to create a hierarchy environment for the new version OS. I will update the example later.

whowutwut:
@tingtli   How come you merged this pull request? Did you review it?  Or are you going to review it?

daniceexi:
@whowutwut 

I think @tingtli has a lot of discussion with @chenglch on this topic. 
I also have take a quick review of this change, it wrote well.

whowutwut:
Ok, thanks for clarifying. 

I came across the document and had some comments, I can make the changes if no one objects since it's hard to comment without the pull request being active.  I have to point out in my text.   

The comments are on the overall flow of the hierarchy document, Im just looking at the main section titles.. 

1. I removed the "(optional)" from the Stateful/Diskful installation step.  This is because we can't have both Stateful and Stateless as optional steps.  This means that the user can skip both steps...  but then i think hierarchy won't work. 

2.  I would suggest the following changes to the headings: 
* Introduction
* Setup the ~~MN Hierarchical~~ Management Node Database
* Define ~~the~~ S~~s~~ervice N~~n~~odes in the database
* Configure DHCP
* Install ~~Setup~~ Service Node**s**
* ~~Set Up the Service Nodes for Stateful (Diskful) Installation~~
* ~~Setup the Service Node for Stateless Deployment (optional)~~
* Test Service Node installation
* Appendix A: Install ~~Setup~~ backup Service Nodes
* Appendix B: Diagnostics
* Appendix C: Migrating a Management Node to a Service Node
* Appendix D: Set up Hierarchical Conserver

3. I think the Stateful and Stateless steps should go under "Install Service Nodes" where we can have some information there saying they need to pick one or the other.   

I think our docs should demonstrate the path that we recommend from the development team ,even though they are optional... we should document the recommended path and mark the optional ones.

PR #149, opened 2015-09-02
 upload doc

junxiawang:
could review according to link: http://10.4.30.1/install/html


whowutwut:
@junxiawang  please add some comments into the pull request about what the changes are. Also. I think your config has some errors because you are committing as "root" user. Please fix and re open  pull request ...

daniceexi:
Close because a new pull request will be created for the change.

PR #153, opened 2015-09-06
Update remoteshell to avoid provsioning hang-on when CN don't support ecdsa

SStar1314:
Update remoteshell to avoid provsioning hang-on when MN support ecdsa but CN don't support ecdsa. Already merge this code to matser branch, but we need to merge in 2.9.1-pcm branch for PCM team.


daniceexi:
I have reviewed this change and agree to check in.

daniceexi:
@SStar1314 Be aware that if pcm want this change get into 2.10.x branch, please create another pull request for 2.10.

SStar1314:
Thank you. I checked with team leader. Pcm will achieve xCat upgrade features in next release, but we don't know which version xCat will be used so far. 
If they make the decision, then I will create a new pull request. Thanks again.

PR #156, opened 2015-09-07
add otherpkg,ospkg,postscripts,syncfile doc

junxiawang:


bybai:
@junxiawang  
1, different OS is used in different examples. I think we have format docs to specify how to divided sections based on different OS; If they are common for all OS, we can use the same osimage definition, our topics are under "IBM Power LE / OpenPOWER" now, so lt is better to pick up related osimage definition.
2, Our docs are under "IBM Power LE / OpenPOWER", but arch in examples are x86_64, Many content are common for both powerLE and x86_64, could you reconsider how to make this more common?
3, We talk about packages for rpm, but there is no content for ubuntu deb packages, we should add contents for deb packages.

daniceexi:
I have reviewed the doc change and agree to check in.

Issue #158, opened 2015-09-08
xCAT cannot generate proper kernel parameter when osimage.nodebootif is set


daniceexi:
In current code logic, when osimage.nodebootif is set, only the 'ifname' kernel parameter is generated. But when installnic is set to 'interface name', the 'netdev', 'ksdevice', 'netdevice' parameters are generated. 

The different between 'osimage.nodebootif' and 'node.installnic' is they belong to different objects, but they should have same impact to a target node. Both of them mean that which interface the node should use to boot.

So we should make the behavior to be consitent between them.

PR #159, opened 2015-09-08
For Issue 158: 

daniceexi:
fix the generated kernel parameters for OS deployment when osimage.nodebootif is set. 
Refer to #158


zet809:
The modification looks good to me, I agree to check in it.

Issue #157, opened 2015-09-08
rcons does not work on sles11.x MN


daniceexi:
xCAT: 2.10
MN: sles11.4

When run rcons against a ipmi based compute node, it will display the following error message.
~~~~
Acquiring startup lock...done
Unable to open socket connection to xcatd daemon on localhost:3001.
Verify that the xcatd daemon is running and that your SSL setup is correct.
Connection failure: SSL connect attempt failed with unknown errorerror:00000000:lib(0):func(0):reason(0) at /opt/xcat/lib/perl/xCAT
/Client.pm line 265.
~~~~

The conserver executes '/opt/xcat/share/xcat/cons/ipmi' to access BMC to get console. /opt/xcat/share/xcat/cons/ipmi needs to access xcatd to get the user/passwd of the BMC. The root cause was for sles11.x, when /opt/xcat/share/xcat/cons/ipmi call Client.pm to access xcatd, the environment variable XCATSSLVER must be set. But XCATSSLVER was not loaded by conserver daemon, then no XCATSSLVER was passed to /opt/xcat/share/xcat/cons/ipmi.

The fix is to add XCATSSLVER when call /opt/xcat/share/xcat/cons/ipmi in conserver.cf.

PR #161, opened 2015-09-08
For Issue 157: 

daniceexi:
add XCATSSLVER environment variable when it's set on sles11.x mn in conserver.cf for calling cons script to communicate with xcatd through tls
Refer to #157


immarvin:
this fix is ok for me, I agree to merge it to the code base.

PR #168, opened 2015-09-09
Simplify the steps to update forked repository

daniceexi:
Refer to http://10.4.14.20/developers/github/syncing_forks.html

### DIFF
--- a/docs/source/advanced/restapi/set_up_ws.rst
+++ b/docs/source/advanced/restapi/set_up_ws.rst
@@ -14,7 +14,7 @@ To improve the security between the REST API client and server, enabling the HTT
 		service httpd restart
 		yum install perl-JSON
 
-* **[RHEL 5 (ppc64)]** ::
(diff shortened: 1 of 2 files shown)


whowutwut:
I'm ok to merge this

whowutwut:
I was just following these steps and just saw i was in a different branch...  should we add a step "git checkout master" before the "git pull upstream/master"...  otherwise we may update the wrong branch..

daniceexi:
@whowutwut Yes I agree to add the 'git checkout master' before pull from upstream.

Since it's a reviewed checkin, I just check it in directly. See https://github.com/daniceexi/xcat-core/commit/529a8ba7c97c412f442f814d8940a684195fe43c

samveen:
It may be advisable to introduce the use of `git stash` in there as well, so that the `git checkout master` doesn't wipe any of the uncommitted work.

Issue #165, opened 2015-09-08
The xcatd command log writer needs a better method to handle command line arguments.


neo954:
Currently, there is no way to distinct a command line argument with spaces in it or multiple command line arguments.

    # copycds nonexist-file.iso
    Error: The management server was unable to find/read /root/nonexist-file.iso. Ensure that file exists on the server at the specified location.
    # copycds 'nonexist file.iso'
    Error: The management server was unable to find/read /root/nonexist file.iso. Ensure that file exists on the server at the specified location.
    # copycds 'nonexist    file.iso'
    Error: The management server was unable to find/read /root/nonexist    file.iso. Ensure that file exists on the server at the specified location.

    ====================================================
    [Date]       2015-8-8 4:41:41
    [ClientType] cli 
    [Request]    copycds nonexist-file.iso 
    [Response]
    Error: The management server was unable to find/read /root/nonexist-file.iso. Ensure that file exists on the server at the specified location.
    ====================================================
    [Date]       2015-8-8 4:43:15
    [ClientType] cli 
    [Request]    copycds nonexist file.iso 
    [Response]
    Error: The management server was unable to find/read /root/nonexist file.iso. Ensure that file exists on the server at the specified location.
    ====================================================
    [Date]       2015-8-8 4:43:26
    [ClientType] cli 
    [Request]    copycds nonexist    file.iso 
    [Response]
    Error: The management server was unable to find/read /root/nonexist    file.iso. Ensure that file exists on the server at the specified location.

I think we can refer to how `bash -x` handle it.

    # copycds nonexist-file.iso
    + copycds nonexist-file.iso
    # copycds 'nonexist file.iso'
    + copycds 'nonexist file.iso'
    # copycds 'nonexist    file.iso'
    + copycds 'nonexist    file.iso'
    # copycds filename-with-signle-quote\'.iso
    + copycds 'filename-with-signle-quote'\''.iso'
    # copycds 'filename-with-line-feed
    > .iso'
    + copycds 'filename-with-line-feed
    .iso'

neo954:
Already fixed. I will close this one.

PR #169, opened 2015-09-10
Fix for GH #165: Quote params

samveen:
Add quotes to the error messages as per #165


daniceexi:
@samveen 
The issue #165 has been fixed by commit c7f3e81. The fix only was to log the argument quated when there's specific characters included in the command arguments. The output message was not covered. We don't think it's necessary to handle the specific characters in the output message. So we decided not accept your fix.

whowutwut:
@daniceexi , you can close this pull request if decided not to merge.

PR #171, opened 2015-09-10
Add information about stashing work before syncing forks

whowutwut:
Added information about stashing work before switching to update the master branch of your forked repo as to not lose any umcommited changes that may be in progress.  Based on comment by @samveen  in Pull Request #168


daniceexi:
@whowutwut @samveen Using **git stash** is a good way to backup your change when you need to update master branch. Thanks.

daniceexi:
I reviewed the change and agree to check in.

Issue #160, opened 2015-09-08
The /var/log/xcat/commands.log needs a better format of datetime


neo954:
Currently, one log entry looks like the following. Beside the off-by-one bug of the month in the ``Date'' field, we need a better datetime format. At least, we need add leading zeros to month, day-of-month, hour, minite and second.

I recommand we use RFC 2822 format. Example: `Mon, 07 Aug 2006 12:34:56 -0600`
Or we may use RFC 3339 format. Example: `2006-08-07 12:34:56-06:00`

    ====================================================
    [Date]       2015-8-8 4:6:33
    [ClientType] cli 
    [Request]    copycds 
    [Response]
    Error: copycds needs at least one full path to ISO currently.

whowutwut:
This seems to be duplicate of Issue #140?

neo954:
Already fixed. I will close this one.

Issue #164, opened 2015-09-08
[fvt]2.11:offline trouble shooting :could not cover rcons getcons text error info in commands.log


junxiawang:
1.down load the latest build
2.makeconsercf
3.rcons node
[root@c910f03c05k06 ~]# rcons c910f03c05k08
[Enter `^Ec?' for help]
ERROR/WARNING: communication with the xCAT server seems to have been ended prematurely

------------->could not get information
4.tail -f /var/log/xcat/commands.log
====================================================
[Date]       2015-8-8 4:8:11
[ClientType] cli
[Request]    makeconservercf
[Response]
====================================================
[Date]       2015-8-8 4:8:12
[ClientType] cli
[Request]    getcons c910f03c05k08 text
[Response]
Error: A fatal error was encountered, the following information may help identify a bug: Can't use string ("Unable to query running VM") as an ARRAY ref while "strict refs" in use at /opt/xcat/sbin/xcatd line 3008.
===================================================
----------------------->This is not the real root cause for getcons text failed.

samveen:
@junxiawang the report will be more readable and understandable if you quote the output as per markdown rules with a simple ```` ``` ```` on either end.

Issue #163, opened 2015-09-08
The xcatd command log writer failed to get the error message of nodels


neo954:
This is obviously. See below for the details of recreation steps.

    # nodels foo
    Error: Invalid nodes and/or groups in noderange: foo
    # lsdef foo
    Error: Could not find an object named 'foo' of type 'node'.

    ====================================================
    [Date]       2015-8-8 4:33:3
    [ClientType] cli 
    [Request]    nodels foo 
    [Response]
    ====================================================
    [Date]       2015-8-8 4:33:17
    [ClientType] cli 
    [Request]    lsdef foo 
    [Response]
    Error: Could not find an object named 'foo' of type 'node'.

neo954:
Already fixed. I will close this one.

Issue #162, opened 2015-09-08
Security issue of the xcatd command log writer


neo954:
Currently, we use a separate `xcatd` daemon named "Command log writer" to do the extra log operating. It bind to 0.0.0.0 and listern TCP port 3003. This is a critical security problem.

First of all, anyone and submit fake log messages by sending anything to the TCP port 3003. Something like `echo "FAKE LOG" | nc 10.2.3.27 3003` will do the work.

Second, by sending a lot of useless log messages, anyone can exhaust the disk space on the log server.

Last, the log writer use a "Stop-and-Wait" protocol. anyone can block the log writer by connect to TCP port 3003, and do nothing. All the remain log will be blocked. And the server may end up with out of memory or get even worse.

    # ps ax | grep -v grep | grep -i xcatd
    17809 ?        Ss     0:00 xcatd: SSL listener
    17810 ?        S      0:10 xcatd: DB Access
    17811 ?        S      0:01 xcatd: UDP listener
    17812 ?        S      0:00 xcatd: Discovery worker
    17813 ?        S      0:00 xcatd: install monitor
    17820 ?        S      0:00 xcatd: Command log writer
    # netstat -nltp | grep 3003
    tcp        0      0 0.0.0.0:3003            0.0.0.0:*               LISTEN      17820/xcatd: Comman 

Conclusion, we should at lease only `bind()` to 127.0.0.1 or use OpenSSL as the other daemons of xcatd.

daniceexi:
Yes, this is a security problem. Make the logwriter only listen to 127.0.0.1 should be good enough.

neo954:
Already fixed. I will close this one.

PR #174, opened 2015-09-13
Make commands.log support rotate mechanism and fix bug#160-165

hu-weihua:
Fix bugs:
#160 The /var/log/xcat/commands.log needs a better format of datetime 
#162 Security issue of the xcatd command log writer
#163 The xcatd command log writer failed to get the error message of nodels
#164 offline trouble shooting :could not cover rcons getcons text error info in commands.log
#165 The xcatd command log writer needs a better method to handle command line arguments

New feature
Make commands.log support rotate mechanism

### DIFF
--- a/xCAT-server/sbin/xcatd
+++ b/xCAT-server/sbin/xcatd
@@ -1048,31 +1048,61 @@ unless ($cmdlog_svrpid){
     my $cmdlogfile;
     my $cmdlogpidfile;
     my $retry=200;
-
+    my $writing=0;
+    my $cmdlogfileswitch=0;
+    my $cmdlogservicefile="/var/run/xcat/cmdlogservice.pid";
+	
     $SIG{USR2} = sub {
-        xCAT::MsgUtils->message("S","INFO xcatd: 'Command log writer' process $$ is terminated by USR2 signal");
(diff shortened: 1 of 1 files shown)


daniceexi:

xCAT-server/sbin/xcatd
```diff
@@ -2450,7 +2502,12 @@ sub service_connection {
     
     if(exists($req->{arg})){
         foreach my $arg (@{$req->{arg}}) {
-	        $cmdlog_alllog .= $arg." ";
+            if($arg =~ /[^A-Za-z0-9.-]/){
+                $arg =~ s/'/'\\''/g;
```
immarvin: hi, this line might caused the issue https://github.com/xcat2/xcat-core/issues/203

daniceexi:
@immarvin Could help to check whether it could fix the issue if removing the change on $arg.

immarvin:
yes.

When this line is commented out, the request passed to xdsh.pm by ``xdsh c910f03c01p08 "echo 'test' >/tmp/test"`` will be:
````
...
            'arg' => [
                       'echo \'test\' >/tmp/test'
                     ],
...
````

Issue #177, opened 2015-09-15
`chdef ` clobbers existing objects with `-n` option


samveen:
Old node definitions:
```
[root@xcatmn01 xCAT_plugin]# lsdef -t node -o n50,spare01-a1
Object name: n50
    bmc=n50-bmc
    chain=runcmd=bmcsetup,shell
    groups=ipmi,idataplex,42perswitch,compute,all
    ip=172.20.101.50
    mgt=ipmi
    netboot=xnba
    ondiscover=nodediscover
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles,create_root_authorized_keys
    profile=compute
    rack=1
    switch=switch2
    switchport=8
    unit=C8
Object name: spare01-a1
    bmc=spare01-a1-ilo
    chain=runcmd=bmcsetup,shell
    groups=ipmi,all
    height=2
    ip=10.2.80.22
    mac=52:55:00:72:42:ff!spare01-A1-m|52:55:00:7b:ba:93!spare01-A1-s|52:55:00:32:de:3d!spare01-A1-d
    mgt=ipmi
    nichostnamesuffixes.eth2=-d
    nichostnamesuffixes.eth1=-m
    nichostnamesuffixes.eth0=-s
    nicips.eth2=10.40.2.4
    nicips.eth1=10.4.3.4
    nicips.eth0=10.2.80.22
    ondiscover=nodediscover
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    rack=A1
    serial=123456789
    unit=1
```
rename node:
```
[root@xcatmn01 xCAT_plugin]# chdef -t node spare01-a1 -n n50
Changed the object name from spare01-a1 to n50.
```
New definitions:
```
[root@xcatmn01 xCAT_plugin]# lsdef -t node -o n50,spare01-a1
Object name: n50
    bmc=spare01-a1-ilo
    chain=runcmd=bmcsetup,shell
    groups=ipmi,idataplex,42perswitch,compute,all
    height=2
    ip=10.2.80.22
    mac=52:55:00:72:42:ff!spare01-A1-m|52:55:00:7b:ba:93!spare01-A1-s|52:55:00:32:de:3d!spare01-A1-d
    mgt=ipmi
    netboot=xnba
    nichostnamesuffixes.eth2=-d
    nichostnamesuffixes.eth1=-m
    nichostnamesuffixes.eth0=-s
    nicips.eth2=10.40.2.4
[...]

PR #178, opened 2015-09-15
Fix for  #177 (`chdef ` clobbers existing objects with `-n` option)

samveen:
Added check to ensure that new name isn't already used on an existing object.


daniceexi:

PR #180, opened 2015-09-15
Fix for #177 (`chdef ` clobbers existing objects with `-n` option) for 2.9 branch

samveen:
Added check to ensure that new name isn't already used on an existing object(see #177).


daniceexi:

Issue #172, opened 2015-09-10
confignics does not handle all new predictable network interface device names


whowutwut:
Opening this Issue on behalf of Vlad.  He his this issue on RHEL 7 running xCAT 2.10 

```
confignics script does not support RHEL 7 enP interfaces
--- if [ `echo $key | grep -E 'e(n|th|m)[0-9a-z]+'` ];then
+++ if [ `echo $key | grep -E 'e(n|th|m|nP)[0-9a-z]+'` ];then 
```

Looking more into the predictable network interface names that part of systemd, It looks like P is a valid  character.   From: http://cgit.freedesktop.org/systemd/systemd/tree/src/udev/udev-builtin-net_id.c#n20

````
 *
 * Two character prefixes based on the type of interface:
 *   en -- ethernet
 *   sl -- serial line IP (slip)
 *   wl -- wlan
 *   ww -- wwan
 *
 * Type of names:
 *   b<number>                             -- BCMA bus core number
 *   ccw<name>                             -- CCW bus group name
 *   o<index>[d<dev_port>]                 -- on-board device index number
 *   s<slot>[f<function>][d<dev_port>]     -- hotplug slot index number
 *   x<MAC>                                -- MAC address
 *   [P<domain>]p<bus>s<slot>[f<function>][d<dev_port>]
 *                                         -- PCI geographical location
 *   [P<domain>]p<bus>s<slot>[f<function>][u<port>][..][c<config>][i<interface>]
 *                                         -- USB port number chain
 *
````

It seems like most options above are satisfied with the original regular expression, I am OK with the suggested change but it seems like we only care of "en" and nothing else that follows... maybe should just add A-Z ....
```
if [ `echo $key | grep -E 'e(n|th|m)[0-9a-zA-Z]+'` ];then 
```

daniceexi:
@whowutwut I agree with your proposal.

samveen:
As some wise man said,  `Predictable interface names are anything but...`. The only thing predictable is that the device will always have same name, not that the name will follow any predictable scheme. As an example, network interface naming changed from RH 6.6 to Centos 7 to Centos 7.1. ( MN: xCAT 2.9.1 ON C7, HP ProLiant DL380p Gen8 nodes). Please take that into consideration.

daniceexi:
@samveen The naming of interface determined by the kernel and udev/systemd system, xCAT has not good way to control that. Do you have any good idea?

whowutwut:
[...]

PR #188, opened 2015-09-16
Fix for xcat-core/issues/177 2.10 branch, predictable nic names

whowutwut:
Add to regular expression for ethernet devices to handle the
predictable nic naming conventions that are starting to be used
in the later OS releases

xcat2/xcat-core#172


daniceexi:

PR #194, opened 2015-09-17
create links for rescanplugins

penguhyang:
rescanplugins command was not created for ubuntu MN, need to add this link.


daniceexi:

daniceexi:
@penguhyang did you make a new ubuntu build and tested it with a new/update install?

penguhyang:
yes, @tingtli helped me to make a new ubuntu build and I tested it with a update install. It works fine.

PR #195, opened 2015-09-18
correct runcmdinstaller code to return right value

penguhyang:
2.11:runcmdinstaller -h return wrong value.

### DIFF
--- a/xCAT-server/sbin/runcmdinstaller
+++ b/xCAT-server/sbin/runcmdinstaller
@@ -7,10 +7,15 @@
 awk  -v argc="$#" -v node="$1" -v cmd="$2" 'BEGIN {
 port = 3001
 action = "sh"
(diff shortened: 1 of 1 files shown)


xCAT-server/sbin/runcmdinstaller
```diff
@@ -10,7 +10,7 @@ action = "sh"
 if( node=="-h" || argc !=2 || ! node || ! cmd){
   print "Usage:\n\n   runcmdinstaller <node> \"<command>\"\n"
   print           "   make sure all the commands are quoted by \"\"\n";
-  exit 1;
```
immarvin: hi, "runcmdinstaller -h" should return 0

penguhyang: yes, "runcmdinstaller -h" should return 0 not 1.

Issue #197, opened 2015-09-18
[DEV] open tcp port of rsyslog on xCAT MN to receive xcat debug logs forwarded via tcp


immarvin:
when xcatdebugmode=1, some debug logs in the os installer or installed nodes are forwarded to MN via tcp by default, we should turn on tcp port of rsyslog to receive these logs

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/198

Issue #196, opened 2015-09-18
[DEV] duplicate syslog entries in /var/log/messages on sles12 ppc64le mn


immarvin:
On the sles12 ppc64le mn, the log entries in /var/log/messages include duplicate entries :
```
  2015-09-18T04:10:13-04:00 c910f03c05k27 root: xxxx
  2015-09-18T04:10:13-04:00 c910f03c05k27 root: xxxx
  2015-09-18T04:10:25-04:00 c910f03c05k27 root: yyyyyy
  2015-09-18T04:10:25-04:00 c910f03c05k27 root: yyyyyy
  2015-09-18T03:52:51.790961-04:00 c910f03c04k04 xcat[10993]: xCAT: Allowing getcons to     c910f03c05k27 text for root from localhost
  2015-09-18T03:52:51.790961-04:00 c910f03c04k04 xcat[10993]: xCAT: Allowing getcons to c910f03c05k27 text for root from localhost
```
the rsyslog installed on the mn is:
```
# rsyslogd -v
rsyslogd 8.4.0, compiled with:
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              Yes
        FEATURE_DEBUG (debug build, slow code): No
        32bit Atomic operations supported:      Yes
        64bit Atomic operations supported:      Yes
        memory allocator:                       system default
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
        Number of Bits in RainerScript integers: 64

See http://www.rsyslog.com for more information.
```

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/198

PR #198, opened 2015-09-18
fix defect # 197 and #196 on "syslog" postscript

immarvin:
https://github.com/xcat2/xcat-core/issues/197
https://github.com/xcat2/xcat-core/issues/196

listen on tcp port to enable receiving the xcat debug logs forwarded  via tcp protocol; 

comment out the rules which will write logs to /var/log/messages to prevent duplicate logs after <syslog> script is run


samveen:
may I suggest the following paradigm:
```
TF=$(mktemp)
sed 's*' $SRC >$TF
chmod --reference $SRC $TF ;  chown --reference $SRC $TF
mv $SRC ${SRC}.xcatback
mv $TF $SRC
```

immarvin:
hi,@samveen , can you explain more on your paradigm? thx

samveen:
@immarvin 
* Editing any file in-place is slightly problematic as multiple writing processes can clobber the contents.
* On POSIX systems, `mv` semantics are atomic, which guarantees that file contents will never be clobbered.
* `mktemp` guarantees a unique not-previously-existing temp file.

Thus this sequence of commands guarantee that in case multiple processes are trying to modify the file, you still have a sane output. 

I realised I had the command order messed up:
```
TF=$(mktemp)
mv $SRC ${SRC}.xcatback
sed 's*' ${SRC}.xcatback >$TF
chmod --reference ${SRC}.xcatback $TF ; chown --reference ${SRC}.xcatback $TF
mv $TF $SRC
```

daniceexi:
@samveen What's the meaning of this line? It does not look like a correct syntax for 's' - replacement.

sed 's*' ${SRC}.xcatback >$TF

immarvin:
@daniceexi ,I think this is just a place holder for  sed substitution.

daniceexi:
Ok, get it. Then I think @samveen 's idea is good that we can 1. make backup of changed file; 2. work on temp file and then move it back to correct path.

immarvin:
@samveen , the [sed manual] (http://www.gnu.org/software/sed/manual/sed.html) mentioned :

````
-i[SUFFIX]
--in-place[=SUFFIX]
    This option specifies that files are to be edited in-place. GNU sed does this by creating a temporary file and sending output to this file rather than to the standard output.1.

    This option implies -s.

    When the end of the file is reached, the temporary file is renamed to the output file's original name. The extension, if supplied, is used to modify the name of the old file before renaming the temporary file, thereby making a backup copy2).

    This rule is followed: if the extension doesn't contain a *, then it is appended to the end of the current filename as a suffix; if the extension does contain one or more * characters, then each asterisk is replaced with the current filename. This allows you to add a prefix to the backup file, instead of (or in addition to) a suffix, or even to place backup copies of the original files into another directory (provided the directory already exists).

    If no extension is supplied, the original file is overwritten without making a backup. 

````
I am wondering whether `sed -i$$*.bak ...` invokes the similar process as your paradigm, what do you think?

**Comment by samveen** [comment]
@immarvin That is a much better implementation than what I suggested(and I learn something new, yet again). I suggest combining multiple substitutions to the same file into the same sed command, using the semicolon or newline as separator.

```
sed -i.xcatback 's*;s*;s*' $FILE
```

immarvin:
@samveen , thanks for your comment.

I will try to combine the successive  substitutions in the same conditional branches.

would you please let me know after your investigation on the ownership and permissions?  thanks

samveen:
From reading the [sed 4.2.2 code](http://git.savannah.gnu.org/cgit/sed.git/tree/sed/execute.c?id=v4.2.2#n675), it seems that sed changes the temporary file's permission to match the source file, so using `-i` with a suffix is the best option.

immarvin:
@samveen , it seems that ``-i$$*.bak`` will backup the original file just before renaming the temporary file to the the original file name after substitutions, ``sed -i`` will create a temporary file and write the output to the temporary file even without the ``$$*.bak`` specified. So I think ``sed -i`` is ok, we do not need to specify the ``-i$$*.bak``. Any thoughts on this?

````
    When the end of the file is reached, the temporary file is renamed to the output file's original name. The extension, if supplied, is used to modify the name of the old file before renaming the temporary file, thereby making a backup copy2).
   If no extension is supplied, the original file is overwritten without making a backup. 
````

immarvin:
yes, SN receives the logs from CN first and then forwards the logs to MN.

immarvin:
updated the code logic to enable log receiving and forwarding on xCAT SN

daniceexi:

samveen:
@daniceexi, you are right. The `-i` option works well with or without a specified suffix. However, as a policy, I suspect it may be better to create a backup 'Just in case', as done by the `make*` commands.

PR #201, opened 2015-09-21
modify makeconservercf and add addtional packages testcase

junxiawang:
modify makeconservercf testcase and add additional  packages  testcases


daniceexi:
I am good with the change for the xCAT-test/autotest/testcase/makeconservercf/cases0

@tingtli could you review the following part?

tingtli:
I will merge the cases since auto-test failed for the makeconservercf cases.

Need to update the otherpkgs and ospkgs list later

PR #202, opened 2015-09-21
Story50687log

neo954:
Send signal HUP to xCAT log writer when log rotate.


whowutwut:
Just trying to understand why we got so many commit history with this pull request. Because of an upstream merge after creating the branch?

neo954:
@whowutwut Fixed :-P

daniceexi:
I have reviewed the code change and agree to check in.

Issue #71, opened 2015-08-10
[FVT] An obsolete util-linux package is installed for Ubuntu 14.04.3 on ppc64el diskless node.


neo954:
The osimage definition looked like the following. Ubuntu 14.04.3 was used on ppc64el architechture.

```
root@c910f03c04k07:~# lsdef -t osimage compute_1533a
Object name: compute_1533a
    exlist=/opt/xcat/share/xcat/netboot/ubuntu/compute.exlist
    imagetype=linux
    kernelver=3.19.0-25-generic
    osarch=ppc64el
    osname=Linux
    osvers=ubuntu14.04.3
    otherpkgdir=/install/post/otherpkgs/ubuntu14.04.3/ppc64el
    permission=755
    pkgdir=/install/ubuntu14.04.3/ppc64el
    pkglist=/opt/xcat/share/xcat/netboot/ubuntu/compute.ubuntu14.04.ppc64el.pkglist
    profile=compute
    provmethod=netboot
    rootimgdir=/install/netboot/ubuntu14.04.3/ppc64el/compute_1533a
```

After the genimage was running, I found out package util-linux installed in the osimage was an old versoin came from Ubuntu remote mirror. But a newer version was shipped in the iso file. Please check the attachment for all the detail genimage output.

[genimage_compute_1533a.out.txt](https://github.com/xcat2/xcat-core/files/1281/genimage_compute_1533a.out.txt)

By look into this problem, as you can see from the following outputs, it is due to the dependency changing of the newer util-linux package.

```
root@c910f03c04k07:~# chroot /install/netboot/ubuntu14.04.3/ppc64el/compute_1533a/rootimg apt-cache policy util-linux
util-linux:
  Installed: 2.20.1-5.1ubuntu20
  Candidate: 2.20.1-5.1ubuntu20.6
  Version table:
     2.20.1-5.1ubuntu20.6 0
        500 http://10.3.4.7/install/ubuntu14.04.3/ppc64el/ trusty/main ppc64el Packages
 *** 2.20.1-5.1ubuntu20 0
        100 /var/lib/dpkg/status
root@c910f03c04k07:~# chroot /install/netboot/ubuntu14.04.3/ppc64el/compute_1533a/rootimg apt-get install util-linux
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Suggested packages:
  util-linux-locales dosfstools
The following packages will be REMOVED:
  mac-fdisk pmac-fdisk ubuntu-minimal
[...]

PR #204, opened 2015-09-21
using dist-upgrade to upgrade the most important packages

penguhyang:
dist-upgrade will attempt to upgrade the most important packages at the expense of less important ones if necessary. The dist-upgrade command may therefore remove some packages. Thus an obsolete util-linux package will not installed for Ubuntu 14.04.3 on ppc64el diskless node.

### DIFF
--- a/xCAT-server/share/xcat/netboot/ubuntu/genimage
+++ b/xCAT-server/share/xcat/netboot/ubuntu/genimage
@@ -354,8 +354,8 @@ unless ($onlyinitrd) {
 
    close($aptconfig);
 
-   # run apt-get upgrade to update any installed debs
-   my $aptgetcmd_update = $aptgetcmd . "&&". $aptgetcmdby . " upgrade  ";
+   # run apt-get dist-upgrade to update any installed debs
+   my $aptgetcmd_update = $aptgetcmd . "&&". $aptgetcmdby . " dist-upgrade  ";
    $rc = system("$aptgetcmd_update");
         
    # Start to install pkgs in pkglist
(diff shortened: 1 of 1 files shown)


immarvin:
hi,  on `apt-get upgrade` and `apt-get dist-upgrade` , there is a discussion [here](http://askubuntu.com/questions/194651/why-use-apt-get-upgrade-instead-of-apt-get-dist-upgrade),
````
I typically upgrade my machines with:

sudo apt-get update && time sudo apt-get dist-upgrade

Below is an excerpt from man apt-get. Using upgrade keeps to the rule: under no circumstances are currently installed packages removed, or packages not already installed retrieved and installed. If that's important to you, use apt-get upgrade. If you want things to "just work", you probably want apt-get dist-upgrade to ensure dependencies are resolved.

To expand on why you'd want upgrade instead of dist-upgrade, if you are a systems administrator, you need predictability. You might be using advanced features like apt pinning or pulling from a collection of PPAs (perhaps you have an in-house PPA), with various automations in place to inspect your system and available upgrades instead of always eagerly upgrading all available packages. You would get very frustrated when apt performs unscripted behavior, particularly if this leads to downtime of a production service.

upgrade
    upgrade is used to install the newest versions of all packages
    currently installed on the system from the sources enumerated in
    /etc/apt/sources.list. Packages currently installed with new
    versions available are retrieved and upgraded; under no
    circumstances are currently installed packages removed, or packages
    not already installed retrieved and installed. New versions of
    currently installed packages that cannot be upgraded without
    changing the install status of another package will be left at
    their current version. An update must be performed first so that
    apt-get knows that new versions of packages are available.

dist-upgrade
    dist-upgrade in addition to performing the function of upgrade,
    also intelligently handles changing dependencies with new versions
    of packages; apt-get has a "smart" conflict resolution system, and
    it will attempt to upgrade the most important packages at the
    expense of less important ones if necessary. So, dist-upgrade
    command may remove some packages. The /etc/apt/sources.list file
    contains a list of locations from which to retrieve desired package
    files. See also apt_preferences(5) for a mechanism for overriding
    the general settings for individual packages.

```` 

The `apt-get upgrade` is called twice in ubuntu/genimage, since there is some tricky things here to use `upgrade` instead of `dist-upgrade` . I suggest you discuss with @zet809 on this.

penguhyang:
yes, apt-get upgrade is restricted to the case where packages are to be replaced by newer versions with no package added or removed. apt-get upgrade will refuse to work when there are additions or removals required by the updated versions. In this issue https://github.com/xcat2/xcat-core/issues/71 , apt-get upgrade will not upgrade util-linux since it breaks the mac-fdisk and pmac-fdisk.

PR #206, opened 2015-09-22
[docs,xCAT-test]Remove editor temp files

samveen:
Remove editor temp files (`vim` `.swp` files)


whowutwut:
I have seen this a lot too.  I think this is a by product of using the ```git commit -a``` instead of adding files manually...

PR #213, opened 2015-09-24
enhance bmcdiscover to get mtms for Firestone machine

zet809:


bybai:
@zet809 As we discussed, should enhance bmc username or password check. After bmcdiscover found the bmc, it can define  node, if bmc username or password is not correct, give warning message.

daniceexi:
I am ok with the code logic but maybe we need to consider the performance when implement features like don't call 'chdef' directly. Access database once to write entries for all nodes instead of access db for node one by one.

PR #214, opened 2015-09-24
fix issue pping cannot connect sqlsite under non-root user

bybai:
When root execute pping, the code logical is the same with before.
When non-root user execute pping, using::
 tabdump site|grep nmapoptions|awk -F, '{print $2}'|sed -e 's/\"//g'|awk -F, '{print $1}'

to get value of nmapoptions,
instead of ::
xCAT::TableUtils->get_site_attribute("nmapoptions");

Testing covered site.nmapoptions was empty or " --min-rtt-timeout 1s"


daniceexi:

PR #215, opened 2015-09-24
fix issue 212: rinv all doesn't work for firestone JMET machine

zet809:


daniceexi:

PR #220, opened 2015-09-25
Fix some bugs found during support IB on ubuntu14.04.3 and ppc64le

hu-weihua:
Fix some bugs found during support ubuntu14.04.3 on ppc64le
1. Use URL instead of ip address of www.ports.ubuntu.com
2. Unset ARCH before invoking Mellanox script, let Mellanox figure out the arch by itself, to avoid xcat input wrong ARCH to Mellanox
3. Refine some prompt message
4. Avoid IB adapter firmware upgrade in all scenario


daniceexi:
I reviewed the change and agree to check in the code.

PR #222, opened 2015-09-28
make the run_ps subroutine to be consistent

penguhyang:
make the run_ps to be consistent in ubuntu.post, xcat.post and xcatdsklspost

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/post.ubuntu
+++ b/xCAT-server/share/xcat/install/scripts/post.ubuntu
@@ -152,9 +152,11 @@ return_value=0
 # subroutine used to run postscripts
 run_ps () {
  local ret_local=0
+ mkdir -p "\"/var/log/xcat\""
  local logfile=\"/var/log/xcat/xcat.log\"
 
- if [ -f \$1 ]; then
+ if [ -f \$1 ]; then 
+  echo \"\`date\` Running postscript: \$*\" 
   msgutil_r \"\$MASTER_IP\" \"info\" "\"\`date\` Running postscript: \$*\"" \"\$logfile\"
   if [ \"\$XCATDEBUGMODE\" = \"1\" ]; then
      bash -x ./\$@ 2>&1 | tee -a \$logfile
@@ -163,12 +165,14 @@ run_ps () {
      ./\$@ 2>&1 | tee -a \$logfile
      ret_local=\${PIPESTATUS[0]}
   fi
-  msgutil_r \"\$MASTER_IP\" \"info\" "\"\`date\` postscript \$* return with \$ret_local\"" \"\$logfile\"
+
   if [ \"\$ret_local\" -ne \"0\" ]; then
      return_value=\$ret_local
   fi
-
+  echo \"Postscript: \$* exited with code \$ret_local\" 
+  msgutil_r \"\$MASTER_IP\" \"info\" "\"\`date\` postscript \$* return with \$ret_local\"" \"\$logfile\"
  else
+  echo \"\`date\` Postscript \$1 does NOT exist.\" 
   msgutil_r \"\$MASTER_IP\" \"info\" "\"\`date\` Postscript \$1 does NOT exist.\"" \"\$logfile\"
   return_value=-1
  fi
--- a/xCAT-server/share/xcat/install/scripts/post.xcat
+++ b/xCAT-server/share/xcat/install/scripts/post.xcat
@@ -137,9 +137,11 @@ return_value=0
 # subroutine used to run postscripts
 run_ps () {
  local ret_local=0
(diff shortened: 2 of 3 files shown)


xCAT-server/share/xcat/install/scripts/post.ubuntu
```diff
@@ -151,11 +151,20 @@ echo "
...
  local ret_local=0
+ logdir=\"/var/log/xcat\"
+ mkdir -p \$logdir
  local logfile=\"/var/log/xcat/xcat.log\"
 
- if [ -f \$1 ]; then
-  msgutil_r \"\$MASTER_IP\" \"info\" "\"\`date\` Running postscript: \$*\"" \"\$logfile\"
+ if [ -f \$1 ]; then 
+  echo \"\`date\` Running postscript: \$*\" 
+  if [ "$file_name" == "xcatdsklspost" ]; then
```
immarvin: there is no difference between the run_ps code logic in %post section of installation and the xcatdsklspost during updatenode. you do not need to add this conditional branch. 
what's more,  the condition  ``if [ "$file_name" == "xcatdsklspost" ]; then``  will never be matched , since during "updatenode", the run_ps is never run in a file named "xcatdsklspost"

xCAT-server/share/xcat/install/scripts/post.xcat
```diff
@@ -148,16 +157,26 @@ run_ps () {
...
      ret_local=\${PIPESTATUS[0]}
   fi
-  msgutil_r \"\$MASTER_IP\" \"info\" "\"\`date\` postscript \$* return with \$ret_local\"" \"\$logfile\"
+
   if [ \"\$ret_local\" -ne \"0\" ]; then
      return_value=\$ret_local
   fi
-
+  echo \"Postscript: \$* exited with code \$ret_local\" 
+  if [ "$file_name" == "xcatdsklspost" ]; then
```
immarvin: this conditional branch is unnecessary,  ``exited with code`` and ``return with`` are just different expressions on the same meaning, use either one is ok.

 the priority of the message can be "info", there is no need to add branch to log the message as "debug"

xCAT/postscripts/xcatdsklspost
```diff
@@ -841,10 +847,18 @@ run_ps () {
...
-  msgutil_r \$MASTER_IP \"debug\" \"\`date\` Postscript: \$* exited with code \$ret_local\" \"\$logfile\"
+  if [ "$file_name" == "xcatdsklspost" ]; then
+     msgutil_r \$MASTER_IP \"debug\" \"\`date\` Postscript: \$* exited with code \$ret_local\" \"\$logfile\"
+  else
+     msgutil_r \"\$MASTER_IP\" \"info\" "\"\`date\` postscript \$* return with \$ret_local\"" \"\$logfile\"
+  fi
  else
   echo \"\`date\` Postscript \$1 does NOT exist.\" 
-  msgutil_r \$MASTER_IP \"debug\" \"\`date\` Postscript: \$* exited with code \$ret_local\" \"\$logfile\"
+  if [ "$file_name" == "xcatdsklspost" ]; then
```
immarvin: the code logic is totally wrong. this branch is under the condition that the given postscript does not exist. we do not need to add the conditional branch

immarvin:
the comments above apply to all the files in your pull request

penguhyang:
@immarvin Files have been corrected appropriately.

xCAT-server/share/xcat/install/scripts/post.ubuntu
```diff
@@ -152,9 +152,12 @@ return_value=0
 # subroutine used to run postscripts
 run_ps () {
  local ret_local=0
- local logfile=\"/var/log/xcat/xcat.log\"
+ logdir=\"/var/log/xcat\"
```
immarvin: there is no need to define ``logdir``
````
+ logdir=\"/var/log/xcat\"
+ mkdir -p \$logdir
````
 can be substituted with 
````
mkdir -p "\"/var/log/xcat\""
````

File: xCAT-server/share/xcat/install/scripts/post.ubuntu [hunk 5]
```diff
@@ -152,9 +152,12 @@ return_value=0
...
+ mkdir -p \$logdir
+ logfile=\"/var/log/xcat/xcat.log\"
```
immarvin: ``logfile`` should also be declared as ``local`` variable

immarvin:
the code looks good, I agree to merge it.

daniceexi:
@penguhyang @immarvin 
If the subroutines in different files are totally same, we should make sure that people should only change one file and copy the subroutine from one to another two.

PR #223, opened 2015-09-28
Fix bug#203 xdsh could not recognize single quote mark

hu-weihua:
Fix bug#203 xdsh could not recognize single quote mark
Modified the original attribute accidently when generate command log


immarvin:
the code seems ok to me, I agree to merge it.

PR #225, opened 2015-09-28
replace lsslp with bmcdiscover for discover BMC/FSP

zet809:


zet809:
The http://10.3.5.10/install/xcat-docs/

daniceexi:
I have reviewed the code change and discussed with @zet809  that the description of using bmcdiscovery still has some problem. I'll merge the change first, @zet809  will fix it later.

PR #228, opened 2015-09-29
check the postscripts before running it with 'bash -x'

penguhyang:
When enable the site.xcatdebugmode, the postscript will be run with 'bash -x' to output more debug messages. But not all of xCAT postscripts are shell script, so do the check before running it with 'bash -x'.

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/post.ubuntu
+++ b/xCAT-server/share/xcat/install/scripts/post.ubuntu
@@ -159,8 +159,12 @@ run_ps () {
   echo \"\`date\` Running postscript: \$*\" 
   msgutil_r \"\$MASTER_IP\" \"info\" "\"\`date\` Running postscript: \$*\"" \"\$logfile\"
   if [ \"\$XCATDEBUGMODE\" = \"1\" ]; then
-     bash -x ./\$@ 2>&1 | tee -a \$logfile
-     ret_local=\${PIPESTATUS[0]}
+     local compt=\$(file \$1)
+     local reg=\"shell script\"
+     if [[ \"\$compt\" =~ \$reg ]]; then
+        bash -x ./\$@ 2>&1 | tee -a \$logfile
+        ret_local=\${PIPESTATUS[0]}
+     fi
   else
      ./\$@ 2>&1 | tee -a \$logfile
      ret_local=\${PIPESTATUS[0]}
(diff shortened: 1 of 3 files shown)


xCAT/postscripts/xcatdsklspost
```diff
@@ -829,7 +829,11 @@ run_ps () {
   echo \"\`date\` Running postscript: \$*\" 
   msgutil_r \"\$MASTER_IP\" \"info\" "\"\`date\` Running postscript: \$*\"" \"\$logfile\"
   if [ \"\$XCATDEBUGMODE\" = \"1\" ]; then
-     bash -x ./\$@ 2>&1 | tee -a \$logfile
+     for compt in `ls ./`; do
```
immarvin: the typical usage of run_ps is:
````
run_ps syslog
run_ps remoteshell
````
the argument is just one postscript file name, no loop needed

File: xCAT/postscripts/xcatdsklspost [hunk 2]
```diff
@@ -829,7 +829,11 @@ run_ps () {
...
+        if [ `file $compt | grep -c "shell script" ` -gt 0 ]; then
```
immarvin: the conditional expression ``if [ `file $compt | grep -c "shell script" ` -gt 0 ]; `` is invalid for the following example:
````
kvm2:/xcatpost # `file syslog|grep -c 'shell script'`
If '1' is not a typo you can use command-not-found to lookup the package that contains it, like this:
    cnf 1
````

so, please use pattern matching in the output string of  the ``stdout`` output of command ``file``

**Comment by immarvin** [comment]
all the comments apply to all the files in your pull request.

just a reminder, fixes of this kind are very important in xcat, any error might cause problems in all the OS and arch, so please make sure you have fully understand the code logic before you start the code modification.

File: xCAT-server/share/xcat/install/scripts/post.ubuntu [hunk 3]
```diff
@@ -159,8 +159,11 @@ run_ps () {
   echo \"\`date\` Running postscript: \$*\" 
   msgutil_r \"\$MASTER_IP\" \"info\" "\"\`date\` Running postscript: \$*\"" \"\$logfile\"
   if [ \"\$XCATDEBUGMODE\" = \"1\" ]; then
-     bash -x ./\$@ 2>&1 | tee -a \$logfile
-     ret_local=\${PIPESTATUS[0]}
+     compt=`file $1`
```
immarvin: the variable  compt should be "local"

penguhyang:
@immarvin Files have been corrected appropriately.

xCAT-server/share/xcat/install/scripts/post.ubuntu
```diff
@@ -159,8 +159,11 @@ run_ps () {
...
+     if [ $compt =~ "shell script" ]; then
```
immarvin: the pattern matching in bash should be ``[[  "compt " =~ PATTERN  ]]``

immarvin:
the code looks fine, I agree to merge it

PR #230, opened 2015-09-29
Ting

tingtli:


daniceexi:
@tingtli  you need to descriptions when create a pull request.

whowutwut:
Also should change the title to be something more descriptive

junxiawang:
1.please add description for each testcase
2.only one testcase setup synclist folder but each test case should add the folder : mkdir -p /install/custom/install/__GETNODEATTR($$CN,os)__/  
3.only diskfull testcase covered about syncfile,please  diskless test case about sync
4.please add test case for servicenode
5.please add test case for syncing File periodically

samveen:
@junxiawang   Please don't merge requests that still need work (as per your comments).

CC: @whowutwut @daniceexi

whowutwut:
Yes, completely agree.   I'm not sure exactly why this was merged in since there were comments made...  I wonder if it was hit by accident?  The changes were for autotesting framework.... but really we should address the issues in the conversations before we merge a pull request.  

Some members of the team are on holiday until 10/12 so may not get a response for a bit...

tingtli:
@junxiawang 
Please see my response to your comments.

1.please add description for each testcase
LTT: The description is optional for each case. Anyway, I will add description for each case. 

2.only one testcase setup synclist folder but each test case should add the folder : mkdir -p /install/custom/install/GETNODEATTR($$CN,os)/
LTT: I don't think other cases need to create this folder.

3.only diskfull testcase covered about syncfile,please diskless test case about sync
LTT: The cases only cover syncfile cases on diskful node with updatenode scenarios. If we need to add diskless cases. the cases are not added in the updatenode folder.  We need to put it in the installation cases. It is in low priority. If we think we need it, we can have a new task to trace it. 

4.please add test case for servicenode
LTT: service node is also a compute node, i don't think we need add cases for servicenode.

5.please add test case for syncing File periodically
LTT: We use crontab to Sync files periodically. It put the updatenode scenarios in crontab list. crontab is linux system functions.  Since we have already cover sync file functions itself. We don't need to test linux system functions.

PR #231, opened 2015-09-29
Cleanup buildcore.sh file in preparation for xcat.org 

whowutwut:
Made some changes mostly to get the yum repo paths updated for xcat.org new directory structure. 
The other changes are to make things better when developers want to build locally.

### DIFF
--- a/buildcore.sh
+++ b/buildcore.sh
@@ -39,7 +39,7 @@
 UPLOADUSER=litingt
 USER=xcat
 SERVER=xcat.org
-FILES_PATH="files_new"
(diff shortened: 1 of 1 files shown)


whowutwut:
Ting Ting, I created the autobuild framework to do the  daily/weekly/monthly builds and upload them as snap builds to xcat.org... 

I think we should actually remove the upload/promote concepts from buildcore.sh and leave that as just a build script for xCAT.

whowutwut:
@tingtli  going to merge this so I can run some tests in the next week or so.

PR #237, opened 2015-09-30
Fixed some of the WARNING and ERROR messages in the document build. 

whowutwut:
When doing a 'make clean' then 'make html' for the documentation. 
There are LOTS and LOTS of formatting errors.  The 'not included in toctree' 
is something that we know about ... but there are Lots of formatting errors.

I'm doing a cleanup for this time, and I think we should all work to look at the build output 
before we submit a pull request.  

Some of the links were broken after developers added new pages and renamed pages without looking at the pages that reference the old ones... 

Lots of syntax errors for wrapping code blocks with ::  needs to have spaces


whowutwut:
Since this is really for clean up of some ERROR/WARNING msg in the docs,  I will plan to merge it

PR #238, opened 2015-09-30
Adding a new script that will use pod2rst to create the rst files to put in docs

whowutwut:
Check in the initial pages that are generated from the pods

Once these are in, will look into how to auto-push changes from the xCAT build scripts
when POD files are changed.  This seems to be the quickest way to get the man pages
into the documentation at this point.


daniceexi:
The script create_man_pages.py looks good for me.

daniceexi:
@whowutwut  
So currently our decesion is that the one who make the xcat build should manually run the script 'create_man_pages.py' to generate .rst format of man page files and then manually push it to xcat-core repository.

In the future we may add code in buildcore to do the push automatically or just give our a prompt to the builder that there's something need to do to push the .rst files to remote repository.

whowutwut:
Yea I'm not sure what the best way is.... Currently I created a autobuild framework that will build xcat-core using a cron job. So really the only manual intervention is to stop the cron job at the end of the release and make one of the snap builds GA.   

I can open an issue to look into the automatic push to GitHub or we can ask the developer who is making changes to the pod files to run the create script and also push the rst file  in the same pull request. Maybe the code reviews will catch this...

daniceexi:
Temporarily, we can just make owner who change the pod file to update the rst file. And the reviewer will guarantee the miss.

PR #242, opened 2015-10-01
Fix #218,  Check file rsyslogd.pid before issuing kill on the pid

### LINKED ISSUE #218: [FVT]2.11: xCAT-* rpm installation on sles11.3 returns  rsyslogd.pid warning message, please remove it.
Please remove following warning message when xCAT-* rpm installation. 

Installing: xCAT-test-4:2.11-snap201509230241 [done]
Installing: xCAT-buildkit-4:2.11-snap201509230242 [done]
Installing: xCAT-client-4:2.11-snap201509230241 [done]
Installing: xCAT-server-4:2.11-snap201509230241 [done]
Installing: xCAT-2.11-snap201509230242 [done]
Additional rpm output:
/var/tmp/rpm-tmp.12746: line 21: /var/run/rsyslogd.pid: No such file or director                                                                                        y
Restarting xCATd ..done
Running '/opt/xcat/sbin/mknb ppc64', triggered by the installation/update of xCA                                                                                        T-genesis-scripts-ppc64 ...
Creating genesis.fs.ppc64.gz in /tftpboot/xcat
The 'mknb ppc64' command completed successfully.

whowutwut:
Added a test for rsylogd.pid before just assuming that is the fileanme on the system.
The older systems is syslogd.pid.  Only of those files are there do we issue a kill -HUP on it. 

Fix Issues #218

### DIFF
--- a/xCAT/debian/postinst
+++ b/xCAT/debian/postinst
@@ -46,7 +46,11 @@ case "$1" in
 
 	/etc/init.d/apache2 restart
 
-	kill -HUP $(</var/run/rsyslogd.pid) >/dev/null 2>&1 || :
+        if [ -e /var/run/rsyslogd.pid ]; then
+            kill -HUP $(</var/run/rsyslogd.pid) >/dev/null 2>&1 || :
+        elif [ -e /var/run/syslogd.pid ]; then
+            kill -HUP $(</var/run/syslogd.pid) >/dev/null 2>&1 || :
+        fi
     ;;
 
     abort-upgrade|abort-remove|abort-deconfigure)
--- a/xCAT/debian/postinst
+++ b/xCAT/debian/postinst
@@ -46,6 +46,7 @@ case "$1" in
 
 	/etc/init.d/apache2 restart
 
+        # Let rsyslogd perform close of any open files
         if [ -e /var/run/rsyslogd.pid ]; then
             kill -HUP $(</var/run/rsyslogd.pid) >/dev/null 2>&1 || :
         elif [ -e /var/run/syslogd.pid ]; then
--- a/xCAT/xCAT.spec
+++ b/xCAT/xCAT.spec
@@ -201,7 +201,7 @@ then
    cp /etc/xcat/conf.orig/xcat.conf.apach24 /etc/apache2/conf.d/xcat.conf
 fi
 
-# Lets rsyslogd perform close all open files
+# Let rsyslogd perform close of any open files
 if [ -e /var/run/rsyslogd.pid ]; then
     kill -HUP $(</var/run/rsyslogd.pid) >/dev/null 2>&1 || :
 elif [ -e /var/run/syslogd.pid ]; then
--- a/xCAT/xCAT.spec
+++ b/xCAT/xCAT.spec
@@ -202,7 +202,11 @@ then
 fi
 
 # Lets rsyslogd perform close all open files
-kill -HUP $(</var/run/rsyslogd.pid) >/dev/null 2>&1 || :
+if [ -e /var/run/rsyslogd.pid ]; then
+    kill -HUP $(</var/run/rsyslogd.pid) >/dev/null 2>&1 || :
+elif [ -e /var/run/syslogd.pid ]; then
+    kill -HUP $(</var/run/syslogd.pid) >/dev/null 2>&1 || :
+fi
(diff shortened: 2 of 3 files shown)


whowutwut:
Assigned  to @neo954  for review

xCAT/debian/postinst
```diff
@@ -46,7 +46,11 @@ case "$1" in
 
 	/etc/init.d/apache2 restart
 
-	kill -HUP $(</var/run/rsyslogd.pid) >/dev/null 2>&1 || :
+        if [ -e /var/run/rsyslogd.pid ]; then
+            kill -HUP $(</var/run/rsyslogd.pid) >/dev/null 2>&1 || :
```
samveen: From the [supported OS list](http://xcat-docs.readthedocs.org/en/latest/xcat2_support_list.html#rhels-os-support-label), I can make out that all supported versions have Bash-4.0 or higher as part of base system, and  as per [Bash 4.0 release notes](https://lists.gnu.org/archive/html/info-gnu/2009-02/msg00018.html),
```
cc. There is a new &>> redirection operator, which appends the standard output
    and standard error to the named file.
```
It's probably better to change all instances of `>/dev/null 2>&1` to `&>>/dev/null`

whowutwut: @samveen 

Good suggestion... let me think about this some more before we merge this.  I think moving forward it's a good change but I'm worried that we may come across shells that do not support this...  (maybe unlikely)..   Is the ``2>&1`` more readable/understandable?

I think the supported OS list needs some updating before we release 2.11 ..  I did look into the ``&>>`` and it looks fairly safe.  The older ``&>`` command seems to be deprecated, per this site: http://wiki.bash-hackers.org/scripting/obsolete

samveen: @whowutwut 
As everything is usually  "appended" to `/dev/null`, that's why I suggested `&>>` instead of `&>`. However it doesn't make a difference either way when we're talking about `/dev/null`. I expect it boils down to personal preferences in this case.

daniceexi:
uniformly using '>/dev/null 2>&1' is good in xCAT. We also still need to consider that aix might still be supported for a while.

david-shapiro:
Victor, these are the symptoms. The BAR errors caused the runtime errors

c71f3n8-n11 all have messages like this in the log at boot time which would impact gpu performance  

[Thu Oct  8 13:38:34 2015] pci 0003:02:01.0: BAR 13: no space for [io  size 0x1000]
[Thu Oct  8 13:38:34 2015] pci 0003:02:01.0: BAR 13: failed to assign [io  size 0x1000]
[Thu Oct  8 13:38:34 2015] pci 0003:02:0a.0: BAR 13: no space for [io  size 0x1000]
[Thu Oct  8 13:38:34 2015] pci 0003:02:0a.0: BAR 13: failed to assign [io  size 0x1000]

Lots of messages like this when gpu jobs are running which would impact the performance
[Fri Oct  9 05:18:46 2015] NVRM: the requested mapping cannot be fulfilled because the address (0x3b0000000000) or size (0x4000) is not aligned to a page boundary!

To work around after the xcat (using image xcat_compute_1541a)
 Mark uninstalled and re-installled cuda on one node (still bad 
compilers) and it looks good so far.  No Bar error messages at 
boot and no NVRM messages during the run
The job will run a long time but I think its safe to say the 
the xCat NVIDA install is still not workiing

whowutwut:
@david-shapiro , You are commenting in the wrong issue..

whowutwut:
@neo954  @daniceexi  , if okay with the changes, please merge the Pull request.

daniceexi:
@whowutwut I think @neo954 missed the change for xCATsn.spec. Could you fix that?

whowutwut:
@daniceexi  Thanks , I added it.  
Can you explain the syntax ``||:`` at the end of those lines?

daniceexi:
@whowutwut 
: means 'true' in shell, so ||: means if the first command fail, then return 'true' still. This is used to make sure this line won't return false.

refer to the explaination here: http://stackoverflow.com/questions/3224878/what-is-the-purpose-of-the-colon-gnu-bash-builtin

daniceexi:

whowutwut:
@daniceexi  Thanks for the explanation and link!

PR #245, opened 2015-10-07
Reorganize the Advanced section and topics in the documentation

whowutwut:
Reorganizing the entire 'Advanced' section of the documentation. 

Hopefully establish a "good practice" when writing the docs.  I've removed "xCAT" from the title of the topics since this site only discusses xCAT.   Created categories and place documentation under those categories. Also split up some of the larger docs into smaller multiple docs.  This allows the user to focus on a specific task , then click "Next" without looking at a documentation for pages with multiple sub topics.   More of this is needed to be done.


cxhong:
looks good!

PR #248, opened 2015-10-08
failed to check postgresql package and ids

cxhong:
Noticed on some of test system, the postgresql-libs was installed which cause pgsqlsetup scripts bypass the checking of "rpm -qa | grep postgresql" ,  "rpm -q postgresql"  works better for this case.  

"yum install postgresql"  will only installed postgresql-9.2...... and postgresql-libs....,   the postgres id will not created unless the postgresql-server is installed.   Add a checking for postgres id in the pgsqlsetup scripts to make sure all required dependency are set.

### DIFF
--- a/xCAT-client/bin/pgsqlsetup
+++ b/xCAT-client/bin/pgsqlsetup
@@ -164,7 +164,7 @@ if ($::RUNCMD_RC != 0)
 #
 # check to see if postgresql is installed
 #
-my $cmd = "rpm -qa | grep postgresql";
+my $cmd = "rpm -q postgresql";
 if ($debianflag){
     $cmd = "dpkg -l | grep postgresql | awk '{print \$2}'";
 }
@@ -179,6 +179,20 @@ if ($::RUNCMD_RC != 0)
     xCAT::MsgUtils->message("E", " $cmd failed. $message");
     exit(1);
 }
+
+#
+#check to see if postgres user is created
+#
+my $cmd = "id postgres";
+my @output=xCAT::Utils->runcmd($cmd, 0);
+if ($::RUNCMD_RC != 0)
+{
+    my $message = "\nPostgreSQL user is not created, please make sure postgresql-server packages are installed.";
+
+    xCAT::MsgUtils->message("E", " $cmd failed. $message");
+    exit(1);
+}
+
 # check if 9.X release not built by us is installed,  setup different
 # SLES used default dir
 if  ( (grep(/postgresql9/, @output)) && ($::linuxos !~ /sles/) ){  # postgresql 9.x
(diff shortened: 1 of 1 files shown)


whowutwut:
Can we show the actual error in the description?

xCAT-client/bin/pgsqlsetup
```diff
@@ -164,7 +164,7 @@ if ($::RUNCMD_RC != 0)
 #
 # check to see if postgresql is installed
 #
-my $cmd = "rpm -qa | grep postgresql";
+my $cmd = "rpm -q postgresql";
```
whowutwut: My thought here is that we should search for ``postgresql-server``.    That should be required to start the PostgreSQL database if the user is switching from SQLite to PostgreSQL....

This postgresql-server rpm package requires postgresql and postgresql-libs.

xCAT-client/bin/pgsqlsetup
```diff
@@ -164,7 +164,7 @@ if ($::RUNCMD_RC != 0)
...
 if ($debianflag){
     $cmd = "dpkg -l | grep postgresql | awk '{print \$2}'";
```
whowutwut: Maybe need to also check on Ubuntu what the pkg name is.... ?

xCAT-client/bin/pgsqlsetup
```diff
@@ -179,6 +179,20 @@ if ($::RUNCMD_RC != 0)
     xCAT::MsgUtils->message("E", " $cmd failed. $message");
     exit(1);
 }
+
+#
```
whowutwut: If we check for postgresql-server... maybe this addition is no longer needed because will error above...

cxhong:
for both case, the error message will show:
.....................................................................................................................................
su: user postgres does not exist
sh: /var/lib/pgsql/data/pg_hba.conf.xcat: No such file or directory
Command failed: awk '{gsub("IPv4 local connections:","IPv4 local connections:\nhost    all          all        10.5.33.254/32      md5  "); print}'   /var/lib/pgsql/data/pg_hba.conf > /var/lib/pgsql/data/pg_hba.conf.xcat 2>&1. Error message: .

 awk '{gsub("IPv4 local connections:","IPv4 local connections:\nhost    all          all        10.5.33.254/32      md5  "); print}'   /var/lib/pgsql/data/pg_hba.conf > /var/lib/pgsql/data/pg_hba.conf.xcat failed.
................................................................................................................................................

Yeh, We should install postgresql-server first before running pgsqlsetup, but sometime user will run it without follow the documentation.  This error message will help to fast determine what the issues are.

cxhong:
yeh, thanks . postgresql-server will catch all.  Ubuntu didn't have this issue.

whowutwut:
Great!

Issue #233, opened 2015-09-30
[FVT]: 2.11 rinv issues summary on IBM Power S822LC RHEL 7.2


tingtli:
1) rinv NODE all
```
[root@c910f03c05k08 /]# rinv c910f05c37 all
c910f05c37: Error: Received incorrect data from BMC
c910f05c37: Error: Received incorrect data from BMC
c910f05c37: BMC Firmware: 2.06
c910f05c37: System Description: OpenPOWER Firmware
...
c910f05c37: System Firmware Product Additional Info: hostboot-binaries-4
c910f05c37: System Firmware Product Additional Info: firestone-xml-9163b
c910f05c37: System Firmware Product Additional Info: capp-ucode-105cb8f
c910f05c37: Error: Received incorrect data from BMC
[root@c910f03c05k08 /]# echo $?
1
```
2) rinv NODE pci     ("rinv X-node pci" has no output) 
```
[root@c910f03c05k08 /]# rinv c910f05c37 pci
c910f05c37: Error: Received incorrect data from BMC
c910f05c37: Error: Received incorrect data from BMC
c910f05c37: Error: Received incorrect data from BMC
```
3) rinv NODE model 
```
[root@c910f03c05k08 /]# rinv c910f05c37 model
c910f05c37: System Description: OpenPOWER Firmware
```
Here is no model information for Firestone.  For X node, there is output.
```
c910f04x31:~ # rinv c910f04x35 model
c910f04x35: System Description: iDataPlex dx360 M4
c910f04x35: System Model/MTM: 7912AC1
```

4) rinv NODE serial 
```
[root@c910f03c05k08 /]# rinv c910f05c37 serial
```
Nothing output!

X-Node has output
```
c910f04x31:~ # rinv c910f04x35 serial
c910f04x35: System Serial Number: 06CGTR0
c910f04x35: Chassis Serial Number: 06CGTR0
```
5) rinv NODE asset 
```
[root@c910f03c05k08 /]# rinv c910f05c37 asset
```
nothing output

X-Node has output, but no number is displayed.
```
c910f04x31:~ # rinv c910f04x35  asset
c910f04x35: System Asset Number:
```
6) rinv NODE vpd 
```
[root@c910f03c05k08 /]# rinv c910f05c37 vpd
c910f05c37: Manufacturer ID: 0
c910f05c37: Device ID: 32
c910f05c37: BMC Firmware: 2.06
c910f05c37: System Description: OpenPOWER Firmware
[...]

PR #250, opened 2015-10-09
Fix for xcat-core/issues/233, rinv recieve incorrect data from BMC

whowutwut:
Commit 317d7cf contains the fix for Issue #233 , Running rinv results in the following messages coming back:
```
Error: Received incorrect data from BMC
```

The array contains uninitialized values that are not 0 or 1, so check if defined before processing the value in the array.  If not defined, return undef

The other commits are to fix some of the tab/space formatting in ipmi.pm.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -2207,66 +2207,66 @@ sub beacon_answer {
 
 sub inv {
     my $sessdata = shift;
-	my $subcommand = $sessdata->{subcommand};
+    my $subcommand = $sessdata->{subcommand};
 
-	my $rc = 0;
-	my $text;
-	my @output;
-	my @types;
+    my $rc = 0;
+    my $text;
+    my @output;
+    my @types;
 
 
     unless ($subcommand) {
         $subcommand = "all";
     }
-	if($subcommand eq "all") {
-		@types = qw(model serial deviceid mprom guid misc hw asset firmware mac wwn);
-	}
-	elsif($subcommand eq "asset") {
+    if($subcommand eq "all") {
+        @types = qw(model serial deviceid mprom guid misc hw asset firmware mac wwn);
+    }
+    elsif($subcommand eq "asset") {
         $sessdata->{skipotherfru}=1;
-		@types = qw(asset);
-	}
-	elsif($subcommand eq "firm" || $subcommand eq "firmware") {
+        @types = qw(asset);
+    }
+    elsif($subcommand eq "firm" || $subcommand eq "firmware") {
         $sessdata->{skipotherfru}=1;
-		@types = qw(firmware);
-	}
-	elsif($subcommand eq "model") {
+        @types = qw(firmware);
+    }
+    elsif($subcommand eq "model") {
         $sessdata->{skipotherfru}=1;
-		@types = qw(model);
-	}
-	elsif($subcommand eq "serial") {
+        @types = qw(model);
+    }
+    elsif($subcommand eq "serial") {
         $sessdata->{skipotherfru}=1;
-		@types = qw(serial);
-	}
-	elsif($subcommand eq "vpd") {
+        @types = qw(serial);
+    }
+    elsif($subcommand eq "vpd") {
         $sessdata->{skipotherfru}=1;
-		@types = qw(model serial deviceid mprom);
-	}
-	elsif($subcommand eq "mprom") {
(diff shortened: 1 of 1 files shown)


whowutwut:
This doesn't address all the issues in #233, will close for now and re-open if I do figure out how to fix the other issues.

whowutwut:
@zet809  There are some issues with the ``rinv`` command on the Power S822LC machines that we have because of the problem with the FRU.  I think we need to open a ticket and get the FRU unit replaced ( @pdlun92 is working on that...  ) before we are able to successfully print out inventory information with ``rinv``.  Maybe if we can find a machine that has a working FRU, we can get onto there to do more devel work and possibly fix this in 2.11.1 if we can't get it fixed in 2.11.  

But for now, it seems to me that we are receiving data and if it's less than the chunksize, we print out the msg, but I think at the end of the data, the chunk received may be less than the size set... so are we printing out a msg that should not be printed out?  Can you review...? 

The code changes I made are in commit 317d7cf and 400711b.  The other 3 commits are fixing the space and formatting issues in ipmi.pm.  

If you don't agree. I can look further.....  

I think the issues @tingtli  brings up about the missing data can't be resolved until we fix the FRU unit...

zet809:
@whowutwut I agree with you that not to output the error msg until we fix the FRU issue. 
Print out err msg for rinv or rvitials help us nothing but confusing users.

whowutwut:
@zet809  if the changes are acceptable, please merge the pull request. Thanks!

whowutwut:
@daniceexi @zet809 @pdlun92  There's a way to fix the FRU, I'm going to look into that, please don't merge this pull request until I see if we can fix the FRU information on the 822LC...

cxhong:
regarding this issue,  the total fru size return back from fru inventory area info command (0x10) is 256.  for each fru id, we read FRU Data (command 0x11) from offset 0 to one chunk at a time, each chunk is 16 bytes, util we finish read all the 256 bytes.  But for some fru,  we asked for to read 16 bytes, but only have less than 16 bytes left,  I don't think this is incorrect data from BMC. so the code
````
       my @data = @{$rsp->{data}};
        if ($data[0] != $sessdata->{currfruchunk}) {
            add_fruhash($sessdata);
            xCAT::SvrUtils::sendmsg([1,"Received incorrect data from BMC"],$callback,$sessdata->{node},%allerrornodes);
            return;
        }
```````
I suggest to change to:
`````
       my @data = @{$rsp->{data}};     
        if ($data[0] < $sessdata->{currfruchunk}) {
            $chunk = $data[0];
            $sessdata->{currfrudone}=1;
        }
````````````
so we shrink chunk to only get the remainder data.

File: xCAT-server/lib/xcat/plugins/ipmi.pm [hunk 1]
```diff
@@ -545,98 +544,101 @@ sub on_bmc_connect {
...
+    }
+    else {
+        $rc = 1;
+        $text = "unsupported command $command $subcommand";
+    }
 
-	if($text) {
-		push(@output,$text);
-	}
+    if($debug) {
```
zet809: There is a "return" in line 540, so this section won't work at all.

zet809:
@cxhong The currfruchunk is calculated with the code below, so it must be fru error if returned data length not equal to currfrusize. 

if ($sessdata->{currfruoffset}+16  >= $sessdata->{currfrusize}) {
        $chunk =  $sessdata->{currfrusize}-$sessdata->{currfruoffset}; # shrink chunk to only get the remainder data
        $sessdata->{currfrudone}=1;
    }

whowutwut:
After some discussion... For the 2.11 GA, will modify the ``rinv`` incorrect data message to be more desciptive and contain the FRU ID.  When we get access to a Power S822LC machine that has a good VPD info from MFG, will address this issue further and see if it's resolved.

New output will be as follows:
```
# rinv c910f05c33 bios
c910f05c33: Received incorrect data from BMC for FRU ID: 52
c910f05c33: Received incorrect data from BMC for FRU ID: 44
c910f05c33: Received incorrect data from BMC for FRU ID: 49
c910f05c33: Received incorrect data from BMC for FRU ID: 47
```

@zet809 , can you review the changes in the last two commits for this pull request 317d7cf08eb01f68be66d17bbf5f98888975ebed and 743960c5a81d8fc495ba6fdfa91a24595174f084 and merge the request if OK.  The others are formatting related.

zet809:
The code looks good to me, I agree to merge it.

Issue #252, opened 2015-10-13
[FVT]:2.11 modify buildcore.sh in prior releases to upload xcat package to xcat.org


tingtli:
This issue is opened to trace the work to modify buildcore.sh in prior release such as 2.7, 2.8 and 2.9 etc to upload xcat packages to xcat.org.

whowutwut:
@tingtli I am going to do a demo of the new autobuild framework in our meeting on Thurs. 

The autobuild framework handles the uploading of the builds for xcat.org.  There's a separate script that we can run on xcat.org that will "promote" builds to GA.  This removes the complexity of the code in the buildcore.sh script.  I don't think I removed the code in buildcore.sh yet though.  I wanted the team to take a look and decide before deleting the code. 

However, I do think that for the older release, we should fix the buildcore.sh for 2.7 and 2.8, if needed.  The 2.9 build can be handled by the autobuild scripts.

tingtli:
I will update xCAT 2.7 and 2.8 first. After you introduce the autobuild framework, I will work on 2.9

tingtli:
@whowutwut  I am thinking if we need to modify 2.7, 2.8 and 2.9 to upload the build to xcat.org.
It seems there is almost no customers to use this build.  And the customers who use the build probably are farmilar with SF.  And the release note, xcat-dep is also on SF. 
Do you think we need to update the script?

whowutwut:
@tingtli  Are you saying we should just leave 2.7 and 2.8 uploading to SF for now until we see a greater demand for builds ?  You're right, maybe no one is using this anymore to make the changes..

daniceexi:
@whowutwut @tingtli I agree we don't need to put effort for 2.7/2.8 build script, we can just manually copy the released files to xcat.org if necessary.

whowutwut:
Sounds good, I closed the 2.7 pull request, we can close out this issue.

PR #259, opened 2015-10-14
modify upload location to xcat.org for xCAT 2.7

tingtli:
modify upload location to xcat.org for 2.7

### DIFF
--- a/buildcore.sh
+++ b/buildcore.sh
@@ -34,7 +34,11 @@
 
 # you can change this if you need to
 UPLOADUSER=litingt
-FRS=/home/frs/project/x/xc/xcat
+USER=xcat
+SERVER=xcat.org
+FILES_PATH="files"
+FRS=/var/www/${SERVER}/${FILES_PATH}
+RELEASE=github.com/xcat2/xcat-core/releases
 
 # Process cmd line variable assignments, assigning each attr=val pair to a variable of same name
 for i in $*; do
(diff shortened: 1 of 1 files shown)


tingtli:
@whowutwut  Hi victor, I just modify 2.7 upload location to xcat.org.
I did the least modification.Please help review it.

whowutwut:
This pull request is in reference to issue #252

buildcore.sh
```diff
@@ -34,7 +34,11 @@
 
 # you can change this if you need to
 UPLOADUSER=litingt
-FRS=/home/frs/project/x/xc/xcat
+USER=xcat
+SERVER=xcat.org
+FILES_PATH="files"
+FRS=/var/www/${SERVER}/${FILES_PATH}
+RELEASE=github.com/xcat2/xcat-core/releases
```
whowutwut: Are we going to upload 2.7 to github? I think this is not being used but I guess we can leave it in for now to be consistent with the other buildcore.sh 

Or is github release going to point to xcat.org... maybe one less place to maintain if we do it that way.... 
@daniceexi  advice?

whowutwut:
I am ok with the changes

whowutwut:
@tingtli  Do we think we will be creating the 2.7 build?   If so, then I will merge this in...

whowutwut:
Decided in #252 that we will not modify for the 2.7 release, closing for now.

PR #267, opened 2015-10-15
Fix bug#155 There are errors when restart xcatd for UDP service's bug

hu-weihua:
Fix bug#155 There are errors when restart xcatd for UDP service's bug

### DIFF
--- a/xCAT-server/sbin/xcatd
+++ b/xCAT-server/sbin/xcatd
@@ -728,7 +728,8 @@ sub do_udp_service { #This function opens up a UDP port
 	   $part = $socket->recv($data,2000);
            $packets{$part} = [$part,$data];
            } elsif ($hdl == $sslctl) { 
-	   	update_udpcontext_from_sslctl(udpcontext=>$udpcontext,select=>$select);
+               next;
+               #update_udpcontext_from_sslctl(udpcontext=>$udpcontext,select=>$select);
            } elsif ($hdl == $discoctl) {  #got a discovery response....
            } else { 
 		print "Something is wrong in udp process (search xcatd for this string)\n";
@@ -948,7 +949,7 @@ $SIG{TERM} = $SIG{INT} = sub {
       kill 2, $_;
    }
    if ($pid_UDP) {
-      kill 2, $pid_UDP;
+      kill 12, $pid_UDP;
    }
    if ($pid_MON) {
       kill 2, $pid_MON;


hu-weihua:
@daniceexi  
For the "Why you have to add a 'next' sentence" question,
Yes. it's not necessary to add 'next'. the key point is we have to keep "elsif ($hdl == $sslctl)" line, otherwise the code will output error message due to we do have "$sslctl" in "@hdls".  if without 'next', that will be blank between '{' and '}'  . that's fine, the code works correctly. but it's little confuse for reader to understand about what's the purpose of this 'elsif' branch.

For the "kill 2' problem:
Actually,  the fork sequence is  "main process" -> "UDP process" -> "Discovery process"
when we stop xcatd, the main process will save "TERM" signal, below code will be triggered  

    942 $SIG{TERM} = $SIG{INT} = sub {
    ....
    951    if ($pid_UDP) {
    952       kill 12, $pid_UDP;
    953    }
    .....
    971    };

then UDP process will receive signal, in fact, in UDP process, it just handles signal "USR2" (i.e. signal 12), not handle other signal else,  that's why I changed the code from **"kill 2, $pid_UDP"** to **"kill 12, $pid_UDP"** 

    647   $SIG{USR2} = sub {
    648      if ($socket) {
    649          #only clear out pid file when we still have socket.
    650          unlink("/var/run/xcat/udpservice.pid"); close($socket); $quit=1; $socket=0;
    651          $udpctl=0;
    652          xCAT::MsgUtils->message("S","xcatd udp service $$ quiescing");
    653      }
    654      kill(15,$discopid);
    655   };

As what the code show above, when UDP process handle "USR2" signal, it will send "TERM" (i.e. signal 15) to Discovery process, when Discovery process receive "TERM", below code will be run

    566 sub do_discovery_process {
    567    $SIG{TERM} = 'DEFAULT';
    568    $SIG{INT} = 'DEFAULT';
     ......

up to now, "Discovery process",  "UDP process" and "main process" will stop one by one.

hu-weihua:
sorry @daniceexi 
I am wrong for UDP process doesn't handle other signal except "USR2",  it also handle TERM and INI, the problem is when it handle "TERM and INI", it didn't handle all necessary key attributes. that cause the code unsafe stop, so we should still send "12" to UDP process but not '2'

daniceexi:
After discuss with Wei Hua, I agree the change is good for the bug 155.

PR #258, opened 2015-10-14
add cuda installation based on ubuntu document

penguhyang:
See the page from my local web server: http://10.2.9.4/advanced/gpu/index.html

PR #272, opened 2015-10-16
Revert "add cuda installation based on ubuntu document"

whowutwut:
Reverts xcat2/xcat-core#258


whowutwut:
@penguhyang   I had merged pull request #258 because I wanted to re-organize the pages, but I think the responsibility should be on the developer adding the pages to not create duplicate pages, so I am un-doing the merge.

whowutwut:
To follow up a bit, the result of pull request #258 is duplication of the CUDA documents

image


We have a lot of the same contents.. the overview should be shared and we should have sections for Ubuntu and RH if they are different enough.

PR #277, opened 2015-10-19
Handle ',,' in nodelist

jjohnson42:
If ',,' is in nodelist, then confluent is
told to deal with empty groups.  Filter
that phenomenon out.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/confluent.pm
+++ b/xCAT-server/lib/xcat/plugins/confluent.pm
@@ -362,7 +362,7 @@ sub donodeent {
   }
   my @cfgroups;
   foreach (keys %$groupdata) {
-      push @cfgroups, split /,/,$groupdata->{$_}->[0]->{groups};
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/confluent.pm
```diff
@@ -419,7 +419,7 @@ foreach my $node (sort keys %$cfgenthash) {
   } elsif ($::XCATSITEVALS{'consoleondemand'} and $::XCATSITEVALS{'consoleondemand'} !~ m/^n/) {
     $parameters{'console.logging'} = 'none';
   }
-  $parameters{'groups'} = [split /,/,$groupdata->{$node}->[0]->{'groups'}];
+  $parameters{'groups'} = [grep {defined && length} split /,/,$groupdata->{$node}->[0]->{'groups'}];
```
samveen: I live and learn.

PR #293, opened 2015-10-22
correct doc for word and syntax

penguhyang:
I found some word and syntax needs to correct.
Test doc link: http://10.2.9.4/overview/index.html


whowutwut:
Did something happen to this pull request?  Intended to be closed?

penguhyang:
This pull request is not in the right way because it has two mixed up commits. 
@whowutwut Is there some kind of method to delete the invalid pull request ?

daniceexi:
@whowutwut 
@penguhyang closed 293 and opened 294 to submit this change.
I think close an unmerged pull request should not make any harm.

whowutwut:
Ok, yep, I think it's okay,
I didn't  know if it was accidentally closed or if we should review it. :-)

PR #294, opened 2015-10-22
correct doc for word and syntax

penguhyang:
I found some word and syntax needs to correct.


daniceexi:

@penguhyang Thanks for the doc change.

PR #295, opened 2015-10-22
upgrade the cuda pkglist for diskfull install

penguhyang:
add linux-headers-generic-lts-utopic to install nvidia-346-uvm


daniceexi:

PR #301, opened 2015-10-23
add flag -t for bmcdiscovery to generate a temp bmc node

zet809:


zet809:
Modified, pls review again.

whowutwut:
I see a .pod file change, can you run  ``create_man_pages.py`` to get the associated .rst page change and include in the pull request?

perl-xCAT/xCAT/Utils.pm
```diff
@@ -4432,7 +4432,14 @@ sub cleanup_for_powerLE_hardware_discovery {
...
         xCAT::MsgUtils->message("S", "Discovery info: configure password for pbmc_node:$pbmc_node.");
         if (defined($new_bmc_username) and $new_bmc_username ne '') {
-            `rspconfig $pbmc_node username=$new_bmc_username password=$new_bmc_password`;
+            if ($new_bmc_username eq "ADMIN" or $new_bmc_username eq 'USERID') {
+                # ADMIN is username for OpenPOWER server, it is not allowed to modify at present
+                # USERID is username for IBM system x server, just modify password
+                `rspconfig $pbmc_node userid=2 password=$new_bmc_password`;
+            } else {
+                # For other username, we'd better create new user for them
+                `rspconfig $pbmc_node userid=3 username=$new_bmc_username password=$new_bmc_password`;
```
whowutwut: What happens if this cmd is executed another time with a new user name specified?  Do we need to check the userid and increment?

whowutwut:
Just for my information, what is a temp bmc node for?

perl-xCAT/xCAT/Utils.pm
daniceexi: @whowutwut 

This subroutine is called automatically by discovery process to initialize BMC. So even the discovery procedure wil be run again, we till do same thing to initialize.

If customer want to set user by themselves, they should run rspconfig explicitly.

zet809:
For the temp bmc, it is used to do hardware ctrl before the node is identified by xCAT.

daniceexi:

perl-xCAT/xCAT/Utils.pm
whowutwut: I meant since we are creating new user... We have uid static set to 3.... If they call rspconfig with user1 that gets assigned if 3... Then they want to add another user, user2.. We are using uid=3. Won't that fail?

zet809: No, it won't fail. It just means the username for userid=3 had been changed from user1 to user2.
In BMC, the users are identified by userid, not username. It means if you want to create a new user, you must specify a userid that haven't been used before. 

In rspconfig case, the rule is:
1. If no userid, username specified,   changing password will be for user with userid 1
2. If no userid specified, but username specified, changing password will be for userid 2.
3. Else, changing username and/or password for the specified userid.

The reason that we set static 3 for userid is:
1. For firestone BMCs, only username ADMIN can be used for userid = 2. 
2. userid = 3 is free to use for both OperPOWER servers and system X machines.

PR #303, opened 2015-10-23
update hdwr discovery doc with -t flag for bmcdiscover

zet809:


zet809:
Yes, the nodetype and hwtype won't be added by default. It is only added when '-t' flag is specified.

daniceexi:

PR #306, opened 2015-10-23
SLES12 SP1 - Modify pgsqlsetup to support detecting postgresql-server package

cxhong:
For the sles12 SP1,  the postgres server has different name than other OS
````
 # rpm -qa | grep postgres
postgresql94-server-9.4.5-1.4.ppc64le
````
So checking "rpm -q postgresql-server" will not working for sles12 SP1.


whowutwut:
I have reviewed the changes

PR #307, opened 2015-10-24
Added documentation for building rst man pages 

whowutwut:
Rearranged the developers area to group topics together

Added documentation primarily for building the man pages locally
There is a requirement for pod2rst Perl module.  Modified the 
create_man_page.py script to allow for a directory to be specified
containing the Perl modules 

http://10.2.5.2/developer_guide


whowutwut:
Pulled in the instructions for creating the man page using the ``create_man_pages.py`` script if you are runnning the script on a machine that does not allow root access to install the pod2rst perl modules.  Install the perl modules to a local directory and then point the directory to ``create_man_pages.py``

PR #309, opened 2015-10-24
Reorganize the man page section on RTD, add in perl-xCAT pages

whowutwut:
Re-organize the man page section over on ReadTheDocs.

Instead of having all the man pages in a single page, separate them based on the man page number.
Also enhanced the create_man_pages.py script to execute db2man

http://10.2.5.2/db2man/guides/admin-guides/references/index.html#xcat-man-pages


daniceexi:

PR #311, opened 2015-10-24
add kvm features: cpu pining, memory binding and pci passthrough

immarvin:
add kvm features:cpu pining,memory binding,pci pass through
supported commands: mkvm,chvm(online/offline)
chvm manpage, vm.othersettings schema
fix issue:"virsh create" failed with "error: internal error: qemu does not support SGA"


xCAT-client/pods/man1/chvm.1.pod
```diff
@@ -215,6 +223,30 @@ To delete physical slots which are specified by the I<drc_index1,drc_index2...>.
 
 To delete a virtual adapter specified by the I<slotid>.
 
+=item B<--cpupin hostcpuset>
+
+To pin guest domain virtual CPUs to physical host CPUs specified with I<hostcpuset>.
```
whowutwut: Suggest: 

Pinning  virtual CPUs to physical CPUs can be done by using the --cpupin option. 
*hostcpuset* is a comma separated list of physical CPU numbers and special markup
using '-' and '^'.  '-' is used to specify a range of CPUs and ^' denotes exclusive.

Then we should add some example...   i don't understand the difference between 0-15,^8 vs ^8,0-15..

xCAT-client/pods/man1/chvm.1.pod
```diff
@@ -215,6 +223,30 @@ To delete physical slots which are specified by the I<drc_index1,drc_index2...>.
...
+markup using '-' and '^' (ex. '0-4', '0-3,^2') can also be allowed. The '-' denotes the range and
+the '^' denotes exclusive. 
+
+Note: The expression is sequentially evaluated, so "0-15,^8" is identical to "9-14,0-7,15" but not
+identical to "^8,0-15".
+
+=item B<--membind numanodeset>
+
+It is possible to restrict a guest to allocate memory from the specified set of NUMA nodes I<numanodeset>. 
+If the guest vCPUs are also pinned to a set of cores located on that same set of NUMA nodes, memory
```
whowutwut: suggest:

If the guest vCPUs are also pinned to a set of cores located on that same set of NUMA nodes, memory access ~~will be~~ is local~~, which~~ and improves memory access performance.

xCAT-client/pods/man1/chvm.1.pod
```diff
@@ -215,6 +223,30 @@ To delete physical slots which are specified by the I<drc_index1,drc_index2...>.
...
+access will be local, which improves memory access performance.
+
+=item B<--devpassthru pcidevice1,pcidevice2...>
+
+The PCI passthroug gives a guest VM direct access to I/O devices I<pcidevice1,pcidevice2...>.
```
whowutwut: passthroug -> passthru

xCAT-client/pods/man1/chvm.1.pod
```diff
@@ -215,6 +223,30 @@ To delete physical slots which are specified by the I<drc_index1,drc_index2...>.
...
+
+It is possible to restrict a guest to allocate memory from the specified set of NUMA nodes I<numanodeset>. 
+If the guest vCPUs are also pinned to a set of cores located on that same set of NUMA nodes, memory
+access will be local, which improves memory access performance.
+
+=item B<--devpassthru pcidevice1,pcidevice2...>
+
+The PCI passthroug gives a guest VM direct access to I/O devices I<pcidevice1,pcidevice2...>. 
+The PCI devices are assigned to a virtual machine, and the virtual machine can use this I/O exclusively.
+The devices list are a list of PCI device names delimited with comma, the PCI device names can be obtained
```
whowutwut: I think we should use "Comma separated" like above to be consistent..

"The device list is a comma separated list of PCI device names which can be obtained by running *virsh noddev-list* on the host.

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -440,7 +438,14 @@ sub reconfigvm {
 sub build_oshash {
     my %rethash;
     $rethash{type}->{content}='hvm';
-    $rethash{bios}->{useserial}='yes';
+    #$rethash{bios}->{useserial}='yes';
```
whowutwut: just delete this line?

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -688,7 +693,78 @@ sub build_xmldesc {
     } else {
         $xtree{memory}->{content}=524288;
     }
-    if ($hypcpumodel eq "ppc64") {
+    
+    print "xxx $confdata->{vm}->{$node}->[0]->{othersettings} ";
```
whowutwut: Is this debug line?

whowutwut:
@immarvin  What is SGA?

whowutwut:
I will need some more time to review the code changes, I just reviewed the externals (documentation)

immarvin:
@whowutwut SGA=Serial Graphics Adapter

xCAT-server/lib/xcat/plugins/kvm.pm
immarvin: Some guys from LTC pointed out that the tag <bios useserial="yes" /> doesn't make sense on Power platform as bios is not the firmware.

xCAT-client/pods/man1/chvm.1.pod
immarvin: Victor,  the following sentence is mentioned in the manpage:
````
Note: The expression is sequentially evaluated, so "0-15,^8" is identical to "9-14,0-7,15" but not identical to "^8,0-15".
````
take "0-15,^8" for example, the cpus "#0-#15" are selected first, then cpu #8 is excluded in the cpu selection when "^8" is evaluated, so the selected cpus  should be "0-7" and "9-15". 
For "^8,0-15", "^8" is first evaluated, it makes no sense as cpu #8 is not in the cpu selection at that time, the cpus "#0-#15" are selected, the final selected cpus will be "0-15".

Comment (whowutwut) [inline xCAT-client/pods/man1/chvm.1.pod]: Oh, so should ^8,0-15 be an error? Because you can't exclude a CPU from a list that is not specified first?

Comment (immarvin) [inline xCAT-client/pods/man1/chvm.1.pod]: added a code slice to throw an error message in this case.

File: xCAT-server/lib/xcat/plugins/kvm.pm (hunk 5 above)
Comment (whowutwut) [inline xCAT-server/lib/xcat/plugins/kvm.pm]: yes, i mean should we just remove the line out of code instead of comment out? do we still need it for other VMs not PowerKVM?

**Comment by whowutwut** [comment]
I added some documentation for the process of building the man pages after changing the pod files:  http://xcat-docs.readthedocs.org/en/latest/developers/guides/code/builds.html#man-pages 

The steps are for incase you don't have root access to install the perl modules, you can install them with PREFIX and then run the create_man_pages.py script in xcat-core to generate the rst changes.. 

I'm not sure where your have your xcat-core clone and whether you have root user permission. For example, I don't think we should install into login nodes for c910 

But that will generate the manpage changes for your pod file change and could include into this pull request.

File: xCAT-server/lib/xcat/plugins/kvm.pm [hunk 7]
```diff
@@ -440,7 +438,14 @@ sub reconfigvm {
 sub build_oshash {
     my %rethash;
     $rethash{type}->{content}='hvm';
-    $rethash{bios}->{useserial}='yes';
+    #$rethash{bios}->{useserial}='yes';
+   
+    
+    my $hypcpumodel = $confdata->{$confdata->{vm}->{$node}->[0]->{host}}->{cpumodel};
```
whowutwut: Trying to understand this condition, I don't see "cpumodel" field described in anywhere.. Should that be in Schema.pm?  What should this evaluate to? 

Then at line 445, if the hypcpumodel is "ppc64le", we are going to define the useserial="yes"

But just above we removed it for Power platform??

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -688,7 +693,77 @@ sub build_xmldesc {
...
+         push(@prdevarray,\%tmphash);
+                
+      } 
+    }
+   
+    $xtree{devices}->{hostdev}=\@prdevarray;
+
+
+    if ($hypcpumodel eq "ppc64" or $hypcpumodel eq "ppc64le") {
+    #if ($hypcpumodel eq "ppc64") {
```
whowutwut: Can delete this line, it's no longer used.

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -771,7 +846,7 @@ sub build_xmldesc {
     if (defined($hypcpumodel) and $hypcpumodel eq 'ppc64') {
         $xtree{devices}->{emulator}->{content} = "/usr/bin/qemu-system-ppc64";
     } else {
-        $xtree{devices}->{sound}->{model}='ac97';
+        #$xtree{devices}->{sound}->{model}='ac97';
```
whowutwut: suggest delete the else clause completely. 

} ~~else {
        #$xtree{devices}->{sound}->{model}='ac97';
    }~~

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -688,7 +693,77 @@ sub build_xmldesc {
     } else {
         $xtree{memory}->{content}=524288;
     }
-    if ($hypcpumodel eq "ppc64") {
+    
+    my %cpupinhash;
```
whowutwut: Possible to add some comments here in the code to help understand what is being done?

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -1625,6 +1700,9 @@ sub chvm {
     my $memory;
     my $cdrom;
     my $eject;
+    my $pcpuset;
```
whowutwut: I would almost suggest changing to ``pcpuset`` to ``pincpuset`` for easy readability. 

In vi, can do: `` %s/pcpuset/pincpuset/g``

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -1625,6 +1700,9 @@ sub chvm {
...
+    my $numanodeset;
+    my $prdevices;
```
whowutwut: Similar comment, would suggest changing ``prdevices`` to ``passthrudevices`` for readability. 

vi: ``%s/prdevices/passthrudevices/g``

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -1914,6 +2011,223 @@ sub chvm {
...
+             };
+             if ($@) {
+               xCAT::SvrUtils::sendmsg([1,"$@"],$callback,$node);
+             }
+          }
+
+          $vmxml=$dom->get_xml_description();
+          my $parsed=$parser->parse_string($vmxml);
+          my $ref=$parsed->findnodes("/domain/vcpu");
+          #$ref->[0]->setAttribute("cpuset","$pcpuset");
```
whowutwut: need this line?

xCAT-server/lib/xcat/plugins/kvm.pm
immarvin: 1. for the "cpumodel", its value is initialized by the following line: 
````
  if (($command eq 'mkvm' or $command eq 'chvm') and $hypconn) {
      my $nodeinfo = $hypconn->get_node_info();
      if (exists($nodeinfo->{model})) {
          $confdata->{$hyp}->{cpumodel} = $nodeinfo->{model};
````
get_node_info() is a method in Sys::Virt, it returns a hash reference summarizing the capabilities of the host node, for more details, please see http://search.cpan.org/~danberr/Sys-Virt/lib/Sys/Virt.pm#METHODS

2. the lines around 445:
````
    unless (defined($hypcpumodel) and $hypcpumodel eq "ppc64le") {
       $rethash{bios}->{useserial}='yes';
    }
```` 
it will define useserial="yes" if the hypcpumodel is not "ppc64le"

xCAT-server/lib/xcat/plugins/kvm.pm
immarvin: removed

xCAT-server/lib/xcat/plugins/kvm.pm
immarvin: added

xCAT-server/lib/xcat/plugins/kvm.pm
immarvin: removed

xCAT-server/lib/xcat/plugins/kvm.pm
immarvin: ``pcpuset`` is short for ``physical cpu set``, which means the physical cpus on the hypervisor, compared with ``vcpu``="virtual cpus on inside the kvm guest"

xCAT-server/lib/xcat/plugins/kvm.pm
immarvin: modified

xCAT-server/lib/xcat/plugins/kvm.pm
immarvin: build the chvm manpage according to the pod changes

immarvin:
hi Victor, is there any more comments?

whowutwut:
I don't have any more comment.  I think there is some conflict that needs to be resolved and I can merge

immarvin:
hi, I have resolved the conflicts, would you please try again?

thans

immarvin:
fixed the reasonable code issues  in https://github.com/xcat2/xcat-core/pull/524/files

Issue #317, opened 2015-10-26
[fvt]2.11 rcons will stay in anaconda interface in rh7.2 ppc64le


junxiawang:
1.download the latest build
2.root@c910f03c05k06 ~]# cat /etc/*-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.2 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.2"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.2 Beta (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.2:beta:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.2
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.2 Beta"
Red Hat Enterprise Linux Server release 7.2 Beta (Maipo)
Red Hat Enterprise Linux Server release 7.2 Beta (Maipo)
[root@c910f03c05k06 ~]# uname -a
Linux c910f03c05k06.pok.stglabs.ibm.com 3.10.0-306.0.1.el7.ppc64le #1 SMP Wed Aug 19 16:06:58 EDT 2015 ppc64le ppc64le ppc64le GNU/Linux
[root@c910f03c05k06 ~]# 
3. nodeset c910f03c05k080  osimage
rpower c910f03c05k080 boot 
rcons c910f03c05k080
anaconda root@c910f03c05k080 /]# 
---------------------->during install still stay in anaconda interface

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/319

PR #319, opened 2015-10-26
fix issue [fvt]2.11 rcons will stay in anaconda interface in rh7.2 ppc64le

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/317


whowutwut:
Is this needed for the ESP build?  If so we need to get approval to integrate...

daniceexi:
@whowutwut 
This only affects the console screen display the install process, no function impact. We can merge this after ESP build.

immarvin:
we have found this problem in diskless provision before, refer to the [link] (http://sourceforge.net/p/xcat/bugs/4502/) for more details.

daniceexi:

PR #325, opened 2015-10-27
If TFTPDIR does not begin with /, correct it

jjohnson42:
TFTPDIR could be set without a leading /.  Support
this syntax by putting a slash in front and hoping
for the best


daniceexi:
I am ok with the fix.

PR #326, opened 2015-10-27
If TFTPDIR does not begin with /, correct it

jjohnson42:
TFTPDIR could be set without a leading /.  Support
this syntax by putting a slash in front and hoping
for the best


daniceexi:
@jjohnson42 
I think if the value of site.tftpdir or site.installdir missed the leading /, it will cause a lot problem in tftp, http. So maybe we should check it in the nodeset command.

daniceexi:
I am ok to merge the change.

PR #327, opened 2015-10-27
Move over the tools page off of SF to RTD

whowutwut:
There's a tools README page on SF that I've moved over to the References page on RTD.  The original page is here: http://xcat.sourceforge.net/tools/README.html


daniceexi:
This is a good addition for our new RTD doc, thanks.

PR #328, opened 2015-10-27
Remove references to sourceforge in xcat-core

whowutwut:
This is a initial pass for removing the references to SF in the xcat-core code

Ran the following command:
``git grep sourceforge  | grep -v "xcat-users@" ``

There still are a bunch of references but I think we need to handle them once the documentation is updated to be more complete

Also removed the getxcatdocs command and simply printing out a message to go to readthedocs to get the documentation


whowutwut:
@daniceexi  we may not want to merge this pull request until after the build this week.

whowutwut:
I updated based on the review comments,

whowutwut:
@daniceexi  is okay to merge?

daniceexi:

Issue #61, opened 2015-08-04
#4673 There's no netbooting status when provision ubuntu stateless node


chudegao:
There's no netbooting status when provision ubuntu stateless node. The node status will changed from 'define' to 'booted' directly.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/329

PR #329, opened 2015-10-28
fix issue #4673 There's no netbooting status when provision ubuntu stateless node

immarvin:
some commands used in node status reporting code does not exist in the ubuntu diskless initrd, use commands in busybox instead


immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/61

penguhyang:
the code looks fine

Issue #314, opened 2015-10-26
[FVT] For CUDA 7.5 support for diskless compute node, need a place to hack command uname while run genimage on Red Hat environment


neo954:
While trying to install CUDA 7.5 for diskless node on RHEL 7.2 Snapshot 2. The management node runs RHEL 7.2 Beta 1. That means, the management node has kernel version ``3.10.0-313.el7.ppc64le``. And the compute node will use kernel version ``3.10.0-316.el7.ppc64le``. In this case the kernel module of CUDA failed to compile during the ``genimage`` procedure.

We need a place to hack command ``uname``.

neo954:
Need run ``mknod /dev/null c 1 3`` in the chroot environment as well.

zet809:
The fix had been checked in. Pls verify.

neo954:
Already fixed. I will close this one.

PR #330, opened 2015-10-28
fix issue 314: CUDA 7.5 support for rh7.2 diskless

zet809:


daniceexi:
See the issue #314 

PR #331, opened 2015-10-28
Fix Cuda related issues found when testing CUDA support

whowutwut:
Add new postscript to configure cuda related settings
Renamed the install pkglist to be more specific supporing only ``ppc64le`` similar to netboot file


cxhong:
Thanks, Victor.  
I will merge this.

Issue #315, opened 2015-10-26
[fvt]2.11 could not generate *stateful-mgmtnode  (osimage) on rh7.0


junxiawang:
How to reproduce:
1.download the latest build
2.on rh7:
[root@c910f02c01p13 /]# cat /etc/*-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.0 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.0"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.0 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.0:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.0
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION=7.0
Red Hat Enterprise Linux Server release 7.0 (Maipo)
Red Hat Enterprise Linux Server release 7.0 (Maipo)
[root@c910f02c01p13 /]# uname -a
Linux c910f02c01p13.pok.stglabs.ibm.com 3.10.0-123.el7.ppc64 #1 SMP Mon May 5 11:18:37 EDT 2014 ppc64 ppc64 ppc64 GNU/Linux
[root@c910f02c01p13 /]# 
3.copycds RHEL-7.0-20140507.0-Server-ppc64-dvd1.iso 
4.lsdef -t osimage
rhels7-ppc64-install-compute  (osimage)
rhels7-ppc64-install-service  (osimage)
rhels7-ppc64-netboot-compute  (osimage)
-------------------------------------> no rhels7-ppc64-stateful-mgmtnode  (osimage) here

whowutwut:
I noticed this a week or so ago but I was not sure what osimage list is suppose to be supported out of the box.  I did think it was strange that we are not supporting state lite...

Wondering if we can add a flag to copycds or just have copycds (without -n) fail if the "required" osimage are not produced.  That would make copycd more "self testing"

immarvin:
hi, according to the previous plan, the "statelite" mode will not be supported anymore in the future xCAT releases.

[...]

PR #333, opened 2015-10-29
correct to generate stateful-mgmtnode osimage on rh7.0

penguhyang:
The return value for $myosver not equal $osver.

### DIFF
--- a/xCAT-server/lib/perl/xCAT/SvrUtils.pm
+++ b/xCAT-server/lib/perl/xCAT/SvrUtils.pm
@@ -829,7 +829,7 @@ sub  update_tables_with_mgt_image
     #if the osver does not match the osver of MN, return
     my $myosver = xCAT::Utils->osver("all");
     $myosver =~ s/,//;
-    if ( $osver ne $myosver ) {
+    if(xCAT::Utils->version_cmp("$osver","$myosver")!=0){
         return 0;
     }
 


penguhyang:
fix issue #315

perl-xCAT/xCAT/Utils.pm
```diff
@@ -2238,8 +2238,9 @@ sub osver
         my $verrel=$line;
         $ver=$line;
         if ( $type ) {
-            $verrel =~ s/[^0-9]*([0-9.]+).*/$1/;
-            ($ver,$rel) = split /\./, $verrel;
+            $verrel =~ /([^0-9]+)\.?(.*)/;
```
immarvin: I don't think your code modification will fix the problem.
1. for redhat7.0 ,it goes into the "if (-f "/etc/os-release"){" branch
2. This problem is not caused by the code logic in xCAT::Utils, it is caused by the the pattern matching loigic in subroutine update_tables_with_mgt_image in xCAT::SvrUtils 
3. your fix might break the  xCAT::Utils

penguhyang:
@immarvin  for redhat, it goes into the "if (-f "/etc/os-release"){" branch and "elsif (-f "/etc/redhat-release")" branch including redhat7.0. I correct the "elsif (-f "/etc/redhat-release")" branch under osver function in Utils.pm since "my $myosver = xCAT::Utils->osver("all");" return the value not equal "$osver" in SvrUtils.pm for redhat7.0.

penguhyang:
@immarvin I have upgrade the code using the "xCAT::Utils->version_cmp("$osver","$myosver");"

xCAT-server/lib/perl/xCAT/SvrUtils.pm
```diff
@@ -829,9 +829,7 @@ sub  update_tables_with_mgt_image
     #if the osver does not match the osver of MN, return
     my $myosver = xCAT::Utils->osver("all");
     $myosver =~ s/,//;
-    if ( $osver ne $myosver ) {
-        return 0;
-    }
+    xCAT::Utils->version_cmp("$osver","$myosver");
```
immarvin: I don't think this line did anything

penguhyang: You are right, now it works fine.

immarvin:
the modified code seems ok

PR #340, opened 2015-10-30
[2.9]for issue 338: rinv CN model -t returns hmc plugin bug

zet809:


daniceexi:

PR #341, opened 2015-10-30
[master]for issue 338: rinv CN model -t returns hmc plugin bug

zet809:
This is for master branch


daniceexi:

PR #292, opened 2015-10-21
Improvements to install guide address issues from #290

whowutwut:
Addresses comments brought up with issue 290, and completed
missing documents for installing SLES and Ubuntu for the install-guide

Address the comments brought up with #290 

http://10.2.5.2/issue290/

PR #342, opened 2015-10-30
add configure_xcat file

junxiawang:
add configure_xcat document  .
Link:
http://10.4.30.1/install/html/guides/install-guides/yum/configure_xcat.html


whowutwut:
@junxiawang  We need to wait for @caomengmeng  to merge pull request #292 because it adds sections into the install guide for SLES and also Ubuntu.  The configure_xcat page will need to change after that is merged...

daniceexi:
@whowutwut Yes, this change is in a very initial state, I still discuss with Jun Xia of how to change it. @junxiawang will make a new pull request base on @caomengmeng 's merge.

whowutwut:
@junxiawang  I think this pull request needs to have the conflicts resolved,  please resolve and I will review.

whowutwut:
@daniceexi , let's continue the discussion about the Configure xCAT in this pull request... You said: 

*For the doc configure xcat which owned by Jun Xia, since it should be done when user has some thought of his cluster like what the networks like, what the dns stuff like domain, ip_forwarding, what the password should be for deployed OS and hardware control points, I prefer to put this part of doc in the admin_guide just before the hardware discovery for certain arch.*

Maybe being newer to xCAT product, we set up new mgmt machines and always hit some problems, but as we know more , we just make the simple changed to continue.  I think it's these steps that I want to document is to help the VERY new user bring up a single node cluster without any issues.  We can document for OpenPower first, RH 7.2 LE, and once we have that document correct, then we can add additional ones like: 
- Open Power
- x86_64 ipmi-based
- VM on KVM hypervisor 

Adding @cxhong  and @pdlun92  to comment also...

I see the goal for Configure xCAT page to take a brand new xCAT install on a management node and bring up 1 single diskful compute node.  We should not go into anything more complicated than that.  We should not define multiple networks...  just use the single network that is on the mgmt node.  

We simply need to define the base environment: 
```
management node=mn1.xcat.org  (10.0.0.1)
site root password=changeme

Compute Node=cn1.xcat.org (10.0.0.2)
BMC IP=50.0.0.1
BMC User=USERID
BMC Pass=PASSW0RD
```

Then give the **exact** xCAT commands in the steps based on this simple environment.  Do not document things like "replace the node with your node name." or replace the IP with your defined IP,  just give the exact commands based on an environment that we define.  Users can understand how to replace the values with their environment specifics... Once we start writing like <arch>.<os> thing like this, it quickly gets confusing to new users.  Too much flexibility... 

Steps could include the following: 
- set the domain
- set the site password
- defined the compute node, add the bmc information
- rpower compute1 stat  
- nodeset
- rnetboot
- rpower... 
- etc...

whowutwut:
The above URL  does not reflect the changes in this pull request..

whowutwut:
Closing this pull request as it seems like the URL is taking a different approach... please re-open if fixed....

Issue #334, opened 2015-10-29
[fvt]2.11  copycds  non-debain iso the cluster.log will show doing debian copycds


junxiawang:
How to reproduce:

[root@c910f03c05k07 /]# cat /etc/*-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.1 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.1"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.1 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.1:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.1
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION=7.1
Red Hat Enterprise Linux Server release 7.1 (Maipo)
Red Hat Enterprise Linux Server release 7.1 (Maipo)
[root@c910f03c05k07 /]# uname -a
Linux c910f03c05k07.pok.stglabs.ibm.com 3.10.0-229.ael7b.ppc64le #1 SMP Fri Jan 30 12:03:50 EST 2015 ppc64le ppc64le ppc64le GNU/Linux
[root@c910f03c05k07 /]# cd /
[root@c910f03c05k07 /]# ls
0  210  211  bin  boot  dev  etc  home  install  lib  lib64  media  mnt  opt  proc  RHEL-LE-7.1-20150219.1-Server-ppc64le-dvd1.iso  root  run  sbin  srv  sys  tftpboot  tmp  usr  var  xcatpost
[root@c910f03c05k07 /]# copycds RHEL-LE-7.1-20150219.1-Server-ppc64le-dvd1.iso 
 cat /var/log/xcat/cluster.log 
Oct 29 00:23:57 c910f03c05k07 xcat[30358]: xCAT: Allowing copycds RHEL-LE-7.1-20150219.1-Server-ppc64le-dvd1.iso for root from localhost
Oct 29 00:23:58 c910f03c05k07 xcat[30359]: Doing debian copycds

immarvin:
Yang peng, could you take a look at this bug?

PR #343, opened 2015-10-30
correct the cluster.log show doing debian copycds

penguhyang:
No need to add this line.


penguhyang:
fix issue  #334

immarvin:
the code looks fine

Issue #97, opened 2015-08-19
[fvt] Can not get full debuginfo in management node for diskless provision after xcatdebugmode=1 


junxiawang:
How to reproduce:
1.
```
c910f02c02p28 ~ # tabdump site |grep xcatdebug
"xcatdebugmode","1",,
910f02c02p28 ~ # cat /etc/*-release
LSB_VERSION="core-2.0-noarch:core-3.2-noarch:core-4.0-noarch:core-2.0-ppc64:core-3.2-ppc64:core-4.0-ppc64"
NAME="SLES"
VERSION="11.4"
VERSION_ID="11.4"
PRETTY_NAME="SUSE Linux Enterprise Server 11 SP4"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:11:4"
SUSE Linux Enterprise Server 11 (ppc64)
VERSION = 11
PATCHLEVEL = 4
c910f02c02p28:~ # uname -a
Linux c910f02c02p28 3.0.101-63-ppc64 #1 SMP Tue Jun 23 16:02:31 UTC 2015 (4b89d0c) ppc64 ppc64 ppc64 GNU/Linux
c910f02c02p28:~ # 
```
2.nodeset cn osimage=sles11.4-ppc64-netboot-compute
3.from management node still could not get debug info
```
tail -f /var/log/messages
Jul 27 14:44:22 c910f02c01p08 xcat: xcatdsklspost : trying to download postscripts from http://10.2.2.28/install/postscripts/
Jul 27 14:44:22 c910f02c01p08 xcat: /opt/xcat/xcatdsklspost : download_postscripts return successfully
Jul 27 14:44:22 c910f02c01p08 xcat: xcatdsklspost: downloaded postscripts successfully
Jul 27 14:44:22 c910f02c01p08 xcat: /opt/xcat/xcatdsklspost :  trying to download http://10.2.2.28/tftpboot/mypostscripts/mypostscript.c910f02c01p08
Jul 27 14:44:22 c910f02c01p08 xcat: /opt/xcat/xcatdsklspost : download_mypostscript failed
Jul 27 14:44:22 c910f02c01p08 xcat: xcatdsklspost: failed to download mypostscript.<nodename>, trying to call getpostscript.awk...
Jul 27 14:44:22 c910f02c02p28 xcat[32357]: xCAT: Allowing getpostscript from c910f02c01p08
[...]

Issue #92, opened 2015-08-19
[fvt]after set xcatdebugmode=1 compute node's log could not fully saved in management node log 


junxiawang:
1.download the latest build
2.xatdebugmode=1
3.nodeset compute node osimage
rpower compute node boot
4.env
[root@c910f04x27v02 ~]# uname -a
Linux c910f04x27v02 2.6.32-431.el6.x86_64 #1 SMP Sun Nov 10 22:19:54 EST 2013 x86_64 x86_64 x86_64 GNU/Linux
[root@c910f04x27v02 ~]# cat /etc/*-release
Red Hat Enterprise Linux Server release 6.5 (Santiago)
Red Hat Enterprise Linux Server release 6.5 (Santiago)
[root@c910f04x27v02 ~]#

5.
on computenode :
cat /var/log/xcat/xcat.log
++ uname -s
++ tr A-Z a-z
+ '[' linux = linux ']'
++ dirname ./syslog
+ str_dir_name=.
+ . ./xcatlib.sh
++ declare -a array_nic_params
++ declare -a array_extra_param_names
++ declare -a array_extra_param_values
+ master=c910f04x27v03
+ conf_file=/etc/syslog.conf
+ sysconfig=/etc/sysconfig/syslog
+ init=/etc/init.d/syslog
on management node:
---------> could not find log in management node

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/344

PR #344, opened 2015-10-30
fix issue #97 [fvt]2.11:still could not get l full debuginfo in management node for diskless provision after xcatdebugmode=1

immarvin:
fix issue #97  and #92


daniceexi:

PR #345, opened 2015-10-30
Change the cuda install doc for rh

daniceexi:
Make it follow the otherpkgs.pkglist rule to
  add cuda and dkms

### DIFF
--- a/docs/source/advanced/gpu/nvidia/osimage/rhels.rst
+++ b/docs/source/advanced/gpu/nvidia/osimage/rhels.rst
@@ -24,12 +24,13 @@ cudafull
 
 #. Add the CUDA repo created in the previous step to the ``pkgdir`` attribute: ::
 
-    chdef -t osimage -o rhels7.2-ppc64le-install-cudafull -p pkgdir=/install/cuda-repo
+    chdef -t osimage -o rhels7.2-ppc64le-install-cudafull -p \
+      pkgdir=/install/cuda-7.5/ppc64le/cuda-core,/install/cuda-7.5/ppc64le/cuda-deps
 
 #. Use the provided ``cudafull`` pkglist to install the CUDA packages: ::
 
     chdef -t osimage -o rhels7.2-ppc64le-install-cudafull \
-    pkglist=/opt/xcat/share/xcat/instal/rh/cudafull.rhels7.ppc64le.pkglist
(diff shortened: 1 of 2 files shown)


docs/source/advanced/gpu/nvidia/osimage/rhels.rst
```diff
@@ -42,23 +43,19 @@ cudaruntime
...
     chdef -t osimage -o rhels7.2-ppc64le-install-cudaruntime \
-    pkglist=/opt/xcat/share/xcat/instal/rh/cudaruntime.rhels7.ppc64le.pkglist
+      pkglist=/opt/xcat/share/xcat/instal/rh/cudaruntime.rhels7.ppc64le.pkglist
 
 Diskless images
 ---------------
 
 The following examples will create diskless images for ``cudafull`` and ``cudaruntime``.  The osimage definitions will be created from the base ``rhels7.2-ppc64le-netboot-compute`` osimage. 
 
-xCAT provides a sample package list files for CUDA. You can find them at:
```
whowutwut: If we remove these file list, should we also do the same for the above diskful sample list?   I guess we don't really need to list the files explicitly.  ... We could also say... 

- diskful
   xCAT provides sample package list files in ``/opt/xcat/share/xcat/install/rh``

- diskless
xCAT provides sample package list files in ``/opt/xcat/share/xcat/netboot/rh``

whowutwut:
I have reviewed the changes. 

I think since this is late for the ESP release, this is OK, but we tried very hard to provide scripts that would work "out of the box" for CUDA.  Now the instructions are more complicated and user needs to make changes to the scripts and customize..  This make the whole process more difficult.  Similar to IB steps, it's very confusing for a new user. 

I still think for the GA we should modify the sample scripts that we ship so the user only need to do chdef to add to osimage.  Not have to copy into custom scripts...   but that's just my opinion...

zet809:
I agree with you that we shall modify the cuda diskless otherpkglist when 2.11GA. 
One reason that we doc that to make changes to the scripts and customize is that our diskless otherpkglist is incorrect. In order not to spin a ESP build, we need to modify the doc to at least show the correct file content to do cuda installation.

daniceexi:
To use otherpkglist, we have to put the dependencies in the pkglist. That means to install cuda, both the pkglist and otherpkgslist have to changed. This caused the complicated. Correct the xCAT shipped cuda pkglist and othetpkgslist is something we can do for 2.11 GA.

PR #355, opened 2015-11-03
rflash to upgrade firmware with ipmitool-xcat

chenglch:
- check ipmitool-xcat version (1.8.15 or above)
- check hpm file
- upgrade firmware with ipmitool-xcat

### DIFF
--- a/perl-xCAT/xCAT/Usage.pm
+++ b/perl-xCAT/xCAT/Usage.pm
@@ -305,7 +305,9 @@ my %usage = (
     PPC (using Direct FSP Management) specific:
 	rflash <noderange> -p <rpm_directory> --activate <disruptive|deferred> [-d <data_directory>]
 	rflash <noderange> [--commit | --recover] [-V|--verbose]
-        rflash <noderange> [--bpa_acdl]",
+        rflash <noderange> [--bpa_acdl]
+    PPC64LE (using BMC Management) specific:
+        rflash <noderange> [-c | --check] <hpm_file>",
     "mkhwconn" =>
 "Usage:
     mkhwconn [-h|--help]
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -36,6 +36,9 @@ my $iem_support;
 my $vpdhash;
 my %allerrornodes=();
 my $global_sessdata;
+
+my $IPMIXCAT = "/opt/xcat/bin/ipmitool-xcat";
+my $NON_BLOCK = 1;
 require xCAT::data::ibmhwtypes;
 
 eval {
@@ -69,7 +72,6 @@ sub handled_commands {
   }
 }
 
-#use Data::Dumper;
 use POSIX "WNOHANG";
 use IO::Handle;
 use IO::Socket;
@@ -107,8 +109,6 @@ my %sdr_caches; #store sdr cachecs in memory indexed such that identical nodes d
 # Firmware data hash
 
 my %hpm_data_hash = ();
-my $hpm_context_string = '';
-
 my %idpxthermprofiles = (
     '0z' => [0x37,0x41,0,0,0,0,5,0xa,0x3c,0xa,0xa,0x1e],
     '1a' => [0x30,0x3c,0,0,0,0,5,0xa,0x3c,0xa,0xa,0x1e], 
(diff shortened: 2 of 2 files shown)


whowutwut:
@chenglch, does this require documentation or man page updates related to firmware flashing?

@pdlun92 , can you review the changes?

chenglch:
@whowutwut , rflash require documentation or man page, but this request will not include this as rflash depends on ipmitool-xcat 1.8.15 which should be patched and built in xcat-dep. I will commit a another pull request in xcat-dep, then commit the doc or man page.

whowutwut:
@chenglch  is ipmitool-xcat just regular ipmitool that is built in the xcat-deps package? so we rename it to ipmitool-xcat?

perl-xCAT/xCAT/Usage.pm
```diff
@@ -305,7 +305,9 @@ my %usage = (
     PPC (using Direct FSP Management) specific:
 	rflash <noderange> -p <rpm_directory> --activate <disruptive|deferred> [-d <data_directory>]
 	rflash <noderange> [--commit | --recover] [-V|--verbose]
-        rflash <noderange> [--bpa_acdl]",
+        rflash <noderange> [--bpa_acdl]
+    PPC64LE (using BMC Management) specific:
+        rflash <noderange> [-c] <hpm_file>",
```
whowutwut: Is the only option for rflash (with BMC) the -c option?  What does -c stand for? 
If other options are specified, do we exit with error?

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1514,38 +1513,157 @@ sub isfirestone {
...
+    }
+    $$version_ptr = (split(/ /, $output))[2];
+    return 0;
+}
+
+#----------------------------------------------------------------#
+# Calculate IPMItool Version format:
+# Example : 1.8.15 --> 1* 1 0000 0000 + 8 * 0000 0000 + 15
+#----------------------------------------------------------------#
+sub cal_ipmitool_version {
```
whowutwut: suggest change function name to "calc_.." instead of "cal"

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1514,38 +1513,157 @@ sub isfirestone {
...
+    my $sessdata = shift;
+    my $ret;
+    my $ipmitool_ver;
+    $ret = get_ipmitool_version(\$ipmitool_ver);
+    return $ret if $ret <0;
+    # only 1.8.15 or above support hpm update for firestone machines.
+    if (cal_ipmitool_version($ipmitool_ver) < cal_ipmitool_version("1.8.15")) {
+        $callback->({error=>"IPMITool $ipmitool_ver do not support firmware update for ".
+                            "firestone mathines, please setup IPMITool 1.8.15 or above.",
+                     errorcode=>1});
```
whowutwut: some alignment issue here? compared to the line above..

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1514,38 +1513,157 @@ sub isfirestone {
...
+    $cmd = "$IPMIXCAT -H $bmc_addr -I lan -U $bmc_userid -P $bmc_password".
+           " mc reset cold";
+    $output = xCAT::Utils->runcmd($cmd, -1);
+    if ($::RUNCMD_RC != 0) {
+        xCAT::SvrUtils::sendmsg ([1,"Running ipmitool command $cmd failed: $output"],
+            $callback,$sessdata->{node},%allerrornodes);
+            return -1;
+    }
+    #wait reset
+    sleep(60);
```
whowutwut: Is there a way to query for the status and continue? or is the sleep the best way?

perl-xCAT/xCAT/Usage.pm
whowutwut: I see below, --check... does that just check the firmware version?  Should we change ``[-c | --check]`` in the Usage?

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1514,38 +1513,157 @@ sub isfirestone {
...
+    #step 3 protect network
+    $cmd = "$IPMIXCAT -H $bmc_addr -I lan -U $bmc_userid -P $bmc_password".
```
pdlun92: I believe the the interface option should be "-I lanplus" as opposed to "-I lan" for this step, according to ipmitool flash directions found here http://pfd.austin.ibm.com/releasenotes/firenze/op810/Open%20Power%20Firmware%20Install%20Instructions_v5.html

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1514,38 +1513,157 @@ sub isfirestone {
...
+           " raw 0x32 0xba 0x18 0x00";
+    $output = xCAT::Utils->runcmd($cmd, -1);
+    if ($::RUNCMD_RC != 0) {
+        xCAT::SvrUtils::sendmsg ([1,"Running ipmitool command $cmd failed: $output"],
+            $callback,$sessdata->{node},%allerrornodes);
+            return -1;
+    }
+    # step 4 upgrade firmware
+    $cmd = "$IPMIXCAT -H $bmc_addr -I lan -U $bmc_userid -P $bmc_password".
```
pdlun92: Same as above. "-I lanplus" for this step as well

xCAT-server/lib/xcat/plugins/ipmi.pm
chenglch: I'm not sure the alignment here as this is a hash. Actually I hope to run perltidy for this file.

chenglch:
@pdlun92, @whowutwut Thanks for your review, I have updated these files again to cover your concerns, besides that I add the thread and lock support to execute rflash

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1514,38 +1514,197 @@ sub isfirestone {
...
+    my $pre_cmd = "$IPMIXCAT -H $bmc_addr -I lanplus -U $bmc_userid";
+    if ($bmc_password) {
+        $pre_cmd = $pre_cmd." -P $bmc_password";
+    }
+    xCAT::SvrUtils::sendmsg ("rflash started, please wait.......",
+            $callback,$sessdata->{node},%allerrornodes);
+    # step 1 power off
+    my $cmd = $pre_cmd." chassis power off";
+    $output = xCAT::Utils->runcmd($cmd, -1);
+    if ($::RUNCMD_RC != 0) {
```
chenglch: As both $::RUNCMD_RC and $callback are global variables, will not use multi thread to implement rflash.

whowutwut:
I think when we do Perltidy to clean up, we should have a 2nd commit, rather than merge the changes into the single commit. This makes it hard to review formatting changes vs actual code changes.  The comment was actually lost in the changes , i saw it in my email notification.

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1514,38 +1514,197 @@ sub isfirestone {
...
+    if (defined($sessdata->{ipmisession}->{password})) {
+        $bmc_password = $sessdata->{ipmisession}->{password};
+    }
+    my $hpm_file = $sessdata->{extraargs}->[0];
+    if ($hpm_file !~ /^\//) {
+        $hpm_file = xCAT::Utils->full_path($hpm_file, $::cwd);
+    }
+    # NOTE (chenglch) lanplus should be used for the task of hpm update 
+    # which indicate the bmc support ipmi protocol version 2.0.
+    my $pre_cmd = "$IPMIXCAT -H $bmc_addr -I lanplus -U $bmc_userid";
```
whowutwut: Nice change...  I like how you made this a variable  :-) 

I was going to comment on in the first review, but was torn between having a single variable for the command or having it spelled out each time for readability on what steps we are issuing.

whowutwut:
We reviewed this code and seems okay.  I know @zet809  was assigned to this Pull request but I would like to get this into tonight weekly build so we can test with it ASAP.

chenglch:
As the race condition exists and the thread is not suggested in perl, I add a new pull request https://github.com/xcat2/xcat-core/pull/382. This implementation make use of Coro, and AnyEvent library which is very similar to the eventlet in python, but we have to add new dependency for Coro library.

whowutwut:
Did you add the threading in the 2nd commit to this Pull request?  I think i saw but didn't review very carefully. sorry.  Is the threading to speed up the rflash over a noderange? 

@daniceexi   While it's convenient to ammend the commits, I don't think we should do this in our instructions...  in step2.... 
http://xcat-docs.readthedocs.org/en/latest/developers/github/pull_request.html#changing-pull-request

As a reviewer, it's hard to see the additional changes if the whole file changes, if we have multiple commits, we can just look at the 2nd or 3rd, or 4th one to see the deltas... rather than review again from scratch...

daniceexi:
@whowutwut 
Regarding to use the --amend for commit:

* Pros
Easier for later reference (after merge) since all the change for certain task could  be found in one commit.

*Cons
As you mentioned, the change for comments cannot be distinguished from the original changes.

Maybe a compromises could be that commit can separated if the change has multiple lines. If the change is trivial, maybe use --amend is better.

Issue #234, opened 2015-09-30
[FVT]:2.11 rvitals issue summary on IBM Power S822LC running RHEL7.2


tingtli:
[root@c910f03c05k08 /]# rvitals c910f05c37 all
c910f05c37: Power Status: on
c910f05c37: Power Overload: false
c910f05c37: Power Interlock: inactive
c910f05c37: Power Fault: false
c910f05c37: Power Control Fault: false
c910f05c37: Power Restore Policy: Always off
c910f05c37: Last Power Event: By Request
c910f05c37: Chassis intrusion: inactive
c910f05c37: Front Panel Lockout: inactive
c910f05c37: Drive Fault: false
c910f05c37: Cooling Fault: false
c910f05c37: Identify Light: off
c910f05c37: 12V Sense: 0 Watts (0 BTUs/hr)
c910f05c37: APSS Fault: 0
c910f05c37: Activate Pwr Lt: 0                    (xCAT needs to add support for 198)   <==confusion output
c910f05c37: All PGood: 0
c910f05c37: Ambient Temp: 28 C (82 F)
c910f05c37: BIOS Golden Side: 0                    (xCAT needs to add support for 195)
c910f05c37: BMC Golden Side: 0                    (xCAT needs to add support for 195)
c910f05c37: Boot Count: 0                    (xCAT needs to add support for 195)
c910f05c37: CPU Core Func 1: 0
c910f05c37: CPU Core Func 10: 0
c910f05c37: CPU Core Func 11: 0
c910f05c37: CPU Core Func 12: 0
c910f05c37: CPU Core Func 13: 0
c910f05c37: CPU Core Func 14: 0
c910f05c37: CPU Core Func 15: 0
c910f05c37: CPU Core Func 16: 0
c910f05c37: CPU Core Func 17: 0
c910f05c37: CPU Core Func 18: 0
c910f05c37: CPU Core Func 19: 0
c910f05c37: CPU Core Func 2: 0
c910f05c37: CPU Core Func 20: 0
c910f05c37: CPU Core Func 21: 0
c910f05c37: CPU Core Func 22: 0
c910f05c37: CPU Core Func 23: 0
c910f05c37: CPU Core Func 24: 0
c910f05c37: CPU Core Func 3: 0
c910f05c37: CPU Core Func 4: 0
c910f05c37: CPU Core Func 5: 0
c910f05c37: CPU Core Func 6: 0
c910f05c37: CPU Core Func 7: 0
[...]

PR #357, opened 2015-11-03
Github issue #234, add more sensor type support for rvitals command

cxhong:
add more sensor type to support rvitals comman,  like type 1f, 22, etc


whowutwut:
I reviewed the changes

PR #358, opened 2015-11-04
Refine IB doc

hu-weihua:
what are refined 
1. add how to verify the IB driver works well
2. modify some example
3. add notice, when the kernel version is different between  generate diskless image and management node, add ``--add-kernel-support`` attribute


cxhong:
Thanks, weihua,  I will merge this.

PR #359, opened 2015-11-04
add debug document

junxiawang:
Add debug_problems document.
Link:http://10.4.30.1/install/html/developers/guides/code/debug.html

### DIFF
--- a/docs/source/developers/guides/code/debug.rst
+++ b/docs/source/developers/guides/code/debug.rst
@@ -1,2 +1,783 @@
-Debug Problems
-==============
+Debug xCAT Problems
+===================
+
+Debugging general issues
+------------------------
+
+Verify xcatd is running. 
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+::     
+
+   lsxcatd -a
+   export XCATBYPASS=1 to be able to run xcat cmds without xcatd (to look at values and use the perl debugger)
+   Restart xcatd with output going to the screen: service xcatd stop; /opt/xcat/sbin/xcatd -f
+
+xcat plugin fails to load
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+if an xcatd plugin is failing to load: ::
+
+   cd /opt/xcat/lib/perl/xCAT_plugin
+   perl -c -I ../ <plugin.pm>
+
+if the perl -c returns "syntax ok", check if the plugin.pm ends with a line "1".
+
+When you installed xCAT, did you also use the latest xcat-dep tarball from **http://xcat.org/files/xcat/xcat-dep**
+
+Verify DB access
+~~~~~~~~~~~~~~~~
+
+::
+
+   tabdump site
+   lsxcatd -d
+
+Verify that some key site attributes are correct: ::
+
+   domain
+   master
+
+Verify the networks table correctly describes the networks in your cluster using ``man networks``
+
+Name resolution
+~~~~~~~~~~~~~~~
+
+Make sure you have name resolution for every defined ip address for the Management Node and the Service Nodes . ::
+
+   run ifconfig -a | grep inet
+
+Take the list of ips that are output and run getent hosts <ipaddres>
+make sure each comes back with a resolved hostname
+If not add the ip address to /etc/hosts
+See [Cluster_Name_Resolution]
+
+Name resolution issues often cause many problems that are difficult to debug. Some symptoms that have been seen: ::
+
+    Statefull (full-disk) installs may hang, fail, or go into "infinite" re-install loops
+    Diskless and statelite installs may hang, be extremely slow, or have problems running xCAT postscripts
+    If an /etc/resolv.conf file has an entry that is not reachable during node install or boot, this can cause long delays, timeouts, and hangs.
+    Hierarchical commands such as xdsh will fail or take a long time to run, if name resolution is not set up correctly on the Management Node and Service nodes.
+
+Node deployment problems
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+    Display problem node attributes using lsdef
+    
+    DHCP: ::
+
+        tail -f /var/log/messages during node deployment to track DHCP requests, etc.
+        Are there two dhcp servers running on the same subnet? Check by running this new xcat tool on any node that is connected to the installation network (can be the MN or SN): /opt/xcat/share/xcat/tools/detect_dhcpd -i <network-facing-NIC
+        *Is dhcp listening on the correct interfaces? Check site.dhcpinterfaces and only specify your cluster-facing network interfaces, then run 'makedhcp -n' to create a new dhcp conf file.
+
+Security ssh or xcat keys/certificates problem
+'''''''''''''''''''''''''''''''''''''''''''''''
+
+    See if **ssh <node> date** can run w/o prompting for pw
+    Run ``xdsh -K`` to update the ssh keys on the nodes
+    Run ``updatenode -k`` to update keys/certificates on the servicenodes
+    If nothing else works, run xcatconfig with appropriate flag(-k,-s,-c). See man page xcatconfig for what you must do after running these commands.
+    Check the auditlog xcat table for denials from the policy table
+
+Verify the intended pkgs are being picked up during genimage or node deployment 
+'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+::
+
+   lsdef <node> \--osimage
+   check image table: lsdef -t osimage <osimage>
+
+Verify all necessary linux services are running 
+'''''''''''''''''''''''''''''''''''''''''''''''
+
+::
+
+   tftp, nfs, http, dhcp, conserver, named
+   Verify no file systems are full on MN or SNs
+   Verify SELinux is disabled, especially on RHEL*
+
+Service Node specific
+'''''''''''''''''''''
+    
+::
+
+    Verify it can access the db on the MN and that the /etc/xcat/cfgloc contents is correct
+    lsxcatd -a
+    Make sure the xCAT versions are the same on MN and SNs
+    lsxcatd -a on MN and SN
+    Make sure the date/time is the same between the MN and SNs
+
+If /install is not mounted on the SN, make sure it is sync'd to all the SNs
+
+Provide xCAT information for development
+''''''''''''''''''''''''''''''''''''''''
+
+Run xcatsnap and send to the xCAT development team
+
+Using xcatdebug to Trace xCAT Code
+----------------------------------
+
+xcatdebug offers two trace facilities to debug the xCAT:
+
+Subroutine calling trace
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Display the calling trace for subroutine when it is called.
+
+The flag -c is used to specify the subroutine list for subroutine calling trace, it can only work with -f. The value of -c can be a configuration file or a subroutine list. ::
+
+   configuration file: a file contains multiple lines of SUBROUTINE_DEFINITION
+   subroutine list:    SUBROUTINE_DEFINITION|SUBROUTINE_DEFINITION|...
+
+SUBROUTINE_DEFINITION: is the element for the -c to specify the subroutine list. 
+
+The format of SUBROUTINE_DEFINITION: plugin
+
+If ignoring the [plugin], the subroutines in the () should be defined in the xcatd. ::
+
+   e.g. (daemonize,do_installm_service,do_udp_service)
+
+Otherwise, the package name of the plugin should be specified. ::
+
+   e.g. xCAT::Utils(isMN,Version)
+   e.g. xCAT_plugin::DBobjectdefs(defls,process_request)
+
+The trace log will be written to **/var/log/xcat/subcallingtrace**. The log file subcallingtrace will be backed up for each running of the xcatdebug -f enable.
+
+Enable the subroutine calling trace for all the subroutines in the xcatd and plugin modules. ::
+
+    xcatdebug -f enable
+
+Enable the subroutine calling trace for the plugin_command in xcatd and defls,process_request in the xCAT_plugin::DBobjectdefs module. ::
+
+    xcatdebug -f enable -c "xCAT_plugin::DBobjectdefs(defls,process_request)|(plugin_command)"
+
+Commented trace log
+~~~~~~~~~~~~~~~~~~~
+
+The trace log code is presented as comments in the code of xCAT. In general mode, it will be kept as comments. But in debug mode, it will be commented back as common code to display the trace log.
+
+This facility offers two formats for the trace log code:
+
+Trace section ::
+
+    ## TRACE_BEGIN
+    # print "In the debug\n";
+    ## TRACE_END
+
+Trace in a single line ::
+
+     ## TRACE_LINE print "In the trace line\n";
+
+Enable the commented trace log ::
+
+    xcatdebug -d enable
+
+This facility also can be enabled by passing the ENABLE_TRACE_CODE=1 global variable when running the xcatd. ::
+
+    ENABLE_TRACE_CODE=1 xcatd -f
+
+How to debug a general OS deployment process
+--------------------------------------------
+
+Network Related Issues
+~~~~~~~~~~~~~~~~~~~~~~
+
+Look at syslog on the MN
+
+By default the syslog is written into the file **/var/log/messages**. You can tail -f it to watch new entries as your nodes deploy.
+
+Following messages should be noticed: ::
+
+    DHCP - The dhcpd on MN/SN should get the dhcp request from node and send back the offer/ack to the node. If no dhcp request is received: 1. try to make sure the node is attempting to boot from network (not its local disk); 2. verify the dhcp server has been started. (for bootp on aix, it will be start by xinetd automatically)
+
+    atftp - To see whether the bootloader file xnba.kpxe/pxelinux.0/yaboot has been transferred to the node.
+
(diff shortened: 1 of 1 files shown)


whowutwut:
@junxiawang  I think we should put each of the debug tips into its own .rst file, this way it's easy to add and remove debugging topics...  If we keep adding to this single page, the page will grow and the user cannot easily find the various topics.... only see the break down in code.html page and the page is already quite large.. 

``cd docs/source/developers/guides/code``
``mkdir debug``
``vi index.rst``  and change ``debug.rst -> debug/index.rst``
``cd debug ``
``cp ../index.rst .`` and use this file as a starting point and add in the files for the various debug topics..

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
-Debug Problems
-==============
+Debug xCAT Problems
+===================
+
+Debugging general issues
+------------------------
+
+Verify xcatd is running.
```
whowutwut: I don't think this is the right description for what we are describing below..    verify xcatd is running could be done using ``service xcatd status``

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+Debugging general issues
+------------------------
+
+Verify xcatd is running. 
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+::     
+
+   lsxcatd -a
```
whowutwut: don't think we need this line.. unless we are looking for information in the ``lsxcatd -a`` output..

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+   export XCATBYPASS=1 to be able to run xcat cmds without xcatd (to look at values and use the perl debugger)
```
whowutwut: This section is all "code" block,  I think we should have description and then code under it...  for example ==>

* To run xCAT commands without going throught the ``xcatd`` daemon, export XCATBYPASS environment variable: 
    ```
    export XCATBYPASS=1
    ```

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+   Restart xcatd with output going to the screen: service xcatd stop; /opt/xcat/sbin/xcatd -f
```
whowutwut: Suggest : 

Stop and then start ``xcatd`` using the -f option to run ``xcatd`` in the foreground: 

``` 
service xcatd stop
/opt/xcat/sbin/xcatd -f 
```

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+xcat plugin fails to load
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+if an xcatd plugin is failing to load: ::
```
whowutwut: xcatd should be xCAT 

If an xCAT plugin fails to load, run the plugin directly with the ``perl -c`` command

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+   cd /opt/xcat/lib/perl/xCAT_plugin
+   perl -c -I ../ <plugin.pm>
+
+if the perl -c returns "syntax ok", check if the plugin.pm ends with a line "1".
+
+When you installed xCAT, did you also use the latest xcat-dep tarball from **http://xcat.org/files/xcat/xcat-dep**
```
whowutwut: what problems would occur if not the latest xcat-dep is used?  Would that cause plugin problems? I think this might be an OLD scenario that we should not have to mention for 2.11 and onwards...

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+Verify DB access
+~~~~~~~~~~~~~~~~
+
+::
+
+   tabdump site
+   lsxcatd -d
+
+Verify that some key site attributes are correct: ::
```
whowutwut: How does the user verify that the attributes are correct? what should they see?

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+   domain
+   master
+
+Verify the networks table correctly describes the networks in your cluster using ``man networks``
```
whowutwut: We can directly link to the man page reference in RTD.  But is that man page sufficient to help the user understand that their network tables is incorrectly defined?  Why would the network table be a problem with "DB access"  ?  This section is "Verify DB access" ..

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+Name resolution
+~~~~~~~~~~~~~~~
+
+Make sure you have name resolution for every defined ip address for the Management Node and the Service Nodes . ::
+
+   run ifconfig -a | grep inet
```
whowutwut: I think we should change this reference to use the ``ip`` command instead of ifconfig.  I think ifconfig is slowly going to be deprecated because it can't support everything...

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+Take the list of ips that are output and run getent hosts <ipaddres>
+make sure each comes back with a resolved hostname
+If not add the ip address to /etc/hosts
+See [Cluster_Name_Resolution]
```
whowutwut: This does not link to a document

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+Name resolution issues often cause many problems that are difficult to debug. Some symptoms that have been seen: ::
```
whowutwut: The user may not know their infinite reinstall loops has to do with name resolution.  Maybe we should have a page for issues, then suggest resolutions...   

Common problems  
1.
2. 
3.  
etc..

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+    Hierarchical commands such as xdsh will fail or take a long time to run, if name resolution is not set up correctly on the Management Node and Service nodes.
+
+Node deployment problems
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+    Display problem node attributes using lsdef
+    
+    DHCP: ::
+
+        tail -f /var/log/messages during node deployment to track DHCP requests, etc.
```
whowutwut: These code block too long.  One thing I noticed that if we have the horizontal scroll on the code block, the PDF document that is generated does not render correctly  . (Want to make @daniceexi aware of this)  So I've been using the \ in the code block to go to the next line and keep it from horizontally scrolling

screen shot 2015-11-07 at 11 00 43 am

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+        Are there two dhcp servers running on the same subnet? Check by running this new xcat tool on any node that is connected to the installation network (can be the MN or SN): /opt/xcat/share/xcat/tools/detect_dhcpd -i <network-facing-NIC
+        *Is dhcp listening on the correct interfaces? Check site.dhcpinterfaces and only specify your cluster-facing network interfaces, then run 'makedhcp -n' to create a new dhcp conf file.
+
+Security ssh or xcat keys/certificates problem
+'''''''''''''''''''''''''''''''''''''''''''''''
+
+    See if **ssh <node> date** can run w/o prompting for pw
```
whowutwut: If there are multiple steps, suggest using the ``#.`` syntax to create a numbered list for easier readability

Also suggest writing out the text instead of abbreviating =>  See if **ssh <node> date** can ~~run~~ return ~~w/o~~ without prompting for ~~pw~~ password

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+    Run ``xdsh -K`` to update the ssh keys on the nodes
+    Run ``updatenode -k`` to update keys/certificates on the servicenodes
+    If nothing else works, run xcatconfig with appropriate flag(-k,-s,-c). See man page xcatconfig for what you must do after running these commands.
+    Check the auditlog xcat table for denials from the policy table
+
+Verify the intended pkgs are being picked up during genimage or node deployment
```
whowutwut: pkgs => packages

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+    If nothing else works, run xcatconfig with appropriate flag(-k,-s,-c). See man page xcatconfig for what you must do after running these commands.
+    Check the auditlog xcat table for denials from the policy table
+
+Verify the intended pkgs are being picked up during genimage or node deployment 
+'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+::
+
+   lsdef <node> \--osimage
+   check image table: lsdef -t osimage <osimage>
```
whowutwut: Help text should not be in the code block

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+    Run ``updatenode -k`` to update keys/certificates on the servicenodes
+    If nothing else works, run xcatconfig with appropriate flag(-k,-s,-c). See man page xcatconfig for what you must do after running these commands.
+    Check the auditlog xcat table for denials from the policy table
+
+Verify the intended pkgs are being picked up during genimage or node deployment 
+'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+::
+
+   lsdef <node> \--osimage
```
whowutwut: what is \ --osimage?  not aware of that syntax..

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+   check image table: lsdef -t osimage <osimage>
+
+Verify all necessary linux services are running
```
whowutwut: I suggest putting the problem in it's own file, for example this is a sub bullet on Node Deployment Problem, but it's difficult to see what problem this is a sub bullet for..

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+    Make sure the xCAT versions are the same on MN and SNs
+    lsxcatd -a on MN and SN
+    Make sure the date/time is the same between the MN and SNs
+
+If /install is not mounted on the SN, make sure it is sync'd to all the SNs
+
+Provide xCAT information for development
+''''''''''''''''''''''''''''''''''''''''
+
+Run xcatsnap and send to the xCAT development team
```
whowutwut: How does the user send to the development team?  This is only the case if the customer has purchased support.. otherwise the avenues are mailing list or create Issue??

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+Using xcatdebug to Trace xCAT Code
+----------------------------------
+
+xcatdebug offers two trace facilities to debug the xCAT:
```
whowutwut: I think we should indicate the two or let's not even use the number, if we provide another trace facility, we have to modify the document to change two -> three.  I'm always thinking about how to maintain the doc with the LEAST amount of changes 

``xcatdebug`` offers the following trace facilities to debug xCAT: 
1. Subroutine calling trace
2. Commented trace log 

Then continue to describe both  or even just have them under the number/bullets.  To do that, need to just keep the indentation so that sphinx knows how to render.

```
#.  Subroutine calling trace 
      everything under here 
      is indented
      
      some code
            <-- indented more

#. Commented trace log

```

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+How to debug a general OS deployment process
+--------------------------------------------
+
+Network Related Issues
+~~~~~~~~~~~~~~~~~~~~~~
+
+Look at syslog on the MN
+
+By default the syslog is written into the file **/var/log/messages**. You can tail -f it to watch new entries as your nodes deploy.
```
whowutwut: suggest just wrapping /var/log/messages in the code syntax instead of bold.

whowutwut: also think we should grep on the mac address, tail -f /var/log/message may be too much output to find the node you are debugging

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+* The hostname of SN/CN should can be resolved from MN. And the hostname of MN/SN should can be resolved from CN. See [Cluster_Name_Resolution]
+
+* You should can run the command ``nslookkup node`` to see whether the dns setup for the node has been done. If you don't want to use dns, just run ``ping node`` to see the IP of node has been worked out by ping command.
+
+* Run following command to setup DNS, see man page for options and [Cluster_Name_Resolution]. ::
+
+   makedns
+
+Setup conserver
+~~~~~~~~~~~~~~
```
whowutwut: I think this needs another ~ under..  If you look at the make html for errors, you should see this one

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+~~~~~~~~~~~~~
+
+Should not have multiple dhcpd daemons in one network/vlan
+
+If there are two or more dhcpd daemons in the installation network/vlan responding to node requests, it will cause strange issues, so you should verify that there is only one dhcp server in the network. This can be done by running a new xcat tool on any node that is connected to the installation network (can be the MN or SN): ::
+
+   /opt/xcat/share/xcat/tools/detect_dhcpd -i <network-facing-NIC>
+
+postscripts issues
+~~~~~~~~~~~~~~~~~~~
```
whowutwut: too many ~ under line 316

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+During the OS deployment, some files will be transported by http. Monitoring the http log could help for debugging.
+
+* The location of http log - For readhat: /var/log/httpd/access_log; For sles: /var/log/apache2/access_log
+* If you use the xnba as boot method, the kernel, image will be gotten from http.
+* If you failed to get xnba file and get error messages such as "Filename: http://$ip/$tftpdir/xcat/xnba/nodes/cn http://$ip/$tftpdir/xcat/xnba/nodes/cn... Connection reset" .
+
+Try to check ::
+  
+   1)if firewall is stopped.
```
whowutwut: Don't think we need to put these steps in code block.  Also always suggest using #. to create numbered lists instead of writing out 1 and 2..

docs/source/developers/guides/code/debug.rst
```diff
@@ -1,2 +1,783 @@
...
+
+For netboot with xnba: the files nbfs.<arch>.gz and nbk.<arch> should be installed in /tftpboot/xcat. The configuration files should be generated in the **/tftpboot/xcat/xnba/nodes/<node>** with node name.
+Note: If the kernel or initrd/osimage can not be loaded correctly, look in the bootloader configuration file to verify that the files IP/Path/file_names have been installed.
+
+Look in the log of the http server
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+During the OS deployment, some files will be transported by http. Monitoring the http log could help for debugging.
+
+* The location of http log - For readhat: /var/log/httpd/access_log; For sles: /var/log/apache2/access_log
```
whowutwut: We try to be consistent in how we are naming the OS.  Redhat is spelled wrong. 
But I think in the guidelines,  RHEL and SLES are used.  I don't know if we need to say where it is.... maybe we can just say..   

Check the httpd ``access_log`` to monitor the files served by the http server

whowutwut:
This is a fairly large document, I will continue to review next week...

whowutwut:
Moving this to 2.12 for now... I don't think will get this section reviewed and updated in 2.11 time frame...  If we do, can move back to 2.11

whowutwut:
@pdlun92  Pat can you also take a look at the changes.  Let us know if it helps with some of the debug.

daniceexi:
@junxiawang Sorry for the later review. Looks like your branch has been poisoned by the merge from master branch. Could you clear up the branch?

whowutwut:
Interesting.. do we know how this happened?  Were you doing other work in the docnew branch?  
A guess is that in our git cheat sheet, we list the update upstream as ``git pull upstream`` and I think the better way is to split this to ``git merge upstream``, then ``git checkout <branch>`` and ``git merge upstream/<branch>`` to have more control.  

Each branch should be created for a separate pull request as well to prevent these kind of merges from happening.  There's really no need to update or rebase unless there's a conflict in the pull request.

daniceexi:
@whowutwut Yes, I agree we should use a branch only for one purpose instead of reusing it.

whowutwut:
@daniceexi ,  It looks like ``git pull upstream`` was done while sitting on the ``docnew`` branch. I  think we have to undo the last two commits from junxiawang:docnew branch and then re-push to get rid of the last 2 commits.  Then she can address the comments in the review.

whowutwut:
@junxiawang  please help to resolve the comments in the pull request when you have time, and then re-open it when it's ready for review again! Thanks!

daniceexi:
@whowutwut We are working on the ``xcatprobe`` command, most of the xcatdebug tips can be covered by the ``xcatprobe``, so we could refine the xcatdebug doc when ``xcatprobe`` is ready.

Issue #299, opened 2015-10-23
[fvt]2.11 makedhcp -n return error info in rhel7 enviroment 


junxiawang:
How to reproduce:
[root@c910f02c01p13 ~]# cat /etc/*-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.0 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.0"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.0 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.0:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.0
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION=7.0
Red Hat Enterprise Linux Server release 7.0 (Maipo)
Red Hat Enterprise Linux 
[root@c910f02c01p13 ~]# lsdef -v
lsdef - Version 2.11 (git commit b1dff58244d7c51a7ddc061b8befa6261fc0dd77, built Thu Oct 22 06:17:32 EDT 2015)
[root@c910f02c01p13 ~]# makedhcp -n
Renamed existing dhcp configuration file to  /etc/dhcp/dhcpd.conf.xcatbak

Error: The file /etc/sysconfig/dhcpd6 doesn't exist, check the dhcp server
--------------------------------------> error info here.
[root@c910f02c01p13 ~]# tabdump networks
#netname,net,mask,mgtifname,gateway,dhcpserver,tftpserver,nameservers,ntpservers,logservers,dynamicrange,staticrange,staticrangeincrement,nodehostname,ddnsdomain,vlanid,domain,comments,disable
"10_0_0_0-255_0_0_0","10.0.0.0","255.0.0.0","eth0","10.2.1.12",,"10.2.1.13",,,,,,,,,,,,
"fd03:2e76:8631::/64","fd03:2e76:8631::/64","/64","eth0",,,,,,,,,,,,,,,
[root@c910f02c01p13 ~]# ls -al /etc/sysconfig/dhcpd6
[...]

PR #363, opened 2015-11-04
makedhcp -n on rhels7 complains about missing dhcpd6 file

whowutwut:
To address Issue #299 

Similar to SLES 11.3 and later, ``dhcp`` on rhels7 stores the dhcpd configuration 
in the /etc/sysconfig/dhcpd file. There's no separate file for ipv4 and ipv6. 

Should we be checking rhels7 and higher? or is this sufficient for now?

### DIFF
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -1601,10 +1601,15 @@ sub process_request
                 my $generatedpath = "$syspath/$dhcpver";
                 my $dhcpd_key = "DHCPDARGS";
 
-                if ($os =~ /sles/i || $os =~ /rhels7/i) {
+                # For SLES11+ and RHEL7+ Operating system releases, the 
+                # dhcpd/dhcpd6 configuration is stored in the same file
+                my $os_ver = $os;
+                $os_ver =~ s/[^0-9.^0-9]//g;
+                if (($os =~ /sles/i && $os_ver >= 11) || 
+                    ($os =~ /rhels/i && $os_ver >= 7)) {
+
                     $dhcpd_key = "DHCPD_INTERFACE";
                     if ($usingipv6 and $dhcpver eq "dhcpd6") {
-                        # For SLES 11 and RHEL7, the dhcpd6/dhcpd config are stored in the same file
                         $dhcpd_key = "DHCPD6_INTERFACE";
                         $generatedpath = "$syspath/dhcpd";
                     }
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -1601,10 +1601,10 @@ sub process_request
                 my $generatedpath = "$syspath/$dhcpver";
                 my $dhcpd_key = "DHCPDARGS";
 
-                if ($os =~ /sles/i) {
+                if ($os =~ /sles/i || $os =~ /rhels7/i) {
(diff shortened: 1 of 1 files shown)


penguhyang:
I think we should be checking rhels7 and higher, because the dhcp and dhcp6 files under /etc/sysconfig/ will not used anymore in rhels7 and higher. Centos7 also has this problem.

whowutwut:
@penguhyang  can you review this and merge if OK?

whowutwut:
@penguhyang , right, we might want to check > rhels7 during the rhels8 support?

penguhyang:
@whowutwut ``my $os = xCAT::Utils->osver();`` the ``$os`` is ``rhserver7`` #366 
maybe redhat8 has this problem too since the dhcpd and dhcpd6 file will not used anymore in latest dhcp package.
```
[root@c910f02c09p09 sysconfig]# more /etc/sysconfig/dhcpd
# WARNING: This file is NOT used anymore.

# If you are here to restrict what interfaces should dhcpd listen on,
# be aware that dhcpd listens *only* on interfaces for which it finds subnet
# declaration in dhcpd.conf. It means that explicitly enumerating interfaces
# also on command line should not be required in most cases.

# If you still insist on adding some command line options,
# copy dhcpd.service from /lib/systemd/system to /etc/systemd/system and modify
# it there.
# https://fedoraproject.org/wiki/Systemd#How_do_I_customize_a_unit_file.2F_add_a_custom_unit_file.3F

# example:
# $ cp /usr/lib/systemd/system/dhcpd.service /etc/systemd/system/
# $ vi /etc/systemd/system/dhcpd.service
# $ ExecStart=/usr/sbin/dhcpd -f -cf /etc/dhcp/dhcpd.conf -user dhcpd -group dhcpd --no-pid <your_interface_name(s)>
# $ systemctl --system daemon-reload
# $ systemctl restart dhcpd.service
```

whowutwut:
@penguhyang  I added some code to check version with the last commit, can you please review and merge if acceptable. 

Based on this small test program...

```
#!/usr/bin/perl
my $os = "rhels6.5";
#my $os = "sles11";
#my $os = "sles10";
#my $os = "rhels7";
#my $os = "rhels7.2";

print "Looking at ... $os\n";

my $os_ver = $os;
$os_ver =~ s/[^0-9.^0-9]//g;

if (($os =~ /sles/i && $os_ver >= 11) ||
    ($os =~ /rhels/i && $os_ver >= 7)) {

    print "Yes, >= sles 11 or >= rhels7\n";
}
```

penguhyang:
The code looks good

xCAT-server/lib/xcat/plugins/dhcp.pm
```diff
@@ -1601,10 +1601,15 @@ sub process_request
                 my $generatedpath = "$syspath/$dhcpver";
                 my $dhcpd_key = "DHCPDARGS";
 
-                if ($os =~ /sles/i) {
+                # For SLES11+ and RHEL7+ Operating system releases, the 
+                # dhcpd/dhcpd6 configuration is stored in the same file
+                my $os_ver = $os;
+                $os_ver =~ s/[^0-9.^0-9]//g;
+                if (($os =~ /sles/i && $os_ver >= 11) || 
+                    ($os =~ /rhels/i && $os_ver >= 7)) {
```
penguhyang: Sorry for my previous comments, for redhat 7 the ``my $os = xCAT::Utils->osver();`` ``$os`` is rhels7 not rhserver7.

PR #364, opened 2015-11-05
AIX -  fix ssl_version issue for aix which openssl less 1.0

daniceexi:
*  export XCATSSLVER=TLSv1 when install xCAT-client
*  Explicitly using XCATSSLVER=TLSv1 in postscripts: xcataixpost,startsyncfiles.aix and aixremoteshell
*  change xCATaixCN61.bnd, make high version of perl-Net_SSLeay.pm to be default


zet809:
Looks good to me. I agree to merge it.

Issue #360, opened 2015-11-04
updatenode #node -P ospkgs report Errors and quit with code 1


wwzhubj:
1. install a node , with image cuda_snpsht2_1543a. 
```
[root@c712ems2 ~]# lsdef -t osimage cuda_snpsht2_1543a
Object name: cuda_snpsht2_1543a
    exlist=/install/osimages/cuda_snpsht2_1543a/kits/KIT_COMPONENTS.exlist
    imagetype=linux
    kitcomponents=xlf.license-compute-15.1.2-0-rhels-7.2-ppc64le,xlf.rte-compute-15.1.2-0-rhels-7.2-ppc64le,xlc.license-compute-13.1.2-0-rhels-7.2-ppc64le,xlc.rte-compute-13.1.2-0-rhels-7.2-ppc64le,xlc.rte-compute-13.1.2-0-rhels-7.2-ppc64le,xlf.compiler-compute-15.1.2-0-rhels-7.2-ppc64le,xlc.compiler-compute-13.1.2-0-rhels-7.2-ppc64le,xlf.license-compute-15.1.2-0-rhels-7.2-ppc64le,xlc.license-compute-13.1.2-0-rhels-7.2-ppc64le,xlf.rte-compute-15.1.2-0-rhels-7.2-ppc64le,essl-license-5.4.0-0-rhels-7.2-ppc64le,essl-license-5.4.0-0-rhels-7.2-ppc64le,essl-computenode-5.4.0-0-rhels-7.2-ppc64le,essl-loginnode-5.4.0-0-rhels-7.2-ppc64le,pessl-license-5.2.0-0-rhels-7.2-ppc64le,pessl-loginnode-5.2.0-0-rhels-7.2-ppc64le,pessl-license-5.2.0-0-rhels-7.2-ppc64le,pessl-computenode-5.2.0-0-rhels-7.2-ppc64le,pperte-license-2.3.0.0-1544a-rhels-7.2-ppc64le,pperte-login-2.3.0.0-1544a-rhels-7.2-ppc64le,pperte-compute-2.3.0.0-1544a-rhels-7.2-ppc64le,pperte-license-2.3.0.0-1544a-rhels-7.2-ppc64le,min-pperte-compute-2.3.0.0-1544a-rhels-7.2-ppc64le
    osarch=ppc64le
    osdistroname=rhels7.2snapshot2-ppc64le
    osname=Linux
    osvers=rhels7.2snapshot2
[...]

PR #365, opened 2015-11-05
fixed issue #360, updatenode -P ospkgs failed to install cuda

cxhong:
if cuda package needs to install from the command  "updatenode -P ospkgs",  it will add to cudapkgs, not pkgs,  and cudapkgs only install in the Ubuntu, not redhat.

for the redhat install flow, it only installed pkgs,  so we need to add cudapkgs to install flow too.


zet809:
The code looks good to me.

PR #368, opened 2015-11-05
correct makedhcp -n return error info in rhel7 environment

penguhyang:
for redhat7 and centos7, the dhcpd6 file will not be used any more.


penguhyang:
fix issue #299

penguhyang:
close this pull request for my mistake that conflict with #363

Issue #367, opened 2015-11-05
Executing "updatenode node -k -F" requires password without prompting message directly


caomengmeng:
Steps:
1. Execute "updatenode node -k -F"

Current result: 
Require password, then after input, there is message remainder it is not allowed to user -k and -F together.

Excepted result: 
There should be message pops up directly to remainder: If you use the -k flag, you cannot specify the -S,-P,-f  or -F flags. Don't need to input password.

caomengmeng:
Verified with latest version and Pass.
See logs below:
[root@c910f02c04p07 xcat-core]# lsxcatd -v
Version 2.11 (git commit 51316df775cb2e410708aab98c3afe9c8ef6ea1d, built Thu Nov 19 03:01:29 EST 2015)
[root@c910f02c04p07 xcat-core]# updatenode c910f02c04p08 -k -F
If you use the -k flag, you cannot specify the -S,-P,-f or -F flags.
[root@c910f02c04p07 xcat-core]# updatenode c910f02c04p08 -k -P
If you use the -k flag, you cannot specify the -S,-P,-f or -F flags.
[root@c910f02c04p07 xcat-core]# updatenode c910f02c04p08 -k -f
If you use the -k flag, you cannot specify the -S,-P,-f or -F flags.
[root@c910f02c04p07 xcat-core]# updatenode c910f02c04p08 -k -S
If you use the -k flag, you cannot specify the -S,-P,-f or -F flags.

PR #370, opened 2015-11-05
correct 'updatenode node -k -F' without prompting message directly

penguhyang:
There should be message pops up directly to remainder: If you use the -k flag, you cannot specify the -S,-P,-f or -F flags. Don't need to input password.

### DIFF
--- a/xCAT-client/bin/updatenode
+++ b/xCAT-client/bin/updatenode
@@ -79,71 +79,77 @@ if (!($tmp =~ / (--help|-h|-v|--version)/)) {
 
 push (@{$cmdref->{arg}}, @ARGV);
 # check the syntax
+my ($ALLSW,$CMDLINE,$ALTSRC,$HELP,$VERSION,$VERBOSE,$FILESYNC,$GENMYPOST,$USER,$SNFILESYNC,$SWMAINTENANCE,$SETSERVER,$RERUNPS,$SECURITY,$OS,$fanout,$timeout,$NOVERIFY);
 if (
     !GetOptions(
-                'A|updateallsw'    => \$::ALLSW,
-                'c|cmdlineonly'    => \$::CMDLINE,
-                'd=s'              => \$::ALTSRC,
-                'h|help'           => \$::HELP,
-                'v|version'        => \$::VERSION,
-                'V|verbose'        => \$::VERBOSE,
-                'F|sync'           => \$::FILESYNC,
-                'g|genmypost'      => \$::GENMYPOST,
-                'f|snsync'         => \$::SNFILESYNC,
-                'l|user:s'         => \$::USER,
-                'S|sw'             => \$::SWMAINTENANCE,
-                's|sn'             => \$::SETSERVER,
-                'P|scripts:s'      => \$::RERUNPS,
-                'k|security'       => \$::SECURITY,
-                'o|os:s'      	   => \$::OS,
-                'fanout=i'         => \$::fanout,
-                't|timetout=i'     => \$::timeout,
-                'n|noverify'       => \$::NOVERIFY,
+                'A|updateallsw'    => \$ALLSW,
+                'c|cmdlineonly'    => \$CMDLINE,
+                'd=s'              => \$ALTSRC,
+                'h|help'           => \$HELP,
+                'v|version'        => \$VERSION,
+                'V|verbose'        => \$VERBOSE,
+                'F|sync'           => \$FILESYNC,
+                'g|genmypost'      => \$GENMYPOST,
+                'f|snsync'         => \$SNFILESYNC,
+                'l|user=s'         => \$USER,
+                'S|sw'             => \$SWMAINTENANCE,
+                's|sn'             => \$SETSERVER,
+                'P|scripts:s'      => \$RERUNPS,
+                'k|security'       => \$SECURITY,
+                'o|os=s'      	   => \$OS,
+                'fanout=i'         => \$fanout,
+                't|timetout=i'     => \$timeout,
+                'n|noverify'       => \$NOVERIFY,
     )
   ) {
     &updatenode_usage();
     exit 1;
 }
 
-if ($::HELP)
+if ($HELP)
 {
-  &updatenode_usage();
-  exit 0;
+    &updatenode_usage();
+    exit 0;
 }
-if (($::USER) && ($::SECURITY))
+if (($USER) && ($SECURITY))
 {
-     my $msg = "-l option is not allowed with -k option.";
-     xCAT::MsgUtils->message("E", $msg);
-     exit 1;
+    my $msg = "-l option is not allowed with -k option.";
+    xCAT::MsgUtils->message("E", $msg);
+    exit 1;
 }
-if (($::USER) && ($::SNFILESYNC))
+if (($USER) && ($SNFILESYNC))
 {
-     my $msg = "-l option is not allowed with -f options.";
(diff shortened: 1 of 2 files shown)


penguhyang:
fix issue #367

xCAT-client/bin/updatenode
```diff
@@ -107,34 +107,39 @@ if (
...
 }
 if (($::FILESYNC) && ($::SNFILESYNC)){  # only one
-     my $msg = "Choose either -f to sync the service nodes, or -F  to sync the nodes not both.";
-     xCAT::MsgUtils->message("E", $msg);
-     exit 1;
+    my $msg = "Choose either -f to sync the service nodes, or -F  to sync the nodes not both.";
+    xCAT::MsgUtils->message("E", $msg);
+    exit 1;
 } 
+if (($::SECURITY) && (($::SWMAINTENANCE) || ($::RERUNPS) || defined($::RERUNPS) || ($::FILESYNC) || ($::SNFILESYNC))){
```
whowutwut: Why is RERUNPS checked twice. Once with defined(). The other variables are not checked this way

penguhyang:
@whowutwut The defined function returns false for undef and true for everything else. updatenode <node> -k -P will skip this verification, updatenode <node> -k -P script will not skip this.

xCAT-client/bin/updatenode
immarvin: hi, Victor, for option  "-P [script1,script2...]", the scripts are optional.
````
       To run all the customization scripts that have been designated for the nodes, (in the "postscripts and    postbootscripts" attributes), type:

         updatenode <noderange> -P

       To run the "syslog" script for the nodes, type:

         updatenode <noderange> -P syslog

       To run a list of scripts, type:

         updatenode <noderange> -P "script1 p1 p2,script2"
````
hi Yang Peng,
I agree with Victor that the conditional statement ``defined($::RERUNPS)`` covers the logic of ``$::RERUNPS`` , in your code, we should just keep ``defined($::RERUNPS)`` here

**Comment by penguhyang** [comment]
@whowutwut @immarvin Right, we should just keep defined($::RERUNPS) here. 
Code has been changed.

File: xCAT-client/bin/updatenode [hunk 2]
```diff
@@ -107,34 +107,39 @@ if (
...
 }
 if (($::FILESYNC) && ($::SNFILESYNC)){  # only one
-     my $msg = "Choose either -f to sync the service nodes, or -F  to sync the nodes not both.";
-     xCAT::MsgUtils->message("E", $msg);
-     exit 1;
+    my $msg = "Choose either -f to sync the service nodes, or -F  to sync the nodes not both.";
+    xCAT::MsgUtils->message("E", $msg);
+    exit 1;
 } 
+if (($::SECURITY) && (($::SWMAINTENANCE) || defined($::RERUNPS) || ($::FILESYNC) || ($::SNFILESYNC))){
```
whowutwut: going back to this... is there a reason we only use ``defined()`` around ``$::RERUNPS`` and not the others in this file?  Or why use ``defined()`` if the other don't have it? 

Just trying to be consistent throughout the code...

penguhyang: I think the reason why we only use ``defined()`` around ``$::RERUNPS`` and not the others is that ``-P`` has a argument while the others have not. Maybe we should add both``($::RERUNPS)`` and ``defined($::RERUNPS)``  to consistent with the updatenode.pm ?

immarvin:
hi @penguhyang ,

Victor's question is why not change 

````
 if (($::SECURITY) && (($::SWMAINTENANCE) || defined($::RERUNPS) || ($::FILESYNC) || ($::SNFILESYNC))){ 
````
to 

````
if (defined($::SECURITY) && (defined($::SWMAINTENANCE) || defined($::RERUNPS) ||defined($::FILESYNC) || defined($::SNFILESYNC))){ 

````
?

penguhyang:
Sorry for my misunderstanding. @whowutwut We can use ``defined()`` for all of them. But I think we don't need to do it since except the ``($::RERUNPS)`` the rest of all can work appropriately.

immarvin:
hi @whowutwut , 

for your last question, I made a quick investigation this afternoon, some points:

* on ``GetOptions`` behavior:

for ``GetOptions`` in our code:
````
if (
     !GetOptions(
                 'A|updateallsw'    => \$::ALLSW,
                 'c|cmdlineonly'    => \$::CMDLINE,
                 'd=s'              => \$::ALTSRC,
                 'h|help'           => \$::HELP,
                 'v|version'        => \$::VERSION,
                 'V|verbose'        => \$::VERBOSE,
                 'F|sync'           => \$::FILESYNC,
                 'g|genmypost'      => \$::GENMYPOST,
                 'f|snsync'         => \$::SNFILESYNC,
                 'l|user:s'         => \$::USER,
                 'S|sw'             => \$::SWMAINTENANCE,
                 's|sn'             => \$::SETSERVER,
                 'P|scripts:s'      => \$::RERUNPS,
                 'k|security'       => \$::SECURITY,
                 'o|os:s'      	   => \$::OS,
                 'fanout=i'         => \$::fanout,
                 't|timetout=i'     => \$::timeout,
                 'n|noverify'       => \$::NOVERIFY,
     )
````
the perldoc mentions:
````
The call to GetOptions() parses the command line arguments that are present in @ARGV and sets the option variable to the value 1 if the option did occur on the command line. Otherwise, the option variable is not touched. Setting the option value to true is often called enabling the option.

Using a colon : instead of the equals sign indicates that the option value is optional. In this case, if no suitable value is supplied, string valued options get an empty string '' assigned, while numeric options are set to 0 .
````
 take ``'k|security'       => \$::SECURITY`` for example, if the variable ``$::SECURITY`` is not initialized, the possible values might be:
undef => "updatenode -k" is not specified
````
"1"     => "updatenode -k" is specified
````
for `` 'P|scripts:s'      => \$::RERUNPS``, if the variable "$::RERUNPS"  it not initialized, the possible values might be:
````
 undef                      => "updatenode -P" is not specified
 empty string("")       => "updatenode -P" is specified 
"postscript"              => "updatenode -P postscript" 
````



*  on the variables  referenced in the "xCAT-client/bin/updatenode",  such as ``$::RERUNPS``, .$::SECURITY``...

the perldoc mentions:
````
 You can refer to variables and filehandles in other packages by prefixing the identifier with the package name and a double colon: $Package::Variable . If the package name is null, the main package is assumed. That is, $::sail is equivalent to $main::sail .

...

Implicitly, there's a package main; at the top of your scripts; that is, unless you explicitly declare a different package, all the variables you declare (keeping the caveat about my in mind) will be in main. Variables that live in a package are reasonably called "package globals", because they are accessible by default to every operator and subroutine that lives in the same package (and, if you're explicit about their names, outside the package, too). 
````
Back to our code,these variables are supposed to be references of the variables with the same name of package "main", which are possible to be read/modified by other perl packages/modules. However, I looked through the code,  the only place they are defined/declared and initialized is "xCAT-client/bin/updatenode" itself. 

According to the above information, my suggestion is to declare them with ``my`` to restrict its namespace in "xCAT-client/bin/updatenode" and use ``defined `` to check whether the related options are specified. 

any thought on this?

whowutwut:
Thanks for that investigation @immarvin ... i agree that maybe it does make sense to declare with ``my`` since we are not using these variables elsewhere... 

Would we want to do that for 2.11.... or maybe just the pull request as-is, then make the changes for 2.12?

immarvin:
I agree to merge the code

PR #374, opened 2015-11-05
[xCAT-client]Improvements to `bin/xdshbak`

samveen:
*    Make xdshbak output match manpage description.
  * Current output:
     ```
[root@xcatmn1 ~]# for op in '-x' '-x -q' '-c' '-c -q' '' '-q'; do 
>     echo "==== xdshbak $op ===" 
>     (  /bin/yes |head -n 1001 |sed 's#^#node1: #'; \
>        /bin/yes |head -n 1001 |sed 's#^#node2: #'  ) | xdshbak $op | head -n 5
> done
==== xdshbak -x ===
....................
node1: y
y
y
y
==== xdshbak -x -q ===

node1: y
y
y
y
==== xdshbak -c ===
..
HOSTS:
-------------------------------------------------------------------------
node1, node2
-------------------------------------------------------------------------------
==== xdshbak -c -q ===

HOSTS:
-------------------------------------------------------------------------
node1, node2
-------------------------------------------------------------------------------
==== xdshbak  ===
..
HOST:node1

-----------
y
==== xdshbak -q ===

HOST:node1

-----------
y
```
  * Manpage described output:

     ```
[root@xcatmn1 ~]# for op in '-x' '-x -q' '-c' '-c -q' '' '-q'; do 
>     echo "==== xdshbak $op ===" 
>     (  /bin/yes |head -n 1001 |sed 's#^#node1: #'; \
>        /bin/yes |head -n 1001 |sed 's#^#node2: #'  ) | xdshbak $op | head -n 5
> done
==== xdshbak -x ===
..
node1: y
node1: y
node1: y
node1: y
==== xdshbak -x -q ===
node1: y
node1: y
node1: y
node1: y
node1: y
==== xdshbak -c ===
..
HOSTS -------------------------------------------------------------------------
node1, node2
-------------------------------------------------------------------------------
y
==== xdshbak -c -q ===
HOSTS -------------------------------------------------------------------------
node1, node2
-------------------------------------------------------------------------------
y
y
==== xdshbak  ===
..
HOST:node1
-----------
y
y
==== xdshbak -q ===
HOST:node1
-----------
y
y
y
```
* Add `-b` option to strip leading hostnames in conjunction with `-x` option (the utility is that I generate a script snippet on each node and combine the generated snippets into a single script on the management node):
```
[root@xcatmn1 ~]# echo spare{1..5}| sed "s# #\\n#g"| sed "s#\$#: executable_line#" \
>     | xdshbak -x
spare1: executable_line
spare2: executable_line
spare3: executable_line
spare4: executable_line
spare5: executable_line
[root@xcatmn1 ~]# echo spare{1..5}| sed "s# #\\n#g"| sed "s#\$#: executable_line#" \
>     | xdshbak -x -b
executable_line
executable_line
executable_line
executable_line
executable_line
```
*   Sort the hostnames using `version` sort instead of string value sort:
  * String value sort:

     ```
[root@inframgtlivexcatmn1.hkdc ~]# echo spare{1..5} spare{10..50..10} \
>     | sed "s# #\\n#g"| sed "s#\$#: $(date)#" |xdshbak -x 
spare1: Thu Nov  5 11:47:31 UTC 2015
spare10: Thu Nov  5 11:47:31 UTC 2015
spare2: Thu Nov  5 11:47:31 UTC 2015
spare20: Thu Nov  5 11:47:31 UTC 2015
spare3: Thu Nov  5 11:47:31 UTC 2015
spare30: Thu Nov  5 11:47:31 UTC 2015
spare4: Thu Nov  5 11:47:31 UTC 2015
spare40: Thu Nov  5 11:47:31 UTC 2015
spare5: Thu Nov  5 11:47:31 UTC 2015
spare50: Thu Nov  5 11:47:31 UTC 2015
```
  * `version` sort:

     ```
[root@xcatmn1 ~]# echo spare{1..5} spare{10..50..10}| sed "s# #\\n#g" \
>     | sed "s#\$#: $(date)#" |xdshbak -x
spare1: Thu Nov  5 06:43:32 EST 2015
spare2: Thu Nov  5 06:43:32 EST 2015
spare3: Thu Nov  5 06:43:32 EST 2015
spare4: Thu Nov  5 06:43:32 EST 2015
spare5: Thu Nov  5 06:43:32 EST 2015
spare10: Thu Nov  5 06:43:32 EST 2015
spare20: Thu Nov  5 06:43:32 EST 2015
spare30: Thu Nov  5 06:43:32 EST 2015
spare40: Thu Nov  5 06:43:32 EST 2015
spare50: Thu Nov  5 06:43:32 EST 2015
```
  * The `version` sort uses the perl package `perl-Sort-Versions` from `[epel]` on RH/Centos {5,6} systems and from `[base]` in RH/Centos 7.
  * The same package is provided by the package `libsort-versions-perl` on Debian/Ubuntu as part of `optional`/`universe` groups.

### DIFF
--- a/xCAT-client/debian/control
+++ b/xCAT-client/debian/control
@@ -7,6 +7,6 @@ Standards-Version: 3.7.2
 
 Package: xcat-client
 Architecture: all
-Depends: ${perl:Depends}, nmap, perl-xCAT
+Depends: ${perl:Depends}, nmap, perl-xCAT, libsort-versions-perl
 Description: Core executables and data of the xCAT management project
  xCAT-client provides the fundamental xCAT commands (chtab, chnode, rpower, etc) helpful in administrating systems at scale, with particular attention paid to large HPC clusters.
--- a/xCAT-client/pods/man1/xdshbak.1.pod
+++ b/xCAT-client/pods/man1/xdshbak.1.pod
@@ -4,7 +4,7 @@ B<xdshbak> - Formats the output of the B<xdsh> command.
 
 =head1 B<SYNOPSIS>
 
-B<xdshbak> [B<-c> | B<-x> | B<-h> | B<-q>]
+B<xdshbak> [B<-c> | B<-x> [ B<-b> ] | B<-h> | B<-q>]
 
 =head1 DESCRIPTION
 
@@ -62,6 +62,9 @@ the output by hostname for easier viewing:
  .
  .
 
+If the B<-b> flag is specified in addition to B<-x>, the hostname at the beginning
+of each line is stripped.
+
 =head2 Standard Error
 
 When the B<xdshbak> filter is used and standard error messages are generated,
@@ -71,6 +74,11 @@ output messages. This is true with and without the B<-c> flag.
 =head1 OPTIONS
 
 =over 6
+
+=item B<-b>
+
+Strip the host prefix from the beginning of the lines. This only
+works with the B<-x> option.
 
 =item B<-c>
 
--- a/xCAT-client/xCAT-client.spec
+++ b/xCAT-client/xCAT-client.spec
@@ -27,7 +27,7 @@ Requires: perl-xCAT >= %{epoch}:%(cat Version)
 
 # fping or nmap is needed by pping (in case xCAT-client is installed by itself on a remote client)
 %ifos linux
-Requires: nmap perl-XML-Simple perl-XML-Parser
+Requires: nmap perl-XML-Simple perl-XML-Parser perl-Sort-Versions
 %else
 Requires: expat 
 %endif
--- a/xCAT-client/bin/xdshbak
+++ b/xCAT-client/bin/xdshbak
@@ -10,6 +10,9 @@ use xCAT::MsgUtils;
 use xCAT::DSHCLI;
 use locale;
 use Getopt::Std;
+
+eval "use Sort::Versions qw/versioncmp/; 1;" or  *versioncmp = sub ($$) { ($a,$b)= @_ ; return $a cmp $b };
+
 #####################################################################
 #                                                                   #
 # Module: xdshbak                                                   #
@@ -231,7 +234,7 @@ sub print_list
 
     local (@lines, $numhosts, $hn_string, $l_string);
 
-    foreach $hostname (sort @hs)
+    foreach $hostname (sort { versioncmp($a, $b) } @hs)
     {
         if (!$::opt_x) { ($num_hosts >= 1) && print "\n"; }
         $num_hosts++;
@@ -295,7 +298,7 @@ sub print_tree
         ($num_hosts >= 1) && print "\n";
         $num_hosts++;
         @wc = split(/:/, $hdr{$index});
-        @wc = sort(@wc);
+        @wc = sort { versioncmp($a, $b) } @wc;
 
         #system "$SPMSG DSH $MSGCAT  INFO511 'HOSTS '";
         xCAT::MsgUtils->message("I",
--- a/xCAT-client/bin/xdshbak
+++ b/xCAT-client/bin/xdshbak
@@ -236,9 +236,9 @@ sub print_list
         if (!$::opt_x) { ($num_hosts >= 1) && print "\n"; }
         $num_hosts++;
 
-        if ($::opt_x) { print "$hostname: $ls{$hostname}"; }
-        else
-        {
+        if ($::opt_x) {
+            map { print "$hostname: $_\n" } split(/\n/, $ls{$hostname});
+        } else {
 
             #$hn_string = `$SPMSG DSH $MSGCAT  INFO510 'HOST: %1\$s\n' $hostname`;
             xCAT::MsgUtils->message("I", "HOST:$hostname\n");
--- a/xCAT-client/bin/xdshbak
+++ b/xCAT-client/bin/xdshbak
@@ -241,7 +241,7 @@ sub print_list
         } else {
 
             #$hn_string = `$SPMSG DSH $MSGCAT  INFO510 'HOST: %1\$s\n' $hostname`;
-            xCAT::MsgUtils->message("I", "HOST:$hostname\n");
+            xCAT::MsgUtils->message("I", "HOST:$hostname");
 
             printf '%.' . (6 + length($hostname)) . "s\n",
               '---------------------------------------------------------------';
@@ -298,10 +298,9 @@ sub print_tree
         @wc = sort(@wc);
 
         #system "$SPMSG DSH $MSGCAT  INFO511 'HOSTS '";
-        xCAT::MsgUtils->message("I", "HOSTS:");
-
-        print
-          "-------------------------------------------------------------------------\n";
+        xCAT::MsgUtils->message("I",
+          "HOSTS -------------------------------------------------------------------------"
+        );
         &display_wc;
         print
           "-------------------------------------------------------------------------------\n";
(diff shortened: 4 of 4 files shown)


samveen:
More details on my use case for the `-b` option:

On a subset of our servers the `genesis image` is unable to get the information about the 10G fibre ethernet cards, until they have been initialized. Unfortunately as it doesn't know about the cards, it can't initialize them, leaving our discovered nodes with incomplete information about their network interfaces.

I need to run a script on each of my nodes to initialize all the network interfaces once and then create a script snippet for that host, which I then run on the management node to update the complete the node definitions for all my incompletely defined nodes. (in this case, the missing interface mac addresses).

whowutwut:
@samveen  we are in invite only mode for 2.11 release and only merging in defect fixes for the 2.11 release .. Going to move this over to the 2.12 milestone.  we Will be branching 2.11 shortly and this can go into master as planned (2.12)

samveen:
Fine by me.

samveen:
@whowutwut Can someone review this, Victor?

whowutwut:
@samveen  Thanks for the reminder, I'll take a look at this.

whowutwut:
@samveen ,  you said above..
. 
*The version sort uses the perl package perl-Sort-Versions from [epel] on RH/Centos {5,6} systems and from [base] in RH/Centos 7.
The same package is provided by the package libsort-versions-perl on Debian/Ubuntu as part of optional/universe groups.*

I searched for the perl-Sort-Versions and it seems like it's in the server-optional repositories and not provided on the base OS from RHEL (at least)..   

But you look for the library and if it's not there you define a equivalent function: 
```
eval "use Sort::Versions qw/versioncmp/; 1;" or  *versioncmp = sub ($$) { ($a,$b)= @_ ; return $a cmp $b };
```
So really we don't need to hard require ``perl-Sort-Versions`` which would cause a failure in the xCAT install if the repos are not defined before hand...

xCAT-client/xCAT-client.spec
```diff
@@ -27,7 +27,7 @@ Requires: perl-xCAT >= %{epoch}:%(cat Version)
 
 # fping or nmap is needed by pping (in case xCAT-client is installed by itself on a remote client)
 %ifos linux
-Requires: nmap perl-XML-Simple perl-XML-Parser
+Requires: nmap perl-XML-Simple perl-XML-Parser perl-Sort-Versions
```
whowutwut: Kind of hesitant about using a hard requires on this since you do have a function defined if it's not there.

xCAT-client/debian/control
```diff
@@ -7,6 +7,6 @@ Standards-Version: 3.7.2
 
 Package: xcat-client
 Architecture: all
-Depends: ${perl:Depends}, nmap, perl-xCAT
+Depends: ${perl:Depends}, nmap, perl-xCAT, libsort-versions-perl
```
whowutwut: Similar to my comment in the .spec file, is this needed if we are doing an eval on "use Sort::Versions" ?

xCAT-client/pods/man1/xdshbak.1.pod
```diff
@@ -4,7 +4,7 @@ B<xdshbak> - Formats the output of the B<xdsh> command.
 
 =head1 B<SYNOPSIS>
 
-B<xdshbak> [B<-c> | B<-x> | B<-h> | B<-q>]
+B<xdshbak> [B<-c> | B<-x> [ B<-b> ] | B<-h> | B<-q>]
```
whowutwut: In the xcat-core directory, there's a ``create_man_pages.py`` script that you can execute in your branch that will automatically create the .rst file for the documentation.  You can add that change to the pull request...  If you can't run it, no worries, we periodically run it to make sure the RTD man pages are in sync with the code.

daniceexi:
@samveen We commend people to use **xcoll** command. Could xcoll command satisfy your requirement?

samveen:
@whowutwut Consolidated reply:
* I think it might be better to leave `HOSTS` in the starting line `-------` . I case there is a  large number of hosts, it's a good differentiation of lines.
* Thank you. 
* I did not want to introduce a new dependency for xCAT. However, the sorting provided by `Sort::Versions` is so compellingly convenient, that I hoped to introduce the xCAT core development team to it via this merge request :).
* On `Centos7` (I had hoped the repos were identical to `RH7`, thus my erroneous assumption), `yum search` for `perl-Sort-Versions` gives me the source repo as `[base]`:
```
[samveen@xcatmn1 ~]$ cat /etc/system-release
CentOS Linux release 7.1.1503 (Core) 
[samveen@xcatmn1 ~]$ yum info perl-Sort-Versions
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: centos.usonyx.net
 * epel: mirror.smartmedia.net.id
 * extras: centos.usonyx.net
 * updates: mirror.vodien.com
Available Packages
Name        : perl-Sort-Versions
Arch        : noarch
Version     : 1.5
Release     : 22.el7
Size        : 13 k
Repo        : base/7/x86_64
Summary     : Perl module for sorting of revision-like numbers
URL         : http://search.cpan.org/dist/Sort-Versions/
License     : GPL+ or Artistic
Description : A perl 5 module for sorting of revision-like numbers
            : 
            : Sort::Versions allows easy sorting of mixed non-numeric and numeric strings,
            : like the 'version numbers' that many shared library systems and revision
            : control packages use. This is quite useful if you are trying to deal with
            : shared libraries. It can also be applied to applications that intersperse
            : variable-width numeric fields within text. Other applications can
            : undoubtedly be found.
[samveen@xcatmn1 ~]$ find /rsync/mirror/OSes/centos/  -name '*Versions*'
/rsync/mirror/OSes/centos/7/os/x86_64/Packages/perl-Sort-Versions-1.5-22.el7.noarch.rpm
```
* I can remove the package from `Requires` in the RPM spec. 
* There should be no issue with the package availability for [Ubuntu](https://launchpad.net/ubuntu/trusty/+package/libsort-versions-perl), so I hope to let the dependency stay in the control file. However, if you want, I'll change it to a `Recommends`.

Please let me know about the decision on the `Requires`/`Depends`, and I'll update the change request appropriately.

samveen:
@daniceexi 
* I took a look at `xcoll`. It collates the output of the commands, which I don't require (not that it matters, because, for my use case, the output of each node differs subtly).
* I've added an option to `xdshbak` which allows for more customization of it's output(don't output hostname), to further fit my requirements.
* I believe it is dangerous to let the command's output be broken, as compared to the description as given in it's manpage.

whowutwut:
@samveen thanks for the detailed explanation! i must have missed the update 6 days ago, sorry.  Will  discuss tonight with the team and let you know our thoughts.

daniceexi:
@samveen @whowutwut Sort-Versions is very short which just includes 50 lines to handle the version comparison. Please take a look whether Utils.pm->CheckVersion() can replace it.

whowutwut:
@samveen This pull request has some conflicts and I think we have not decided on what direction to go, will closer for now. Please re-open if you feel this is required.

samveen:
@whowutwut I was already working on rebasing this onto HEAD. I'll make the appropriate changes and create a new change request.

On a side note, did the team reach any decision on using `Sort::Version` ?

whowutwut:
@samveen  Thanks, the team over in China is on holiday, will discuss this when they are back on Monday with them.  

I did make a comment on the pull request earlier where if we are checking ``Sort::Versions`` is there and if not, using our own comparison function,  maybe we don't need to have to add requires for the perl-Sort-Versions in the spec files.  

That would allow us to not have to worry about the older OS's that are not shipping ``Sort::Versions`` as default

samveen:
True.

However, it could be added to the `recommends` section in the debian control files.

Unfortunately (or maybe fortunately), `yum` doesn't have a concept equivalent to the `apt`/`dpkg` `recommends`, so the information that xCAT would perform better sorting if the module was installed might need documentating somewhere (preferably a high visibility location in the docs)

PR #378, opened 2015-11-06
correct words and syntax for Using Postscript doc

penguhyang:
I found some words and syntax need to correct when I read about the Using Postscript document.
html doc can be found from: http://10.2.9.4/guides/admin-guides/manage_clusters/common/deployment/prepostscripts/post_script.html?highlight=mypostscript


whowutwut:
Looks good!

Issue #217, opened 2015-09-25
[FVT]2.11: log classification does not work on ubuntu 14.04.02


tingtli:
root@c910f03c04k03:/var/log/xcat# cat /etc/*release*
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.2 LTS"
NAME="Ubuntu"
VERSION="14.04.2 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.04.2 LTS"
VERSION_ID="14.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
root@c910f03c04k03:/var/log/xcat# dpkg -l |grep -i xcat
ii  conserver-xcat                      8.1.16-3                         ppc64el      Serial console server daemon/client
ii  elilo-xcat                          3.14-4                           all          Open source implementation of Service Location Protocol V2
ii  grub2-xcat                          2.02-0.16.el7.snap201506090204   all          grub2 resources generated by grub2-mknetdir
ii  ipmitool-xcat                       1.8.11-1                         ppc64el      utility for IPMI control with kernel driver or LAN interface
ii  perl-xcat                           2.11-snap201509240430            all          xCAT perl libraries
ii  xcat                                2.11-snap201509240430            ppc64el      Server and configuration utilities of the xCAT management project
ii  xcat-buildkit                       2.11-snap201509080319            all          xCAT-buildkit provides the buildkit tool and sample kit files to build an xCAT kit.
ii  xcat-client                         2.11-snap201509240430            all          Core executables and data of the xCAT management project
ii  xcat-genesis-base-ppc64             2.10-snap201505172314            all          xCAT Genesis netboot image
ii  xcat-genesis-scripts                2.10-snap201507240530            ppc64el      xCAT genesis
ii  xcat-server                         2.11-snap201509240430            all          Server and configuration utilities of the xCAT management project
ii  xnba-undi                           1.0.3-7                          all          xCAT Network Boot Agent for x86 PXE hosts
root@c910f03c04k03:/var/log/xcat# ls -l /etc/rsyslog.d/
total 16
-rw-r--r-- 1 root   root     311 Mar 17  2012 20-ufw.conf
[...]

PR #380, opened 2015-11-06
Implement log classification on xCAT MN - Bug Fix

neo954:
Fix issue #217.

### DIFF
--- a/xCAT/etc/logrotate.d/xcat
+++ b/xCAT/etc/logrotate.d/xcat
@@ -2,8 +2,8 @@
     missingok
     sharedscripts
     postrotate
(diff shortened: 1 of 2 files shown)


whowutwut:
Hi @neo954 , I would suggest that we write the story title into the title of this pull request rather than the story number...  "Implement log classification on xCAT MN".  

The RTC tool we are using is internal and if we do not have any descriptive title, we will not benefit from feedback for any community members who may be interested in this topic.   It's more work for us, but I think it's worth it.  We can mention the story number in the task for our own cross reference.

immarvin:
the code looks fine, i agree to merge it

Issue #348, opened 2015-11-02
Fix missed file links in documentation


caomengmeng:
1. Link: https://xcat-docs.readthedocs.org/en/latest/overview/quick_start.html, 
Description: step6 miss Hardware Managment link -- the page source is right, but why no link shown? --Done
2. Link: http://xcat-docs.readthedocs.org/en/latest/guides/admin-guides/manage_clusters/ppc64le/diskful/customize_image/index.html
Description: "Initialize the Compute for Deployment" link missed --- Done.

3. Link： https://xcat-docs.readthedocs.org/en/latest/guides/admin-guides/manage_clusters/ppc64le/updatenode.html
Description:
Item1: For "Using_Clone_to_Deploy_Server#Update_Nodes_Later_On ", miss links, need to update.
there is the content: https://xcat-docs.readthedocs.org/en/latest/advanced/sysclone/sysclone.html
Item2: For "Use instuctions in Sync-Files-label",  is it also a file link?
https://xcat-docs.readthedocs.org/en/latest/guides/admin-guides/manage_clusters/common/deployment/syncfile/syncfile_synclist_file.html
The location of synclist file for updatenode and install process

4. Link: https://xcat-docs.readthedocs.org/en/latest/advanced/hierarchy/configure_dhcp.html
Description:
Failed to link doc for 
 XCAT_pLinux_Clusters/#setup-dhcp: 
 https://xcat-docs.readthedocs.org/en/latest/guides/admin-guides/manage_clusters/ppc64le/discovery/config_environment.html

caomengmeng:
@bybai  Would you please check if the files links should be added or not?

bybai:
@caomengmeng When we wrote the updatenode.rst, these docs were not there, and we just left link there. Now we have these docs, so you can add these 2 links in updatenode.rst.

whowutwut:
We decided this Issue should be expanded this task to search the TODO in the documentation and add the missing links throughout the doc.

caomengmeng:
issue1 and 2 are fixed by https://github.com/xcat2/xcat-core/pull/381

caomengmeng:
Issue 3 and Issue 4 are fixed by https://github.com/xcat2/xcat-core/pull/397

PR #381, opened 2015-11-06
Update missed file links

caomengmeng:
Fix issue1 and issue2 in https://github.com/xcat2/xcat-core/issues/348

### DIFF
--- a/docs/source/guides/admin-guides/manage_clusters/ppc64le/diskful/customize_image/index.rst
+++ b/docs/source/guides/admin-guides/manage_clusters/ppc64le/diskful/customize_image/index.rst
@@ -1,6 +1,6 @@
 Customize osimage (Optional)
 ============================
-Optional means all the subitems in this page are not necessary to finish an OS deployment. If you are new to xCAT, you can just jump to `Initialize the Compute for Deployment`.
+Optional means all the subitems in this page are not necessary to finish an OS deployment. If you are new to xCAT, you can just jump to `: :doc:`Initialize the Compute for Deployment <../deploy_os/index>`.
 
 .. toctree::
    :maxdepth: 2
--- a/docs/source/overview/quick_start.rst
+++ b/docs/source/overview/quick_start.rst
@@ -33,7 +33,7 @@ If xCAT looks suitable for your requirement, following steps are recommended pro
 
    Now you have the node definition. Take a try to confirm the hardware control for defined nodes is working. e.g. ``rpower <node> stat``.
 
-   Refer to the doc: :doc:`Hardware Management <../guides/admin-guides/manage_clusters/ppc64le/management>` to learn how to perform the remote hardware control.
+   Refer to the doc: :doc:`Hardware Management </guides/admin-guides/manage_clusters/ppc64le/management>` to learn how to perform the remote hardware control.
 
 #. Deploy OS for the target nodes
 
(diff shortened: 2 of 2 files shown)


docs/source/guides/admin-guides/manage_clusters/ppc64le/diskful/customize_image/index.rst
```diff
@@ -1,6 +1,6 @@
 Customize osimage (Optional)
 ============================
-Optional means all the subitems in this page are not necessary to finish an OS deployment. If you are new to xCAT, you can just jump to `Initialize the Compute for Deployment`.
+Optional means all the subitems in this page are not necessary to finish an OS deployment. If you are new to xCAT, you can just jump to `: :doc:`Initialize the Compute for Deployment <../deploy_os/index>`.
```
whowutwut: I would prefer to remove this line completely.  I don't think we need to define what optional means, do we??

docs/source/overview/quick_start.rst
```diff
@@ -33,7 +33,7 @@ If xCAT looks suitable for your requirement, following steps are recommended pro
 
    Now you have the node definition. Take a try to confirm the hardware control for defined nodes is working. e.g. ``rpower <node> stat``.
 
-   Refer to the doc: :doc:`Hardware Management <../guides/admin-guides/manage_clusters/ppc64le/management/index>` to learn how to perform the remote hardware control.
+   Refer to the doc: :doc:`Hardware Management <../guides/admin-guides/manage_clusters/ppc64le/management>` to learn how to perform the remote hardware control.
```
whowutwut: Suggest that we specify the absolute path instead of the relative path.  So just remove the ``../``
This way if we ever move this file around.. it won't have any effect on the link. 

Also suggest:
Refer to ~~the doc:~~ *Hardware Management* to learn how to perform the remote hardware control. 

There are other places in the document that we also say "Refer to the doc:"

whowutwut:
@caomengmeng  You didn't write the quick start, but since you made a change in the file, I was looking at the quick start guide.. .but then as I read the steps, this is really not "quick start..." There are 9 steps that link to huge amounts of xCAT documentation. 

I think my vision for the install guide is really a quick start guide for new users to get xCAT installed and up and running with 1 mgmt and 1 compute node .  All the other features of xCAT can be learned and explored by the user after we have xCAT installed

What is your opinion on this?

docs/source/guides/admin-guides/manage_clusters/ppc64le/diskful/customize_image/index.rst
daniceexi: @whowutwut This was a long discussion that how to indicate new user that he/she can ignore this 'Customize osimage' part directly if just want to try the basic OS deployment. Do you think it's meaningful for a new user?

docs/source/overview/quick_start.rst
daniceexi: Use absolute path is a good idea

caomengmeng:
Hi @whowutwut and @daniceexi,
Thank you very much for reply.
1. I will change the two links to be absolute ones.
2. About "Refer to the doc" to be "Refer to", both are OK?
3. Yes, one MN and one CN is really a quick guide for new users just like me, but it seems the optional ones are also helpful to let new users to understand xCAT. 

Finally, I make the two changes only because there should be a link here, but absolutely we missed, and I think this situation belongs to https://github.com/xcat2/xcat-core/issues/348, so can I just complete the missed links issue for this PR?

caomengmeng:
Hi @whowutwut and @daniceexi , already update the absolute path and apply for merge.

whowutwut:
Thanks, looks good.  @daniceexi  I think we will need to revisit this section or these links when add more sections for x86_64, right now we are linking always to the Power section, or maybe that should be the guideline moving forward...

PR #382, opened 2015-11-06
Add Coro and AnyEvent lib to replace original thread for rflash

chenglch:
Original implementation has some race condition in dealing with the
glocal variables, fix this issue by replacing the thread with
coroutine.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -32,6 +32,8 @@ use Thread qw(yield);
 use LWP 5.64;
 use HTTP::Request::Common;
 use Time::HiRes qw/time/;
+use Coro;
+use AnyEvent::Util;
 my $iem_support;
 my $vpdhash;
 my %allerrornodes=();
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1515,6 +1517,56 @@ sub isfirestone {
...
 
+#----------------------------------------------------------------#
+# sleep in coroutine
+#
+#  Arguments:
+#        time: in second
+#----------------------------------------------------------------#
+sub ae_sleep {
+    my ($secs) = @_;
+    my $cv = AE::cv();
```
whowutwut: any reason why we are doing AE::cv() here and AnyEvent->condvar() below? Make consistent? 

Should we spell out AnyEvent to make it easier to know AE = AnyEvent? 
Same comment regarding the AE::timer below..

chenglch:
@whowutwut , I have uploaded a new pull request https://github.com/xcat2/xcat-core/pull/391 which fork one process for each node while running rflash procedure. Personally speaking, I prefer current implementation which make use of the event driven model, but it will take too much effort to maintain the Coro and AnyEvent dependency for rhels7 ( rhels6 and fedora have these libraries by default ).

PR #384, opened 2015-11-09
modify autotest installation case for rhels7.2

junxiawang:


xCAT-test/autotest/testcase/installation/reg_linux_diskfull_installation_flat
```diff
@@ -23,9 +23,9 @@ cmd:sleep 10
 cmd:if cat /etc/*release |grep SUSE >/dev/null;then cat /var/lib/dhcp/db/dhcpd.leases|grep $$CN;elif cat /etc/*release |grep "Red Hat" >/dev/null;then cat /var/lib/dhcpd/dhcpd.leases|grep $$CN;fi     
 check:output=~$$CN
 
-cmd:copycds   $$ISO
+cmd:if cat /etc/*release |grep "7.2" |grep "Red Hat">/dev/null;then copycds $$ISO -n rhels7.2;else copycds $$ISO;fi
```
whowutwut: I think we should not specify ``-n rhels7.2`` because on the normal case, once we add in the DVD identifier, this testcase should pass. So my point of view is let this test case fail, so that we are reminded that we need to add the DVD identifier.

whowutwut: Also, trying to understand. is this testcase used for all x86_64, ppc64, ppc64le?

junxiawang: I will modify this code in autotest configure file.
 
 
----- Original message -----From: Victor Hu <notifications@github.com>To: xcat2/xcat-core <xcat-core@noreply.github.com>Cc: Jun Xia Wang/China/IBM@IBMCNSubject: Re: [xcat-core] modify autotest installation case for rhels7.2 (#384)Date: Thu, Nov 12, 2015 5:55 AM 
In xCAT-test/autotest/testcase/installation/reg_linux_diskfull_installation_flat:
> @@ -23,9 +23,9 @@ cmd:sleep 10>  cmd:if cat /etc/*release |grep SUSE >/dev/null;then cat /var/lib/dhcp/db/dhcpd.leases|grep $$CN;elif cat /etc/*release |grep "Red Hat" >/dev/null;then cat /var/lib/dhcpd/dhcpd.leases|grep $$CN;fi    >  check:output=~$$CN>  > -cmd:copycds   $$ISO> +cmd:if cat /etc/*release |grep "7.2" |grep "Red Hat">/dev/null;then copycds $$ISO -n rhels7.2;else copycds $$ISO;fi
Also, trying to understand. is this testcase used for all x86_64, ppc64, ppc64le?

whowutwut: Interesting, @junxiawang  I didn't know you could reply to the GitHub emails and it will come into this as a note... It does come out a bit messy though :-D

PR #387, opened 2015-11-09
correct the message to make the updatenode -k output looks neat

penguhyang:
```
[root]# updatenode c910f02c09p08 -k
Enter the password for the userid: root on the node where the ssh keys
will be updated:
```
or in this way
```
[root]# updatenode c910f02c09p11 -k
Enter the password for the userid: root
on the node where the ssh keys will be updated.
password:
```

### DIFF
--- a/xCAT-client/bin/updatenode
+++ b/xCAT-client/bin/updatenode
@@ -159,7 +159,7 @@ if ($::SECURITY) {
         # prompt for the password for the userid on the node that will be setup
         my $userpw;
         $msg =
-          "Enter the password for the userid: $to_userid on the node where the ssh keys \nwill be updated:\n";
+          "The ssh keys will be updated for '$to_userid' on the node(s).\nPassword:";
         xCAT::MsgUtils->message("I", $msg);
         system("stty -echo");    # turn off keyboard
         chop($userpw = <STDIN>);
--- a/xCAT-client/bin/xdsh
+++ b/xCAT-client/bin/xdsh
@@ -405,7 +405,7 @@ sub parse_args_xdsh
              # prompt for the password for the userid on the node that will be setup
             my $userpw;
             $msg =
-              "Enter the password for the userid: $to_userid on the node where the ssh keys \nwill be updated:\n";
+              "The ssh keys will be updated for '$to_userid' on the node(s).\nPassword:";
             xCAT::MsgUtils->message("I", $msg);
             system("stty -echo");    # turn off keyboard
             chop($userpw = <STDIN>);


whowutwut:
What about taking a more ssh login approach and prompting Password:
```
[root]# updatenode c910f02c09p11 -k
The ssh keys will be updated for 'root' on c910f02c09p11
Password: 
```

Is there any other user ID that is valid besides ``root``?    If not, then can we just prompt? 
```
[root]# updatenode c910f02c09p11 -k
Password:
```

Will changing this affect any existing ``expect`` scripts that some system admins might have written to expect the prompts???

penguhyang:
Taking a more ssh login approach and prompting Password looks good.
```
[root]# updatenode c910f02c09p11 -k
The ssh keys will be updated for 'root' on c910f02c09p11
Password: 
```
I think we should keep the prompts.

whowutwut:
The only other question i have is do we think this will break existing customers ``expect`` scripts?  I'm okay to merge this change, maybe updating the ssh keys is not something that is often done using a script.. so maybe this is okay??  @daniceexi  ?

xCAT-client/bin/updatenode
```diff
@@ -159,7 +159,7 @@ if ($::SECURITY) {
         # prompt for the password for the userid on the node that will be setup
         my $userpw;
         $msg =
-          "Enter the password for the userid: $to_userid on the node where the ssh keys \nwill be updated:\n";
+          "The ssh keys will be updated for '$to_userid' on the node\nPassword:";
```
whowutwut: One additional comment, should we write "node(s)" since we can provide a noderange.. it could be 1 or more nodes... ?

```
The ssh keys will be updated for `<user>` on the node(s).
Password:
```

penguhyang: yes, we should write node(s) since we can provide a noderange.

penguhyang:
@whowutwut I updated the xdsh too.

whowutwut:
@samveen, question for you, as a user of xCAT, do you think that changing the two files in this pull request would cause damage to users scripts, maybe they are using expect to enter root password?

samveen:
@whowutwut This may cause some damage, but I would think that most expect scripts would probably use `-re '.: '` instead of  `'updated: '`(I know mine would). Maybe a post on xcat user list might give more insight (we don't so much of expect based scripting (yet), here in my current shop).

whowutwut:
Thanks!

 @daniceexi @penguhyang , can we post this to the mailing list before we merge?

bybai:
Hi @whowutwut , daniceexi is on the way for business travel today. I think we can post this to the mailing list.

whowutwut:
I did not see any objections to the mailing list, so I'm going to merge this in for tomorrows RC1 build.  @penguhyang  can you work with FVT to make sure this is executed and no issues..

PR #389, opened 2015-11-10
Update discinfo.pm(merge to trunk)

lyu571:
Add the Centos6.7 to discinfo.pm for PCM Rhel6.7 support(add centos6.7 os distro in rhel6.7).
Merge it to trunk.


daniceexi:
I am ok with the change. Thanks.

PR #391, opened 2015-11-11
Fork a rflash process for each node

chenglch:
Original implementation use thread to deal with the rflash
procedure for multiple nodes, but thread is not suggested
by perl. This patch aims to replace the thread with process.


zet809:
The code seems good to me, I agree to merge it.

PR #393, opened 2015-11-11
add ubuntu netboot initrd.gz when creating osimage definition

zet809:


daniceexi:

PR #399, opened 2015-11-12
Add 4 new updatenode cases

caomengmeng:
@tingtli, Would you please review these new cases?
Add 4 new cases as below order:
1. updatenode,  without option, it will run the "-S", "-P" and "-F" options in this order.  
2. updatenode noderange -F -P script
3. updatenode noderange script, without -P, this can also be run correctly.
4. updatenode noderange -P script1 p1, script2 p1

### DIFF
--- a/xCAT-test/autotest/testcase/updatenode/cases0
+++ b/xCAT-test/autotest/testcase/updatenode/cases0
@@ -348,6 +348,67 @@ cmd:rm -f /install/script1
 end
 
 
+start:updatenode_script3
+cmd:echo "#!/bin/bash" > /tmp/updatenode_P_script_input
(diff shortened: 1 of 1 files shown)


bybai:
@caomengmeng You can use ``git commit --amend``, so that there will be one commit.

whowutwut:
Need to fix the commits to be "non root" before we can merge this.

caomengmeng:
@bybai  Thank you very much for you suggestions, it's great help for me to improve all the 4 cases and to understand the updatenode command.
@whowutwut the commits user info is updated.

See: https://github.com/xcat2/xcat-core/pull/414
And this will be closed

PR #401, opened 2015-11-12
Fix bug#398 IB driver can not be installed via /opt/xcat/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2 

hu-weihua:
When restructure code, made a mistake. fix it now

### DIFF
--- a/xCAT-server/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2
+++ b/xCAT-server/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2
@@ -114,7 +114,7 @@ do
     shift
 done
 
-if [ ! -f "$OFED_PATH" ]; then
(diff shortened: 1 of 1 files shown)


neo954:
We are good to go.

xCAT-server/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2
```diff
@@ -114,7 +114,7 @@ do
     shift
 done
 
-if [ ! -f "$OFED_PATH" ]; then
+if [ -z "$OFED_PATH" ]; then
     echo "[Error] Without Mellanox OFED file path, please assign correct path" >&2
```
whowutwut: Since the -p option is required, really, if we find OFED_PATH is not set, we should print the usage and exit.

Issue #405, opened 2015-11-13
Miss 3 dependence package when install IB driver in rhel7.1+x86_64 environment


hu-weihua:
When install IB drivers in rhel7.1+x86_64 environment, mias **gtk2 atk cairo** 3 dependence package.

hu-weihua:
fixed by pull request #406

PR #406, opened 2015-11-13
Fix bug #405 Miss 3 dependence package when install IB driver in rhel7.1+x86_64 environment 

hu-weihua:
Add these **gtk2 atk cairo** 3 dependence package into ib.rhels7.ppc64le.pkglist


daniceexi:

PR #407, opened 2015-11-14
xcat-core/docs: minor documentation fixes

mfoliveira:
BMC (Baseboard Management Controller) is used as the SP (Service Processor) on IBM S822LC, for example.


whowutwut:
@mfoliveira, looks good,  thanks for making these improvements!

PR #411, opened 2015-11-16
fix issue 352: 2.9.2rspconfig HMC sshcfg does not show disable state …

zet809:
…even the state is disabled


daniceexi:

PR #413, opened 2015-11-16
add configure for for xcat document

junxiawang:
add doc for configure xcat
links for review:
http://10.4.30.1/install/html/guides/admin-guides/manage_clusters/ppc64le/configure_xcat.html


whowutwut:
I think you have two commits here, one is for DNS and switch, the other is for configure xCAT... .  I think the pull request is only for configure_xcat right?  

Maybe you forgot to switch branch when doing the commit ? 

I think you can fix this by doing the following... 
```
git checkout docl
git reset --soft HEAD~
```
That should remove the last commit.. You would want to save those changes if you need to commit for another branch... 

Then I think you need to force push...

```
git push origin docl -f 
```

I'm not 100% sure that will do it... anyone else know?

junxiawang:
thanks @whowutwut I have modified it

bybai:
It is good for me. I will merge it.

PR #414, opened 2015-11-16
Improve updatenode 4 new cases and git user info

caomengmeng:
@tingtli, Would you please review these new cases?
Add 4 new cases as below order:
1. updatenode, without option, it will run the "-S", "-P" and "-F" options in this order.

2. updatenode noderange -F -P script
3. updatenode noderange script, without -P, this can also be run correctly.
4. updatenode noderange -P script1 p1, script2 p1

### DIFF
--- a/xCAT-test/autotest/testcase/updatenode/cases0
+++ b/xCAT-test/autotest/testcase/updatenode/cases0
@@ -9,7 +9,6 @@ check:rc==0
 check:output=~Version
 end
 
-
 start:updatenode_diskful_syncfiles
 cmd:mkdir -p /install/custom/install/__GETNODEATTR($$CN,os)__/ 
 check:rc==0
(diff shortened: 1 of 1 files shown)


caomengmeng:
@tingtli update cases:
1. Update case name for first case to be: updatenode_without_flag
2. Add descriptoin for each case
3. Add "chdef noderange postscripts=/install/postscripts" before command "updatenode xxx" and "chdef noderang postscripts=" after command "updatenode xxx" for the last 3 new cases.

whowutwut:
Set for 2.11 milestone, testing cases should be safe to merge since they do not affect the mainline code.  Do we want to add a label for **testcase** or **fvt:testcase** ?

caomengmeng:
@daniceexi  would you please look at Victor's comments for the fvt:testcase label issue?

tingtli:
@whowutwut 
Do you mean all test cases code need a lable?

daniceexi:
@whowutwut @caomengmeng I added a lable **component:test** for all the issues and pull request which related to test like automation framework or test cases.

daniceexi:
@whowutwut For test cases, I'd like it's also reviewed by someone to make sure it's really testing the stuff it declears to. In the future, xCAT will totally rely on the automation test tool to guarantee the quality of xCAT release if we want to release xCAT more agilely

whowutwut:
My comment about the label is not applicable since Xiao Peng created the component labels. 

@daniceexi I agree we should review testing cases and make sure we are testing effectively.  My comment on merge was really referring to pull request during code freeze time. I think it's ok to merge changes to the test case similar to document changes because it doesn't affect the main code and customers

daniceexi:
@whowutwut Yes I agree that the checkin for test and doc can do in anytime.

PR #415, opened 2015-11-16
Fix defect of the rflash option to check

chenglch:
rflash for ipmi also supports the fpc machines,
delete the check constraints for unknown options.


zet809:
Looks OK to me, I agree to merge it.

Issue #347, opened 2015-11-02
[FVT]Confused issues for Prescripts and Postscripts pages


caomengmeng:
1. Link: https://xcat-docs.readthedocs.org/en/latest/guides/admin-guides/manage_clusters/common/deployment/prepostscripts/post_script.html

Description:

Item1:   There is a sentence : See Suggestions for writing scripts. Should we add a link to this file?

Item2:  Another sentence:　The mypostscript file is generated according to the mypostscript.tmpl file.
When Ｉ see this mypostscript.tmpl file, just wonder what is it?
Then notice that below section in this page describes the tmpl file, so do we need to add a current page link to this tmpl file when it appears in this sentence?
Not sure this is a problem or not?

Below issues are all updated directly to GitHub
Item3: In sentence "xCAT provides a way for the admin to customize the information that will be provide to the postscripts/postbootscripts when they run on the node", do we need to update the provide to be provided? or other?

Item4: In section: site table precreatemypostscripts attribute, there are two words "xcat", do we need to change them to be "xCAT"?, and do we need to update the section title first word "site" to be "Site"?

Item5: In sentence "Note the attributes are not always defined exactly as", is it changed to be "Note: the xxxx"? just like other "Note:" reminders?

Item6: In sentence:＂For example, to get the new updatstatus attribute from the nodelist table:＂ change the "updatstatus" to be "updatestatus"

caomengmeng:
@junxiawang, I am reading these pages, and will update if there are anything confused.
Not sure all should be updated just mark them here.

daniceexi:
@caomengmeng Could you talk with @tingtli @junxiawang to figure out how to fix the issues in your list?

caomengmeng:
@daniceexi , I talked with @junxiawang  before opening this issue, and she suggests me to do so. 
I just list the issues maybe should updated, but not sure by myself and @junxiawang  will make final decision.
And for some obvious spelling mistakes, I just updated directly.

junxiawang:
@daniceexi ,I will modify it after meng meng read all the document.

whowutwut:
Assigned to @junxiawang , seems like she is working on this issue.

caomengmeng:
Track with PR: https://github.com/xcat2/xcat-core/pull/417

tingtli:
@caomengmeng  please verify and close it asap

caomengmeng:
Description:

Item1: No need to add  "See Suggestions for writing scripts". 

[...]

PR #417, opened 2015-11-17
modify prepostscripts according to bug 347

junxiawang:
modify prepostscripts file according to bug 347

### DIFF
--- a/docs/source/guides/admin-guides/manage_clusters/common/deployment/prepostscripts/post_script.rst
+++ b/docs/source/guides/admin-guides/manage_clusters/common/deployment/prepostscripts/post_script.rst
@@ -89,7 +89,7 @@ When your script is executed on the node, all the attributes in the site table a
 
 To run the postscripts, a script is built, so the above exported variables can be input. You can usually find that script in /xcatpost on the node and for example in the Linux case it is call mypostscript. A good way to debug problems is to go to the node and just run mypostscript and see errors. You can also check the syslog on the Management Node for errors.
 
-When writing you postscripts, it is good to follow the example of the current postscripts and write errors to syslog and in shell. See Suggestions for writing scripts.
+When writing you postscripts, it is good to follow the example of the current postscripts and write errors to syslog and in shell. See :ref:`Suggestions-for-write-scripts-label` .
 
 All attributes in the site table are exported and available to the postscript/postbootscript during execution. See the mypostscript file, which is generated and executed on the nodes to run the postscripts.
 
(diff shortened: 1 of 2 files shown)


whowutwut:
@junxiawang  you need to put a # before 347 to make the GitHub reference, you can edit the post above to add the # so that we can have a reference to the issue...

docs/source/guides/admin-guides/manage_clusters/common/deployment/prepostscripts/post_script.rst
```diff
@@ -89,7 +89,7 @@ When your script is executed on the node, all the attributes in the site table a
 
 To run the postscripts, a script is built, so the above exported variables can be input. You can usually find that script in /xcatpost on the node and for example in the Linux case it is call mypostscript. A good way to debug problems is to go to the node and just run mypostscript and see errors. You can also check the syslog on the Management Node for errors.
 
-When writing you postscripts, it is good to follow the example of the current postscripts and write errors to syslog and in shell. See Suggestions for writing scripts.
+When writing you postscripts, it is good to follow the example of the current postscripts and write errors to syslog and in shell. See :ref:`Suggestions-for-write-scripts-label` .
```
whowutwut: This section was copied over from SourceForge directly :-(   The fix here is not to create a link to a blank page, but to see whether the content is even worth bringing over to RTD. 

Here's the original page... 
http://sourceforge.net/p/xcat/wiki/Postscripts_and_Prescripts/#suggestions-for-writing-scripts

I think those tips are not very helpful without any examples.  I would suggest removing the link here...

docs/source/guides/admin-guides/manage_clusters/common/deployment/prepostscripts/suggestions.rst
```diff
@@ -1,3 +1,5 @@
+.. _Suggestions-for-write-scripts-label:
```
whowutwut: Suggest removing this page.  Add it back in later if we feel it's necessary.  Right now I don't think that "Suggestions for writing scripts" page over on SourceForge is adding anything valuable to the documentation.

docs/source/guides/admin-guides/manage_clusters/common/deployment/prepostscripts/post_script.rst
```diff
@@ -216,7 +216,7 @@ Example of mypostscript  ::
     run_ps script2
     # postscripts-end-here\n
 
-The mypostscript file is generated according to the mypostscript.tmpl file.
+The mypostscript file is generated according to the mypostscript.tmpl file and template details will show in below section.
```
whowutwut: I don't know how this improves the understanding...

whowutwut:
@junxiawang , I don't think we should merge this pull request.  There are 6 items in @caomengmeng Issue #347  and the changes here only attempt to resolve 2 of the 6.. what about the other 4?  Closing for now, please re-open when the changes are more complete.

tingtli:
Yes. I agree. Let's trace it in 2.12

Issue #418, opened 2015-11-17
 `addkitcomp -n` man page need to be  updated for its extra function


wwzhubj:
*man addkitcomp*
   *-n|--noupgrade*
            *Allow multiple versions of kitcomponent to be installed into*
              *the osimage, instead of kitcomponent upgrade*
 In some situation, we add "-n "not for "noupgrade", but for the package installation order, which is , those packages with "-n " will be installed behind those packages without "-n".

bybai:
manpage for "-n | --noupgrage" will be modified as:
1, Allow multiple versions of kitcomponent to be installed into the osimage, instead of kitcomponent upgrade
2, Allow these batch of kitcomponents to be installed separately, these kitcomponents will be installed behind those kitcomponents using addkitcomp without -n option

wwzhubj:
I have verified it.
[root@c910f03c05k07 ~]# lsxcatd -v
Version 2.11 (git commit 51316df775cb2e410708aab98c3afe9c8ef6ea1d, built Thu Nov 19 03:01:29 EST 2015)

PR #419, opened 2015-11-17
addkitcomp -n man page need to be updated for its extra fun…

bybai:
Old man page only explain --noupgrage function, details : #418

Add another function explanation in man page for "-n | --noupgrage" as:
1.  Allow multiple versions of kitcomponent to be installed into the osimage, instead of kitcomponent upgrade
2. Allow these batch of kitcomponents to be installed separately, these kitcomponents will be installed behind those kitcomponents using addkitcomp without -n option

### DIFF
--- a/xCAT-client/pods/man1/addkitcomp.1.pod
+++ b/xCAT-client/pods/man1/addkitcomp.1.pod
@@ -44,7 +44,9 @@ The osimage name that the kit component is assigning to.
 
 =item B<-n|--noupgrade>
 
-Allow multiple versions of kitcomponent to be installed into the osimage, instead of kitcomponent upgrade
+1. Allow multiple versions of kitcomponent to be installed into the osimage, instead of kitcomponent upgrade.
+
+2. Kit components added by addkitcomp -n will be installed separately behind all other ones which have been added.
 
 =item B<--noscripts>
 


bybai:
@daniceexi 
The original design for ``-n|--noupgrade`` is "allow multiple versions of kitcomponent to be installed into the osimage, instead of kitcomponent upgrade"; So we use noupgrade as long name.

Currently, we use -n to invoke new yum install session to resolve some mixed software dependencies. So -n has another useful usage that is not mentioned in man page or kit doc;
This pull request is to append this usage in man page.
But I do not think --noupgrade should be changed, it has been used years. 

Based on your suggestion, I think this is accuracy.
Kit components added by -n will be installed separately behind all other ones which have been added without -n.

bybai:
@daniceexi The additional explanation for ``-n`` is as following, please review:
2. Kit components added by addkitcomp -n will be installed separately behind all other ones which have been added.

daniceexi:

whowutwut:
Sorry, A little late for the review, the .rst file is not updated for the RTD manpage documentation. That is done using "create_man_pages.py" script, instructions are here: 
http://xcat-docs.readthedocs.org/en/latest/developers/guides/code/builds.html#man-pages
Some of the team already has configured their environment to be able to run this.  

We can push the missing .rst files into xcat-core/docs before we GA 12/14, but just FYI

Issue #420, opened 2015-11-17
[DEV] need to support "updatenode -o " option


penguhyang:
1. there are no description about updatenode -o option in man page.
2. need to add verification about the -o option if not enter OS.
3. updatenode -o did not work normally.
```
[root@c910f02c09p07 ~]# updatenode c910f02c09p08 -o rh6.7 
c910f02c09p08: Upgrading c910f02c09p08 to 
c910f02c09p08: (Error) Cannot not update rh67 to .  Linux distribution does not match
```

PR #423, opened 2015-11-17
support updatenode -o option to use for update

penguhyang:
fix issue #420

### DIFF
--- a/xCAT-server/lib/xcat/plugins/updatenode.pm
+++ b/xCAT-server/lib/xcat/plugins/updatenode.pm
@@ -283,6 +283,11 @@ sub preprocess_updatenode
     } else {
        undef $::SETSERVER;
     }
+    if (defined($OS)) {
+       $::OS=$OS;
+    } else {
+       undef $::OS;
+    }
 
     # display the usage if -h or --help is specified
     if ($HELP)
@@ -1153,6 +1158,11 @@ sub updatenode
        $::SETSERVER=$SETSERVER;
     } else {
        undef $::SETSERVER;
+    }
+    if (defined($OS)) {
+       $::OS=$OS;
+    } else {
+       undef $::OS;
     }
 
     #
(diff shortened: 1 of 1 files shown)


whowutwut:
Can you provide the new output that appears with the code change? In #420 you show the problem..

penguhyang:
@whowutwut  Thanks for review.
before fix
```
[root@c910f02c09p07 ~]# updatenode c910f02c09p08 -o rh6.7 
c910f02c09p08: Upgrading c910f02c09p08 to 
c910f02c09p08: (Error) Cannot not update rh67 to .  Linux distribution does not match
```
after fix
```
[root@c910f02c09p07 ~]# updatenode c910f02c09p08 -o rh6.7
c910f02c09p08: Upgrading c910f02c09p08 to rh6.7
c910f02c09p08: (Error) Missing install directory /install/rh6.7/ppc64/Server/
```

bybai:
@whowutwut @penguhyang 
``-o`` option is not in updatenode manpage/help/documention.
I think the code for -o in updatenode is not used for a long time.
We can reconsider to delete this -o option from code or pick up this option in later release.

penguhyang:
@bybai  right, I think the -o option is used for upgrade in the same os version for different release(for example rhels7 to rhels7.1). If user want to upgrade the OS version (for example rhels6 to rhels7) maybe reprovision is a good idea.

whowutwut:
What is the conclusion here... should we remove that option or fix it ?

bybai:
@whowutwut , It is not a good time to investigate it in current stage in 2.11?  We discussed this issue in china team, this issue will be removed into 2.12. We can reconsider remove it or fix it after a careful investigate in 2.12;

whowutwut:
Ok sounds good!

xCAT-server/lib/xcat/plugins/updatenode.pm
```diff
@@ -3247,14 +3257,30 @@ sub updateOS
     elsif ("$installOS$version" =~ m/rh/i)
     {
 
-        # Red Hat repository path - http://10.0.0.1/install/rhel5.4/s390x/Server/
-        $path = "http://$http$installDIR/$os/$arch/Server/";
-        if (!(-e "$installDIR/$os/$arch/Server/"))
+        my $verifyOS = $os;
+        $verifyOS =~ s/[^0-9]*([0-9]+).*/$1/;
+        if ($verifyOS < '7')
```
immarvin: please use ``if(xCAT::Utils->version_cmp($kversion,"7.0") < 0){``

PR #425, opened 2015-11-17
More minor documentation changes

mfoliveira:
…ines so they actually become separate lines.

### DIFF
--- a/docs/source/guides/admin-guides/basic_concepts/network_planning/xcat_net_planning.rst
+++ b/docs/source/guides/admin-guides/basic_concepts/network_planning/xcat_net_planning.rst
@@ -11,8 +11,8 @@ Service Nodes
 For very large clusters, xCAT has the ability to distribute the management operations to service nodes. This allows the management node to delegate all management responsibilities for a set of compute or storage nodes to a service node so that the management node doesn't get overloaded. Although xCAT automates a lot of the aspects of deploying and configuring the services, it still adds complexity to your cluster. So the question is: at what size cluster do you need to start using service nodes?? The exact answer depends on a lot of factors (mgmt node size, network speed, node type, OS, frequency of node deployment, etc.), but here are some general guidelines for how many nodes a single mgmt node (or single service node) can handle:
 
 * **[Linux]:**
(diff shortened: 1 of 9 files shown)


mfoliveira:
Hi..
I'm going through some more sections, and may put more commits in this branch soon.

bybai:
hi @mfoliveira, 
There are many commits in your pull requests. 
Could you use ``git commit --amend`` next time ?  
So that all the commits will be formed as one commit. 
You can refer to doc http://xcat-docs.readthedocs.org/en/latest/developers/github/pull_request.html#changing-pull-request

mfoliveira:
Hi @bybay

Sure! No problem.

It turned out to be so many commits because I've been editing the files
directly on github as I went through them.

I can amend them next time. If this is required for this PR too, just let
me know.

Thanks!

mfoliveira:
s/@bybay/@bybai/

whowutwut:
@mfoliveira I think you are replying directly to the GitHub notifications, but you can edit your post directly from the GitHub UI also.. if you made a mistake spelling. Just FYI 

Anyway, I think this is kind of a debate....    I do see @bybai  point about reducing the number of commits, but I don't think git cares how many commits we make.  I also don't like the use of ``--amend`` unless I made a mistake with my commit. 

For example, if I'm working on a small function and make ``commit1``, but then I found I spelled something wrong, I would much rather fix the spelling mistake in ``commit1`` rather then creating ``commit2``.   Or maybe I'm working on a function and I think I have completed a checkpoint, I might commit that set of work.  The next day, I find I forgot something relating to that functional set, and so I would amend to include the missing piece.

Other than that I think it's **not** good to amend after we start reviewing pull requests.  I might spend 1-2 hours reviewing some large code change, then make 2-3 small comments,... If the originator uses amend, I CANNOT quickly see the 2 or 3 changes I requested.  I have to review EVERYTHING again.   it's better to just commit again and I can just look at the new differences from where I last reviewed.

mfoliveira:
Hi @whowutwut 

Oh, cool. Thanks for the clarifications and rationale - really appreciated.

Issue #426, opened 2015-11-18
[fvt]2.11:bmcdiscover return message need modified on ubuntu 


junxiawang:
1.download the  latest build
2.env:
root@c910f03c05k05:~# cat /etc/*-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.3 LTS"
NAME="Ubuntu"
VERSION="14.04.3 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.04.3 LTS"
VERSION_ID="14.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
root@c910f03c05k05:~# uname -a
Linux c910f03c05k05 3.16.0-34-generic #47~14.04.1-Ubuntu SMP Fri Apr 10 17:49:46 UTC 2015 ppc64le ppc64le ppc64le GNU/Linux
root@c910f03c05k05:~# 
3.root@c910f03c05k05:~#  bmcdiscover -s nmap --range 50.3.9.1
Warning: other error for node-32030901
50.3.9.1
root@c910f03c05k05:~# echo $?
0
------------------------->I find the bmc but still give error info here so the message need modified

bybai:
In bmcdiscover.pm:
```
        } elsif ($output =~ /error : unauthorized name/){
            xCAT::MsgUtils->message("I", {data=>["Warning: bmc username is incorrect for $node"]}, $::CALLBACK);
        } elsif ($output =~ /RAKP \S* \S* is invalid/) {
            xCAT::MsgUtils->message("I", {data=>["Warning: bmc password is incorrect for $node"]}, $::CALLBACK);
        }
else {
        xCAT::MsgUtils->message("I", {data=>["Warning: other error for $node"]}, $::CALLBACK);-----------------> useless.
  }
```

PR #427, opened 2015-11-18
issue426 bmcdiscover return message need modified on ubuntu

bybai:
Details refer to #426 .

bmcdiscover just checks if username or password is correct for bmc, and return bmcip or node definition. So there is a confused and useless warning message "Warning: other error for $node".


zet809:
Seems OK to me, I agree to merge.

PR #429, opened 2015-11-18
Dirty hack, fix the permission of direcotry /var/log/xcat

neo954:
On Ubuntu, the ``rsyslogd`` daemon write log files with ``syslog:adm`` permissison. And in some case, the directory ``/var/log/xcat`` was created by xCAT, and had ``root:root`` ownership. In this way, ``rsyslogd`` did not have enough permission to write to log files under this directory.

As a dirty hack, change the ownership of directory ``/var/log/xcat`` to the same ownership of directory ``/var/log``.

### DIFF
--- a/xCAT/debian/postinst
+++ b/xCAT/debian/postinst
@@ -20,12 +20,12 @@ set -e
 
 case "$1" in
     configure)
-         #make dir for the current pid
-         mkdir -p /var/run/xcat
-	 cp -f /opt/xcat/share/xcat/scripts/xHRM /install/postscripts/
-	. /etc/profile.d/xcat.sh
-	if [ -f /tmp/xCAT_upgrade.tmp ]
-	then
+        # Make dir for the current pid
+        mkdir -p /var/run/xcat
+        cp -f /opt/xcat/share/xcat/scripts/xHRM /install/postscripts/
+        . /etc/profile.d/xcat.sh
+        if [ -f /tmp/xCAT_upgrade.tmp ]
+        then
             if [ -r "/tmp/xcat/installservice.pid" ]; then
                 mv /tmp/xcat/installservice.pid /var/run/xcat/installservice.pid
             fi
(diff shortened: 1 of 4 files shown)


immarvin:
hi @neo954 , I have 2 comments:
1. the ``chown`` logic should be placed in ``syslog`` postscript instead of the spec file
2. maybe we can consider to use ``chown --reference`` 
````
       --reference=RFILE
              use RFILE's owner and group rather than specifying OWNER:GROUP values
````

File: xCAT/debian/postinst [hunk 1]
```diff
@@ -20,9 +20,9 @@ set -e
 
 case "$1" in
     configure)
-         #make dir for the current pid
-         mkdir -p /var/run/xcat
-	 cp -f /opt/xcat/share/xcat/scripts/xHRM /install/postscripts/
+        # Make dir for the current pid
+        mkdir -p /var/run/xcat
+	cp -f /opt/xcat/share/xcat/scripts/xHRM /install/postscripts/
```
samveen: Is the indentation still off?

xCAT/debian/postinst
```diff
@@ -41,11 +41,23 @@ case "$1" in
 	    xcatconfig -i -d -s
 	fi
 
-#        [ -e /etc/apache2/conf-enabled/xcat.conf ] && rm /etc/apache2/conf-enabled/xcat.conf
-         ln -s -f /etc/apache2/conf-available/xcat.conf.apach24 /etc/apache2/conf-enabled/xcat.conf
+        ln -s -f /etc/apache2/conf-available/xcat.conf.apach24 /etc/apache2/conf-enabled/xcat.conf
```
samveen: Here too?

xCAT/debian/postinst
```diff
@@ -41,11 +41,23 @@ case "$1" in
...
 
 	/etc/init.d/apache2 restart
 
+        # On Ubuntu, the rsyslogd daemon write log files with syslog:adm
```
samveen: Here onwards as well, by the looks of it.

xCAT/debian/postinst
whowutwut: This line is still off in the indentation, it's a tab where the others are spaces.

whowutwut: Line 26-28 and 39-42 also tabs..

PR #430, opened 2015-11-18
Fix mix of invalid token and grouped nodes

jjohnson42:
When a parenthetical group occurred after an invalid token,
it failed to report the error.  Address this problem.


daniceexi:
@whowutwut This change has been checked in to 2.11 1 month ago.

daniceexi:
Refer to commit: ab5a84d3952ecf228d4595587afb7dd19ed97081

PR #441, opened 2015-11-20
Fix-issue 442: Add sleep function before upgrade firmware

chenglch:
Firmware or ipmitool-xcat may not stable enough. BMC can handle
the ipmi request from ipmitool, but can not be upgraded succesfully,
add a sleep function as a work around.

Fix-issue: #442

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -1659,6 +1659,7 @@ sub do_firmware_update {
             return -1;
     }
     #check reset status
+    sleep(10);
     unless (check_bmc_status_with_ipmitool($pre_cmd, 5, 12)) {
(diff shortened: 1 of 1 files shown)


whowutwut:
@pdlun92  Can you take a look at the code changes? 

@chenglch  Is sleep the best way to handle this?  What would happen in the case of running the firmware update on the cmdline using ipmitool....  would it also fail?

chenglch:
@whowutwut , The phenomenon is that, the ipmitool command(power status, raw code) works correctly, the last step hpm upgrade will print the upgrade process table( this output is right), but the upgrade process stays at 0%, then the upgrade command will return the error code.  

@tingting  has already tested this fix, it works well.  But I haven't found a better way to control the time to execute the upgrade command after cold reset of bmc.

whowutwut:
Ok, thanks for the explanation

Issue #440, opened 2015-11-20
IBM Power S812LC using rspconfig to set bmc ip address failed if the bmc is in dhcp mode 


zet809:
# rspconfig dn22 ip
dn22: BMC IP: 50.2.0.28

 # rspconfig dn22 ip=50.2.0.72
dn22: BMC IP: 50.2.0.28

 # rspconfig dn22 ip
dn22: BMC IP: 50.2.0.28

 # ipmitool-xcat -I lanplus -H 50.2.0.28 -U xxx -P xxx lan print 1
Set in Progress         : Set Complete
Auth Type Support       : MD5 
Auth Type Enable        : Callback : MD5 
                        : User     : MD5 
                        : Operator : MD5 
                        : Admin    : MD5 
                        : OEM      : MD5 
IP Address Source       : DHCP Address
IP Address              : 50.2.0.28
Subnet Mask             : 255.0.0.0
MAC Address             : a0:42:3f:30:62:5e
SNMP Community String   : AMI
IP Header               : TTL=0x40 Flags=0x40 Precedence=0x00 TOS=0x10
BMC ARP Control         : ARP Responses Enabled, Gratuitous ARP Disabled
Gratituous ARP Intrvl   : 0.0 seconds
Default Gateway IP      : 0.0.0.0

Default Gateway MAC     : 00:00:00:00:00:00
Backup Gateway IP       : 0.0.0.0
Backup Gateway MAC      : 00:00:00:00:c0:a4
802.1q VLAN ID          : Disabled
802.1q VLAN Priority    : 0
RMCP+ Cipher Suites     : 
Cipher Suite Priv Max   : Not Available

zet809:
Pull request 443.

whowutwut:
@zet809  Can you modify the above and wrap in markdown.  

You are trying to set dn22 to 50.2.0.72 and it does not work? Can you use rspconfig to display the existing configured IP address? Before and after?  
For my knowledge, looking at ipmitool output against 50.2.0.28, how can you tell that BMC is in DHCP mode? Does the rspconfig work if the BMC is not in DHCP mode? 

Trying to understand the problem so I can better review the fix.

zet809:
@whowutwut I have modify the defect description.
For the 2nd question, you can see line "IP Address Source : DHCP Address" from output of ipmitool which means the BMC get ip address through DHCP. 
[...]

PR #443, opened 2015-11-20
fix issue 440: IBM Power S812LC using rspconfig to set bmc ip address…

zet809:
… failed if the bmc is in dhcp mode

Fix for Issue #440

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -827,6 +827,12 @@ sub setuseraccess {
     return;
 }
 
+sub next_setnetinfo {
+    my $rsp = shift;
+    my $sessdata = shift;
+    &setnetinfo($sessdata);
+}
+
 sub setnetinfo {
     my $sessdata = shift;
     my $subcommand = $sessdata->{subcommand};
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -931,7 +945,8 @@ sub setnetinfo {
     }
     unless ($sessdata->{netinfo_setinprogress}) {
         $sessdata->{netinfo_setinprogress} = '1';
-        $sessdata->{ipmisession}->subcmd(netfn=>$netfun, command=>0x01, data=>[$channel_number,0x0,0x1], callback=>\&setnetinfo,callback_args=>$sessdata);
+        $sessdata->{ipmisession}->subcmd(netfn=>$netfun, command=>0x01, data=>[$channel_number,0x0,0x1], callback=>\&next_setnetinfo,callback_args=>$sessdata);
+        return;
```
whowutwut: If I'm reading this correctly, this means that we are going to set "in progress" for any command that we are going to issue to the BMC in this function.   Once it's set to "in progress" it's going to call the ``next_setnetinfo`` in the call back and run through the code again.. is that right?

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -907,6 +913,14 @@ sub setnetinfo {
     } elsif ($subcommand =~ m/ip/ and $argument =~ m/dhcp/) {
         @cmd = (0x01,$channel_number,0x4,0x2);
     } elsif ($subcommand =~ m/ip/) {
+        # Need to set ipsrc to static for IBM POWER S822LC and S812LC
+        if (exists($sessdata->{netinfo_setinprogress})) {
+            unless($sessdata->{set_ipsrc_static}) {
+                $sessdata->{set_ipsrc_static} = 1;
+                $sessdata->{ipmisession}->subcmd(netfn=>$netfun, command=>0x01, data=>[$channel_number,0x04,0x1], callback=>\&next_setnetinfo,callback_args=>$sessdata);
```
whowutwut: Again the same comment as below, if it's not set to static IP , we are going to change BMC to static IP, then send the command... then on the call back we will run though this code and then set the IP address.   Is that why we need to return early here instead of allowing the command to run down on line 952?

xCAT-server/lib/xcat/plugins/ipmi.pm
zet809: Yes, correct. The logic is:
1. setting in-progress
2. set ip src to be static
3. set ip address
4. setting done.

whowutwut:
@zet809  Thanks for the response.  I reviewed the code and will merge

xCAT-server/lib/xcat/plugins/ipmi.pm
zet809: Yes, if we haven't set ip source to static, we need to do that and then go into setnetinfo from next_setnetinfo (called by the callback process) to set ip address.

PR #454, opened 2015-11-23
fix issue 444: Timeout while executing "renergy noderange temperature…

zet809:
…|powerusage"


chenglch:
I understand the cause for ipmi timeout, but I think we should modify the comment for OpenPOWER server which get in 'ipmi.pm'   https://github.com/xcat2/xcat-core/blob/master/perl-xCAT/xCAT/Utils.pm#L3516

zet809:
Add more comments, pls review.

chenglch:
LGTM

PR #456, opened 2015-11-23
add disc selecting logic for powerkvm3.1.0

immarvin:


xCAT-server/share/xcat/install/scripts/pre.pkvm3
```diff
@@ -0,0 +1,58 @@
...
+		"mptsas"|"mpt2sas"|"mpt3sas") #*PROBABLY* not SAN, but SAS SAN is possible
+			if [ -z "$probablyfirstdirectdisk" ]; then probablyfirstdirectdisk=$disk; fi #remember first disk as a guess of medium resort
+			;;
+		*)
+			if [ -z "$firstdisk" ]; then firstdisk=$disk; fi #remember first disk as a guess of medium resort
+			;;
+		esac
+	done
+fi
+
```
zet809: For the 2 sections: 1 check /proc/paritions and 2 check /dev/sd*, why not merge them into 1 like this?
For every disk that can list:
    check the driver
         driver is ok, check diskname/devwwn, and write into temp files.
         Not ok, go the next partition device.
For the temp files, sort the output and choose the inst_disk.

xCAT-server/share/xcat/install/scripts/pre.pkvm3
```diff
@@ -0,0 +1,58 @@
...
+#first take the disk with the MAX WWN numer to be the installdisk
+if [ -z "$instdisk" ]; then
+   disks=$(awk -F' '  '{print $4  }' /proc/partitions |grep -e "sd[a-z]$")
+   for disk in $disks; do
+      diskinfo=$(udevadm info --name $disk)
+      diskname=$(IFS= ;echo $diskinfo|grep '\<DEVNAME\>'|cut -d "=" -f2|tr A-Z a-z)
+      devtype=$(IFS= ;echo $diskinfo|grep '\<DEVTYPE\>'|cut -d "=" -f2|tr A-Z a-z)
+      devwwn=$(IFS= ;echo $diskinfo|grep '\<ID_WWN\>'|cut -d "=" -f2|tr A-Z a-z)
+      [  "$devtype" != "disk" ]  && continue
```
zet809: The devtype is useless if check the device driver first.

xCAT-server/share/xcat/install/scripts/pre.pkvm3
immarvin: thelogic here is from redhat/sles  prescript, I would like to keep it here as the backup when the newly added logic failed to determine the disk to partition. I am wondering whether we should delete or keep the logic here.  I suggest that we keep it for this release and refine it and make it consistent for redhat,sles and ubuntu in next release.

xCAT-server/share/xcat/install/scripts/pre.pkvm3
immarvin: yes

xCAT-server/share/xcat/install/scripts/pre.pkvm3
```diff
@@ -0,0 +1,58 @@
+#find first available block device (sda or vda likely)
+#TODO: pick a likely non-SAN target if possible
+
+shopt -s nullglob
+
+#first take the disk with the MAX WWN numer to be the installdisk
+if [ -z "$instdisk" ]; then
+   disks=$(awk -F' '  '{print $4  }' /proc/partitions |grep -e "sd[a-z]$")
```
zet809: I think it shall sd[a-z]+  here for the scenario that there is a lot of disks.

xCAT-server/share/xcat/install/scripts/pre.pkvm3
zet809: I mean is it possible the it is not a disk in $disks?
We don't need to check the devtype here, right?

immarvin: this piece of code does not focus on the device driver, it just check the WWN. The guess on device driver is done in the logic below if the newly added logic cannot determine the disc.

xCAT-server/share/xcat/install/scripts/pre.pkvm3
immarvin: the available disk should be "sda,sdb...", I think "sd[a-z]" is ok. "sda1,sda2" are partitions, which cannot be used to create partitions.

xCAT-server/share/xcat/install/scripts/pre.pkvm3
zet809: OK

zet809:

PR #468, opened 2015-11-25
fix issue 462: The result of "rvitals wattage" is not right on firest…

zet809:
…one and habanero


bybai:
I think it is a safe fix.

Issue #474, opened 2015-11-26
[dev]pasu cannot work in sles11.4 and return SSL_verify_mode error at /opt/xcat/bin/pasu line 283 in RH7.1


bybai:
This is from PMR .
 pasu return SSL_verify_mode error at /opt/xcat/bin/pasu line 283 error  in redhat7.1
 pasu Connection failure: Connection reset by peer

On redhat7.1
```
/]# lsxcatd -v
Version 2.11 (git commit 06c950d57848ee04786c19eff6bf075e6f9f0ea9, built Thu Nov 19 19:30:55 EST 2015)

8 /]# pasu c910f04x20 show
*******************************************************************
 Using the default of SSL_verify_mode of SSL_VERIFY_NONE for client
 is deprecated! Please set SSL_verify_mode to SSL_VERIFY_PEER
 together with SSL_ca_file|SSL_ca_path for verification.
 If you really don't want to verify the certificate and keep the
 connection open to Man-In-The-Middle attacks please set
 SSL_verify_mode explicitly to SSL_VERIFY_NONE in your application.
*******************************************************************
  at /opt/xcat/bin/pasu line 283.
c910f04x20: IMM.SSH_SERVER_KEY=Installed
c910f04x20: IMM.SSL_HTTPS_SERVER_CERT=Private Key and CA-signed cert installed.
c910f04x20: IMM.SSL_HTTPS_SERVER_CSR=Private Key and CA-signed cert installed.
c910f04x20: IMM.SSL_LDAP_CLIENT_CERT=Private Key and Cert/CSR not available.
```

On sles11.4:
```
/opt/xcat/bin # lsxcatd -v
Version 2.11 (git commit 6d9ff85e1c412943012fb0ebbeb7f0f44d53b29b, built Wed Nov 25 15:30:52 EST 2015)
/opt/xcat/bin # pasu c910f04x18 show
Connection failure: Connection reset by peer
```

bybai:
Related code line:
```
  my $client = IO::Socket::SSL->new(
                PeerAddr=>$xcathost,
                SSL_key_file=>$homedir."/.xcat/client-cred.pem",
                SSL_cert_file=>$homedir."/.xcat/client-cred.pem",
                SSL_ca_file => $homedir."/.xcat/ca.pem",
                SSL_use_cert => 1,
                #SSL_verify_mode => 1,
             );
```

caomengmeng:
[...]

PR #475, opened 2015-11-26
pasu/psh/pscp Connection failure in sles11.4,error info in rh7.1 

bybai:
details refer to :  #474 #479 
pasu return SSL_verify_mode error at /opt/xcat/bin/pasu line 283 in RH7.1
pasu Connection failure: Connection reset by peer on sles11

psh and pscp have the same problem.

### DIFF
--- a/xCAT-client/bin/pasu
+++ b/xCAT-client/bin/pasu
@@ -231,13 +231,19 @@ sub expandnoderange {
   my @nodes;
   my @user = getpwuid($>);
   my $homedir=$user[7];
+  my %sslargs;
+  if (defined($ENV{'XCATSSLVER'})) {
+      $sslargs{SSL_version} = $ENV{'XCATSSLVER'};
+  }
+
   my $client = IO::Socket::SSL->new(
                 PeerAddr=>$xcathost,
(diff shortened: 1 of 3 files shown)


xCAT-client/bin/pasu
```diff
@@ -286,6 +286,7 @@ sub getipmiattrs {
                 SSL_cert_file=>$homedir."/.xcat/client-cred.pem",
                 SSL_ca_file => $homedir."/.xcat/ca.pem",
                 SSL_use_cert => 1,
+                SSL_verify_mode => "SSL_VERIFY_PEER",
```
chenglch: A little strange, I think SSL_VERIFY_PEER is a enum variable or constant value.
Maybe it is  SSL_verify_mode => SSL_VERIFY_PEER
Correct me if I am wrong.

bybai:
Hi @chenglch , Both SSL_verify_mode => SSL_VERIFY_PEER and SSL_verify_mode => 1 can work here. But it is not enough for the defect. see my new fix.

neo954:
Okay, I think the change is minor, and the patch looks good.

chenglch:
LGTM

PR #477, opened 2015-11-26
add xcat_port document

xuweibj:
Task 75882

http://10.3.5.4/install/xcat-doc/html/advanced/ports/xcat_ports.html

### DIFF
--- a/docs/source/advanced/ports/xcat_ports.rst
+++ b/docs/source/advanced/ports/xcat_ports.rst
@@ -1,4 +1,281 @@
 Port Usage
 ==========
 
-TODO: http://sourceforge.net/p/xcat/wiki/XCAT_Port_Usage/?version=4
+The following table lists the ports that must be open between the xCAT management node and the nodes it manages, unless otherwise noted. The xCAT service nodes use the same ports as the management node. A service (or protocol) applies to both AIX and Linux, unless stated otherwise. Service names are typical strings that appear in the /etc/services file, or in firewall/IP filtering logs. Local customization of the /etc/services files, daemon configuration options, like overriding the default port number, and differences in software source implementations, may yield other service information results.
+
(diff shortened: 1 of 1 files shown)


docs/source/advanced/ports/xcat_ports.rst
```diff
@@ -1,4 +1,273 @@
...
+
+would require that the /etc/services file on ms2112 be updated with something similar to the following:
+
+mountd 33333/tcp mountd 33334/udp
+
+For mountd to detect its new port values you must stop and start rpc.mountd. The stopping and starting of mountd takes place on the same host where the /etc/services file mountd updates were made. In the above example, ms2112's mountd is stopped and started. You can verify that mountd is using the new port definitions by issuing the rpcinfo command.
+
+This procedure shows how to change ports used by mountd:
+
+lssrc -s rpc.mountd
```
immarvin: just a reminder on the format
1. the command and its output should be in a literal block, like:
````
lssrc -s rpc.mountd
```` 
The files and service names should be in decorated words or key words, such as ``/etc/services`` or **syslogd**

please refer to the https://xcat-docs.readthedocs.org/en/latest/developers/doc_guidelines.html#list-structure

bybai:
It is a useful port documentation. I will merge it first.

whowutwut:
A request came to me asking how to run SeLINUX with xCAT.   I don't think we have a clear understanding or directions to help user configure this. Can  we should enhance  our port section give details on what firewall configurations must be made in order to run in a more secure environment?

daniceexi:
@immarvin Could we take a try what's going on if we do NOT disable the SELINUX setting in the kickstart configuration file? Maybe the installation or postscript process cannot continue. We can take a look how many work around we need to make it finish the OS deployment with SELINUX.

Issue #471, opened 2015-11-25
[FVT]2.11 fireware update  failed on some of the Firestone  machine when concurrently do rflash


tingtli:
xCAT 2.11 on rh7.2
[root@c910f03c05k23 xcat-dep]# lsxcatd -v
Version 2.11 (git commit 9885d9acba568e21137ca1be624d0ba8b1655091, built Tue Nov 24 03:05:32 EST 2015)
[root@c910f03c05k23 firmware]# rflash c910f05c37,c910f05c33 8335_810.1549.20151116c.hpm
c910f05c37: rflash started, please wait.......
c910f05c33: rflash started, please wait.......
c910f05c33: Error: Running ipmitool command /opt/xcat/bin/ipmitool-xcat -H 50.5.33.10 -I lanplus -U ADMIN -P passw0rd -z 30000 hpm upgrade /firmware/8335_810.1549.20151116c.hpm force failed.
c910f05c37: rflash completed.

chenglch:
When I set the sleep time to 120s @tingtli has success to update 3 Firestone machines  concurrently.
I haven't found the root cause, so I prepare to  add a timeout option to set the sleep time for rflash command  at first

whowutwut:
@chenglch , yeah this is the concern when adding sleep time... It would be best if we could query for some status (say.. every 5-10 seconds)  then continue if we can detect some status change.... Otherwise we are always guessing the right timeout.. which won't work every time..

chenglch:
The root cause for this issue is from internal of ipmitool. ``rflash`` is not stable enough and a workaround is provided. We'd better not to use  ``concurrently`` to describe this issue as ``rflash`` also fails in some case with only one node.

whowutwut:
@tingtli , @pdlun92 made pull request #818 which should improve the stability of rflash.

whowutwut:
Closing this issue for now and I think xCAT is doing the right thing for the ``rflash`` command with the pull requests associated with this issue.  

Will open new issues if we see them on the large systems when running rflash with 2.11.1 build.

PR #486, opened 2015-11-27
Fix-issue 471: Add a user interface to set the delay time for rflash

chenglch:
Currently abort failure(0x81) occur during the initiate stage as
the firmware state is not stable enough after cold reset. As the
delay time is not a constant value, this change add a interface in
site table for user to set this value.

Close-issue: #471

### DIFF
--- a/perl-xCAT/xCAT/Schema.pm
+++ b/perl-xCAT/xCAT/Schema.pm
@@ -1070,6 +1070,9 @@ site => {
    "              through HMC. Default is 0.\n\n".
    " snmpc:  The snmp community string that xcat should use when communicating with the\n".
    "         switches.\n\n".
+   " rflash_delay:  Only for rflash command to upgrade firmware on OpenPOWER servers.\n".
+   "                This attribute is delay time to wait for the firmware initialization\n".
+   "                after cold reset. Default value is 120 in seconds. If set to 0, no delay.\n\n".
    " ---------------------------\n".
    "INSTALL/DEPLOYMENT ATTRIBUTES\n".
    " ---------------------------\n".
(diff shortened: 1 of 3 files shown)


xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1692,7 +1692,11 @@ sub do_firmware_update {
     # NOTE(chenglch) some firmware may not stable enough, it can handle the ipmi session
     # request, but failed to upgrade, add sleep function as a work around here to avoid of
     # error.
-    sleep(60);
+    my $rflash_delay = 120;
+    my $sitetab = xCAT::Table->new("site");
+    my $delay_config = $sitetab->getAttribs({key => "rflash_delay"}, 'value');
```
zet809: Using xCAT::TableUtils->get_site_attribute is better.

perl-xCAT/xCAT/Schema.pm
```diff
@@ -1070,6 +1070,9 @@ site => {
    "              through HMC. Default is 0.\n\n".
    " snmpc:  The snmp community string that xcat should use when communicating with the\n".
    "         switches.\n\n".
+   " rflash_delay:  Only for rflash command to upgrade firmware on OpenPOWER servers.\n".
+   "                This attribute is delay time to wait for the firmware initialization\n".
+   "                after cold reset. Default value is 120 in seconds.\n\n".
```
zet809: No explain for the value 0.  0 means not waiting or waiting until succeed?

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1692,7 +1692,11 @@ sub do_firmware_update {
...
+    $rflash_delay = $delay_config->{'value'} if $delay_config;
```
zet809: If $delay_config is 0, the code will use 120s. Correct?

bybai: If do not set rflash_delay in site table, the default sleep is 120s.

zet809:
Looks OK to me, I agree to merge.

Issue #488, opened 2015-11-27
[fvt]2.11:xcatsn installation error  in rhels7.2


junxiawang:
In xcatsn.spec:
````
version=`rpm -qi httpd 2>&1 | grep 'Version' | awk '$NF { print $3 }' | cut -d. -f1,2` 
4:53:34 PM: Song Yang: compare_version="2.4" 
4:53:39 PM: Song Yang: version=`rpm -qi apache2 2>&1 | grep '^Version' | awk '$NF { print $3 }' | cut -d. -f1,2` 
4:53:48 PM: Song Yang: bc <<< "$version >= $compare_version" 
5:40:31 PM: Song Yang: 

%post
%ifos linux
#Apply the correct httpd/apache configuration file according to the httpd/apache version

compare_version="2.4"
version=`rpm -qi httpd 2>&1 | grep 'Version' | awk '$NF { print $3 }' | cut -d. -f1,2`
if [ -n "$version" ]; then
    if (( `bc <<< "$version >= $compare_version"` )); then
        rm -rf /etc/httpd/conf.d/xcat.conf
        cp /etc/xcat/conf.orig/xcat.conf.apach24 /etc/httpd/conf.d/xcat.conf
    fi
fi

version=`rpm -qi apache2 2>&1 | grep '^Version' | awk '$NF { print $3 }' | cut -d. -f1,2`
if [ -n "$version" ]; then
    if (( `bc <<< "$version >= $compare_version"` )); then
        rm -rf /etc/apache2/conf.d/xcat.conf
        cp /etc/xcat/conf.orig/xcat.conf.apach24 /etc/apache2/conf.d/xcat.conf
    fi
fi
````
could not find bc in xcatsn

immarvin:
pull request https://github.com/xcat2/xcat-core/pull/489

PR #489, opened 2015-11-27
fix defect [fvt]2.11:xcatsn installation error in rhels7.2 #488

immarvin:
[fvt]2.11:xcatsn installation error in rhels7.2 #488

### DIFF
--- a/xCATsn/xCATsn.spec
+++ b/xCATsn/xCATsn.spec
@@ -28,6 +28,7 @@ Conflicts: xCAT
 # on RHEL7, need to specify it explicitly
 Requires: net-tools
 Requires: /usr/bin/killall
+Requires: /usr/bin/bc
 # yaboot-xcat is pulled in so any SN can manage ppc nodes
 Requires: httpd nfs-utils nmap bind
 # On RHEL this pulls in dhcp, on SLES it pulls in dhcp-server


whowutwut:
The change will just pull in a requirement for bc by the xCATsn rpm.  I'm assuming that when we provision the ``*-service`` osimage definitions, it will add the bc rpm in as a dependency and it will get pulled in.  I don't think we need to add to the pkglist right?

immarvin:
yes, since ``xCATsn`` is in the ``service-*.pkglist``, ``bc`` will be installed as a dependency of ``xCATsn``, we do not need to add it to pkglist explicitly.

whowutwut:
Thanks. Looks good

daniceexi:
@immarvin The code logic is used to generate/copy the corresponding httpd configuration file base on httpd version. I think the code logic should be same between xCAT and xCATsn. But current they are not. We need to take a look what was missed.

immarvin:
@daniceexi , xCATsn.spec is modified while fixing this defect https://sourceforge.net/p/xcat/bugs/4642/.
````
commit 896877deb7ebef264a46a67743db80af059a81a0
Author: Casandra Qiu <cxhong@us.ibm.com>
Date:   Wed Apr 15 14:08:40 2015 -0400

    Bug #4642 httpd didn't start up for rhels7.1 diskless server node

diff --git a/xCATsn/xCATsn.spec b/xCATsn/xCATsn.spec
index df24475..e617c30 100644
--- a/xCATsn/xCATsn.spec
+++ b/xCATsn/xCATsn.spec
@@ -137,16 +137,22 @@ fi
 %post
commit 896877deb7ebef264a46a67743db80af059a81a0
Author: Casandra Qiu <cxhong@us.ibm.com>
Date:   Wed Apr 15 14:08:40 2015 -0400

    Bug #4642 httpd didn't start up for rhels7.1 diskless server node

diff --git a/xCATsn/xCATsn.spec b/xCATsn/xCATsn.spec
index df24475..e617c30 100644
--- a/xCATsn/xCATsn.spec
+++ b/xCATsn/xCATsn.spec
@@ -137,16 +137,22 @@ fi
 %post
 %ifos linux
 #Apply the correct httpd/apache configuration file according to the httpd/apache version
-if [ -n "$(httpd -v 2>&1 |grep -e '^Server version\s*:.*\/2.4')" ]
-then
-   rm -rf /etc/httpd/conf.d/xcat.conf
-   cp /etc/xcat/conf.orig/xcat.conf.apach24 /etc/httpd/conf.d/xcat.conf
+
+compare_version="2.4"
+version=`rpm -qi httpd 2>&1 | grep 'Version' | awk '$NF { print $3 }' | cut -d. -f1,2`
+if [ -n "$version" ]; then
+    if (( `bc <<< "$version >= $compare_version"` )); then
+        rm -rf /etc/httpd/conf.d/xcat.conf
+        cp /etc/xcat/conf.orig/xcat.conf.apach24 /etc/httpd/conf.d/xcat.conf
+    fi
 fi
 
-if [ -n "$(apachectl -v 2>&1 |grep -e '^Server version\s*:.*\/2.4')" ]
-then
-   rm -rf /etc/apache2/conf.d/xcat.conf
-   cp /etc/xcat/conf.orig/xcat.conf.apach24 /etc/apache2/conf.d/xcat.conf
+version=`rpm -qi apache2 2>&1 | grep '^Version' | awk '$NF { print $3 }' | cut -d. -f1,2`
+if [ -n "$version" ]; then
+    if (( `bc <<< "$version >= $compare_version"` )); then
+        rm -rf /etc/apache2/conf.d/xcat.conf
+        cp /etc/xcat/conf.orig/xcat.conf.apach24 /etc/apache2/conf.d/xcat.conf
+    fi
 fi
````
I think ``bc`` is a powerful tool for version compare, we should use it in the logic of xCAT.spec and xCAT::Utils::version_cmp.

but I think the problem of using ``rpm -qi `` instead of  ``httpd -v`` is that ``rpm`` is not supported in ubuntu.

whowutwut:
@cxhong FYI , the discussion above...

PR #491, opened 2015-11-28
Cuda pkglist for 2.11

zet809:


bybai:
Discussed this pull request in team meeting,cuda version should not be specified in otherpkgs.list;
 If CUDA version is specified in otherpkgs.pkglist as "cuda-7.5/ppc64le/cuda-deps/dkms", cuda version may be changed soon, and this otherpkgs.pkglist is till invalid; 
Since current doc+code works, Er tao will only add enhancements in doc later. this pull request will be closed.

PR #493, opened 2015-11-30
Remove the x86_64 documentation pages and direct to SourceForge for 2.11 GA

whowutwut:
Cleanup and remove the empty pages for the x86_64 documentation as we prepare for the 2.11 GA.  This is so we don't have blank pages in the documentation. 

Remove the blank framework pages  for managing xCAT cluster and point users to the existing SourceForge documentation for now until we have an opportunity to add the documentation for System X.  This should be done after the System P documentation is in good shape to mimic.


whowutwut:
@immarvin, @tingtli , lets try to merge this before creating the 2.11 branch.

immarvin:
I will merge this now, and we can roll back to the point before this commit when we start to construct the X86_64 documentation..

PR #497, opened 2015-12-01
Add a global cfg to enable the longnamefirst

xuweibj:
For a lot of no-HPC users, they will use `resolveip` or other tools or functions shipped by program language to check the hostname of a specified IP address. They will prefer the long hostname (FQDN) which mean the hostname plus domain name to be available first. For xCAT, the short hostname is put first in /etc/hosts which will cause the no-HPC user concerns. 
So we plan to provide a method by which the generated /etc/hosts entry will be have the long hostname appeared before shot hostname.


xCAT-server/sbin/xcatconfig
```diff
@@ -1173,7 +1173,7 @@ sub initDB
         $chtabcmds .= "$::XCATROOT/sbin/chtab key=auditnosyslog site.value=0;";
         $chtabcmds .= "$::XCATROOT/sbin/chtab key=xcatsslversion site.value=TLSv1;";
         #$chtabcmds .= "$::XCATROOT/sbin/chtab key=useflowcontrol site.value=yes;"; # need to fix 4031
-
+
```
neo954: One redundance space is here, should be removed.

docs/source/guides/admin-guides/references/man8/makehosts.8.rst
```diff
@@ -42,6 +42,7 @@ Note: If your node hostnames and IP addresses follow a regular pattern, you can
 
 If you specify additional network interfaces in your xCAT node definitions they will also be added to the /etc/hosts file.  You can specify additional network interface information (NICs) using the following node attributes: nicips, nichostnamesuffixes, nictypes, niccustomscripts, nicnetworks.  You can get a description of these attributes by running "lsdef -t node -h | more" or "man nics".
 
+After xCAT is installed, the attribute name "longnamefirst" can be added into "site" table manually.  If the value is set as "1", "yes" or "enable", the /etc/hosts entries generated by "makehosts" will be have the long hostname put before the short hostname.  If the value is set as "0", "no", "disable" or doesn't set at all, the original behavior won't change for "makehosts".
```
neo954: "... will ~~be~~ have the long hostname ..." I think the word ``be`` should be deleted.

And in the code, there is no place to check if the value is ``0`` or ``no`` or ``disable``. So do not say it in the document. You can say ...

"Otherwise, the original behavior will be performed."

docs/source/guides/admin-guides/references/man8/makehosts.8.rst
```diff
@@ -42,6 +42,7 @@ Note: If your node hostnames and IP addresses follow a regular pattern, you can
 
 If you specify additional network interfaces in your xCAT node definitions they will also be added to the /etc/hosts file.  You can specify additional network interface information (NICs) using the following node attributes: nicips, nichostnamesuffixes, nictypes, niccustomscripts, nicnetworks.  You can get a description of these attributes by running "lsdef -t node -h | more" or "man nics".
 
+After xCAT is installed, the attribute name "longnamefirst" can be added into "site" table manually.  If the value is set as "1", "yes" or "enable", the /etc/hosts entries generated by "makehosts" will have the long hostname put before the short hostname.  Otherwise, the original behavior will be performed.
```
whowutwut: This change should not be in the .rst file directly.  
The man page is generated from the .pod file so the changes need to be made in the pod file:
``xcat-core/xCAT-client/pods/man8/makehosts.8.pod``

Then there's a script to generate the man pages under docs... ``xcat-core/create_man_pages.py``

Instructions for running the create_man_pages.py script are here: http://xcat-docs.readthedocs.org/en/2.11/developers/guides/code/builds.html#man-pages

xCAT-server/lib/xcat/plugins/hosts.pm
```diff
@@ -175,7 +175,12 @@ sub build_line
     }
 
     $othernames = join(' ', @o_names);
-    if    ($LONGNAME)        { return "$ip $longname $shortname $othernames\n"; }
+
+    # read site.longnamefirst attribute
+    my $longnamefirst = $::XCATSITEVALS{longnamefirst};
+
```
zet809: Why not put the getting site attribute "longnamefirst" logic under line 404 and set the global variable $LONGNAME? Then there is not needed to add a special case for "$longnamefirst".
And, pls modify the code around line 398 and use "xCAT::TableUtils->get_site_attribute" to get site attribute "domain" and/or "longnamefirst".

perl-xCAT/xCAT/Schema.pm
```diff
@@ -1174,6 +1174,9 @@ site => {
    "              This setting does not apply to the service node. For sn\n".
    "              you need to set servicenode.ftpserver=1 if you want xcatd to\n".
    "              bring up vsftpd.\n\n".
+   "longnamefirst:   If set to 1/yes/enable, the /etc/hosts entries generated by 'makehosts' \n".
+   "                 will be have the long hostname put before the short hostname. If set to 0/no/disable \n".
+   "                 or doesn't set at all, the original behavior won't change for 'makehosts'.\n\n".
```
zet809: Pls remove the 0/no/disable to keep persistent with manpage.

whowutwut:
Can we add some text to describe the purpose of long name first in this pull request? 

"Longnamefirst" seems not to be the best keyword name...

If there is some description I can help think if anything is bether

xuweibj:
@whowutwut 
The background information I added to the comments. 
Looking forward to your help.

chenglch:
Maybe we can use Fully Qualified Domain Name (FQDN) instead of longname

whowutwut:
I agree with @chenglch , fqdn is probably a more standard naming for the long hostname. 

In my RHEL 7 machine, the default loopback is: 
```
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4 
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
```
have we looked into why there is localhost4 and localhost4.localdomain4 and 6?

xuweibj:
@chenglch , I have already modified the name as your comments.

chenglch:
For the code, LGTM

Only one question for @whowutwut about ``create_man_page.py``.
As we can use this script to generate rst from pod( or docstring in pm) , do we need to upload rst file in the pull request?  For this pull request, both ``site.5.rst`` and ``makehosts.8.rst`` are affected.

daniceexi:
@whowutwut I am a little tend to make the changer of pod file to handle the creating of rst file and commit it in github. The goodness for this way is the owner and reviewer can take a quick look of the output in RTD even we believe the pod2rst tool can handle the converting correctly.

perl-xCAT/xCAT/Schema.pm
```diff
@@ -1174,6 +1174,9 @@ site => {
    "              This setting does not apply to the service node. For sn\n".
    "              you need to set servicenode.ftpserver=1 if you want xcatd to\n".
    "              bring up vsftpd.\n\n".
+   "FQDNfirst: Fully Qualified Domain Name first. If set to 1/yes/enable, the /etc/hosts \n".
```
chenglch: I think ``create_man_page.py`` will generate a rst file called ``site.5.rst``. You can upload this file at the same time.

whowutwut: Yea it's a manual process to generate but running git status should show the .rst file changed by modifying the pod file.   If other .rst files appear it's probably bexause someone else missed running the tool. It's ok. I'm not sure what other way we can do this since converting all the pods is probably not a good option

chenglch:
LGTM for this change

PR #500, opened 2015-12-01
Set the version of the document to 2.11

whowutwut:
The title of the documentation pages being created page of the RTD puts the version at 0.0.1, update to 2.11 for the 2.11 branch

PR #502, opened 2015-12-01
Revert "Set the version of the document to 2.11"

whowutwut:
Reverts xcat2/xcat-core#500


daniceexi:
@whowutwut I don't know what's this conf.py for. Could you give some information?

We have made the decision that each xCAT release will have a corresponding doc version, so that means the doc for 2.11 will bind to the 2.11 branch, right? 

Current we still have some doc changes after creating the branch 2.11, an easy way is we can just change all stuff in master only now, then port back the whole docs directory from master to 2.11 branch just before the GA.

whowutwut:
Conf.py is the configuration file for the Sphinx documentation.  The version number, copyright, project name, etc are set in this file as well as any themes being used.  

The branch had already been made so how do you suggest we back port. Without a pull request or cherry pick all the commits into one single pull request before GA?

daniceexi:
@whowutwut easiest way is to just copy the whole dir docs from master to 2.11 and then add the docs dir to commit. Maybe there's a better way to do this through git command.

whowutwut:
Ok that might be good way

whowutwut:
@daniceexi  I tried this out today. I don't think it's a good solution because there's many places for error.   I can get a pretty good idea of what we need to do to sync up, but really it should not be one or two people to clean up. . the author of the change should create two pull requests.

It was our decision as a team to branch 2.11 before GA and @tingtli had brought up the warning that if we branch early and we need to put in a fixes, we need to cherry pick both branches.. and I think the agreement in that meeting was it was NOT a problem for dev to do that...  So why now are we trying to make exception at this time and make it easier...  ??   I have been doing 2 pull requests for the changes that went in and will continue to do so... 

I think we should stick to this standard practice of creating two pull request and have each author manage this so we can all understand this operating scenario...

daniceexi:
@whowutwut I totally agree that every code changer should cover the pull request for all necessary barnches. If it's not easy to sync the whole doc dirctory to GA branch, that's fine to require a pull request for each doc change for 2.11 branch.

PR #508, opened 2015-12-02
Remove test case get_xcat_install_loginfo

neo954:
Remove test case get_xcat_install_loginfo. This test case only run on RHEL ppc64 env. And this test case always failed, since the RHEL installer did not produce the old style log output any more.


tingtli:
Need to redesign the cases. Remove it right now.

whowutwut:
Is there a equivalent test case for RHEL 7? Is this to test that xCAT is installing packages correctly?

PR #512, opened 2015-12-03
Update the documentation landing page title for 2.11

whowutwut:
The currently landing page has a title of "Content"
Create a heading for xCAT and remove the index/module index/search 
at the bottom since those do not work.


bybai:
The new page is here: http://10.5.106.1/install/landing2.11/xcat-core/docs/build/html/
It looks better than before, I am OK to merge it.

PR #513, opened 2015-12-03
Add reboot again recommendation in IB doc for rh7

hu-weihua:
You can review it through http://10.2.2.30/advanced/networks/infiniband/mlnxofed_ib_install_v2_diskful.html.

### DIFF
--- a/docs/source/advanced/networks/infiniband/mlnxofed_ib_install_v2_diskful.rst
+++ b/docs/source/advanced/networks/infiniband/mlnxofed_ib_install_v2_diskful.rst
@@ -55,12 +55,12 @@ Configuration for Diskful Installation
 	rsetboot <node> net
 	rpower <node> reset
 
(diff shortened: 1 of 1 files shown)


daniceexi:

daniceexi:
@hu-weihua Please make another pull request for 2.11 branch.

whowutwut:
This really should only go into 2.11 branch for now.. But is ok. We need to see how to solve this in 2.11.1 without reboot as the step..

daniceexi:
@hu-weihua @neo954 I think you have figure out a way that run OFED install in postscript phase so don't need additional reboot. What's the progress?

hu-weihua:
Up to now, after some verification, @neo954 and I think postscript solution can work in redhat and ubuntu. but still has problem in sles. we are keep trying for this problem.  If we confirm postscript solution can't work in sles finally, maybe we can modify document, tell customer they have two options in redhat and ubuntu, but for sles, we recommend reboot again. this is the only one option.

PR #514, opened 2015-12-03
add man page about updatenode -o option

penguhyang:
Since there is not man page about the -o option, we need to add it.


whowutwut:
@penguhyang  why is this one closed?  We need to run the "create_man_page.py" script to generate the associated .rst file for the pod file

penguhyang:
@whowutwut We discussed this option in china team, we thought this option will be used for internal use.

daniceexi:
@penguhyang Please give a strong justificaiton when close close a issue without fix.

PR #517, opened 2015-12-04
Add attention message in doc for rflash

chenglch:
The rflash command has problems working well, so adding information text to the user for adding delays before issuing the command.

### DIFF
--- a/docs/source/guides/admin-guides/manage_clusters/ppc64le/management.rst
+++ b/docs/source/guides/admin-guides/manage_clusters/ppc64le/management.rst
@@ -100,15 +100,24 @@ As an example, get only the temperature information of a particular machine. ::
 Firmware Updating
 `````````````````
 
-For OpenPower machine, the firmware updating feature is implement in ``rflash``
-command.
+For OpenPOWER machines, use the ``rflash`` command to update firmware.
 
-Collect firmware version of the node and the HPM file:  ::
+Check firmware version of the node and the HPM file:  ::
 
     rflash cn1 -c /firmware/8335_810.1543.20151021b_update.hpm
 
-Update node firmware to the version of the HPM file:  ::
+Update node firmware to the version of the HPM file
 
+**ATTENTION**
+
+* Currently rflash is not stable enough to select accurate time before
+  upgrading. A ``rflash_delay`` attribute can be set in site table to adjust the
+  wait time before upgrading. We suggest setting this value to 70 which means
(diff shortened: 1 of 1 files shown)


docs/source/guides/admin-guides/manage_clusters/ppc64le/management.rst
```diff
@@ -100,15 +100,26 @@ As an example, get only the temperature information of a particular machine. ::
 Firmware Updating
 `````````````````
 
-For OpenPower machine, the firmware updating feature is implement in ``rflash``
+For OpenPOWER machine, the firmware updating feature is implement in ``rflash``
```
whowutwut: Suggest:  

For OpenPOWER machine**s**, use the ``rflash`` command to update firmware

docs/source/guides/admin-guides/manage_clusters/ppc64le/management.rst
```diff
@@ -100,15 +100,26 @@ As an example, get only the temperature information of a particular machine. ::
...
 command.
 
 Collect firmware version of the node and the HPM file:  ::
```
whowutwut: This should be "check" instead of "collect"?

docs/source/guides/admin-guides/manage_clusters/ppc64le/management.rst
```diff
@@ -100,15 +100,26 @@ As an example, get only the temperature information of a particular machine. ::
...
 
-Update node firmware to the version of the HPM file:  ::
 
+
+Update node firmware to the version of the HPM file
+
+**ATTENTION**
+
+* Currently rflash is not stable enough to select accurate time before
+  upgrading. A 'rflash_delay' attribute can be set in site table to adjust the
```
whowutwut: suggest ``rflash_delay`` format in code with the backticks.

whowutwut:
@chenglch  I wonder if we can actually print out the ipmi commands as we are doing the rflash as debug, so the user can see the raw commands being issued to the machine? or if we have a -V verbose mode for rflash.. maybe something to think about for 2.12?

chenglch:
@whowutwut , I have update this commit following your comments. 

For the verbose option, it is useful and we can simply add ``-v`` option when running ipmitool command,but the output is not just raw commands.

whowutwut:
I think this one needs to cherry pick to 2.11 branch

PR #528, opened 2015-12-11
Merge the pages for Large Cluster and Hierarchical Clusters

whowutwut:
Realized that the large cluster support and the advanced hierarchical pages are the same.
This initial commit does some refactoring of the pages.

Left around the define_service_node.rst page but it's not in the index.
We need to verify that the content is correct in define_service_nodes.rst


whowutwut:
http://10.2.5.2/hierarchical/  if anyone wants to view a preview.   

@cxhong  I'm done with the changes for this.  

@mfoliveira  Thanks for bringing this up.  What I've done for the short term is to move the Large Cluster to the Hierarchical Section under Advanced Topics since those sections are the same. I was just trying to re-write it into the "Large Cluster" section.   Maybe will be enhanced in 2.12 docs "latest" version

PR #531, opened 2015-12-14
Get_install_disk,Task 75933

xuweibj:
For RedHat/Ubuntu/Sles/PowerKVM, the files for choosing install disk are different, but with similar logic. It is inconvenient for maintenance. And because of different methods for choosing disk and the disk name maybe different from last boot (physical machine), the OS may be installed on different disk. To avoid this situation, we want to use the same logic/file for all.
The idea is,  for RedHat/Ubuntu/Sles/PowerKVM we use the same file to get the install disk.
The file will get install disk, and then output it to the file "/tmp/install_disk".

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/getinstdisk
+++ b/xCAT-server/share/xcat/install/scripts/getinstdisk
@@ -1,0 +1,155 @@
+#!/bin/bash
+
+#-----------------------------------------------------------
+#
+# Get proper disk to install OS
+#
+# 1. Check all partitions list in /proc/partitions, whether 
+#    there is disk had OS installed. Select the one with the
+#    smallest WWN.
+# 2. Check all disks list in /proc/partitions, sort them by
+#    WWN and driver type, select the first one.
+# 3. Select the default one: /dev/sda.   
+#
+# Output: /tmp/install_disk
+#
+#-----------------------------------------------------------
+
+install_disk=""
+rm /tmp/install_disk
+
+# Check if any disk have installed OS
+if [ -z "$install_disk" ]; then
+
+    entries=$(awk -F ' '  '{print $4}' /proc/partitions | grep -v "name")
+    
+    for entry in $entries; do
+
+        dev_type=$(udevadm info --query=property --name=/dev/$entry | grep -i "DEVTYPE" | awk -F = '{print $2}' | tr A-Z a-z)
+        
+        if [ "$dev_type" == "disk" ]; then
+            disks=$disks"$entry"
+        elif [ "$dev_type" == "partition" ]; then
+            partitions=$partitions"$entry"
+        fi
+
+    done
+
+    for partition in $partitions; do
+
+        echo Check the partition $partition >> /tmp/getinstalldisk_log
+
+        if [ -e "/tmp/tmp_${partition%%[0-9]*}" ]; then
+            echo The disk had OS installed, check next partition. >> /tmp/getinstalldisk_log
+            continue
+        fi
+
+        mkdir -p /tmp/mymount;
+
+        fs_type=$(udevadm info --query=property --name=/dev/$partition | grep -i "FS_TYPE" | awk -F = '{print $2}')
+
+        rc=255
+
+        # mount partition based on fs type, if fs_type is swap, do not mount it, jump to next partition.
+        if [ -z "$fs_type" ]; then
+            mount /dev/$partition /tmp/mymount 
+            rc=$?
+        elif [ "$fs_type" != "swap" ]; then
+            mount -t $fs_type /dev/$partition /tmp/mymount 
+            rc=$?
+        fi
+
+        if [ $rc -eq 0 ]; then
+
+            ker_dir="/tmp/mymount"            
+            
+            if [ -d /tmp/mymount/boot ]; then
+                ker_dir="/tmp/mymount/boot"
+            fi
+
+            for i in $ker_dir/vmlinuz*; do
+                disk_part=${partition%%[0-9]*}
+                disk_wwn=$(udevadm info --query=property --name=$disk_part | grep '\<ID_WWN\>' | cut -d "=" -f2 | tr A-Z a-z)
+                disk_array=$disk_array"$disk_part $disk_wwn\n"
+                touch /tmp/tmp_$disk_part
+                break
+            done
+             
+            echo -e The partition $partition has kernel file. >> /tmp/getinstalldisk_log
+
+            umount /tmp/mymount || echo $partition umount failed. >> /tmp/getinstalldisk_log
+        else
+            echo Partition $partition mount failed or the partition is swap. >> /tmp/getinstalldisk_log
+        fi
+    done
+    
+    if [ "$disk_array" ]; then    
+        echo -e The disk_array: >> /tmp/getinstalldisk_log
+        echo -e $disk_array >> /tmp/getinstalldisk_log
+
+        install_disk=/dev/$(echo -e $disk_array | grep -v "^$" | sort -t : -k 2 -b | cut -d " " -f1 | head -n 1)
+   
+        echo -e The install_disk is $install_disk. >> /tmp/getinstalldisk_log
+    fi
+    rm -f /tmp/tmp_*;
+    rmdir /tmp/mymount;
+fi
+
+# Sort all disks based on their WWN and driver type, choose the first one
+if [ -z "$install_disk" ]; then
+    
+    rm -f /tmp/tmp_*;
+
+    for disk in $disks; do
+
+        disk_wwn=$(udevadm info --query=property --name=$disk | grep '\<ID_WWN\>' | cut -d "=" -f2 | tr A-Z a-z)
+ 
+        if [ "$disk_wwn" ]; then
+
+            disk_driver=$(udevadm info --attribute-walk --name=$disk | grep DRIVERS| grep -v '""'| grep -v '"sd"'| head -n 1| sed -e 's/[^"]*"//' -e 's/"//' | tr A-Z a-z)
+
+            echo -e The disk $disk information: disk_wwn=$disk_wwn disk_driver=$disk_driver. >> /tmp/getinstalldisk_log
+
+            case "$disk_driver" in
+
+            "ata_piix4"|"PMC MaxRAID"|"ahci"|"megaraid_sas")
+                echo "$disk $disk_wwn" >> /tmp/tmp_firstdisks
+                echo "Add disk:" $disk $driver $disk_wwn " into firstdisks" >> tmp/tmp/getinstalldisk_log
+                ;;
+            "mptsas"|"mpt2sas"|"mpt3sas")
+                echo "$disk $disk_wwn" >> /tmp/tmp_seconddisks
+                echo "Add disk:" $disk $driver $disk_wwn " into seconddisks" >> tmp/tmp/getinstalldisk_log
+                ;;
+            *)
+                echo "$disk $disk_wwn" >> /tmp/tmp_thirddisks
+                echo "Add disk:" $disk $driver $disk_wwn " into thirddisks" >> tmp/tmp/getinstalldisk_log
+                ;;
+ 
+            esac
+        fi
+    done
+
+    for seq in first second third; do
+        if [ -s /tmp/tmp_${seq}disks ]; then
+            install_file="/tmp/tmp_${seq}disks"
+            break
+        fi
+    done
+        
+
+    if [ "$install_file" ] && [ -s $install_file ]; then
+        install_disk=/dev/$(cat $install_file | grep -v "^$" | sort -t : -k 2 -b | cut -d " " -f1 | head -n 1)
+        echo -e The install_disk is $install_disk by sorting WWN and DRIVER. >> /tmp/getinstalldisk_log
+        rm -f /tmp/tmp_*;
+    fi
+      
+fi
+
+# Cannot find proper disk for OS install, select the default one "/dev/sda"
+if [ -z "$install_disk" ]; then
+    install_disk="/dev/sda"
+    echo -e The default install_disk is $install_disk. >> /tmp/getinstalldisk_log
+fi
+
+# Output the result to /tmp/install_disk file
+echo $install_disk > /tmp/install_disk


xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,111 @@
...
+
+#-----------------------------------------------------------
+#
+# Get proper disk to install OS
+#   
+# Output: /tmp/install_disk
+#
+#-----------------------------------------------------------
+
+# Check if any disk have installed OS
```
zet809: My idea is: if the variable install_disk have value, we need to clear it. Since we'd like to use this script to get the install_disk.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,111 @@
...
+if [ -z "$install_disk" ]; then
+
```
zet809: You may need to add some debug info or log here.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,111 @@
...
+#-----------------------------------------------------------
+
+# Check if any disk have installed OS
+if [ -z "$install_disk" ]; then
+    
+    partitions=$(awk -F ' '  '{print $4}' /proc/partitions | grep -e "[s|v|h]d[a-z][0-9]\+$")    
+    disks=$(awk -F ' '  '{print $4}' /proc/partitions | grep -e "[s|v|h]d[a-z]$")
+
+    for partition in $partitions; do
+
```
zet809: Cut the first 3 character will cause a potential issue for the disk name like sdaa...
Move the numbers our are enough.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,111 @@
...
+        fs_type=$(udevadm info --query=property --name=/dev/$partition | grep -i "FS_TYPE" | awk -F = '{print $2}')
+
+        rc=255
+        if [ -z "$fs_type" ]; then
+            mount /dev/$partition /tmp/mymount 
+            rc=$?
+        elif [ "$fs_type" != "swap" ]; then
+            mount -t $fs_type /dev/$partition /tmp/mymount 
+            rc=$?
+        fi
```
zet809: If the fs_type is "swap", we can go to next partition directly, right?

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,111 @@
...
+        if [ $rc -eq 0 ]; then
+
+            ker_dir="/tmp/mymount"            
+            
+            if [ -d /tmp/mymount/boot ]; then
+                ker_dir="/tmp/mymount/boot"
+            fi
+
+            for i in $ker_dir/vmlinuz*; do
+                disk_part=${partition:0:3}
```
zet809: Same issue with line 19.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,111 @@
...
+
+    if [ -s /tmp/tmp_firstdisks ]; then
+        install_file="/tmp/tmp_firstdisks"
+    elif [ -s /tmp/tmp_seconddisks ]; then
+        install_file="/tmp/tmp_seconddisks"
+    elif [ -s /tmp/tmp_thirddisks ]; then
+        install_file="/tmp/tmp_thirddisks"
+    fi
+
+    if [ "$install_file" ]; then
```
zet809: I think checking whether the $install_file exist will be better.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,126 @@
...
+install_disk=""
+# Check if any disk have installed OS
+if [ -z "$install_disk" ]; then
+    
+    partitions=$(awk -F ' '  '{print $4}' /proc/partitions | grep -e "[s|v|h]d[a-z][0-9]\+$")    
+    disks=$(awk -F ' '  '{print $4}' /proc/partitions | grep -e "[s|v|h]d[a-z]$")
+
+    for partition in $partitions; do
+
+        if [ -e "/tmp/tmp_$(echo $partition | tr -d '[0-9]')" ]; then
```
neo954: You can remove the number suffix by using ``${partition%%[0-9]*}``. It is something like the following line. In this way, you will save two ``fork()``s.

``if [ -e "/tmp/tmp_${partition%%[0-9]*}" ]; then``

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,126 @@
...
+
+            for i in $ker_dir/vmlinuz*; do
+                disk_part=$(echo $partition | tr -d '[0-9]')
+                disk_wwn=$(udevadm info --query=property --name=$disk_part | grep '\<ID_WWN\>' | cut -d "=" -f2 | tr A-Z a-z)
+                disk_array=$disk_array"$disk_part $disk_wwn\n"
+                touch /tmp/tmp_$disk_part
+                break
+            done
+             
+            echo -e The partition $partition has kernel file. >> /tmp/getinstalldisk_log
```
neo954: The ``-e`` argument of ``echo`` is no use here.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,126 @@
...
+
+            umount /tmp/mymount;
```
neo954: Check the ``$?`` value here. What if the ``umount`` command failed? If the ``umount`` failed here, and no error handling, when we go to line 70, everything in this partition will be deleted.

I believe this is improper.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,126 @@
...
+        echo -e The disk $disk information: disk_wwn=$disk_wwn disk_driver=$disk_driver. >> /tmp/getinstalldisk_log
+    done
+
+    if [ -s /tmp/tmp_firstdisks ]; then
+        install_file="/tmp/tmp_firstdisks"
+    elif [ -s /tmp/tmp_seconddisks ]; then
+        install_file="/tmp/tmp_seconddisks"
+    elif [ -s /tmp/tmp_thirddisks ]; then
+        install_file="/tmp/tmp_thirddisks"
+    fi
```
neo954: You can change line 103 to 109 to the following codes, in this way, repeated lines can be avoid.

    for seq in first second third
    do
        if [ -s "/tmp/tmp_${seq}disks" ]; then
            install_file="/tmp/tmp_${seq}disks"
            break
        fi
    done

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,126 @@
...
+
+    if [ "$install_file" ] && [ -s $install_file ]; then
```
neo954: ``test -s`` is already checked through line 103 to 109, I think there is no reason to check it again here.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,126 @@
...
+# Check if any disk have installed OS
+if [ -z "$install_disk" ]; then
+    
+    partitions=$(awk -F ' '  '{print $4}' /proc/partitions | grep -e "[s|v|h]d[a-z][0-9]\+$")    
+    disks=$(awk -F ' '  '{print $4}' /proc/partitions | grep -e "[s|v|h]d[a-z]$")
+
+    for partition in $partitions; do
+
+        if [ -e "/tmp/tmp_$(echo $partition | tr -d '[0-9]')" ]; then
+            break
```
neo954: I believe ``break`` is improper here. The word ``continue`` should be used.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,126 @@
...
+    if [ "$disk_array" ]; then    
+        echo -e The disk_array: >> /tmp/getinstalldisk_log
+        echo -e $disk_array >> /tmp/getinstalldisk_log
+
+        install_disk=/dev/$(echo -e $disk_array | grep -v "^$" | sort -t : -k 2 -b | cut -d " " -f1 | head -n 1)
+   
+        echo -e The install_disk is $install_disk. >> /tmp/getinstalldisk_log
+    fi
+    rm -f /tmp/tmp_*;
+    rm -rf /tmp/mymount;
```
neo954: At least, run ``rmdir /tmp/mymount`` to make sure the directory to be removed is an empty directory.

xCAT-server/share/xcat/install/scripts/getinstdisk
neo954: Or more eval, something like ...

    for install_file in /tmp/tmp_{first,second,third}disks
    do
        [ -s "$install_file" ] && break
    done

In this way, you can ``test -s`` in line 111 again.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,126 @@
...
+                echo $disk $disk_wwn >> /tmp/tmp_thirddisks
+                ;;
+ 
+            esac
+        fi
+        echo -e The disk $disk information: disk_wwn=$disk_wwn disk_driver=$disk_driver. >> /tmp/getinstalldisk_log
+    done
+
+    for seq in first second third; do
+        if [-s /tmp/${seq}disks]; then
```
neo954: Between ``[`` and ``-s``, there should be a space.

xuweibj:
@neo954 ， I have modified some, please help review again.
Thank you!

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,126 @@
...
+            "mptsas"|"mpt2sas"|"mpt3sas")
+                echo $disk $disk_wwn >> /tmp/tmp_seconddisks
+                ;;
+            *)
+                echo $disk $disk_wwn >> /tmp/tmp_thirddisks
+                ;;
+ 
+            esac
+        fi
+        echo -e The disk $disk information: disk_wwn=$disk_wwn disk_driver=$disk_driver. >> /tmp/getinstalldisk_log
```
caomengmeng: How about adding something to point it clearly where this disk information is wrote into, the first, second or third file?
Then without checking the /tmp/tmp_files, then we can deduce which is the install_disk by reading this /tmp/getinstalldisk_log.
Other question, can I execute some command to check if the disk information are right or not after installing?

xCAT-server/share/xcat/install/scripts/getinstdisk
caomengmeng: Can we add the WWN of the disk into the /tmp/geinstalldisk_log also? This can be used to verify if the collected information is right or not.

xuweibj: Hi @caomengmeng , the WWN has been added into disk_array, and will be printed into log in line 63.

caomengmeng: Yes @xuweibj, I saw this disk_array too. But except this result disk_array, there isn't any log to track if the partition and WWN are corresponding or not, then to verify the disk and WWN are correctly corresponding.
Not sure if it is necessary.

xCAT-server/share/xcat/install/scripts/getinstdisk
xuweibj: Hi @caomengmeng , I modified the code, is this OK?

xuweibj:
Hi @daniceexi ,  I modified, please help review again.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,144 @@
...
+#-----------------------------------------------------------
+
+install_disk=""
+rm /tmp/install_disk
+
+# Check if any disk have installed OS
+if [ -z "$install_disk" ]; then
+    
+    partitions=$(awk -F ' '  '{print $4}' /proc/partitions | grep -e "[s|v|h]d[a-z][0-9]\+$")    
+    disks=$(awk -F ' '  '{print $4}' /proc/partitions | grep -e "[s|v|h]d[a-z]$")
```
immarvin: I suggest to traverse all the entries in ``/proc/partitions`` with ``udevadm`` , take the entries with "DEVTYPE=disk" to be disks and take the ones with "DEVTYPE=partition" to be partitions. The logic  to differentiate the devices based on the device names are not reliable enough. We can also get the "ID_WWN" and "DRIVERs" attributes for the disks in this iteration.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,144 @@
...
+        fs_type=$(udevadm info --query=property --name=/dev/$partition | grep -i "FS_TYPE" | awk -F = '{print $2}')
+
+        rc=255
+
+        # mount partition based on fs type, if fs_type is swap, do not mount it, jump to next partition.
+        if [ -z "$fs_type" ]; then
+            mount /dev/$partition /tmp/mymount 
+            rc=$?
+        elif [ "$fs_type" != "swap" ]; then
+            mount -t $fs_type /dev/$partition /tmp/mymount
```
immarvin: I think we do not need to try to mount ``swap``

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,144 @@
...
+        if [ "$disk_wwn" ]; then
+
+            disk_driver=$(udevadm info --attribute-walk --name=$disk | grep DRIVERS| grep -v '""'| grep -v '"sd"'| head -n 1| sed -e 's/[^"]*"//' -e 's/"//')
+
+            echo -e The disk $disk information: disk_wwn=$disk_wwn disk_driver=$disk_driver. >> /tmp/getinstalldisk_log
+
+            case "$disk_driver" in
+
+            "ata_piix4"|"PMC MaxRAID"|"ahci"|"megaraid_sas")
+                echo $disk $disk_wwn >> /tmp/tmp_firstdisks
```
immarvin: since these files are format sensitive, I suggest to quote the string after``echo`` with ""

xuweibj:
Hi @immarvin , I modified, please help review again.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,155 @@
...
+#
+#-----------------------------------------------------------
+
+install_disk=""
+rm /tmp/install_disk
+
+# Check if any disk have installed OS
+if [ -z "$install_disk" ]; then
+
+    entries=$(awk -F ' '  '{print $4}' /proc/partitions)
```
immarvin: The header of ``/proc/partitions`` should be filtered out
[root@c910f03c05k21 ~]# entries=$(awk -F ' '  '{print $4}' /proc/partitions)
[root@c910f03c05k21 ~]# echo $entries
name sr0 sda sda1 sda2 sda3 sda4 sda5 dm-0 loop0 loop1 loop2 loop3

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -0,0 +1,155 @@
...
+    
+    for entry in $entries; do
+
+        dev_type=$(udevadm info --query=property --name=/dev/$entry | grep -i "DEVTYPE" | awk -F = '{print $2}')
+        
+        if [ $dev_type == "disk" ]; then
```
immarvin: for comparing of string, the arguments in both side of the operator should be double quoted, since bash will raise a error if either of them is empty.

daniceexi:
@immarvin , @zet809 , @neo954 , @caomengmeng have reviewed the code, it'd good enough to merge. Thanks all reviewer.

daniceexi:
@xuweibj and @zet809 
I did not see the description of how to use /tmp/getinstalldisk_log. Does user have to ssh to the installer to check this log file? Could we copy this log file to the installed system when running the post script?

immarvin:
hi @daniceexi , I suggest to add the description of the partition log file into https://sourceforge.net/p/xcat/wiki/Debugging_xCAT_Problems/#collect-information-and-logs-from-the-installer-during-os-installation, btw, this doc has not been moved to RTD yet.

It's a good idea to copy this file into the installed file system. But there is a technical problem, since the copy action should be taken after the installation is completed and before ``chrooted`` to the installed file system and before 1st reboot, for sles, the script can be added into the "chroot.sles", but for redhat, this requires the post script be run with "--nochroot", this option is turned off by default

immarvin:
for redhat, there is an option for Pre-installation Script,
````
--log=
    Logs the script's output into the specified log file. For example:

    %pre --log=/root/ks-pre.log
````

Issue #516, opened 2015-12-04
updatenode -k  does not check if the passwd is correct, only check if the passwd is input or not.


wwzhubj:
1. After I installed a node, ssh the node, below message always showed up:
[root@c712ems2 setupnic]# ssh c712f5n12
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
IT IS POSSIBLE THAT SOMEONE IS DOING SOMETHING NASTY!
Someone could be eavesdropping on you right now (man-in-the-middle attack)!
It is also possible that a host key has just been changed.
The fingerprint for the ECDSA key sent by the remote host is
e7:ac:1a:26:bf:77:40:93:35:18:80:a4:57:84:c8:45.
Please contact your system administrator.
Add correct host key in /root/.ssh/known_hosts to get rid of this message.
Offending ECDSA key in /root/.ssh/known_hosts:40
Password authentication is disabled to avoid man-in-the-middle attacks.
Keyboard-interactive authentication is disabled to avoid man-in-the-middle attacks.
2. Then I ran ' updatenode c712f5n12 -k'
[root@c712ems2 setupnic]# updatenode c712f5n12 -k
Enter the password for the userid: root on the node where the ssh keys
will be updated:
##(here I entered a wrong passwd)##

c712f5n12: Setup ssh keys has completed.
c712f5n12: xcatdsklspost: downloaded postscripts successfully
c712f5n12: Fri Dec  4 01:07:59 EST 2015 Running postscript: remoteshell
c712f5n12:
c712f5n12: root      25900  0.5  0.0  27136 17792 ?        Ss   01:07   0:00 sshd: root@notty
c712f5n12: root      26112  0.0  0.0  17920 11712 ?        Ss   01:08   0:00 /usr/sbin/sshd -D
c712f5n12: Postscript: remoteshell exited with code 0
c712f5n12: Running of postscripts has completed.
3.  then ssh c712f5n12 again, upper messages in step 1 are still there, no any change. 
What I expect is the messages in step 1 are disappeared after I ran `updatenode -k`.

*So there are 2 problems in this case:
[...]

PR #509, opened 2015-12-02
Remove white space at the end of line for spawn strings.

cxhong:
IBM Cloud Analytics team report a "xdsh -K" issue on their Softlayer system.  The compute node was provisioned by different Manager Node.  The nodes are in the different datacenter and hooked up via tunnels. 
````
# xdsh uklon02hyp00014ianra -K
Enter the password for the userid: root on the node where the ssh keys 
will be updated:

Error: copykeys:uklon02hyp00014ianra has error,1:TIMEOUT
Error: copy.sh:uklon02hyp00014ianra has error,3:Child PID 14494 exited with status 32512
Error: remoteshellexp failed sending keys to enablenodes.
Error: SSH setup failed for the following nodes: uklon02hyp00014ianra.
return code = 1
 
but 
# ssh uklon02hyp00014ianra
root@uklon02hyp00014ianra's password: 
````

Issue #280, opened 2015-10-20
[Customer]`nodeset $node osimage=$image` should clean out keys for $node from `~root/.ssh/known_hosts`


samveen:
Re-imaging a node always changes the ssh host key of the node, and will always trigger the `ssh` `man-in-the-middle` warning. A better solution would be to `remove the offending keys` from the `known_hosts` file.

daniceexi:
@samveen The issue has been fixed by pull request #542. Please confirm.

samveen:
@daniceexi Looks like the fix should work. Closing ticket.

whowutwut:
Using the latest build, we still see this issue in 2.12.1.... 
```
[root@fs1 ~]# ssh c910f05c31
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
IT IS POSSIBLE THAT SOMEONE IS DOING SOMETHING NASTY!
Someone could be eavesdropping on you right now (man-in-the-middle attack)!
It is also possible that a host key has just been changed.
The fingerprint for the ECDSA key sent by the remote host is
46:5d:dd:3f:58:b5:fc:7f:53:4e:6a:73:ed:62:8f:5b.
Please contact your system administrator.
Add correct host key in /root/.ssh/known_hosts to get rid of this message.
Offending ECDSA key in /root/.ssh/known_hosts:22
Password authentication is disabled to avoid man-in-the-middle attacks.
Keyboard-interactive authentication is disabled to avoid man-in-the-middle attacks.
[root@c910f05c31 ~]# ```

Going to reopen....

gurevichmark:
I am wondering is the same should be done for issue #450. If user issues `nodeset <name> offline` should we remove the entry from **known_hosts** ?

whowutwut:
Yes, good point. I think we should.  If offline means we need to re-provision when bringing back online, the host key will change.

daniceexi:
``offline`` means deconfigure the provision configuration. But the node is still in power on status. Do we want this impacts the xdsh command?

whowutwut:
[...]

PR #542, opened 2015-12-17
Remove hostname from known_hosts file after nodes reset

cxhong:
Issue #516 , this fix will remove hostname from known_hosts file on the MN to avoid "man-in-the-middle attack" message after nodes reset.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/credentials.pm
+++ b/xCAT-server/lib/xcat/plugins/credentials.pm
@@ -320,9 +320,6 @@ sub process_request
            @filecontent=();
        }
     }
-     `logger -t xcat -p local4.info "credentials: remove $client from known_hosts"` ;
-    system("ssh-keygen -R $client -f /root/.ssh/known_hosts");
-
     if (defined $rsp->{data}->[0]) {
 	#if we got the data from the file, send the data message to the client
         xCAT::MsgUtils->message("D", $rsp, $callback, 0);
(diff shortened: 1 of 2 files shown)


cxhong:
Pull request #509 will take care "updatenode -k"  problem seeing on Issue #516

daniceexi:
@cxhong I understand your change could fix this issue through removing the original public host key for the target node, and when reconnect to the node with ssh a new host key will be created in know_hosts by default. But it depens on our current configuration that xcat mn just accept the public host key without checkout the authority.

I think a better fix is to run the makeknownhosts in the credentials.pm to regenerate the correct entry in the 'known_hosts' file. **BUT** since this rewrite of known_hosts happens for each node when install a banch of nodes, there may be conflict for multiple processes to wirte konwn_hosts file. At least the performance will be impacted.

Then whether require user to run 'makeknowhosts' manually is an acceptable solution?

cxhong:
Actually updatenode -k is another options to modify  known_host file after node provisioned.  pull request #509 had fixes for "updatenode -k".  User has to manually run it.

daniceexi:
@cxhong 'updatenode -k' does work because it calls the 'makeknownhosts' internally. 
I recommend we don't fix this issue, but tell user to run 'makeknownhosts' command to fix it.

cxhong:
Actually, the root cause for this issue was the file '/etc/ssh/ssh_host_ecdsa_key' was NOT existed when running remoteshell postscripts.  we will use ssh-keygen command to figure out the compute node is supporting 'ecdsa' instead of by checking the existing of file '/etc/ssh/ssh_host_ecdsa_key'.

daniceexi:
I have reviewed this change and agree to merge. @cxhong Thanks for your fix.

whowutwut:
This pull request attempted to address issue #280 but looks like it does not.  Re-opened that issue

PR #543, opened 2015-12-18
Add getadapter in genesis build spec

chenglch:
Add getadapter in genesis build spec to avoid of build error


daniceexi:
The change is OK.

Issue #470, opened 2015-11-25
[FVT]Need to support detach operation for PowerKVM PCI pass-through


neo954:
xCAT 2.11 support PCI pass-through feature for PowerKVM. With this feature, now, the end user can attach PCI adapter for PowerKVM guest. But for xCAT support, there is no detach operation. While virsh and PowerKVM do support the detach operation.

I think xCAT need to support detach operation for PowerKVM PCI pass-through.

immarvin:
https://github.com/xcat2/xcat-core/pull/524

neo954:
This bug is verified. I will close this one.

PR #548, opened 2015-12-21
 fix issue Need to support detach operation for PowerKVM PCI pass-through #470 #524 

immarvin:
add support for device detaching from kvm guest,
add a new option --devdetach for chvm


whowutwut:
@immarvin, I see there's a .pod file change. Can you run the create man pages script and include the .rst file changes in the pull request too?

daniceexi:
I have reviewed the changes and agree to merge.

PR #551, opened 2015-12-22
Fix GitHub issue 505. Run hack_uname() while mlnxofed_ib_install.v2 is run in install stage

neo954:
Fix GitHub issue 505. Run hack_uname() while mlnxofed_ib_install.v2 is run in install stage.

### DIFF
--- a/xCAT-server/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2
+++ b/xCAT-server/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2
@@ -172,6 +172,11 @@ function cleanup()
     local -i max_retry=99
 	
     if [ "$NODESETSTATE" != "genimage" ]; then
+        # Clean up the uname hacking
+        if [ -f "/bin/uname.save" ]; then
+            mv -f "/bin/uname.save" "/bin/uname"
+        fi
+
         if mount | grep -q "/tmp/ofed/mountpoint"; then
             while ! umount "/tmp/ofed/mountpoint"
             do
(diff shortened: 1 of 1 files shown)


hu-weihua:
I have reviewed the changing code, I agree to merge.

xCAT-server/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2
```diff
@@ -172,6 +172,11 @@ function cleanup()
     local -i max_retry=99
 	
     if [ "$NODESETSTATE" != "genimage" ]; then
+        # Clean up the uname hacking
+        if [ -f "/bin/uname.save" ]; then
```
whowutwut: I always like to use variable here.. should we have called the file ``/bin/uname.xcatsave`` to be more unique? 
Also, above in hack_uname(), we reference the uname path with $1...  ``mv $1/bin/uname``.  Are we ever calling hack_uname with an argument?  The cleanup function doesn't take that into account.. what if the file is $1/bin/uname?

neo954: @whowutwut, there is no chance to pass argument to function ``cleanup()``. Yes, we do pass argument to ``hack_uname()``. And the current implement of ``cleanup()`` count that in. It will not clean up the wrong thing.

hu-weihua:
@whowutwut ,  thanks for reviewing again.  ``mlnxofed_ib_install`` is used in 2 scenario. One is running in CN during diskfull installation, another is running in MN during ``genimage``.  When running in MN during ``genimage``,  ``hack_uname`` needs to know the  the root path of image, so it need a attribute ``$1`` to indicate the root path of image.  But when ``mlnxofed_ib_install`` runs in CN during installation,  ``chroot`` has ready done. We call ``hack_uname`` without attribute( i.e. ``$1=null``). so this time ``uname`` file is ``/bin/uname``.

We can know depending on line 174,  the changing code belongs to first scenario.  So the code could work correctly.

Issue #559, opened 2015-12-28
[Customer] bmcdiscover command cannot get the MTMS attributes for Habanero Server


fenghaizhu:
**Environment:** 
RHEL7.2 (GA)
Habanero server

**Issue:** MTMS_based_discovery didn't work for Habanero .  
The MTMs/serials are not discovered as bellow.
```
[root@smn~]# bmcdiscover -s nmap --range 50.2.0.65-80 -u ADMIN -p admin  -z
node-32020041:
	objtype=node
	groups=all
	bmc=50.2.0.65
	cons=ipmi
	mgt=ipmi
	bmcusername=ADMIN
	bmcpassword=admin

node-32020042:
	objtype=node
	groups=all
	bmc=50.2.0.66
	cons=ipmi
	mgt=ipmi
	bmcusername=ADMIN
	bmcpassword=admin

node-32020044:
	objtype=node
	groups=all
	bmc=50.2.0.68
	cons=ipmi
	mgt=ipmi
	bmcusername=ADMIN
	bmcpassword=admin
```

whowutwut:
@xuweibj  Can you please add some response to this issue as to what cause the problem and how did we fix it? There is no information provided in the pull reqeust, the only way to know is to look at the code.

daniceexi:
@whowutwut I changed the subject to a reasonable text. And set the lable: ``status:pending`` since the code has been merged.

whowutwut:
Thanks @daniceexi ... What I was looking for was a solution to the Issue.  The customer opens an issue but we don't move the issue discussion through to completion.  In this case we just reference a pull request that claims to fix the issue.  In the pull request, we also do not have any text indicating what the fix was, that would have been OK also.   What is left is that the customer needs to go into the code and look at the code change to understand why this did not work in the first place.   

So I was looking for some simple response to explain the cause of the problem and the solution. Something like: 

*The problem is caused by Tyan OpenPower (Habanero) servers having the MTMS information in FRU2 and we are not checking that FRU.  Historically, SystemX and Tuletta Power machines have the MTMS is in FRU0, For OpenPower S822LC, it is in FRU3.  Will add a check in the code to look for MTMS information in FRU2 for Habanero*

daniceexi:
@whowutwut Thanks for the remind. Yes, the root cause was xCAT code could not find the MTMS for Habanero nodes from FRU0. The fix was to try to go through FRU0, 2, 3 until to find a MTMS information.

[...]

PR #570, opened 2016-01-05
Fix Issue 559, MTMS_based_discovery work for Habanero. 

xuweibj:
MTMS_based_discovery didn't work for Habanero. #559


whowutwut:
@cxhong  maybe you can take a look at this since we did some discovery work this week.

daniceexi:
@cxhong @whowutwut I have reviewed the code and talked comments with @xuweibj direclty. The code change is OK, I will merge the change.

Issue #564, opened 2016-01-04
Mistake for creating install disk logs


caomengmeng:
ENV:
    OS and platform: Lpar PPC64, RH 7.2
xCAT-core and dep:
    20151231.1930-core-rpms-snap.tar.bz2
    xcat-dep-201511181550.tar.bz2
How to reproduce:
    Install the CN then using debugmode to track logs.

Currente: Failed to write some logs for the misktake: tmp/tmp/getinstalldisk_log, which should be /tmp/getinstalldisk_log

Excepted: All logs should be wrote into  /tmp/getinstalldisk_log

xuweibj:
Pull Request #565

tingtli:
@caomengmeng 
when you open a defect, please write down which xcat version you use, which platform you test. How could developer reproduce your defect etc.

whowutwut:
@xuweibj If your pull request fixing this issue is merge, please set to pending state to @caomengmeng  knows that it's ready for verification

caomengmeng:
passed

PR #571, opened 2016-01-05
Fix issue 564 and 579, creating right install disk logs

xuweibj:
I am so sorry, I forgot the "/" before tmp.
Refer to issue #564

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/getinstdisk
+++ b/xCAT-server/share/xcat/install/scripts/getinstdisk
@@ -17,6 +17,10 @@
 
 install_disk=""
 rm /tmp/install_disk
+tmpdir="/tmp/xcat.getinstalldisk"
+mkdir -p $tmpdir
+logfile="/tmp/xcat_getinstalldisk_log"
+tmpfile=$tmpdir"/getinstalldisk_"
 
 # Check if any disk have installed OS
 if [ -z "$install_disk" ]; then
@@ -35,16 +39,17 @@ if [ -z "$install_disk" ]; then
 
     done
 
+    mount_dir=$tmpdir"/xcat.getinstalldisk.mount"
+    mkdir -p $mount_dir;
+
     for partition in $partitions; do
 
-        echo Check the partition $partition >> /tmp/getinstalldisk_log
+        echo "Check the partition $partition." >> $logfile
 
-        if [ -e "/tmp/tmp_${partition%%[0-9]*}" ]; then
-            echo The disk had OS installed, check next partition. >> /tmp/getinstalldisk_log
+        if [ -e "$tmpfile${partition%%[0-9]*}" ]; then
+            echo "    The disk ${partition%%[0-9]*} had OS installed, check next partition." >> $logfile
             continue
         fi
-
-        mkdir -p /tmp/mymount;
 
         fs_type=$(udevadm info --query=property --name=/dev/$partition | grep -i "FS_TYPE" | awk -F = '{print $2}')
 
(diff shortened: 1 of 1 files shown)


xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -61,6 +61,8 @@ if [ -z "$install_disk" ]; then
 
         if [ $rc -eq 0 ]; then
 
+            echo Partition $partition mount success. >> /tmp/getinstalldisk_log
```
whowutwut: Is there a specific reason we don't use quotes around the text that will be echo out? 
```
echo "Partition $partition mount success." >> /tmp/getinstalldisk_log
```

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -61,6 +61,8 @@ if [ -z "$install_disk" ]; then
...
+
             ker_dir="/tmp/mymount"
```
whowutwut: What is ``ker_dir`` stand for?

whowutwut: actually, this is not even used, we should delete it.

whowutwut:
I did not review the initial check in of the code but this pull request caught my attention. I have some comments on the script. 

1. I don't like the name of this script.  ``getinstdisk``.   It's very difficult for me to know what this script is about.   I think we should come up with a script naming convention, whether to use _ or camel case. 

2. The log file is getinstalldisk_log, why do we call it ``getinstdisk``, then the log file is /tmp/getinstalldisk_log..    Suggest changing the name of the script to ``getinstalldisk`` to be consistent

3. The log file, I would suggest being renamed to ``getinstalldisk.log``.  We could also use $0.log 
Should we even be saving logs to /tmp? What about ``/var/log/xcat/$0.log`` so now if we called this script ``getinstalldisk``, it would create ``/var/log/xcat/getinstalldisk.log``

4. The log file should be a variable instead of `` >> /tmp/getinstalldisk_log`` repeated 

5. The ``/tmp/mymount`` should also be a variable and I don't like "mymount" as the mount point. It's not unique enough.  It should be something like ``/tmp/xcat.getinstalldisk.mount`` or something that we are fairly confident that a user would not also create

6. Line 47,  don't see a check that /tmp/mymount is not already mounted before we do a mkdir -p.. 
What would happen if something else is mounted there already? 

7. Line 40, The first time that we log, we should not append but do ``>`` so that it creates a new file. 
Otherwise, we should have a logging function that will produce a timestamp.  If we run this multiple times, it's going to keep appending to the log file.  ``echo Check the partition $partition > /tmp/getinstalldisk_log``

8. Around line 95 and 102 and 143 there's a `` rm -f /tmp/tmp_*;``   I think this is very risky to execute...  If we are creating many files, we should really do something like...`` mkdir -p /tmp/xcat.getinstalldisk/``
Then create all the files under /tmp/xcat.getinstalldisk/, then do a `` rm -rf /tmp/xcat.getinstalldisk`` at the end. 

9. /tmp/install_disk  should be a variable so if the user or anyone wants to change the name of this file in the future, we only have to change in one place

whowutwut:
Discussed with team about the review comments in a interlock call, the script is run during node provisioning so some of the issues that I bring up are not applicable and will not be changed to the code..  However, I do think many of the comments should be applied in the future as good practices in general to help maintainability.

daniceexi:
@xcat2/developers Yes, @whowutwut 's comments are inspirable for us to name the variables and files

zet809:
Seems OK to me, merge it.

Issue #573, opened 2016-01-06
[FVT]2.9.2:aix720 install error message "xcatd: possible BUG encountered by xCAT UDP service"


junxiawang:
How to reproduce:
1.install xcat on aix 720
During install
```
xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716.

[...]

PR #575, opened 2016-01-06
Fix defect#573：aix720 install error message "xcatd: possible BUG encountered by xCAT UDP service"

hu-weihua:
Fix issue #573 
In aix7.2 with xcat2.9.2,  get error message "xcatd: possible BUG encountered by xCAT UDP service: Undefined subroutine &main::update_udpcontext_from_sslctl called at /opt/xcat/sbin/xcatd line 716." when xcatd stop.

This issue has been fixed in xcat2.11.


daniceexi:
I have reviewed this change and agree to check in.

Issue #569, opened 2016-01-04
[DEV]Update xCAT 2.10 to xCAT 2.11 will overwrite the site."xcatsslversion" attribute.


daniceexi:
It's a bug in the xcatconfig that it always overwrite the "xcatsslversion" attribute when do the package update.

whowutwut:
To add some more information about the original problem that was encountered.  The cluster is updated to xCAT 2.11 by changing the yum repo config files and running ``yum update '*xCAT*'`` .  The following problem is seen: 
```
# lsxcatd -a
Unable to open socket connection to xcatd daemon on localhost:3001.
Verify that the xcatd daemon is running and that your SSL setup is correct.
Connection failure: SSL connect attempt failed with unknown error at /opt/xcat/lib/perl/xCAT/Client.pm line 265.
```

xCAT was running fine before with xCAT 2.10 and stops working updating to xCAT 2.11

Upon investigation, the ``xcatsslciphers`` and ``xcatsslversion`` was set at this site according to the Poodle attack security advisory (https://github.com/xcat2/xcat-core/wiki/2014_11_04_Poodle_Attack) 

```
# tabdump site | grep ssl
"xcatsslciphers","kDH:kEDH:kRSA:!SSLv3:!SSLv2:!aNULL:!eNULL:!MEDIUM:!LOW:!MD5:!EXPORT:!CAMELLIA:!ECDH",,
"xcatsslversion","TLSv1",,
```

However the ciphers only is supported with versions higher than TLSv1, so we modified to TLSv12 and things started working again.  @daniceexi  investigated and realized that we are overwriting the value of ``xcatsslversion`` and opened this Issue.

daniceexi:
Code has been checked in:

* 2.12: ec4ce73d5be41c86da62d0b1ff936526fb7d65fd
* 2.11: 3eb292a80f91ef449988b6a2b1558f6b7a019832
* 2.9: 4dd1cd6a128d731ad9958e5e366c87f8a181be53

caomengmeng:
Hi @daniceexi 
I am using the latest build.
1. change xcatsslversion value to be TLSv12
2. run: xcatconfig -f
3. check the xcatsslversion value, it is still the default value TLSv1

```
[root@c910f02c04p07 ~]# rpm -qa | grep -i xCAT
conserver-xcat-8.1.16-10.ppc64
perl-xCAT-2.12-snap201601271530.noarch
xCAT-buildkit-2.12-snap201601271531.noarch
[...]

PR #576, opened 2016-01-06
Issue 569: check the value of site.xcatsslversion before setting default value

daniceexi:
Refer to issue #569

### DIFF
--- a/xCAT-server/sbin/xcatconfig
+++ b/xCAT-server/sbin/xcatconfig
@@ -1426,16 +1426,17 @@ sub initDB
             }
         }
 
-        # add default value to site.xcatsslversion
-        $cmds = "$::XCATROOT/sbin/chtab key=xcatsslversion site.value=TLSv1;";
-        $outref = xCAT::Utils->runcmd("$cmds", 0);
-        if ($::RUNCMD_RC != 0)
-        {
-            xCAT::MsgUtils->message('E',"Could not add default value for site.xcatsslversion.");
+        # Set default value for site.xcatsslversion when update xcat
+        $cmds = "XCATBYPASS=Y $::XCATROOT/sbin/tabdump site 2>/dev/null |grep xcatsslversion";
+        xCAT::Utils->runcmd("$cmds", -1);
+        if ($::RUNCMD_RC != 0) {
+            # if site.xcatsslversion was not set, then set the default value TLSv1
+            $cmds = "$::XCATROOT/sbin/chtab key=xcatsslversion site.value=TLSv1;";
+            xCAT::Utils->runcmd("$cmds", 0);
+            if ($::RUNCMD_RC != 0) {
+                xCAT::MsgUtils->message('E',"Could not add default value for site.xcatsslversion.");
+            }
         }
-
-
-
    } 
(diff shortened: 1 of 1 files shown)


whowutwut:
@daniceexi I just copied over the security advisory over to GitHub about this and read and fixed up some of the content... https://github.com/xcat2/xcat-core/wiki/2014_11_04_Poodle_Attack. 

I noticed that for RHEL and SLES, it's **TLSv12** and for Ubuntu it's **TLSv1_2**, do you know why the difference in spelling 12 vs 1_2? 

The other thing worth considering, for xCAT 2.11 onwards, should we default to TLSv1.2 instead of TLSv1 to increase the level of security by default?   I think the security advisory that we wrote is not accurate stating that RHEL 6.x only supports TLSv1, RHEL 6.6 does support TLSv1.2 because I think openssl was updated at some point. 

```
[root@c910f02c05p07 ~]# cat /etc/*release*
Red Hat Enterprise Linux Server release 6.6 (Santiago)
Red Hat Enterprise Linux Server release 6.6 (Santiago)
cpe:/o:redhat:enterprise_linux:6server:ga:server

[root@c910f02c05p07 ~]# tabdump site | grep ssl
"xcatsslversion","TLSv12",,
"xcatsslciphers","kDH:kEDH:kRSA:!SSLv3:!SSLv2:!aNULL:!eNULL:!MEDIUM:!LOW:!MD5:!EXPORT:!CAMELLIA:!ECDH",,

[root@c910f02c05p07 ~]# lsxcatd -a 
Version 2.11 (git commit 9ea36ca6163392bf9ab684830217f017193815be, built Mon Nov 30 05:43:11 EST 2015)
This is a Management Node
dbengine=SQLite

[root@c910f02c05p07 ~]# openssl s_client -connect 127.0.0.1:3001 -tls1_2
...
---
SSL handshake has read 2052 bytes and written 579 bytes
---
New, TLSv1/SSLv3, Cipher is AES256-GCM-SHA384
Server public key is 2048 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : AES256-GCM-SHA384
...
```

daniceexi:
@whowutwut 
My point is xcat set the default value for xcatsslversion to TLSv1 make sure the poodle attack has been blocked. If customer wants to get new features from newer TLS version, they need to cover the ssl configuration by themselves. I suggest that don't involve xCAT code to make different ssl configuration base on the OS version, it will make xcat code complicated and hard to maintain.

The name of TLS like TLSv12 or TLSv1_2 is read by Perl ssl library. Because of the difference of perl ssl lib between the rh and ubuntu, the name of TLS version are different.

whowutwut:
OK, @tingtli , could you make sure that we have a testcase that verifies poodle attack is blocked using TLSv1 and also another one changing the xcatsslversion and xcatsslciphers combinations to make sure that we are providing a good error message if configured incorrectly...?  I asked @gurevichmark to see if we can detect and provide a more meaningful error message that today. It seems like we are printing a catch all message and it doesn't really point to the root cause.

PR #582, opened 2016-01-08
Use fifo pipe to enhance getadapter

chenglch:
Use fifo pipe to enhance getadapter

Currently the getadapter implementation store the nics information
in multiple files. This implementaion will store the records in the
database and add the following changes:

1. Add taskstate table to record the getadapter request.
2. Add route_request hook in xcatd to avoid of too much process.
3. Add adapterinfo column to keep the adapter information.
4. Use fifo pipe to send/recv the adapter message and wake up the waiting
    process.
5. Use alarm signal to handle the timeout event.


daniceexi:
@chenglch @hu-weihua 
Solution team requires to support the ping test feature likes we did for getmacs against Power Server. 

So please output the following messages for getadapters command:
* IP of each adapter if it has
* If this adapter is pingable to the xCAT MN in the install network.

xCAT-server/lib/xcat/plugins/getadapter.pm
```diff
@@ -125,328 +201,383 @@ sub handle_getadapters{
...
+
+    $SIG{CHLD} = 'DEFAULT';
+    $SIG{ALRM} = sub {
+        xCAT::MsgUtils->trace($VERBOSE, "d", "getadapter: alarm signal received");
+        xCAT::MsgUtils->message("W", "getadapter: alarm signal received");
+        exit(clean_up());
+    };
+    alarm(ALARM_TIMEOUT);
+    foreach my $node (@nodes) {
+        $pid = xCAT::Utils->xfork();
```
chenglch: I have some concern for this code which forks a process for every node. Actually, these commands do not take too much time. I have some thought to leverage process worker pool which has max limitation, If needed,  request a process from this pool. But I do not have plan to cover this concern here, as this is a common problem for xcat. We copy too much code to fork process, then wait process. Some implementation even have bug to handle the signal. I think xcat should provide a common library to solve this problem.

xCAT-server/lib/xcat/plugins/getadapter.pm
```diff
@@ -125,328 +226,487 @@ sub handle_getadapters{
...
+    my $parse_request_message_func = sub {
+        my $xml = shift;
+        my $xs  = new XML::Simple();
+        return $xs->XMLin($xml);
+    };
+
+    my $timeout_output_func = sub {
+        my $nodes_ptr = shift;
+        my $callback  = shift;
+        foreach my $node (@nodes) {
```
zet809: Where does nodes come from for this sub function?

zet809:

PR #583, opened 2016-01-08
support rscan command to scan kvmhost to get the virtual machine list

penguhyang:
Currently, the rscan command didn't support to get the kvm guests information from the kvmhost/hypervisor, and can't write the kvm guests information into the xCAT database if required.
This implementation will support rscan command to scan kvmhost to get the virtual machine list.
Support Options:
    -w  write the different kvm guest information in xCAT tables and output conflict information if occurs.
    -u  update the same kvm guest.
    -n  write the different kvm guest information in xCAT tables.
    -z  stanza output.


xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,176 @@ sub power {
...
+    foreach $dom (@doms) {
+        my $currxml=$dom->get_xml_description();
+        unless ($currxml) {
+            xCAT::SvrUtils::sendmsg([1,"VM does not appear to exist"], $callback,$hyper);
+            return;
+        }
+        my $domain=$parser->parse_string($currxml);
+        my $uuid = $domain->findnodes("/domain/uuid")->[0]->to_literal;
+        $uuid =~ s/^(..)(..)(..)(..)-(..)(..)-(..)(..)/$4$3$2$1-$6$5-$8$7/;
+        my $type = $domain->findnodes("/domain")->[0]->getAttribute("type");
```
immarvin: for operations on such nested hash structure, you need to verify that keys at each level is defined

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,182 @@ sub power {
...
+    my $dom;
+    eval {
+        @doms = $hypconn->list_all_domains();
+    };
+    my %host2kvm;
+    my @displaymsg;
+    if ($stanza) {
+        push @displaymsg, "# <xCAT data object stanza file>\n";
+    }
+
```
immarvin: can -w/-u/-n specified together? if not. a checking logic should be added

penguhyang: -w/-u/-n can be specified together.

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,199 @@ sub power {
...
+        foreach (@tabs) {
+            $db{$_} = xCAT::Table->new( $_, -create=>1, -autocommit=>0 );
+            if (!$db{$_}) {
+                xCAT::SvrUtils::sendmsg([1,"Error opening '$_' table for $name"], $callback,$hyper);
+                next;
+            }
+        }
+
+        #get 'node' and 'host' attributes from vm table...
+        my %table2kvm;
```
immarvin: the variable name ``table2kvm`` is not meaningful enough, please pick another name which can give more hints

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,199 @@ sub power {
...
+    foreach $dom (@doms) {
+        my $name=$dom->get_name();
+        my $currxml=$dom->get_xml_description();
+        unless ($currxml) {
+            xCAT::SvrUtils::sendmsg([1,"$name does not appear to exist"], $callback,$hyper);
+            next;
+        }
+        my $domain=$parser->parse_string($currxml);
+        my @paths = $domain->findnodes("/domain");
+        if (scalar @paths != 1) {
```
daniceexi: I did not see the posibility of return xml has multiple domain. Even it really has multiple domain, the correct procedure should be
* output a warning message that 'xcat will only handle the first one'
* only handle the the first one

xCAT-server/lib/xcat/plugins/kvm.pm
penguhyang: changed to hash_vm2host

xCAT-server/lib/xcat/plugins/kvm.pm
penguhyang: This if construct used to verify whether the ``/domain`` nodes can be found.
changed the msg to ``Can not find $name domain definition``

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,199 @@ sub power {
...
+            xCAT::SvrUtils::sendmsg([1,"$name should only have one domain"], $callback,$hyper);
+            next;
+        }
+        my ($uuid, $node, $vmcpus, $vmmemory, $vmnics, $vmstorage, $arch, $mac, $vmnicnicmodel);
+        if (defined $domain->findnodes("/domain/uuid")->[0]) {
```
daniceexi: The code has problem to check the availability of a hash key when **findnodes("/domain/uuid")** cannot find an obj for /domain/uuid.  When this happens, an error like following will be displayed:

~~~~
Can't use an undefined value as an ARRAY reference
~~~~

I recommended it to be: (If you agree, all the similar code need be changed)
~~~~
my $uuidobj = $domain->findnodes("/domain/uuid");
if ($uuidobj and defined($uuidobj->[0])) {
bla ...
}
~~~~

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,199 @@ sub power {
...
+        if (defined $domain->findnodes("/domain/devices/interface/model")->[0]) {
+            $vmnicnicmodel = $domain->findnodes("/domain/devices/interface/model")->[0]->getAttribute("type");
+        }
+        push @{$host2kvm{$uuid}}, join( ",", $type,$node,$hypervisor,$id,$vmcpus,$vmmemory,$vmnics,$vmstorage,$arch,$mac,$vmnicnicmodel );
+        my %db = ();
+        my @tabs = qw(vm);
+        foreach (@tabs) {
+            $db{$_} = xCAT::Table->new( $_, -create=>1, -autocommit=>0 );
+            if (!$db{$_}) {
+                xCAT::SvrUtils::sendmsg([1,"Error opening '$_' table for $name"], $callback,$hyper);
```
daniceexi: @penguhyang Could you confirm what will be displayed when this error happens? I think the 'Error' word is not necessary.

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,199 @@ sub power {
...
+                next;
+            }
+        }
+
+        #get 'node' and 'host' attributes from vm table...
+        my %table2kvm;
+        my @vm_nodes_hosts = $db{vm}->getAllNodeAttribs(['node','host']);
```
daniceexi: What will happen if opening table vm failed in above code?

daniceexi: Why do you read all the nodes from vm table in each node loop? It should be put before the for each node loop.

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,199 @@ sub power {
...
+                $updatetable->{vm}->{$node}->{memory} = $vmmemory;
+                $updatetable->{vm}->{$node}->{cpus} = $vmcpus;
+                $updatetable->{vm}->{$node}->{nics} = $vmnics;
+                $updatetable->{vm}->{$node}->{nicmodel} = $vmnicnicmodel;
+                $updatetable->{mac}->{$node}->{mac} = $mac;
+                $updatetable->{vpd}->{$node}->{uuid} = $uuid;
+                $updatetable->{nodetype}->{$node}->{arch} = $arch;
+                $updatetable->{nodehm}->{$node}->{mgt} = $type;
+            }
+            elsif ((exists $table2kvm{$node}) and ($table2kvm{$node} eq $hypervisor)) {
```
daniceexi: I think the logic is not correct here: you should write the vm to table too when it's really the same one with the one in the vm table instead of doing nothing.

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,199 @@ sub power {
...
+                #mark this node to delete in 'vm' 'mac' vpd' 'nodetype' 'nodehm' tables
+                $updatetable->{vm}->{'!*XCATNODESTODELETE*!'}->{$node} = $node;
+
+                $updatetable->{vm}->{$node}->{host} = $hypervisor;
+                $updatetable->{vm}->{$node}->{storage} = $vmstorage;
+                $updatetable->{vm}->{$node}->{memory} = $vmmemory;
+                $updatetable->{vm}->{$node}->{cpus} = $vmcpus;
+                $updatetable->{vm}->{$node}->{nics} = $vmnics;
+                $updatetable->{vm}->{$node}->{nicmodel} = $vmnicnicmodel;
+                $updatetable->{mac}->{'!*XCATNODESTODELETE*!'}->{$node} = $node;
```
daniceexi: what's this **XCATNODESTODELETE** for?

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,199 @@ sub power {
...
+                $updatetable->{nodehm}->{'!*XCATNODESTODELETE*!'}->{$node} = $node;
+                $updatetable->{nodehm}->{$node}->{mgt} = $type;
+            }
+        }
+        if ($create) {
+            unless (exists $table2kvm{$node}) {
+                $updatetable->{vm}->{$node}->{host} = $hypervisor;
+                $updatetable->{vm}->{$node}->{storage} = $vmstorage;
+                $updatetable->{vm}->{$node}->{memory} = $vmmemory;
+                $updatetable->{vm}->{$node}->{cpus} = $vmcpus;
```
daniceexi: I strongly recommend to reuse the code to set ``$updatetable``

xCAT-server/lib/xcat/plugins/kvm.pm
penguhyang: **!*XCATNODESTODELETE*!** is used to mark the kvm host to delete in tables.
```
    foreach (keys %$updatetable) {
        my $tabhandle = xCAT::Table->new($_, -create => 1);
        my $updates = $updatetable->{$_};
        if ($updates->{'!*XCATNODESTODELETE*!'}) {
            my @delkeys;
            foreach (keys %{ $updates->{'!*XCATNODESTODELETE*!'} }) {
                if ($_) { push @delkeys, { node => $_ }; }
            }
            if (@delkeys) { $tabhandle->delEntries(\@delkeys); }
            delete $updates->{'!*XCATNODESTODELETE*!'};
        }
        $tabhandle->setNodesAttribs($updatetable->{$_});
    }
```
this block of code is defined in ``sub dohyp()`` function.

xCAT-server/lib/xcat/plugins/kvm.pm
penguhyang: Yes, it should be put before the for each node loop since the vm table is same for each node.

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,233 @@ sub power {
...
+    };
+    if ($@) {
+        print STDERR "Unable to list all domains for $hyper" . $@->message . "\n";
+    }
+    my %host2kvm;
+    my @displaymsg;
+    if ($stanza) {
+        push @displaymsg, "# <xCAT data object stanza file>\n";
+    }
+    my %handle_vmtab = ();
```
immarvin: why define %handle_vmtab as a hash here?

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,233 @@ sub power {
...
+        'u' => \$update,
+        'z' => \$stanza,
+        'n' => \$create,
+    );
+    my @doms;
+    my $dom;
+    eval {
+        @doms = $hypconn->list_all_domains();
+    };
+    if ($@) {
```
immarvin: $@->message ? is the $@ a hash?

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,233 @@ sub power {
...
+        }
+        my $domain=$parser->parse_string($currxml);
+        my @paths = $domain->findnodes("/domain");
+        if (scalar @paths != 1) {
+            xCAT::SvrUtils::sendmsg([1,"Can not find $name domain definition"], $callback,$hyper);
+            next;
+        }
+        my ($uuid, $node, $vmcpus, $vmmemory, $vmnics, $vmstorage, $arch, $mac, $vmnicnicmodel);
+        my @uuidobj = $domain->findnodes("/domain/uuid");
+        if (@uuidobj and defined(@uuidobj->[0])) {
```
immarvin: ``@uuidobj->[0]`` is there any syntax error? can this pass perl -c?

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,233 @@ sub power {
...
+        if (@macobj and defined(@macobj->[0])) {
+            $mac = @macobj->[0]->getAttribute("address");
+        }
+        my @vmnicnicmodelobj = $domain->findnodes("/domain/devices/interface/model");
+        if (@vmnicnicmodelobj and defined(@vmnicnicmodelobj->[0])) {
+            $vmnicnicmodel = @vmnicnicmodelobj->[0]->getAttribute("type");
+        }
+        push @{$host2kvm{$uuid}}, join( ",", $type,$node,$hypervisor,$id,$vmcpus,$vmmemory,$vmnics,$vmstorage,$arch,$mac,$vmnicnicmodel );
+        if ($write) {
+            unless (exists $hash_vm2host{$node}) {
```
immarvin: $hash_vm2host was created at the very begining of sub rscan(), it might be outdated during the loop iterating. please consider this

penguhyang: $hash_vm2host keep the same in the loop iterating, so it will not outdated.

immarvin: will $hash_vm2host be updated when some nodes are added into vm table during iteration?

penguhyang: no, the $hash_vm2host will not be updated.
The vm table will not be updated during interation.

xCAT-server/lib/xcat/plugins/kvm.pm
penguhyang: $@ is set to the error message or $@ is set to the empty string.

xCAT-server/lib/xcat/plugins/kvm.pm
penguhyang: there is no syntax error.

xCAT-server/lib/xcat/plugins/kvm.pm
immarvin: why the kvm host should be deleted

xCAT-server/lib/xcat/plugins/kvm.pm
penguhyang: changed to $handle_vmtab

xCAT-server/lib/xcat/plugins/kvm.pm
penguhyang: the update procedure is: delete this entry in tables and write the new value about this entry in tables.

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,233 @@ sub power {
...
+        }
+        my @macobj = $domain->findnodes("/domain/devices/interface/mac");
+        if (@macobj and defined(@macobj->[0])) {
+            $mac = @macobj->[0]->getAttribute("address");
+        }
+        my @vmnicnicmodelobj = $domain->findnodes("/domain/devices/interface/model");
+        if (@vmnicnicmodelobj and defined(@vmnicnicmodelobj->[0])) {
+            $vmnicnicmodel = @vmnicnicmodelobj->[0]->getAttribute("type");
+        }
+        push @{$host2kvm{$uuid}}, join( ",", $type,$node,$hypervisor,$id,$vmcpus,$vmmemory,$vmnics,$vmstorage,$arch,$mac,$vmnicnicmodel );
```
immarvin: according to the design, the nodes with conflicting names should be notified , where is the code logic ?

penguhyang: please see 3159 line.

xCAT-server/lib/xcat/plugins/kvm.pm
immarvin: $@ is a string?  $@->message assumes $@ to be a hash.
in xcat plugin, use "xCAT::SvrUtils::sendmsg" to output message, do not use print

penguhyang: changed to use ``xCAT::SvrUtils::sendmsg``

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,233 @@ sub power {
...
+        my $hypervisor = $hyper;
+        my $id = $domain->findnodes("/domain")->[0]->getAttribute("id");
+        my @vmcpusobj = $domain->findnodes("/domain/vcpu");
+        if (@vmcpusobj and defined(@vmcpusobj->[0])) {
+            $vmcpus = @vmcpusobj->[0]->to_literal;
+        }
+        my @vmmemoryobj = $domain->findnodes("/domain/memory");
+        if (@vmmemoryobj and defined(@vmmemoryobj->[0])) {
+            my $mem = @vmmemoryobj->[0]->to_literal;
+            $vmmemory=$mem/1024;
```
immarvin:
````
 Memory Allocation

<domain>
  ...
  <maxMemory slots='16' unit='KiB'>1524288</maxMemory>
  <memory unit='KiB'>524288</memory>
  <currentMemory unit='KiB'>524288</currentMemory>
  ...
</domain>

memory
    The maximum allocation of memory for the guest at boot time. The memory allocation includes possible additional memory devices specified at start or hotplugged later. The units for this value are determined by the optional attribute unit, which defaults to "KiB" (kibibytes, 210 or blocks of 1024 bytes). Valid units are "b" or "bytes" for bytes, "KB" for kilobytes (103 or 1,000 bytes), "k" or "KiB" for kibibytes (1024 bytes), "MB" for megabytes (106 or 1,000,000 bytes), "M" or "MiB" for mebibytes (220 or 1,048,576 bytes), "GB" for gigabytes (109 or 1,000,000,000 bytes), "G" or "GiB" for gibibytes (230 or 1,073,741,824 bytes), "TB" for terabytes (1012 or 1,000,000,000,000 bytes), or "T" or "TiB" for tebibytes (240 or 1,099,511,627,776 bytes). However, the value will be rounded up to the nearest kibibyte by libvirt, and may be further rounded to the granularity supported by the hypervisor. Some hypervisors also enforce a minimum, such as 4000KiB. In case NUMA is configured for the guest the memory element can be omitted. In the case of crash, optional attribute dumpCore can be used to control whether the guest memory should be included in the generated coredump or not (values "on", "off"). unit since 0.9.11, dumpCore since 0.10.2 (QEMU only)

````

I think should consider unit of the memory instead of using 1024 in the ``$vmmemory=$mem/1024``

File: xCAT-server/lib/xcat/plugins/kvm.pm [hunk 17]
```diff
@@ -3016,6 +3016,233 @@ sub power {
...
+                my @vmnicsobj = $domain->findnodes("/domain/devices/interface/source");
+                if (@vmnicsobj and defined(@vmnicsobj->[0])) {
+                    $vmnics = @vmnicsobj->[0]->getAttribute("bridge");
+                    last;
+                }
+            }
+        }
+        my @vmstoragediskobjs = $domain->findnodes("/domain/devices/disk");
+        foreach my $vmstoragediskobj (@vmstoragediskobjs) {
+            if (($vmstoragediskobj->getAttribute("type") eq "file") and ($vmstoragediskobj->getAttribute("device") eq "disk")) {
```
immarvin:
````
device attribute since 0.1.4
    Indicates how the disk is to be exposed to the guest OS. Possible values for this attribute are "floppy", "disk", "cdrom", and "lun", defaulting to "disk".

    Using "lun" (since 0.9.10) is only valid when the type is "block" or "network" for protocol='iscsi' or when the type is "volume" when using an iSCSI source pool for mode "host" or as an NPIV virtual Host Bus Adapter (vHBA) using a Fibre Channel storage pool. Configured in this manner, the LUN behaves identically to "disk", except that generic SCSI commands from the guest are accepted and passed through to the physical device. Also note that device='lun' will only be recognized for actual raw devices, but never for individual partitions or LVM partitions (in those cases, the kernel will reject the generic SCSI commands, making it identical to device='disk').

````

I think we can only pick up the disk device with "device" is "disk", but do not restrict the "type" to be "file", since xCAT supports "dir://","nfs://"...

File: xCAT-server/lib/xcat/plugins/kvm.pm [hunk 18]
```diff
@@ -3016,6 +3016,233 @@ sub power {
...
+                    $vmstorage = @vmstorageobj->[0]->getAttribute("file");
+                    last;
+                }
+            }
+        }
+        my @archobj = $domain->findnodes("/domain/os/type");
+        if (@archobj and defined(@archobj->[0])) {
+            $arch = @archobj->[0]->getAttribute("arch");
+        }
+        my @macobj = $domain->findnodes("/domain/devices/interface/mac");
```
immarvin: if there are multiple network devices in vm guest, what's your will your strategy to select 1 interface out of them? is your code enough robust for this case?

xCAT-client/pods/man1/rscan.1.pod
```diff
@@ -17,6 +17,8 @@ The rscan command lists hardware information for each node managed by the hardwa
 
 For the management module of blade, if the blade server is a Flex system P node, the fsp belongs to the blade server also will be scanned.
 
+For the Virtualization of KVM, the kvm guest information will be scanned for each hypervisor.
```
immarvin: how about “For the KVM host, all the KVM guests on the specified KVM host will be scanned.” ?

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,233 @@ sub power {
...
+                $updatetable->{vm}->{$node}->{memory} = $vmmemory;
+                $updatetable->{vm}->{$node}->{cpus} = $vmcpus;
+                $updatetable->{vm}->{$node}->{nics} = $vmnics;
+                $updatetable->{vm}->{$node}->{nicmodel} = $vmnicnicmodel;
+                $updatetable->{mac}->{$node}->{mac} = $mac;
+                $updatetable->{vpd}->{$node}->{uuid} = $uuid;
+                $updatetable->{nodetype}->{$node}->{arch} = $arch;
+                $updatetable->{nodehm}->{$node}->{mgt} = "kvm";
+            }
+            elsif ((exists $hash_vm2host{$node}) and ($hash_vm2host{$node} eq $hypervisor)) {
```
immarvin: is this condition ``(exists $hash_vm2host{$node})`` still needed?

xCAT-client/pods/man1/rscan.1.pod
```diff
@@ -33,10 +35,16 @@ B<-u>          Updates and then prints out node definitions in the xCAT database
 
 For the Flex system manager, both the blade server and fsp object of xCAT will be updated if the mpa and slot id are matched to the object which has been defined in the xCAT database.
 
+For the Virtualization of KVM, the kvm guest information will be updated if the kvm guest name and the hypervisor matched to the object which has been defined in the xCAT database.
```
immarvin: For the KVM host, the information of the KVM guests which have been defined in xCAT database will be updated.

xCAT-client/pods/man1/rscan.1.pod
```diff
@@ -33,10 +35,16 @@ B<-u>          Updates and then prints out node definitions in the xCAT database
...
+
 Note: only the matched object will be updated.
 
+B<-n>          Only for the Virtualization of KVM, write the kvm guest information to xCAT database which kvm guest name have not been defined in xCAT database.
```
immarvin: For KVM host, the information of the KVM guests, which are not defined in xCAT database yet, will be written in xCAT database.

xCAT-server/lib/xcat/plugins/kvm.pm
penguhyang: there is no need to add `` (exists $hash_vm2host{$node}) ``.

xCAT-client/pods/man1/rscan.1.pod
```diff
@@ -33,10 +35,16 @@ B<-u>          Updates and then prints out node definitions in the xCAT database
...
+
 B<-w>          Writes output to xCAT database.
 
+For the Virtualization of KVM, output conflict information if the hypervisor not matched to the object which has been defined in the xCAT database with the same kvm guest name.
```
immarvin: For KVM host, updates the information of the KVM guests which have been defined in xCAT database with the same node name and KVM host, creates the definition of the KVM guests which do not exist in xCAT database , and  notifies user about the conflicting KVM guests that the name exist in xCAT database but the kvm host is different.

xCAT-client/pods/man1/rscan.1.pod
```diff
@@ -223,6 +231,43 @@ Output is similar to:
   blade  [SN#YL10JH184067]      Matched To =>[cmm01node01]
   blade  [SN#YL10JH184079]      Matched To =>[cmm01node03]
 
+8. To list all kvm guest information managed by hypervisor hyp01 in stanza format and write the output to the xCAT database, enter:
```
immarvin: 8. To scan the KVM host "hyp01", list all the KVM guest information on the KVM host in stanza format and write the KVM guest information into xCAT database.

xCAT-client/pods/man1/rscan.1.pod
```diff
@@ -223,6 +231,43 @@ Output is similar to:
...
+  kvm2
+    arch=x86_64
+    mac=
+    mgt=kvm
+    vmcpus=2
+    vmhost=hyp01
+    vmmemory=1024
+    vmnicnicmodel=virtio
+    vmnics=virbr0
+    vmstorage=
```
immarvin: why  ``vmstorage=`` is empty?

immarvin:
since there's a .pod file change. Can you run the create man pages script and include the .rst file changes in the pull request too?
 
see https://xcat-docs.readthedocs.org/en/latest/developers/guides/code/builds.html#man-pages for details

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3016,6 +3016,252 @@ sub power {
...
+    #operate every domain in current hypervisor
+    foreach $dom (@doms) {
+        my $name=$dom->get_name();
+        my $currxml=$dom->get_xml_description();
+        unless ($currxml) {
+            xCAT::SvrUtils::sendmsg([1,"fail to get the xml definition of $name"], $callback,$hyper);
+            next;
+        }
+        my $domain=$parser->parse_string($currxml);
+        my @paths = $domain->findnodes("/domain");
```
immarvin: @paths is defined,initialized and verified but never used in the following code , is it necessary?

whowutwut:
@penguhyang  Something is wrong with your commit, you are removing so many .rst files....   I don't think we can merge this as is.. I'm going to close this pull request for now so we do not accidently merge it.   Once you fix your commit, you can re-open for review.  

I think it was from the --amend commit.  I **really** think we should stop doing the amend commit unless the review has NOT yet started.  After someone makes a comment, we should commit again to the branch . This is very confusing to review....

penguhyang:
I opened a new pull request to replace this one as this one is complicated.
https://github.com/xcat2/xcat-core/pull/617

PR #584, opened 2016-01-08
docker lifecycle management commands framework

zet809:
The mini-design is available at https://github.com/xcat2/xcat-core/wiki/docker_instances_lifecycle_management
The implemented item are:
1. mkdocker/rmdocker/lsdocker
2. rpower start/stop/restart/pause/unpause/state
3. lsdocker <docker> -l|--logs
The left things are:
1. set ip for docker container
2. add more log info, debug info, explanation in the code.
3. usage info.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/docker.pm
+++ b/xCAT-server/lib/xcat/plugins/docker.pm
@@ -18,6 +18,7 @@ use lib "$::XCATROOT/lib/perl";
 use strict;
 use POSIX qw(WNOHANG nice);
 use POSIX qw(WNOHANG setsid :errno_h);
+use Errno;
 use IO::Select;
 require IO::Socket::SSL; IO::Socket::SSL->import('inet4');
 use Time::HiRes qw(gettimeofday sleep);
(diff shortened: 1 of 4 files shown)


whowutwut:
@zet809 Instead of describing what is left todo in the description, can you provide what it is that you implemented in the pull request?

daniceexi:
@zet809 you should mention in the description that this pull request is just for prereview instead of checking in.

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -90,10 +132,302 @@ sub send_msg {
...
+    if ($select->can_read($timeout)) {
+        my @ready_fds = $select->can_read(0);
+        foreach my $sockfd (@ready_fds) {
+            my $readbytes = 0;
+            my $res = "";
+            my $node_hash = $node_hash_variable{$sockfd};
+            if (defined($node_hash)) {
+                do {
+                    $readbytes = sysread($sockfd, $res, 65535, length($res));
+                } while ($readbytes);
```
chenglch: sysread may return -1

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -261,53 +886,114 @@ sub genreq {
...
-    my $socket = IO::Socket::INET->new( PeerAddr => $doc_hostname,
-                                                              PeerPort => $port,
-                                                              Timeout => 2,
-                                                              Blocking => 0
-                                      );
+    my $socket = IO::Socket::INET->new( PeerHost => $hostname,
+                                        PeerPort => $port,
+                                        Timeout => 2,
+                                        Blocking => 0);
     if ($socket) {
```
chenglch: As this is a non-block socket, do you need to check the $socket->connect to determine whether the connection has been completed?

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -316,121 +1002,25 @@ sub send_req {
                                                    SSL_key_file => $key_file,
                                                    Timeout => 0
                                       );
-
-
         if ($connect) {
```
chenglch: When you upgrade the socket to ssl socket, is this procedure in block mode or non-block mode?
If it is in non-block mode, how do you handle the ssl handshake? 
From my understanding, non-block ssl connection can not created immediately

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -195,30 +605,243 @@ sub preprocess_request {
...
+            last;
+        }
+        if (($pending_nodes) and ($concurrent_ssl_sessions lt $max_concur_ssl_session_allow)) {
+            my $node = shift @nodeargs;
+            my $ssl_connect = init_ssl_connection($node->[1]);
+            if (!defined($ssl_connect)) {
+                $callback->({error=>["Create SSL connection failed for docker $node->[0] on host $node->[1]->{host}"], errorcode=>1});
+            } 
+            else {
+                my $res = sendreq($ssl_connect, @$node);
```
chenglch: Do you need to check 'select->can_write()' before you send the request? This socket may work in connection state

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -195,30 +605,243 @@ sub preprocess_request {
...
+            }
+            elsif (/dockerflag=(.*)$/) {
+                $flagarg = $1;
+            }
+        }     
+        $genreq_ptr = \&genreq_for_mkdocker;
+        $nodetypetab = xCAT::Table->new('nodetype');
+        my $hosttab = xCAT::Table->new('hosts');
+        my $mactab = xCAT::Table->new('mac');
+        if (!defined($hosttab) or !defined($nodetypetab) or !defined($mactab) or !defined($vmtab)) {
```
chenglch: I think you'd better check the table object one by one. If $nodetypetab can not be created, the following request will fail in most cases. There is no need to try and waste the database connections.

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -195,30 +605,243 @@ sub preprocess_request {
...
+            }
+            # Wait for 10 * num_of_sessions 
+            if ($no_res_times > 5) {
+                last;
+            }
+        }        
 
-    return;
 
+        if (($pending_nodes eq 0) and ($pending_res eq 0)) { # No more nodes needed to be process, no more response is expected, end the loop
```
chenglch: what's the difference between L811 and L784 ?

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -195,30 +605,243 @@ sub preprocess_request {
...
+                $pre_pending_res = $pending_res;
+            }
+            elsif ($pre_pending_res eq $pending_res) {
+                $no_res_times++;
+                $timeout += $pending_res;
+            }
+            else {
+                $pre_pending_res = undef;
+                $no_res_times = 0;
+                $timeout = 0;
```
chenglch: As socket has timeout setting,  Can you handle the timeout event in the select method? errno may has this information.

xCAT-server/lib/xcat/plugins/docker.pm
zet809: You are right that the section around line 811 need to be removed.

xCAT-server/lib/xcat/plugins/docker.pm
zet809: I don't think so, the IO::Socket::SSL will check that.

xCAT-server/lib/xcat/plugins/docker.pm
zet809: As we discussed today, we can implement a better non-block select model in the future.

xCAT-server/lib/xcat/plugins/docker.pm
chenglch: If so, why not to use 'IO::Socket::SSL->new' to create a ssl socket directly?

xCAT-server/lib/xcat/plugins/docker.pm
zet809: No, sysread won't return -1. 
undef - for error
0 - at end of file
number of bytes read - success
So, I need to deal with 'undef' here.

xCAT-server/lib/xcat/plugins/docker.pm
chenglch: I'm wrong here, socket can handle the connection timeout(connection alive), but can't handle the recv or send timeout event , this retry logic should be implement in the session layer like what you have done.

chenglch:
As discussed locally, the socket communication part need to be refactored. Maybe we should leverage Coro/AnyEvent/LWP library to make the rest api request process easy to control. So I will not focus on the details and agree to merge.

PR #592, opened 2016-01-13
remove the xcat2_support_list file and instead point to the release notes in docs

whowutwut:
This pull request removes the ``docs/source/xcat2_support_list.rst`` file since that was initially created to track the supported OS versions.  Since then, we have moved over the Release Notes and that is a more complete picture of the OS that xCAT supports for each release. 

Update the link for the install guides to point to the release notes.  http://10.2.5.2/docs/os_support/
This will also need to be cherry picked to 2.11 once merged.


whowutwut:
@gurevichmark  , can you review?

docs/source/guides/install-guides/common_sections.rst
```diff
@@ -1,6 +1,13 @@
 
+.. BEGIN_see_release_notes
+
+For the current list of operating systems supported and verified by the development team for the different releases of xCAT, see the :doc:`xCAT2 Release Notes </overview/xcat2_release>`. 
+
+.. END_see_release_notes
+
 .. BEGIN_install_os_mgmt_node
 
+Install one of the supported operating systems on to your target management node.
```
gurevichmark: Small wording fix. No need for "to". Can just say "Install one of the supported operating systems on your target management node."

gurevichmark:
There is a small inconsistency for 2.11 between the release notes file and the wiki page with release notes.
The "xcat2_release.rst" file shows "UBT 14.4.3 LE" and "UBT 15.10 LE" as new supported OS. 
But the wiki release notes at https://github.com/xcat2/xcat-core/wiki/XCAT_2.11_Release_Notes say "Ubuntu 14.04.03" and "Ubuntu 15.10". No "LE" stated. Not sure how important this is. Not sure which one is correct.

whowutwut:
Thanks Mark, I updated the wording change in the pull request. 

@daniceexi  In the 2.11 Release notes under the Testing Sections, we have testing Ubuntu 14.04.3 on both Power and x86_64 machines.  Should the release notes over on RTD not have a "LE" tag which indicates ppc64le?

PR #595, opened 2016-01-14
Add details for security doc

hu-weihua:
For review conveniently, please refer to 
http://10.2.2.30/advanced/security/security.html

### DIFF
--- a/docs/source/advanced/security/security.rst
+++ b/docs/source/advanced/security/security.rst
@@ -1,0 +1,187 @@
+XCAT Security
+=============
+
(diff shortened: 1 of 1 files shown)


daniceexi:
I have reviewed the change and agree to merge. @hu-weihua Thanks for your effort.

docs/source/advanced/security/security.rst
```diff
@@ -0,0 +1,188 @@
+XCAT Security
```
whowutwut: I think we should keep consistent with 'xCAT' 

I see us mixing xCAT, XCAT and xcat but I think we should stick to one and use **xCAT**

PR #487, opened 2015-11-27
add bmcsetup in discovery doc, add chain document in Advanced Topics …

zet809:
…chapter

PR #599, opened 2016-01-14
Add changes to the chain man page missed in the 2.11 GA

whowutwut:
Add changes back to the 2.11 branch for the chain related man page entries.  This was missed in Schema.pm.  The original pull request is #487


zet809:
Looks good to me, merge it.

Issue #597, opened 2016-01-14
[FVT]Failed to generate expected logs in /var/log/xcat/xcat.log while installing SLES11.4 on x86_64


caomengmeng:
ENV:
MN: x86_64 sles11.04
set site.xcatdebugmode=1

xCAT build infor: 
```
c910f04x31:~ # rpm -qa |grep -i xCAT
syslinux-xcat-3.86-2
xCAT-genesis-scripts-x86_64-2.12-snap201512311931
ipmitool-xcat-1.8.11-3
elilo-xcat-3.14-4
conserver-xcat-8.1.16-10
xCAT-buildkit-2.12-snap201512311931
perl-xCAT-2.12-snap201601121533
xCAT-client-2.12-snap201512311931
xCAT-2.12-snap201601121533
xCAT-genesis-base-x86_64-2.9-snap201504212134
xCAT-server-2.12-snap201512311931
grub2-xcat-2.02-0.16.el7.snap201506090204
```

Steps:
1. Install sles11.04 on CN, then track the CN: /var/log/xcat/xcat.log file to get install information, but not as excepted, only few logs as below, and the status is "booting"
```
c910f04x35:~ # vi /var/log/xcat/xcat.log
+ for karg in '$(cat /proc/cmdline |sed -e '\''s/-/:/g'\'' -e '\''s/ /\n/g'\'')'
+ '[' netdevice = BOOTIF ']'
+ '[' netdevice = netdevice ']'
+ export PRINIC=34:40:B5:B9:5C:9E
+ PRINIC=34:40:B5:B9:5C:9E
+ '[' 17 = 20 ']'
+ for karg in '$(cat /proc/cmdline |sed -e '\''s/-/:/g'\'' -e '\''s/ /\n/g'\'')'
+ '[' UseSSH = BOOTIF ']'
+ '[' UseSSH = netdevice ']'
+ for karg in '$(cat /proc/cmdline |sed -e '\''s/-/:/g'\'' -e '\''s/ /\n/g'\'')'
+ '[' SSHPassword = BOOTIF ']'
+ '[' SSHPassword = netdevice ']'
+ for karg in '$(cat /proc/cmdline |sed -e '\''s/-/:/g'\'' -e '\''s/ /\n/g'\'')'
+ '[' Loghost = BOOTIF ']'
+ '[' Loghost = netdevice ']'
[...]

PR #600, opened 2016-01-15
correct failed to generate expected logs for SLES

penguhyang:
fix issue #597

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/scriptlib
+++ b/xCAT-server/share/xcat/install/scripts/scriptlib
@@ -11,7 +11,7 @@ declare -F msgutil_r &>/dev/null  || function msgutil_r {
    if [ -n "$logserver" ];then
       logger -n $logserver -t xcat -p local4.$msgtype "$msgstr" 
       if [ "$?" != "0" ];then
-         exec 3<>/dev/udp/$logserver/514 >/dev/null 2>&1;logger -s -t xcat -p local4.$msgtype "$msgstr" 1>&3  2>&1
+         exec 3<>/dev/udp/$logserver/514 >/dev/null;logger -s -t xcat -p local4.$msgtype "$msgstr" 2>&3
          if [ "$?" != "0" ];then
             logger -s -t xcat -p local4.$msgtype "$msgstr" 2>&1|nc $logserver 514 >/dev/null 2>&1
             if [ "$?" != "0" ];then


whowutwut:
@penguhyang  can you explain why on SLES we should not be referencing "$MASTER_IP" in the description for this pull request?   Is it not defined for SLES machines? 

I would offer the following change.. rather than testing for /etc/SuSe-release in each print out of the msg, we can test it at the very beginning...

```
if [ -f /etc/SuSE-release ]; then
    export MASTER_IP=""
else
    export MASTER_IP="#ENV:MASTER_IP#"
fi 
```

The only requires one place to change instead of every line we are doing a print out.

xCAT-server/share/xcat/install/scripts/post.xcat
```diff
@@ -371,3 +499,4 @@ then
     set +x
 fi
 
+
```
whowutwut: should not have this extra line... :-)

xCAT-server/share/xcat/install/scripts/post.sles.common
```diff
@@ -59,7 +59,12 @@ HOSTNAME=$(hostname -s)
...
-   msgutil_r "$MASTER_IP" "debug" "PRINIC=$PRINIC,HOSTNAME=$HOSTNAME" "/var/log/xcat/xcat.log"
+    if [ -f /etc/SuSE-release ]; then
+        msgutil_r "" "debug" "PRINIC=$PRINIC,HOSTNAME=$HOSTNAME" "/var/log/xcat/xcat.log"
+    else
+        msgutil_r "$MASTER_IP" "debug" "PRINIC=$PRINIC,HOSTNAME=$HOSTNAME" "/var/log/xcat/xcat.log"
+    fi
 fi
 /sbin/portmap
 
+
```
whowutwut: should not have this extra line

penguhyang: Thanks for review. It's looks better.
```
if [ -f /etc/SuSE-release ]; then
    export MASTER_IP=""
else
    export MASTER_IP="#ENV:MASTER_IP#"
fi 
```

xCAT-server/share/xcat/install/scripts/post.xcat
penguhyang: @whowutwut Thanks for review, I think we should not be referencing "$MASTER_IP" for SLES. As in the function msgutil_r, the logger command not support -n option for sles. 
If using logger with -n option for sles, there will be a error output in the xcat.log, and the following output will lack the necessary procedure.

xCAT-server/share/xcat/install/scripts/post.xcat
```diff
@@ -1,7 +1,12 @@
 #
 # Run xCAT post install
 #
-export MASTER_IP="#ENV:MASTER_IP#"
+if [ -f /etc/SuSE-release ]; then
+    export MASTER_IP=""
+else
+    export MASTER_IP="#ENV:MASTER_IP#"
+fi
+export XCAT_MASTER_IP="#ENV:MASTER_IP#"
```
whowutwut: Why did we define XCAT_MASTER_IP?

xCAT-server/share/xcat/install/scripts/post.xcat
```diff
@@ -1,7 +1,12 @@
 #
 # Run xCAT post install
 #
-export MASTER_IP="#ENV:MASTER_IP#"
+if [ -f /etc/SuSE-release ]; then
```
whowutwut: Can we add some comments here on why we are testing for SLES?

xCAT-server/share/xcat/install/scripts/post.xcat
whowutwut: On SLES machines, this will just show up blank right? and we are adding this into the output message... also down at line we are exporting XCATSERVER with this variable, so that variable will be blank? or something else?

penguhyang: Because of the following code using the ``MASTER_IP`` to download the ``postscripts`` from MN, for example, ``msgutil_r "$MASTER_IP" "debug" "trying to download postscripts from http://$MASTER_IP$INSTALLDIR/postscripts/" "/var/log/xcat/xcat.log"``. The system will halt if the ``MASTER_IP`` is null. So I define a new variable ``XCAT_MASTER_IP`` the store the IP of the MN to make the code right without conflict the ``XCAT_MASTER``

xCAT-server/share/xcat/install/scripts/post.xcat
penguhyang: Thanks for review, changed to add comments.

immarvin:
@penguhyang , according to the design,  the logs in CN should be forwarded to MN if site.xcatdebugmode=1. If this defect is caused by some problems inside function "msgutil_r", we should try to fix  "msgutil_r". Bypassing the problem by not passing MASTER_IP to "msgutil_r" will affect the designed behavior, this is not the right derection.

penguhyang:
@immarvin Thanks for review. 
This error is caused by ``exec 3<>/dev/udp/$logserver/514 >/dev/null 2>&1;logger -s -t xcat -p local4.$msgtype "$msgstr" 1>&3  2>&1`` in msgutil_r. We can skip this line if we set ``MASTER_IP`` to null. 
If we set ``MASTER_IP`` to null, the ``logger -t xcat -p local4.$msgtype "$msgstr"`` will be run. The logs in CN also can be forwarded to MN.

xCAT-server/share/xcat/install/scripts/post.xcat
whowutwut: Still not really understanding why we need to redefine XCAT_MASTER_IP.  I looked at the function ``msgutil_r...  I looked at the define of msgutil_r in scriptlib file.  The XCAT_MASTER_IP is going to be passed into as argument 3, so ``msgstr``.  

I guess my point is that if ``#ENV:MASTER_IP#`` is not evaluating right (or evaluating null), then msgstr is not going to be printed correctly anyway on SLES, so why not just pass in $MASTER_IP like we were doing before? 

Basically, delete line 11 and don't change any code below in the calls to msgutil_r.

penguhyang: @whowutwut Thanks for review.
After the code change, the ``/var/log/xcat/xcat.log`` will have a integrated running procedure in the node.  For example
```
+ logger -n 10.2.9.12 -t xcat -p local4.debug '/opt/xcat/xcatinstallpost generat
ed'
logger: invalid option -- 'n'
usage: logger [-is] [-f file] [-p pri] [-t tag] [-u socket] [ message ... ]
+ '[' 1 '!=' 0 ']'
+ exec
+ logger -s -t xcat -p local4.debug '/opt/xcat/xcatinstallpost generated'
+ '[' 0 '!=' 0 ']'
+ '[' -n /var/log/xcat/xcat.log ']'
++ dirname /var/log/xcat/xcat.log
+ local logdir=/var/log/xcat
+ '[' '!' -d /var/log/xcat ']'
+ echo '/opt/xcat/xcatinstallpost generated'
/opt/xcat/xcatinstallpost generated
+ cat
+ chmod 755 /opt/xcat/xcatdsklspost
+ '[' '!' -x /opt/xcat/xcatdsklspost ']'
+ '[' 1 '!=' 0 ']'
+ msgutil_r 10.2.9.12 debug '/opt/xcat/xcatdsklspost generated' /var/log/xcat/xc
at.log
+ local logserver=10.2.9.12
+ local msgtype=debug
+ local 'msgstr=/opt/xcat/xcatdsklspost generated'
+ local logfile=/var/log/xcat/xcat.log
+ '[' -z debug ']'
+ '[' -n 10.2.9.12 ']'
+ logger -n 10.2.9.12 -t xcat -p local4.debug '/opt/xcat/xcatdsklspost generated
'
logger: invalid option -- 'n'
usage: logger [-is] [-f file] [-p pri] [-t tag] [-u socket] [ message ... ]
+ '[' 1 '!=' 0 ']'
+ exec
+ logger -s -t xcat -p local4.debug '/opt/xcat/xcatdsklspost generated'
+ '[' 0 '!=' 0 ']'
+ '[' -n /var/log/xcat/xcat.log ']'
++ dirname /var/log/xcat/xcat.log
+ local logdir=/var/log/xcat
+ '[' '!' -d /var/log/xcat ']'
+ echo '/opt/xcat/xcatdsklspost generated'
/opt/xcat/xcatdsklspost generated
++ sed '/^#\s*postbootscripts-start-here/,/^#\s*postbootscripts-end-here/ d' /xc
atpost/mypostscript
+ TMP='set -x

. /xcatpost/xcatlib.sh

# global value to store the running status of the postbootscripts,the value is n
on-zero if one postbootscript failed
return_value=0
# subroutine used to run postscripts
run_ps () {
 local ret_local=0
 mkdir -p "/var/log/xcat"
 # On some Linux distro, the rsyslogd daemon write log files with permision
 # other than root:root. And in some case, the directory /var/log/xcat was
 # created by xCAT, and had root:root ownership. In this way, rsyslogd
 # did not have enough permission to write to log files under this directory.
 # As a dirty hack, change the ownership of directory /var/log/xcat to the
 # same ownership of directory /var/log.
 chown root:root "/var/log/xcat"
 local logfile="/var/log/xcat/xcat.log"

 if [ -f $1 ]; then
  echo "`date` Running postscript: $*"
  msgutil_r "$MASTER_IP" "info" "`date` Running postscript: $*" "$logfile"
  if [ "$XCATDEBUGMODE" = "1" ]; then
     local compt=$(file $1)
     local reg="shell script"
     if [[ "$compt" =~ $reg ]]; then
        bash -x ./$@ 2>&1
        ret_local=$?
     else
        ./$@ 2>&1 | tee -a $logfile |  logger -t xcat -p debug
        ret_local=${PIPESTATUS[0]}
     fi
  else
     ./$@ 2>&1 | tee -a $logfile
     ret_local=${PIPESTATUS[0]}
  fi

  if [ "$ret_local" -ne "0" ]; then
     return_value=$ret_local
  fi
  echo "Postscript: $* exited with code $ret_local"
  msgutil_r "$MASTER_IP" "info" "`date` postscript $* return with $ret_local" "$
logfile"
 else
  echo "`date` Postscript $1 does NOT exist."
  msgutil_r "$MASTER_IP" "info" "`date` Postscript $1 does NOT exist." "$logfile
"
  return_value=-1
 fi

 return 0
}
# subroutine end
```

penguhyang:
If we set MASTER_IP to null, the logger -t xcat -p local4.$msgtype "$msgstr" will be run. The logs in CN also can be forwarded to MN. 
But as SLES using the syslog_ng, the syslog_ng need configuration in the ``/etc/syslog-ng/syslog-ng.conf``. This configuration was completed in the ``syslog`` scripts which was running in the mypostscript defined in post.xcat.
So before running the ``syslog`` script, the logs can't be forwarded to MN.

penguhyang:
@whowutwut @immarvin Code updated.

immarvin:
looks reasonable for me, since @penguhyang have finished the unit test, i will merge it so that FVT can resume the test

PR #596, opened 2016-01-14
fix defect 94 [DEV]<xcatconfig -d> failed to determine the correct site.forwarder

immarvin:
fix https://github.com/xcat2/xcat-core/issues/594

PR #603, opened 2016-01-15
Revert "fix defect 94 [DEV]<xcatconfig -d> failed to determine the correct site.forwarder"

immarvin:
Reverts xcat2/xcat-core#596


whowutwut:
@immarvin  Looking at this some more, i think this does not do what we need.  I think maybe we did not need to revert because the changes went into the xcat-core:onissue594 branch instead of the master branch :-)  so we just undid the changes to a separate branch that we will delete anyway 

Pull request #604 looks like it's going to the right place now, I deleted the branch for this revert .. .and i think we just need to delete the onissue594 branch to clean up. 

image

Issue #594, opened 2016-01-14
[DEV]"xcatconfig -d" failed to determine the correct "site.forwarder" and "site.domain"


immarvin:
The code logic in line#1063 is wrong, it take ip address "10.1.0.1" the same as "10.1.0.101" by mistake.

````
1059     my @tmpnames =
1060     xCAT::Utils->runcmd(
1061        "/bin/grep ^[^#]*nameserver /etc/resolv.conf | awk '{print \$2}'", -
     1);
1062     foreach my $ip (@tmpnames) {
1063       if (!grep(/$ip/, @MNnodeipaddr))
1064       {     # if not the MN
1065          push @names,$ip;  # add it as a forwarder
1066       }
1067     }
````

The logic to determine the domain failed it "hostname -d" output nothing.
````
1105     else    # linux
1106     {
1107         $domain = `hostname -d`;
1108     }
````

the behavior of "hostname -d" depends on different linux distributions and system configurations. 
For ubuntu. it works when the "/etc/hosts" contains the host entry with the "ip  FQDN hostname "
````
10.5.106.101 xcat2-11mn.clusters.com xcat2-11mn 
````

however, for redhat, this does not work, i think we should look into this

daniceexi:
@immarvin Will you fix this issue? If yes, please assign it to you. Make @penguhyang to fix it also ok.
Please at least set the labels for a new issue.

PR #604, opened 2016-01-15
fix defect 94 [DEV]<xcatconfig -d> failed to determine the correct site.forwarder

immarvin:


whowutwut:
I believe the issue this is fixing is #594, not 94.

PR #611, opened 2016-01-18
fix issue 601: def command does not work for docker instance definition

zet809:
For issue 601


daniceexi:

whowutwut:
@zet809 , you need to add the # before 601 for Github to make the cross reference.. can you modify your post above?   What do we title this "def" command... does all the lsdef and chdef commands break because of that spelling error?

zet809:
@whowutwut Thank you victor, after I modify the description of this pull request, the defect 601 can reference it. 
And for your question, the answer is yes, all the command mkdef/rmdev are all hung. 
To correct the spelling issue can fix this issue, but not exactly sure it is the core reason since it won't cause issue on my dev environment. This issue may be triggered together with other codes.

PR #612, opened 2016-01-18
modify default dockerhost port, modify logic for rpower state

zet809:
1. the setupdockerhost will enable 2375 on dockerhost, so modify default dockerhost to 2375.
2. modify logic for rpower state to deal with both chunked data and content. 
3. Add "rmdocker -f"

### DIFF
--- a/xCAT-server/lib/xcat/plugins/docker.pm
+++ b/xCAT-server/lib/xcat/plugins/docker.pm
@@ -180,6 +180,11 @@ my %command_states = (
         },
     },
     rmdocker => {
+        force => {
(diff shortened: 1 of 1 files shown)


zet809:
#658 Can be fixed with this full request.

chenglch:
The usage of $get_another_pkg and $pending_res is a little confusing, You'd better define a boolean flag variable like 'remain' in '$node_hash_variable{$sockfd}' to  replace the global variable '$pending_res'.  This work is needed in the next phase, for a defect issue it's OK to merge.

PR #621, opened 2016-01-20
Enhance cfm2xcat

daniceexi:
Make cfm2xcat command generate the syncfile which has the new
format that adding a target node list for each cfg entry


hu-weihua:
I think the code can work, I agree to merge

PR #622, opened 2016-01-20
Enhance AIX support

daniceexi:
1. support 'updatenode -S -u' for aix to only do update but
   not install new packages
2. fix an issue in aixremoteshell that .pub files for ssh key
   were not created


hu-weihua:
I think it can work. I agree to merge.

Issue #537, opened 2015-12-15
[FVT]restart xcatd get error message for bugs.


tingtli:
Restart xcatd on 2.11 will get following message.

Dec 15 01:32:42 c910f03c05k23 xcat[3780]: xcatd: possible BUG encountered by xCAT UDP service: Magic number checking on storable file failed at /usr/lib64/perl5/vendor_perl/Storable.pm line 399, at /opt/xcat/lib/perl/xCAT/Table.pm line 110.

hu-weihua:
@tingtli , I tried to reproduce this issue on c910f03c05k23, but didn't hit this issue any more.
Could you offer a specific steps to reproduce this issue? thanks

chenglch:
@hu-weihua @tingtli , Seems that this bug as same as the bug I met when I clean up database in the signal handler.  You can see the detail to reprocess this issue.
https://github.com/chenglch/xcat-core/blob/getadapter/xCAT-server/lib/xcat/plugins/getadapter.pm#L227

I think this bug is not just related to xcatd, but also related to the database. If DB access process is down, 'table->new' will cause this problem. How to handle this situation when the DB access process stopped unexpectedly? Currently the plugin process is blocked when trying to connect to the database.

I have already find the root cause and talked this to @daniceexi last week, he knows the details.

daniceexi:
@tingtli, @hu-weihua We waiting for @chenglch 's fix, so see whether it can be fixed.

whowutwut:
@cxhong  and I saw an issue recently where the PostgreSQL DB was running but there was something wrong with it.  The nodeset commands would just time out.  But I restarted postgresql and the commands still worked so I didn't assume it was the DB.  It wasn't until we changed back to SQLite that xCAT started working again.

daniceexi:
@whowutwut maybe the issue not caused by the DB itself isntead of xCAT DB process. So restart xcatd would resolve the problem. If this is the case, the xcat DB process might get into an unknown state.

whowutwut:
@daniceexi  The restart of xcatd did not resolve the problem.  It was very strange.  The only thing that resolved it was changing back to SQLite.  If it happens again , ill try and debug but that environment is not there anymore

daniceexi:
@tingtli Could you verify whether still can recreate this issue in the latest build?

tingtli:
I didn't see the error in log. But why did I see different log information when i restart xcatd  for three times.



[...]

PR #635, opened 2016-01-21
Partial issue #537, try to detect the error of db access process

chenglch:
This patch add changes to detect the error and enter in direct db access
mode so that the other process can access the data without db access
process temporarily. Direct access is not suggested way and currently
nodelist (noderange) can not support this mode.

Partial issue: #537


daniceexi:

PR #645, opened 2016-01-22
cover rh6.7 for post.xcat restructure

### LINKED ISSUE #638: [FVT] Design Bug: Not Cover RH6.7 for Post.xcat Restructure
The post.xcat restructure design doesn't cover the RH6.7: https://github.com/xcat2/xcat-core/wiki/post.xcat_restructure
As we support the latest two version for each OS, so please take RH6.7 in consideration and make it works.

penguhyang:
fix #638


immarvin:
agree to merge it

PR #646, opened 2016-01-22
refine the document : create the bridge with a script

immarvin:
Since the commands to create the bridge will break the network connection on “eth0”, you’d better create a script such as “mkbridge” to create the bridge instead of running the commands sequentially.  it.


daniceexi:

Issue #649, opened 2016-01-22
[FVT]2.12:docker : command rpower 's  return value need be modified


junxiawang:
Test OS:rh7
Test xCAT version:
Version 2.12 (git commit fa2237d141c49e01a630b73c0f0a8e8a69a877fa, built Thu Jan 21 19:30:51 EST 2016)
How to reproduce:
```
[root@c910f04x30 xcat-core]# rpower host01c01 start
host01c01: Error: container already started
[root@c910f04x30 xcat-core]# echo $?
0
```

PR #652, opened 2016-01-25
Fix issue 649: [FVT]2.12:docker : command rpower 's return value need be modified

zet809:
Fix issue #649


chenglch:
I have reviewed this change and agree to merge

PR #653, opened 2016-01-25
change the pipework command to avoid confuse

immarvin:


immarvin:
@daniceexi , good idea. I have appended it

daniceexi:

Issue #627, opened 2016-01-20
[FVT] The postscript confignetwork run out of control and exited with zero exit code


neo954:
The postscript ``confignetwork`` run out of control and exited with zero exit code.

As you can see from the output of the script, there is usage of the command ``grep``. That may due to the script ``confignetwork`` was passing wrong command line arguments to ``grep``

While reaching fatal error and failing configure the network, the proper behavior of ``confignetwork`` should be exited with non-zero exit code.

    # updatenode c910f02c03p23 confignetwork
    c910f02c03p23: xcatdsklspost: updating VPD database
    c910f02c03p23: xcatdsklspost: downloaded postscripts successfully
    c910f02c03p23: Wed Jan 20 00:16:13 EST 2016 Running postscript: confignetwork
    c910f02c03p23: [I]: NetworkManager is inactive.
    c910f02c03p23: Usage: grep [OPTION]... PATTERN [FILE]...
    c910f02c03p23: Try 'grep --help' for more information.
    c910f02c03p23: [I]: All valid nics and device list:
    c910f02c03p23: [I]:
    c910f02c03p23: [I]: [E]: Error: bond0 eth2@eth1 pair is invalid nic and nicdevice pair.
    c910f02c03p23: ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
    c910f02c03p23: configure nic and its device :
    c910f02c03p23: [E]: Error : please check nictypes for .
    c910f02c03p23: ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
    c910f02c03p23: configure nic and its device : [E]: Error: bond0 eth2@eth1 pair is invalid nic and nicdevice pair.
    c910f02c03p23: [E]: Error : please check nictypes for [E]: Error: bond0 eth2@eth1 pair is invalid nic and nicdevice pair..
    c910f02c03p23: Postscript: confignetwork exited with code 0
    c910f02c03p23: Running of postscripts has completed.

neo954:
This is the node definition of ``c910f02c03p23``. As you can see here, the attributes ``nictypes.eth1`` and ``nictypes.eth2`` are missing. I made this definition on purpose.

    # lsdef c910f02c03p23
    Object name: c910f02c03p23
        arch=ppc64
        cons=hmc
        currchain=boot
        currstate=boot
[...]

PR #661, opened 2016-01-25
fix defect 627 The postscript confignetwork run out of control and ex…

bybai:
#627 The postscript confignetwork run out of control and exited with zero exit code


neo954:
Will merge this one. Only two quotes were added.

Issue #628, opened 2016-01-20
[FVT] The postscript confignetwork failed and exited with zero exit code when command brctl was absent on the compute node


neo954:
The postscript ``confignetwork`` failed and exited with zero exit code when command ``brctl`` was absent on the compute node.

While reaching fatal error and failing configure the network, the proper behavior of confignetwork should be exited with non-zero exit code.

On RHEL 7.2, the command ``brctl`` belongs to the rpm package ``bridge-utils``. But this package is not in the default RHEL 7.2 package list shipped with xCAT.

    # updatenode c910f02c03p23 confignetwork
    c910f02c03p23: xcatdsklspost: updating VPD database
    c910f02c03p23: xcatdsklspost: downloaded postscripts successfully
    c910f02c03p23: Wed Jan 20 00:45:32 EST 2016 Running postscript: confignetwork
    c910f02c03p23: [I]: NetworkManager is inactive.
    c910f02c03p23: [I]: All valid nics and device list:
    c910f02c03p23: [I]: br0 eth1
    c910f02c03p23: ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
    c910f02c03p23: configure nic and its device : br0 eth1
    c910f02c03p23: [I]: create_bridge_interface ifname=br0 _brtype=bridge _port=eth1 _pretype=ethernet
    c910f02c03p23: [I]: Pickup xcatnet, "", from NICNETWORKS for interface "br0".
    c910f02c03p23: [I]: create_raw_eth_interface_for_br ifname=eth1 _bridge=br0 _mtu=
    c910f02c03p23: [I]: create_persistent_ifcfg ifname=eth1 inattrs=ONBOOT=yes,TYPE=Ethernet,BRIDGE=br0,BOOTPROTO=none
    c910f02c03p23: ['ifcfg-eth1']
    c910f02c03p23: [I]: >> DEVICE="eth1"
    c910f02c03p23: [I]: >> BOOTPROTO="none"
    c910f02c03p23: [I]: >> NAME="eth1"
    c910f02c03p23: [I]: >> HWADDR="16:3f:4d:b4:fe:04"
    c910f02c03p23: [I]: >> ONBOOT="yes"
    c910f02c03p23: [I]: >> TYPE="Ethernet"
    c910f02c03p23: [I]: >> BRIDGE="br0"
    c910f02c03p23: [I]: brctl addbr br0
[...]

PR #662, opened 2016-01-25
issue 628 The postscript confignetwork failed and exited with zero ex…

bybai:
#628 The postscript confignetwork failed and exited with zero exit code when command brctl was absent on the compute node

### DIFF
--- a/xCAT/postscripts/nicutils.sh
+++ b/xCAT/postscripts/nicutils.sh
@@ -608,6 +608,8 @@ function migrate_ip {
 
 }
 
+
+
 ###############################################################################
 #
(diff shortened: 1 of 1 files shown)


xCAT/postscripts/nicutils.sh
```diff
@@ -643,6 +647,9 @@ function add_if() {
     BRIDGE=$3
 
     if [[ $BRIDGE == "bridge_ovs" ]]; then
+g_info "ovs-vsctl add-br $BNAME"
```
neo954: What is ``g_info``? I believe something is missing here.

PR #668, opened 2016-01-26
Fix pmr 88439,227,000 xcat bug 573

hu-weihua:
After debug. I found in ``xcatd`` we use number to indicate signal, such like ``12``, ``2``,  but unfortunately the same number has different meaning between AIX and LINUX.  So some our signal handler almost lose efficacy. that result in some unexpected problem, such like core dump. socket pipe broken.  

I change some important signal to macro definition in aix 2.9. (not for all signal just be caution, we have customer is using it. we can't do test totally). 

I also enhance the robustness of ``do_discovery_process`` to avoid ``storable file failed`` in some unexpected case.  I adjust indent of ``do_discovery_process`` at the same time.

### DIFF
--- a/xCAT-server/sbin/xcatd
+++ b/xCAT-server/sbin/xcatd
@@ -306,7 +306,7 @@ my $socket;
       my $pid = <$installpidfile>;
       if ($pid) {
           $retry=100; #grace period for old instance to get out of the way, 5 seconds
-          kill 12,$pid;
+          kill 'USR2',  $pid;
           yield(); # let peer have a shot at closure
       }
       close($installpidfile);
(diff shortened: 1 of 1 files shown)


daniceexi:
The change looks good. I agree to merge. Thanks.

daniceexi:
Please make a pull request for this issue to master branch. I recommend to replace all the signal number with signal name for the master branch.

chenglch:
@daniceexi I have already discussed with @hu-weihua , in master branch, to avoid of merge conflict, I will replace all the signal number with signal name when I fixing the signal handler defect about xcat DB process.

PR #669, opened 2016-01-26
For IBM Power S822LC and S812LC, pause before doing the mc reset cold

whowutwut:
For the IBM Power servers indicated above, there should not be a need to issue a ``mc reset cold`` after applying the BMC configurations inband.  However, there's an issue where the BMC only responds to the new IP for about 10 seconds and then goes dead.  To work around, we want to give the BMC some time to apply the settings before issuing the ``mc reset cold``.


daniceexi:

whowutwut:
Cherry picked this fix to 2.11 branch 5a22467c8875022bcc1bd0733aa553731845bcbf 553271c63954340aefaaeb6b5d993aeb9ab5bf83

PR #670, opened 2016-01-27
Story 85827, Hand the case that the disk does not have WWN

xuweibj:
The method of sort drivers is same with before.
Modified:
If there is disk has ID_WWN, choose the disk has smallest ID_WWN. 
If all the disks has no ID_WWN, check their DEVPATH, choose the one has the smallest DEVPATH.
If all the disks has no ID_WWN and DEVPATH, sort by their name, choose the first one.

For Ubuntu, awk not found at that time, replace awk to cut.

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/getinstdisk
+++ b/xCAT-server/share/xcat/install/scripts/getinstdisk
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 #-----------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


zet809:
@xuweibj A simple discussion in pull request #684 shows that we'd like to use '#!/bin/sh' as the shebang for this script, will you pls modify it in line 1 and verify again in RH and Sles?

xuweibj:
@zet809 I modified the line 1 and verified in RH 6.7/7.2 and Sles11/12, but just run this script, is this enough?

whowutwut:
@zet809  why did we decide to use #!/bin/sh?  
@xuweibj  I think we need to verify on Ubuntu too, that is the os that will symlink #!/bin/sh -> #!/bin/dash.  Will be get some unexpected behavior?

daniceexi:
Ubuntu just supports dash in initrd. So we have to use /bin/sh for system to select the corresponding shell. 
And yes, all the platform need be tested.

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -113,36 +153,61 @@ if [ -z "$install_disk" ]; then
...
-                echo "$disk $disk_wwn" >> "$tmpfile""thirddisks"
-                echo "Add disk: $disk $driver $disk_wwn into thirddisks" >> $logfile
-                ;;
+        elif [ "$disk_path" ]; then
+        
+            has_path=1
+            file_pre="path"
+            disk_data=$disk_path
+
+        elif [ $has_path -eq 1 ]; then
```
zet809: I think we shall add some log info here to describe why go to the next loop.

xuweibj:
Hi @zet809 I added some log info, please help review again.

zet809:

Issue #660, opened 2016-01-25
[FVT] lsdocker node -l --logs return too many  illegal code


junxiawang:
Test OS: rhels7
Test xCAT version:
[root@c910f04x30 ~]# lsdef -v
lsdef - Version 2.12 (git commit fa2237d141c49e01a630b73c0f0a8e8a69a877fa, built Thu Jan 21 19:30:51 EST 2016)
How to reproduce:
```

[root@c910f04x30 ~]# lsdocker host01c01  -l --logs
host01c01: ????????home
????????lib
????????lib64
????????media
????????mnt
????????opt
????????proc
????????root
????????run
????????
host01c01: sbin
????????srv
????????sys
????????tmp
????????usr
????????var
???????*/bin/bash: line 4: bin: command not found
???????:eth0      Link encap:Ethernet  HWaddr 02:42:0a:00:00:01  
???????<          inet addr:10.0.0.1  Bcast:0.0.0.0  Mask:255.0.0.0
???????8          inet6 addr: fe80::42:aff:fe00:1/64 Scope:Link
???????=          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
???????@          RX packets:2300 errors:0 dropped:3 overruns:0 frame:0

host01c01: ????????          TX packets:9 errors:0 dropped:0 overruns:0 carrier:0
???????%          collisions:0 txqueuelen:0 
???????=          RX bytes:296822 (296.8 KB)  TX bytes:734 (734.0 B)
????????
???????&lo        Link encap:Local Loopback  
???????.          inet addr:127.0.0.1  Mask:255.0.0.0
???????)          inet6 addr: ::1/128 Scope:Host
???????3          UP LOOPBACK RUNNING  MTU:65536  Metric:1

host01c01: Error: Timeout to wait for response
```

PR #673, opened 2016-01-27
fix issue 660: lsdocker node -l --logs return too many illegal code

zet809:
The fix for issue #660


chenglch:
I have reviewed this patch and agree to merge.

Issue #639, opened 2016-01-21
[FVT] The postscript confignetwork created ethernet bridge interface successfully, Spanning Tree Protocol is not enabled by default


neo954:
Below are the ``nicdevices`` and ``nictypes`` attributes of node ``c910f02c03p23``. Please notice there is a bridge interface named ``foo``. And ethernet interface ``eth1`` is connected to it. And there is no IP address assigned to interface ``foo``. Which is intentional.

    # lsdef c910f02c03p23 -i nicdevices,nictypes
    Object name: c910f02c03p23
        nicdevices.foo=eth1
        nictypes.eth2=ethernet
        nictypes.eth1=ethernet
        nictypes.foo=bridge

After run postscript ``confignetwork``, on the compute node, interface ``foo`` is created and launched up. But STP is not enabled. Spanning Tree Protocol is a very important part for layer-2 bridge of ethernet. I believe it should be enabled by default.

    [root@c910f02c03p23 ~]# ip addr
    1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN
        link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
        inet 127.0.0.1/8 scope host lo
           valid_lft forever preferred_lft forever
        inet6 ::1/128 scope host
           valid_lft forever preferred_lft forever
    2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UNKNOWN qlen 1000
        link/ether 16:3f:4d:b4:fe:03 brd ff:ff:ff:ff:ff:ff
        inet 10.2.3.23/8 brd 10.255.255.255 scope global dynamic eth0
           valid_lft 40230sec preferred_lft 40230sec
        inet6 fe80::143f:4dff:feb4:fe03/64 scope link
           valid_lft forever preferred_lft forever
    3: eth1: <BROADCAST,MULTICAST> mtu 1500 qdisc pfifo_fast master foo state DOWN qlen 1000
        link/ether 16:3f:4d:b4:fe:04 brd ff:ff:ff:ff:ff:ff
    4: eth2: <BROADCAST,MULTICAST> mtu 1500 qdisc mq state DOWN qlen 1000
        link/ether 00:21:5e:a9:54:56 brd ff:ff:ff:ff:ff:ff
[...]

Issue #632, opened 2016-01-20
[FVT] The postscript confignetwork failed to create tagged VLAN interface with leading zeros in the VLAN id.


neo954:
Below are the ``nicdevices`` and ``nictypes`` attributes of node ``c910f02c03p23``. Please notice, ``bond0.0009`` is defined, with leading zeros as a part of the VLAN id.

In this case, postscript ``confignetwork`` failed to create the network interface of VLAN 9.

    # lsdef c910f02c03p23 -i nicdevices,nictypes
    Object name: c910f02c03p23
        nicdevices.bond0.0009=bond0
        nicdevices.bond0=eth2|eth1
        nicdevices.bond0.4094=bond0
        nictypes.bond0.0009=vlan
        nictypes.eth2=ethernet
        nictypes.bond0=bond
        nictypes.eth1=ethernet
        nictypes.bond0.4094=vlan

Below are the outputs of ``confignetwork``.

    # updatenode c910f02c03p23 confignetwork
    c910f02c03p23: xcatdsklspost: updating VPD database
    c910f02c03p23: xcatdsklspost: downloaded postscripts successfully
    c910f02c03p23: Wed Jan 20 06:46:04 EST 2016 Running postscript: confignetwork
    c910f02c03p23: [I]: NetworkManager is inactive.
    c910f02c03p23: [I]: All valid nics and device list:
    c910f02c03p23: [I]: bond0 eth2@eth1
    c910f02c03p23: [I]: bond0.0009 bond0
    c910f02c03p23: [I]: bond0.4094 bond0
    c910f02c03p23: ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
    c910f02c03p23: configure nic and its device : bond0 eth2@eth1
    c910f02c03p23: [I]: create_bond_interface ifname=bond0 slave_ports=eth2,eth1
    c910f02c03p23: [I]: Pickup xcatnet, "", from NICNETWORKS for interface "bond0".
    c910f02c03p23: [I]: ip link set bond0 down
    c910f02c03p23: [I]: [bond.down] >> 5: bond0: <BROADCAST,MULTICAST,MASTER> mtu 1500 qdisc noqueue state DOWN mode DEFAULT
    c910f02c03p23: [I]: [bond.down] >>     link/ether 16:3f:4d:b4:fe:04 brd ff:ff:ff:ff:ff:ff
[...]

Issue #629, opened 2016-01-20
[FVT] The postscript confignetwork failed to configure link aggregation out of the box


neo954:
Below is the detailed output. Please notice the error message ``[E]: Fail to load kernel module "bonding"``, and ``[I]: >> ./nicutils.sh: line 243: bonding: command not found``.

    # updatenode c910f02c03p22 confignetwork
    c910f02c03p22: xcatdsklspost: updating VPD database
    c910f02c03p22: xcatdsklspost: downloaded postscripts successfully
    c910f02c03p22: Wed Jan 20 01:49:18 EST 2016 Running postscript: confignetwork
    c910f02c03p22: [I]: NetworkManager is inactive.
    c910f02c03p22: [I]: All valid nics and device list:
    c910f02c03p22: [I]: bond0 eth1@eth2
    c910f02c03p22: ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
    c910f02c03p22: configure nic and its device : bond0 eth1@eth2
    c910f02c03p22: [I]: create_bond_interface ifname=bond0 slave_ports=eth1,eth2
    c910f02c03p22: [I]: Pickup xcatnet, "", from NICNETWORKS for interface "bond0".
    c910f02c03p22: [E]: Fail to load kernel module "bonding"
    c910f02c03p22: [I]: >> ./nicutils.sh: line 243: bonding: command not found
    c910f02c03p22: ./nicutils.sh: line 1198: /sys/class/net/bonding_masters: No such file or directory
    c910f02c03p22: [E]: stage 0: Fail to create bond device "bond0"
    c910f02c03p22: [I]: create_persistent_ifcfg ifname=bond0 xcatnet= inattrs=ONBOOT=yes,USERCTL=no,TYPE=Bond,BONDING_MASTER=yes,BONDING_OPTS='mode=802.3ad miimon=100',BOOTPROTO=none,DHCLIENTARGS='-timeout 200'
    c910f02c03p22: ['ifcfg-bond0']
    c910f02c03p22: [I]: >> DEVICE="bond0"
    c910f02c03p22: [I]: >> BOOTPROTO="none"
    c910f02c03p22: [I]: >> NAME="bond0"
    c910f02c03p22: [I]: >> ONBOOT="yes"
    c910f02c03p22: [I]: >> USERCTL="no"
    c910f02c03p22: [I]: >> TYPE="Bond"
[...]

PR #677, opened 2016-01-28
issue 632， issue 629 and issue639

bybai:
#632 failed to create tagged VLAN interface with leading zeros i…
#629 failed to configure link aggregation out of the box
#639 The postscript confignetwork created ethernet bridge interface successfully, Spanning Tree Protocol is not enabled by default


neo954:
Merge it, anyway.

PR #678, opened 2016-01-28
Fix issue667: command log writer process failed to start in xcat docker container 

hu-weihua:
Due to command log operates ``/var/log/xcat/commands.log``, but when directory ``/var/log/xcat/`` didn't exist before command log process starting,  command log process will exit.  

Add code to create ``/var/log/xcat/`` if ``/var/log/xcat/`` didn't exist before.

### DIFF
--- a/xCAT-server/sbin/xcatd
+++ b/xCAT-server/sbin/xcatd
@@ -1128,6 +1128,11 @@ unless ($cmdlog_svrpid){
         exit(0);
     }
 
+    my @tmp =  split "/",$cmdlog_logfile;
+    splice (@tmp, -1);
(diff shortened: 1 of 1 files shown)


xCAT-server/sbin/xcatd
```diff
@@ -1128,6 +1128,11 @@ unless ($cmdlog_svrpid){
         exit(0);
     }
 
+    my @tmp =  split "/",$cmdlog_logfile;
```
immarvin: hi, perl provided subroutine ``dirname`` to get the path of a file, it is widely use in xCAT code

PR #682, opened 2016-01-29
Partial issue #537, do not kill db process if immediate child exists

chenglch:
This change includes the following:

1. Refactor xcatd to make the process model more clear.
2. Replace the signal number with signal name.
3. stop db process after all the immediate children stopped.
4. fix a defect in TableUtils that error to close the table object.
5. enhance the is_pid_exists subroutine in utils.

Close issue: #537


chenglch:
As too much changes in one commit, close this pull request and create a new one 
https://github.com/xcat2/xcat-core/pull/824 
to fix the bug.

Issue #657, opened 2016-01-25
[FVT]"mkdocker command= " could not set commad  for container correctly


junxiawang:
Test OS:rhels 7
Test xCAT version:
lsdef - Version 2.12 (git commit fa2237d141c49e01a630b73c0f0a8e8a69a877fa, built Thu Jan 21 19:30:51 EST 2016)
How  to reproduce:
```
# mkdocker host01c02 image=0d28948ebb14 command=/bin/bash dockerflag="{\"AttachStdin\":true,\"AttachStdout\":true,\"AttachStderr\":true,\"OpenStdin\":true}"
[root@c910f04x30 ~]# lsdocker host01c02
host01c02: 85588e772a78	0d28948ebb14	/bin/bash	2016-01-25 - 10:06:29	running	/host01c02
root@c910f04x30v01:~# docker ps -a
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
85588e772a78        0d28948ebb14        "/bin/startservice.sh"   13 minutes ago      Up 11 minutes                           host01c02
---------------------> the command could not be changed
```

zet809:
Fixed within another issue #732

PR #683, opened 2016-01-29
Fix issue #657: "mkdocker command= " could not set command correctly

zet809:
2 reasons for this issue #657:
1. The image running contains an Entrypoint which can not be overwrote by setting command
2. The lsdocker output the command set by mkdocker rather than the Entrypoint within the image.

The Fix is to modify the lsdocker output since overwrite the Entrypoint may destroy the image.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/docker.pm
+++ b/xCAT-server/lib/xcat/plugins/docker.pm
@@ -488,6 +488,37 @@ sub single_state_engine {
 
 #-------------------------------------------------------
 
+=head3 deal_with_space_in_array_entry
+
+  The function to add '' for entries that have spaces
+  Input:
+        $array: The string array whose entires may have spaces
+  Return:
+        A string that join the entries in input $array with space, 
+        for entries have spaces, they will be put in "'"
+  Usage example:
+
+=cut  
+
+#-------------------------------------------------------
+
+sub deal_with_space_in_array_entry {
+    my $array = shift;
+    my @ret_array = ();
+    push @ret_array, shift @$array;
+    foreach (@$array) {
+        if (/\s/) {
+            push @ret_array, "'$_'";
+        }
+        else {
+            push @ret_array, $_;
+        }
+    }
+    return join(' ', @ret_array);
+}
+
+#-------------------------------------------------------
+
 =head3 parse_docker_list_info
 
   The function to parse the content returned by the lsdocker command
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -488,6 +488,36 @@ sub single_state_engine {
...
+=cut  
+
+#-------------------------------------------------------
+
+sub deal_with_space_in_array_entry {
+    my $array = shift;
+    my @ret_array = ();
+    foreach (@$array) {
+        if (/\s/) {
+            push @ret_array, "'$_'";
```
chenglch: Seems only the first item is command, the other is argument. This output for cmd is not similar to the `docker ps`

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -509,13 +539,12 @@ sub parse_docker_list_info {
...
     if ($flag) {
-        $image = substr($docker_info_hash->{'Image'}, 0, 20);
+        $image = $docker_info_hash->{'Image'};
         $command = $docker_info_hash->{'Command'};
         $created = $docker_info_hash->{'Created'};
         $status = $docker_info_hash->{'Status'};
 
-        $command = substr($command,0, 20);
-        $names = join(',',@{$docker_info_hash->{'Names'}});
+        $names = $docker_info_hash->{'Names'}->[0];
```
chenglch: I'm not sure which version of docker rest api is used. I haven't found the `Names` property.

zet809: For the 1st question, it is GET /containers/json?
For the 2nd, there are 2 scenarios, 1st is have the content length, I will get all the content and then decode them, 2nd is the chunked date, I decode every http pkg.

xCAT-server/lib/xcat/plugins/docker.pm
zet809: No, the docker support to run a groups of commands, such as "date ls pwd 'ip addr show'".
It will show as "date ls pwd 'ip addr" for both 'docker ps' and 'lsdocker <dockerhost>'.

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -523,13 +553,18 @@ sub parse_docker_list_info {
     }
     else {
         $image = $docker_info_hash->{Config}->{'Image'};
-        $command = join(',', @{$docker_info_hash->{Config}->{'Cmd'}});
+        $command = deal_with_space_in_array_entry($docker_info_hash->{Config}->{'Cmd'});
+        if (defined($docker_info_hash->{Config}->{'Entrypoint'})) {
+            $command = deal_with_space_in_array_entry($docker_info_hash->{Config}->{'Entrypoint'});
```
chenglch: If Entrypoint is defined in the docker image, CMD is passed to docker as arguments.

chenglch:
Since @zet809  prepare to leverage Http::Async package to process rest api request, there is no need to fix the json parse issue for chunked data. I also open a issue to locate the error about docker command. For current fix, I agree to merge.

Issue #679, opened 2016-01-29
[Customer] updatenode failed against Ubuntu node


daniceexi:
A customer found the updatenode command failed to run against Ubuntu node, the root cause was the default /bin/sh links to the /bin/dash. xCAT updatenode framework script **xcatdsklspost** is running with **/bin/sh**, then for Ubuntu the **/bin/dash** will be used to run the script. But the script **xcatdsklspost** was designed to run with general shell or bash. 

xCAT has discussion before that all the postscript will only support bash. All the postscript is using **/bin/dash** as sha-bang. So my thought to fix this issue is to change the sha-bang for **xcatdsklspost** to **/bin/dash**. Customer has verified this fix and it worked. xCAT team needs to consider whether there's side-effect of this change.

immarvin:
on ubuntu, "/bin/sh" is a symbol link to "/bin/bash" or "/bin/dash", so I think modify the she-bang to "/bin/bash" is ok. The autotest can cover it

whowutwut:
@immarvin ...A little confused, above @daniceexi  mentions his thought to change the #!/bin/sh -> #!/bin/dash for xcatdkslspost..  but in the pull request, you are changing to #!/bin/bash. 

However, I think we should **not** use #!/bin/sh and use #!/bin/bash like you made... so @daniceexi why do you want to use #!/bin/dash?

daniceexi:
@whowutwut my typo, should use bash in static for the xcatdskless.

daniceexi:
I have verified this change, thanks.

PR #686, opened 2016-01-30
fix defect [Customer] updatenode failed against Ubuntu node #679

immarvin:
replace she-bang "/bin/sh" with "/bin/bash"

### DIFF
--- a/xCAT/postscripts/xcatdsklspost
+++ b/xCAT/postscripts/xcatdsklspost
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
 #####################################################
 #


immarvin:
FVT should carry on sufficient tests on this

daniceexi:
@immarvin Did you cover the following two impacted scenarios in your UT? If no, you need to confirm with FVT that the scenarios has been covered by xcat automation test bucket.

* Updatenode for Ubuntu and rh
* Diskless node for Ubuntu and rh

daniceexi:

immarvin:
on redhat 7.2:

[root@c910f03c05k21 ~]# cat /etc/os-release 
NAME="Red Hat Enterprise Linux Server"
VERSION="7.2 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.2"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.2 Beta (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.2:beta:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.2
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.2 Beta"
[root@c910f03c05k21 ~]# ll /bin/sh          
lrwxrwxrwx. 1 root root 4 Sep 27 22:41 /bin/sh -> bash

on ubuntu:
generally, ``/bin/sh`` is symbol link to either ``bash`` or ``dash``,  xcatdsklspost will work when``/bin/sh`` is a symbol link to ``/bin/bash``,  it won't work when otherwise.

The only thing I need to make sure  is that ``bash`` is in the ubuntu pkglist files for diskful and diskless provision

whowutwut:
I remember that in our instructions we asked the user to change the symlink on Ubuntu from ``/bin/sh`` pointing to ``/bin/dash`` to ``/bin/bash``.  Did we ever change that? 

For 2.12, why not just make a hard change and always make our scripts use ``/bin/bash`` instead of leaving it up to the OS default shell?

daniceexi:
@whowutwut I decision was to use /bin/bash as shebang for all the post/postbootscripts instead of change the default /bin/sh to /bin/bash. The reason was there might be customers want to use /bin/dash instead of /bin/bash. If just arbitrarily changed link to /bin/dash, customers script may fail, even some system scripts may fail.

Issue #630, opened 2016-01-20
[FVT] While creating link aggregation through confignetwork, only the nictypes of the first lower interface is checked


neo954:
Below are the ``nicdevices`` and ``nictypes`` attributes of node ``c910f02c03p23``. At this situation, run postscript ``confignetwork`` will fail, as I mentioned in issue #627.

    # lsdef c910f02c03p23 -i nicdevices,nictypes
    Object name: c910f02c03p23
        nicdevices.bond0=eth2|eth1
        nictypes.bond0=bond

But, if attribute ``nictypes.eth2=ethernet`` is added, postscript ``confignetwork`` will create ``bond0`` successfully. If attribute ``nictypes.eth2=ethernet`` is absent but attribute ``nictypes.eth1=ethernet`` is added, then the script will fail.

For a quick summary, please see the table below.

    --|------------------------|--------
      | nictypes.eth2=         |
    1 | nictypes.eth1=         | Failed
    --|------------------------|--------
      | nictypes.eth2=ethernet |
    2 | nictypes.eth1=         | Succeed
    --|------------------------|--------
      | nictypes.eth2=         |
    3 | nictypes.eth1=ethernet | Failed
    --|------------------------|--------
      | nictypes.eth2=ethernet |
    4 | nictypes.eth1=ethernet | Succeed
    --|------------------------|--------
   
It seems only the ``nictypes`` of the first lower interface is checked and used. I think this behavior is improper and not obey the "Zero one infinity rule". All the lower interfaces should be checked.

For the details of "Zero one infinity rule", please check Wikipedia from the URL below.
https://en.wikipedia.org/wiki/Zero_one_infinity_rule

neo954:
This is the version of xCAT used in the testing.

    Version 2.12 (git commit fbd91febf512026bdada2ab10120957b90bf8831, built Mon Jan 18 15:30:50 EST 2016)

neo954:
Will close this one.

PR #688, opened 2016-02-01
fix 630 only the nictypes of the first lower interface is checked

bybai:
#630 only the nictypes of the first lower interface is checked

### DIFF
--- a/xCAT/postscripts/confignetwork
+++ b/xCAT/postscripts/confignetwork
@@ -249,11 +249,21 @@ function sort_nics_device_order {
         #find nicdevice type as base_nic_type
         base_nic_dev=`echo "$nics_list" |sed -n "${num}p"|awk '{print $2}'`
         if echo "$base_nic_dev"|grep "@" >/dev/null; then
-            temp_base_nic_dev=`echo $base_nic_dev|awk -F@ '{print $1}'`
+            for i in `echo "$base_nic_dev" |sed 's/@/ /g'`
+            do
+                temp_base_nic_type=`find_nic_type "$i"`
+                #all base ni devices types should be ethernet
+                #if one nic type is not ethernet, it is invalid
+                if [ x"$temp_base_nic_type" != "xethernet" ]; then
+                    break
+                fi
+            done
         else
             temp_base_nic_dev=$base_nic_dev
+            temp_base_nic_type=`find_nic_type "temp_base_nic_dev"`
         fi
-        base_nic_type=`find_nic_type "$temp_base_nic_dev"`
+
+        base_nic_type=$temp_base_nic_type
         nic_dev=`echo "$nics_list" |sed -n "${num}p"|awk '{print $1}'`
         nic_dev_type=`find_nic_type "$nic_dev"`
 
--- a/xCAT/postscripts/confignetwork
+++ b/xCAT/postscripts/confignetwork
@@ -252,9 +252,9 @@ function sort_nics_device_order {
             for i in `echo "$base_nic_dev" |sed 's/@/ /g'`
             do
                 temp_base_nic_type=`find_nic_type "$i"`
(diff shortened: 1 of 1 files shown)


xCAT/postscripts/confignetwork
```diff
@@ -249,11 +249,21 @@ function sort_nics_device_order {
         #find nicdevice type as base_nic_type
         base_nic_dev=`echo "$nics_list" |sed -n "${num}p"|awk '{print $2}'`
         if echo "$base_nic_dev"|grep "@" >/dev/null; then
-            temp_base_nic_dev=`echo $base_nic_dev|awk -F@ '{print $1}'`
+            for i in `echo "$base_nic_dev" |sed 's/@/ /g'`
+            do
+                temp_base_nic_type=`find_nic_type "$i"`
+                #all base ni devices types should be ethernet
+                #if one nic type is not ethernet, it is invalid
+                if [ x"$temp_base_nic_type" != "xethernet" ]; then
```
neo954: For link aggregation, bonding multiple infiniband interfaces are useful. And in the HPC environment, this kind of scenario is common.

While bonding ethernet interfaces and infiniband interfaces together is meaningless.

neo954: I think you should check if the ``find_nic_type "$i"`` return the same value in each iteration of the loop.

xCAT/postscripts/confignetwork
```diff
@@ -249,11 +249,21 @@ function sort_nics_device_order {
         #find nicdevice type as base_nic_type
         base_nic_dev=`echo "$nics_list" |sed -n "${num}p"|awk '{print $2}'`
         if echo "$base_nic_dev"|grep "@" >/dev/null; then
-            temp_base_nic_dev=`echo $base_nic_dev|awk -F@ '{print $1}'`
+            for i in `echo "$base_nic_dev" |sed 's/@/ /g'`
+            do
+                temp_base_nic_type=`find_nic_type "$i"`
+                temp_base_nic_type_one=$temp_base_nic_type
+                #all base nic devices types should be the same
+                if [ x"$temp_base_nic_type" != x"$temp_base_nic_type_one" ]; then
```
neo954: You do the variable assignment in line 255, and then do the variable  comparison here. Every time, both variables will have exact same value.

xCAT/postscripts/confignetwork
```diff
@@ -249,11 +249,25 @@ function sort_nics_device_order {
...
+                temp_base_nic_type=`find_nic_type "$i"`
+                if [ x"$temp_base_nic_type_one" = x ]; then
+                    temp_base_nic_type_one=$temp_base_nic_type
+                else
+                    #all base nic devices types should be the same
+                    if [ x"$temp_base_nic_type" != x"$temp_base_nic_type_one" ]; then
+                        break
+                    fi
+                    temp_base_nic_type_one=$temp_base_nic_type
+                fi
```
neo954: Thus, line 262 is no use at all. Right?

xCAT/postscripts/confignetwork
```diff
@@ -249,11 +249,25 @@ function sort_nics_device_order {
...
+            do
+                temp_base_nic_type=`find_nic_type "$i"`
+                if [ x"$temp_base_nic_type_one" = x ]; then
+                    temp_base_nic_type_one=$temp_base_nic_type
+                else
+                    #all base nic devices types should be the same
+                    if [ x"$temp_base_nic_type" != x"$temp_base_nic_type_one" ]; then
+                        break
+                    fi
+                    temp_base_nic_type_one=$temp_base_nic_type
```
neo954: Thus, line 262 is no use. Right?

neo954:
The new patch looks good. I will merge this one.

PR #690, opened 2016-02-01
Improve the content in the xCAT kit related command man page, better formatting on RTD 

whowutwut:
I'm working with some of the kit related commands so I spent some time to improve the format of the .pod files so they would render a little nicer on RTD. 

Also clean up some of the text wording

### DIFF
--- a/docs/source/guides/admin-guides/references/man1/addkit.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/addkit.1.rst
@@ -33,7 +33,7 @@ DESCRIPTION
 
 The \ **addkit**\  command install a kit on the xCAT management node from a kit tarfile or directory, creating xCAT database definitions for kit, kitrepo, kitcomponent.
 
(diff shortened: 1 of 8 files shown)


whowutwut:
@bybai  If you cannot get around to review before your holiday, I will assign to Casandra.

whowutwut:
@bybai, yes this is an issue that I have noticed when we are rendering man pages over on RTD.  @gurevichmark actually working on a fix for it where the ``create_man_pages.py`` script will apply a sed to the .rst files so that the double dash is rendered correctly.  I will let his changes fix the issues for all the man pages.

whowutwut:
@cxhong  Can you review and merge if acceptable.  Regarding the comment @bybai  mentioned, Mark is working on a fix that will address all the pages.

gurevichmark:
Double dash and other man page formatting problems have been fixed by pull requests 703, 710 and 717

Issue #692, opened 2016-02-02
[FVT]While using XCATBYPASS=YES with command, there are unexcepted logs in terminal.


caomengmeng:
Description: While using XCATBYPASS=YES with command, there are unexcepted logs in terminal.

ENV：
```
[root@c910f02c04p08 ~]# cat /etc/*release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.1 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.1"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.1 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.1:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.1
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.1"
Red Hat Enterprise Linux Server release 7.1 (Maipo)
Red Hat Enterprise Linux Server release 7.1 (Maipo)
[root@c910f02c04p08 ~]# rpm -qa | grep -i xCAT
grub2-xcat-2.02-0.16.el7.snap201506090204.noarch
xCAT-client-2.12-snap201601281930.noarch
xCAT-2.12-snap201601281931.ppc64
conserver-xcat-8.1.16-10.ppc64
xCAT-genesis-base-x86_64-2.9-snap201504212134.noarch
perl-xCAT-2.12-snap201601281930.noarch
xCAT-server-2.12-snap201601281930.noarch
xCAT-genesis-scripts-ppc64-2.12-snap201601281931.noarch
xCAT-test-2.12-snap201601281930.noarch
ipmitool-xcat-1.8.11-3.ppc64
xCAT-genesis-base-ppc64-2.11-snap201511160819.noarch
xCAT-buildkit-2.12-snap201601281930.noarch
```

Steps to reproduce::
1. make sure the xcatd is running
[...]

PR #693, opened 2016-02-02
correct While using XCATBYPASS=YES with command, there are unexcepted logs in terminal

penguhyang:
fix issue #692

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -3064,25 +3064,25 @@ sub rscan {
         my $domain=$parser->parse_string($currxml);
         my ($uuid, $node, $vmcpus, $vmmemory, $vmnics, $vmstorage, $arch, $mac, $vmnicnicmodel);
         my @uuidobj = $domain->findnodes("/domain/uuid");
-        if (@uuidobj and defined(@uuidobj->[0])) {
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -3064,25 +3064,25 @@ sub rscan {
         my $domain=$parser->parse_string($currxml);
         my ($uuid, $node, $vmcpus, $vmmemory, $vmnics, $vmstorage, $arch, $mac, $vmnicnicmodel);
         my @uuidobj = $domain->findnodes("/domain/uuid");
-        if (@uuidobj and defined(@uuidobj->[0])) {
-            $uuid = @uuidobj->[0]->to_literal;
+        if (@uuidobj and defined(@uuidobj[0])) {
+            $uuid = @uuidobj[0]->to_literal;
```
whowutwut: I  don't think the fix here to just remove the dereference ``->``? 
Since we are referencing a specific index in the array, we should be using the ``$`` sign... so ==>  ``$uuidobj[0]``

The same comment applies to all the changes in this file.

penguhyang: @whowutwut Thanks for review. Code updated.

whowutwut:
Looks good, I've reviewed the changes, will merge.

PR #695, opened 2016-02-02
[FVT]Automation Case: Add xcatd related cases.

caomengmeng:
@tingtli @daniceexi 
Add 5 xcatd related test cases:
1. Run service xcatd stop; check all process of xcatd have been stopped; service xcatd start; chck all process are started;
2. Run service xcatd restart more than 10 times to make sure xcatd still works well;
3. Stop xcatd, then run xcat command with XCATBYPASS=YES with commands, it works well;
4. If remove the 'root' policy in the policy table, the xcat cmd will fail;
5. If only add permission for certain cmd, check that only this cmd is working but other cmd won't.


daniceexi:
@caomengmeng It's good enough to have the basic testing cases for xcatd. Thanks.

caomengmeng:
@tingtli  Thank you very much for Comments and Great Help!

1. Divide the stop and start case into two: start, stop;
2. Add a loop to run the restart case
Please have a look.

PR #699, opened 2016-02-04
Provide 'txc' mode for packimage

jjohnson42:
tar.xz files will preserve an increasing number of things cpio will
not.  For now, implement for EL7 type oses to explore before making
it more widespread.


daniceexi:
@jjohnson42 Please answer my question that what's the meaning of 'txc'?

daniceexi:
No reply from @jjohnson42. Checking the code and test it with the method name 'txc'.

Issue #651, opened 2016-01-22
[DEV] addkit command does not behave exactly like manpage indicates


whowutwut:
I'm in the process of cleaning up the kit related man pages... while running the commands, I found some inconsistencies and places we could be improving.  

1. The files are extracted to /tmp/tmpkit and are not cleaned up 
```
[root@c910f02c05p05 hpcsw]# addkit essl-5.4.0-0-ppc64le.tar.bz2
Adding Kit essl-5.4.0-0-ppc64le
Kit essl-5.4.0-0-ppc64le was successfully added.

[root@c910f02c05p05 hpcsw]# ls -ltr /install/kits/
total 0
drwxr-xr-x 7 root root 94 Jan 22 15:42 essl-5.4.0-0-ppc64le

[root@c910f02c05p05 hpcsw]# ls -ltr /tmp/tmpkit/
total 0
drwxr-xr-x 7 root root 94 Jan 22 15:42 essl-5.4.0-0-ppc64le
```

2 . Why do we need to extract twice? 

It seems like regardless we are extracting once to /tmp/tmpkit to look at the contents. Once we look at the contents why are we extracting again to deploy the kit? 

3 . The message should be fixed to make more sense... 
```
[root@c910f02c05p05 hpcsw]# addkit -V  -p /tmp essl-5.4.0-0-ppc64le.tar.bz2
Extract Kit /root/hpcsw/essl-5.4.0-0-ppc64le.tar.bz2 to /tmp
Reading kit configuration file /tmp/tmpkit/essl-5.4.0-0-ppc64le/kit.conf
Extract Kit /root/hpcsw/essl-5.4.0-0-ppc64le.tar.bz2 to /tmp
Adding Kit essl-5.4.0-0-ppc64le
...
```

The first msg should be Extract to ``/tmp/tmpkit`` not ``/tmp`` because we are actually putting in ``/tmp/tmpkit``.  

The 2nd extract msg... if we still need it , should point to the final destination...  Although I think we should not need to extract twice 

4 . I think we should prevent -p and -i options to be specified at the same time.  Although it does not extract, we should not allow that combination

5 . I also think we should prevent -i and -V but it's okay to have verbose output, we are not really adding the kit

PR #702, opened 2016-02-05
Remove /tmp/tmpkit dir after addkit and extract kit tar file only once

cxhong:
for issue #651

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kit.pm
+++ b/xCAT-server/lib/xcat/plugins/kit.pm
@@ -104,6 +104,7 @@ sub process_request
         $rc = lskit($request, $callback, $request_command);
     } elsif ($command eq "addkit"){
         $rc = addkit($request, $callback, $request_command);
+        system("rm -rf /tmp/tmpkit/");
     } elsif ($command eq "rmkit"){
         $rc = rmkit($request, $callback, $request_command);
     } elsif ($command eq "lskitcomp"){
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/kit.pm
```diff
@@ -1246,7 +1247,7 @@ sub addkit
             
             if($::VERBOSE){
                 my %rsp;
-                push@{ $rsp{data} }, "Extract Kit $kit to /tmp";
+                push@{ $rsp{data} }, "Extract Kit $kit to /tmp/tmpkit";
```
whowutwut: Can we make the ``/tmp/tmpkit`` a variable?

whowutwut: Line 1252 and 1254 are essentially doing the same thing except for verbose we are adding the ``v`` option in tar.   However, we are not sending back this output to the user on the console.  
```
[root@c910f02c05p03 ESSL]# addkit -i essl-5.4.0-0-ppc64le.tar.bz2 -V 
Extract Kit /root/ESSL/essl-5.4.0-0-ppc64le.tar.bz2 to /tmp/tmpkit
Reading kit configuration file /tmp/tmpkit/essl-5.4.0-0-ppc64le/kit.conf
kitname=essl-5.4.0-0-ppc64le
    description=essl for Linux
    version=5.4.0
    ostype=Linux
```

See the above, there is no tar output.  But then, I don't think we should print out the verbose tar output either because it might be too much.  And we are not doing it today, so the users don't really care that it's not printed out :-)  

So maybe we should just remove the two lines doing the same thing and just print out the msg when -V is selected. 

```
           if($::VERBOSE) {
                my %rsp;
                push@{ $rsp{data} }, "Extract Kit $kit to /tmp/tmpkit";
                xCAT::MsgUtils->message( "I", \%rsp, $callback );
           }
           $rc = system("tar jxf $kit -C /tmp/tmpkit/ --wildcards */kit.conf");
```

whowutwut:
I have reviewed the changes

PR #703, opened 2016-02-08
man page fixes for man8 section

gurevichmark:
Fixes for man pages in man8 section:
1. POD files modified:
    a. Consistent use of italics and bold fonts - bold for fixed options or parameters, italics for user substitutions
    b. Consistent EXAMPLES section - examples are numbered, example commands are shows as command text
    c. Consistent use of [ ] and { } - bracket for options, braces for choices
    d. Consistent RETURN VALUE section - return values and explanation on a single line with dot separation
    e. Misc wording fixed - some EXAMPLE section descriptions added
2. create_man_pages.py modified to fix generated .rst files
   a. Vertical bar fix to not appear as a link in some instances
   b. "- -" option fix to not appear as a single long dash.

### DIFF
--- a/create_man_pages.py
+++ b/create_man_pages.py
@@ -80,7 +80,22 @@ def cleanup_db_man_pages_dir():
     shutil.rmtree("perl-xCAT/pods")
     shutil.rmtree("perl-xCAT/share")
 
-   
+def fix_vertical_bar(rst_file):
+    print "Removing | from file %s, version %s" %(rst_file, man_ver)
+    # Verical bar can not appear with spaces around it, otherwise
+    # it gets displayed as a link in .html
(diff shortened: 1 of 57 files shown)


create_man_pages.py
```diff
@@ -119,6 +134,9 @@ def cleanup_db_man_pages_dir():
                 cmd += " --infile=%s --outfile=%s --title=%s.%s" %(pod_input, rst_output, title, man_ver)
                 print cmd 
                 os.system(cmd)
+		if man_ver == '8':
```
whowutwut: Is this just to do a "test run" for the man8 pages and slowly roll out the changes once we think it's good?

gurevichmark: Yes. I did not want one massive checkin. So my plan is to do man8 first, then in a separate pull request, man1 section changes. I do not think other man sections need these type of changes.

whowutwut:
Thanks for the clean up, I reviewed the changes and will merge.

PR #705, opened 2016-02-09
Refactor MTMS hardware discovery section in documentation

whowutwut:
I've placed the changes here on an internal server: http://10.2.5.2/docs/mtms_discovery/

This is fairly large change broken into 3 separate commits. 
- Moved common installation properties out to the install guide section so we do not need to re-iterate the basic configuration for xCAT in the hardware discovery section
- Modify the configure xCAT section under the hardware discovery section
- Update the MTMS discovery process, intended to clarify some of the issues we hit when doing HW discovery. 

I'd like to get some feedback and once we decide the content is good, I think this should be cherry picked back into the 2.11 release.


whowutwut:
Assigned to @zet809 ... But would also appreciate some others taking a look... @daniceexi  and @cxhong. Fairly new to HW discovery and I tried to write the docs in a way that I understand..

whowutwut:
@zet809  in the documentation, I specified the bmc IP address in the predefined node. This will set it statically. .What happens if the user does not want the BMC IP address to set static and continue to leave it DHCP?

daniceexi:
how about not set up for BMC but running bmcsetup.

docs/source/guides/admin-guides/manage_clusters/ppc64le/configure/site.rst
```diff
@@ -0,0 +1,36 @@
...
+   *For more information on the keywords, see the DHCP ATTRIBUTES in the :doc:`site </guides/admin-guides/references/man5/site.5>` table.*
+
+   If the fields are not set or need to be changed, use the xCAT ``chdef`` command: ::
+
+      chdef -t site domain="domain_string"
+      chdef -t site fowarders="forwarders"
+      chdef -t site master="xcat_master_ip"
+      chdef -t site nameservers="nameserver1,nameserver2,etc"
+
+.. [#] The value of the ``master`` attribute in the site table can be set to either the Management Node IP or Service Node IP.
```
zet809: The ''master'' shall be set as the IP of Management Node facing to the Compute Node.

zet809:
@whowutwut One concern is that we may shall not drop the origin process that use the DHCP assigned BMC ip address directly.

whowutwut:
@zet809  Thanks!  

From what I understand, there are 2 scenarios that we should cover. 
1. BMC gets a DHCP address over the dynamic range and does ``bmcsetup`` to configure a static IP address for BMC 
2. BMC gets a DHCP address over the dynamic range.  Admin discovers the node and gets the BMC mac address so that they can leave BMC in dhcp, but xCAT has a node definition for the BMC where it's serving the IP address using DHCP server. 

Is this the two scenarios you are referring to?

zet809:
@whowutwut 
For scenario 1: 
    Use *bmcsetup* to configure static ip address within Dynamic range or out out Dynamic range are different, right?
For scenario 2:
    You are right, we may also need to think over this kind of using. We need to modify our code logic in order to support this use case.

zet809:
@whowutwut Thanks very much for the modification, that is what I want.

whowutwut:
@zet809 still working on some changes. Should be done soon hopefully

whowutwut:
@zet809 , I finished the section and updated on our internal server here: http://10.2.5.2/docs/mtms_new/.  

The bulk of the changes is under the MTMS section here:
http://10.2.5.2/docs/mtms_new/guides/admin-guides/manage_clusters/ppc64le/discovery/mtms/index.html

Does this cover the changes we described above?

daniceexi:
@whowutwut 
In your discovery page, there are two steps to find the **predefined node** and 'discovered node**. It's a little confusing, maybe it's easier to understand if we name the **predefined node** to be **compute node** and name the **discovered node** to be **bmc node**, what do you think?

zet809:
@whowutwut It seems the scenario that set the Dynamic BMC ip as static is missed from the modification.

whowutwut:
@daniceexi, I agree. the predefined and discovered node concept is a but confusing.. let me try to change the terminology and see how the document looks. 

@zet809  For the dynamic BMC IP to static,does step 5 cover it?  (I need to fix the IP address to show up as bold)  
image

zet809:
@whowutwut Sorry, victor, what I mean is the scenario that the user use the assigned dynamic IP address directly. 
The steps may include:
1. bmcdiscover  BMC:   bmcdiscover -s nmap --range 50.0.100.1-100 -z 
2. add chain="runcmd=bmcsetup" for the discovered BMC node
3. power on the BMC node
4. after discovery process done, all info will be updated to the BMC node and it will be the final object node.

whowutwut:
@zet809 Updated the MTMS documentation to cover the two scenarios. 

Here's a test machine to view if needed...
http://10.2.5.2/docs/mtms/guides/admin-guides/manage_clusters/ppc64le/discovery/mtms/index.html

daniceexi:
@whowutwut Thanks for the change, it looks more clear.

Two minor comments:

* I'd like to be clear to distinguish the bmc node and compute node. The following line can be ** Define the discovered-bmc node to xCAT.

  Overview part: 2. Define the **discovered-node** to xCAT

* In the approach of using different IP for bmc. The static ip range might be in the same network with the dynamic ip range. Otherwise the network cfg will be complicated on xCAT MN since it needs handle two networks for BMC.

zet809:
@whowutwut The modification looks good to me, just 1 more comments:
 In "Set static BMC IP using dhcp provided IP address" page step 7 and step8, the node shall be cn01 rather than node-8247-22l-10112ca.

whowutwut:
@daniceexi  @zet809  Thanks for the comments!
I updated based on your suggestions in the last commit. 

I did not refresh the text on the server

zet809:
@whowutwut Looks good to me, no more comment.

daniceexi:
Then we agree to merge the change.

PR #709, opened 2016-02-10
Add "Install HPC software Kits" section

cxhong:
Add a section to describe how to install each HPC Software Kits to osimage.


whowutwut:
Looks good, I think we should cherry pick this one back into 2.11 branch

PR #710, opened 2016-02-11
man1 chages for commands a-n

gurevichmark:
Fixes for man pages in man1 section (commands a-n):
1. POD files modified:
a. Consistent use of italics and bold fonts - bold for fixed options or parameters, italics for user substitutions
b. Consistent EXAMPLES section - examples are numbered, example commands are shows as command text
c. Consistent use of [ ] and { } - bracket for options, braces for choices
d. Various formatting fixes.
2. create_man_pages.py modified to fix generated .rst files
a. Vertical bar fix to not appear as a link in some instances
b. "- -" option fix to not appear as a single long dash.

### DIFF
--- a/create_man_pages.py
+++ b/create_man_pages.py
@@ -91,7 +91,7 @@ def fix_double_dash(rst_file):
     print "Fixing -- in file %s, version %s" %(rst_file, man_ver)
     # -- gets displayed in .html as a sinle long dash, need to insert
     # a non bold space between 2 dashes
-    sed_cmd = "sed 's/--/-\*\*\\\ \*\*-/g' %s.sed1 > %s" %(rst_file, rst_file)
+    sed_cmd = "sed '/\*\*/s/--/-\*\*\\\ \*\*-/g' %s.sed1 > %s" %(rst_file, rst_file)
     os.system(sed_cmd)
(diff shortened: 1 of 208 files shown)


create_man_pages.py
```diff
@@ -91,7 +91,7 @@ def fix_double_dash(rst_file):
     print "Fixing -- in file %s, version %s" %(rst_file, man_ver)
     # -- gets displayed in .html as a sinle long dash, need to insert
     # a non bold space between 2 dashes
-    sed_cmd = "sed 's/--/-\*\*\\\ \*\*-/g' %s.sed1 > %s" %(rst_file, rst_file)
+    sed_cmd = "sed '/\*\*/s/--/-\*\*\\\ \*\*-/g' %s.sed1 > %s" %(rst_file, rst_file)
```
whowutwut: Does this change require the man8 pages to be re committed?

gurevichmark: No, this just makes sure that only lines containing "**" get changes by sed. In man8 all lines containing "--" were also lines with "**", but in man1 there were some lines with "--", but not containing "**". Those lines do not need to be changed.

whowutwut:
Thanks for helping clean up the man pages!

Issue #663, opened 2016-01-25
[FVT]xCAT in docker :genimage failed 


junxiawang:

immarvin:
since this is a bug report, please attach the details on the bug description

immarvin:
Please continue your test with the workaround:

run genimage export the generated diskless osimage with ``imgexport`` on other MN, then import the image to docker container with ``imgimport``

I will keep looking into this problem

immarvin:
````

Selecting previously unselected package vim-runtime.
Preparing to unpack .../vim-runtime_2%3a7.4.052-1ubuntu3_all.deb ...
Leaving 'diversion of /usr/share/vim/vim74/doc/help.txt to /usr/share/vim/vim74/doc/help.txt.vim-tiny by vim-runtime'
Leaving 'diversion of /usr/share/vim/vim74/doc/tags to /usr/share/vim/vim74/doc/tags.vim-tiny by vim-runtime'
Unpacking vim-runtime (2:7.4.052-1ubuntu3) ...
Selecting previously unselected package vim.
Preparing to unpack .../vim_2%3a7.4.052-1ubuntu3_amd64.deb ...
Unpacking vim (2:7.4.052-1ubuntu3) ...
Selecting previously unselected package ssh-import-id.
Preparing to unpack .../ssh-import-id_3.21-0ubuntu1_all.deb ...
Unpacking ssh-import-id (3.21-0ubuntu1) ...
Processing triggers for ureadahead (0.100.0-16) ...
Processing triggers for mime-support (3.54ubuntu1) ...
Setting up udev (204-5ubuntu20.13) ...
initctl: Unable to connect to Upstart: Failed to connect to socket /com/ubuntu/upstart: Connection refused
 * udev requires a mounted sysfs, not started
   ...fail!
invoke-rc.d: initscript udev, action "restart" failed.
dpkg: error processing package udev (--configure):
 subprocess installed post-installation script returned error exit status 1
dpkg: dependency problems prevent configuration of initramfs-tools:
 initramfs-tools depends on udev (>= 147~-5); however:
  Package udev is not configured yet.

dpkg: error processing package initramfs-tools (--configure):
 dependency problems - leaving unconfigured
dpkg: dependency problems prevent configuration of plymouth:
 plymouth depends on initramfs-tools; however:
  Package initramfs-tools is not configured yet.
 plymouth depends on udev (>= 166-0ubuntu4); however:
  Package udev is not configured yet.

[...]

PR #714, opened 2016-02-16
fix defect [FVT]xCAT in docker :genimage failed #663

immarvin:
fix defect [FVT]xCAT in docker :genimage failed #663:
(1). mount /proc /sys read-only to eliminate errors during package installation in the chrooted directory
(2). fake invoke.rc.d to eliminate errors during package installation in the chrooted directory

### DIFF
--- a/xCAT-server/share/xcat/netboot/ubuntu/genimage
+++ b/xCAT-server/share/xcat/netboot/ubuntu/genimage
@@ -427,12 +427,11 @@ unless ($onlyinitrd) {
             $envlist = join(',', @{$pkg_hash{$pass}{ENVLIST}});
         }
         
-        fake_invoke_rc_d();
+        fake_policy_rc_d();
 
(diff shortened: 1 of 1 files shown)


immarvin:
hi @daniceexi @whowutwut ,any comments?

immarvin:
``Is there same issues for rh and sles?``
I am not sure, we did not build xcat docker images from a redhat/sles base image yet.

``Do we need to handle the /dev here too?``
The genimage on ubuntu works without devfs mounted, it is also ok with devfs mounted readonly, do we need to add devfs in chroot_mount?

daniceexi:
@immarvin I meant we need to do the similar thing for rh and sles. At least for rh, we did see the issue caused by the missing of /proc, but I cannot remember why the code for mounting was removed.

And please do the mount for /dev too.

immarvin:
mounted devfs under rootimg during genimage, code updated @daniceexi

daniceexi:

Issue #510, opened 2015-12-02
[FVT] SLES 12.1 Power8 LE: Error: qemu does not support SGA


junxiawang:
1.download the latest build
2.env:
```
910f05c33:/ # cat /etc/*-release
NAME="SLES"
VERSION="12-SP1"
VERSION_ID="12.1"
PRETTY_NAME="SUSE Linux Enterprise Server 12 SP1"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:12:sp1"
SUSE Linux Enterprise Server 12 (ppc64le)
VERSION = 12
PATCHLEVEL = 1
# This file is deprecated and will be removed in a future service pack or release.
# Please check /etc/os-release for details about this release.
c910f05c33:/ # 
```
3.rpower p8levm boot
```
c910f05c33:/ # rpower c910f03fsp03v120 boot
c910f03fsp03v120: Error: internal error: qemu does not support SGA
```

whowutwut:
Hi @junxiawang 

 I went onto ``c910f05c33`` and took a look, somethings were not right....   There was no entry in kvm_nodetata for the VM.. .because the ``useserial`` attribute causes that message...    How did you create this VM? 

```
c910f05c33:~ # rpower c910f03fsp03v120 boot
c910f03fsp03v120: Error: internal error: qemu does not support SGA
c910f05c33:~ # rpower c910f03fsp03v120 stat
c910f03fsp03v120: off
c910f05c33:~ # tabdump kvm_nodedata
#node,xml,comments,disable
c910f05c33:~ #
```

So I removed the qcow2 file on the hypervisor and re-created with mkvm....

```
c910f05c33:~ # mkvm c910f03fsp03v120 -s 20G
c910f03fsp03v120: Error: Storage creation request conflicts with existing file(s)
c910f05c33:~ # ssh 10.3.9.1

[root@node-32030901 ~]# rm /var/lib/libvirt/images/c910f03fsp03v120.sda.qcow2
rm: remove regular file '/var/lib/libvirt/images/c910f03fsp03v120.sda.qcow2'? y
```

Now re-create the VM

```
c910f05c33:~ # mkvm c910f03fsp03v120 -s 20G
```

Now we see kvm_nodedata have an entry.. 
```
c910f05c33:~ # tabdump kvm_nodedata
#node,xml,comments,disable
"c910f03fsp03v120","<domain type=""kvm"">
[...]

Issue #184, opened 2015-09-16
  [FVT] Virtual machine failed to power on with PowerKVM 3.1.0 and latest xCAT version 2.11


neo954:
The latest PowerKVM build used for this testing is 3.1.0.0-31 Alpha. The xCAT used in this testing is 2.11-local20150916, build from the latest source code locally.

    # lsdef c910f03c29k03
    Object name: c910f03c29k03
        arch=ppc64el
        groups=all
        id=3
        mac=42:07:0a:03:1d:03|42:66:0a:03:1d:03!*NOIP*|42:dd:0a:03:1d:03!*NOIP*|42:d6:0a:03:1d:03!*NOIP*
        mgt=kvm
        netboot=grub2
        postbootscripts=otherpkgs
        postscripts=syslog,remoteshell,syncfiles
        serialport=0
        serialspeed=115200
        status=booting
        statustime=09-16-2015 07:52:28
        vmcpus=2
        vmhost=c910f03c29
        vmmemory=6144
        vmnics=brenx97,brenP4p1s0f1,brenP4p1s0f2,brenP4p1s0f3
        vmstorage=phy:/dev/mapper/vdiskvg-vdisk00n03

While trying to power on the virtual machine. The following error messages were shown.

    # rpower c910f03c29k03 on
    Error: kvm plugin bug, pid 12218, process description: 'xCATd SSL: rpower to c910f03c29k03 for root@localhost: kvm instance' with error 'libvirt error code: 1, message: internal error: qemu does not support SGA
' while trying to fulfill request for the following nodes: c910f03c29k03

I striped the virtual machine definition xml file through `tabdump kvm_nodedata`. Here is what I got.

    <domain type="kvm">
      <name>c910f03c29k03</name>
      <clock offset="utc" />
      <devices>
        <console type="pty">
          <target port="1" />
        </console>
        <disk device="disk" type="block">
          <source dev="/dev/mapper/vdiskvg-vdisk00n03" />
          <target bus="scsi" dev="sda" />
        </disk>
        <disk device="cdrom" type="file">
          <target dev="sdd" />
        </disk>
        <graphics autoport="yes" listen="0.0.0.0" password="UrSi7F5T07jMDZPKV8qr" type="vnc" />
        <input bus="usb" type="tablet" />
        <interface type="bridge">
          <mac address="42:07:0a:03:1d:03" />
          <model type="virtio" />
          <source bridge="brenx97" />
        </interface>
        <interface type="bridge">
[...]

PR #718, opened 2016-02-17
Fix issue with SGA when creating diskless VMs on PowerKVM

whowutwut:
When creating a diskless VM, and doing rpower <vm> on, it comes back with ``Error: internal error: qemu does not support SGA``.  The entry useserial is causing this problem. 

This is on power kvm , so we have a check in the build_oshash() to NOT set useserial.  However, when I debugged this, the code is not setting anything in the $confdata.  So I added a condition to set this in dohyp() for the rpower command.  

Address Issue #510 

I think this should also be fixed in 2.11...

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -3873,7 +3873,7 @@ sub dohyp {
             return 1, "General error establishing libvirt communication";
         }
     }
-    if (($command eq 'mkvm' or $command eq 'chvm') and $hypconn) {
+    if (($command eq 'mkvm' or $command eq 'chvm' or $command eq 'rpower') and $hypconn) {
         my $nodeinfo = $hypconn->get_node_info();
         if (exists($nodeinfo->{model})) {
             $confdata->{$hyp}->{cpumodel} = $nodeinfo->{model};


whowutwut:
@immarvin  can you take a look at this change and let me know if this is the right thing to do?

immarvin:
hi Victor, 

have you recreated the issue? I recalled that this defect has been fixed before, see https://github.com/xcat2/xcat-core/issues/184

For this defect https://github.com/xcat2/xcat-core/issues/510, I guess Junxia just upgraded xCAT and the xml definition of c910f03fsp03v120 was created by the old xCAT before upgrade, when there has been  a xml definition for c910f03fsp03v120  in kvm_nodedata, xCAT won't update it on rpower or mkvm. If this is the case, I do not think we need to fix it

whowutwut:
@immarvin Yes, I am able to re-create this issue using xCAT 2.11 GA code.  I'm looking at the issue you specify and I'm not sure why it works there... 

The scenario that I'm running is the following: 
1. From a 2.11 xCAT mgmt node, provision a VM 
2. From the VM, re-define the hypervisor node definition, install xCAT 2.11 on the VM and deploy more VMs from this VM... 

I don't know if this does something different.   In the code for ``useserial`` we are looking at the cpumodel attribute...
```
    my $hypcpumodel = $confdata->{ $confdata->{vm}->{$node}->[0]->{host} }->{cpumodel};
    unless (defined($hypcpumodel) and $hypcpumodel eq "ppc64le") {
        $rethash{bios}->{useserial} = 'yes';
    }
```

The only place this is set is ``sub dohyp()``.. and it's not set for the rpower command so when I ran the code through the perl debugger, the {cpumodel} is undef and so we include useserial into the XML and hit the SGA problem....   So I'm not sure how it worked before....  any idea?

immarvin:
Hi Victor,
This is indeed a bug for VMs without storage. Your fix looks well. 
But I am not sure why the VM xml definitions are not created on "mkvm", is it ok that the VMs running on hypervisor have no xml definition in xCAT kvm_nodedata table?

whowutwut:
Hi Yang Song,  if I remember right from last week, it looks like the VMs (diskless and diskful) do not store anything until rpower is issued.  (Maybe i'm mistaken for diskful). ..  For diskless, it's not until we are doing the rpower that the xml information is created.

immarvin:
agree to merge

PR #726, opened 2016-02-18
Additional small man page fixes

gurevichmark:
Remove "print" from create_man_pages script and fix small spelling and formatting errors.

### DIFF
--- a/create_man_pages.py
+++ b/create_man_pages.py
@@ -81,14 +81,12 @@ def cleanup_db_man_pages_dir():
     shutil.rmtree("perl-xCAT/share")
 
 def fix_vertical_bar(rst_file):
-    print "Removing | from file %s, version %s" %(rst_file, man_ver)
     # Verical bar can not appear with spaces around it, otherwise
     # it gets displayed as a link in .html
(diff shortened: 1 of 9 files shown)


create_man_pages.py
```diff
@@ -81,14 +81,12 @@ def cleanup_db_man_pages_dir():
     shutil.rmtree("perl-xCAT/share")
 
 def fix_vertical_bar(rst_file):
-    print "Removing | from file %s, version %s" %(rst_file, man_ver)
```
whowutwut: Just some thoughts... I didn't add it into this script initially, but we could add a ``--verbose`` option if you think the debug are useful to keep in... 

At the above OptionParser, add the option.... 
```
parser.add_option("--verbose", dest="VERBOSE", help="Print verbose output from the script") 
```

Then you can test...

```
if (options.VERBOSE):
     print "Removing.... " 
```

gurevichmark: Not sure if it is worth it. Those prints were just useful for UT.

whowutwut:
I have reviewed the changes

PR #733, opened 2016-02-22
NextScale NX360 M5 support: add description of -u and -p for rsetboot;patch pasu with the latest version

immarvin:


immarvin:
hi @daniceexi @whowutwut , would you please review the code ASAP, so that we can make a 2.10.1 snapshot  build  and perform FVT.

immarvin:
updated

immarvin:
@daniceexi ,seems [-V| --verbose] was not processed in rsetboot code. But I think we can keep it there to reserve for future enhancement.

daniceexi:
@immarvin If rsetboot does not support -V, we should remove it from the usage message. It's not reasonable to have a unimplemented flag in the usage.

immarvin:
-V|--verbose removed

daniceexi:

PR #734, opened 2016-02-22
Missed checkin of rinstall.rst file

gurevichmark:
Somehow previous man page fixes checkins missed this .rst file which just fixes spelling.


whowutwut:
@gurevichmark Are there other .rst files missed? Maybe we can use this pull request to sync up all the missing rst files in the master branch.

gurevichmark:
Seems to be the only one. I extracted the latest xcat-core and ran **create_man_pages.py**. Out of all the generated .rst files, rinstall was the only one out of sync. Still can not explain, however, how it was missed in the last man files pull request.

PR #737, opened 2016-02-25
Refine the docker instance lifecycle management code with HTTP::Async

zet809:


xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -1172,74 +1008,11 @@ sub sendreq {
     elsif (not ref($http_req)) {
```
chenglch: As  $request->protocol is called in your `genreq` function as follows, I think $request can not be `undef` or `not ref` here.

    my $url = "https://".$dockerhost->{name}.":".$dockerhost->{port}.$api;
    my $request = HTTP::Request->new($method, $url, $header, $content);
    $request->protocol('HTTP/1.1');
    return $request;

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -275,212 +271,99 @@ sub http_state_code_info {
...
         if ($content ne '') {
             $msg[0]->[1] = "$content";
         }
+        elsif ($data->message ne '') {
+            $msg[0]->[1] = $data->message;
+        }
+    }
+
+    if ($data->header("connection") =~ /close/) { 
+        $http_requests_in_progress--;
```
chenglch: Is this the only case to decrease the `$http_requests_in_progress` number here? If timeout happens for the current async id, does the connection header contains `close`?

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -275,212 +271,99 @@ sub http_state_code_info {
...
-                while ($length) {
-                    my $content_hash = decode_json (shift @chunked_array);
-                    if (defined($content_hash->{'State'}->{'Status'})) {
-                        $node_state = $content_hash->{'State'}->{'Status'};
-                        last;
+        if ($data->is_success) {
+            if ($content_type =~ /json/i) {
+                my $node_state = $content_hash->{'State'}->{'Status'};
+                if (defined($node_state)) {
+                    if ($nodelisttab) {
```
chenglch: Why do you init `$nodelisttab` in `process_request` but check and use this variable here?

zet809: The reason is that the node status attribute will only be updated when running 'rpower <node> state'

chenglch: I understand, but why not to init table object here to save the connection buffer for the database or DB process.

xCAT-server/lib/xcat/plugins/docker.pm
zet809: You are right that if http return failed, the connection also need to be closed.

chenglch:
on line 776:

    if (!defined($hosttab) or !defined($nodetypetab) or !defined($mactab) or !defined($vmtab)) { 
             $callback->({error=>["Open table 'nodetype', 'hosts' or 'mac' failed"], errorcode=>1});
             return;
         }

same comments as the previous pull request, you'd better create and check the tab object one by one.

Another issue in the whole file is that you just init the tab object, but no any code to close the tab object.

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -275,213 +269,103 @@ sub http_state_code_info {
...
+                $msg[0] = [1, "The content type: $content_type is unable to be parsed."];
             }
         }
-        elsif ($res->code eq '404') {
+        elsif ($data->code eq '404') {
+            if (!$nodelisttab) {
+                $nodelisttab = xCAT::Table->new('nodelist');
+            }
             if ($nodelisttab) {
                 $nodelisttab->setNodeAttribs($node, {status=>''});
```
chenglch: Seems you clear the status column in database here, how about the other http error status code like 503?

chenglch:
Looks good to me.

PR #738, opened 2016-02-25
Fix issue #573 in the 2.10 branch, error message "xcatd: possible BUG encountered by xCAT UDP service"

whowutwut:
Issue #573, Fix the install error message that appears in the xCAT 2.10 branch

```
xcatd: possible BUG encountered by xCAT UDP service
```


daniceexi:
The change looks good for 2.10.

Issue #729, opened 2016-02-22
[FVT]could not get database value as soon as  copycds finished in sles11.4 using  command "ls* "


junxiawang:
ENV:
```
c910f04x27v05:/ # uname -a
Linux c910f04x27v05 3.0.101-63-default #1 SMP Tue Jun 23 16:02:31 UTC 2015 (4b89d0c) x86_64 x86_64 x86_64 GNU/Linux
c910f04x27v05:/ # cat /etc/*-release
LSB_VERSION="core-2.0-noarch:core-3.2-noarch:core-4.0-noarch:core-2.0-x86_64:core-3.2-x86_64:core-4.0-x86_64"
PRETTY_NAME="SUSE Linux Enterprise Server 11 SP4"
Version:
c910f04x27v05:/ # lsdef -v
lsdef - Version 2.12 (git commit add2c075e4c8eebadf3ae1976c0f38172a632b5d, built Fri Feb 19 15:30:39 EST 2016)
```
How to reproduce:
```
c910f04x27v05:/ # copycds SLES-11-SP4-DVD-x86_64-GM-DVD1.iso 
Copying media to /install/sles11.4/x86_64/1
Media copy operation successful
```

As soon as copycds finished I run: 
```
c910f04x27v05:~ # lsdef c910f04x27v06 -i os
c910f04x27v05:~ # echo $?
0 --------------------------> could not get any value and the command return value is 0 
```

immarvin:
copycds just create the osimage objects, it does not set any node attributes, You should try to get "os" attribute of node object after ``nodeset``

junxiawang:
yes ,my means is I could not get any thing using lsdef command as soon as copycds finished .

junxiawang:
Node definition has been defined before lsdef .I could not get any value using lsdef as soon as copycds finished .But about 5 seconds I could get the right value using lsdef.

junxiawang:
```
c910f04x27v05:/opt/xcat/bin # lsxcatd -v
c910f04x27v05:/opt/xcat/bin # echo $?
0
c910f04x27v05:/opt/xcat/bin # 
```
lsxcatd -v still could not get get as soon as copycds finished.
In my point as soon as copycds finished ,we could not get database information using ls* command.

hu-weihua:
@chenglch, I think it's something related about what you investigate recently. Could you take a look at this issue at the same time? thanks

junxiawang:
After code merged some other errors:
```
[...]

PR #742, opened 2016-02-26
Fix issue #729: xcat do not print message under heavy load scenario

chenglch:
The root cause of this issue is the select time out when polling the
pipe socket. When timeout occurs, child process exit before printing
the message at the parent process side.

close-issue: #729


daniceexi:
The change looks good, I am OK for the merge.

Issue #712, opened 2016-02-15
[DEV]the code logic of "updateflag.awk" and related logic in "xcatd" need to be refined 


immarvin:
1. the following code logic of "updateflag.awk" should be refined to prevent infinite loop on failure. 
````
cat /xcatpost/updateflag.awk
#!/usr/bin/awk -f
# IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html

BEGIN {
    xcatdhost = ARGV[1]
    xcatdport = ARGV[2]
    flag = ARGV[3]

        if (!flag) flag = "next"

        ns = "/inet/tcp/0/" ARGV[1] "/" xcatdport

        while(1) {
                if((ns |& getline) > 0)
                        print $0 | "logger -t xcat -p local4.info"
        else {
#>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
            print "Retrying flag update" | "logger -t xcat -p local4.info"
            close(ns)
            system("sleep 10")

#<<<<<<<<<<<<<<<<<<<<should be some thing like MAX_RETRY here and it should be logged
#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
        }

                if($0 == "ready")
                        print flag |& ns
                if($0 == "done")
                        break
        }

        close(ns)
        exit 0
}
````

2. in  the ``/opt/xcat/sbin/xcatd``, the connection created by updateflag.awk from each node should better be logged on MN, especially, it should be easy to find out the connections from the nodes which are deemed to be invalid from the log.
````
cat /opt/xcat/sbin/xcatd
    ...
    352         next unless $conn = $socket->accept;
    ...
    367         my $client_name;
    368         my $client_aliases;
    369         my @clients;
    370         if ($inet6support) {
    371             ($client_name,$client_aliases) = gethostbyaddr($conn->peeraddr,AF_INET6);
    372             unless ($client_name) { ($client_name,$client_aliases) =  gethostbyaddr($conn->peeraddr,AF_I
        NET); }
    373         } else {
    374             ($client_name,$client_aliases) = gethostbyaddr($conn->peeraddr,AF_INET);
    375         }
    376 
    377         $clients[0] = $client_name;
    378         if ($client_aliases) {
[...]

PR #744, opened 2016-02-26
Detect Error case in the xcatd and avoid loop in the updateflag.awk

cxhong:
Check description of issue #712
1) add trace message if didn't get hostname from gethostbyaddr() function in the xcatd
2) add trace message if didn't find this node in the nodels in the xcatd
3) avoid loop in the updateflag.awk

### DIFF
--- a/xCAT-server/sbin/xcatd
+++ b/xCAT-server/sbin/xcatd
@@ -377,6 +377,9 @@ sub do_installm_service {
         } else {
             ($client_name,$client_aliases) = gethostbyaddr($conn->peeraddr,AF_INET);
         }
+        unless ($client_name) {
+            xCAT::MsgUtils->message("S","xcatd unable to gethostbyaddr");
+        }
 
         $clients[0] = $client_name;
         if ($client_aliases) {
@@ -403,7 +406,11 @@ sub do_installm_service {
             if ($node) {  # Means the source isn't valid
                 $validclient=1;
                 last;
+            } else {
+                xCAT::MsgUtils->message("S","Invalid client $client, please check nodels and it's name resolution");
+
             }
+            
         }
 
         unless ($validclient) {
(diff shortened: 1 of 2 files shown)


xCAT-server/sbin/xcatd
```diff
@@ -377,6 +377,9 @@ sub do_installm_service {
         } else {
             ($client_name,$client_aliases) = gethostbyaddr($conn->peeraddr,AF_INET);
         }
+        unless ($client_name) {
+            xCAT::MsgUtils->message("S","xcatd unable to gethostbyaddr");
```
immarvin: suggest the information to be "xcatd received a connection request from unknown host with ip address \"$conn->peeraddr\", please check whether the reverse name resolution for it works correctly"

xCAT-server/sbin/xcatd
```diff
@@ -403,7 +406,11 @@ sub do_installm_service {
             if ($node) {  # Means the source isn't valid
                 $validclient=1;
                 last;
+            } else {
+                xCAT::MsgUtils->message("S","Invalid client $client, please check nodels and it's name resolution");
```
immarvin: the message should better be "xcatd received a connection request from $client, which can not be found in xCAT nodelist table. The connection request will be ignored"

xCAT/postscripts/updateflag.awk
```diff
@@ -6,27 +6,31 @@ BEGIN {
...
-		if($0 == "ready")
-			print flag |& ns
-		if($0 == "done")
-			break
-	}
+        if($0 == "ready")
+            print flag |& ns
+        if($0 == "done")
+            break
+        if(loop > 10)
```
immarvin: there should better be a message here, something like
````
             print "flag update failed" | "logger -t xcat -p local4.info"
````

cxhong:
Thanks, @immarvin .

I modified the messages

immarvin:
looks good to me, agree to merge

PR #745, opened 2016-02-26
Add bypass flag description to xdcp and xdsh commands

gurevichmark:
Added -B | --bypass flag description to xdcp and xdsh commands man pages.
Also fixed some formatting problems.


whowutwut:
I reviewed, thanks

PR #746, opened 2016-02-29
add sound device into vm xml

### LINKED ISSUE #723: [Customer] The sound device should be included in vm xml
Hello,
 
  I am working on a VDI project using KVM for hosting the virtual machines and SPICE to provide GUI/Sound access to the VMs.  Between the 2.10 and 2.11 releases, the kvm.pm module had sound support removed.  See:
 
https://github.com/xcat2/xcat-core/commit/c5f1075aeed48907b3a9f4cc9302f723f08d4eba
https://github.com/xcat2/xcat-core/commit/fa3f96a94029d659116cd24c2307137393fafbf3
 
Is there a plan to re-introduce sound support?  Were there any technical reasons for removing the support?
 
Thank you,
 
Michael Moore
HPC Systems Engineer
Nuance Communications
Michael.Moore@Nuance.com
 -----------------------------------

penguhyang:
fix #723

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -888,6 +888,8 @@ sub build_xmldesc {
     }
     if (defined($hypcpumodel) and $hypcpumodel eq 'ppc64') {
         $xtree{devices}->{emulator}->{content} = "/usr/bin/qemu-system-ppc64";
+    } else {
+        $xtree{devices}->{sound}->{model}='ac97';
     }
 
     $xtree{devices}->{console}->{type} = 'pty';


daniceexi:
@immarvin There must have reason that we removed the ac97 driver in the kvm xml definition, then is it reasonable that just take it back directly?

immarvin:
from my reply to xcat-user mail list:

hi,

sorry for inconvenient.

the sound device in the kvm xml template was removed when we fixed a bug which was caused by the devices/features in xml which are not supported by cpu model or machine type. 

We removed some devices and features we thought that might not be used, the sound device is one of them, obviously, just as you pointed out, this is not the case.

we will add it back, opened a ticket to trace this https://github.com/xcat2/xcat-core/issues/723. would you please cpu model or arch you need?

thanks

daniceexi:
@immarvin Could you point out the issue and commit ID for fixing the issue you mentioned that certain devices cannot be supported by cpu model?

immarvin:
https://github.com/xcat2/xcat-core/pull/311
https://github.com/xcat2/xcat-core/issues/184

daniceexi:
@immarvin Thanks for the explaination. 

I am OK with the change adn agree to merge.

PR #750, opened 2016-02-29
Add pull image for mkdocker

zet809:


gurevichmark:
With this change, do we still need this text in lifecycle_management.html file:

> Note: The mkdocker can not pull image automatically if the specified image does not exist on dockerhost. Pull the image manually with docker pull <image_name> in dockerhost before running mkdocker. Check available images with docker images

zet809:
@gurevichmark You are right that with this function, the pull image section in the doc will be useless.

chenglch:
This is a issue from last pull request, the logic is correct but a little hard to understand.
    
    unless ($mapping_hash) {
            my $option = '';
            if (defined($args->[0])) {
                $option = $args->[0];
            }
            $callback->({error=>["Not support $command $option"], errorcode=>1});
           return;
       }

As you said `parse_args` has already checked the argument, $args->[0] must be defined. Can we remove the `if (defined($args->[0])` here or put all the check logic  in `parse_args` subroutine?

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -249,6 +255,43 @@ sub http_state_code_info {
 
 #-------------------------------------------------------
 
+=head3 change_node_state
+  To change node state to the state specified, and then send out the HTTP request.
+  Input:
+        $node: the node to change state
+        $to_state_hash: the hash which store the destination state info
+  Return:
```
chenglch: Seems return nothing here.

zet809:
@chenglch Have modified based on your comments, pls review.

Issue #747, opened 2016-02-29
[FVT]Please improve fresh install disk log information


caomengmeng:
Set site.xcatdebugmode=1, then fresh install RH7.2 ppc64, during installing process, using ssh to logon.
In /tmp you will see the xcat_getinstalldisk_log file.
Now the log information for disk information is as below:
```
Check the partition sda1.
    Partition sda1 mount failed or the partition is swap.
Check the partition sda2.
    Partition sda2 mount failed or the partition is swap.
Check the partition sda3.
    Partition sda3 mount success.
    The partition sda3 does not have kernel file.
The disk sda information: disk_wwn= disk_path=/devices/vio/30000002/host0/target0:0:1/0:0:1:0/block/sda disk_driver=ibmvscsi.
    Add disk: sda /devices/vio/30000002/host0/target0:0:1/0:0:1:0/block/sda  into path thichodisks
The install_disk is /dev/sda by sorting path and DRIVER.
```
This is hard to read. so how about improve the disk information like below?
```
Check the partition sda1.
    Partition sda1 mount failed or the partition is swap.
Check the partition sda2.
    Partition sda2 mount failed or the partition is swap.
Check the partition sda3.
    Partition sda3 mount success.
    The partition sda3 does not have kernel file.
The disk sda information: 
  disk_wwn= 
  disk_path=/devices/vio/30000002/host0/target0:0:1/0:0:1:0/block/sda 
  disk_driver=ibmvscsi.
  Add disk: sda /devices/vio/30000002/host0/target0:0:1/0:0:1:0/block/sda  into path thichodisks
The install_disk is /dev/sda by sorting path and DRIVER.
```

caomengmeng:
It is like below now:
```
[get_install_disk]The disk sdd information: 
[get_install_disk]    disk_wwn=0x500a0751036d9e83
[get_install_disk]    disk_path=/devices/pci0000:00/0000:00:03.2/0000:1b:00.0/host6/port-6:3/end_device-6:3/target6:0:3/6:0:3:0/block/sdd
[get_install_disk]    disk_driver=mpt2sas
```
passed

PR #751, opened 2016-02-29
Improve fresh install disk log information, fix issue #747

xuweibj:
To fix issue #747, improve install disk log
Modified logic of reinstall, if there is disk had installed OS, add it to disk_array.
If disk_array is empty, get disks from /proc/partitions. If not, disks equals disk_array.
Step 2, sort disks by WWN/PATH and DRIVER.


whowutwut:
@xuweibj  Is there no more logging of the output to a file?

xuweibj:
@whowutwut , pull request #764 will collect the "ehco" messages and add them to their log file.

PR #753, opened 2016-03-01
modify autotest xCATreg

junxiawang:
@hu-weihua 
I modify xCATreg and please help review

### DIFF
--- a/xCAT-server/share/xcat/tools/xCATreg
+++ b/xCAT-server/share/xcat/tools/xCATreg
@@ -39,14 +39,15 @@ BEGIN
 use lib "$::XCATROOT/lib/perl";
 
 my $date;
-#my $regrootdir="/home/autotest_code";
-#my $mailtitle="P Cluster";
-my $regrootdir="/xCATreg";
-my $mailtitle="X Cluster";
+my $regrootdir="/home/autotest_code";
+my $mailtitle="PPC64LE Cluster";
+#my $regrootdir="/xCATreg";
+#my $mailtitle="X Cluster";
 my $logfiledir = "$regrootdir/log";
 my $logfile="";
 my $configfiledir = "$regrootdir/conf";
 my $configfile = "$regrootdir/conf/xCATreg.conf";
+my $xcatcoredowndir = "$regrootdir/xcatsource/xcat-core-build-down";
 my $xcatcoredir = "$regrootdir/xcatsource/xcat-core";
 my $xcatdepdir = "$regrootdir/xcatsource/xcat-dep";
 my $osisodir = "$regrootdir/iso";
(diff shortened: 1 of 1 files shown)


whowutwut:
Can we add some description on what is being changed with this commit? 

The comment is really generic "modify autotest xCATreg"... Are we adding support for some new test? 

I think there is some formatting inconsistency, the new code is spaced using space :+1: , the old code however is using tabs.. so if we want to run the code  through perltidy, please use a separate commit instead of amend so that we can keep track of the actual change vs cleanup.  Up to you...

xCAT-server/share/xcat/tools/xCATreg
```diff
@@ -1294,49 +1357,49 @@ foreach my $m (keys %mns) {
...
+       # step 3. download xcat core
+       ########################################
+       send_msg(2, "[$m]step 3, Running down_xcat_core...............");
+       $res = down_xcat_core_build($m);
+          if ($res) {
+           syswrite MNWRITE,"REPORTFROM:$m->$mn: download xcat core for $mn failed\n";
+           exit 1;
+        }
+               send_msg(2, "[$m]step 3, Run down_xcat_core...............[OK]");
+
```
hu-weihua: We put ``down_xcat_core_build()`` and ``down_dep()`` in sub-process, that means if we support 5 platform+OS in this control node, we will trigger 5 connections with build server or xcat.org server at same time,  maybe only for downloading the same file.  another thing is we save the same file for many copies in local disk. It's inefficiency and wasteful.

xCAT-server/share/xcat/tools/xCATreg
```diff
@@ -365,6 +376,27 @@ sub down_dep{
...
+    my $downxcatcoredir  = $confkeys{$mn."downxcatcoredir"};
+    my $latestxcatcoredir =  `ssh  $downxcatcoreip ls -rt $downxcatcoredir / |tail -1  |awk '{printf \$0}'`;
+    my $xcatcorename = $confkeys{$mn."xcatcorename"};
+    `scp -r $downxcatcoreip\:$downxcatcoredir/$latestxcatcoredir/$xcatcorename  $downxcatcoredowndir `;
+    if($? != 0){
+    send_msg(0,"[$mn->$confkeys{$mn}][down_xcat_core_build]  failed");
+    return 1;
+    }
+ }
+
```
hu-weihua: Now we download xcat-core from build server instead of downloading source code from github. that's ok, But  when it works with old code logic will result in 2 issues. 

Below things exist at the same time:
1. In the old code logic, at very beginning of xCATreg, we will access github to find out if there is new code check in, if there is, find out who check in code last day and  do the regression.  if not, have rest today. 
2. now, we download xcat-core from build server, but the downloading operation start behind github accessing.
3. There is a delta-time between doing build in build server and doing regression in auto-test environment. it's possible for someone to check in code in the delta-time .

So these will result in 2 problems:
1. Maybe one people's name will show up in the list of who checked in code last day, but his or her code didn't include in the xcat-core.
2. Maybe the xcat-core in build server is the same between 2 continuous days, but regression is triggered. we test the same thing two times. Another situation is xcat-core in build server is changed then there isn't code check-in in last few days, these code changing maybe be tested by xCATreg few days later.

hu-weihua:
Many new codes inserted twistingly, but fortunately it works.  it do have problem because there is delta-time between doing build in build server and doing regression in auto test environment, but we can reduce the probability of the issue happen by adjusting the start time of doing build and doing regression.

Another thing is we are starting to develop new auto test framework, so maybe we can let the old one work like this. @whowutwut and @tingtli , what do you think ?

xCAT-server/share/xcat/tools/xCATreg
daniceexi: The two problems are not big defects, at least they don't impact the running of automation. All of them can be fixed by changing the way to trigger the xCAT build.

hu-weihua:
Ok. it's good for me. I will merge it.

PR #754, opened 2016-03-01
Set the FQDN in the xcatd certificate, use subjectAltNames

xuweibj:
Add subjectAltName in openssl.cnf, set subjectAltName as FQDN and PQDN.
The subject 'SAN' (subjectAltName) used in the certificate can indicate the identity of the server, including FQDN and PQDN.


daniceexi:
I have reviewed the change and agree to change.

@xuweibj Please make sure it works for all the OS: rh, sles and Ubuntu.

PR #757, opened 2016-03-01
Fix race condition issue when printing message in xcatd

chenglch:
related-issue: #729
related-commit: #742


daniceexi:

Issue #761, opened 2016-03-02
[FVT] CA cert ERROR:  SSL connect attempt failed with unknown error


junxiawang:
Env:ppc64le rhels7.2
xcatbuild:latest xcatcode
How to reproduce:
```
1.down the latest build
2.[root@c910f03fsp03v02 /]# lsdef
*******************************************************************
 Using the default of SSL_verify_mode of SSL_VERIFY_NONE for client
 is deprecated! Please set SSL_verify_mode to SSL_VERIFY_PEER
 together with SSL_ca_file|SSL_ca_path for verification.
 If you really don't want to verify the certificate and keep the
 connection open to Man-In-The-Middle attacks please set
 SSL_verify_mode explicitly to SSL_VERIFY_NONE in your application.
*******************************************************************
  at /opt/xcat/lib/perl/xCAT/Client.pm line 256.
Unable to open socket connection to xcatd daemon on localhost:3001.
Verify that the xcatd daemon is running and that your SSL setup is correct.
Connection failure: SSL connect attempt failed with unknown error at /opt/xcat/lib/perl/xCAT/Client.pm line 266.
[root@c910f03fsp03v02 /]# cat /etc/*-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.2 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.2"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.2 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.2:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.2
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.2"
Red Hat Enterprise Linux Server release 7.2 (Maipo)
Red Hat Enterprise Linux Server release 7.2 (Maipo)
```

junxiawang:
still errors 
-----------------
```
[root@c910f03fsp03v02 ~]# lsdef
*******************************************************************
Using the default of SSL_verify_mode of SSL_VERIFY_NONE for client
[...]

PR #762, opened 2016-03-02
To fix issue #761, made a mistake that make certificate generate failed

xuweibj:


daniceexi:
I am good with the change.

whowutwut:
@xuweibj  you need to put the issue into the text of the description for GitHub to make the cross references, not just in the abstract.

xuweibj:
@whowutwut  OK, I will add more information.
The problem is because I made a mistake when added subjectAltName into openssl.cnf, so that the certificate generated failed. Using the default of SSL_verify_mode of SSL_VERIFY_NONE for client
 is deprecated!

daniceexi:
@xuweibj  Victor meant add the issue number in the description for cross reference. I put #761 in this comment, then this pull request will be seen in the issue page #761.

Issue #763, opened 2016-03-02
[DEV] need to logging the pre-script installation for redhat,sles,ubuntu to make debug easy


penguhyang:
Currently, the pre-script file can't be logging.
As a new script ``getinstdisk`` was added, it's necessary to log the installation for easy debug.

PR #743, opened 2016-02-26
split the post.ubuntu, upgrade the post.xcat, logging the ubuntu installation information.

penguhyang:
Currently, there is no big difference between the post.ubuntu and post.xcat.
This implementation will split the post.ubuntu file into two files(post.ubuntu.common and post.xcat) and save the ubuntu installation information.

PR #764, opened 2016-03-02
logging the pre-script to make debug easy

penguhyang:
fix issue #763

### DIFF
--- a/xCAT-server/share/xcat/install/rh/compute.rhels7.tmpl
+++ b/xCAT-server/share/xcat/install/rh/compute.rhels7.tmpl
@@ -147,11 +147,18 @@ reboot
 #INCLUDE_DEFAULT_PKGLIST#
 %end
 %pre
+{
+echo "Running Kickstart Pre-Installation script..."
 #INCLUDE:#ENV:XCATROOT#/share/xcat/install/scripts/pre.rh.rhels7#
+} >>/tmp/pre-install.log 2>&1
 %end
 %post
 mkdir -p /var/log/xcat/
 {
+cat >> /var/log/xcat/xcat.log << "EOF"
+%include /tmp/pre-install.log
+EOF
+echo "Running Kickstart Post-Installation script..."
 #INCLUDE:#ENV:XCATROOT#/share/xcat/install/scripts/post.xcat#
 #INCLUDE:#ENV:XCATROOT#/share/xcat/install/scripts/post.rhels7#
 } >>/var/log/xcat/xcat.log 2>&1
--- a/xCAT-server/share/xcat/install/scripts/chroot.sles
+++ b/xCAT-server/share/xcat/install/scripts/chroot.sles
@@ -7,6 +7,11 @@
 <![CDATA[
 mkdir -p /mnt/var/log/xcat/
 {
+
+cat /tmp/pre-install.log >> /mnt/var/log/xcat/xcat.log
+
+echo "Running AutoYaST Chroot-Installation script..."
+
 #!/bin/sh
 
 AWK=`find / -name awk | head -1`
--- a/xCAT-server/share/xcat/install/scripts/post.sles11
+++ b/xCAT-server/share/xcat/install/scripts/post.sles11
@@ -7,6 +7,7 @@
 <![CDATA[
 mkdir -p /var/log/xcat/
 {
+echo "Running AutoYaST Post-Installation script..."
 #INCLUDE:#ENV:XCATROOT#/share/xcat/install/scripts/post.sles.common#
 #INCLUDE:#ENV:XCATROOT#/share/xcat/install/scripts/post.xcat#
 } >>/var/log/xcat/xcat.log 2>&1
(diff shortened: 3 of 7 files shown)


whowutwut:
@penguhyang Just wondering, what is the reason for the partial pull request?   Why not just keep changes in your forked copy until complete.  Are there going to be multiple commits or amend commits?

daniceexi:
Only goodness I can see is to discuss code with peoples.

penguhyang:
@whowutwut Thanks for review.
This partial pull request has not been finished. We can discuss the code to make sure I'm in right direction.

whowutwut:
Ok, I think that is good,  If we have in the title (don't merge, not complete) then we know that it's partially done... .  I saw you changed the title, does it mean it's ready to be review?

xCAT-server/share/xcat/install/rh/compute.rhels7.tmpl
```diff
@@ -147,11 +147,18 @@ reboot
 #INCLUDE_DEFAULT_PKGLIST#
 %end
 %pre
+{
+echo "Starting Kickstart Pre-Installation..."
```
immarvin: suggest to be ``echo "Running Kickstart Pre-Installation script..."``

xCAT-server/share/xcat/install/rh/compute.rhels7.tmpl
```diff
@@ -147,11 +147,18 @@ reboot
...
 #INCLUDE:#ENV:XCATROOT#/share/xcat/install/scripts/pre.rh.rhels7#
+} >>/tmp/pre-install.log 2>&1
 %end
 %post
 mkdir -p /var/log/xcat/
 {
+cat >> /var/log/xcat/xcat.log << "EOF"
+%include /tmp/pre-install.log
+EOF
+echo "Starting Kickstart Post-Installation..."
```
immarvin: suggest to be ``echo "Running Kickstart Post-Installation script..."``

xCAT-server/share/xcat/install/scripts/chroot.sles
```diff
@@ -7,6 +7,11 @@
 <![CDATA[
 mkdir -p /mnt/var/log/xcat/
 {
+
+cat /tmp/pre-install.log >> /mnt/var/log/xcat/xcat.log
+
+echo "Starting AutoYaST Chroot-Installation..."
```
immarvin: suggest to be ``echo "Running AutoYaST Chroot-Installation script..."``

xCAT-server/share/xcat/install/scripts/post.sles11
```diff
@@ -7,6 +7,7 @@
 <![CDATA[
 mkdir -p /var/log/xcat/
 {
+echo "Starting AutoYaST Post-Installation..."
```
immarvin: suggest to be echo "Running AutoYaST Post-Installation script..."

xCAT-server/share/xcat/install/scripts/pre.sles
```diff
@@ -5,6 +5,9 @@
           <source>
 
 <![CDATA[
+{
+echo "Starting AutoYaST Pre-Installation..."
```
immarvin: suggest to be echo "Running AutoYaST Pre-Installation script..."

xCAT-server/share/xcat/install/scripts/pre.rh.rhels7
```diff
@@ -1,3 +1,9 @@
+export XCATDEBUGMODE="#TABLEBLANKOKAY:site:key=xcatdebugmode:value#"
+if [ "$XCATDEBUGMODE" != "0" ] && [ "$XCATDEBUGMODE" != "" ]
```
whowutwut: Just a general coding style questions... , should we use single ``[`` or double ``[[``?  safer to use POSIX?  But if we are specifying ``#!/bin/bash`` in scripts going forward, then it should not matter?

Does anyone care if we should sync up about whether we should write 

```
if [   ]; then 
fi
```

or

```
if [   ]
then
fi 
```

xCAT-server/share/xcat/install/scripts/pre.rh.rhels7
```diff
@@ -200,4 +206,7 @@ NEXTSERVER=${NEXTSERVER%%:*}
 export nextserver=$NEXTSERVER
 #INSTALL_SOURCES_IN_PRE#
 
-
+if [ "$XCATDEBUGMODE" != "0" ] && [ "$XCATDEBUGMODE" != "" ]
+then
```
whowutwut: why does set -x need to be done twice?

penguhyang: @whowutwut Thanks for review.
set -x: Mark  names for export to subsequent commands via the environment.
Using ``+`` instead of ``-`` turns off the attribute instead.

penguhyang: This pull request realize logging the pre-script for ``redhat`` and ``sles``.
For ``ubuntu``,  I will put that in this one #743.

whowutwut: Thanks @penguhyang ,Sorry, I read that too fast and thought they were both ``set -x`` :-).

penguhyang:
@immarvin Thanks for review. The output messages have been updated.

immarvin:
hi @whowutwut @daniceexi , is there any comments?

immarvin:
the code looks well for me, I will merge it. The discussion on the code style question from @whowutwut can continue in ths thread

PR #768, opened 2016-03-03
Add Rhelhpc7.1 support to discinfo.pm(master)

lyu571:
#Update discinfo.pm for support Rhelhpc7.1(RHEL ComputeNode version) in mainline trunk.


daniceexi:

Issue #755, opened 2016-03-01
[fvt]rspconfig on aix2.9.3 could not work


junxiawang:
Env: AIX720

xCAT Build: lsdef - Version 2.9.3 (git commit 5345643cf8fd145fb6c66397624445b6538d9d24, built Fri Jan 29 02:25:14 CST 2016)

```
bash-3.2# rspconfig c910hmc01 sshcfg
xCAT: Allowing rspconfig to c910hmc01 sshcfg for root from loopback
Can't locate HTML/Form.pm in @INC (you may need to install the HTML::Form module) (@INC contains: /usr/opt/perl5/lib/5.8.2/aix-thread-multi /usr/opt/perl5/lib/5.8.2 /usr/opt/perl5/lib/site_perl/5.8.2/aix-thread-multi /usr/opt/perl5/lib/site_perl/5.8.2 /opt/xcat/lib/perl /usr/opt/perl5/lib/5.8.2/aix-thread-multi /usr/opt/perl5/lib/5.8.2 /usr/opt/perl5/lib/site_perl/5.8.2/aix-thread-multi /usr/opt/perl5/lib/site_perl/5.8.2 /usr/opt/perl5/lib/site_perl/5.8.2/aix-thread-multi  /usr/opt/perl5/lib/site_perl/5.20.1/aix-thread-multi /usr/opt/perl5/lib/site_perl/5.20.1 /usr/opt/perl5/lib/5.20.1/aix-thread-multi /usr/opt/perl5/lib/5.20.1 /usr/opt/perl5/lib/site_perl .) at /opt/xcat/lib/perl/xCAT/PPCfsp.pm line 8.
BEGIN failed--compilation aborted at /opt/xcat/lib/perl/xCAT/PPCfsp.pm line 8.
Compilation failed in require at /opt/xcat/lib/perl/xCAT/PPCcfg.pm line 7.
BEGIN failed--compilation aborted at /opt/xcat/lib/perl/xCAT/PPCcfg.pm line 7.
Compilation failed in require at (eval 34) line 2.
```

PR #769, opened 2016-03-03
Fix issue 755: Disable the module which is using HTML::Form (only for 2.9 branch)

daniceexi:
Since the perl module HTML::Form has been removed from default perl install and xCAT has deprecated the features which is using HTML::Form to runn direct fsp access, the fix is to require the HTML::Form module when using it. If the require operation fails, just display a message.

This fix is for #755.


zet809:

PR #770, opened 2016-03-03
Make auto test environment stop when one key test case failed, keep environment for debug

hu-weihua:
This pull request add a feature for current auto test environment,  when one key test case failed, the whole auto test stop for keeping environment to debug.

Usage:
If you think one test case is very important,  you can add a label ``stop=yes`` in this test case. Take test case ``lsdef_t_o_l``  for example, you can edit it like below:

    start:lsdef_t_o_l
    description: lsdef -o -l
    stop:yes
    cmd:mkdef -t node -o testnode groups=all
    check:rc==0
    cmd:mkdef -t network -o testnetwork net=1.2.3.0 mask=255.255.255.0
    .....
    end

when  ``lsdef_t_o_l`` failed.  you will get report in mail like below:

    [rhels6.5+PPC64]
	    Time consumption: 0 hours 0 minutes 19 seconds
	    TotalCase 7 TotalRun 3 Failed 2
	    Failed cases: lsdef_null, lsdef_t_o_l,
	    Final environment stop at lsdef_t_o_l

``Final environment stop at`` label will indicate the whole auto test environment stop at which case.

Some important things:
1. If the all test cases complete without breaking, the ``Final environment stop at`` label will indicate the last one case we plan to run
2. If there are some unimportant test case (without ``stop=yes`` label)  failed before the key test case. these case's name still appear in ``Failed cases:`` list and be counted into failed case in report mail. ``TotalRun`` will indicate how many cases we have run before breaking.

### DIFF
--- a/xCAT-server/share/xcat/tools/xCATreg
+++ b/xCAT-server/share/xcat/tools/xCATreg
@@ -39,7 +39,7 @@ BEGIN
 use lib "$::XCATROOT/lib/perl";
 
 my $date;
-my $regrootdir="/autotest/xCATreg/";
+my $regrootdir="/home/autotest_code";
 my $mailtitle="PPC64LE Cluster";
 #my $regrootdir="/xCATreg";
 #my $mailtitle="X Cluster";
--- a/xCAT-server/share/xcat/tools/xCATreg
+++ b/xCAT-server/share/xcat/tools/xCATreg
@@ -736,7 +736,7 @@ sub do_test {
                 $batbundle="le_ubuntu_cmd.bundle" if($mn =~ /ubuntuEL/);
         }	
         if(($mn !~ /ubuntu/) && ($mn !~ /E/)){
-            for(my $loop=0; $loop<1; $loop++){
+            {
                 my $rc;
 
 		#diskfull installation in flat mode test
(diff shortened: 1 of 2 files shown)


xCAT-server/share/xcat/tools/xCATreg
```diff
@@ -813,40 +822,57 @@ sub do_test {
...
 
 		#statelite installation in hierarchy mode test
 		send_msg(2, "[$mn->$confkeys{$mn}][do_test] doing statelite installation test in hierarchy mode in $mn.....");
 		system("xdsh $confkeys{$mn} \"xcattest -f /opt/xcat/share/xcat/tools/autotest/default.conf  -t $statelite_installation_hierarchy_testcase\" >/dev/null 2>&1");
 		send_msg(2, "[$mn->$confkeys{$mn}][do_test] do statelite installation test in hierarchy mode in $mn.....[done]");	
-		
+                $casestop=0;
+        }		
     }else{
+         for(my $loop=0; $loop<1; $loop++){
```
neo954: So, this loop loop only once. Thus there is no loop.

xCAT-server/share/xcat/tools/xCATreg
```diff
@@ -734,25 +736,32 @@ sub do_test {
                 $batbundle="le_ubuntu_cmd.bundle" if($mn =~ /ubuntuEL/);
         }	
         if(($mn !~ /ubuntu/) && ($mn !~ /E/)){
+            for(my $loop=0; $loop<1; $loop++){
```
neo954: So, this loop loop only once. Thus, there is no loop.

tingtli:
I agree to merge

PR #773, opened 2016-03-04
Support the new format 'file -> (noderange for permitted nodes) file' in synclist file for file syncing

tingtli:
Support the new format 'file -> (noderange for permitted nodes) file' in synclist file for file syncing


daniceexi:
The change looks good for me, I agree to merge it.

PR #775, opened 2016-03-04
fixes #708, first try at getting the confignics -s fixing

### LINKED ISSUE #708: [Customer] confignics -s, sets up config files, but interface loses IP after dhcp lease expire
Hi,

I have started to use `confignics -s` as a way to configure the install nic. But as the lease of the IP still exists in DHCP on the server and on the client, it is still not configured on the live system to be static. When the lease expires, it no longer grabs the IP anymore, and lose all connectivity. We then have to login via `rcons`, or reboot the machines to get the network on the machine back up again.

Maybe we need to setup the `configeth` such that the `bool_restart_flag` is set to 1, if we are configuring the install nic. It probably makes sense with the section that restarts the nic, that we do a ifdown before we do a ifup, to cover all scenarios, such as this one as well.

I will test this later, in the next few days, and see if I can do a PR

arif-ali:
This is a first attempt at trying to fix Issue #708, that the config changes with the confignics -s, but is not being applied

This is currently in testing, and will update soon, please don't commit this yet

### DIFF
--- a/xCAT/postscripts/configeth
+++ b/xCAT/postscripts/configeth
@@ -502,6 +502,9 @@ elif [ "$1" = "-s" ];then
 		parse_nic_extra_params "$str_extra_params"
 	fi 
 
+    # Bring the interface down before cofniguring the interface
+    ifdown $str_inst_nic
+
     if [ -f "/etc/debian_version" ];then
         str_conf_file="/etc/network/interfaces.d/${str_inst_nic}"
         echo "auto ${str_inst_nic}" > $str_conf_file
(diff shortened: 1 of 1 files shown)


arif-ali:
This patch has resolved the problem for me

whowutwut:
@arif-ali  I added the issue number into your original post so GitHub can create the reference.  If in the title, it is unable to create a link to the Issue.

arif-ali:
can we also cherry-pick this into 2.11 as well?

daniceexi:
The change looks good for me.

daniceexi:
@arif-ali I have cherry-picked the change to 2.11 branch.

arif-ali:
thanks @daniceexi

PR #778, opened 2016-03-06
simplify the process of running xCAT in Docker with Docker 1.10

immarvin:
http://10.3.5.21/doc/advanced/docker/run_xcat_in_docker.html

### DIFF
--- a/docs/source/advanced/docker/run_xcat_in_docker.rst
+++ b/docs/source/advanced/docker/run_xcat_in_docker.rst
@@ -3,15 +3,19 @@ Run xCAT in Docker Container
 
 `Docker <https://www.docker.com/>`_ is a popular application containment environment. With Docker, applications/Services are shipped as **Docker images** and run in **Docker containers**. **Docker containers** include the application and all of its dependencies, but share the kernel with other containers. They run as an isolated process in userspace on the host operating system. The server on which  **Docker containers** run is called **Docker host**.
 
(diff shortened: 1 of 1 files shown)


daniceexi:
@immarvin I am note sure how many people has been moved to docker 1.10. Since we already has the doc to make it works well for docker 1.09 or lower version, we suggest we can keep the original doc and make this new doc to be the one for v1.10 and newer.

immarvin:
On the loop device problem, i added the code logic to create loop device in startservice.sh

immarvin:
Hi, i do not think there is anyone using xcat in docker now, since we have not released it yet. Since docker is in the fast development phase,  the behavior changes and new features are introduced in each major release, some options are deprecated,some are added, the behavior of some existing options are changed, this will continue until it is matured and stable enough in the production environment.   It seems docker does not care backward compatibility at this time,  so I think it is ok for xcat to support the latest docker major release, docker.docs provided the steps on docker upgrade , we do not need to worry about this.

I think we should focus on xcat release in xcat docker  image instead of docker release, since the xcat docker image on docker hub is built and based on the features of docker v1.10 release, i am not sure whether it still works on docker 1.9, and i do not want to put any effort to make sure this:)

immarvin:
Ping @daniceexi

daniceexi:
OK. If our xCAT docker image is built base on docker v1.10, it's reasonable to only support v1.10 now.

PR #780, opened 2016-03-07
modified subjectAltName method, did not use ENV

xuweibj:
Issue #761


xuweibj:
@daniceexi 
About the first limitation, I modified. Is this OK?
When run xcatconfig -c, it will generate openssl.cnf by openssl.cnf.tmpl. So it will be covered by the default value"#XCATCASAN#", and then be changed to `hostname`

daniceexi:
@xuweibj The change is not good enough. I just check in the code to avoid the failure of automation. We need to consider a better fix later.

whowutwut:
@daniceexi  Is there an issue to track the better fix?

daniceexi:
@whowutwut I have discussed with @xuweibj, looks like there was no a better and simple fix. From the current usage process, this fix works fine. So we don't plan to make an enhancement for this fix any more.

PR #785, opened 2016-03-08
refine the "run xcat in docker" doc

immarvin:
refine the "Save and Restore xCAT data" section


daniceexi:

Issue #784, opened 2016-03-08
[FVT]The command " rcons node -f -s" is not as excepted.


caomengmeng:
Description: The -f flag can not be used with the -s flag for the rcons command. But the return message is not right and return value is 0 not 1.

ENV:
```
[root@c910f02c04p07 ~]# rpm -qa | grep -i xCAT
conserver-xcat-8.1.16-10.ppc64
xCAT-server-2.12-snap201602251930.noarch
xCAT-test-2.12-snap201602251930.noarch
grub2-xcat-2.02-0.16.el7.snap201506090204.noarch
xCAT-genesis-base-ppc64-2.11-snap201511160819.noarch
syslinux-xcat-3.86-2.noarch
xCAT-client-2.12-snap201602251930.noarch
xCAT-buildkit-2.12-snap201602251931.noarch
xCAT-2.12-snap201602251931.ppc64
ipmitool-xcat-1.8.11-3.ppc64
perl-xCAT-2.12-snap201602251930.noarch
xCAT-genesis-scripts-ppc64-2.12-snap201602251931.noarch
[root@c910f02c04p07 ~]# cat /etc/*release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.1 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.1"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.1 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.1:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.1
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.1"
Red Hat Enterprise Linux Server release 7.1 (Maipo)
Red Hat Enterprise Linux Server release 7.1 (Maipo)
```

Steps to reproduce:
```
[root@c910f02c04p07 ~]# rcons c910f02c04p07 -f -s
console: gethostbyname(-f): host lookup error
[root@c910f02c04p07 ~]# echo $?
0
```

zet809:
[...]

PR #788, opened 2016-03-09
To fix issue #784, output error message when use rcons node -f -s

xuweibj:
#784
When use "rcons node -f -s", output "Error: -f and -s could not be used at the same". 
And exit 1.

### DIFF
--- a/xCAT-client/bin/rcons
+++ b/xCAT-client/bin/rcons
@@ -28,28 +28,35 @@ if [ "$1" = "-v" -o "$1" = "--version" ]; then
     exit 0
 fi
 
-if [ -n "$3" ] && [ "$3" = "-f" ]; then
-    FORCE=-f
-    CONSERVER=$2
-fi
+param_n=0
+for parameter in $@; do
+    ((param_n++))
 
-#console spy mode
-if [ -n "$3" ] && [ "$3" = "-s" ]; then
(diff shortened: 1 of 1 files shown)


daniceexi:
@xuweibj The original code for option check  was not good. If adding your code, there were three blocks of code to handle the parameters. Even so it still cannot handle the case where there were more than two options like 'rcons <node> -a -s -f'.

I have proposal, you can review:

~~~~
firstparam="yes"
for param in $@; do
    # check the first parameter to be conserver
    if [ "$firstparam" = "yes" ] && [ "$1" != "-s" ] && [ "$1" != "-f" ]; then
        CONSERVER=$1
        firstparam="no"
        continue
    fi
    firstparam="no"

    if [ "$param" = "-f" ] || [ "$param" = "-s" ]; then
        if [ -z "$FORCE" ]; then
            FORCE=$param
        else
            echo "Error: cannot use options -s and -f together."
            exit 1
        fi
    elif [ -n "$param" ]; then
        echo "Error: cannot recognize parameter \"$param\"."
        exit 1
    fi
done
~~~~

xuweibj:
@daniceexi , I modified based on your comment. But made some changes. Please help review.

daniceexi:
I have reviewed the change and agree to merge. Thanks.

Issue #787, opened 2016-03-09
[DEV] site.xcatdebugmode need a fixed value (0 1 blank)


penguhyang:
Currently, xcatdebugmode can be any value.If xcatdebugmode=0 or xcatdebugmode value is blank, that means xcatdebugmode is off. Using other values turns on xcatdebugmode.
However, most scripts have a verification between 0 and 1. It's necessary to make a fixed value scope for xcatdebugmode.

daniceexi:
@penguhyang Do you plan to support the valid value like '1|Y|y|yes, 0|N|n|no' for site.xcatdebugmode attribute?

penguhyang:
@daniceexi Yes, these valid value will be supported.

PR #789, opened 2016-03-09
make a fixed value scope for site.xcatdebugmode(0 1 2)

penguhyang:
fix issue #787

### DIFF
--- a/docs/source/guides/admin-guides/basic_concepts/global_cfg/index.rst
+++ b/docs/source/guides/admin-guides/basic_concepts/global_cfg/index.rst
@@ -79,7 +79,13 @@ Install/Deployment Attributes
   (``yes/1`` or ``no/0``). Default is ``no``. If yes, it will instruct xCAT at ``nodeset`` and ``updatenode`` time to query the db once for all of the nodes passed into the cmd and create the mypostscript file for each node, and put them in a directory of tftpdir(such as: /tftpboot). If no, it will not generate the mypostscript file in the ``tftpdir``.
 
 * xcatdebugmode:  
-  the value of xcat debug mode. Currently supported values: ``0: off(default); 1: on``
+  the xCAT debug level. xCAT provides a batch of techniques to help user debug problems while using xCAT, especially on OS provision, such as collecting logs of the whole installation process and accessing the installing system via ssh, etc. These techniques will be enabled according to different xCAT debug levels specified by 'xcatdebugmode', currently supported values: ::
+
+    '0':  disable debug mode
+    '1':  enable basic debug mode
+    '2':  enalbe expert debug mode
+
+  For the details on 'basic debug mode' and 'expert debug mode', please refer to xCAT documentation.
 
 
 Remoteshell Attributes
--- a/xCAT-server/lib/perl/xCAT/Template.pm
+++ b/xCAT-server/lib/perl/xCAT/Template.pm
@@ -351,8 +351,12 @@ sub subvars {
       $inc =~ s/#GETNODEDOMAIN:([^#]+)#/get_node_domain($1)/eg;
       $inc =~ s/#GETPRINICMAC:([^#]+)#/xCAT::Utils::parseMacTabEntry(tabdb("mac",$1,"mac"),$1)/eg;
    
-      if($::XCATSITEVALS{xcatdebugmode} eq "1"){
-           $inc =~ s/#UNCOMMENONXCATDEBUGMODE#/ /g;
+      if(($::XCATSITEVALS{xcatdebugmode} eq "1") or ($::XCATSITEVALS{xcatdebugmode} eq "2")){
+           $inc =~ s/#UNCOMMENTOENABLEDEBUGPORT#/ /g;
+      }
+
+      if($::XCATSITEVALS{xcatdebugmode} eq "2"){
+           $inc =~ s/#UNCOMMENTOENABLESSH#/ /g;
       }
  
       my $nrtab = xCAT::Table->new("noderes");
@@ -1604,6 +1608,13 @@ sub tabdb
       }
       return "";
       #return "#TABLEBAD:$table:field $field not found#";
+    } else {
+        # check for site.xcatdebugmode
+        if (($table =~ /site/) and ($key =~ /xcatdebugmode/)) {
+            if ((($ent->{$field}) ne "0") and (($ent->{$field}) ne "1") and (($ent->{$field}) ne "2")) {
+                $tmplerr="Unable to recognise filed <$field> from table <$table>, with key <$key>. Please enter '0' '1' or '2'"
+            }
+        }
     }
     return $ent->{$field};
 
--- a/xCAT-server/lib/xcat/plugins/anaconda.pm
+++ b/xCAT-server/lib/xcat/plugins/anaconda.pm
@@ -1528,15 +1528,17 @@ sub mkinstall
                 }
            }
               
-           if($::XCATSITEVALS{xcatdebugmode} eq "1"){
+           if(($::XCATSITEVALS{xcatdebugmode} eq "1") or ($::XCATSITEVALS{xcatdebugmode} eq "2")){
                  unless($instserver eq '!myipfn!'){
                     my($host,$ip)=xCAT::NetworkUtils->gethostnameandip($instserver);
                     $instserver=$ip;
                  } 
  
                  if (xCAT::Utils->version_cmp($kversion,"7.0") >= 0){
-                    #enable ssh access during installation
-                    $kcmdline .= " inst.sshd";  
+                    if($::XCATSITEVALS{xcatdebugmode} eq "2"){
+                       #enable ssh access during installation
+                       $kcmdline .= " inst.sshd";  
+                    }
                            
                     #set minimum level of messages to be logged on the console
                     #to be "debug"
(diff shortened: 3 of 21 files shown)


immarvin:
hi @penguhyang , we'd better not to restrict user input in site table.  
for the site.xcatdebugmode,  we should turn the debug mode on if  site.xcatdebugmode="1|Y|y|yes", otherwise, the debugmode should be turned off. The modification should be made on all the places we test the value of site.xcatdebugmode

penguhyang:
@immarvin Yes, that's a good idea.

immarvin:
run ``grep -i xcatdebugmode `` on the "xcat-core" code, they are many places  need modification

whowutwut:
@immarvin @penguhyang with the comment above, site.xcatdebugmode=1|Y|y|yes, why do we allow that? In the documentation, there is 2 values supported, 0, 1 and 2.  I don't think we should allow yes because what debug level does that equate to?  Is there a way to change this value using a chdef command where we can flag the error?

penguhyang:
@whowutwut Thanks for review. 
We discussed about the value of site.xcatdebugmode in China team.
site.xcatdebugmode=1|Y|y|yes will be abandoned. New values are 0 1 2.

xCAT-server/lib/perl/xCAT/Postage.pm
```diff
@@ -787,8 +787,23 @@ sub getAllAttribsFromSiteTab {
             #$result .= "export $attribute\n";
 
         }
+        elsif ($attribute eq "XCATDEBUGMODE")
```
immarvin: suggest pass the raw value of "site.xcatdebugmode" to mypostscript and mypostscript.post directly without mapping, since there might be some features added in the future which perform different  operations according to the value of xcatdebugmode

xCAT-server/lib/perl/xCAT/Template.pm
```diff
@@ -351,9 +351,13 @@ sub subvars {
       $inc =~ s/#GETNODEDOMAIN:([^#]+)#/get_node_domain($1)/eg;
       $inc =~ s/#GETPRINICMAC:([^#]+)#/xCAT::Utils::parseMacTabEntry(tabdb("mac",$1,"mac"),$1)/eg;
    
-      if($::XCATSITEVALS{xcatdebugmode} eq "1"){
+      if(($::XCATSITEVALS{xcatdebugmode} eq "1") or ($::XCATSITEVALS{xcatdebugmode} eq "2")){
```
immarvin: why keep these lines?

xCAT-server/lib/perl/xCAT/Template.pm
```diff
@@ -1601,9 +1605,24 @@ sub tabdb
          } else {
             $tmplerr="Unable to find requested filed <$field> from table <$table>, with key <$key>"
          }
+      } else {
```
immarvin: do not map the value of xcatdebugmode

xCAT-server/lib/perl/xCAT/Template.pm
```diff
@@ -1601,9 +1605,24 @@ sub tabdb
...
+         if (($table =~ /site/) and ($key =~ /xcatdebugmode/)) {
+            return "off";
+         }
       }
       return "";
       #return "#TABLEBAD:$table:field $field not found#";
+    } else {
+        # check for site.xcatdebugmode
```
immarvin: do not map the value of xcatdebugmode

xCAT-server/lib/xcat/plugins/anaconda.pm
```diff
@@ -745,7 +745,7 @@ sub mknetboot
            $kcmdline .= " nonodestatus ";
         }
 
-        if($::XCATSITEVALS{xcatdebugmode} eq "1"){
+        if(($::XCATSITEVALS{xcatdebugmode} eq "1") or ($::XCATSITEVALS{xcatdebugmode} eq "2")){
```
immarvin: mknetboot is for diskless provision, please do not touch

immarvin:
I noticed you did not split  the feature of "enablessh" and "remote log" according to xcatdebugmode different levels, can you tell the code slices to process these 2 features exactly? if not, please read the code first and make yourself clear enough on the code logic before you make any change.

xCAT-server/lib/perl/xCAT/Postage.pm
penguhyang: fixed.

docs/source/guides/admin-guides/basic_concepts/global_cfg/index.rst
```diff
@@ -79,7 +79,7 @@ Install/Deployment Attributes
   (``yes/1`` or ``no/0``). Default is ``no``. If yes, it will instruct xCAT at ``nodeset`` and ``updatenode`` time to query the db once for all of the nodes passed into the cmd and create the mypostscript file for each node, and put them in a directory of tftpdir(such as: /tftpboot). If no, it will not generate the mypostscript file in the ``tftpdir``.
 
 * xcatdebugmode:  
-  the value of xcat debug mode. Currently supported values: ``0: off(default); 1: on``
+  the value of xcat debug mode. Currently supported values: ``0: off(default); 1: on(debug level 1); 2: on(debug level 2)``
```
penguhyang: fixed

xCAT-server/lib/perl/xCAT/Template.pm
penguhyang: fixed

xCAT-server/lib/perl/xCAT/Template.pm
penguhyang: fix for ubuntu

penguhyang:
@immarvin Thanks for review. Code updated.

xCAT-server/lib/perl/xCAT/Template.pm
```diff
@@ -351,9 +351,13 @@ sub subvars {
       $inc =~ s/#GETNODEDOMAIN:([^#]+)#/get_node_domain($1)/eg;
       $inc =~ s/#GETPRINICMAC:([^#]+)#/xCAT::Utils::parseMacTabEntry(tabdb("mac",$1,"mac"),$1)/eg;
    
-      if($::XCATSITEVALS{xcatdebugmode} eq "1"){
+      if(($::XCATSITEVALS{xcatdebugmode} eq "1") or ($::XCATSITEVALS{xcatdebugmode} eq "2")){
            $inc =~ s/#UNCOMMENONXCATDEBUGMODE#/ /g;
       }
+
+      if($::XCATSITEVALS{xcatdebugmode} eq "2"){
+           $inc =~ s/#UNCOMMENTOENABLESSH#/ /g;
```
immarvin: what's the difference between "UNCOMMENONXCATDEBUGMODE" and "UNCOMMENTOENABLESSH"? it's not clear enough

xCAT-server/lib/xcat/plugins/anaconda.pm
penguhyang: ok, as this pull request only support diskful install. I will support diskless install for xcatdebugmode in future.

xCAT-server/lib/perl/xCAT/Template.pm
penguhyang: changed "UNCOMMENONXCATDEBUGMODE" to "UNCOMMENTOENABLEDEBUGPORT" to make code more clear.

xCAT-server/lib/perl/xCAT/Template.pm
```diff
@@ -1604,6 +1608,13 @@ sub tabdb
       }
       return "";
       #return "#TABLEBAD:$table:field $field not found#";
+    } else {
+        # check for site.xcatdebugmode
+        if (($table =~ /site/) and ($key =~ /xcatdebugmode/)) {
```
immarvin: no need to check the value, we only make sure when user sets values mentioned in our doc, the logic will be triggered, otherwise, nothing will happen

penguhyang: In our latest design, we need to check the value of xcatdebugmode. 
If it is not 0 1 2 or null, then a message will remind the user to set the right values. Isn't it?

immarvin: ok, if this is the case, please make sure the case that "xcatdebugmode" is not set is checked. And is this the only place you need to check the value of "xcatdebugmode"?

penguhyang: yes, I think this is the only place need to check the value of "xcatdebugmode". If xcatdebugmode is not set then xcatdebug is off.

immarvin: if site.xcatdebugmode is not set, will "$tmplerr="Unable to recognise filed <$field> from table <$table>, with key <$key>. Please enter '0' '1' or '2'"  be returned?

penguhyang: no, "$tmplerr="Unable to recognise filed <$field> from table <$table>, with key <$key>. Please enter '0' '1' or '2'" used for check the defined value "$ent and  defined($ent->{$field})"

immarvin: have you checked ""$ent and defined($ent->{$field})""?

penguhyang: yes

PR #791, opened 2016-03-10
configure static IP for container without using other tool

zet809:


zet809:
@daniceexi I have modified code and add some comments, pls help to review.

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -653,7 +733,7 @@ sub parse_args {
                 return ( [1, "Option $op is not supported for $cmd"]);
             }
         }
-        $request->{arg}->[0] = "force";
+        $request->{state_machine_hash} = $command_states{$cmd}{disconnect};
```
daniceexi: Please just pass the state_machine name instead of the address of hash

xCAT-server/lib/xcat/plugins/docker.pm
```diff
@@ -307,6 +348,38 @@ sub change_node_state {
 =cut  
 
 #-------------------------------------------------------
+#-----------------------------------------------------------------------------
```
daniceexi: Please move the descriptions for state-machines to location which defines the state-machine.

daniceexi: Please change the name of single_state_engine to something like default_engine since it's not only for single state case. It's used for a many state by default.

daniceexi:
I have reviewed the change and agree to merge. @zet809 Thanks for your great effort.

PR #794, opened 2016-03-11
modify setupdockerhost to create network object based on the bridge s…

zet809:
…pecified for 1.10 and higher

### DIFF
--- a/xCAT/postscripts/setupdockerhost
+++ b/xCAT/postscripts/setupdockerhost
@@ -3,12 +3,10 @@
 # IBM(c) 2014 EPL license http://www.eclipse.org/legal/epl-v10.html
 #-------------------------------------------------------------------------------
 #=head1  setupdockerhost
-#=head2  Used on Linux only. Configure docker host
+#=head2  Used on Linux only. Configure SSL connection and docker network object on docker host
 #
 #  You can run the following commands on MN:
-#       updatenode noderange setupdockerhost (To use the default bridge mydocker0 for docker services)
-#       updatenode noderange "setupdockerhost -b=dockerbr0" (To specify the bridge which will be used by docker services)
-#
+#       updatenode noderange -P "setupdockerhost netname=net/mask@gateway:nicname"
 #
 #=cut
 #-------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


zet809:
@daniceexi Modified, pls help to review.

xCAT/postscripts/setupdockerhost
```diff
@@ -140,8 +106,68 @@ fi
...
     echo "Docker service starting failed"
     exit 1
 fi
+
+# TO check whether the docker server version is lower than or equal to 1.10
+#    If so, need to create docker network object based on the bridge specified
+#    Else, use the brdige as an option for docker daemon
+
+
+# For docker 1.8.x and higher
```
daniceexi: meaningless comment

xCAT/postscripts/setupdockerhost
```diff
@@ -140,8 +106,68 @@ fi
...
+DOCKER_VERSION=`docker info 2>/dev/null | grep "Server Version:" | awk -F: '{print $2}'`
+if [ -z "$DOCKER_VERSION" ];then
+    # For docker 1.7.x and lower
```
daniceexi: meaningless comment

xCAT/postscripts/setupdockerhost
```diff
@@ -140,8 +106,68 @@ fi
...
+    DOCKER_VERSION=`docker version | grep "Server version" | awk -F: '{print $2}'`
+    if [ -z "$DOCKER_VERSION" ];then
+        echo "Failed to get docker server version"
+        exit 1
+    fi
+fi
+MAX_VERSION=`echo $DOCKER_VERSION | awk -F. '{print $1}'`
```
daniceexi: It should use Major, Minor version

xCAT/postscripts/setupdockerhost
```diff
@@ -140,8 +106,68 @@ fi
...
+fi
+
+FLAG=0
+if [  $MAX_VERSION -le 1 -a $MIN_VERSION -le 10 ]; then
+    FLAG=1
+fi
+
+# The docker newer than 1.10.x can accept NICNAME directly.
+# But we treat it as 1.10.x in case it is not supported immediately.
+FLAG=1
```
daniceexi: why did you set it in static here?

daniceexi:
I have reviewed the change and agreed to merge. Thanks.

whowutwut:
@daniceexi @zet809 I just noticed the comments that were made is regarding checking the version of docker installed.  This might be a good time to bring up pull request #374 again.  Instead of doing these parsing of each major and minor version, can we look into the suggestion to use the Sort::Versions instead?   I think we should use the libraries that are provided rather than doing the parsing like we have done...

daniceexi:
Yes, I agree with that. This is a shell script, we could make a version comparison function in xcatlib so that all the postscripts can use it.

Issue #795, opened 2016-03-11
[DEV] updatenode command keeps going after detecting missing node attributes.


gurevichmark:
updatenode command checks all DB tables to build a hash of all node attributes for the specified noderange.
If a node does not have "os" or "arch" attribute specified, an error message is displayed but the updatenode command will still be performed on that node. Also, after one such node is detected in the noderange, the hash of node attributes for the remaining nodes in the noderange is not performed.

Comment from Wang Xiaopeng on this issue - "We should enable the checking that display an error message and return directly."

PR #796, opened 2016-03-11
Updatenode to skip processing nodes which have some attributes missing

gurevichmark:
This is a fix for issue #795 .

When a node is detected in Postage.pm with some attributes not defined, such as "os" or "arch", the node is added to exclude_nodes list. Later in updatenode.pm, a list of nodes in the request is modified, to remove all the nodes found in the exclude_nodes list. As the result, the updatenode action is only executed on the nodes remaining in the request and not on the nodes found in the exclude_nodes list.


gurevichmark:
Added a message as recommended by Xiaopeng Wang.

daniceexi:
The change looks good. @gurevichmark Thanks.

PR #797, opened 2016-03-14
[FVT]Automation Case: Add analyze xCAT log file case to track known error even the OS is installed successfully 

caomengmeng:
@tingtli 

Here is the case used to track the known error message even the OS is installed successfully.
Would you please review it.


tingtli:
@caomengmeng  Please modify the case according to my comments. Thanks!

caomengmeng:
@tingtli 
Update OS information, and move the "support rh and sles" into the description.

Issue #801, opened 2016-03-15
[FVT] Ubuntu packaging problem, xcat-client depends perl-xCAT, should be perl-xcat


neo954:
In file ``xCAT-client/debian``, it shows ``xcat-client`` depends ``perl-xCAT``. But package ``perl-xCAT`` is non-existence. We do not have ``perl-xCAT``, instead, we have ``perl-xcat``.

Based on Debian policy, section 5.6.1 and section 5.6.7 [1], there is no uppercase letters allowed in the package names.

> Package names (both source and binary, see Package, Section 5.6.7) must consist only of lower case letters (a-z), digits (0-9), plus (+) and minus (-) signs, and periods (.). They must be at least two characters long and must start with an alphanumeric character.

[1] https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Source

    Source: xcat-client
    Section: admin
    Priority: extra
    Maintainer: xCAT <xcat-user@lists.sourceforge.net>
    Build-Depends: debhelper (>= 5)
    Standards-Version: 3.7.2

    Package: xcat-client
    Architecture: all
    Depends: ${perl:Depends}, nmap, perl-xCAT
    Description: Core executables and data of the xCAT management project
     xCAT-client provides the fundamental xCAT commands (chtab, chnode, rpower, etc) helpful in administrating systems at scale, with particular attention paid to large HPC clusters.

neo954:
Create pull request #802 for the fixing.

neo954:
Will close this one.

Issue #790, opened 2016-03-10
[FVT]:Update  build-ubunturepo for xcat build to support Ubuntu 16.04 (Xenial Xerus)


tingtli:
We need to update build-ubunturepo to support the new Ubuntu distribution Ubuntu 16.04 (Xenial Xerus).

Here is the code which needs update.
```
# Supported distributions
dists="saucy trusty utopic"    <== Add Xenial in the distribution list.

```
And for xcat-core, following code needs update for "Xenial"
```
    for dist in $dists; do
       if [ "$dist" = "trusty" ] || [ "$dist" = "utopic" ]; then   
           tmp_out_arch="amd64 ppc64el"
       else
           tmp_out_arch="amd64"
       fi
...
    for dist in $dists; do
      if [ "$dist" = "trusty" ] || [ "$dist" = "utopic" ]; then
          deb_files=$all_files
      else
          deb_files=$amd_files
      fi

```

For xcat-dep , following codes needs to update
```
for dist in $dists; do
       if [ "$dist" = "trusty" ] || [ "$dist" = "utopic" ]; then
           tmp_out_arch="amd64 ppc64el"
       else
           tmp_out_arch="amd64"
       fi

...

    for dist in $dists; do
      if [ "$dist" = "trusty" ] || [ "$dist" = "utopic" ]; then
          deb_files=$all_files
      else
          deb_files=$amd_files
      fi

```

neo954:
This issue depends on issue #801. Need fix issue #801 first.

whowutwut:
@neo954, thanks I will try my pkg changes again with the change you made

neo954:
Create pull request #830 for all the templates and pkglists.

PR #802, opened 2016-03-15
Fix isse #801, change the package name perl-xCAT to perl-xcat.

neo954:
Fixing the Ubuntu packaging problem.

### DIFF
--- a/xCAT-client/debian/control
+++ b/xCAT-client/debian/control
@@ -7,6 +7,6 @@ Standards-Version: 3.7.2
 
 Package: xcat-client
 Architecture: all
-Depends: ${perl:Depends}, nmap, perl-xCAT
+Depends: ${perl:Depends}, nmap, perl-xcat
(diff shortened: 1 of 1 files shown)


daniceexi:
Refer to the issue #790, this fix is reasonable to complain the issue why there's dependency issue for 16.04.
It's OK to merge.

whowutwut:
@neo954 @daniceexi  Sorry for this late review, but we should have probably checked all control files to make sure we are doing lowercase in every spot.  We missed ``xCAT-client`` in:
https://github.com/xcat2/xcat-core/blob/master/xCAT-OpenStack-baremetal/debian/control
https://github.com/xcat2/xcat-core/blob/master/xCAT-vlan/debian/control

daniceexi:
Good point, we need to change all the dependency packages name to be low case in the Debian deb control file. @neo954

neo954:
Created new pull request #823.

PR #803, opened 2016-03-15
Add release info for 2.9.3

daniceexi:


whowutwut:
@tingtli  I updated the index.html page on xcat.org to show the 2.9.3 release https://xcat.org

We should decide what we want to do for the SourceForge site.  I was going to update it but we didn't include the 2.11 release information so I'm not sure how we want to proceed moving forward. 

I see the News in https://github.com/xcat2/xcat-core/wiki was updated, should we get rid of the news bar on xcat.org so we don't have this many places to update?

PR #804, opened 2016-03-15
Changes to support installing xCAT on Ubuntu 16.04 to unblock #790

whowutwut:
Fixes Issue #790 

- Change the SSL_VERIFY_PEER in commit ed02740 to be a number instead of a string.  
- Looks like ``apt-get install xcat`` needs to be in lowercase, most strict in this version of Ubuntu? 
- Documentation change for the instructions
- change the build script so that we only need to update the release name moving forward instead of looking for each release of Ubuntu

Followed the steps in http://xcat-docs.readthedocs.org/en/latest/guides/install-guides/apt/install_xcat.html  and xCAT installs OK with my sandbox build

```
root@fs1vm202:/tmp/xcat-dep# source /etc/profile.d/xcat.sh 
root@fs1vm202:/tmp/xcat-dep# lsxcatd -a
Version 2.12 (git commit ed02740d1e18595f24181be7865aa0f5bd07b0b4, built Tue Mar 15 14:06:50 EDT 2016)
This is a Management Node
dbengine=SQLite
```

Documentation change: 
image

### DIFF
--- a/build-ubunturepo
+++ b/build-ubunturepo
@@ -63,7 +63,7 @@ for i in $*; do
 done
 
 # Supported distributions
-dists="saucy trusty utopic"
+dists="saucy trusty utopic xenial"
 
 c_flag=            # xcat-core (trunk-delvel) path
 d_flag=            # xcat-dep (trunk) path
@@ -287,12 +287,13 @@ then
     mkdir conf
 
     for dist in $dists; do
-       if [ "$dist" = "trusty" ] || [ "$dist" = "utopic" ]; then
-           tmp_out_arch="amd64 ppc64el"
-       else
-           tmp_out_arch="amd64"
-       fi
-       cat << __EOF__ >> conf/distributions
+        # for all releases moving forward, support amd64 and ppc64el
+        tmp_out_arch="amd64 ppc64el"
+        if [ "$dist" = "saucy" ]; then
+            # for older releases of Ubuntu that does not support ppc64el
+            tmp_out_arch="amd64"
+        fi
+        cat << __EOF__ >> conf/distributions
 Origin: xCAT internal repository
 Label: xcat-core bazaar repository
 Codename: $dist
@@ -314,14 +315,14 @@ __EOF__
     amd_files=`ls ../$package_dir_name/*.deb | grep -v "ppc64el"`
     all_files=`ls ../$package_dir_name/*.deb`
     for dist in $dists; do
-      if [ "$dist" = "trusty" ] || [ "$dist" = "utopic" ]; then
-          deb_files=$all_files
-      else
-          deb_files=$amd_files
-      fi
-      for file in $deb_files; do
-	reprepro -b ./ includedeb $dist $file;
-      done
+        deb_files=$all_files
+        if [ "$dist" = "saucy" ]; then
+            # for older releases of Ubuntu that does not support ppc64el
+            deb_files=$amd_files
+        fi
+        for file in $deb_files; do
+            reprepro -b ./ includedeb $dist $file;
+        done
     done
     #create the mklocalrepo script
     cat << '__EOF__' > mklocalrepo.sh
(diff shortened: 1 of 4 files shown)


whowutwut:
@daniceexi Should we cherry-pick this back to 2.11 or just release Ubuntu support in 2.12 mid 2016?  Need to discuss in next mtg.

build-ubunturepo
```diff
@@ -45,11 +45,16 @@ if [[ ! -f /etc/lsb-release ]]; then
 fi 
 . /etc/lsb-release
 
-REPREPRO=`which reprepro >> /dev/null 2>&1; echo $?`
-if [[ ${REPREPRO} != 0 ]]; then
-    echo "ERROR: Could not find reprepro, verify that reprepro is installed on this machine.  Cannot continue!"
-    exit 1
-fi
+# Check the necessary packages before starting the build
+declare -a packages=( "reprepro" "devscripts" "debhelper" "libsoap-lite-perl" "libdbi-perl" "quilt" )
```
tingtli: @whowutwut  It seems we need other packages.
Here is my list to install a ubuntu build server
apt-get -y install build-essential dpkg-dev dh-make debhelper fakeroot gnupg lintian pbuilder quilt reprepro libsoap-lite-perl libdbi-perl

Anyway, if you think you packages are enough, I am ok. But I didn't try that.

whowutwut: @tingtli  Your list might be complete for everything that we used to be doing with the build-ubunturepo build script prior to the autobuild scripts.   The list that I created was ONLY for doing a Ubuntu build successfully in a Docker container... we can update this list as needed.

whowutwut:
@daniceexi any issue with the SSL_VERIFY_PEER changes?

daniceexi:
@whowutwut Yes, using the constant is a better than a number. But I think the library Net::SSLeay has been loaded automatically when using ssl certificates. So I guess we don't need to require Net::SSLeay explicitly.

daniceexi:
@whowutwut The change for Ubutnu 16.04 is just for packaging of xCAT, that means even without these changes xcat 2.11 still can support 16.04 (if we cannot find more issues for ubuntu 16.04 support).

whowutwut:
@daniceexi  I tried removing the `use Net::SSLeay` from Client.pm, and it results in this... 

```
root@fs1vm202:~# service xcatd restart 
root@fs1vm202:~# lsxcatd -a 
Undefined subroutine &Net::SSLeay::VERIFY_PEER called at /opt/xcat/lib/perl/xCAT/Client.pm line 18.
BEGIN failed--compilation aborted at /opt/xcat/lib/perl/xCAT/Client.pm line 18.
Compilation failed in require at /opt/xcat/bin/lsxcatd line 13.
BEGIN failed--compilation aborted at /opt/xcat/bin/lsxcatd line 13.
```

daniceexi:
@whowutwut I meant you can try to remove the call of Net::SSLeay::VERIFY_PEER too. 
The constant **SSL_VERIFY_PEER** should work in the Client.pm before your change. Do you think the **SSL_VERIFY_PEER** was not replaced with correct value in the original Client.pm? 

~~~~
 +use Net::SSLeay;
+use constant SSL_VERIFY_PEER => Net::SSLeay::VERIFY_PEER();
~~~~

whowutwut:
Originally in Client.pm , we are passing the string "SSL_VERIFY_PEER" into the ``start_SSL`` function.  I think the older versions can handle that and translate the string to a ENUM . 

With the new version of SSL.pm shipped with Ubuntu, they don't allow a string to be passed in, the error message ``SSL_verify_mode must be a number and not a string at /usr/share/perl5/IO/Socket/SSL.pm line 2166.`` comes out.. . So the function is used to get the value of the enum.   I think ``Net::SSLeay::VERIFY_PEER();`` returns 1.  

So we have to either define it ourselves as the constant 1, or get it from the SSLeay::VERIFY_PEER() function.  I think it's better to get from the function, in case they change something in the future.

daniceexi:
@whowutwut 
I think we always use the constant **SSL_VERIFY_PEER** instead of string in Client.pm. **SSL_VERIFY_PEER** was got from Net::SSLeay automatically since the Net::SSLeay should have been loaded when SSL function was used. So we don't need to assign it in our code explicitly. 

Using **SSL_VERIFY_PEER** as string in docker.pm was an incorrect usage.

whowutwut:
@daniceexi  Thanks, so the issue with Client.pm is that we do not have  ``use IO::Socket::SSL``.  We have some eval tags requiring it, but that does not set the SSL_VERIFY_PEER enum.  I made some changes to the pull request, can you review again .

daniceexi:
@whowutwut Thanks for the change, everything looks good now. Thanks for @tingtli 's review. I'll merge it.

Issue #810, opened 2016-03-17
[FVT]Failed to update booted status for ubuntu14.04.3 on ppc64le 


caomengmeng:
ENV:
MN:
```
[root@c910f02c01p17 autotest_code]# rpm -qa | grep -i xCAT
xCAT-client-2.12-snap201603081530.noarch
xCAT-2.12-snap201603081531.ppc64
ipmitool-xcat-1.8.11-3.ppc64
conserver-xcat-8.1.16-10.ppc64
xCAT-genesis-base-ppc64-2.11-snap201511160819.noarch
perl-xCAT-2.12-snap201603081530.noarch
xCAT-server-2.12-snap201603081530.noarch
xCAT-genesis-scripts-ppc64-2.12-snap201603081531.noarch
grub2-xcat-2.02-0.16.el7.snap201506090204.noarch
xCAT-buildkit-2.12-snap201603081530.noarch
[root@c910f02c01p17 autotest_code]# cat /etc/*release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.2 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.2"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.2 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.2:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.2
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.2"
Red Hat Enterprise Linux Server release 7.2 (Maipo)
Red Hat Enterprise Linux Server release 7.2 (Maipo)
```

Description:
In PPC64le autotest environment, it is failed to install the CN node.
The CN status is booting, the scripts for install are not executed completely, and logs in CN /var/log/xcat/xcat.log are as below:

```
Wed Mar 16 10:12:46 EDT 2016 Running postscript: syslog
initctl: Unable to connect to Upstart: Failed to connect to socket /com/ubuntu/upstart: Connection refused
initctl: Unable to connect to Upstart: Failed to connect to socket /com/ubuntu/upstart: Connection refused
[...]

PR #814, opened 2016-03-17
refine updateflag.awk

daniceexi:
Fix loop issue;
  print error message to /var/log/xcat/xcat.log


daniceexi:
I found the updateflag.awk does not work well when handle the issue #810, then I make this refine.

whowutwut:
@cxhong Is this related to the updateflag issues that you saw?

cxhong:
yeh..  XiaoPeng added more messages here in this file and logged to xcat.log

PR #817, opened 2016-03-17
Find switch vendor using snmpwalk command

cxhong:
For the RTC Task #70903.

switchdiscover is based on the nmap command to find switch information, like Mac address, hostname and vendors, but in the some environment,  the nmap couldn't find the vendor info.  In this task, it will add snmpwalk command to enhance switchdiscover and use OID.enterprises to match vendor info and sysName to find it's hostname.

The metrix for some OID and switches we are supporting as following;
```
enterprises.20301.1.18.13    IBM Flex System Fabric EN4093/EN2092 Scalable Switch
enterprises.26543.1.7.1       IBM Networking Operating System RackSwitch G8000-RS
enterprises.26543.1.7.4       IBM Networking Operating System RackSwitch G8124
enterprises.26543.1.7.6       IBM Networking Operating System RackSwitch G8264
enterprises.26543.1.7.7       IBM Networking Operating System RackSwitch G8052
enterprises.33049.1.1.1.6036  IBM Mellanox switch SX6036
enterprises.33049.1.1.1.6512  IBM Mellanox switch SX6512
enterprises.33049.2.1            IBM Mellanox IB switch IB6131
```

### DIFF
--- a/docs/source/guides/admin-guides/references/man1/switchdiscover.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/switchdiscover.1.rst
@@ -29,6 +29,8 @@ To view all the switches defined in the xCAT databasee use \ **lsdef -w "nodetyp
 
 For lldp method, please make sure that lldpd package is installed and lldpd is running on the xCAT management node. lldpd comes from xcat-dep packge or you can get it from http://vincentbernat.github.io/lldpd/installation.html.
 
+For snmp method, please make sure that snmpwalk command is installed and snmp is enabled for switches. To install snmpwalk, "yum install net-snmp-utils" for redhat and sles,  "apt-get install snmp" for Ubuntu.
+
 
 *******
 OPTIONS
@@ -60,6 +62,8 @@ OPTIONS
  Specify one or more IP ranges. Each can be an ip address (10.1.2.3) or an ip range (10.1.2.0/24). If the range is huge, for example, 192.168.1.1/8, the switch discover may take a very long time to scan. So the range should be exactly specified.
  
  For nmap scan method, it accepts multiple formats. For example, 192.168.1.1/24, 40-41.1-2.3-4.1-100.
+
+For snmp scan method, it accepts multiple formats.  For example: 192.168.1.1/24, 1.2-3.4.5, 1.2.3-4.5, 1.2.3.4-5
  
  If the range is not specified, the command scans all the subnets that the active network interfaces (eth0, eth1) are on where this command is issued.
  
@@ -74,7 +78,7 @@ OPTIONS
 \ **-s**\ 
  
  It is a comma separated list of methods for switch discovery. 
- The possible switch scan methods are: lldp and nmap. The default is nmap.
+ The possible switch scan methods are: lldp, nmap or snmp. The default is nmap.
  
 
 
--- a/docs/source/guides/admin-guides/references/man1/switchdiscover.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/switchdiscover.1.rst
@@ -61,9 +61,8 @@ OPTIONS
  
  Specify one or more IP ranges. Each can be an ip address (10.1.2.3) or an ip range (10.1.2.0/24). If the range is huge, for example, 192.168.1.1/8, the switch discover may take a very long time to scan. So the range should be exactly specified.
  
- For nmap scan method, it accepts multiple formats. For example, 192.168.1.1/24, 40-41.1-2.3-4.1-100.
+ For nmap and snmp scan method, it accepts multiple formats. For example, 192.168.1.1/24, 40-41.1-2.3-4.1-100.
 
-For snmp scan method, it accepts multiple formats.  For example: 192.168.1.1/24, 1.2-3.4.5, 1.2.3-4.5, 1.2.3.4-5
  
  If the range is not specified, the command scans all the subnets that the active network interfaces (eth0, eth1) are on where this command is issued.
  
--- a/docs/source/guides/admin-guides/references/man1/switchdiscover.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/switchdiscover.1.rst
@@ -23,7 +23,7 @@ DESCRIPTION
 ***********
 
 
-The switchdiscover command scans the subnets and discovers all the swithches on the subnets. The command takes a list of subnets as input. The default subnets are the ones that the xCAT management node is on. It uses nmap command to discover the switches. However, you can specify other discovery methods such as lldp with \ **-s**\  flag. You can write the discovered switches into xCAT database with \ **-w**\  flag. This command supports may output formats such as xml(\ **-x**\ ), raw(\ **-r**\ ) and stanza(\ **-z**\ ) in addition to the default format.
+The switchdiscover command scans the subnets and discovers all the swithches on the subnets. The command takes a list of subnets as input. The default subnets are the ones that the xCAT management node is on. It uses nmap command as default to discover the switches. However, you can specify other discovery methods such as lldp or snmp with \ **-s**\  flag. You can write the discovered switches into xCAT database with \ **-w**\  flag. This command supports may output formats such as xml(\ **-x**\ ), raw(\ **-r**\ ) and stanza(\ **-z**\ ) in addition to the default format.
 
 To view all the switches defined in the xCAT databasee use \ **lsdef -w "nodetype=switch"**\  command.
 
@@ -61,8 +61,7 @@ OPTIONS
  
  Specify one or more IP ranges. Each can be an ip address (10.1.2.3) or an ip range (10.1.2.0/24). If the range is huge, for example, 192.168.1.1/8, the switch discover may take a very long time to scan. So the range should be exactly specified.
  
- For nmap and snmp scan method, it accepts multiple formats. For example, 192.168.1.1/24, 40-41.1-2.3-4.1-100.
-
+ For nmap and snmp scan method, it accepts multiple formats. For example: 192.168.1.1/24, 40-41.1-2.3-4.1-100.
  
  If the range is not specified, the command scans all the subnets that the active network interfaces (eth0, eth1) are on where this command is issued.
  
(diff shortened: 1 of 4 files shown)


cxhong:
@bybai Can u review this?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -807,6 +826,68 @@ sub snmp_scan {
...
+      vendor:  vendor info  of the switch
+=cut
+#--------------------------------------------------------------------------------
+sub get_vendorinfo {
+    my $request = shift;
+    my $ip  = shift;
+    my $snmpwalk_vendor;
+
+
+    if (-x "/usr/bin/snmpwalk" ){
```
daniceexi: Does snmpwalk locate in the /usr/bin for all supported OS? rh, sles, Ubuntu ...

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -807,6 +826,68 @@ sub snmp_scan {
...
+        if (exists($globalopt{verbose}))    {
```
daniceexi: I am afraid this message should be displayed even without **-V**

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -807,6 +826,68 @@ sub snmp_scan {
...
+            send_msg($request, 0, "Discovering switch vendor using snmpwalk.....");
+        }
+    } else {
+        if (exists($globalopt{verbose}))    {
+            send_msg($request, 0, "snmpwalk is not available, please install snmpwalk command for better results");
+        }
+        return $snmpwalk_vendor;
+    }
+
+    my $ccmd = "snmpwalk -Os -v1 -c public $ip system | grep enterprises";
```
daniceexi: @cxhong Could you give a rough time that this command will take? Since it need to be run for each IP.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -807,6 +826,68 @@ sub snmp_scan {
...
+    }
+    my $search_string = join '|', keys(%global_OID_model_metrix);
+    if ($oid =~ /($search_string)/) {
+        $key = $1;
+        $snmpwalk_vendor = $global_OID_model_metrix{$key};
+        if (exists($globalopt{verbose}))    {
+            send_msg($request, 0, "find vendor = $snmpwalk_vendor from snmpwalk command\n");
+        }
+    } else {
+        if (exists($globalopt{verbose}))    {
```
daniceexi: @cxhong We have a list of mapping between **oid<->'vendor info'** for known vendors. How about directly display the vendor infor from the snmpwalk if cannot found it in the **%global_OID_model_metrix**?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -814,14 +895,34 @@ sub snmp_scan {
...
 sub get_hostname {
     my $host = shift;
     my $ip = shift;
 
+    if ($host) {
+        return $host;
+    }
+
+    # run snmpwalk to get hostname
+    my $ccmd = "snmpwalk -Os -v1 -c public $ip system | grep sysName";
```
daniceexi: Currently we are using snmp v1. Do we know that generally which type of swhich enabled snmp v1 by default, and which enabled snmp v3 by default?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -695,15 +709,20 @@ sub nmap_scan {
                             $mac="nomac_nmap_$counter";
                             $counter++;
                         }
-                        if ($addr->{vendor}) {
+                        my $vendor = $addr->{vendor};
+                        # run snmpwalk command to find vendor
+                        if (!$vendor) {
+                            $vendor = get_vendorinfo($request, $ip);
```
daniceexi: Do we want to enable snmp check in the **nmap** mode through a flag like ** --snmp**?
We were discussing to support snmp-based discovery ' like **nmap -s snmp**. Then how do we position the **-s nmap** and **-s snmp**? Do we want to merge them into one method?

cxhong: for this switchdiscover command, only nmap works so far, snmp didn't implemented and lldpd scan method implemented but don't think it's usdeful.   The default scan is nmap,  we can add few lines to documentation to describable which commands we used to discover the switches

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: yeh, the snmpwalk is located at /usr/bin.   but most xCAT MN didn't come with this command,  we have to yum install net-snmp-utils or apt-get install snmp.   Should we add this to pkglist for each osimage?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: snmpwalk has better performance than nmap.  I don't think this is a issue here.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
whowutwut: If It's required for this to work, I would suggest that we add a ``Requires:`` tag for the rpm in the rpm that ships this ``.pm`` so that the package is automatically installed.   For SLES, Is this something on the SDK or on the normal disc1?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
daniceexi: First, we need position the role of snmpwalk in the nmap-based switch discovery. What should we do when the snmpwalk is not installed? And what should we do when the snmp access failed?

To make the user interface be simple, is that acceptable that keep the original code logic for nmap-based only. And add your current code logic to be snmp-based discovery? (switchdiscover -s snmp)

cxhong:
I will restructure the code to snmp-based discovery.

cxhong:
@daniceexi , can u review the code?

docs/source/guides/admin-guides/references/man1/switchdiscover.1.rst
```diff
@@ -60,6 +62,8 @@ OPTIONS
  Specify one or more IP ranges. Each can be an ip address (10.1.2.3) or an ip range (10.1.2.0/24). If the range is huge, for example, 192.168.1.1/8, the switch discover may take a very long time to scan. So the range should be exactly specified.
  
  For nmap scan method, it accepts multiple formats. For example, 192.168.1.1/24, 40-41.1-2.3-4.1-100.
+
+For snmp scan method, it accepts multiple formats.  For example: 192.168.1.1/24, 1.2-3.4.5, 1.2.3-4.5, 1.2.3.4-5
```
daniceexi: I think the Ip range for **nmap** and **snmp** should be same.

docs/source/guides/admin-guides/references/man1/switchdiscover.1.rst
```diff
@@ -74,7 +78,7 @@ OPTIONS
 \ **-s**\ 
  
  It is a comma separated list of methods for switch discovery. 
- The possible switch scan methods are: lldp and nmap. The default is nmap.
+ The possible switch scan methods are: lldp, nmap or snmp. The default is nmap.
```
daniceexi: Since we does not support lldp well, should we remove the lldp method from the command help and man page?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -818,17 +804,98 @@ sub nmap_scan {
...
+
+    #################################################
+    # If --range options, take iprange, if noderange is defined
+    # us the ip addresses of the nodes. If none is define, use the
+    # subnets for all the interfaces.
+    ##################################################
+    my $ranges = get_ip_ranges($request);
+
+    #push each ip to ip list (need better code?)
+    foreach my $ips (@$ranges) {
```
daniceexi: @cxhong I am thinking that we can leverage nmap to test the snmp port first, and then use the snmp active node to run the snmp_scan. Advantages:
* save the code to handle the ip range
* the ip range format can be same as nmap_scan
* since nmap has a good perforamnce, we don't need to test snmp for each IP in the ip range.

daniceexi:
A general comment, could you add more verbose log for this command?

cxhong:
@daniceexi , as we discussed , we will not remove lldp method from the command.  I removed Ip address parse in the previous code, and use nmap to parse the ip address and find if snmp port is enabled .  the snmpwalk command only return information if snmp port is enabled.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -824,63 +823,46 @@ sub snmp_scan {
...
+    if ($::RUNCMD_RC != 0)
+    {
+        send_msg($request, 1, "Could not process this command: $ccmd" );
+        return 1;
+    }
+
+    #################################################
+    #display the raw output
+    #################################################
+    if (exists($globalopt{r})) {
```
daniceexi: Could we make this block to be:
~~~~
if (defined($globalopt{r}) || defined($globalopt{verbose}) {
    send_msg($request, 0, "$result\n" );
}
~~~~

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -824,63 +823,46 @@ sub snmp_scan {
...
+    } else {
+        if (exists($globalopt{verbose})) {
+            send_msg($request, 0, "$result\n" );
+        }
     }
+    my @lines = split /\n/, $result;
+
+    # only open port will be scan
+    foreach my $line (@lines) {
+        my $ip = `echo "$line\n" |awk '{printf \$6}'`;
```
daniceexi: Generally we don't call external command only if we must have to since call external cmd will fork another process. The efficiency is not good.
We could split the string and then use the index 6 against the output array.

daniceexi:
I have reviewed the change and agree to merge. Thanks @cxhong

PR #819, opened 2016-03-21
modify xCATreg for ubuntu's dns server could not work

junxiawang:
@hu-weihua  modify xCATreg for ubuntu's dns server could not work


hu-weihua:
I agree to merge

Issue #815, opened 2016-03-17
[FVT]Failed to choose correct install disk from multi-disks with WWN on X86(installing sles12.1)


caomengmeng:
ENV:
MN:
```
[root@c910f04x21 rh6.7]# rpm -qa  | grep -i xCAT
xCAT-genesis-scripts-x86_64-2.12-snap201603161531.noarch
syslinux-xcat-3.86-2.noarch
xCAT-genesis-base-x86_64-2.9-snap201504212134.noarch
perl-xCAT-2.12-snap201603161530.noarch
xCAT-server-2.12-snap201603161531.noarch
xCAT-buildkit-2.12-snap201603161531.noarch
elilo-xcat-3.14-4.noarch
conserver-xcat-8.1.16-10.x86_64
grub2-xcat-2.02-0.16.el7.snap201506090204.noarch
xCAT-client-2.12-snap201603161530.noarch
xCAT-2.12-snap201603161531.x86_64
ipmitool-xcat-1.8.11-3.x86_64
[root@c910f04x21 rh6.7]# cat /etc/*release
Red Hat Enterprise Linux Server release 6.7 (Santiago)
Red Hat Enterprise Linux Server release 6.7 (Santiago)
```

CN:
```
c910f04x20:/var/log/xcat # udevadm info --query=property --name=sda| grep WWN
ID_WWN=0x500a0751036d9e61
c910f04x20:/var/log/xcat # udevadm info --query=property --name=sdb| grep WWN
ID_WWN=0x500a0751036d9c24
c910f04x20:/var/log/xcat # udevadm info --query=property --name=sdc| grep WWN
ID_WWN=0x500a0751036d9e7a
c910f04x20:/var/log/xcat # udevadm info --query=property --name=sdd| grep WWN
ID_WWN=0x500a0751036d9e83
```

Description:
Set xcatdebugmode=1, then install sles12.1 on CN. 
After installed successfully, logon CN, track /var/log/xcat/xcat.log, then the /dev/sda is chosen as the install disk, but it should be the /dev/sdb

you will see the below logs:
[xcat_log_sles12_x86_1_fresh.txt](https://github.com/xcat2/xcat-core/files/177524/xcat_log_sles12_x86_1_fresh.txt)

caomengmeng:
Passed, correctly select disk.

PR #820, opened 2016-03-21
To fix issue 815, get the disk name with the smallest data

xuweibj:
#815


daniceexi:

PR #823, opened 2016-03-22
Based on Debian policy, all package names should not use upper-case letters

neo954:
Change depends xCAT-client -> xcat-client


neo954:
Related issue #790, #801.
Related pull request #802.

daniceexi:
I have reviewed the change and agreed to merge. Thanks. @neo954

PR #824, opened 2016-03-22
Do not kill db process if immediate child process exists

chenglch:
This change includes the following:

- Replace the signal number with signal name.
- Stop db process after all the immediate children stopped.
- Fix a defect in TableUtils when closing the table object.
- Enhance the is_process_exists subroutine in utils.

Close issue: #537


daniceexi:

PR #826, opened 2016-03-22
Add attribute 'syslog', avoid the syslog configuration for certain node

xuweibj:
Story 86265
If set 'syslog' as ignore, the syslog script will not run in postscript.

### DIFF
--- a/perl-xCAT/xCAT/Schema.pm
+++ b/perl-xCAT/xCAT/Schema.pm
@@ -1,5 +1,6 @@
 # IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
 package xCAT::Schema;
+
 BEGIN
 {
     $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : -d '/opt/xcat' ? '/opt/xcat' : '/usr';
@@ -19,11 +20,11 @@ use xCAT::ExtTab;
 #  as attributes under any conditions:
 #  http://www.postgresql.org/docs/8.3/static/sql-keywords-appendix.html
 #
-#  Current SQL reserved words being used in this Schema with special 
+#  Current SQL reserved words being used in this Schema with special
 #  processing are the
 #  following:
-#   
-#Word     Table                   Databases that will not allow 
+#
+#Word     Table                   Databases that will not allow
 # key      site,passwd,prodkey,monsetting      MySQL, DB2,SQL Server 2000
 # dump     nimimage                            SQL Server 2000 (microsoft)
 # power    nodehm                              SQL Server 2000
@@ -31,7 +32,7 @@ use xCAT::ExtTab;
 # parameters  policy              DB2,SQL Server Future Keywords,ISO/ANSI,SQL99
 # time        policy              DB2,SQL Server Future Keywords,ISO/ANSI,SQL99
 # rule        policy              SQL Server 2000
-# value       site,monsetting     ODBC, DB2, SQL Server 
+# value       site,monsetting     ODBC, DB2, SQL Server
 #                                 Future Keywords,ISO/ANSI,SQL99
 #
 # !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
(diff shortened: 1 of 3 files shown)


xCAT/postscripts/syslog
```diff
@@ -13,6 +13,10 @@
 #           OSVER: possible values are sles10, fedora8, redhat5, aix etc.   
 #=cut
 #-------------------------------------------------------------------------------
+if [ "$SYSLOG" = "ignore" ]; then
+    exit 0
+fi
```
zet809: If the value is not "ignore", will still run this script?

perl-xCAT/xCAT/Schema.pm
```diff
@@ -668,6 +668,7 @@ noderes => {
   routenames => 'A comma separated list of route names that refer to rows in the routes table. These are the routes that should be defined on this node when it is deployed.',
   nameservers => 'An optional node/group specific override for name server list.  Most people want to stick to site or network defined nameserver configuration.',
   proxydhcp => 'To specify whether the node supports proxydhcp protocol. Valid values: yes or 1, no or 0. Default value is yes.',
+  syslog => "Set to 'ignore', will not run syslog script in postscript",
```
zet809: I think you can add more comments here, such as it is used to configure syslog for the compute node, currently, only 'ignore' is supported.

perl-xCAT/xCAT/Schema.pm
```diff
@@ -1846,6 +1847,10 @@ my @nodeattrs = (
                  tabentry => 'noderes.proxydhcp',
                  access_tabentry => 'noderes.node=attr:node',
   },
+        {attr_name => 'syslog',
```
zet809: Pls keep the format to be consistent.

perl-xCAT/xCAT/Schema.pm
```diff
@@ -668,6 +668,7 @@ noderes => {
   routenames => 'A comma separated list of route names that refer to rows in the routes table. These are the routes that should be defined on this node when it is deployed.',
   nameservers => 'An optional node/group specific override for name server list.  Most people want to stick to site or network defined nameserver configuration.',
   proxydhcp => 'To specify whether the node supports proxydhcp protocol. Valid values: yes or 1, no or 0. Default value is yes.',
+  syslog => "It is used to configure syslog for the compute node. Currently, only 'ignore' is supported, which means will not run syslog script in postscript",
```
daniceexi: I suggest the message to be:
~~~~
To configure how to configure syslog for compute node. Valid values:blank, ignore. blank - run postscript syslog.; ignore - Do NOT run postscript syslog;
~~~~

zet809:

whowutwut:
@xuweibj  In the description of this pull request, can you add the external usage for this new option.  
For users of xCAT, how would they set it? 

I also think we should have FVT create a testcase for this to verify that syslog is executed or not executed.

whowutwut:
Another comment I have... with the changes to Schema.pm, why is there so much differences? Did you run perl tidy?  If we are doing any type of tidy code, we should put that first into it's own commit, then have code changes in another commit.  There's no way to review the changes when git cannot display the diffs .   @daniceexi  FYI

whowutwut:
@zet809  looking at your comments before and my email notifications, it looks like the commit was done using the --amend and caused over 3,000 changes to Schema.pm.. what happened here? 

Is this really what we should be doing once we start a code review? It's very difficult for us to see the changes made from the review comments...

daniceexi:
@xuweibj @zet809 @whowutwut I totally agree, for the change of tidying a file, it must have an individual commit so that we can just review it quickly to make sure there's no format issue.

whowutwut:
with this changes in Schema.pm, this resulted in changes to the .rst files when running ``create_man_pages.py``..  I think we need to build the rst file and look at the resulting output. 
```
[vhu@victorvm7 xcat-core]$ git status
# On branch master
# Changes not staged for commit:
#   (use "git add <file>..." to update what will be committed)
#   (use "git checkout -- <file>..." to discard changes in working directory)
#
#	modified:   docs/source/guides/admin-guides/references/man5/mpa.5.rst
#	modified:   docs/source/guides/admin-guides/references/man5/noderes.5.rst
#	modified:   docs/source/guides/admin-guides/references/man7/group.7.rst
#	modified:   docs/source/guides/admin-guides/references/man7/node.7.rst
```

I'm not sure if the changes are correct...   I quickly looked at a diff for one page..  mpa.rst.
```
[vhu@victorvm7 xcat-core]$ git diff docs/source/guides/admin-guides/references/man5/mpa.5.rst
diff --git a/docs/source/guides/admin-guides/references/man5/mpa.5.rst b/docs/source/guides/admin-guides/references/man5/mpa.5.rst
index f06f0bb..e44e0f8 100644
--- a/docs/source/guides/admin-guides/references/man5/mpa.5.rst
+++ b/docs/source/guides/admin-guides/references/man5/mpa.5.rst
@@ -63,9 +63,9 @@ mpa Attributes:
 \ **slots**\ 
  
  The number of available slots in the chassis. For PCM, this attribute is used to store the number of slots in the following format:  <slot rows>,<slot columns>,<slot orient
-    <slot rows>  = number of rows of slots in chassis
-    <slot columns> = number of columns of slots in chassis
-    <slot orientation> = set to 0 if slots are vertical, and set to 1 if slots of horizontal
+                  <slot rows>  = number of rows of slots in chassis
+                  <slot columns> = number of columns of slots in chassis
+                  <slot orientation> = set to 0 if slots are vertical, and set to 1 if slots of horizontal
  
```
I didn't build the html man pages but that doesn't look like it would render right..    maybe it will..  @xuweibj  can you double check?

daniceexi:
@whowutwut @xuweibj made the perltidy against the Schema.pm so that there were some indent changes in the description string. Then it caused the change in the map page.

PR #828, opened 2016-03-22
Fix issue 732: Error to process CMD and Entrypoint in Docker life cyc…

zet809:
…le management


daniceexi:
I have reviewed the change and agreed to merge.

PR #829, opened 2016-03-22
Add Ubuntu 14.04.4 support

neo954:


daniceexi:
I wil review it when @neo954 finished the test.

whowutwut:
@neo954  can you provide some description about this in the top so I have some background...    I'm wondering why we need to go down to the minor release level, 14.04.04?  Can we just create 14.04 generically that should work on all the minor release, then if we find that 14.04.05 needs something specific, we can override with a more specific file. This way we do not need to create new files for every OS support. 

We don't do this for RHEL or SLES, so I think we should not for Ubuntu.  For RHEL we have rhels7 file.... not rhels7.1, rhels7.2 .etc...

daniceexi:
@whowutwut In our current code logic, if the os version is xxx.yyy, the code would check the existence of xxx.yyy first. If failed, then check the existence of xxx. Apply this logic to 'Ubuntu14.04.4', the xxx is Ubuntu14 and yyy is '04.4'. But obviously the template file ubutnu14 cannot satisfy the ubuntu14.04.4. 
BTW, generally if there are common template files for different release, we just create a symbol link which saves space and easy to manage.
 
@neo954 Looked like you did not use symbol link for these template files.

whowutwut:
@daniceexi how come the ubuntu14 cannot apply to all ubuntu14.* because it's the similar concept for rhels7.1, xxx=7 and yyy=1 but in the template directory, we only have files like compute.rhels7.tmpl.

daniceexi:
@whowutwut ubuntu14 does can apply to ubuntu14.*. But it can only cover one pkglist cfg. For example, if ubuntu14.04 and ubuntu14.04.1 use **pkg1**, ubuntu 14.04.2 and ubuntu 14.04.3 use **pkg2**, do you want ubuntu14 to be **pk1** or **pk2**?
Current logic is ubuntu14 keeps as the early version of OS, in this case it should be **pkg1**, then for ubuntu14.04.4, we have to make link to ubuntu14.04.3.
I think you point is to make the ubuntu14 to be the common cfg for latest release. That means to create the pkglist for ubuntu14.04 and ubuntu 14.04.1, and make the ubuntu14 to be the one of ubuntu14.04.2 so that we don't need to make specific pkglist for ubuntu14.04.3/4 since ubuntu14 can cover it.

whowutwut:
@daniceexi  Right, since our package lists are pretty generic, unless there's something specific that is not provided in Ubuntu 14.X, it's easier to maintain if we first create ubuntu14 file, then only add a file that overrides if there's something different in 14.04.3 or 14.04.4.. if not, then we don't need to keep creating files for each minor release, just the major release.

daniceexi:

PR #832, opened 2016-03-22
Fix some issues for xCAT on Power 775

billwaj:
request pull for three commits ...


daniceexi:
I am OK with the change. Thanks @billwaj

Issue #822, opened 2016-03-22
[FVT]Install redhat6.7, failed to get the determine install disk logs


caomengmeng:
ENV:
 MN:
```
[root@c910f04x21 rh6.7]# rpm -qa  | grep -i xCAT
xCAT-genesis-scripts-x86_64-2.12-snap201603161531.noarch
syslinux-xcat-3.86-2.noarch
xCAT-genesis-base-x86_64-2.9-snap201504212134.noarch
perl-xCAT-2.12-snap201603161530.noarch
xCAT-server-2.12-snap201603161531.noarch
xCAT-buildkit-2.12-snap201603161531.noarch
elilo-xcat-3.14-4.noarch
conserver-xcat-8.1.16-10.x86_64
grub2-xcat-2.02-0.16.el7.snap201506090204.noarch
xCAT-client-2.12-snap201603161530.noarch
xCAT-2.12-snap201603161531.x86_64
ipmitool-xcat-1.8.11-3.x86_64
[root@c910f04x21 rh6.7]# cat /etc/*release
Red Hat Enterprise Linux Server release 6.7 (Santiago)
Red Hat Enterprise Linux Server release 6.7 (Santiago)
```

Description: 
Set site.xcatdebugmode=0, install rh6.7 on CN, after successful installation, then track logs in /var/log/xcat/xcat.log.
There isn't determining install disk logs shown:
```
[root@c910f04x20 xcat]# vi xcat.log
Running Kickstart Pre-Installation script...
mknod: `/dev/vcs1': File exists
mknod: `/dev/vcsa1': File exists
mknod: `/dev/vcs2': File exists
mknod: `/dev/vcsa2': File exists
mknod: `/dev/vcs3': File exists
mknod: `/dev/vcsa3': File exists
mknod: `/dev/vcs4': File exists
mknod: `/dev/vcsa4': File exists
mknod: `/dev/vcs5': File exists
mknod: `/dev/vcsa5': File exists
mknod: `/dev/vcs6': File exists
mknod: `/dev/vcsa6': File exists
mknod: `/dev/vcs7': File exists
mknod: `/dev/vcsa7': File exists
mknod: `/dev/vcs8': File exists
mknod: `/dev/vcsa8': File exists
installstatus installing
BOOT_IMAGE=net13:/tftpboot/xcat/osimage/rhels6.7-x86_64-install-compute/vmlinuz
[...]

PR #835, opened 2016-03-23
To fix issue 822, run getinstdisk script when install rh6.7

xuweibj:
#822 
Run getinstdisk script when install rh6.7
Use 'sed' in place of 'tr'.


zet809:

Issue #756, opened 2016-03-01
ubuntu 14.04.2 installation boot loop when one more pkglist is added


wwzhubj:
**Issue: **When I added below pkglist, the location is at  9.114.39.157: /install/custom/install/compute.pkglist.full, the installation can not 100% finish, when it goes to 95% , system reboots and goes back to start from xCAT installation, in an endless loop(install, reboot, install reboot...)
When I removed the pkglist, only use the /opt/xcat/share/xcat/install/ubuntu/compute.ubuntu14.04.2.ppc64le.pkglist, it can be installed successfully. 
```
_root@c656ems2:~# cat /install/custom/install/compute.pkglist.full
# nfs and automountd
#nfs-common
autofs
# package management
apt-file
dpkg
debhelper
dkms
dpatch
# useful tools
lsvpd
curl
libcurl4-gnutls-dev
# Debug
apport
gdb
kdump-tools
crash
# make, compile and development tools
autotools-dev
autoconf
libtool
automake1.10
automake
m4
tcl
tcl8.4
chrpath
swig
graphviz
tcl-dev
tcl8.4-dev
tk-dev
tk8.4-dev
bison
flex
zlib1g-dev
python
libxml2
libvirt-bin
libvirt0
libnl-dev
libglib2.0-dev
libgfortran3
gcc
make
# Performance
numactl
libnuma1
hugepages
#linux-tools
ksh
at
build-essential
#linux-headers-generic
#linux-image-generic
zlib1g-dev
cpufrequtils

```

wwzhubj:
the ems is 9.114.39.157
the image I use is: oldcompilers
cn is :c656f6n01

daniceexi:
@penguhyang Please take a look of this issue in @wwzhubj 's environment.

immarvin:
hi, I looked into this problem on @wwzhubj 's env. There are 2 problems here:
1) the size of "/var/lib/dhcpd/dhcpd.leases" file was around 60M+, this made the dhcp response from MN too slow and caused timeout on the dhcp request from compute nodes. In the /var/log/message:
````
...
und
Mar  2 01:59:10 c656ems2 dhcpd: DHCPDISCOVER from 00:0a:f7:73:7d:d0 via eth0
Mar  2 01:59:10 c656ems2 dhcpd: DHCPOFFER on 9.114.39.101 to 00:0a:f7:73:7d:d0 via eth0
[...]

PR #843, opened 2016-03-24
add support for multiple pkglist for ubuntu

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/756

### DIFF
--- a/xCAT-server/lib/perl/xCAT/Postage.pm
+++ b/xCAT-server/lib/perl/xCAT/Postage.pm
@@ -1588,6 +1588,10 @@ sub get_envlist
 sub get_pkglist_tex
 {
     my $allfiles_pkglist   = shift;
+    if($allfiles_pkglist =~ "xCAT::"){
+       $allfiles_pkglist   = shift;
+    }
+
     my $allfiles_pkgtext;
     foreach my $pkglist (split(/,/, $allfiles_pkglist))
     {
@@ -1657,7 +1661,7 @@ sub includefile
         $file = $idir . "/" . $file;
     }
 
-    open(INCLUDE, $file) || \return "#INCLUDEBAD:cannot open $file#";
+    open(INCLUDE, $file) || return "#INCLUDEBAD:cannot open pkglist file $file#";
 
     while (<INCLUDE>)
     {
--- a/xCAT-server/lib/perl/xCAT/Template.pm
+++ b/xCAT-server/lib/perl/xCAT/Template.pm
@@ -134,51 +134,57 @@ sub subvars {
   }
 
   if (("ubuntu" eq $platform) || ("debian" eq $platform)) {
-    # since debian/ubuntu uses a preseed file instead of a kickstart file, pkglist
-    # must be included via simple string replacement instead of using includefile()
+      # since debian/ubuntu uses a preseed file instead of a kickstart file, pkglist
+      # must be included via simple string replacement instead of using includefile()
 
-    # the first line of $pkglistfile is the space-delimited package list
-    # the additional lines are considered preseed directives and included as is
+      # the first line of $pkglistfile is the space-delimited package list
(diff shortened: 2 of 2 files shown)


whowutwut:
@immarvin  Can we add some description to the pull requests? For example if we can have a small description describing the changes, it will be much easier for the reviewer to review the code with some context.  The issue describes the problem but does not talk about the intended solution.  In the issue we discuss about printing an error message, but it looks like this is not what is being done in this pull request.   If we can show the example ERROR preseed and what the preseed will look like after. 

If i understand the code right, we are going to parse through the include file and fix the error so that it's correct for Ubuntu preseed file?  I don't know if we should do that because it masks the format of a preseed file....

immarvin:
hi @whowutwut , what do you mean by "masks the format of a preseed file"?

daniceexi:
@immarvin I known in the xcat, it's not easy always to display an error/warning message in any code. But it's easy to call the **xCAT::MsgUtils->trace** to generate a message to xcat/cluster.log. Could you add some trace log for the code you changed?

whowutwut:
@immarvin  Sorry, I'm not familiar with the preseed file for ubuntu..  so maybe my understanding is wrong.   In your comments in the issue...  you said that the user should create a new package list file, with multiple includes.   So this is what the code is parsing through right? 

How did we get into this scenario where xCAT leaves a line in the preseed file after nodeset? 
````
d-i pkgsel/include string #INCLUDE_DEFAULT_PKGLIST_PRESEED#
````

I think in that scenario we need to put out an error message.  I don't see that in the pull request.  Is this very different behavior than on RHEL/SLES?

immarvin:
hi @whowutwut , I agree that we should throw a message out if the pkglists in osimage.pkglist or "#INCLUDE xxxxxxxxx#" does not exist.

My fix is to support multiple pkglist files in ubuntu. The osimage.pkglist can be a list of pkglist files seperated with ",". "nodeset" will expand the "#INCLUDE xxx#" entries in all the specified pkglist files, then  substitute  "#INCLUDE_DEFAULT_PKGLIST_PRESEED#" in preseed template with the package list specified in the pkglist files.  Error message will be thrown out when the specified pkglist cannot be opened.

immarvin:
I have updated the code

immarvin:
Currently, the multiple package list is not supported on Redhat and Sles, open a ticket https://github.com/xcat2/xcat-core/issues/851 for redhat and sles

daniceexi:
@immarvin I am not sure the root cause why @chenglch made a new subroutine **debian_includefile** for the debian specific. Invite @chenglch to review this change.

immarvin:
hi @daniceexi , is @cxhong made the change?

daniceexi:
@immarvin right, the subroutine **debian_includefile** made by @chenglch.

daniceexi:
@immarvin  @chenglch  Sorry, My fault, should be @cxhong!

cxhong:
debian_includefile was a fix for this issue:
https://sourceforge.net/p/xcat/bugs/4475/

I will try to recreate this....

daniceexi:
@cxhong Did you make any progress?

cxhong:
I ran both diskfull and diskless test on my test environment: with current code , and didn't hit any issues with  this compute pkglist file :

````
# cat compute.pkg
bash
nfs-common
openssl
isc-dhcp-client
libc-bin
openssh-server
openssh-client
wget
vim
ntp
rsync
busybox-static
gawk
dnsutils
#INCLUDE:/tmp/include_pkg#
#INCLUDE:/tmp/include1_pkg#
````

daniceexi:
@cxhong Thanks for your effort. If no issue found, we assume @immarvin 's change work for Ubuntu too.

@tingtli Could you confirm do we have test case to cover the #INCLUDE function in the pkglist?

daniceexi:
Then I agree to merge it now.

PR #848, opened 2016-03-25
refine the doc on dhcp in large cluster; correcta typo

immarvin:
refine the doc on dhcp in large cluster; 
correct a typo 

see http://10.3.5.21/doc/advanced/hierarchy/configure_dhcp.html


daniceexi:
I am ok with the change.

Issue #496, opened 2015-12-01
[Customer]xCAT 2.11: lsf_startup use incorrect shell profile file on RHEL 7.2


mfoliveira:
xCAT 2.11

After a succesful execution of lsf_started (see PR #495), the import of LSF profile is done via "~/.profile", which doesn't work on RHEL 7.2, as it uses "~/.bash_profile".

```
[root@xcat-mn lsf]# updatenode $lsf_master --scripts lsf_startup
p8r1n1: xcatdsklspost: downloaded postscripts successfully
p8r1n1: Mon Nov 30 21:52:19 EST 2015 Running postscript: lsf_startup
<...>
p8r1n1: INFO: Set LSF environment for root and LSF_ADMINS
<...>
p8r1n1: Postscript: lsf_startup exited with code 0
p8r1n1: Running of postscripts has completed.
```

The LSF commands don't work by default. It's required to source the LSF conf from .bash_profile:

```
[root@xcat-mn lsf]# xdsh $lsf_master "su -l $lsf_username -c lsclusters"
p8r1n1: -bash: lsclusters: command not found

[root@xcat-mn lsf]# xdsh $lsf_master "su -l $lsf_username -c 'cat ~/.profile >> ~/.bash_profile'"

[root@xcat-mn lsf]# xdsh $lsf_master "su -l $lsf_username -c lsclusters"
p8r1n1: CLUSTER_NAME   STATUS   MASTER_HOST               ADMIN    HOSTS  SERVERS
p8r1n1: lsf-cluster    ok       p8r1n1                 lsfadmin        1        1
```

bybai:
Hi @mfoliveira, since you have workaround for this issue, and 2.11 entered invite mode, move it into 2.12. Will handle PR #495 in 2.12

PR #849, opened 2016-03-25
fix issue 496 lsf_startup use incorrect shell profile file on RHEL 7.2

bybai:
fix issue #496 
lsf_startup use incorrect shell profile file on RHEL 7.2 #496

### DIFF
--- a/xCAT/postscripts/lsf_startup
+++ b/xCAT/postscripts/lsf_startup
@@ -59,7 +59,14 @@ for lsfnode in $ALL_LSF_NODES
 do
     if [[ x${lsfnode} == x$NODE ]]
     then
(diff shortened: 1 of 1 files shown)


whowutwut:
Are these settings something that all LFS users should be able to get into their profile?

daniceexi:
Sorry @whowutwut, I did some discussion with @bybai before, but did not put it as a comment.

I thought the current fix might satisfy the requirement of installing LSF. But there are still two things we need to figure out:

* What we should do for the upcoming OS like rh7.3?
* We need to figure out the reason why rh7.2 must use /root/.bash_profile instead of /root/.profile.

neo954:
Cool patch, will merge this one.

Issue #694, opened 2016-02-02
[FVT] Disable root permission in policy table, the chtab command still work well.


caomengmeng:
ENV:

```
[root@c910f02c04p07 autotest]# cat /etc/*release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.1 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.1"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.1 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.1:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.1
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.1"
Red Hat Enterprise Linux Server release 7.1 (Maipo)
Red Hat Enterprise Linux Server release 7.1 (Maipo)
[root@c910f02c04p07 autotest]# rpm -qa | grep -i xCAT
conserver-xcat-8.1.16-10.ppc64
xCAT-client-2.12-snap201601281930.noarch
xCAT-2.12-snap201601281931.ppc64
grub2-xcat-2.02-0.16.el7.snap201506090204.noarch
xCAT-genesis-base-ppc64-2.11-snap201511160819.noarch
syslinux-xcat-3.86-2.noarch
perl-xCAT-2.12-snap201601281930.noarch
xCAT-server-2.12-snap201601281930.noarch
xCAT-buildkit-2.12-snap201601281930.noarch
xCAT-test-2.12-snap201601281930.noarch
ipmitool-xcat-1.8.11-3.ppc64
xCAT-genesis-scripts-ppc64-2.12-snap201601281931.noarch
```
Steps to reproduce:
1. disable root permission 
2. run xcat command, for example: lsdef. Error message appears for permission error.
3. run XCATBYPASS=YES with xcat command, for example: XCATBYPASS=YES lsdef, it works.
4. run "chtab name=root policy.rule=allow" to change the policy table, it works too.
But this is not excepted, without XCATBYPASS=YES, all the xcat command will be failed.
```
[...]

PR #850, opened 2016-03-25
fix issue 694: refine chtab's man page

hu-weihua:
Fix issue #694, Disable root permission in policy table, the chtab command still work well.

It don't need to change code, chtab command is designed to work without ``xcatd``, so it's out of control of ``policy`` mechanism.  

I refine the ``chtab`` man page to specify  it.


caomengmeng:
Yes, this explains well.

whowutwut:
whoops , forgot to run the create_man_pages.py for this .pod file, i pushed the changes into master directly aee2b62c25af95e8410dafa401634116d7228031

daniceexi:
@whowutwut Looks like it's hard for all people to remember to run the create_man_pages.py to create .rst file when changing the pod file or Schema.pm. Could you run the create_man_pages.py at end of the build process, if there are any .rst files need be updated, the build script could send an email to **xcat all** to require an update.

whowutwut:
Yea let me look into this

PR #853, opened 2016-03-25
Fix issue to acess db in getadapter

chenglch:
Previously the database access in getadapter ignore
the group cases, this patch intent to fix it.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/getadapter.pm
+++ b/xCAT-server/lib/xcat/plugins/getadapter.pm
@@ -521,7 +521,7 @@ sub do_inspect {
     my $callback       = shift;
     my $subreq         = shift;
     my @nodes          = @{$nodesptr};
-    my %updates;
+    my $updates;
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/getadapter.pm
```diff
@@ -682,14 +679,12 @@ sub inspect_adapter {
...
         my @checked_nodes;
-        foreach my $checked (@entries) {
-            push(@checked_nodes, $checked->{'node'});
+        foreach my $tmp (keys %{$entries}) {
+            if($entries->{$tmp}->[0]->{nicsadapter}) {
+                push(@checked_nodes, $tmp);
+            }
         }
 
         foreach my $node (@{ $request->{node} }) {
```
zet809: We can check whether the node have attributes in nics table in this loop, the loop in line 684 is useless.

zet809:

Issue #730, opened 2016-02-22
[Customer] Missed the /proc file system in the chroot environment when running genimage


daniceexi:
A customer complained the genimage for RH has problem. When install rpm to root image, certain rpms display that missing of /proc FS.

We had  a lot of disucssion of how to handle the issue that rpm installing complains the missing of /proc, /sys  and /dev FS in the chroot env. The basic thought was to mount the /proc, /sys and /dev from host OS to the chroot root image directory and unmount them after the rpm install. But this method has two disadvantages: 

* The rpm might write something to the /proc /sys or /dev which were mounted from the host OS. That means the running of geniamge might impact the host OS.
* If genimage command stopped unusual, the /proc /sys or /dev won't be umounted before exiting genimage command. Then the subsequent operation against the root image dir might impact the host OS.

@immarvin mentioned that the safer way is to mount the /proc /dev and /sys from host OS in readonly so that anything happens in the root image won't impact the host OS. And this method has been used in the genimage for Ubuntu. We should apply it for RH and SLES too.

~~~~
  Installing : 1:telnet-0.17-48.el6.x86_64 
 224/224
cat: /proc/cmdline: No such file or directory
cat: /proc/cmdline: No such file or directory
or
warning: %posttrans(nfs-utils-1:1.2.3-64.el6.x86_64) scriptlet failed, 
exit status 6
Non-fatal POSTTRANS scriptlet failure in rpm package 
1:nfs-utils-1.2.3-64.el6.x86_64
/proc issue : is /proc bind-mounted or something like that while 
~~~~

PR #855, opened 2016-03-26
mount /dev /sys and /proc from host readonly during genimage for redhat;correct a typo in ubuntu genimage

immarvin:
fix https://github.com/xcat2/xcat-core/issues/730

mount /dev /sys and /proc from host readonly during genimage for redhat;
correct a typo in ubuntu genimage


daniceexi:
The change looks good for me. Thanks @immarvin

Issue #857, opened 2016-03-28
[FVT]Failed to install xCAT on sles11 x86_64 using latest daily build


caomengmeng:
Steps to reproduce:
When in MN, using the latest xcat-core build in /xcat/build/xcat2_autobuild_daily_builds/20160327.0730/linux_rpm/devel
and xcat-dep: xcat-dep-201603231124.tar.bz2
After execute the "./mklocalrepo.sh", then run "zypper install xCAT" to install xCAT, but you can see below:
```
c910f04x27v05:/xcat-core # zypper install xCAT
Loading repository data...
Reading installed packages...
Resolving package dependencies...

Problem: xCAT-2.12-snap201603270731.x86_64 requires perl(xCAT::MsgUtils), but this requirement cannot be provided
  uninstallable providers: perl-xCAT-4:2.12-snap201603270730.noarch[xcat-2-core]
 Solution 1: do not install xCAT-2.12-snap201603270731.x86_64
 Solution 2: break xCAT-2.12-snap201603270731.x86_64 by ignoring some of its dependencies

Choose from above solutions by number or cancel [1/2/c] (c):
```

Test ENV:
MN:
```
c910f04x27v05:/ # cat /etc/*release
LSB_VERSION="core-2.0-noarch:core-3.2-noarch:core-4.0-noarch:core-2.0-x86_64:core-3.2-x86_64:core-4.0-x86_64"
NAME="SLES"
VERSION="11.4"
VERSION_ID="11.4"
PRETTY_NAME="SUSE Linux Enterprise Server 11 SP4"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:11:4"
SUSE Linux Enterprise Server 11 (x86_64)
VERSION = 11
PATCHLEVEL = 4
```

Log:in /var/log/xcat/install_xcat.log
```
Loading repository data...
Reading installed packages...
Resolving package dependencies...

Problem: xCAT-2.12-snap201603270731.x86_64 requires perl(xCAT::MsgUtils), but this requirement cannot be provided
  uninstallable providers: perl-xCAT-4:2.12-snap201603270730.noarch[xCAT-core]
 Solution 1: do not install xCAT-2.12-snap201603270731.x86_64
 Solution 2: break xCAT-2.12-snap201603270731.x86_64 by ignoring some of its dependencies

Choose from above solutions by number or cancel [1/2/c] (c): c
```

immarvin:
[...]

PR #865, opened 2016-03-29
Add check logic before using perl-Error library

chenglch:
perl-Error lib is not included in the system package of sles 11.4
by default. Load Error lib with eval and require statement to avoid
of dependency error.

Close issue: #857


zet809:

PR #866, opened 2016-03-29
modify p8le test case for mini.iso

junxiawang:
@hu-weihua please help review


xCAT-test/autotest/testcase/installation/ubuntu_full_installation_p8le
```diff
@@ -4,13 +4,14 @@ stop=yes
 
 cmd:copycds $$ISO
 check:rc==0
-cmd:makedns -n
+cmd:if [[ "__GETNODEATTR($$CN,os)__" =~ "ubuntu" ]];then mkdir /tmp/iso; mount -o loop $$MINIISO /tmp/iso ;  mkdir -p  /install/__GETNODEATTR($$CN,os)__/__GETNODEATTR($$CN,arch)__/install/netboot; cp  /tmp/iso/install/initrd.gz  /install/__GETNODEATTR($$CN,os)__/__GETNODEATTR($$CN,arch)__/install/netboot;fi
```
whowutwut: I would suggest also adding 
```
umount /tmp/iso; rmdir /tmp/iso
```
at the end of this block 

Is the xCAT mgmt node already re-installed on each run of this testcase?  If not, then we should probably make the mount point more unique...  will we ever hit a scenario where the mount fails? Or the mini.iso file is invalid?

xCAT-test/autotest/testcase/installation/ubuntu_full_installation_p8le
```diff
@@ -4,13 +4,14 @@ stop=yes
...
 check:rc==0
-cmd:sleep 60
 cmd:makedhcp -n
 check:rc==0
 cmd:makedhcp -a
```
whowutwut: Just a general question,   why do we need to run -n and -a for makedhcp? 
```
makedhcp -n
makedhcp -a
```

junxiawang: @whowutwut  just want to test both makedhcp -n and -a command

xCAT-test/autotest/testcase/installation/ubuntu_full_installation_p8le
```diff
@@ -27,4 +28,8 @@ check:output=~booted
 cmd:xdsh $$CN date
 check:rc==0
 check:output=~\d\d:\d\d:\d\d
+cmd:umount /tmp/iso
+check:rc==0
+cmd:rmdir /tmp/iso
+check:rc==0
```
hu-weihua: I agree with @whowutwut,  @junxiawang  you should  ``umount /tmp/iso`` and ``rmdir /tmp/iso``. But I don't think you should add these two operations in the end of test case.  What you want to do is copy ``initrd.gz`` out, so after ``cp`` operation, ``/tmp/iso`` is already useless. So these 2 operations should add behind ``cp``.

hu-weihua: Maybe it should be looked like this

      cmd:if [[ "__GETNODEATTR($$CN,os)__" =~ "ubuntu" ]];then mkdir /tmp/iso; mount -o loop $$MINIISO /tmp/iso ;  mkdir -p  /install/__GETNODEATTR($$CN,os)__/__GETNODEATTR($$CN,arch)__/install/netboot; cp  /tmp/iso/install/initrd.gz  /install/__GETNODEATTR($$CN,os)__/__GETNODEATTR($$CN,arch)__/install/netboot; umount /tmp/iso; rmdir /tmp/iso; fi

PR #875, opened 2016-03-31
Make genesis to include bmc ip in the findme package

xuweibj:
Get bmcip when CN run dodiscovery script and save in findme package.
Update the bmc ip only when the matched pre-defined node has NOT bmc ip was set.

### DIFF
--- a/xCAT-genesis-scripts/bin/dodiscovery
+++ b/xCAT-genesis-scripts/bin/dodiscovery
@@ -132,6 +132,15 @@ fi
 if [ "$PLATFORM" != "unknown" ]; then
         echo "<platform>$PLATFORM</platform>" >> /tmp/discopacket
 fi
+
+IsStatic=`ipmitool lan print 1 | grep 'IP Address Source' | grep 'Static'`
+if [ "$IsStatic" ]; then
+    BMCIPADDR=`ipmitool lan print 1 | grep 'IP Address' | grep -v 'IP Address Source' | cut -d ":" -f2 | sed 's/ //'`
+    if [ "$BMCIPADDR" ]; then
+        echo "<bmc>$BMCIPADDR</bmc>" >> /tmp/discopacket
+    fi
+fi
+
 #so we have some cases where DMI table explictly records every function number, and some where only first function is enumerated
 #so for each onboard device, we record it.  If it is also the first function, we'll seek out other function numbers and append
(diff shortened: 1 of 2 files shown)


xCAT-server/lib/xcat/plugins/nodediscover.pm
```diff
@@ -154,6 +154,17 @@ sub process_request {
         }
     }
 
+    # if there is no bmc ip in ipmi table, save bmc ip into ipmi table
+      if (defined($request->{bmc})) {
+        my $ipmitab = xCAT::Table->new("ipmi", -create=>1);
```
daniceexi: Is it better to be:

~~~~
if (defined($request->{bmc}) && $request->{bmc}->[0]) {
    my $ipmitab = xCAT::Table->new("ipmi", -create=>1);
~~~~

daniceexi: Please check the validaty of $ipmitab before using it.

whowutwut:
@gurevichmark can you also take a quick review of this since you have been looking into this code?

gurevichmark:
Looks good to me. I have not really worked with the files changed for this pull request.

xuweibj:
@daniceexi , I modified as your comments, please help review again.

xCAT-genesis-scripts/bin/dodiscovery
```diff
@@ -132,6 +132,10 @@ fi
 if [ "$PLATFORM" != "unknown" ]; then
         echo "<platform>$PLATFORM</platform>" >> /tmp/discopacket
 fi
+
```
zet809: It seems we only need to deal with the static BMC IP. It can check by:
 ipmitool-xcat lan print 1 | grep -i "IP Address Source" | grep -i "Static"
IP Address Source       : Static Address

xCAT-server/lib/xcat/plugins/nodediscover.pm
```diff
@@ -154,6 +154,19 @@ sub process_request {
         }
     }
 
+    # if there is no bmc ip in ipmi table, save bmc ip into ipmi table
+      if (defined($request->{bmc}) && $request->{bmc}->[0]) {
```
zet809: The format of this line.

zet809:

PR #879, opened 2016-03-31
Modified error message of my_ip_facing

xuweibj:
If there is no appropriate MN and  SN for node, the error message will be "The IP address of node $node is in an undefined subnet".


daniceexi:
I have reviewed the code change and agreed to merge.

PR #881, opened 2016-03-31
Do not send warning message as the output is complete

chenglch:
Although handshake between parent and child process fails
due to the heavy IO loading, the message received by parent
process  is complete as the block operation is used.


daniceexi:

PR #888, opened 2016-04-04
Eliminate some Genesis boot errors

gurevichmark:
This pull request fixes several Genesis boot errors in **doxcat** and **dhclinet-script** scripts.
This pull request is related to RTC task 93486.

1. When sshd starts on ppc64 it complains about some keys not being generated. 
     - Changed to call to **ssh-keygen** with **-A** option to generate all the keys.

2. When **hwclock** command is called on ppc64, it returns an error if **/dev/rtc** file is not there. 
     - Changed code to check for this file before calling **hwclock**.

3. Several errors from **rsyslogd** command. On ppc64 a later version of **rsyslogd** is used than on x86. 
    - Added code to check for **rsyslogd** version and use appropriate command invocation
    - Fixed **rsyslog.conf** file to initially log locally into **/var/log/xcat.genesis** file. Later when xCAT MN node is known, the **rsyslog.conf** file is changed to sent all log messages to xCAT MN
    - removed _*.emerg *_ from **rsyslog.conf** file which was intended to call **wall** if emergency message is received. But it does not work, probably because **wall** does not appears to be a valid command while running in Genesis shell.

### DIFF
--- a/xCAT-genesis-builder/dhclient-script
+++ b/xCAT-genesis-builder/dhclient-script
@@ -36,15 +36,26 @@ elif [ $reason = "BOUND" ]; then
 	for gw in $new_routers; do
 		ip route add default via $gw
 	done
-        if [ ! -z "$new_log_servers" ]; then
+	# Up to this point we were logging to the local /var/log/xcat.genesis
+	# file. But now we know the xCAT MN, so replace logging to the local
+	# file with logging to the xCAT MN
+	if [ ! -z "$new_log_servers" ]; then
 		head -n -1 /etc/rsyslog.conf > /etc/rsyslog.conf.new
 		cp /etc/rsyslog.conf.new /etc/rsyslog.conf
 	fi
         for ls in $new_log_servers; do
-		echo *.* @$ls >> /etc/rsyslog.conf
+		echo "*.* @$ls" >> /etc/rsyslog.conf
 	done
 	kill `cat /var/run/syslogd.pid`
-	( sleep 3;/sbin/rsyslogd -c4 ) &
+	sleep 3
+	RSYSLOGD_VERSION=`rsyslogd -v | grep "rsyslogd" | cut -d" " -f2 | cut -d"." -f1`
+	if [ $RSYSLOGD_VERSION -ge 8 ]; then
+        	# Newer versions of rsyslogd do not support -c flag anymore
+		/sbin/rsyslogd
+	else
+		/sbin/rsyslogd -c4
+	fi
+
 
 	if [ ! -z "$new_tcode" -a -r "/usr/share/zoneinfo/posix/$new_tcode" ]; then
 		cp "/usr/share/zoneinfo/posix/$new_tcode" /etc/localtime
--- a/xCAT-genesis-builder/rsyslog.conf
+++ b/xCAT-genesis-builder/rsyslog.conf
@@ -1,5 +1,4 @@
 $ModLoad imuxsock
 $ModLoad immark
 $MarkMessagePeriod 1200
-*.emerg                                                 *
-*.*                ~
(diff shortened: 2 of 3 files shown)


daniceexi:
@gurevichmark The code change looks good for me. One question is that upon this change, the xCAT MN could get all the system log after the rsyslog is configured to forward message to MN, right?

gurevichmark:
Yes. After dhclient-script finishes, all log messages will be forwarded to xCAT MN.

daniceexi:
@whowutwut I am Ok with the change. You can merge it If you have not more comment.

whowutwut:
Thanks! looks good.

whowutwut:
@daniceexi  so the changes that were made to the files in this pull request does not automatically get build to xCAT-genesis-base package. That needs to be manually done.  On the interlock call, you said we should be putting changes into the xCAT-genesis-scripts which will overlay.   

The potential issue we see here is that if we ever update xCAT-genesis-base in the future, and the user updates *xCAT*, without any changes to the xCAT-genesis-scripts, the files will be overwritten.  

If we have the files provided by the xCAT-genesis-scripts, shouldn't we remove those from the base rpm so the next time we build the base-rpm, it will not overwrite files provided by the scripts.

daniceexi:
@whowutwut 
You are right. if a file are existed in both genesis-scripts and genesis-base, there would be an overwriting when install one of them. 

It looks like not proper to put the dhclient-script in our genesis-script. One proposal is to just enable rsyslog in dhclient-script. But we can add a rsysylog cfg in our genesis-script to config rsyslog.

gurevichmark:
Pull request #1016 used to move rsyslog.conf and dhclient-script files to genesis-scripts location.

daniceexi:
The file /usr/sbin/dhclient-script is shipped the dhclient rpm package. Is there a /usr/sbin/dhclient-script file in the genesis-base without our customization?

If we just changed the dhclient to enable the rsyslog, we may move it to doxcat and keep the dhclient-script untouched.

gurevichmark:
@daniceexi there is no dhclient-script file in the genesis-base **without** our customization. xCAT started modifying this file in 2011 and, it seems to me, is currently very different from the dhclient-script that gets shipped with the os.

A better design is probably to **not** modify dhclient-script at all. Instead, the xCAT code currently in dhclient-script could be extracted into an external script. This script can be configured to be called by dhclient-script and can be stored in genesis-scripts direcgtory. But I think this kind of work is better left for a future release.

I propose, we merge pull request #1016 to move our custom dhclient-script into genesis-scripts directory so that it can be part of xcat-core in 2.12. Then, we can open a "new feature" issue for 2.13 release to revisit this.

daniceexi:
@gurevichmark  
I am ok with this.
Could you give a diff of your changed dhclient-script with the OS shipped one?

gurevichmark:
@daniceexi I do not think we include an OS shipped version of the **dhclient-script**. xCAT has been shipping it own version of the **dhclient-script** since 2011. I just have made a few changes to the **dhclient-scrip**t that xCAT ships. Here are the changes I have made:

image

whowutwut:
According to the man page, http://linux.die.net/man/8/dhclient-script , it looks like the way we should be handling this is in the exit hooks.  We leave the `dhclient-script` alone and ship an exit hook in the genesis-scripts. But this would be in an effort to redesign, not to simply cleanup what we have today.

daniceexi:
Yes, looks we should redesign the dhclient network setting in next release. If we agree with this, I suggest we keep our current code since it works instead of taking risk to merge the pull request #1016.

gurevichmark:
The problem is that without #888 and #1016 when genesis kernel boots into shell, the rsyslog is not properly configured and none of syslog messages are logged.

Issue #715, opened 2016-02-16
[DEV] the format of "ubuntu/genimage" need to be refined


immarvin:
1. indent the code with ``perltidy -w -syn -g -opt -i=4 -nt -io -nbbc -kbl=2 -pscf=-c -aws -pt=2 -bbc ./genimage`` automatically
2. indent manually for the lines not processed in step#1
3.remove the code commented out

immarvin:
PR: https://github.com/xcat2/xcat-core/pull/890

PR #890, opened 2016-04-05
fix defect  [DEV] the format of "ubuntu/genimage" need to be refined #715 

immarvin:
fix defect https://github.com/xcat2/xcat-core/issues/715

1.    indent the code with ``perltidy -w -syn -g -opt -i=4 -nt -io -nbbc -kbl=2 -pscf=-c -aws -pt=2 -bbc ./genimage`` 
 2.   indent manually for the lines not processed in step#1 3.remove the code commented out

### DIFF
--- a/xCAT-server/share/xcat/netboot/ubuntu/genimage
+++ b/xCAT-server/share/xcat/netboot/ubuntu/genimage
@@ -23,10 +23,10 @@ use imgutils;
 Getopt::Long::Configure("bundling");
 Getopt::Long::Configure("pass_through");
 
-my $dracutmode; #Indicate whether this is a dracut style initrd
-my $dracutdir = "dracut";  # The default directory name of dracut
-my $prinic; #TODO be flexible on node primary nic
-my $othernics; #TODO be flexible on node primary nic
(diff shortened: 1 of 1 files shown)


whowutwut:
Sorry, I missed this email notification. Is the only change here perl tidy? 

The changes are too large to see the diff..so I don't know how to review...

if it's only the format change using perltidy, maybe it's okay to merge in for 2.12 GA. What do you think @daniceexi @tingtli ?

daniceexi:
It may safe. But my idea is we should follow the rule that to consider the code safe before the release, the unnecessary checkin should be postponed to next release.

whowutwut:
Ok, if we leave this pull request...  then the master will be 2.13... moving to 2.13...

daniceexi:
@whowutwut It should be 2.12.1, right? We don't plan for 2.13 now.

whowutwut:
Yes, 2.12.1, @immarvin sorry for the delay here ,but I think you might need to resolve the conflicts or re-create the pull request with the perl tidy..

immarvin:
hi @whowutwut ,since ``ubuntu/genimage`` has been changed a lot and this PR has expired, I will close this PR and create a new one later

Issue #445, opened 2015-11-20
[FVT]2.12 deleteraid/configraid enhancement: return warning message when RAID Status is in Rebuilt other than hanging there


tingtli:
This is an enhancement for deleteraid/configraid. 
In current code, when RAID status is in Rebuilt, if customer would like to delete it and create a new RAID,  it hangs for a very very long time.  Could we just return a warning message when the RAID is in Rebuilt status?  I think if the RAID is in Rebuilt status,  delete and config RAID operation should be prohibitted.

bybai:
This will be fixed in 2.12.

bybai:
while the RAID status is Rebuilding, configraid will not delete raid, it will wait until its status reach Optimized. The old timeout is 6 hours, it is too long; I will change the timeout to be 2 minutes. Outputs will be as following, code will be check in 2.12:
``` 
 ~]# xdsh p8euh 'configraid delete_raid="sda"'
p8euh: [I]: Renew available target arrays which will be deleting, "0 0:2:0:0"
p8euh: [I]: Round 1: Deletable array >>
p8euh: [I]: Round 1: ipr info before delete >> Name PCI/SCSI Location Description Status
p8euh: [I]: Round 1: ipr info before delete >> ------ ------------------------- ------------------------- -----------------
p8euh: [I]: Round 1: ipr info before delete >> 0001:08:00.0/0: PCI-E SAS RAID Adapter Operational
p8euh: [I]: Round 1: ipr info before delete >> sda 0001:08:00.0/0:2:0:0 RAID 10 Disk Array 61% Rebuilt
p8euh: [I]: Round 1: ipr info before delete >> 0001:08:00.0/0:0:0:0 RAID 10 Array Member Active
p8euh: [I]: Round 1: ipr info before delete >> 0001:08:00.0/0:0:1:0 RAID 10 Array Member Active
p8euh: [I]: Round 1: ipr info before delete >> 0003:04:00.0/1: PCI-E SAS RAID Adapter Operational
p8euh: [I]: Round 1: delete_ipr_array, "0:2:0:0"
p8euh: [W]: Round 1: Array 0:2:0:0 is un-deletable at present.
p8euh: [I]: Round 1: All remaining target arrayes are un-deletable now.
p8euh: [I]: Round 1: Wait for these un-deletable arrayes deletable with tryCnt=2, tryInt=60.
[...]

PR #891, opened 2016-04-05
fix issue445 deleteraid/configraid enhancement

bybai:
issue #445 
deleteraid/configraid enhancement: return warning message when RAID Status is in Rebuilt other than hanging there


tingtli:
I don't have test environment now. Will merge this and add a comment in defect 445 @bybai

PR #910, opened 2016-04-12
Update the ipmitool-xcat requires 

whowutwut:
For OpenPower machines 1.8.15 is required.  
For x86_64, update to 1.8.11 because we have not built 1.8.15 on all platform, only the following:

```
 find ./ -name "ipmitool-xcat*" -print  | grep 1.8.15
./rh7/ppc64le/ipmitool-xcat-1.8.15-1.ppc64le.rpm
./rh7/x86_64/ipmitool-xcat-1.8.15-1.x86_64.rpm
./rh7/ppc64/ipmitool-xcat-1.8.15-1.ppc64.rpm
./sles12/ppc64le/ipmitool-xcat-1.8.15-1.ppc64le.rpm
```

This needs to be cherry picked to master branch after review and merged

### DIFF
--- a/xCAT/xCAT.spec
+++ b/xCAT/xCAT.spec
@@ -61,13 +61,13 @@ Requires: conserver-xcat
 
 %ifarch i386 i586 i686 x86 x86_64
 Requires: syslinux xCAT-genesis-scripts-x86_64 elilo-xcat
-Requires: ipmitool-xcat >= 1.8.9
+Requires: ipmitool-xcat >= 1.8.11
 Requires: xnba-undi
 %endif
 %ifos linux
 %ifarch ppc ppc64 ppc64le
 Requires: xCAT-genesis-scripts-ppc64
-Requires: ipmitool-xcat >= 1.8.9
+Requires: ipmitool-xcat >= 1.8.15
 %endif
 %endif
 
--- a/xCATsn/xCATsn.spec
+++ b/xCATsn/xCATsn.spec
@@ -56,13 +56,13 @@ Requires: conserver-xcat
 
 %ifarch i386 i586 i686 x86 x86_64
 Requires: syslinux xCAT-genesis-scripts-x86_64 elilo-xcat
-Requires: ipmitool-xcat >= 1.8.9
+Requires: ipmitool-xcat >= 1.8.11
 Requires: xnba-undi
 %endif
 %ifos linux
 %ifarch ppc ppc64 ppc64le
 Requires: xCAT-genesis-scripts-ppc64
-Requires: ipmitool-xcat >= 1.8.9
+Requires: ipmitool-xcat >= 1.8.15
 %endif
 %endif
 


whowutwut:
@daniceexi  @zet809 , do you agree with leaving the requires at 1.8.11 for x86_64 and only bumping up 1.8.15 for ppc64le?

zet809:
@whowutwut No, in fact, the ipmitool is used by the xCAT MN to manage object servers. If the admin use x86_64 MN to manage S822LC servers, ipmitool-1.8.11 won't work. 
So, I advice to update all ipmitool-1.8.11 to be ipmitool-1.8.15 to make sure any kind of MN OS can manage the S822LC servers.

daniceexi:
I agree with @zet809, but we can merge this change now. And make the 1-8.15 later for x86_64.

whowutwut:
@zet809  good point, 

So I think we want to re-build ipmitool-xcat for the x86_64 for the OS that we want to support in the mixed cluster environment.  Can we deprecate the OS or should we build for all? 

We built ipmitool 1.8.11 for the following OS for x86_64... 
```
$ find ./ -name "ipmitool-xcat*" -print  | grep x86_64
./SL6/x86_64/ipmitool-xcat-1.8.11-3.x86_64.rpm
./fedora8/x86_64/ipmitool-xcat-1.8.11-3.x86_64.rpm
./sles10/x86_64/ipmitool-xcat-1.8.11-3.x86_64.rpm
./rh6/x86_64/ipmitool-xcat-1.8.11-3.x86_64.rpm
./SL5/x86_64/ipmitool-xcat-1.8.11-1.x86_64.rpm
./sles11/x86_64/ipmitool-xcat-1.8.11-3.x86_64.rpm
./fedora9/x86_64/ipmitool-xcat-1.8.11-1.x86_64.rpm
./rh5/x86_64/ipmitool-xcat-1.8.11-3.x86_64.rpm
./rh7/x86_64/ipmitool-xcat-1.8.15-1.x86_64.rpm
./fedora13/x86_64/ipmitool-xcat-1.8.11-1.x86_64.rpm
./fedora15/x86_64/ipmitool-xcat-1.8.11-1.x86_64.rpm
./sles12/x86_64/ipmitool-xcat-1.8.11-3.x86_64.rpm
./fedora12/x86_64/ipmitool-xcat-1.8.11-1.x86_64.rpm
./fedora14/x86_64/ipmitool-xcat-1.8.11-1.x86_64.rpm
```

zet809:
What about having ipmitool pkg for all the os link to ipmitool-xcat-1.8.15-1.x86_64.rpm? Then we don't need to put the rpm package for all of them. 
And I think it won't be a problem for the OS which using 1.8.11 before to use ipmitool-xcat-1.8.15. 
What is your idea?

daniceexi:
The build on rh6 might not work on rh7, reverse is same. So we should choose the new os to build and test whether it works for old version. For example, we could make the build on rh7.2 and test it on rh6.6. If it does work on rh6, then make a link. If does not, make a new build for rh6.

whowutwut:
@daniceexi  had a comment above,

> I agree with @zet809, but we can merge this change now. And make the 1-8.15 later for x86_64.  

Can we merge and open another issue to test the ipmitool-xcat built for x86_64 rhel7 onto the earlier OS?

daniceexi:
I will merge this pull request, the continues work will be done in #1000

PR #913, opened 2016-04-12
Add new IB pkglist files

hu-weihua:
For support rhels7.2, ubuntu14.04.3 and sles12.1 on ppc64le. add new and update pkglist file.
Also add pkglist file for rhels7.1 on ppc64 environment


neo954:
Looks good. Will merge this one

Issue #916, opened 2016-04-13
[DEV]"buildcore.sh" and "builddep.sh" report "System error: exec format error"


immarvin:
When "buildcore.sh"  and "builddep.sh" are called in the docker container, it complains:

````
[root@c910f05c01bc07 build]# docker run -it --privileged=true  -v /sys/fs/cgroup:/sys/fs/cgroup:ro  xcatbuilder:rpm UP=0 BUILDLALL=1      
exec format error
docker: Error response from daemon: Cannot start container 08760581df4b6b0657e739d988abd678faad4faf410873708da784ce6ecea761: [9] System error: exec format error.
````
This is caused by the fact that the shebang of the 2 scripts are not in the 1st line, the 1st line in the files are some description cooments. We should avoid this in the future coding

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/917

PR #917, opened 2016-04-13
the shebang of bash scripts should be the 1st line of script file

immarvin:
fix https://github.com/xcat2/xcat-core/issues/916 

The shebang of our build srcipts are not in the 1st line of the script files. The 1st line in these files are some description comments, the shebang below will be ignored as comment lines, which introduced problems

### DIFF
--- a/buildcore.sh
+++ b/buildcore.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-# The shell is commented out so that it will run in bash on linux and ksh on aix
 # Build and upload the xcat-core code, on either linux or aix.
 
 # Getting Started:
--- a/buildcore.sh
+++ b/buildcore.sh
@@ -1,6 +1,6 @@
-# The shell is commented out so that it will run in bash on linux and ksh on aix
-#  !/bin/bash
+#!/bin/bash
 
+# The shell is commented out so that it will run in bash on linux and ksh on aix
 # Build and upload the xcat-core code, on either linux or aix.
 
 # Getting Started:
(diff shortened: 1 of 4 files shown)


buildcore.sh
```diff
@@ -1,6 +1,6 @@
-# The shell is commented out so that it will run in bash on linux and ksh on aix
-#  !/bin/bash
+#!/bin/bash
```
daniceexi: @immarvin The comment said that the owner did that intentionally so that this script can be supported for both Linux and AIX.
Currently we only support Linux after 2.9 branch, so you should remove the comment also.

buildcore.sh
```diff
@@ -1,6 +1,6 @@
...
 
+# The shell is commented out so that it will run in bash on linux and ksh on aix
```
tingtli: We commented out !/bin/bash because we use ksh on aix,  user need to decide which shell this script uses on different OS. 
If we obsolete AIX, we need to remove this comment.

makerpm
```diff
@@ -1,6 +1,6 @@
-# The shell is commented out so that it will run in bash on linux and ksh on aix
-#  !/bin/bash
+#!/bin/bash
 
+# The shell is commented out so that it will run in bash on linux and ksh on aix
```
tingtli: Please remove this sentence since it will cause confusion.

buildcore.sh
immarvin: removed

immarvin:
hi @tingtli , the comments lines on AIX have been removed

PR #920, opened 2016-04-13
mini refine for genesis script doxcat to have it run into bash if no …

zet809:
…NIC with IP facing xcat MN found

### DIFF
--- a/xCAT-genesis-scripts/bin/doxcat
+++ b/xCAT-genesis-scripts/bin/doxcat
@@ -188,6 +188,7 @@ else
             done
 
 	    echo -n "Acquiring network addresses.."
+            tries=0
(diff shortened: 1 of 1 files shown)


zet809:
@gurevichmark Hi, mark, will you pls help to review this pull request and merge it? Thx!

gurevichmark:
Before we would check every 2 seconds to see if we can ping MN on a NIC.
The new code will only check once. Is there a possibility that initially, on the first check,  MN is not pingable, but some short time later it would be ?

zet809:
@gurevichmark Thx very much for your comment, I have modified to wait 20 seconds more for NICs to get IP address. Pls help to verify.

gurevichmark:
Looks good to me.

PR #924, opened 2016-04-14
sync up test case in 2.11 with 2.12

tingtli:


tingtli:
@junxiawang  please review

PR #927, opened 2016-04-14
replace the digest algorithm of xCAT ssl certificate from sha1 to sha256

immarvin:
"xCAT-server/share/xcat/ca/openssl.cnf.tmpl" is used as a template for the openssl.conf, which was the configuration file xCAT use to sign certificates for ssl communication between xcat client and xcatd, the message digest algorithm should be changed to "sha256" instead of "sha1" due to security reasons.

````
default_md      = sha256                # which md to use.
````

### DIFF
--- a/xCAT-server/share/xcat/ca/openssl.cnf.tmpl
+++ b/xCAT-server/share/xcat/ca/openssl.cnf.tmpl
@@ -67,7 +67,7 @@ cert_opt 	= ca_default		# Certificate field options
 
 default_days	= 7300			# how long to certify for
 default_crl_days= 30			# how long before next CRL
-default_md	= sha1			# which md to use.
+default_md	= sha256		# which md to use.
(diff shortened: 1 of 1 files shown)


immarvin:
the PR https://github.com/xcat2/xcat-core/pull/771 from @pdlun92 is on the same subject. If this PR is acceptable, is it ok to close it? @whowutwut @daniceexi

daniceexi:
I think this fix is better. @pdlun92 could you close the pull request #771?

whowutwut:
@immarvin @daniceexi  I closed #771 , however, it's strange because we tried the modifications in this pull request on a test machine without any success because it seemed like keywords are missing from the openssl.cnf file..    we can merge and then see if the behavior is correct.  Should merge soon so we have time to verify for 2.12

daniceexi:
OK, I merge it first and if there's problem we can reopen it.

PR #934, opened 2016-04-15
modify docker command testcase during to issue 882

junxiawang:
@tingtli  please help review


daniceexi:
The change looks OK for merge.

Issue #886, opened 2016-04-04
[DEV]nodeset <noderange> shell does not work until osimage is assigned to the node


whowutwut:
Tried this using Virtual Machines... I'm not sure if the behavior is identical for a bare metal node.

``nodeset <noderange> shell`` , does not seem to work until an osimage is associated with the node.  Is this working as designed, @zet809?   

The goal here was to try and get the VM booted into the genesis kernel.  Didn't expect to need to associate a osimage with the node... 

```
[root@c910f02c05p03 etc]# nodeset fs2vm102 osimage=rhels7.2-ppc64le-install-compute
fs2vm102: Error: Unable to identify plugin for this command, check relevant tables: noderes.netboot=(grub2[-]?.*);nodehm.mgt
[root@c910f02c05p03 etc]# nodeset fs2vm102 shell
fs2vm102: Error: Unable to identify plugin for this command, check relevant tables: noderes.netboot=(grub2[-]?.*);nodehm.mgt

[root@c910f02c05p03 etc]# chdef fs2vm102 netboot=grub2
1 object definitions have been created or modified.
[root@c910f02c05p03 etc]# nodeset fs2vm102 shell
fs2vm102: shell
Error: stop configuration, missing /tftpboot/boot/grub2/grub2..

[root@c910f02c05p03 etc]# nodeset fs2vm102 osimage=rhels7.2-ppc64le-install-compute
fs2vm102: install rhels7.2-ppc64le-compute
[root@c910f02c05p03 etc]# nodeset fs2vm102 shell
fs2vm102: shell
```

zet809:
@whowutwut Have got the point, will fix it soon.

zet809:
@whowutwut Have got the point, will fix it soon.

gurevichmark:
When "netboot" attr is set to grub2, the grub2.pm is called. It checks the "arch" attr of **osimage** specified on **nodeset** command. This "arch" attr is used to make a link grub2-_compute node_ to grub2._arch_ in /tftpboot/boot/grub2 directory. 
However if "shell" or "shutdown" specified for **nodeset** command, there is no **osimage** and therefore no "arch" specified. The error message is displayed.

Talked to Xiao Peng about possible solution. He agreed that if no **osimage** specified for the **nodeset** command, we can use the "arch" attribute from the _compute node_ definition.

gurevichmark:
Fixed by pull request #936

whowutwut:
Using the following build for Verification: 
[...]

PR #936, opened 2016-04-15
Fix nodeset command for shell and shutdown options and fix usage and man page

gurevichmark:
This pull request fixes the problem described in issue #886 and fixes some usage and man page errors for nodeset command.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/grub2.pm
+++ b/xCAT-server/lib/xcat/plugins/grub2.pm
@@ -23,7 +23,7 @@ my $globaltftpdir = xCAT::TableUtils->getTftpDir();
 
 
 my %usage = (
-    "nodeset" => "Usage: nodeset <noderange> [shell|boot|runcmd=bmcsetup|iscsiboot|osimage[=<imagename>]|offline]",
+    "nodeset" => "Usage: nodeset <noderange> [shell|boot|runcmd=bmcsetup|iscsiboot|osimage[=<imagename>]|offline|shutdown|stat]",
 );
 sub handled_commands {
     # process noderes:netboot like "grub2-<transfer protocol>"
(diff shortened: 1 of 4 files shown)


xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -620,6 +621,20 @@ sub process_request {
...
+            else
+               {
+                 # Can not determine arch from osimage definition. This is most likely
+                 # the case when nodeset is shell or shutdown. Look at node definition, to
+                 # figure out arch to use.
+                 foreach my $this_node (@{$osimagenodehash{$osimage}})
+                   {
+                     $nodetypeent = $nodetypetab->getAttribs({node => $this_node },'arch');
+                   }
+                   if($nodetypeent and $nodetypeent->{'arch'})
```
zet809: 1. It seems the vairable $nodetypeent is undeclared. 
2. The $nodetypeent is only used once here, why it is set for every node in $osimagenodehash{$osimage}?

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -620,6 +621,20 @@ sub process_request {
                {
                  $validarch = $osimgent->{'osarch'};
                }
+            else
+               {
+                 # Can not determine arch from osimage definition. This is most likely
+                 # the case when nodeset is shell or shutdown. Look at node definition, to
+                 # figure out arch to use.
+                 foreach my $this_node (@{$osimagenodehash{$osimage}})
```
zet809: As I know, the 'arch' attribute of a node is assigned through 'nodeset node osimage=<image>', if it haven't been run before, there won't be arch attribute assigned (For baremetal server, the arch attribute can be set during the hardware discovery process).
And there is another loop in line 637 to generate symlink for every normal node, you can use $nodetypetab->getNodesAttribs to get 'arch' attribute from all the normal nodes, and then if no arch for this osimage, check whether the special node have 'arch' attribute in 'nodetype' table. 
For the nodes haven't set 'arch' attribute, an error msg shall be reported.

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -620,6 +620,31 @@ sub process_request {
...
+                     {
+                       # Extracted arch from node definition
+                       $validarch = $node_entry->{'arch'};
+                     }
+                 else
+                     {
+                       # At this point we can not determine architecture either
+                       # from osimage or node definition.
+                       my $rsp;
+                       push @{$rsp->{data}}, "Not able to determine architecture of node $node_name. Verify arch attribute setting.\n";
```
whowutwut: Instead of "Verify arch attribute setting", can we do some more and help the user. 
"Verify the 'arch' attribute is set for the node (lsdef $nodename -i arch). If missing, set it using the chdef command" 

also, is there some arch that is needed, we can't set it to anything.

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -620,6 +620,31 @@ sub process_request {
                {
                  $validarch = $osimgent->{'osarch'};
                }
+            else
```
whowutwut: Little nit-picky here, but I think the code is not using 4 spaces.  I think we should re-format the code while we are in the files... there's a perl tidy command we can run and commit formatting changes in a separate commit to leave it out of the specific code changes

whowutwut:
I have reviewed the changes and will merge. thanks

PR #912, opened 2016-04-12
Update IB docs for bug fix

hu-weihua:
Please refer to http://10.2.2.30/advanced/networks/infiniband/index.html for review

I modified the most part of IB drives installation docs for below reasons:
* fix bugs
#549 ［Customer］Updatenode otherpkgs non-zero return code due to MOFED repository conflicts 
#505 ［Customer］xcat 2.11 provision diskfull RHEL7.2LE and InfiniBand MLNX OFED may need a reboot(?) 

* Add 2 new pages ``The Scenarioes Have Been Verified`` and  ``Known Issue``. Make the doc work as a guide, not a solution step by step. Just in ``The Scenarioes Have Been Verified``, it can work as step by step.

* Update the example. using the latest os and arch

Issue #549, opened 2015-12-21
[Customer] Updatenode otherpkgs non-zero return code due to MOFED repository conflicts


markprez:
```
Version 2.11 (git commit 9ea36ca6163392bf9ab684830217f017193815be, built Mon Nov 30 05:43:11 EST 2015)
```

After installing Mellanox OFED 3.1-1.0.6.1 on Red Hat 7.2, yum update fails with the following errors:

```
Error: Package: mpitests_openmpi__1_8_8-3.2.16-fe5387c.ppc64le (installed)
           Requires: liboshmem.so.3()(64bit)
           Removing: openmpi-1.8.8-1.ppc64le (installed)
               liboshmem.so.3()(64bit)
           Updated By: openmpi-1.10.0-10.el7.ppc64le (xCAT-rhels7.2rc-path0)
               Not found
Error: Package: libibmad-static-1.3.11.MLNX20150329.b2cb00f-0.1.ppc64le (installed)
           Requires: libibmad = 1.3.11.MLNX20150329.b2cb00f-0.1
           Removing: libibmad-1.3.11.MLNX20150329.b2cb00f-0.1.ppc64le (installed)
               libibmad = 1.3.11.MLNX20150329.b2cb00f-0.1
           Updated By: libibmad-1.3.12-1.el7.ppc64le (xCAT-rhels7.2rc-path0)
               libibmad = 1.3.12-1.el7
Error: Package: libibmad-devel-1.3.11.MLNX20150329.b2cb00f-0.1.ppc64le (installed)
           Requires: libibmad = 1.3.11.MLNX20150329.b2cb00f-0.1
           Removing: libibmad-1.3.11.MLNX20150329.b2cb00f-0.1.ppc64le (installed)
               libibmad = 1.3.11.MLNX20150329.b2cb00f-0.1
           Updated By: libibmad-1.3.12-1.el7.ppc64le (xCAT-rhels7.2rc-path0)
               libibmad = 1.3.12-1.el7
 You could try using --skip-broken to work around the problem
** Found 1 pre-existing rpmdb problem(s), 'yum check' output follows:
```

When we run updatenode ospkgs or otherpkgs, xcat sets up RH72 repositories on the node. Several of the rpm levels from the Red Hat distro are newer versions of the IB rpms from Mellanox.
When xCAT updatenode ospkgs or otherpkgs, "yum upgrade" is run and results in a non-zero return code:     ``Postscript: otherpkgs exited with code 1``

[...]

Issue #505, opened 2015-12-02
［Customer］xcat 2.11 provision diskfull RHEL7.2LE and InfiniBand MLNX OFED may need a reboot(?)


lxguang:
Hi,  I have been following the URL https://xcat-docs.readthedocs.org/en/latest/advanced/networks/infiniband/mlnxofed_ib_install_v2_preparation.html to provision diskfull RHEL7.2 LE with MLNX OFED by xCAT 2.11 running Ubuntu 14.04.2 LE on Power 8.
```
# cp /opt/xcat/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2    /install/postscripts/mlnxofed_ib_install
# chdef compute -p postbootscripts="mlnxofed_ib_install -p /install/rhels7.2/ppc64le/MLNX_OFED_LINUX-3.1-1.0.3.1-rhel7.2-ppc64le.iso -m --without-32bit --without-fw-update --add-kernel-support --force "
# chmod +x /install/postscripts/mlnxofed_ib_install
```
When the FireStone node is provisioned, it looks like the InfiniBand is up with the specified IP address. Ping other IB ip addresses works.  
```
[root@fs04 ~]# ibstat
CA 'mlx4_0'
        CA type: MT4099
        Number of ports: 2
        Firmware version: 2.35.5100
        Hardware version: 0
        Node GUID: 0xe41d2d0300233ea0
        System image GUID: 0xe41d2d0300233ea0
        Port 1:
                State: Down
                Physical state: Disabled
                Rate: 40
                Base lid: 0
                LMC: 0
                SM lid: 0
                Capability mask: 0x04010000
                Port GUID: 0xe61d2dfffe233ea0
                Link layer: Ethernet
        Port 2:
                State: Down
                Physical state: Disabled
                Rate: 40
                Base lid: 0
                LMC: 0
                SM lid: 0
                Capability mask: 0x04010000
                Port GUID: 0xe61d2dfffe233ea1
                Link layer: Ethernet
CA 'mlx5_0'
        CA type: MT4113
        Number of ports: 2
        Firmware version: 10.12.1100
        Hardware version: 0
        Node GUID: 0xf4521403006d5010
        System image GUID: 0xf4521403006d5010
        Port 1:
                State: Active
                Physical state: LinkUp
                Rate: 56
                Base lid: 14
                LMC: 0
                SM lid: 1
                Capability mask: 0x26596848
                Port GUID: 0xf4521403006d5010
[...]

PR #937, opened 2016-04-15
Refactor Infiniband Documentation (do not merge)

whowutwut:
This pull request is being used to revisit the process for users to configure the IB support in xCAT.
The attempt is to simplify the process and instruction set for the normal use case.


whowutwut:
I'd like to get rid of the v1 and v2 concept for 2.12.  I think if we think version2 should be used moving forward, people can still obtain version1 by looking at the source tree... we don't need to ship it. Having these two versions is confusing.  @hu-weihua  What levels of OFED do we use with v1 vs v2?

hu-weihua:
@whowutwut, I rewrote the InfiniBand docs recently. At first I just want to fix 2 bugs(#549 and #505), but I found the doc was not convenient to cover new scenarios, so I almost rewrote InfiniBand drivers installation part, and also add 2 new pages to list ``The Scenarioes Have Been Verified`` and ``Known Issues``.  The related pull request is #912.  @neo954 has reviewed first round.  I did simplify the process and instruction, but I am not sure it's good enough for you. Maybe you can review again and give me your comments.  

I agree to get rid of v1. We kept it before just worry some customer was using it. but up to now we didn't get any feedback for v1, I suppose no customer are using it now.

hu-weihua:
@whowutwut, In the new page ``The Scenarioes Have Been Verified`` I list the certain OFED version and attributes in every scenario. But this list just covers the latest scenarios I have verified(``rhels7.2+ppc64le``  ``ubuntu14.4.3+ppc64le``....),  Some scenario we had ever verified long time ago isn't contained in this list for I lost such information, we can verify again these old scenarios (for example rhels6+x86) and complete this list. But I think maybe no customer care about these old scenario. We can list all the scenarios we think we should list, and do some verify again.

whowutwut:
Closing this pull request.

PR #956, opened 2016-04-21
fix a typo in rh/genimage

immarvin:
fix https://github.com/xcat2/xcat-core/issues/953


daniceexi:
The change looks good to check in.

Issue #115, opened 2015-08-24
[Customer]Postscripts set in groups don't apply correctly to nodes inheriting postscripts from multiple groups


samveen:
Postscript definitions for relevant groups:
```
[root@inframgtlivexcatmn1 ~]# lsdef -t group -o AllRackA25-AU121,ID-LZD-ALICE-STAG,centos7,ilo,all -i postscripts
Object name: AllRackA25-AU121
    postscripts=
Object name: ID-LZD-ALICE-STAG
    postscripts=
Object name: all
    postscripts=
Object name: centos7
    postscripts=disableNetwork_Manager,alter_yum_centos,enable_epel_centos
Object name: ilo
    postscripts=
```

Relevant node fields:
```
[root@inframgtlivexcatmn1 ~]# lsdef -t node -o idlzdstagalice1 -i groups,postscripts
Object name: idlzdstagalice1
    groups=AllRackA25-AU121,ID-LZD-ALICE-STAG,centos7,ilo,all
    postscripts=syslog,remoteshell,syncfiles,disableNetwork_Manager,alter_yum_centos,enable_epel_centos
```
Changing the post scripts:
```
[root@inframgtlivexcatmn1 ~]# chdef -t group -o AllRackA25-AU121 -p postscripts=confignics
1 object definitions have been created or modified.
[root@inframgtlivexcatmn1 ~]# lsdef -t group -o AllRackA25-AU121,ID-LZD-ALICE-STAG,centos7,ilo,all -i postscripts
Object name: AllRackA25-AU121
    postscripts=confignics
Object name: ID-LZD-ALICE-STAG
    postscripts=
Object name: all
    postscripts=
Object name: centos7
    postscripts=disableNetwork_Manager,alter_yum_centos,enable_epel_centos
Object name: ilo
    postscripts=
```
Relevant postscripts after changes to group:
```
[root@inframgtlivexcatmn1 ~]# lsdef -t node -o idlzdstagalice1 -i groups,postscripts
Object name: idlzdstagalice1
    groups=AllRackA25-AU121,ID-LZD-ALICE-STAG,centos7,ilo,all
    postscripts=syslog,remoteshell,syncfiles,confignics
```
Change back to old config:
```
[root@inframgtlivexcatmn1 ~]# chdef -t group -o AllRackA25-AU121 -m postscripts=confignics
1 object definitions have been created or modified.
[...]

PR #961, opened 2016-04-22
Add configuration in site table to inherit attributes

chenglch:
'heirarchicalattrs' is added in site table to apply attributes
from multiple groups in order.

close-issue #115


daniceexi:

Issue #948, opened 2016-04-19
[DEV] RHEL 7.2 can't install on ppc64le Non-Virtualized


penguhyang:
how to reproduce:
```
rsetboot c910f05c33 net -p
nodeset c910f05c33 osimage=rhels7.2-ppc64le-install-compute
rpower c910f05c33 boot
```
The console output:
```
...
Installation

 1) [x] Language settings                 2) [x] Timezone settings
        (English (United States))                (US/Eastern timezone)
 3) [x] Installation source               4) [x] Software selection
        (http://10.3.5.10:80/install/rh          (Minimal Install)
        els7.2/ppc64le)                   6) [x] Kdump
 5) [!] Installation Destination                 (Kdump is enabled)
        (Error checking storage configu   8) [ ] User creation
        ration)                                  (No user will be created)
 7) [x] Network configuration
        (Wired (enP3s1f3) connected)
  Please make your choice from above ['q' to quit | 'b' to begin installation |
  'r' to refresh]:
```
The environment:
```
root@c910f03c05k10:~# lsxcatd -v
Version 2.12 (git commit 15abaac70885e8d08e8c1bf225b751b921390b18, built Tue Mar 22 23:23:42 EDT 2016)
root@c910f03c05k10:~# lsdef -t node c910f05c33
Object name: c910f05c33
    arch=ppc64le
    bmc=50.5.33.10
    bmcpassword=passw0rd
    bmcusername=ADMIN
    chain=bmcsetup,shell
    cons=ipmi
    currchain=boot
    currstate=install rhels7.2-ppc64le-compute
    groups=pkvm,all
    hidden=0
    initrd=xcat/osimage/rhels7.2-ppc64le-install-compute/initrd.img
    installnic=mac
    ip=10.5.33.254
    kcmdline=quiet inst.repo=http://!myipfn!:80/install/rhels7.2/ppc64le inst.ks=http://!myipfn!:80/install/autoinst/c910f05c33 BOOTIF=98:be:94:59:ea:27  ip=dhcp  inst.sshd inst.loglevel=debug inst.syslog=!myipfn! inst.cmdline  console=tty0 console=hvc0,115200
    kernel=xcat/osimage/rhels7.2-ppc64le-install-compute/vmlinuz
    mac=98:be:94:59:ea:27
    mgt=ipmi
    netboot=petitboot
    nodetype=mp
    os=rhels7.2
    postbootscripts=otherpkgs
[...]

PR #965, opened 2016-04-25
fix issue 948 [DEV] RHEL 7.2 can't install on ppc64le Non-Virtualized

zet809:
Fix issue #948


daniceexi:
I reviewed the change and agree to merge.

Issue #938, opened 2016-04-18
[Solution] Man page of bmcdiscover missed the -u -p in SYNOPSIS for scan operation


daniceexi:
The command **bmcdiscover** does support to specify the -u/-p to indicate which user and password should be used to verify the discovered bmc device, but in the man page it does not indicate that.

~~~~
bmcdiscover [-s scan_method] --range ip_ranges [-z] [-w] [-t]
~~~~

whowutwut:
@daniceexi  I took this since I touched the man pages recently.. In the following 2.12 build:
```
[root@fs2vm100 ~]# lsxcatd -a
Version 2.12 (git commit 641af3c6a0fcec382c4482d8c0abc6717c06af3d, built Thu Apr 14 09:30:34 EDT 2016)
This is a Management Node
dbengine=SQLite
```
We have the following man page: 
```
BMCDISCOVER.1(1)                              User Contributed Perl Documentation                             BMCDISCOVER.1(1)

NAME
       bmcdiscover - Discover Baseboard Management Controllers (BMCs) using a scan method

SYNOPSIS
       bmcdiscover [-?|-h|--help]

       bmcdiscover [-v|--version]

       bmcdiscover [-s scan_method] --range ip_ranges [-z] [-w] [-t]

       bmcdiscover {-i|--bmcip} bmc_ip {-u|--bmcuser} bmc_username {-p|--bmcpasswd} bmc_password --check

       bmcdiscover {-i|--bmcip} bmc_ip {-u|--bmcuser} bmc_username {-p|--bmcpasswd} bmc_password --ipsource
```

What version of the 2.12 build are you looking at?

daniceexi:
@whowutwut I meant the -u/-p should be added for the following specific line. Otherwise the user would think the -u/-p only works  with **--check or --ipsource**

~~~~
bmcdiscover [-s scan_method] --range ip_ranges [-z] [-w] [-t]
~~~~

whowutwut:
Thanks, @daniceexi  I'll work on making the change.

whowutwut:
I think the format of the option should be the following: 

```
bmcdiscover [-s scan_method] --range ip_range [{-u|--bmcuser} bmc_user {-p|--bmcpasswd} bmc_password] [-z] [-w] [-t]
```

If one of -u or -p is specified, then both must be specified.

daniceexi:
I have verified this bug.

PR #967, opened 2016-04-25
Add the -u and -p option back into the man page for bmcdiscover

whowutwut:
Fixes Issue #938

### DIFF
--- a/docs/source/guides/admin-guides/references/man1/bmcdiscover.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/bmcdiscover.1.rst
@@ -23,11 +23,11 @@ SYNOPSIS
 
 \ **bmcdiscover**\  [\ **-v | -**\ **-version**\ ]
 
-\ **bmcdiscover**\  [\ **-s**\  \ *scan_method*\ ] \ **-**\ **-range**\  \ *ip_ranges*\  [{\ **-u | -**\ **-bmcuser**\ } \ *bmc_username*\  {\ **-p | -**\ **-bmcpasswd**\ }] [\ **-z**\ ] [\ **-w**\ ] [\ **-t**\ ]
+\ **bmcdiscover**\  [\ **-s**\  \ *scan_method*\ ] [\ **-u**\  \ *bmc_user*\ ] [\ **-p**\  \ *bmc_passwd*\ ] [\ **-z**\ ] [\ **-w**\ ] [\ **-t**\ ] \ **-**\ **-range**\  \ *ip_ranges*\ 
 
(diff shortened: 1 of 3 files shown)


daniceexi:
* I think we should skip to use {} in the man page for xCAT command . We was using {} to express the parameter is mandatory. But in the Linux, without any bracket  means mandatory. 

* Put the mandatory parameter at end of the **SYNOPSIS** (refer to the **man man**). Even in xcat cmd, we always put the mandatory **noderange** as the first parameter. 

So I recommend the synopsis changes from 

~~~~
B<bmcdiscover> [B<-s> I<scan_method>] B<--range> I<ip_ranges> [{B<-u>|B<--bmcuser>} I<bmc_username> {B<-p>|B<--bmcpasswd>}] [B<-z>] [B<-w>] [B<-t>]
~~~~

to (P.S: I don't like to have both short and long format for one option. It's not necessary.)
~~~~
B<bmcdiscover> [B<-s> I<scan_method>] [B<-u> I<bmc_username> B<-p> I<bmc_password>] [B<-z>] [B<-w>] [B<-t>] B<--range> I<ip_ranges>
~~~~

gurevichmark:
Yes, no brackets means mandatory, [ ] means optional, and { } means "one of".
So technically,  {-i | --bmcip} means either -i or --bmcip needs to be specified. But these 2 options are really the same. So perhaps SYNOPSIS should only list the short version of the option and then in the DESCRIPTION, both short and long can be shown. This way we can get rid of { } for this command.

Also, on the third line of SYNOPSIS, is the "bmc_password" missing following **{-p|--bmcpasswd}**

Finally the SYNOPSIS section does not list -c option only --check. The DESCRIPTION section lists both.

whowutwut:
Propose the following usage: 

```
# bmcdiscover -h

bmcdiscover - Discover BMC (Baseboard Management Controller) using the specified scan method

Usage:
	bmcdiscover [-?|-h|--help]
	bmcdiscover [-v|--version]
	bmcdiscover [-s scan_method] [-u bmc_user] [-p bmc_passwd] [-z] [-w] [-t] --range ip_range

	Check BMC administrator User/Password:

		bmcdiscover -u bmc_user -p bmc_password -i bmc_ip --check

	Check the BMC IP configuration:

		bmcdiscover [-u bmc_user] [-p bmc_passwd] -i bmc_ip --ipsource
```

I'm also going to make changes to have -u and -p optional except for --check option where we are checking the user and password.  In the other case, we should pick up the defaults from the passwd table and overwrite it if the user specifies -u or -p.

whowutwut:
I updated the pull request with the changes

daniceexi:

Issue #864, opened 2016-03-28
[DEV] auditlog table will be updated for each command as auditnosyslog is 0 by default


chenglch:
Currently, auditlog table keeps the record information for every command. As db update operation is slow, is it necessary to open auditlog feature by default?

daniceexi:
My idea is to make the audilog to be disabled by default. If user wants use it, just enable it.

whowutwut:
I also noticed that each entry in the audit log is a few lines for each command.  Is there a way to get it down to one line for each command? What was the reason behind that format? is it a standard?  It's hard to parse or grep the current format entry...

daniceexi:
We can set the **site.auditskipcmds=ALL** by default to avoid the db writing for each command.

Refer to: https://sourceforge.net/p/xcat/wiki/Syslog_and_auditlog/#auditlog

PR #974, opened 2016-04-26
Stop audit command log by default

chenglch:
Audit every command in auditlog table may affect performance, disable
it by default.

close-issue #864


daniceexi:

Issue #977, opened 2016-04-27
[FVT]support running script in xcat test case


tingtli:
Need to update /opt/xcat/bin/xcattest to run some specified scripts in xcat case.
E.g
cmd: /tmp/script.sh

hu-weihua:
@tingtli , I have fixed this defect my pull request #981. Could you please review?

tingtli:
We've used this feature in our autotest. so i will close it.

PR #981, opened 2016-04-27
Fix bug #977 support running script in xcat test case and bug #982 update xcattest log recording time

hu-weihua:
This pull request covers 2 things:
1, Fix bug #977 support running script in xcat test case 
2. Fix bug #982  update xcattest log recording time. We used system time to calculate ``xcattest`` itself running time, but used Greenwich time to calculate each test case running time, so it will cause problem in ``xcattest`` result report file.


daniceexi:
I am ok with the change.

Issue #954, opened 2016-04-20
[FVT]Failed to find expected logs while site.xcatdebugmode=1 for ubuntu on ppc64le


caomengmeng:
Description:
According to the design doc:　https://github.com/xcat2/xcat-core/wiki/support%20new%20values%20for%20site.xcatdebugmode
while the site.xcatdebugmod=1/2 for ubuntu installation, the Y3: installation logs can be forwarded to the MN in /var/log/xcat/computes.log file without export to subsequent commands.
But there isn't this kind of file in MN:/var/log/xcat, only the below two logs:
```
root@c910f03fsp03v10:/var/log/xcat# ls
commands.log  xcat.log
```
And failed to find most of the excepted logs in the two files.
ENV.
```
root@c910f03fsp03v10:/opt/xcat/share/xcat/tools/autotest# lsxcatd -a
Version 2.12 (git commit 641af3c6a0fcec382c4482d8c0abc6717c06af3d, built Thu Apr 14 09:32:00 EDT 2016)
This is a Management Node
dbengine=SQLite
```

penguhyang:
This bug is not only related to permission but also the chroot environment.
The pull request #984 make the logger can save info to MN after provision.
But the log can't be forwarded to MN when provisioning node.
I think it's related to the chroot. The reason for this is that the chroot’ed process can not open /dev/log as this is not within the chrooted filesystem.

daniceexi:
So looks like @penguhyang thinks the issue has not been fixed and I don't see the possibility to fix in 2.12. Since it's a log related issue, move it to 2.12.1

whowutwut:
I'm a little confused. Why does chroot come into play here? Is this related to diskless?

daniceexi:
@whowutwut I think @penguhyang meant the phase that run the post script in the Ubuntu diskful installer. In that phase the installer needs chroot to the new installed file system to run script.

PR #984, opened 2016-04-27
Failed to find expected logs while site.xcatdebugmode=1 for ubuntu on ppc64le

penguhyang:
for ubuntu14, change the permission to make the logs can be saved.
for ubuntu15, fix the procedure to make the logs can be saved.

### DIFF
--- a/xCAT/postscripts/syslog
+++ b/xCAT/postscripts/syslog
@@ -85,8 +85,6 @@ config_Rsyslog_C3()
 
   if [ ! -f $2.XCATORIG ]; then
     cp -f $2 $2.XCATORIG
-  else
-    cat $2 > $2.XCATORIG
   fi
 
   grep "xCAT settings" $conf_file 2>&1 1> /dev/null
@@ -113,6 +111,8 @@ config_Rsyslog_C3()
       touch  /var/log/messages  
       if ( pmatch $OSVER "ubuntu*" ) || ( is_lsb_ubuntu ) || ( pmatch $OSVER "debian*" ); then
         chown syslog:adm /var/log/messages
+        sed -i 's/\$PrivDropToGroup syslog/\$PrivDropToGroup adm/' $2
+        chown syslog:adm /var/log/xcat
       fi
   else
       sed -i 's/^\(\*\..*\)/\#\1/; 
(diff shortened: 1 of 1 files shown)


penguhyang:
fix issue: #954

xCAT/postscripts/syslog
```diff
@@ -152,6 +154,12 @@ config_rsyslog_V8()
     fi
   fi
 
+  # backup rsyslog.conf
+  if [ ! -f $conf_file.XCATORIG ]; then
+    cp -f $conf_file $conf_file.XCATORIG
+  else
+    cat $conf_file > $conf_file.XCATORIG
```
immarvin: the $conf_file.XCATORIG is used to backup the original syslog configuration on rsyslog installation, do not overwrite it when it has already existed

xCAT/postscripts/syslog
```diff
@@ -169,11 +177,13 @@ config_rsyslog_V8()
...
+                     s/#input(type="imudp" port="514")/input(type="imudp" port="514")/; 
+                     s/#module(load="imtcp")/module(load="imtcp")/; 
+                     s/#input(type="imtcp" port="514")/input(type="imtcp" port="514")/; 
+                     s/\$PrivDropToGroup syslog/\$PrivDropToGroup adm/' $conf_file
       chown syslog:adm /var/log/messages
-    fi
-    # backup rsyslog.conf
-    if [ ! -f $conf_file.XCATORIG ]; then
-      cp -f $conf_file $conf_file.XCATORIG
+      chown syslog:adm /var/log/xcat
```
immarvin: whether "-R" option is needed?
````
       -R, --recursive
              operate on files and directories recursively
````
if the log files are created after the postscript "syslog" is run, will the files inherit the owner and group of the directory?if not, maybe we need to create them on xCAT installation

File: xCAT/postscripts/syslog (hunk 1 above)
Comment (penguhyang) [inline xCAT/postscripts/syslog]: fixed.

File: xCAT/postscripts/syslog (hunk 2 above)
Comment (neo954) [inline xCAT/postscripts/syslog]: ``-R`` is not needed.

Line 174 and line 180 is no use actually. ``rsyslog`` will create the log file as needed. No need to ``touch`` it first at all.

File: xCAT/postscripts/syslog [hunk 3]
```diff
@@ -213,6 +219,11 @@ config_rsyslog_V8()
     # add xCAT entries
     echo "# xCAT settings" >> $remoteconf
     echo "*.* @$master" >> $remoteconf
+
+    if ( pmatch $OSVER "ubuntu*" ) || ( is_lsb_ubuntu ) || ( pmatch $OSVER "debian*" ); then
+        echo "# xCAT settings" >> $conf_file
+        echo "*.* @$master" >> $conf_file
+    fi
```
neo954: ``$remoteconf`` is already used, why do the same thing again in the main configuration file ``$conf_file``.

penguhyang: fixed

PR #988, opened 2016-04-27
Add 7 bundle files according to existing automation test environment

caomengmeng:
@tingtli 
Please help to review these bundle files.
Thanks.


tingtli:
This is the new bundle files that we will put all provision cases with command cases

Issue #437, opened 2015-11-19
[FVT]: 2.12: could not go to next state in currchain when currstate is shell


tingtli:
This bug is found when I use test provision after I create the RAID.

Here is the command I used to define node:

[root@c910f03c05k23 ~]# nodeset p8euh runcmd="configraid delete_raid=all stripe_size=256 create_raid=rl#0|pci_slot_name#0001:08:00.0|disk_num#2",shell,osimage=ubuntu14.04.2-ppc64el-install-compute
p8euh: runcmd=configraid delete_raid=all stripe_size=256 create_raid=rl#0|pci_slot_name#0001:08:00.0|disk_num#2
[root@c910f03c05k23 ~]# rsetboot p8euh net
p8euh: Network
[root@c910f03c05k23 ~]# rpower p8euh reset
p8euh: reset

Monitor the console, after raid is created, system goes into shell, but could not go ahead to start provsion.

[root@c910f03c05k23 ~]# lsdef -l p8euh
Object name: p8euh
    arch=ppc64el
    bmc=50.3.15.1
    bmcpassword=passw0rd
    cons=ipmi
    conserver=10.3.5.23
    currchain=osimage=ubuntu14.04.2-ppc64el-install-compute
    currstate=shell
    groups=all,kvm
    initrd=xcat/genesis.fs.ppc64.gz
    installnic=mac
    ip=10.4.41.200
    kcmdline=quiet console=tty0 console=hvc0,115200 xcatd=10.3.5.23:3001 destiny
    kernel=xcat/genesis.kernel.ppc64
    mac=6C:AE:8B:02:12:50
    mgt=ipmi
    netboot=petitboot
    nfsserver=10.3.5.23
    os=ubuntu14.04.2
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    primarynic=mac
    profile=compute
    provmethod=ubuntu14.04.2-ppc64el-install-compute
    serialport=0
    serialspeed=115200
    status=shell
    statustime=11-19-2015 02:52:39
    tftpserver=10.3.5.23

whowutwut:
Since this is waiting in shell, does ssh into shell and exit cause the process to continue?

tingtli:
Verified can closed. 
[...]

PR #994, opened 2016-04-28
Fix issue 437, go to next state when exit shell

xuweibj:
#437 

When exit shell, run /bin/nextdestiny and then get the destiny.


daniceexi:

Issue #972, opened 2016-04-26
[FVT]restart xcatd get different logs 


tingtli:
When i verify bug #537 I get different logs when I restart xcatd.
Here is the 3 logs 

Apr 21 14:18:36 c910f03c17k33 xcat[450]:  INFO xcatd: 'Command log writer' process 450 is terminated by TERM or INT signal
Apr 21 14:18:36 c910f03c17k33 xcat[448]:  xcatd udp service 448 quiescing
Apr 21 14:18:39 c910f03c17k33 xcat[5518]:  INFO xcatd: command log process 5518 start
Apr 21 14:20:02 c910f03c17k33 xcat[5515]:  xcatd udp service 5515 quiescing

Apr 21 14:20:02 c910f03c17k33 xcat[5518]:  INFO xcatd: 'Command log writer' process 5518 is terminated by TERM or INT signal
Apr 21 14:20:02 c910f03c17k33 xcat[5516]:  xcatd install monitor 5516 quiescing
Apr 21 14:20:04 c910f03c17k33 xcat[5614]:  INFO xcatd: command log process 5614 start
Apr 21 14:22:20 c910f03c17k33 xcat[5613]:  xcatd install monitor 5613 quiescing

Apr 21 14:22:20 c910f03c17k33 xcat[5614]:  INFO xcatd: 'Command log writer' process 5614 is terminated by TERM or INT signal
Apr 21 14:22:20 c910f03c17k33 xcat[5611]:  xcatd udp service 5611 quiescing
Apr 21 14:22:23 c910f03c17k33 xcat[5710]:  INFO xcatd: command log process 5710 start

chenglch:
@tingtli  For current code logic, I think the log sequential is not a problem as the time to execute the signal handler depends on the result of kernel scheduler for udp and monitor processes. We can find from the source code below:
https://github.com/xcat2/xcat-core/blob/master/xCAT-server/sbin/xcatd#L662
https://github.com/xcat2/xcat-core/blob/master/xCAT-server/sbin/xcatd#L306

@daniceexi  I don't know too much background of the usage of USR2, why do we define USR2 as a stop handler like TERM or INT?  
https://github.com/xcat2/xcat-core/blob/master/xCAT-server/sbin/xcatd#L965   Here may be a problem, we intent to send two signals to terminate the process.

chenglch:
@tingtli  as the sequence of the log message is not controlled by xcat, just verify the count of log message when starting or stopping xcatd is OK

[...]

PR #996, opened 2016-04-28
Delete signal to stop monitor process

chenglch:
Currently INT signal and USR2 signal are sent to stop the monitor
process at the same time, just keep USR2 as the stop signal in this
patch.

Close-issue: #972


daniceexi:

Issue #995, opened 2016-04-28
［FVT]xcattest does not support vmstorage=dir:///var/lib/libvirt/images/  in default.conf


tingtli:
When we define vmstorage=dir:///var/lib/libvirt/images/  in default.conf, we found xcattest will read this line but define vmstorage=dir: in node defintion. This is not correct. Need to correct.

hu-weihua:
@tingtli,  I have fixed this issue by pull request #997 . Could you please review #997 ?

tingtli:
This is fixed and we use it in our daily autotest regression.

PR #997, opened 2016-04-28
Fix bug 995 [FVT]xcattest does not support vmstorage=dir:///var/lib/libvirt/images/ in default.conf

hu-weihua:
Fix bug #995 xcattest does not support vmstorage=dir:///var/lib/libvirt/images/ in default.conf

### DIFF
--- a/xCAT-test/xcattest
+++ b/xCAT-test/xcattest
@@ -197,7 +197,7 @@ sub getConfig
             }
         }elsif ($type eq "Object") {
             ##OBJECT BLOCK##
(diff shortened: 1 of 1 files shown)


hu-weihua:
@tingtli ,  after fix, the default.conf will be load as below 

    BJECT:c910f02c04p05,TYPE:node
        vmstorage = dir:///var/lib/libvirt/images/;
        parent = c910f02fsp04;
        pprofile = c910f02c04p05;

daniceexi:

PR #999, opened 2016-04-28
modify xCATreg for mini.iso

junxiawang:
@tingtli


junxiawang:
@tingtli  please help review

Issue #987, opened 2016-04-27
[FVT]2.12  output format for switchdiscover should be modified 


junxiawang:
env:[root@c910f04x30 ~]# cat /etc/*-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.1 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.1"
build:
lsdef - Version 2.12 (git commit 750e82f4f2a77a9e2c870b1776169b2a33110300, built Mon Apr 25 09:31:08 EDT 2016)
How to reproduce:
```
[root@c910f04x30 ~]# switchdiscover --range 10.4-5.20-25.1 -s snmp
Discovering switches using snmpwalk.....
ip          	name           	vendor                                  	mac         
------------	------------   	------------                            	------------
10.4.25.1   	Juniper_01     	test juniper switch                     	2c:6b:f5:37:45:7f
10.5.23.1   	switch-10-5-23-1	IBM Networking Operating System RackSwitch G8052	a8:97:dc:5:2e:0
[root@c910f04x30 ~]# switchdiscover --range 10.4-5.20-25.1        
Discovering switches using nmap for 10.4-5.20-25.1. It may take long time...
ip          	name           	vendor                                  	mac         
------------	------------   	------------                            	------------
10.4.25.1   	Juniper_01     	Juniper networks                        	2C:6B:F5:37:45:7F
10.5.23.1   	switch-10-5-23-1	IBM                                     	A8:97:DC:05:2E:00
10.4.21.1   	switch-10-4-21-1	IBM                                     	E4:1F:13:ED:8B:AA
10.4.20.1   	switch-10-4-20-1	IBM                                     	40:F2:E9:0A:8C:64
10.5.24.1   	switch-10-5-24-1	Juniper Networks                        	5C:5E:AB:64:CA:7F
```
--------------------> the output for switchdiscover should be modifed

Issue #973, opened 2016-04-26
[FVT]2.12 switchdiscover could not find switch Ubuntu using snmp


junxiawang:
env:
root@c910f04x30v14:~# cat /etc/*-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.4 LTS"
NAME="Ubuntu"
VERSION="14.04.4 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.04.4 LTS"
VERSION_ID="14.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
root@c910f04x30v14:~# 
buid：lsdef - Version 2.12 (git commit 750e82f4f2a77a9e2c870b1776169b2a33110300, built Mon Apr 25 09:32:19 EDT 2016)
How to reproduce:
```
root@c910f04x27v08:~#  switchdiscover --range 10.4.25.1 -s snmp
Discovering switches using snmpwalk.....
 No switch found 
root@c910f04x27v08:~# echo $?
0
root@c910f04x27v08:~# 
------------------------> in fact there are a switch named 10.4.25.1 
```

cxhong:
snmpwalk common on the ubuntu didn't return object name, it only return OID, so I can't use following command:
`````
snmpwalk -Os -v1 -c public $ip sysDescr.0
snmpwalk -Os -v1 -c public $ip ipNetToMediaPhysAddress | grep $ip
snmpwalk -Os -v1 -c public $ip sysName
`````
for sysDescr.0,  the command has to replace like:
````
snmpwalk -Os -v1 -c public 10.4.25.1 1.3.6.1.2.1.1.1
`````

junxiawang:
@cxhong  your means we do not support switchdiscover on ubuntu using snmp?

cxhong:
@junxiawang , I will modified the code to use OID instead of object name

Issue #971, opened 2016-04-26
[FVT]2.12:could not  discover switch using switchdiscover -s snmp on sles 11.4 environment


junxiawang:
env:
c910f02c01p14:~ # cat /etc/*-release
LSB_VERSION="core-2.0-noarch:core-3.2-noarch:core-4.0-noarch:core-2.0-ppc64:core-3.2-ppc64:core-4.0-ppc64"
NAME="SLES"
VERSION="11.4"
VERSION_ID="11.4"
PRETTY_NAME="SUSE Linux Enterprise Server 11 SP4"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:11:4"
SUSE Linux Enterprise Server 11 (ppc64)
VERSION = 11
PATCHLEVEL = 4
c910f02c01p14:~ # 
build:
lsdef - Version 2.12 (git commit b908019a59bcb53fbb758ae99e1f0cb4a13d648d, built Sun Apr 24 09:30:54 EDT 2016)
How to reproduce:
```
c910f02c01p14:~ # switchdiscover --range 10.4.25.0/24,192.168.0.0/24 -s snmp
Discovering switches using snmpwalk.....
Could not process this command: /usr/bin/nmap -P0 -v -sU -p 161 -oA snmp_scan 10.4.25.0/24 192.168.0.0/24 | grep 'open port 161' 
 No switch found 
c910f02c01p14:~ # echo $?
1
In fact we could find switch 10.4.25.1 on rhels7 env
```

cxhong:
works fine on sles12:
`````
# switchdiscover --range 10.4.25.0/24,192.168.0.0/24 -s snmp
Discovering switches using snmpwalk.....
ip              name            vendor                                          mac
------------    ------------    ------------                                    ------------
10.4.25.1       Juniper_01      test juniper switch                             2c:6b:f5:37:45:7f
``````

on sles12.1:

````
 # switchdiscover --range 10.4.25.0/24,192.168.0.0/24 -s snmp
Discovering switches using snmpwalk.....
ip              name            vendor                                          mac
------------    ------------    ------------                                    ------------
10.4.25.1       Juniper_01      test juniper switch                             2c:6b:f5:37:45:7f

````

I don't have sles11.4 system available right now, 
@junxiawang can u able to rerun this command again on your sles11.4 node?

junxiawang:
@cxhong 
[...]

PR #1003, opened 2016-04-29
Support switchdiscover snmp_scan command for sles11.4 and ubuntu

cxhong:
Fix issue #971 :
on sles11.4, the command /usr/bin/nmap -P0 -v -sU -p 161 -oA snmp_scan 10.4.25.1 had different output than other OS

Fix issue #973   snmpwalk command on the ubuntu only supports OID, not object name

Fix issue #987  increase size of format string

### DIFF
--- a/xCAT-server/lib/xcat/plugins/switchdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/switchdiscover.pm
@@ -836,7 +836,7 @@ sub snmp_scan {
     # command.
     # for version 4.75, the line as :"Host 10.4.25.1 appears to be up ... good."
     # other higher version has line like this: "Discovered open port 161/udp on 10.4.25.1"
-    if ($nmap_version <= 4.75) {
+    if (xCAT::Utils->version_cmp($nmap_version,"4.75") <= 0) {
         $ccmd = "/usr/bin/nmap -P0 -v -sU -p 161 -oA snmp_scan @$ranges | grep up | grep good ";    
     } else {
         $ccmd = "/usr/bin/nmap -P0 -v -sU -p 161 -oA snmp_scan @$ranges | grep 'open port 161' ";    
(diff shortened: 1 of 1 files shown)


cxhong:
@daniceexi , can u review this?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -819,13 +822,29 @@ sub snmp_scan {
     ##################################################
     my $ranges = get_ip_ranges($request);
 
+    #sles11 SP4 has different output than other OS from nmap command
```
daniceexi: @cxhong Generally the different output format caused by the difference of nmap versions. So could you try to use the version of nmap to distinguish which kind of output format should be checked.

* rh7.2
~~~~
# nmap -V

Nmap version 6.40 ( http://nmap.org )
~~~~

* sles11.4
~~~~
# nmap -V

Nmap version 4.75 ( http://nmap.org )

~~~~

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -822,20 +822,21 @@ sub snmp_scan {
...
-    # other OS has line like this: "Discovered open port 161/udp on 10.4.25.1"
-    # only open port will be scan
     #use nmap to find if snmp port is enabled  
-    if ($sles11_sp4 == 1) {
+    # only open port will be scan
+    # currently, we know nmap version 4.75 has different output for snmp_scan
+    # command.
+    # for version 4.75, the line as :"Host 10.4.25.1 appears to be up ... good."
+    # other higher version has line like this: "Discovered open port 161/udp on 10.4.25.1"
+    if ($nmap_version <= 4.75) {
```
daniceexi: @cxhong Could you use the Utils.pm->version_cmp() to make the version comparison?

daniceexi:
There's another place that needs to use version_cpm to replace the string compare directly. 
I will merge the code first and make the minor change directly in the github.

daniceexi:
A minor change for version_cmp has been done in commit: a05c28c05ec91e47f68bfecad468a209fd962f21

Issue #998, opened 2016-04-28
[FVT]Ubuntu14.04.4 is dash now, so automation test cases should be updated.


caomengmeng:
As Ubuntu 14.04.r is using dash by default now,  automation test cases default using bash, which caused so many cases failed in ubuntu 14.04.4, so we need to make changes.

hu-weihua:
@caomengmeng , I have fixed this bug by pull request #1007 .

caomengmeng:
As we discussed, we still use the bash to avoid unexpected issue maybe caused by existed cases.

PR #1007, opened 2016-04-29
Fix 998[FVT] Ubuntu14.04.4 is dash now, so automation test cases should be updated

hu-weihua:
Fix bug #998, Ubuntu14.04.4 is dash now, so automation test cases should be updated


hu-weihua:
@tingtli,  Could you please review and merge?

daniceexi:

Issue #989, opened 2016-04-27
Put the log of DHCP request and reply back to /var/log/messages


neo954:
The log of DHCP is belong to ``daemon.debug``. Need to log these logs back to ``/var/log/messages``.

neo954:
Verified and closed

PR #1010, opened 2016-04-29
Fix github issue #989, put the log of DHCP back to /var/log/messages

neo954:


daniceexi:

PR #1012, opened 2016-04-29
Update provision cases to remove "chdef -t node -o $$CN vmstorage=dir:///var/lib/libvirt/images/"

junxiawang:
@tingtli please help review


tingtli:
@junxiawang This is the provision case, not the vm created cases.
You remove the chdef vm vmstorage attribute in the case to make sure that we could use other vmstorage values.

Issue #492, opened 2015-11-30
[Customer] xcatd service can not be started automatically after reboot for RHEL7.2


fenggaobj:
We find that there is no link to "/etc/init.d/xcatd" under "/etc/rcX.d/" directory for RHEL7.2 after installing xCAT-server(xCAT-server-2.11-snap201511100307.noarch.rpm). It leads to that xcatd service can not be started automatically after reboot. 
Reinstall xcat-server package manually, the result is the same.

But for ubuntu14.04.3 we find that this function works well.  when we install it, the log shows that

root@ubuntu:/mnt/packages/repos/kit-pcm-4.2.1-ubuntu-14-ppc64el# dpkg -i xcat-server_2.11-snap201510260426_all.deb
Selecting previously unselected package xcat-server.
(Reading database ... 127545 files and directories currently installed.)
Preparing to unpack xcat-server_2.11-snap201510260426_all.deb ...
Unpacking xcat-server (2.11-snap201510260426) ...
Setting up xcat-server (2.11-snap201510260426) ...
update-rc.d: warning: default start runlevel arguments (2 3 4 5) do not match xcatd Default-Start values (3 4 5)
update-rc.d: warning: default stop runlevel arguments (0 1 6) do not match xcatd Default-Stop values (none)
Adding system startup for /etc/init.d/xcatd ...
   /etc/rc0.d/K20xcatd -> ../init.d/xcatd
   /etc/rc1.d/K20xcatd -> ../init.d/xcatd
   /etc/rc6.d/K20xcatd -> ../init.d/xcatd
   /etc/rc2.d/S20xcatd -> ../init.d/xcatd
   /etc/rc3.d/S20xcatd -> ../init.d/xcatd
   /etc/rc4.d/S20xcatd -> ../init.d/xcatd
   /etc/rc5.d/S20xcatd -> ../init.d/xcatd
Processing triggers for ureadahead (0.100.0-16) ...

bybai:
Hi @fenggaobj , In our environment, I can find the link to ../init.d/xcatd. Could you check your steps or how reproduce it ?
```
[root@c910f03c05k23 etc]# ls -l rc0.d/K60xcatd
lrwxrwxrwx 1 root root 15 Sep 24 05:11 rc0.d/K60xcatd -> ../init.d/xcatd
[root@c910f03c05k23 etc]# lsxcatd -v
[...]

PR #1014, opened 2016-04-29
fix issue  [Customere]xcatd service can not be started automatically after reboot for RHEL7.2 #492 

immarvin:
/usr/lib/lsb/remove_initd and /usr/lib/lsb/install_initd shipped in redhat-lsb-core has been a link to chkconfig in redhat7.2, whis will cause problem while enabling xcatd service after xCAT-server installation when redhat-lsb-core is installed


daniceexi:

PR #1016, opened 2016-04-29
Move genesis files from xCAT-genesis-builder to xCAT-genesis-scripts

gurevichmark:
Moving 2 genesis kernel files from xCAT-genesis-builder to xCAT-genesis-scripts - rsyslog.conf and dhclient-script
This will allow these 2 files to be built and packaged into xcat-core package.


whowutwut:
@daniceexi  I'm okay with these changes.   I'll let you take a look before we merge this is... 

@gurevichmark and I decided to go this route because having the files in two places `xCAT-genesis-builder` and `xCAT-genesis-scripts` will be confusing to whomever looks at the code...

Regarding your comment in #888... I think that requires a much larger effort to redesign how we are doing things in the genesis kernel.  We agree that we probably are not doing the right thing by modifying the rsyslog.conf file or the dhclient-scripts file (the man pages says not to modify it :-) ) http://linux.die.net/man/8/dhclient-script ... but this is what is being done today.   

I don't think we should redesign the framework, when we are simply trying to clean up some of the error messages at this time.

whowutwut:
I believe from the interlock meeting we decide to hold off on this pull request. Should we leave this or close it and use another once we decide the right way to handle the Genesis scripts and genesis build rpms?

whowutwut:
@gurevichmark this is not high priority...  to understand the genesis-base/genesis-scripts better.. let's make a change in addition to what is already in the pull request to remove rsyslog.conf and dhclient-script from the install, install.ubuntu files in xCAT-Genesis-builder so that we at least have something we can try out and test for our understanding.

whowutwut:
I'm going to close this for now.

Issue #1017, opened 2016-05-02
[customer] xcatd process uses up the cpu resource when xcatd port is scanned


daniceexi:
Two examples shown using a openssl client or nap to scan cat port 3001

*  Run openssl client from IO node

~~~~
$ openssl s_client -quiet -no_ssl3 -no_ssl2 -connect 192.168.45.20:3001 -rand /bin/nice
70728 semi-random bytes loaded
depth=1 CN = xCAT CA
verify error:num=19:self signed certificate in certificate chain
verify return:0
test
~~~~

  Top on xcat management node
~~~~
op - 15:31:13 up 133 days, 15:19,  3 users,  load average: 7.47, 11.31, 5.67
Tasks: 280 total,  27 running, 253 sleeping,   0 stopped,   0 zombie
%Cpu(s): 10.5 us,  2.1 sy,  0.0 ni, 87.4 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
KiB Mem :  6776320 total,   759104 free,   756288 used,  5260928 buff/cache
KiB Swap:  4095936 total,  4073152 free,    22784 used.  5068672 avail Mem 

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                                                                             
16518 root      20   0   58176  49344   7168 R 100.0  0.7   0:40.29 xCATd SSL: Inst                       
~~~~

* From IO node run nmap:
~~~~
$ nmap -sV --reason -PN -n -p 3001 192.168.45.20

Starting Nmap 6.40 ( http://nmap.org ) at 2016-04-29 15:24 PDT
Nmap scan report for 192.168.45.20
Host is up, received arp-response (0.000063s latency).
PORT     STATE SERVICE     REASON  VERSION
3001/tcp open  ssl/nessus? syn-ack
MAC Address: E4:1F:13:06:20:40 (IBM)

Service detection performed. Please report any incorrect results at http://nmap.org/submit/ .
Nmap done: 1 IP address (1 host up) scanned in 151.90 seconds
~~~~

  Top on xcat management node.
~~~~
op - 15:27:01 up 133 days, 15:14,  3 users,  load average: 17.91, 6.16, 2.25
Tasks: 304 total,  50 running, 254 sleeping,   0 stopped,   0 zombie
[...]

PR #1018, opened 2016-05-02
bug 1017: make a sleep before retrying too much

daniceexi:
when xcatd trys to read the request data from an xcatd client,
  if there are too many retry, sleep for a while before retrying
  too much.


daniceexi:
this pull request is for the bug #1017 in 2.9 branch.

daniceexi:
Close this pull request and replace it with #1027

PR #1020, opened 2016-05-02
Fix readthedocs build error

gurevichmark:
An attempt to fix readthedocs build error. Build on readthedocs has been failing since April 1st.


whowutwut:
Let me merge this in. I wonder if the back ticks are causing it. Let's see

whowutwut:
@gurevichmark  Looks good.

whowutwut:
Took a bit of trial and error but I renamed the `xcat2/xcat-core/README -> README.rst `
Then I added the badges  (http://docs.readthedocs.io/en/latest/badges.html) so that we have a place to easily track the status of the documentation build.  The build has been failing for 1 month and went undetected.   

@daniceexi @tingtli FYI

PR #1021, opened 2016-05-03
small doc clenaup - simplify install of postgres instructions, change Redhat->RedHat

whowutwut:
- Clean up the postgres instructions to simplify the yum and zypper command with a wildcard
- Change Redhat to RedHat


cxhong:
looks good

PR #1022, opened 2016-05-03
Enable lsvm for kvm hypervisor

gurevichmark:
Enable existing code to display KVM VMs for kvm hypervisor.
Modify lsvm man page.


xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -79,7 +79,8 @@ sub handled_commands {
         #rvitals => 'nodehm:mgt',
         #rinv => 'nodehm:mgt',
         getrvidparms  => 'nodehm:mgt',
-        lsvm          => 'hypervisor:type',
+        #lsvm          => 'hypervisor:type',
```
whowutwut: I think we should just delete this line.  I don't think kvm supports this.

gurevichmark: Done in the second commit.

Issue #1011, opened 2016-04-29
[FVT]2.12: xcat command will hange after xcatd restart


junxiawang:
env: all platform
build:
lsdef - Version 2.12 (git commit 8bdfc11977a18713e02022ae73567731860c4406, built Thu Apr 28 21:50:44 EDT 2016)
Hot to reproduce:
```
yum install xCAT
then  chdef -t node -o c910f02fsp01 hwtype=cec serial=103A4EP mgt=hmc groups=cec hcp=c910hmc01 nodetype=ppc mtm=8233-E8B
------------------------------> Command will hang  
And after restart xcatd the xcat command will hang sometimes.
```

chenglch:
I add some debug code in Table.pm around Line 2584

     if (!defined($self->{nodelist})) {
        use Data::Dumper;
        xCAT::MsgUtils->message("S","cltest nodelist error ".Dumper($self));
      }
      use Carp 'verbose';
      $SIG{ __DIE__ } = sub { Carp::confess( @_ ) };
      **my ($nodeghash) = $self->{nodelist}->getAttribs({node => $node}, 'groups');**

The debug message is like this, seems a race condition when starting xcatd and get node data from db at the same time.

xcatd: possible BUG encountered by xCAT DB worker Can't call method "getAttribs" on an undefined value at /opt/xcat/lib/perl/xCAT/Table.pm line 2591.
 at /opt/xcat/lib/perl/xCAT/Table.pm line 2589.
        xCAT::Table::__ANON__('Can\'t call method "getAttribs" on an undefined value at /opt...') called at /opt/xcat/lib/perl/xCAT/Table.pm line 2591
        xCAT::Table::getNodeAttribs_nosub_returnany('xCAT::Table=HASH(0x10002886448)', 'c910f02fsp01', 'ARRAY(0x10001aebd50)') called at /opt/xcat/lib/perl/xCAT/Table.pm line 2290
        xCAT::Table::getNodeAttribs('xCAT::Table=HASH(0x10002886448)', 'c910f02fsp01', 'ARRAY(0x10001ad20d0)') called at /opt/xcat/lib/perl/xCAT/Table.pm line 2114
        xCAT::Table::getNodesAttribs('xCAT::Table=HASH(0x10002886448)', 'ARRAY(0x10002931878)', 'ARRAY(0x10002bd8a88)') called at /opt/xcat/lib/perl/xCAT/Table.pm line 346
[...]

PR #1026, opened 2016-05-04
nodelist table object should not fetch the attribs in groups

chenglch:
This patch intent to fix the undef reference for nodelist temporarily.

partial-issue #1011


daniceexi:
I have reviewed the change. The change is just a temporary fix to workaround this issue. @chenglch will dig into the issue and found a completed fix. So just merge it as a workaround.

whowutwut:
Does this issue also occur on 2.11 release?

PR #1027, opened 2016-05-04
bug 1017: make a sleep before retrying too much

daniceexi:
when xcatd trys to read the request data from an xcatd client,
  if there are too many retry, sleep for a while before retrying
  too much.


chenglch:
looks good to me

PR #1028, opened 2016-05-04
Add new SN setup case to support using bundle only in jenkins environment

hu-weihua:


whowutwut:
@hu-weihua  can you fill in the milestone and labels? Thanks
 
Looking at the sleep function, this essentially makes our test framework run much longer each time we have sleep time....  Could we do something better?  Some ping test or even a sshd test.. or even a postscript that touches a file so that we can look for that file in a loop.. then when a postscript writes that file, we can continue.  Adding blind sleep makes everything wait longer.. if we only need 300 seconds, then we now need to sleep 10 minutes. 

Or having sleep in a global configuration so that we could tune the value....

hu-weihua:
@whowutwut , I think the purpose of ``customize_sleep_for_sn`` is just do what you want.  we spend lot of time to sleep for nothing while installation, now, we first sleep a short while(now it's 5 minutes, it's enough for target node reboot), after 5 minutes, we will check the attribute ``status`` of node every 10 seconds, to make sure if the installation is over, ``status=booted``. If it is,  ``customize_sleep_for_sn`` will exit, then the rest steps will start to do.  
The first 5 minutes is a experience time.  the physical X server need 2 minutes to restart , need some time to send dhcp request and download boot loader, if we check the ``status`` earlier, maybe the server is still rebooting, and the ``status`` keep the origin values which come from some other test before, maybe equal ``booted``. So in order to make sure the value of  ``status`` is reliable, so 5 minutes is good option.  it's true 5 minutes is too long for other platform, for example VM and lpar, but depending on experience, some time maybe we hit network problem, or there are too many VMs in one server, the VM restarting is slower than we expected. so 5 minutes is an accepted value.

tingtli:
@whowutwut @hu-weihua  Since i am waiting for this case to kick off my test on Jenkins new environment, I will merge the pull request. We can open the discussion and updates when needed.

whowutwut:
@gurevichmark  did you do some investigation about when the status message is updated and when it is not?  

In the past I think we have seen that the status is not always updated correctly and so it's misleading. Is there a field that we can provide when the postscripts have completed running?  So that users can use xCAT status to script other work?  `=booted` might be only when the OS is done loading and xCAT postscript starts running, or is it when it's completed running?

gurevichmark:
My experience is with VMs only:
When os is loading the status is `installing`.
After the os is installed the vm is rebooted and the status is `booting`.
Then, when login prompt is displayed in the console, status get set to `booted`
I am not sure if `booted` status gets set before or after postscripts run, but one of the simple things we can do is to check the /var/log/xcat/xcat.log file for return from mypostscript.
From MN, something like `xdsh <node> "grep '/xcatpost/mypostscript.post return' /var/log/xcat/xcat.log"`

PR #1035, opened 2016-05-04
Improve existed bundle files and add new bundle files

caomengmeng:
Hi @tingtli 
Add new bundle files:
RH6.7, RH6.8, Sles12.1, Ubuntu14.04.4, Ubuntu16.04 and update x86 to be x86_64.
Also as request from @hu-weihua, at the last line of the files, there should be a return.

Please help to review code.


xCAT-test/autotest/bundle/1
```diff
@@ -0,0 +1,303 @@
+reg_linux_diskfull_installation_flat
```
hu-weihua: this file has a strange name ``1``

tingtli:
@caomengmeng a general problem is for rh6,* we don't have ppc64le support, so ,please remove the rh6.* ppc64le bundle

caomengmeng:
@tingtli  @hu-weihua 
1. Delete the invalid file: 1
2. Delete invalid 6.7/6.8_ppc64le.bundle

whowutwut:
@caomengmeng the testcase requested is to mark issues. If this adds a testcase for an issue that requested automation test you can reference this pull request and change testcase requested to testcase added in the issue. Should not use the label in a pull request

Issue #455, opened 2015-11-23
[FVT]2.12 rmvm -f -p could not work sometimes


junxiawang:
1.download the latest build
2.[root@c910f03fsp03v02 ~]# lsdef -v
lsdef - Version 2.11 (git commit 895ad1fa02cd4a69a35ffd07256ade3ffc6491aa, built Sun Nov 22 15:30:53 EST 2015)
[root@c910f03fsp03v02 ~]# 
3.
[root@c910f03fsp03v02 ~]# rmvm c910f03fsp03v03 -f -p
Error: kvm plugin bug, pid 6020, process description: 'xCATd SSL: rmvm to c910f03fsp03v03 for root@localhost: kvm instance' with error 'libvirt error code: 38, message: cannot unlink file '/var/lib/libvirt/images/c910f03fsp03v03.sda.qcow2': Success
' while trying to fulfill request for the following nodes: c910f03fsp03v03

ssh kvmhost:
[root@node-32030901 ~]# find / -name "c910f03fsp03v03*"
/var/lib/libvirt/images/c910f03fsp03v03.sda.qcow2
---------------------> the file still here

whowutwut:
@zet809  if you are busy, I'd like to give this one to @pdlun92

daniceexi:
In current code logic, the **rmvm -f -p** tried to destory the vm first and then delete the disk. But I am afraid the issue caused by the **destory** action did not finish when calling the delete of disk. We may need to check the power state to be **off** before deleting the disk.

An work around for this issue is, we should use two command to replace **-f -p**

* rpower off the node first
* rmvm -p to delete the vm

tingtli:
Seems it does not work also.
c910f03fsp03v04:~ # rpower c910f03fsp03v05 stat
c910f03fsp03v05: off
c910f03fsp03v04:~ # rmvm c910f03fsp03v05 -f -p
c910f03fsp03v04:~ # mkvm c910f03fsp03v05 -s 20G
c910f03fsp03v05: Error: Storage creation request conflicts with existing file(s)

immarvin:
seems that ``rmvm c910f03fsp03v05 -f -p``  failed to destroy the volume

cannot recreate now

daniceexi:
@immarvin I tried with with following command to replace the rmvmk <node> -f -p, it worked.
~~~~
rpower <node> off
rmvm <node> -p 
~~~~

Could you help FVT to change the current test frame to be following to see whether this is a workaround.
[...]

PR #1039, opened 2016-05-04
call info() before deleting vm volume

gurevichmark:
This pull request fixes issue #455.

1. Changed destroy() to shutdown(). The libvir documentation describes destroy() as an equivalent of pulling a power cord and recommends using shutdown() instead.

2. get_info() needs to be called before deleting a volume. This seems to be the only thing that works. Adding sleeps between shutdown and delete volume did not solve the problem.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -1735,7 +1735,7 @@ sub rmvm {
     if ($currstate eq 'on') {
         if ($force) {
             $currxml = $dom->get_xml_description();
-            $dom->destroy();
+            $dom->shutdown();
         } else {
             xCAT::SvrUtils::sendmsg([ 1, "Cannot rmvm active guest (use -f argument to force)" ], $callback, $node);
             return;
@@ -1753,6 +1753,8 @@ sub rmvm {
             my $file = $disk->getAttribute("file");
             my $vol  = $hypconn->get_storage_volume_by_path($file);
             if ($vol) {
+                # Need to call get_info() before deleting a volume, without that, delete() will fail. Issue #455
+                $vol->get_info();
                 $vol->delete();
             }
         }


whowutwut:
I've reviewed and OK to merge

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -1735,7 +1735,7 @@ sub rmvm {
     if ($currstate eq 'on') {
         if ($force) {
             $currxml = $dom->get_xml_description();
-            $dom->destroy();
+            $dom->shutdown();
```
daniceexi: @gurevichmark As you mentioned in the comment the **shutdown** means 'OS' based shutdown. In our code, it maps to the **rpower softoff**.

Generally the **shutdown** will take more time than destroy and the consumed time cannot be estimated since it depends on the OS. If changing to **shutdown**, I am not sure whether it would impact the rest of disk delete work?

Did you verify this change when the OS is in running status?

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -1753,6 +1753,8 @@ sub rmvm {
             my $file = $disk->getAttribute("file");
             my $vol  = $hypconn->get_storage_volume_by_path($file);
             if ($vol) {
+                # Need to call get_info() before deleting a volume, without that, delete() will fail. Issue #455
```
daniceexi: @gurevichmark Good catch.
But this is a workaround, maybe there's a bug in the libvert. Could we find any clue from internet? We may open a bug to libvert.

xCAT-server/lib/xcat/plugins/kvm.pm
gurevichmark: I think calling $dom->shutdown() calls a ``shutdown`` from libvirt. When reading the description of the destroy(), it seemed to me that shutdown() was better:

``destroy domain [--graceful]
           Immediately terminate the domain domain.  This doesn't give the
           domain OS any chance to react, and it's the equivalent of ripping
           the power cord out on a physical machine.  In most cases you will
           want to use the shutdown command instead.  ``

I tried using ``destroy --graceful`` but that still did not work and volume could not be deleted.
I verified this fix while the vm was running.

xCAT-server/lib/xcat/plugins/kvm.pm
gurevichmark: Online I found a recommendation to run pool-refresh() before calling vol-delete(), but pool-refresh() requires a poolname which does not appear to be readily available inside our rmvm function, so I decided to use get_info() to minimize the changes.

gurevichmark: There is already a bug opened for this issue
https://bugzilla.redhat.com/show_bug.cgi?id=1293804
It was fixed Apr 25, 2016 into version libvirt 1.3.3. But can we assume that the user will have this new version ?

PR #1042, opened 2016-05-05
New xcatdebugmode cases

caomengmeng:
@tingtli 
Add 5 cases for xcatdebugmode according to the design doc: https://github.com/xcat2/xcat-core/wiki/support%20new%20values%20for%20site.xcatdebugmode
and test steps: 
xcatdebugmode_off_rhels
set_xcatdebugmode_one_rhels
set_xcatdebugmode_value_two_rhels
xcatdebugmode_off_sles
set_xcatdebugmode_one_sles
xcatdebugmode_off_ubuntu

Please help to review cases.


whowutwut:
Just a general question, why are all testcases added into a single file under a function.  Like `xcatdebugmode/case0` rather then creating smaller testcase files , case0, case1, case2, etc where each file can test a smaller subset or a specific function and they run one after another.  

Then we have the flexibility to disable certain testcases if we have a bug in a specific command... 

For example.. regression testing could be testing 10 cases... case1-case10. Then FVT could create new automated testcases to test new function being added based on the externals...... those would initially fail because development is not complete but those testcases can be bypassed from normal regression testing until the new function is committed and tested.

tingtli:
@whowutwut 
I remember we support case0, case1... file but the cases which we use is specified in the bundle file which is a different file from the case file. We put the case names in  the bundle file and specify a bundle file name, then We could choose which case could be run.

PR #1043, opened 2016-05-05
Fix issue 998 again,  restore xcattest back to original before bash dash problem

hu-weihua:
Restore xcattest back to original before bash dash problem. using workaround


daniceexi:

PR #1045, opened 2016-05-05
Do not undef nodelist reference in DB process

chenglch:
This patch add check method to wait for the DB process in xcatd.
Enhance the close method in Table.pm to disable the undef operation
for nodelist in DB process as the reference has already been cached
as a weak reference.

close-issue: #1011


whowutwut:
@chenglch please set the side milestone and labels

daniceexi:
The code logic is OK for me, but I have a little comment for the error message. To avoid the failure in automation, I merge this change first.

Issue #834, opened 2016-03-23
[FVT]Man page for bmcdiscover is not synced completely 


caomengmeng:
In bmcdiscover man page, the option "-c" is replaced by "--check", but there are still "-c" existed.

OPTIONS 
```
       -c|--check
                 Check
```

EXAMPLES
```
       4. To check if the username or password is correct against the BMC:

           bmcdiscover -i 10.4.23.254 -u USERID -p PASSW0RD -c
```

Please make changes.

zet809:
@whowutwut Hi, victor, in pull request #704, you have modified supporting of "-c|--check" to be "--check". But the manpage -c is still there, will you pls help to modify it?

whowutwut:
Yea I will make the changes

caomengmeng:
Passed
Build:
```
[root@c910f02c04p08 /]# lsxcatd -a
Version 2.12 (git commit cccd99d10331da69c74c6d7a69bd734595602abf, built Sat May  7 01:15:31 EDT 2016)
This is a Management Node
dbengine=SQLite

```
OPTIONS
```
 --check   Check BMC administrator User/Password.
```

EXAMPLES
```
  4. To check if the username or password is correct against the BMC:

           bmcdiscover -i 10.4.23.254 -u USERID -p PASSW0RD --check
```

PR #1051, opened 2016-05-05
Make the --check example consistent with the usage in bmcdiscover man page

whowutwut:
Fix Issue #834 

The --check is used in the usage but the example below is using the -c option.
Both the --check and -c option will work, but sync up the example to match the usage.


daniceexi:

Issue #1025, opened 2016-05-04
[FVT] File 98-mlx.rules in package xcat-genesis-scripts conflict with the same file in package xcat-genesis-base


junxiawang:
env: rh7.2 (x86_64,ppc64),ubuntu14.04.4 x86_64
build:latest build
How to reproduce:
```
================================================================================
Install  2 Packages (+105 Dependent packages)

Total download size: 73 M
Installed size: 293 M
Downloading packages:
warning: /xcat-dep/rh5/ppc64/conserver-xcat-8.1.16-10.ppc64.rpm: Header V4 DSA/SHA1 Signature, key ID c6565bc9: NOKEY
Public key for conserver-xcat-8.1.16-10.ppc64.rpm is not installed
Public key for perl-xCAT-2.12-snap201605030930.noarch.rpm is not installed
--------------------------------------------------------------------------------
Total                                               32 MB/s |  73 MB  00:02
Retrieving key from file:///xcat-dep/rh7/ppc64/repodata/repomd.xml.key
Importing GPG key 0xC6565BC9:
 Userid     : "xCAT Security Key <xcat@cn.ibm.com>"
 Fingerprint: f75b 1bf6 78b6 44fd f3aa cfc8 60a3 e9ac c656 5bc9
 From       : /xcat-dep/rh7/ppc64/repodata/repomd.xml.key
Running transaction check
Running transaction test


Transaction check error:
  file /opt/xcat/share/xcat/netboot/genesis/ppc64/fs/etc/udev/rules.d/98-mlx.rules conflicts between attempted installs of xCAT-genesis-scripts-ppc64-1:2.12-snap201605030931.noarch and xCAT-genesis-base-ppc64-1:2.11-snap201511160819.noarch
  file /opt/xcat/share/xcat/netboot/genesis/ppc64/fs/sbin/loadmlxeth conflicts between attempted installs of xCAT-genesis-scripts-ppc64-1:2.12-snap201605030931.noarch and xCAT-genesis-base-ppc64-1:2.11-snap201511160819.noarch

Error Summary
-------------

```

junxiawang:
in x86_64 install_xcat.log:
```
Install  2 Packages (+110 Dependent packages)

Total download size: 64 M
Installed size: 190 M
Downloading packages:
warning: /xcat-dep/rh5/x86_64/conserver-xcat-8.1.16-10.x86_64.rpm: Header V4 DSA/SHA1 Signature, key ID c6565bc9: NOKEY
[...]

PR #1052, opened 2016-05-06
Add back in the changes for mellanox workaround 

whowutwut:
After @daniceexi removed the following files from the genesis-base in #1025, adding back the changes to genesis-scripts that @jjohnson42  made


daniceexi:
I have reviewed the change and OK to merge.

PR #1053, opened 2016-05-06
The bmcdiscover command does not support -c option, removing from documentation

whowutwut:
Fix issue #834.  

After @daniceexi reviewed the pull request #1051, he mentioned that -c option is not supported in the code.  That is right, -c is not supported.  Removing it completely from the doc


daniceexi:

Issue #1054, opened 2016-05-06
[FVT]:add rhels6.8 support and remove -n for copycds


tingtli:
xcat 2.12 

when I run copycds , i still need to add -n option to specify the rhels6.8 version.

Please add rhels6.8 official support then we don't need to use -n options for copycds

immarvin:
fixed https://github.com/xcat2/xcat-core/pull/1055

PR #1055, opened 2016-05-06
add discinfo for rhels6.8

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1054
add discinfo for rherls6.8.
The discinfo is from .discinfo of rhels6.8-RC1 iso


daniceexi:

whowutwut:
I still think that we should first try and provide an automatic way to determine the os versions based on the files on the ISO image, then as a fall back, write the ISO number into the code..... then this kind of thing will not block copycds when doing testing...

daniceexi:
@immarvin Could you take a look the possibility to use the configuration files in the disc to determine the OS and image name?

whowutwut:
It's possible. I made the change to do it for powerKVM, first try to figure out. If can't then fail to the mapping...

We discussed it but I don't remember the history behind why we decided that we still needed to keep the identifier mapping method... We only do this mapping for red hat....

Issue #1056, opened 2016-05-06
[FVT]: rcons returns SSLVerifyCallback():  error #7: certificate signature failure on rhels6.8


tingtli:
xCAT 2.12 rhesl6.8
 
[root@c910f02c01p22 /]# rcons c910f02c01p23 -f
console: SSLVerifyCallback(): error with certificate at depth: 0
console: SSLVerifyCallback():  issuer  = /CN=xCAT CA
console: SSLVerifyCallback():  subject = /CN=c910f02c01p22
console: SSLVerifyCallback():  error #7: certificate signature failure
console: SSL negotiation failed
4938:error:0D0C50A1:asn1 encoding routines:ASN1_item_verify:unknown message digest algorithm:a_verify.c:141:
4938:error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed:s3_clnt.c:914:

[root@c910f02c01p22 /]# getmacs c910f02c01p23 -f -D
c910f02c01p23:

daniceexi:
From the error message that **conserver** encountered ``message digest algorithm`` problem when using our new xcat certificates which digested with ``sha256``; Refer to the pull request #927 

The problem only happens for rh6 ppc64, looks like the ssl library on rh6 ppc64 has problem to support sha256.

So the temporary fix is to revert the pull request #927, and then we can take a look of this issue in 2.12.1.

tingtli:
@daniceexi  It seems the rcons does not work with the temporary fix
[root@c910f02c01p22 /]# lsxcatd -v
Version 2.12 (git commit bb5988332333745cbe759f5c8299c8379170b994, built Mon May  9 09:30:39 EDT 2016)
[root@c910f02c01p22 /]# rcons c910f02c01p23 -f
console: SSLVerifyCallback(): error with certificate at depth: 0
console: SSLVerifyCallback():  issuer  = /CN=xCAT CA
console: SSLVerifyCallback():  subject = /CN=c910f02c01p22
console: SSLVerifyCallback():  error #7: certificate signature failure
console: SSL negotiation failed
4979:error:0D0C50A1:asn1 encoding routines:ASN1_item_verify:unknown message digest algorithm:a_verify.c:141:
4979:error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed:s3_clnt.c:914:

daniceexi:
@tingtli 
[...]

PR #1059, opened 2016-05-06
Revert "replace the digest algorithm of xCAT ssl certificate from sha1 to sha256"

daniceexi:
Reverts xcat2/xcat-core#927


daniceexi:
Refer to the bug: #1056.

PR #1066, opened 2016-05-06
bmcdiscover command does not allow -u without -p specified in command

whowutwut:
Specifying the -u option to bmcdiscover command only took effect if the -p option was alos used. 

Moving the check so that either option can be specified to override the default in the site.passwd table


whowutwut:
I think this bug fix is low risk and should be added to the 2.12 GA RC build.

daniceexi:

zet809:
@whowutwut Hi, Victor, I am sorry to bring this modification out again since we encountered an issue when PCM using bmcdiscover to discover FSP for TUL servers. 
For TUL servers, there is no bmcusername enable for it, then our code will failed to get MTMS since we provide a default bmcusername when running ipmitool against it.

whowutwut:
Ok, so it seems like we should undo this change...

zet809:
Yes, will you do that or me?

Issue #1068, opened 2016-05-06
[FVT] vm's mac in database could not same with the vm definition's mac 


junxiawang:
env:sles11
build:c910f04x27v05:/opt/xcat/share/xcat/tools/autotest/testcase/installation # lsdef -v
lsdef - Version 2.12 (git commit 46a07a8d2512b23a68f2049be0c06632702e4193, built Fri May  6 09:30:45 EDT 2016)
How to reproduce:
```
rmvm CN -f -p
mkvm CN -s 40G
c910f04x27v05:/opt/xcat/share/xcat/tools/autotest/testcase/installation # tabdump kvm_nodedata |grep mac
    <type arch='x86_64' machine='rhel6.5.0'>hvm</type>
      <mac address='42:b1:0a:04:1b:06'/>
    <type arch='x86_64' machine='rhel6.5.0'>hvm</type>
      <mac address='42:5c:0a:04:1b:07'/>

c910f04x27v05:/opt/xcat/share/xcat/tools/autotest/testcase/installation # lsdef c910f04x27v06 -i mac
Object name: c910f04x27v06
    mac=42:b1:0a:04:1b:06
c910f04x27v05:/opt/xcat/share/xcat/tools/autotest/testcase/installation # lsdef c910f04x27v07 -i mac
Object name: c910f04x27v07
    mac=42:c1:0a:04:1b:07
c910f04x27v05:/opt/xcat/share/xcat/tools/autotest/testcase/installation # 

```

daniceexi:
The root cause was the commit a0ab377ecb998f0af9772f93fc34c0b338588b1d changed the node removing from ``destroy`` to ``shutdown`` in the ``rmfm -f`` command. But the ``shutdown`` does not really remove the node object from the kvm host, which will cause the next running of ``mkvm`` failed since the node was there in the kvm host.

Then in our automation scenario, we cleaned up the mac address and expected the ``mkvm`` to create a new one. The reality was the ``mkvm`` did create a new one and added to mac table, but it did not take effect in the vm since vm creating failed (because the prior ``rmvm`` did not remove the vm from kvm host).

The fix is change back to ``destroy`` from ``shutdown``;

daniceexi:
The fix has been merged by pull request #1069.

whowutwut:
@gurevichmark @daniceexi  are we still going to have the issue with removing the persistent disk with this change reverted?

gurevichmark:
[...]

PR #1069, opened 2016-05-07
issue 1068: kvm.pm: shutdown in rmvm will not destory the node object…

daniceexi:
… in kvm host, this will cause the next mkvm do nothing


daniceexi:
Fix for #1068.

Issue #760, opened 2016-03-02
[DEV]Missed the dhcp and tftp messages in sytem log


daniceexi:
After we change the way to store xCAT logs in xCAT 2.11, the dhcp and tftp log cannot be recorded in the system log file like /var/log/messages.

I saw this issue on RH MN.

neo954:
DHCP log is back.

Refer issue #989

daniceexi:
I have verified the fix.

PR #1085, opened 2016-05-11
Fix issue #760, log tftp logs to /var/log/messages

neo954:


neo954:
The patch fixed issue #760.

daniceexi:
The fix is good for me. Thanks @neo954 .

Issue #929, opened 2016-04-14
[DEV]zypper install is locked by the y2base on sles12.1 during the provisioning. 


cxhong:
working on sles12.1 hierarchy support for issue  #898,  during the server node provisioning, it hit following error message while it was trying to install otherpkgs.

````
pkgsarray: -perl-doc xcat/xcat-core/xCATsn xcat/xcat-dep/sles12/ppc64le/conserver-xcat, 3
yum: 0, apt: 0, zypper: 1
System management is locked by the application with pid 2729 (y2base).
Close this application before trying again.
System management is locked by the application with pid 2729 (y2base).
Close this application before trying again.
System management is locked by the application with pid 2729 (y2base).
Close this application before trying again.
System management is locked by the application with pid 2729 (y2base).
Close this application before trying again.

`````
after server node provisioned, ran updatenode command and it worked fine

`````
# updatenode fs1vm301 -P otherpkgs
fs1vm301: vpdupdate is not supported on the pSeries KVM Guest
fs1vm301: xcatdsklspost: updating VPD database
fs1vm301: xcatdsklspost: downloaded postscripts successfully
fs1vm301: Fri Nov 27 08:20:11 EST 2015 Running postscript: otherpkgs
fs1vm301: pkgsarray: -perl-doc xcat/xcat-core/xCATsn xcat/xcat-dep/sles12/ppc64le/conserver-xcat, 3
fs1vm301: yum: 0, apt: 0, zypper: 1
fs1vm301: Postscript: otherpkgs exited with code 0
fs1vm301: Running of postscripts has completed.

`````

Yong Song opened a bugzilla issue
https://bugzilla.linux.ibm.com/show_bug.cgi?id=140313

immarvin:
LTC ticket :https://bugzilla.linux.ibm.com/show_bug.cgi?id=140313

daniceexi:
@cxhong Looks like the Yast was trying to install xCATsn or something when the issue happened. Generally, the xCATsn is installed in otherpkgslist. But you were trying to install it in pkglist (Yast phase), right?

immarvin:
hi @daniceexi , for sles, the yast will be run on the 1st reboot after installation, the PBS otherpkgs is called by ``service xcatpostinit1 start``

cxhong:
the bugzilla bug still open.  

workaround:  after service node is up, run:
````
updatenode nodename -P otherpkgs
````

daniceexi:
[...]

PR #1087, opened 2016-05-11
fix issue#929@github,service network restart in <post-install scripts》 might break the service dependecy and will cause conflict between  YaST2-Second-Stage.service and xcatpostinit1 services

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/929

"network.service" will be stopped at the end of <post-install scripts> of autoyast.

for details, please refer to the https://github.com/xcat2/xcat-core/issues/929

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/post.xcat
+++ b/xCAT-server/share/xcat/install/scripts/post.xcat
@@ -421,6 +421,19 @@ updateflag.awk $MASTER 3002
 
 cd /
 
+# fix the issue#929@github, the "service network restart" in <sles post scripts>  
+# makes the 'network' service in start status, then it causes the
+# services YaST2-Second-Stage.service and xcatpostinit1 services to be run in parallel.
+# But the YaST2-Second-Stage.service calls the Yast2 which conflics with the zypper
+# which is called in xcatpostinit1->otherpkgs. 
+# The correct services start order is 'YaST2-Second-Stage.service'->'network'->'xcatpostinit1' in serial.
+# To skip the conflict, the network service should be stoped in the yast first stage. Then base on the service
+# dependency, the service start order will be 'YaST2-Second-Stage.service'->'network'->'xcatpostinit1' in serial.
+# Then the Yast2 will be stopped before running the zypper in otherpkgs.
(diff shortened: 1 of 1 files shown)


immarvin:
hi @cxhong @daniceexi @whowutwut , would you please review the fix so that it can be covered by FVT ASAP? thanks

daniceexi:
I have reviewed the change and merge it first so that it will get into today's build.

@cxhong , @whowutwut if you have any comments, please add it here, we can handle it tomorrow.

xCAT-server/share/xcat/install/scripts/post.xcat
```diff
@@ -421,6 +421,19 @@ updateflag.awk $MASTER 3002
...
+# fix the issue#929@github, the "service network restart" in <sles post scripts>  
+# makes the 'network' service in start status, then it causes the
+# services YaST2-Second-Stage.service and xcatpostinit1 services to be run in parallel.
+# But the YaST2-Second-Stage.service calls the Yast2 which conflics with the zypper
+# which is called in xcatpostinit1->otherpkgs. 
+# The correct services start order is 'YaST2-Second-Stage.service'->'network'->'xcatpostinit1' in serial.
+# To skip the conflict, the network service should be stoped in the yast first stage. Then base on the service
+# dependency, the service start order will be 'YaST2-Second-Stage.service'->'network'->'xcatpostinit1' in serial.
+# Then the Yast2 will be stopped before running the zypper in otherpkgs.
+if [[ $OSVER != sles11* ]]; then
```
whowutwut: Interesting syntax with the wildcard, i didn't know that was possible. 

Why are we testing `! sles11*`? That means SLES 9, SLES 10, SLES 12+ will execute this code...

immarvin:
hi @daniceexi , I agree with @whowutwut  that  
````
 +if [[ $OSVER != sles11* ]]; then 
````
will affect the sles distributions before "sles11",  and it is not safe enough.

I think we should restrict the modification only for sles12
````
 +if [[ $OSVER == sles12* ]]; then 
````
might be better.

immarvin:
hi, I refined the code in a new PR https://github.com/xcat2/xcat-core/pull/1096

PR #1094, opened 2016-05-12
Update cases for ubuntu installation time issue

caomengmeng:
@tingtli 
Update ubuntu installation cases to decrease/increase the wait time.
Please help to review.
Thanks.

### DIFF
--- a/xCAT-test/autotest/testcase/installation/Diskless_installation_flat_p8_le
+++ b/xCAT-test/autotest/testcase/installation/Diskless_installation_flat_p8_le
@@ -30,6 +30,7 @@ check:rc==0
 cmd:rpower $$CN boot
 check:rc==0
 cmd:sleep 200 
+cmd:a=0;while ! `lsdef -l $$CN|grep status|grep booted >/dev/null`; do sleep 10;((a++));if [ $a -gt 300 ];then break;fi done
 cmd:lsdef -l $$CN | grep status
 check:output=~booted
 cmd:xdsh $$CN date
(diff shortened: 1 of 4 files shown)


xCAT-test/autotest/testcase/installation/Diskless_installation_flat_p8_le
```diff
@@ -30,6 +30,7 @@ check:rc==0
 cmd:rpower $$CN boot
 check:rc==0
 cmd:sleep 200 
+cmd:a=0;while ! `lsdef -l $$CN|grep status|grep booted >/dev/null`; do sleep 10;((a++));if [ $a -gt 300 ];then break;fi done
```
daniceexi: I think we should not wait 3000+200 for a diskless install. If the diskless not cannot finish in 20m, it were failed. Even it might success, but from the performance perspective, it failed.

tingtli: We will check the status booted, if it is the status , the cmd exit.
That means we will first sleep 200 or 300 , then check the status.

whowutwut: Why do we need to use sleep?  

I think we should do the following: 
1) Add a timer in the framework so that any testcase that runs over some global timer value will fail, maybe this is 60 minutes
2) In the xCAT code, provide something that can be checked at the END of xCAT provisioning, this includes running the xCAT postscripts.  I think customer will also find this beneficial because they can wait until that status is updated before using the node.  Some ideas... some status in systemd, some entry in /opt/xcat/xcatinfo?  Can xCAT have better status .. like "booting" -> "installing" -> "installed" -> "post" -> "runnnig/complete"  or do we have a defined list of states that the user can reliably use to know the state of the OS?

whowutwut:
@tingtli  @caomengmeng  the "testcase_requested" label is to be used for Issues to help find issues that were missed by autotest that should be added into autotest.  If a new testcase is added for a specific Issue that requested it, the pull request number should be added to the Issue and the "testcase_requested" label should be removed and "testcase_added" issue should replace it to indicate that a testcase was added to autotest.

PR #1096, opened 2016-05-12
refine the code logic to stop network.service for sles12 and above

immarvin:
refine the PR https://github.com/xcat2/xcat-core/pull/1087 ,which is the fix for  issue https://github.com/xcat2/xcat-core/issues/929
only stop network service for sles12 or above

verified on sles12-ppc64le and sles11.4-x86_64


daniceexi:
The change cannot handle the case ``OSVER=sles111``, so the version comparison cannot using string comparison.

daniceexi:
The decision was to keep the old change.

PR #1099, opened 2016-05-13
fix #1083 for redhat and sles, support logging diskless installation when xcatdebugmode=2

### LINKED ISSUE #1083: [DEV] for diskless installation, the log can't forward to MN when xcatdebugmode=2
For redhat, sles, ubuntu.
The diskless installation log can't be forwarded to MN.

penguhyang:
fix #1083 
support redhat and sles for diskless.

### DIFF
--- a/perl-xCAT/xCAT/Schema.pm
+++ b/perl-xCAT/xCAT/Schema.pm
@@ -1131,7 +1131,7 @@ use xCAT::ExtTab;
 "                 currently supported values:\n".
 "                   '0':  disable debug mode\n".
 "                   '1':  enable basic debug mode\n".
-"                   '2':  enalbe expert debug mode\n".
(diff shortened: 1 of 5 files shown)


penguhyang:
@whowutwut Thanks for review.
If we want to add new feature for xcatdebugmode=1 not for xcatdebugmode=2 in the future.
It can be easier to distinguish from each other. So I think it's ok.

whowutwut:
Is there a need for two modes? I still don't know why someone would use xcatdebugmode=1 instead of 2?

daniceexi:
Generally the xcatdebugmode=2 was used to enable the debug feature that customer needs to interact with the system to make it continue, otherwise the OS system would hang. So if a debug function can work without user's interaction, it should be put in the xcatdebugode=1

PR #1101, opened 2016-05-13
Waiting longer time for console ready in SN_setup_case to resolve getmacs failed issue

hu-weihua:


daniceexi:
Talked with @hu-weihua that the failure of getmacs might caused by a timing issue. So change the sleep to 20s what we used in the original automation framework.

PR #1108, opened 2016-05-13
Add RHEL7.2 ppc64le support to discinfo.pm(master)

lyu571:
Update discinfo.pm for support RHEL 7.2 ppc64 le(little endian) in mainline trunk.


lyu571:
RHEL 7.2 ppc64 le has been supported, so close it.

PR #1100, opened 2016-05-13
support logging diskfull installation for PowerKVM

penguhyang:
fix #1093 
support for PowerKVM for diskfull installation.

PR #1111, opened 2016-05-13
Modify the getinstdisk to create files called /tmp/xcat.install_disk to indicate xCAT related files  (DO NOT MERGE until after branch 2.12)

whowutwut:
I believe having this change in name will allow admins to better debug and understand immediately which files are created by xCAT.  This is especially true since we enabled ssh to the installer with `xcatdebugmode`.


whowutwut:
TODO: Will need to also make similar changes based on the files created in Pull request #1100 when that's merged in

xuweibj:
I agree with the code.

PR #1112, opened 2016-05-13
call get_info() for rmvm -f -p

gurevichmark:
Adding back get_info() before the volume delete.

```
Hi Mark,
 
Thanks for the digging.
 
From the RH bugzilla, someone mentioned your workaround that dump the volume definition before the deleting 
operating might help. So with the current retry logic, we can take back the '$vol->get_info();' before the delete 
operation.

Wang Xiaopeng 
```


daniceexi:
I have reviewed the change, it's safe to be merged in 2.12.

PR #1113, opened 2016-05-13
Remove --master from mkvm man page and usage

gurevichmark:
Based on Xiaopeng's recommendation on May 13, 2016 removing `--master` flag from `mkvm` command usage and man page.


daniceexi:

PR #1121, opened 2016-05-16
Add sles, redhat migration test cases

caomengmeng:
@tingtli 
Hi Ting Ting
Please help to review Sles RH migration test cases.
I already run them in automation environment.
2.10/2.11.1 --> 2.12
All passed.
Thanks.


tingtli:
@caomengmeng  some general comments
1) Need to add some values migration verification such as "xcatsslversion" 
2) Need to add the cases in the bundle file to let the cases run in the daily regression at the end of cycle.
Since we are very close to GA.  I will merge at the moment. Please do the case refinement later.

Issue #1103, opened 2016-05-13
xcat2.12 can not get the right disk name for RHEL on PowerKVM


fenggaobj:
When deploying compute node by diskfull method, it will be failed because it can not find the disk to install OS. 
The default disk of PowerKVM is "vda", but the disk name gave by xcat is "sda".

xuweibj:
hi @fenggaobj 
Now in our code, we just choose sd* or hd* as install disk.
Could you tell me why you named disk as "vd*"? And do you have any other names except [s/h/v]d*?

fenggaobj:
The disk name is provided by PowerKVM OS default. we can not configure them.
Or you can refer to the ticket:
https://sourceforge.net/p/xcat/bugs/4665/   
https://sourceforge.net/p/xcat/bugs/4535/

daniceexi:
@fenggaobj If this is a must have fix, we will try it make it gets into 2.12 release.

whowutwut:
@daniceexi @xuweibj  if I can make a guess here. I think we are not cleaning up the drives correctly when we are doing installs ... maybe we are now with the fix in 2.12 code...  @fenggaobj  can you provide the version of xCAT you are using?

daniceexi:
@whowutwut 
@xuweibj has found the root cause, we missed the support when the install disk is named as ``vd[a,b..]``.  We cannot recreate the issue locally, maybe the problem encountered by PCM was because the vm was created by Kimchi. This is a bug, @xuweibj has given PCM a workaround. But we need to test the change on more platforms.

tingtli:
@daniceexi What do you mean by saying test the change on more platforms?
Do you mean run regression on all the supported platform and run test on the vm which was created by Kimchi?

daniceexi:
@tingtli Run the fix on kimchi created vm will be covered PCM. xCAT team needs to verify that the change won't impact the basic functions of ``get-install-disk``. Besides the automation, @caomengmeng needs to verify the coverage which cannot be covered in he automation.

daniceexi:
The code has been merged. And the change has been verified by @fenggaobj , so close this issue.

PR #1123, opened 2016-05-17
fix issue 1103:xcat2.12 can not get the right disk name for RHEL on P…

zet809:
…owerKVM

For issue #1103


daniceexi:
I have reviewed the code change and talked with @zet809 , the change is safe for merging.

PR #1129, opened 2016-05-17
modify bmc testcase

junxiawang:
ting ting please help merge

### DIFF
--- a/xCAT-test/autotest/testcase/bmcdiscover/cases0
+++ b/xCAT-test/autotest/testcase/bmcdiscover/cases0
@@ -7,7 +7,7 @@ check:output=~Usage
 end
 
 start:bmcdiscover_nmap_range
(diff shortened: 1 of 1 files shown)


whowutwut:
@junxiawang do we also have testcase a for the error condition when we do not pass in the correct options to bm discover? Like the user/password.

Looks like we are also not verifying the -w is writing to the xcatdb. check before -w then after to make sure an entry is added? What about testing the error handling for bmcdiscover? Or testing the --check option?

whowutwut:
Also need to fill in the milestone and label for this pull request

junxiawang:
@whowutwut  I will modify the bmc cases but not in this pull request .This pull request is just for some cases's output changed according to bmc code change.

tingtli:
@junxiawang  I will merge your change to remove the error everyday. Please update the cases according to Victor's comments.

Issue #1122, opened 2016-05-16
[Customer]The '/dev' file system is changed to ready-only(ro) after install xcat-core 2.12


fenggaobj:
The '/dev' file system is changed to ready-only(ro) after install xcat-core 2.12

[root@pcmrhelsmoke log]# mount
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime)
proc on /proc type proc (rw,nosuid,nodev,noexec,relatime)
devtmpfs on /dev type devtmpfs (**ro**,nosuid,size=3936960k,nr_inodes=984240,mode=755)
securityfs on /sys/kernel/security type securityfs (rw,nosuid,nodev,noexec,relatime)
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev)

[root@pcmrhelsmoke log]# lsxcatd -a
Version 2.12 (git commit a05c28c05ec91e47f68bfecad468a209fd962f21, built Thu May  5 01:43:04 EDT 2016)
This is a Management Node
cfgloc=Pg:dbname=xcatdb;host=21.21.0.4|xcatadm
dbengine=Pg
dbname=xcatdb
dbhost=21.21.0.4
dbadmin=xcatadm

tingtli:
@fenggaobj  what is the OS of the system?

whowutwut:
@daniceexi  is this planned to be fixed for the GA?

fenggaobj:
The OS is RHEL7.2 X86_64.

tingtli:
@daniceexi @whowutwut 
I could reproduce this defect on our system.
Please take a look at this defect.

daniceexi:
This issue happens on rh7.2 after the running the ``genimage`` command.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1131

daniceexi:
The fix has been merged. But we still want to enhance this change, so move this issue to 2.12.1.

whowutwut:
@daniceexi what is the enhancement that we want to do for the next release?

daniceexi:
@whowutwut Use the ``cp`` to replace the ``cpio``, there's comment in the #1131.

daniceexi:
Looks like cp has problem to handle the specific device files, so keep the original fix.

PR #1131, opened 2016-05-17
fix issue#1122@github;create /dev under chroot env based on /dev on host;use lazy mount for sysfs and procfs to make sure they are umounted no matter whether busy

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/1122
for /proc and /sys, they are safe to be mounted on chrooted env., since they are readonly themselves.
for /dev under chrooted rootimg, we cannot use readonly mount, because it will make /dev/ on the management node readonly; 
we cannot use RW mount either, because when the "genimage" is terminated with either CTRL+C or session disconnect, the genimage will be terminated for user,  but genimage might not call the interrupt handler on SIGINT or SIGTERM correctly, the mounted /dev failed to be umounted, if user run "rm -rf /install/netboot/<osver>/<arch>/compute/rooimg", the /dev/ on management node will be cleaned up and the management node will be broken.
The fix for /dev in this PR is a safe way.

### DIFF
--- a/xCAT-server/share/xcat/netboot/rh/genimage
+++ b/xCAT-server/share/xcat/netboot/rh/genimage
@@ -91,15 +91,18 @@ sub mount_chroot {
     my $rootimage_dir = shift;
 
     #postinstall script of some packages might access the /proc, /sys and /dev filesystem
-    #For Redhat7 or above, mount these directories readonly from host to avoid error messages
+    #For Redhat7 or above, mount /proc,/sys from host to avoid error messages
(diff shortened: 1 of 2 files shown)


daniceexi:
I have reviewed the change, the code logic is good OK. To catch the 2.12 release, this change will be checked in now.
**But I want to use the ``cp`` to replace the ``cpio`` when copy the /dev from xCAT MN to chroot directory for new image. This can be done 2.12.1.**

Issue #1120, opened 2016-05-16
[FVT]2.12: could not get bmc information on sles11.4 using bmc discover 


junxiawang:
env:sles11.4
build:c910f02c01p14:~ # lsdef -v
lsdef - Version 2.12 (git commit 6e6c63c8157bd2af9341d3586c2d58a8b8f10971, built Sun May 15 09:30:32 EDT 2016)
How to reproduce:
```
c910f02c01p14:~ # cat /etc/*-release 
LSB_VERSION="core-2.0-noarch:core-3.2-noarch:core-4.0-noarch:core-2.0-ppc64:core-3.2-ppc64:core-4.0-ppc64"
NAME="SLES"
VERSION="11.4"
VERSION_ID="11.4"

c910f02c01p14:~ # bmcdiscover --range "10.4.30-40.254" -u USERID -p PASSW0RD -z
c910f02c01p14:~ # bmcdiscover --range "10.4.30-40.254" -u USERID -p PASSW0RD -z
c910f02c01p14:~ # echo $?
0
c910f02c01p14:~ # lsdef -v
lsdef - Version 2.12 (git commit 5ee31c9, built Wed May 11 09:30:34 EDT 2016)
c910f02c01p14:~ #
--------------------------------> using this command could not get any bmc info
But before the code change:
root@c910f03c05k10:~# bmcdiscover --range "10.4.30-40.254" -u USERID -p PASSW0RD
10.4.30.254,7912AC1,06VAG36,USERID,PASSW0RD
10.4.31.254,7912AC1,06VAE26,USERID,PASSW0RD
10.4.32.254,7912AC1,06VAD90,USERID,PASSW0RD
10.4.33.254,7912AC1,06VAE52,USERID,PASSW0RD
10.4.34.254,7912AC1,06VAF01,USERID,PASSW0RD
10.4.35.254,7912AC1,06CGTR0,USERID,PASSW0RD
10.4.36.254,7912AC1,06VAG60,USERID,PASSW0RD
10.4.37.254,7912AC1,06CGTR6,USERID,PASSW0RD
10.4.38.254,7912AC1,06VAG62,USERID,PASSW0RD
10.4.39.254,7912AC1,06BPKK9,USERID,PASSW0RD
10.4.40.254,7912AC1,06VCM35,USERID,PASSW0RD
root@c910f03c05k10:~# 
```

zet809:
[...]

PR #1134, opened 2016-05-17
Support lower version of nmap on sles11.4

cxhong:
for issue #1120 ,  for bmcdiscover command, it will call nmap to discovery host ip address.  on sles11.4, nmap has lower version 4.75 than other OS,  it used nmap -sP instead of nmap -sn for scan Host.   the output of this command is also different.

on sles11.4:
````
 # nmap -sP -n 10.4.30.254

Starting Nmap 4.75 ( http://nmap.org ) at 2016-05-17 20:15 EDT
Host 10.4.30.254 appears to be up.
MAC Address: 34:40:B5:B9:BA:D7 (Unknown)
Nmap done: 1 IP address (1 host up) scanned in 0.52 seconds
``````

on redhat:
````
# nmap -sn -n 10.4.30.254

Starting Nmap 6.40 ( http://nmap.org ) at 2016-05-17 16:14 EDT
Nmap scan report for 10.4.30.254
Host is up (0.00033s latency).
MAC Address: 34:40:B5:B9:BA:D7 (IBM)
Nmap done: 1 IP address (1 host up) scanned in 0.39 seconds
````

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -443,6 +443,7 @@ sub scan_process{
     my $callback = $::CALLBACK;
     my $children;    # The number of child process
     my %sp_children;    # Record the pid of child process
+    my $bcmd;
     my $sub_fds = new IO::Select;    # Record the parent fd for each child process
   
     if ( !defined($method) )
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: xCAT-server/lib/xcat/plugins/bmcdiscover.pm [hunk 1]
```diff
@@ -455,7 +456,18 @@ sub scan_process{
     # get live ip list
     ###########################################################
     if ( $method eq "nmap" ) {
-        my $bcmd = join(" ",$nmap_path," -sn -n $range | grep for |cut -d ' ' -f5 |tr -s '\n' ' ' ");
+        #check nmap version first
+        my $ccmd = "$nmap_path -V | grep version";
+        my $version_result = xCAT::Utils->runcmd($ccmd, 0);
+        my @version_array = split / /, $version_result;
+        my $nmap_version = $version_array[2];
+        # the output of nmap is different for version under 4.75
```
zet809: In fact, from https://nmap.org/changelog.html#5.10BETA1, it shows that from nmap 5.10,  the option -sP is replaced with -sn
So, 4.75 shall be replaced with (< 5.10) here.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -455,7 +456,18 @@ sub scan_process{
...
+        if (xCAT::Utils->version_cmp($nmap_version,"4.75") <= 0) {
+            $bcmd = join(" ",$nmap_path," -sP -n $range | grep Host |cut -d ' ' -f2 |tr -s '\n' ' ' ");
```
zet809: I think grep "appears to be up" is more precise, what is you idea?
So the command will be "$bcmd = join(" ",$nmap_path," -sP -n $range | grep \"appears to be up\" |cut -d ' ' -f2 |tr -s '\n' ' ' ");"

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -455,7 +456,18 @@ sub scan_process{
...
+        } else {
+            $bcmd = join(" ",$nmap_path," -sn -n $range | grep for |cut -d ' ' -f5 |tr -s '\n' ' ' ");
```
zet809: I think replace "grep for" with "grep -B1 up  | grep \"Nmap scan report\"" will be more precise.

zet809:
@cxhong Hi, casandra, since it is very near to release, we'd like to merge it first, and make modification later in 2.12.1.

PR #1141, opened 2016-05-18
Change the kernel package name to linux-image-generic for Ubuntu 16.04

neo954:
... Based on the actual shape of the Ubuntu GA ISO image.


daniceexi:

PR #1143, opened 2016-05-18
fix issue#1139@github, run <dpkg-reconfigure locales> noninteractive

immarvin:
the "ubuntu/compute.postinstall" includes 
````
dpkg-reconfigure locales
````
which raise a interactive dialog box to select locales, we have no chance to interact with it during genimage, the command should be run with "DEBIAN_FRONTEND=noninteractive"


daniceexi:

Issue #1148, opened 2016-05-19
[FVT]:initctl: command not found on Ubuntu16.04


tingtli:
When install xCAT 2.12 on Ubuntu16.04, we found following error message.
```
Certificate is to be certified until May 13 17:07:30 2036 GMT (7300 days)^M
Sign the certificate? [y/n]:^M
^M
1 out of 1 certificate requests certified, commit? [y/n]Write out database with 1 new entries^M
Data Base Updated^M
/^M
Created xCAT certificate.^M
dns server has been enabled on boot.^M
httpd has been restarted.^M
sh: initctl: command not found^M
sh: initctl: command not found^M
Failed to stop firewall^M
xCAT is now running, it is recommended to tabedit networks ^M
and set a dynamic ip address range on any networks where nodes ^M
are to be discovered. Then, run makedhcp -n to create a new dhcpd ^M
configuration file, and /etc/init.d/dhcpd restart. Either examine sample ^M
configuration templates, or write your own, or specify a value per ^M
node with nodeadd or tabedit.^M
Running '/opt/xcat/sbin/mknb ppc64', triggered by the installation/update of xCAT-genesis-scripts-ppc64 ...^M
Creating genesis.fs.ppc64.lzma in /tftpboot/xcat^M
The 'mknb ppc64' command completed successfully.^M
Restarting apache2 (via systemctl): apache2.service.^M
Setting up xcat-test (2.12-snap201605180931) ...^M
Processing triggers for libc-bin (2.23-0ubuntu3) ...^M
Processing triggers for ureadahead (0.100.0-19) ...^M
Processing triggers for systemd (229-4ubuntu4) ...^M
Processing triggers for ufw (0.35-0ubuntu2) ...^M
```

Then when we provision a Ubuntu16.04 compute node, i found some error in the xcat.log
```
Wed May 18 13:03:33 EDT 2016 Running postscript: syslog
grep: /etc/rsyslog.d/remote.conf: No such file or directory
cp: cannot stat '/etc/rsyslog.d/remote.conf': No such file or directory
./xcatlib.sh: line 467: initctl: command not found
./xcatlib.sh: line 481: initctl: command not found
Wed May 18 13:03:33 EDT 2016 postscript syslog return with 0
Wed May 18 13:03:33 EDT 2016 Running postscript: remoteshell

root     16773  0.0  0.0  15360  6144 ?        Ss   13:03   0:00 /usr/sbin/sshd -f /etc/ssh/sshd_config
[...]

PR #1150, opened 2016-05-19
fix issue#1148@github, support service management with systemd on ubuntu16.04

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1148

from ubuntu16.04, "systemd" is used instead of "upstart" as the default service management framework. "upstart" does not installed on ubuntu16.04 by default.

The "systemd" on ubuntu 16.04 put all the service unit files under "/lib/systemd/system" instead of "/usr/lib/systemd/system" , which is the case for systemd on ubuntu and sles.

We need to enhance the service management subroutines in Utils.pm and xcatlib.sh to support the systemd on ubuntu16.04

### DIFF
--- a/perl-xCAT/xCAT/Utils.pm
+++ b/perl-xCAT/xCAT/Utils.pm
@@ -3896,30 +3896,55 @@ sub servicemap{
   my $path=undef;
   my $postfix="";
   my $retdefault=$svcname;
+  my $svcmgrcmd;
   if($svcmgrtype == 0){
+     #for sysvinit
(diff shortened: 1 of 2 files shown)


immarvin:
verified on rh7.2 and ubuntu 16.04

zet809:
Looks good to me. Thx!

daniceexi:
I reviewed the change too. This code has been reviewed the @zet809 and @daniceexi, should be safe for checking in.

Issue #1146, opened 2016-05-19
Format of hosts is wrong when add input in /etc/hosts


xyzcaoli:
Platform: VM
OS:redhat 7.2
Steps in Converge:
Step 1: Install Converge
Step 2: Enable HA
Step 3: Create a cs cluster with 3 mn and 1 cn
After that, check /etc/hosts, the format is wrong

Looks this is makehosts issue, following step to re-produce in our env.
[root@cnv1m1 ~]# cat /etc/hosts
12.12.12.2 cnv1m1-eno50338560
12.12.12.3 cnv1m2-eno50338560
12.12.12.4 cnv1m3-eno50338560

[root@cnv1m1 ~]# makehosts

[root@cnv1m1 ~]# cat /etc/hosts
12.12.12.2 cnv1m1-eno50338560
cnv1m1-eno50338560
.private.zone cnv1m1-eno50338560
12.12.12.3 cnv1m2-eno50338560
cnv1m2-eno50338560
.private.zone cnv1m2-eno50338560
12.12.12.4 cnv1m3-eno50338560
cnv1m3-eno50338560
.private.zone cnv1m3-eno50338560
11.11.11.2 cnv1m1 cnv1m1.private.zone cnv1m1-eno33559296
11.11.11.3 cnv1m2 cnv1m2.private.zone cnv1m2-eno33559296
11.11.11.19 cnv1 cnv1.private.zone cnv1-eno33559296
11.11.11.4 cnv1m3 cnv1m3.private.zone cnv1m3-eno33559296
11.11.11.6 cnv1c1 cnv1c1.private.zone cnv1c1-eno33559296
11.11.11.8 cnv1c2 cnv1c2.private.zone cnv1c2-eno33559296
172.29.5.11 cnv1m1-eno16780032 cnv1m1-eno16780032.ma.platformlab.ibm.com
172.29.5.12 cnv1m2-eno16780032 cnv1m2-eno16780032.ma.platformlab.ibm.com
172.29.5.19 cnv1-eno16780032 cnv1-eno16780032.ma.platformlab.ibm.com
172.29.5.13 cnv1m3-eno16780032 cnv1m3-eno16780032.ma.platformlab.ibm.com
172.29.5.14 cnv1c1-eno16780032 cnv1c1-eno16780032.ma.platformlab.ibm.com
11.11.11.5 cnv1c1-bmc cnv1c1-bmc.private.zone
[...]

PR #1151, opened 2016-05-19
fix issue 1146: Format of hosts is wrong when add input in /etc/hosts

zet809:
Fix issue #1146


daniceexi:
I have reviewed the change, which is safe for 2.12 release.

Issue #1115, opened 2016-05-16
[DEV] Don't run rspconfig after hardware discovery finished for servers Other than S822L(TLT)


zet809:
For S822L servers (Tuleta), the in-band ipmi operation is not supported (It use FSP to simulate the function of BMC such as supporting ipmi interface). In order to configure FSP ip, user account, we have to run 'rspconfig' to configure FSP out-of-band. 
But for later servers such as S822LC (Firestone), S812LC (Habanero) and even Garrison, there is full function of BMC is supported in-band from the host OS. We have bmcsetup to configuration BMC.

So, the purpose of this defect is:
1. For servers that support in-band BMC configuration, just remove the temp BMC node if there is.
2. If no in-band BMC configuration is supported, go to the original progress to use rspconfig to set BMC ip and user account.

whowutwut:
@zet809  can you add some description for this issue and what the problem is when we run rspconfig after hardware discovery?  Should be S822L**C** right? what is TUT?  Tuletta?

zet809:
@whowutwut Not S822LC, what I mean is S822L which is Tuleta servers.

whowutwut:
Thanks for the clarification

PR #1153, opened 2016-05-19
fix issue 1115: Don't run rspconfig after hardware discovery finished…

zet809:
… for servers Other than S822L(TLT)

For issue #1115 
The modifications are:
1. in findme request, add bmcinband attribute if the BMC can be accessed in-band
2. within hardware discovery process, the bmcinband is set, just run "rmdef" against the temp BMC node.


whowutwut:
I think if we want this to be in 2.12.1, we need to re-create the pull request for 2.12 branch, or we need to cherry pick back the fix after GA.

daniceexi:

Issue #1149, opened 2016-05-19
[FVT]ship ntpdate in Ubuntu16.04 to support postscript setupntp


tingtli:
xCAT 2.12 on Ubuntu16.04 
Run postscript setupntp and get following error on Ubuntu16.04
```
Wed May 18 13:03:35 EDT 2016 Running postscript: setupntp
0 HPCbootstatus HPCbootstatus.aix HPCbootstatus.linux HPCbootstatus.linuxinit _ssh _xcat addcudakey addsiteyum aixlitesetup aixremoteshell aixswupdate aixvgsetup allowcred.awk ca chef-client confGang config_bootnicsea config_chef_client config_chef_server config_chef_workstation config_cuda config_puppet_client config_puppet_server configbond configefi configeth configeth_aix configfirewall confighfi configib confignics confignimsh createFS cvs_template.pl cvs_template.sh db2install disableconsistentNICrename efibootmgr enabledebug enablekdump enablesysreq getcredentials.awk getpostscript.awk hardeths hostkeys install_chef_client install_chef_server install_chef_workstation install_lsf install_puppet_client install_puppet_server joindomain killsyslog locktftpdir.awk lsf_startup make_sn_fs makeinitrd mkhyperv mkresolvconf mountinstall mypostscript mypostscript.post odbcsetup ospkgs otherpkgs reboot redirectps remoteshell replace_byid_device repos resyslog routeop runxcatpost serialconsole servicenode setbootfromdisk setbootfromnet setiscsiparms.awk setroute setupLDAP setupdockerhost setupesx setupnfsv4replication setupntp setuppostbootscripts setupscratch startsyncfiles.aix startsyncfiles.awk sudoer syncfiles syscloneimgupdate syslog umountpost unlocktftpdir.awk updateflag.awk updatenetwork upflagx64.exe upflagx86.exe uploadboot xHRM xcataixpost xcataixscript xcatclient xcatdsklspost xcatflowrequest xcatinstallpost xcatlib.sh xcatpostinit xcatpostinit1 xcatserver NTP server is not running
./setupntp: line 131: ntpdate: command not found
  ntpdate -t5 10.4.31.1 failed
```
```
root@c910f03c17k27:/var/log/xcat# which ntpdate
root@c910f03c17k27:/var/log/xcat#
```
on rhels7.2 , we have ntpdate installed.
```
[root@c910f03c17 ~]# which ntpdate
/usr/sbin/ntpdate
[...]

PR #1155, opened 2016-05-19
issue 1149: add ntpdate in the pkglist for ubuntu16.04

daniceexi:


daniceexi:
Fix issue for #1149 that has not ``ntpdate`` to be installed for ubuntu 16.04 diskful and diskless compute node.
``ntpdate`` was installed by default, now for ubuntu 16.04, we have to require it explicitly.

immarvin:
agree to merge

whowutwut:
Why do we still update the base compute.pkglist file? Which version of Ubuntu should use that?

daniceexi:
all the x86_64 for ubuntu will share this pkglist file

Issue #1159, opened 2016-05-19
[FVT]:some errors when execute syslog on Ubuntu16.04


tingtli:
xCAT 2.12 on Ubuntu16.04
```
 Thu May 19 10:53:39 EDT 2016 Running postscript: syslog
 2547 grep: /etc/rsyslog.d/remote.conf: No such file or directory
 2548 grep: /etc/rsyslog.d/remote.conf: No such file or directory
 2549 cp: cannot stat '/etc/rsyslog.d/remote.conf': No such file or directory
 2550 cp: cannot stat '/etc/rsyslog.d/remote.conf': No such file or directory
```

immarvin:
The rsyslog shipped in ubuntu 16.04 is:
````
root@c910f03c17k27:/xcatpost# rsyslogd -v
rsyslogd 8.16.0, compiled with:
        PLATFORM:                               powerpc64le-unknown-linux-gnu
        PLATFORM (lsb_release -d):
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              Yes
        FEATURE_DEBUG (debug build, slow code): No
        32bit Atomic operations supported:      Yes
        64bit Atomic operations supported:      Yes
        memory allocator:                       system default
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
        Number of Bits in RainerScript integers: 64

See http://www.rsyslog.com for more information.
````
The rsyslogd 8.16.0 will not install ``/etc/rsyslog.d/remote.conf`` on package installation and the configuration for UDP/TCP reception is commented out in rsyslog.conf. However, the syslog postscript suppose configuration for UDP/TCP reception is included in``/etc/rsyslog.d/remote.conf`` ,which is shipped by rsyslog package.

immarvin:
This is caused by the packaging change from rsyslog 8.16.
````
root@c910f03c17k27:/xcatpost# rsyslogd -v
rsyslogd 8.16.0, compiled with:
        PLATFORM:                               powerpc64le-unknown-linux-gnu
        PLATFORM (lsb_release -d):
        FEATURE_REGEXP:                         Yes
        GSSAPI Kerberos 5 support:              Yes
        FEATURE_DEBUG (debug build, slow code): No
        32bit Atomic operations supported:      Yes
        64bit Atomic operations supported:      Yes
        memory allocator:                       system default
        Runtime Instrumentation (slow code):    No
        uuid support:                           Yes
[...]

PR #1163, opened 2016-05-20
fix issue#1159@github,add support for udp/tcp syslog reception for rsyslog shipped in ubuntu16.04

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/1159

The rsyslogd 8.16.0 will not install /etc/rsyslog.d/remote.conf on package installation and the configuration for UDP/TCP reception is commented out in rsyslog.conf. However, the syslog postscript suppose configuration for UDP/TCP reception is included in/etc/rsyslog.d/remote.conf ,which is shipped by rsyslog package.


daniceexi:
This change has been reviewed by @neo954 and @daniceexi.

Issue #1162, opened 2016-05-20
[FVT]:network: unrecognized service


tingtli:
xCAT 2.12 Ubuntu16.04 on ppc64le
We found there is an error for network is not recognized service in xcat.log. Do we use network service in some script on Ubuntu16.04 , please check 
```
Generating grub configuration file ...
Found linux image: /boot/vmlinux-4.4.0-22-generic
Found initrd image: /boot/initrd.img-4.4.0-22-generic
done
network: unrecognized service   <===here
running /xcatpost/mypostscript.post
Thu May 19 22:57:18 EDT 2016 Running postscript: otherpkgs
./otherpkgs: no extra rpms to install
Thu May 19 22:57:18 EDT 2016 postscript otherpkgs return with 0
/xcatpost/mypostscript.post return
xcatdsklspost: downloaded postscripts successfully
Thu May 19 23:13:12 EDT 2016 Running postscript: ospkgs
```

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1164

PR #1164, opened 2016-05-20
fix issue#1162@github, correct a code logic error

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1162


daniceexi:
I have reviewed this change and agree to merge.

Issue #1166, opened 2016-05-20
[FVT]Failed to show the installing status after rnetboot CN even after 5 minutes.


caomengmeng:
Test ENV:
```
[root@c910f result]# rpm -qa | grep -i xCAT
xCAT-server-2.12-snap201605190710.noarch
xCAT-test-2.12-snap201605190711.noarch
syslinux-xcat-3.86-2.noarch
xCAT-genesis-base-x86_64-2.12-snap201605051534.noarch
ipmitool-xcat-1.8.15-1.x86_64
xCAT-client-2.12-snap201605190710.noarch
xCAT-buildkit-2.12-snap201605190711.noarch
xCAT-2.12-snap201605190711.x86_64
elilo-xcat-3.14-4.noarch
conserver-xcat-8.1.16-10.x86_64
grub2-xcat-2.02-0.16.el7.snap201506090204.noarch
perl-xCAT-2.12-snap201605190710.noarch
xCAT-genesis-scripts-x86_64-2.12-snap201605190711.noarch
[root@c910f result]# cat /etc/*release
Red Hat Enterprise Linux Server release 6.7 (Santiago)
Red Hat Enterprise Linux Server release 6.7 (Santiago)
```
Description: Install CN with autotest case, case begin at 1:07PM, 5 minutes later after rnetboot CN, check the CN status, it is booted, but the time is 00:20PM, this is the last booted time, that means 5 minutes later, the CN status is not changed to show this installation. Then the case failed at 1:19PM, rcons CN, it shows that the CN is still being installed.
```
------START:reg_linux_diskfull_installation_flat::Time:Fri May 20 01:07:20 2016------
RUN:lsdef -l c9 | grep status

[lsdef -l c910 | grep status] Running Time:0 sec
RETURN: rc = 0
OUTPUT:
    status=booted
    statustime=05-20-2016 00:20:17
------END::reg_linux_diskfull_installation_flat::Failed::Time:Fri May 20 01:15:19 2016 ::Duration::479 sec------
```
At 1:27 PM this OS is installed successfully.
```
[root@c910f0 result]# lsdef c910f04x20 | grep status
    status=booted
    statustime=05-20-2016 01:27:26
```

daniceexi:
We already had a story to track this issue that showing the ``power on`` status after running the ``rpower on``.

caomengmeng:
@daniceexi  This reproduced in ppc64le rh7.2 test environment when install diskful CN in today's autotest enviroment.
[...]

PR #1170, opened 2016-05-20
To avoid rnetboot cn status not changed issue

caomengmeng:
@tingtli 
1. To fix below issue to add "chdef CN status=" before start nodeset for always failed cases.
2. update wait time from 5 minutes to 10 minutes
https://github.com/xcat2/xcat-core/issues/1166
Please help review.

### DIFF
--- a/xCAT-server/share/xcat/tools/xCATreg
+++ b/xCAT-server/share/xcat/tools/xCATreg
@@ -297,6 +297,8 @@ sub mn_install {
 
         &runcmd("makedhcp -n");
         &runcmd("makedns -n");
+        &runcmd("chdef $confkeys{$mn} status=");
+        &runcmd("lsdef -l $confkeys{$mn}|grep status");
 
 	&runcmd("nodeset $confkeys{$mn} osimage=$osimage");
 	if( $arch =~ /x86_64/ || $arch =~ /ppc64le/ || $arch =~ /ppc64el/){
@@ -305,12 +307,15 @@ sub mn_install {
 	    &runcmd("rnetboot $confkeys{$mn}");
 	}
 	
-	if($mn !~ /ubuntu/){
-	    &runcmd("sleep 300");
-	    &runcmd("a=0;while ! `lsdef -l $confkeys{$mn}|grep status|grep booted >/dev/null`; do sleep 10;((a++));if [ \$a -gt 200 ];then break;fi done");}
-	else{
-	    &runcmd("sleep 1800");	
-	}
+         &runcmd("sleep 600");
+         &runcmd("a=0;while ! `lsdef -l $confkeys{$mn}|grep status|grep booted >/dev/null`; do sleep 30;((a++));if [ \$a -gt 100 ];then break;fi done");
+
+#	if($mn !~ /ubuntu/){
+#	    &runcmd("sleep 300");
+#	    &runcmd("a=0;while ! `lsdef -l $confkeys{$mn}|grep status|grep booted >/dev/null`; do sleep 10;((a++));if [ \$a -gt 200 ];then break;fi done");}
+#	else{
+#	    &runcmd("sleep 1800");	
+#	}
 	
 	&runcmd("lsdef -l $confkeys{$mn}|grep status|grep booted >/dev/null");
 	if($?){
--- a/xCAT-server/share/xcat/tools/xCATreg
+++ b/xCAT-server/share/xcat/tools/xCATreg
@@ -298,7 +298,11 @@ sub mn_install {
         &runcmd("makedhcp -n");
         &runcmd("makedns -n");
         &runcmd("chdef $confkeys{$mn} status=");
-        &runcmd("lsdef -l $confkeys{$mn}|grep status");
+        if($?){
+            send_msg(0, "[$mn->$confkeys{$mn}][mn_install] Failed to change MN status in current control node");
+            return 1;
+        }
+
 
 	&runcmd("nodeset $confkeys{$mn} osimage=$osimage");
 	if( $arch =~ /x86_64/ || $arch =~ /ppc64le/ || $arch =~ /ppc64el/){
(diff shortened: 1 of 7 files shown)


xCAT-server/share/xcat/tools/xCATreg
```diff
@@ -297,6 +297,8 @@ sub mn_install {
 
         &runcmd("makedhcp -n");
         &runcmd("makedns -n");
+        &runcmd("chdef $confkeys{$mn} status=");
+        &runcmd("lsdef -l $confkeys{$mn}|grep status");
```
tingtli: After you set status=blank, you will get nothing for "lsdef -l $confkeys{$mn}|grep status"
what is the goal for this sentence "&runcmd("lsdef -l $confkeys{$mn}|grep status");" ?

caomengmeng:
@tingtli Update the xCATreg file to check the chdef command works well or not.

whowutwut:
Recommend changing "Full_installation_flat_p8_le" to "Diskful_installation_flat_p8_le" to be consistent with the testcase file naming to match Diskless. 

Also, the **testcase_added** tag should not be used for the pull requests... it should be used to tag an existing issue that had previously been tagged with **testcase_requested** so we can indicate that FVT added a missing testcase into the testing suite.

xCAT-test/autotest/testcase/installation/Full_installation_flat_p8_le
```diff
@@ -20,7 +20,7 @@ cmd:nodeset $$CN osimage=__GETNODEATTR($$CN,os)__-__GETNODEATTR($$CN,arch)__-ins
 check:rc==0
 cmd:rpower $$CN boot
 check:rc==0
-cmd:sleep 300
+cmd:sleep 600
```
whowutwut: 👎  Do we need to use the sleep?   This value will continue to change as we move up in OS because it normally will take longer for the installation to finish. What is wrong with just running the command below on line 24.. it just sleep 1 minute and recheck the installation status? 

@daniceexi  I think we should also look into adding a new status when EVERYTHING is complete from an xCAT point of view so that we know postscripts have completed and the machine is "ready" to be used....

xCAT-test/autotest/testcase/installation/reg_linux_SN_installation_hierarchy
```diff
@@ -67,8 +67,8 @@ check:rc==0
 cmd:if [[ "__GETNODEATTR($$SN,arch)__" =~ "ppc64" ]]; then rnetboot $$SN;elif [[ "__GETNODEATTR($$SN,arch)__" =~ "x86_64" ]];then rpower $$SN boot; fi
 check:rc==0
 
-cmd:sleep 1800
-cmd:a=0;while ! `lsdef -l $$SN|grep status|grep booted >/dev/null`; do sleep 10;((a++));if [ $a -gt 400 ];then break;fi done
+cmd:sleep 600
+cmd:a=0;while ! `lsdef -l $$SN|grep status|grep booted >/dev/null`; do sleep 60;((a++));if [ $a -gt 70 ];then break;fi done
```
whowutwut: This one waits 70 iterations, but the changes below in the other test case files wait 50 iterations.   I think this should all be sync'ed up.  Can we have this be a function since it's being called so many times... 

Looks like this is waiting 10 minutes above on the sleep, then 70 minutes.. so 80 min total... 
The changes in the other files below wait 60 minutes..  We could have a worse case because since this is looping, if the OS completes before that time, it will continue.

whowutwut: After we hit the time out, we break out of the while loop.. shouldn't we exit ERROR in that case?

xCAT-test/autotest/testcase/installation/Full_installation_flat_p8_le
caomengmeng: @whowutwut  I don't think we need to do so many times recheck one minutes after another at the beginning, as experience, the shortest time for OS to installed successfully maybe more than 7 minutes,  so used the sleep 5 minutes at before, why sleep 10 minutes is because this issue: https://github.com/xcat2/xcat-core/issues/1166, which is a known issue but not fixed till now, and it randomly impacts all the os diskful installation in every platform.

xCAT-test/autotest/testcase/installation/reg_linux_SN_installation_hierarchy
caomengmeng: After this while loop, there is check and will exit with error.
Automation test cases is limited by framework, so can't do this directly in this while loop.
And for the whole time, if there is a sub function that will be great, this is also the framework limited so far as I know, I will discuss with @hu-weihua  for this to see if we can do it now easily. If the effort is huge to complete this, as these old test environment will move to Jenkins later, we will do the sub function issues later.
For the total time, do you thinks it impacts? I don't think so, if the installation works well, it will jump out quickly. but if it is failed to install, 60, 70 or 80 minutes, it doesn't matter. As all the cases is existed before, so I just improve them a little. If we need to unify them, there should be a story to track it, but think about the effort and result, I don't think it is necessary. And this can also be fixed if sub-function can be used easily in existed automaton framework.

xCAT-test/autotest/testcase/installation/Full_installation_flat_p8_le
caomengmeng: @whowutwut  After rneboot, or rpower reset, how long the CN status will be changed and changed to which? If xCAT defines this clearly, in our automation test case, to track the status is easy for code things. But it is the xCAT production issue for this :-1:  question, not automation code now.

tingtli:
@caomengmeng @whowutwut  We have realized usring script or sub_function in the test case section  "cmd:" just short time ago.  So I think we can think about refining all the diskful/diskless provision test cases to check the status after the provision.  The problem is what is the reasonable time to wait before it becomes from netbootiing to booted?  In the sub_function to check the status,we have different to process, e.g. for a baremetal machine to install rhels, we probably need 30-40 minutes to wait, for a virtual machine, we probably need 10 minutes to wait. This now all based on our test results.  so, in the sub-function, for different machine type, and for different os, for different provision type, we need to define a  "the most long" time out. If time our, we think the provision fails.

Issue #1114, opened 2016-05-14
[DEV] Switch-based discovery finished, but bmc node definition didn't get to remove from xcat db


cxhong:
Working on switch-based discovery,  BMC finished discovery, we can see from compute node, all the attributes included Mac address are set.  but bmc node definition still there.
````
# nodels
c910f05c31
node-8335-gtb-100469a
switch-10-5-23-1

# lsdef c910f05c31
Object name: c910f05c31
    arch=ppc64
    bmc=50.5.31.1
    cons=ipmi
    cpucount=128
    cputype=POWER8NVL (raw), altivec supported
    disksize=sda:1000GB,sdb:1000GB
    groups=all,compute
    installnic=mac
    ip=192.168.5.31
    mac=70:e2:84:14:0a:29
    memory=261506MB
    mgt=ipmi
    mtm=8335-GTB
    netboot=petitboot
    nodetype=mp
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    serial=100469A
    status=standingby
    statustime=05-13-2016 21:53:15
    supportedarchs=ppc64
    switch=switch-10-5-23-1
    switchport=2


# tabdump vpd
#node,serial,mtm,side,asset,uuid,comments,disable
"node-8335-gtb-100469a","100469A","8335-GTB",,,,,
"c910f05c31","100469A","8335-GTB",,,"911c8d68-6ddf-6944-96a3-e090bd0eb87f",,
````

but the results are not consistent, I had some successfully runs.  I think there are race condition in this piece of code in the switch.pm.

````
 if ($req->{'mtm'}->[0] and $req->{'serial'}->[0]) {
            my $mtms = $req->{'mtm'}->[0]."*".$req->{'serial'}->[0];
            **my $tmp_nodes = $::XCATVPDHASH{$mtms};**
            foreach (@$tmp_nodes) {
                if ($::XCATMPHASH{$_}) {
                    $pbmc_node = $_;
                }
            }
        }
```````

for the failed case, XCATVPDHASH will return back with compute node
for the successfully case, XCATVPDHASH will return back with bmc node.

zet809:
@cxhong The fix had been merged, will you pls close this defect?

PR #1186, opened 2016-05-24
Temp BMC node definition hadn't been removed for switch-based discovery

cxhong:
for issue #1114 .

ErTao's note for this issue:
for the issue that temp BMC node can not be found, a possible reason seems as:
```
sub do_discovery_process {
    $SIG{TERM} = 'DEFAULT';
    $SIG{INT} = 'DEFAULT';
    my %args =@_;
    my $broker = $args{broker};
    my $quit=0;
    my $vintage = time();
    $dispatch_requests=0;
    populate_site_hash();
    populate_vpd_hash();
    populate_mp_hash();
    while (not $quit) {
        if ((time()-$vintage)> 15) {
            populate_site_hash();
            populate_vpd_hash();
            populate_mp_hash();
            $vintage = time();
        } # site table reread every 15 second
        my $msg = fd_retrieve($broker);

```
As you can see the sub-routine do_discovery_process, it will read vpd and mp table before going into the while loop, but if there is no findme request got, the loop will hung at "fd_retrieve", which mean if there is any modification for vpd and mp after xcatd start, they can not be read since it have no change to check whether it is time to refresh vpd and mp hash again.
 
Then, if there is a findme come in, it may won't be able to match any node from vpd and mp hash.

### DIFF
--- a/xCAT-server/sbin/xcatd
+++ b/xCAT-server/sbin/xcatd
@@ -585,13 +585,13 @@ sub do_discovery_process {
     populate_vpd_hash();
     populate_mp_hash();
     while (not $quit) {
+        my $msg = fd_retrieve($broker);
         if ((time()-$vintage)> 15) {
             populate_site_hash();
             populate_vpd_hash();
             populate_mp_hash();
             $vintage = time();
         } # site table reread every 15 second
-        my $msg = fd_retrieve($broker);
         my $data;
         my $client;
         my $clientn;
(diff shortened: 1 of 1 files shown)


cxhong:
I modified the code as ErTao's suggestion.  I didn't hit this issue any more after changes

zet809:
Looks good to me, merge it. 
@cxhong Will you pls verify and close defect #1114 ?

PR #1193, opened 2016-05-24
Update centos7 template to mimic rhels7

arif-ali:


daniceexi:
I have reviewed the change. Thanks @arif-ali

PR #1194, opened 2016-05-24
Modify xCATreg about issue 1068 

junxiawang:
@ting ting please help review

### DIFF
--- a/xCAT-server/share/xcat/tools/xCATreg
+++ b/xCAT-server/share/xcat/tools/xCATreg
@@ -294,6 +294,12 @@ sub mn_install {
 	}
 	
 	send_msg(2, "[$mn->$confkeys{$mn}][mn_install] installing $osimage for node $confkeys{$mn}................");
+        
+        if( $arch =~ /x86_64/ || $arch =~ /ppc64le/ || $arch =~ /ppc64el/){
+        &runcmd("rmvm $confkeys{$mn} -f -p ");
+        &runcmd("chdef $confkeys{$mn} mac= ");
+        &runcmd("mkvm $confkeys{$mn} -s 40G");
+        }
 
         &runcmd("makedhcp -n");
         &runcmd("makedns -n");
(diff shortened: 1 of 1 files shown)


xCAT-server/share/xcat/tools/xCATreg
```diff
@@ -870,13 +881,15 @@ sub do_test {
                 $rc=system("ssh -t $confkeys{$mn} 'exec bash -l -i -c \"xcattest -f /opt/xcat/share/xcat/tools/autotest/default.conf  -t $diskfull_installation_flat_testcase\"' >/dev/null 2>&1");
                 send_msg(2, "[$mn->$confkeys{$mn}][do_test] do diskfull installation test in flat mode in $mn.....[done]");
                 last if($rc != 0);
-
+                
+                if($mn =~ /ubuntux/){
                 #set up dockerhosts
```
whowutwut: Since you added this test, you should indent the code block following 4 spaces so that it's easier to read.

xCAT-server/share/xcat/tools/xCATreg
```diff
@@ -294,6 +294,12 @@ sub mn_install {
 	}
 	
 	send_msg(2, "[$mn->$confkeys{$mn}][mn_install] installing $osimage for node $confkeys{$mn}................");
+        
+        if( $arch =~ /x86_64/ || $arch =~ /ppc64le/ || $arch =~ /ppc64el/){
+        &runcmd("rmvm $confkeys{$mn} -f -p ");
```
whowutwut: This code is not formatted correctly, should be indented 4 space

xCAT-server/share/xcat/tools/xCATreg
```diff
@@ -294,6 +294,12 @@ sub mn_install {
 	}
 	
 	send_msg(2, "[$mn->$confkeys{$mn}][mn_install] installing $osimage for node $confkeys{$mn}................");
+        
+        if( $arch =~ /x86_64/ || $arch =~ /ppc64le/ || $arch =~ /ppc64el/){
```
whowutwut: Why do we have this test here?  why only do this for x86_64 and ppc64le? why not include ppc64 be?

whowutwut:
@junxiawang  can you please add a description to the issue describing the code changes. That will help in the code review.  Also please set the label and milestone and assign to @tingtli . 

It's not making sense why the code changes are made "According to the GA build?"  Doesn't seem like anything changed to be GA version specific.

xCAT-server/share/xcat/tools/xCATreg
junxiawang: @whowutwut  only virtual machine need mkvm command

junxiawang:
@whowutwut  xCATreg code modified before GA about xcat version check and virtual machine set up

tingtli:
@junxiawang  please modify the commit information

tingtli:
@junxiawang Would you please modify the description for this commit to say why we modify this.
I don't think this is because we sync the files. We may fix some bugs or add some support in this commit.

junxiawang:
@tingtli 
1.for rmvm bug
2.for docker support
3.for ubuntu's bash issue

PR #1195, opened 2016-05-24
Update migration test cases

caomengmeng:
@tingtli please help review code:
1. update config file to change the latest_version to be 2.12.1 as last night meeting
2. update redhat test cases to fix the rh6.7 test environment issue: already passed with rh6.7 test and verified in autotest environment for rh7.2 test.
Thanks.

### DIFF
--- a/xCAT-test/autotest/testcase/migration/redhat_migration
+++ b/xCAT-test/autotest/testcase/migration/redhat_migration
@@ -63,7 +63,7 @@ cmd:xdsh $$CN "tar -jxvf /oldxcat/xcat-core* -C /oldxcat"
 check:rc==0
 cmd:xdsh $$CN "/oldxcat/xcat-core/mklocalrepo.sh"
 check:rc==0
-cmd:xdsh $$CN "cd /oldxcat/xcat-dep/rh`more /etc/*release|grep VERSION_ID|cut -c13`/__GETNODEATTR($$CN,arch)__/;./mklocalrepo.sh"
+cmd:xdsh $$CN "cd /oldxcat/xcat-dep/rh`echo __GETNODEATTR($$CN,os)__ | cut -c6`/__GETNODEATTR($$CN,arch)__/;./mklocalrepo.sh"
 check:rc==0
 cmd:xdsh $$CN "yum clean all;yum -y install xCAT"
 check:rc==0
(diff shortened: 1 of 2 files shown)


xCAT-test/autotest/testcase/migration/redhat_migration
```diff
@@ -63,7 +63,7 @@ cmd:xdsh $$CN "tar -jxvf /oldxcat/xcat-core* -C /oldxcat"
 check:rc==0
 cmd:xdsh $$CN "/oldxcat/xcat-core/mklocalrepo.sh"
 check:rc==0
-cmd:xdsh $$CN "cd /oldxcat/xcat-dep/rh`more /etc/*release|grep VERSION_ID|cut -c13`/__GETNODEATTR($$CN,arch)__/;./mklocalrepo.sh"
+cmd:xdsh $$CN "cd /oldxcat/xcat-dep/rh`echo __GETNODEATTR($$CN,os)__ | cut -c6`/__GETNODEATTR($$CN,arch)__/;./mklocalrepo.sh"
```
whowutwut: 👍

whowutwut:
@caomengmeng , please set the label and milestone ! :)  

* I don't understand how the code changes fix issue in the rhel6.7 test environment?  Using the `echo __GETNODEATTR($$CN,os)__ ` instead of looking at the `/etc/*release` file? 

* Is it possible to have the directories be variables in the test case? 

```
 cmd:xdsh $$CN "rm -rf /oldxcat"
 cmd:xdsh $$CN "mkdir -p /oldxcat"
 check:rc==0
 cmd:xdsh $$CN "yum -y install bzip2"
 check:rc==0
 cmd:xdsh $$CN "cd /oldxcat;wget $$MIGRATION1_DEP"
 check:rc==0
 cmd:xdsh $$CN "cd /oldxcat;wget $$MIGRATION1_CORE"
 check:rc==0
 cmd:xdsh $$CN "tar -jxvf /oldxcat/xcat-dep* -C /oldxcat"
 check:rc==0
 cmd:xdsh $$CN "tar -jxvf /oldxcat/xcat-core* -C /oldxcat"
```
Can `/oldxcat` be a variable.. or does it need to be spelled out each time?

* Can we add some description comments to explain what the difference is between migration case 1 and migration case 2? 

Seems like we are not doing anything to check during the migration that the site table variables are not changed... is this something we can add?    @tingtli Is there another issue open to have that?

tingtli:
@whowutwut  The reason why we don't use more /etc/*release is because rhels7.* and rhels6.* has different for /etc/*release*
/oldxcat is a temporary directory, if it is a variable, we will have so many variables. I guess that is why mengmeng dead code .
For the site value.  we have a case cover xcatsslversion value updates. But we don't have values check for other site attributes. I think we can add that in this case.

tingtli:
@whowutwut I think the site value could be refinement for the case. I will merge this case first for the daily run succeed. Then ask mengmeng to update that. 
@caomengmeng  I add a task and assign to you for the site value check when we do migration.

PR #1196, opened 2016-05-24
fix #954, ubuntu, diskfull, enable the pre-script log and post-script log can be forwarded to the MN

### LINKED ISSUE #954: [FVT]Failed to find expected logs while site.xcatdebugmode=1 for ubuntu on ppc64le

penguhyang:
fix #954 
For ubuntu diskfull installation, enable the pre-script log and post-script log can be forwarded to the MN.


whowutwut:
In the description, you mention pre and post script log can be forwarded to the mgmt node, but the files affected is only pre.*, is something missing? 

Also some of the code change is for adding `xcatdebugmode` code processing, that is missed in the description 

Shouldn't we have these file specify the ubuntu version? ubuntu16, ubuntu14? what OS/ARCH is are using the pre.ubuntu file?

xCAT-server/share/xcat/install/scripts/pre.ubuntu
```diff
@@ -11,6 +11,24 @@ fi
 
 debconf-get open-iscsi/targets > /tmp/q
 
+cat >/tmp/syslogd.sh <<EOF
```
whowutwut: Suggest that we specify a filename with xcat in it... `/tmp/xcat.syslogd.sh`

xCAT-server/share/xcat/install/scripts/pre.ubuntu.ppc64
```diff
@@ -6,6 +11,24 @@ fi
 
 debconf-get open-iscsi/targets > /tmp/q
 
+cat >/tmp/syslogd.sh <<EOF
```
whowutwut: Suggest that we specify a filename with xcat in it.., something like: `/tmp/xcat.syslogd.sh`

penguhyang:
@whowutwut Thanks for review.
I create a new process to connect master using busybox syslogd when running pre-script file. So the pre script(run in installer) and post script(run in chroot /target) log can be forwarded to the mgmt node. The config file created in /tmp... it's only used in installer time.

xCAT-server/share/xcat/install/scripts/pre.ubuntu
penguhyang: fixed.

xCAT-server/share/xcat/install/scripts/pre.ubuntu.ppc64
penguhyang: fixed.

immarvin:
inside the installer, there is already a syslogd running, which write all the logs into /var/log/syslog:
````
/sbin/syslogd -m 0 -O /var/log/syslog -S
````
we should 1st kill the existing syslogd process  first and then start the syslogd with remote log forwarding.

the following lines can be added into the beginning of "/xCAT-server/share/xcat/install/scripts/pre.ubuntu" and "/xCAT-server/share/xcat/install/scripts/pre.ubuntu.ppc64"

````
if [ "$XCATDEBUGMODE" -gt "0" ]; then
    killall syslogd
    syslogd -R #XCATVAR:XCATMASTER#
fi
````
btw, the documentation on busybox is https://busybox.net/downloads/BusyBox.html

penguhyang:
@immarvin Thanks for review.
Code updated.
1. 1st kill the existing syslogd process
2. use syslogd without create a new file.

immarvin:
agree to merge

PR #1207, opened 2016-05-25
Add stop label for makedns case as it failed for serveral time 

caomengmeng:
@tingtli 
Add stop label for makedns case as it failed for serveral time but can't reproduced while run by manual.
Thanks.


daniceexi:
It's ok to add this ``stop`` label to help to debug the makedns issue.

PR #1208, opened 2016-05-25
Add -1 value for $dbworkerpid to make the db access mode more clear.

chenglch:
$dbworkerpid = -1 db process has not been started,
               access db in direct access mode.
$dbworkerpid == 0 db process itself.
$dbworkerpid >0 db process is started, access db in cache mode.


daniceexi:

PR #1210, opened 2016-05-25
modify template for docker host setup

junxiawang:
@tingtli  modify template for docker kvm host setup


tingtli:
@junxiawang please set lables, milestone and assignee
Please double check the template, make sure we will not hit the "attributes missing" error we use the default template

PR #1215, opened 2016-05-26
build xcat-core script for pcm

fenggaobj:
Add the code for building xcat-core for PCM. 
Include DEB and RPM.

### DIFF
--- a/build-ubunturepo
+++ b/build-ubunturepo
@@ -35,7 +35,6 @@
 ############################
 
 #Begin pcm
-PCM_USER=root
 PCM_SERVER=9.21.55.37
 PCM_UPLOAD_PATH="/xcat_pcm_release/DEB"
 
@@ -435,8 +434,8 @@ __EOF__
         tar_new_name=`date +%Y%m%d%H%M`."PCM".$tar_name
         cp $tar_name $tar_new_name
         i=0
-        echo "Uploading $tar_new_name to $PCM_USER@$PCM_SERVER:$PCM_UPLOAD_PATH"
-        while [ $((i+=1)) -le 5 ] && ! rsync -v --force $tar_new_name $PCM_USER@$PCM_SERVER:$PCM_UPLOAD_PATH
+        echo "Uploading $tar_new_name to $UPLOADUSER@$SERVER:$RELEASE/$PCM_UPLOAD_PATH"
+        while [ $((i+=1)) -le 5 ] && ! rsync -v --force $tar_new_name $UPLOADUSER@$SERVER:$RELEASE/$PCM_UPLOAD_PATH
         do : ; done
         exit 0;
     fi
--- a/build-ubunturepo
+++ b/build-ubunturepo
@@ -20,6 +20,7 @@
 #                           being uploaded to http://xcat.org/files/xcat/repos/apt
 #                           (but the tarball file name will be like a released tarball, not a snap build).  
 #                           When you are ready to release this build, use PROMOTE=1 without PREGA
+#               EMBED=<embedded-environment> - the environment for which a minimal version of xcat should be built, e.g. pcm
 #               BUILDALL=1 - build all rpms, whether they changed or not.  Should be used for snap builds that are in 
 #                            prep for a release.
 #               UP=0 or UP=1 - override the default upload behavior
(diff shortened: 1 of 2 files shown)


build-ubunturepo
```diff
@@ -32,6 +33,53 @@
 #                             3. the built xcat-dep deb packages tarball can be found in "../../xcat-dep" 
 #                                related to the path of this script
 ############################
+
+#Begin pcm
+PCM_USER=root
```
whowutwut: I also made this comment to @tingtli  when I was looking at her changes when she was supporting the PCM build. 
Since we now allow users to clone the xCAT project and build from source on their local machine, I don't think we should put any information that would potentially change into the build script.   

So the PCM_USER, PCM_SERVER, PCM_UPLOAD_PATH  (upload path might be OK) 

The reason is now if you change any of those values, it results in another commit being made to the build script. 

Rather we should do error checking and make sure it's exported in the environment on the build machine. For example...
```
[vhu@victorvm7 tmp]$ cat test.sh 
#!/bin/sh

if [[ ${EMBED} == "pcm" ]]; then
    # pre-verification checking for pcm build
    if [[ -z $PCM_USER ]]; then 
       echo "PCM_USER must be specified in an environment variable, cannot continue"
       exit 1
    fi

    echo "PCM_USER=$PCM_USER"
fi 
```

You can even go one step further and only  check this when we are uploading (UP=1)

build-ubunturepo
```diff
@@ -32,6 +33,53 @@
...
+PCM_SERVER=9.21.55.37
+PCM_UPLOAD_PATH="/xcat_pcm_release/DEB"
+
+#For pcm replace and recover spec file
+new_spec_path="/CONTROL/"
```
whowutwut: I don't agree with these changes to replace and recover the files...  The files is in `/CONTOL` on the build server? 
What happens if our build server dies? The files that you want to change to should be checked into the code repo and maintained there.

build-ubunturepo
```diff
@@ -32,6 +33,52 @@
...
+new_control_xcat_server="xCAT-server-control"
+new_constrol_xcat_sn="xCATsn-control"
+old_control_name="control"
+old_control_xcat="xCAT/debian"
+old_control_xcat_server="xCAT-server/debian"
+old_constrol_xcat_sn="xCATsn/debian"
+
+replace_pcm_spec()
+{
+   echo "PCM begin to replace spec file[`pwd`]"
```
whowutwut: I don't think this is good design.  If you add the new control files for PCM into a subdirectory under `debian`... for example, if we look at `new_control_xcat_server="xCAT-server-control`

This file is intended to replace: `xCAT-server/debian/control`
If you create this new file under a "pcm" sub directory.. `xCAT-server/debian/pcm/control` then you can have a script look for all control files under xcat-core that looks like: `debian/pcm/control*` and copy it to the directory above.  

Then you do not need to have those hard coded paths

whowutwut: A better option is to have your pcm branch maintain all these changes, and we would just build your 2.11-pcm or 2.12-pcm branch..  The downside is that the master build will never have these changes... so maybe that's not a good option :)

build-ubunturepo
```diff
@@ -364,13 +414,32 @@ __EOF__
         ln -s xcat-core core-snap
     fi
 
+    #For pcm
+    if [ "$EMBED" = "pcm" ]; then
```
whowutwut: If you take the comment above to commit the control files specific to PCM and you copy them over in a generic loop, then this step is not necessary.  You can use git to checkout the files from the repo, which will undo any of the copys that you will make.

buildcore.sh
```diff
@@ -41,6 +41,10 @@ SERVER=xcat.org
 FILES_PATH="files"
 FRS=/var/www/${SERVER}/${FILES_PATH}
 RELEASE=github.com/xcat2/xcat-core/releases
+#for pcm
+PCM_USER=root
```
whowutwut: Same comment as with the build-ubunturepo script, I don't like these 3 variable to be defined inside the script.

whowutwut:
@fenggaobj  I was looking at cleaning up some of my branches, and I came across this branch that I worked on back in Feb of this year. 

https://github.com/whowutwut/xcat-core/commits/pcm_build_master

I think I did this after I talked to @tingtli  about the changes to pcm build script.  Seems like there is a lot of duplicate work going on... You can take a look at the commits for examples of what I think you should be doing...

https://github.com/whowutwut/xcat-core/commit/a87448f8b423f477d9ae65cbd979916a834740f8
https://github.com/whowutwut/xcat-core/commit/c70c9813158cc30ebfec81bd6948b9b5f5561e28

whowutwut:
@fenggaobj  I don't want to leave this pull request open,  can you make the changes to your branch and then re-open the pull request when it's ready for review.

PR #1216, opened 2016-05-26
The code of xcat probe component

hu-weihua:
These 3 files are the first batch files of xcat probe component.
xcatprobe : the main script of probe framework
xcatmn     : the first checking tool
probe_utils.pm : include common functions which can be share between checking tools developed by perl

Below is the structure of xcat probe component after xcat installation

    /opt/xcat/bin/xcatprobe
    /opt/xcat/probe/subcmds/xcatmn
    /opt/xcat/probe/lib/perl/ProbeUtils.pm

Below is the structure of xcat probe code saved in github

    ~/xCAT-probe/xcatprobe
    ~/xCAT-probe/subcmds/xcatmn
    ~/xCAT-probe/lib/perl/ProbeUtils.pm

For  detailed design please refer to https://github.com/xcat2/xcat-core/wiki/xcatprobe_framework
Any comments about code logic or code save structure are welcome.


xCAT-probe/lib/perl/ProbeUtils.pm
```diff
@@ -0,0 +1,317 @@
...
+        print LOGFILE "$flag $msg\n";
+        close LOGFILE;
+    }
+    return 0;
+}
+
+#------------------------------------------
+=head3
+    Description:
+        Judge if a string is a IP address
```
daniceexi: Please replace the Judge with ``test``

xCAT-probe/lib/perl/ProbeUtils.pm
```diff
@@ -0,0 +1,317 @@
+package ProbeUtils;
```
daniceexi: Please add the licence head at beginning of the file

~~~~
# IBM(c) 2016 EPL license http://www.eclipse.org/legal/epl-v10.html
~~~~

xCAT-probe/lib/perl/ProbeUtils.pm
```diff
@@ -0,0 +1,317 @@
...
+    Description:
+        Judge if a string is a IP address 
+    Attribute:  list by input sequence
+        addr: the string want to be judged 
+    Return value:
+        1 : yes
+        0 : no
+=cut
+#------------------------------------------
+sub isIpAddr{
```
daniceexi: Please change the naming of subroutines and variables to be ``str_str_str``. 
Please follow the coding guide. http://xcat-docs.readthedocs.io/en/stable/developers/guides/code/code_standard.html

xCAT-probe/lib/perl/ProbeUtils.pm
```diff
@@ -0,0 +1,317 @@
...
+        Judge if a IP address is a static IP address 
+    Attribute:  list by input sequence
+        ip:   a ip address
+        nic:  the network adapter which ip belongs to
+    Return value:
+        1 : yes
+        0 : no
+=cut
+#------------------------------------------
+sub isStaticIp{
```
daniceexi: I'd like the subcommand ``xcatmn`` require the 'MN IP' as a must have argument. And your code should test the IP is a valid IP for this xCAT MN (through ``ip addr`` command)

xCAT-probe/lib/perl/ProbeUtils.pm
```diff
@@ -0,0 +1,317 @@
...
+    Return value:
+        1 : yes
+        0 : no
+=cut
+#------------------------------------------
+sub isHttpReady{
+     my $mnip=shift;
+     $mnip=shift if(($mnip) && ($mnip =~ /ProbeUtils/));
+      
+     my $http = "http://$mnip/install/postscripts/efibootmgr";
```
daniceexi: I'd like you could try to download the file ``syslog``. The efibootmgr might be removed someday later.

xCAT-probe/lib/perl/ProbeUtils.pm
```diff
@@ -0,0 +1,317 @@
...
+#------------------------------------------
+sub isTftpReady{
+     my $mnip=shift;
+     $mnip=shift if(($mnip) && ($mnip =~ /ProbeUtils/));
+  
+     rename("/tftpboot/tftptestt.tmp", "/tftpboot/tftptestt.tmp.old") if(-e "/tftpboot/tftptestt.tmp");
+     rename("./tftptestt.tmp", "./tftptestt.tmp.old") if(-e "./tftptestt.tmp");
+
+     system("touch /tftpboot/tftptestt.tmp");
+     my $output = `tftp -4 -v $mnip  -c get tftptestt.tmp`;
```
daniceexi: ``tftp`` command is not installed by default for rh6/rh7. So you need to check whether the ``tftp`` has been installed first. Otherwise, you need to check whether the tftp process is running. 

You also can give a warning and require user to install tftp and run xcatprobe again.

xCAT-probe/subcmds/xcatmn
```diff
@@ -0,0 +1,475 @@
...
+     }
+     exit 0;
+}
+
+if($test){
+   ProbeUtils->send_msg("$output",3,"After xcat installation, use this command to check if xcat has been installed correctly and is ready for use. Before using this command        , please install tftp, nslookup and wget commands ahead. The platform supported are redhat, sles and ubuntu.");
+   exit 0;
+}
+
+my $hostname=`hostname -s`;
```
daniceexi: You have an assumption that the hostname of the node is the one which maps to the ip of MN which passed in from parameter, but this is not always true. We may add a parameter for user to pass in the hostname of the MN IP.

xCAT-probe/subcmds/xcatmn
```diff
@@ -0,0 +1,475 @@
...
+
+Description:
+    After xcat installation, use this command to check if xcat has been installed correctly and is ready for use. 
+
+Options:
+    -h : get usage information of $proname
+    -t : to verify if $proname can work, reserve option for probe framework
+    -m : specify current MN's IP address, if not specified, use master ip in site table
+    -f : specify where should $proname pass its output to. if not specified, pass output to STDOUT by default.
+    -v : output more information for debug
```
daniceexi: In xCAT **low-case** ``-v`` means version. The Verbose should use capital ``-V``.

xCAT-probe/lib/perl/ProbeUtils.pm
```diff
@@ -0,0 +1,317 @@
...
+=cut
+#----------------------------------------
+sub send_msg {
+    my $output=shift;
+    $output=shift if(($output) && ($output =~ /ProbeUtils/));
+    my $num = shift;
+    my $msg = shift;
+    my $flag;
+
+    if ($num == 0) {
```
daniceexi: could we use the ``d``, ``w``, ``f``, ``o`` to replace 1,2,3,4?

xCAT-probe/subcmds/xcatmn
```diff
@@ -0,0 +1,475 @@
...
+    $rst=1; 
+}else{
+    ProbeUtils->send_msg("$output",3,"$msg");
+}
+
+my $xcatiport=`tabdump site 2>&1| awk -F',' '/xcatiport/ { gsub(/"/, "", \$2) ; print \$2 }'`;
+chomp($xcatiport);
+ProbeUtils->send_msg("$output",0,"The port used by xcatd to receive install status updates from nodes is $xcatiport") if($verbose);
+$msg="xcatd is listening on port $xcatiport";
+$cmdoutput = `netstat -ant|grep LISTEN|grep $xcatiport`;
```
daniceexi: we may use 'netstat -antp | grep -i xcatd' to display the program name info too so that we can make the port is opened by xCAT.

xCAT-probe/subcmds/xcatmn
```diff
@@ -0,0 +1,475 @@
...
+    if($size<$expected){
+        ProbeUtils->send_msg("$output",1,"The free space of /tmp is less than $expected G");
+    }else{
+        ProbeUtils->send_msg("$output",3,"$msg");
+    }
+}
+
+
+$expected=10;
+$msg="The free space of /install directory is more than $expected G";
```
daniceexi: Generally, the path of install and tftpboot should be gotten from site table.

xCAT-probe/subcmds/xcatmn
```diff
@@ -0,0 +1,475 @@
...
+    if($?){
+         returncmdoutput($tmp) if($verbose);
+         ProbeUtils->send_msg("$output",0,"Simulate a node by chdef failed") if($verbose);
+         $rc=1;
+         last;
+    }else{
+         ProbeUtils->send_msg("$output",0,"Simulate a node xcatmntest<ip=$mnip mac=aa:aa:aa:aa:aa:aa> to do dhcp test") if($verbose);
+    } 
+ 
+    $tmp=`makehosts xcatmntest`;
```
daniceexi: do you want to check the test node has been added into /etc/hosts?

xCAT-probe/xcatprobe
```diff
@@ -0,0 +1,258 @@
...
+use Data::Dumper;
+use File::Path;
+use POSIX qw(WNOHANG setsid :errno_h);
+use Term::ANSIColor qw(:constants);
+$Term::ANSIColor::AUTORESET=1;
+
+my $pro_name=basename($0);
+my $pro_dir=dirname($0);
+#my $plugin_dir="$pro_dir/subcmds";
+my $plugin_dir="/opt/xcat/probe/subcmds";
```
daniceexi: You need to use following code to check the location of xcat

~~~~
$::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : '/opt/xcat';
~~~~

xCAT-probe/xcatprobe
```diff
@@ -0,0 +1,258 @@
...
+        $output = `$_ -t 2>&1`;
+        chomp($output);
+     
+        print "\n-->$_\n[OUTPUT]:\n$output\n" if($verbose);
+        if($output !~ /\[(\w+)\]\s*:\s*(.+)/){
+            print "skip $_ for doing '$_ -t' failed, bad format\n" if($verbose);
+            next;
+        }else{
+            my $desc =$2;
+            unless(@validflag ~~ /$1/){
```
daniceexi: ``~~``?

xCAT-probe/xcatprobe
```diff
@@ -0,0 +1,258 @@
...
+        my $cmdname = basename("$_");
+        $output = `$_ -t 2>&1`;
+        chomp($output);
+     
+        print "\n-->$_\n[OUTPUT]:\n$output\n" if($verbose);
+        if($output !~ /\[(\w+)\]\s*:\s*(.+)/){
+            print "skip $_ for doing '$_ -t' failed, bad format\n" if($verbose);
+            next;
+        }else{
+            my $desc =$2;
```
daniceexi: Do we need to check the ``$1`` is ``ok``?

xCAT-probe/xcatprobe
```diff
@@ -0,0 +1,258 @@
...
+            }else{
+                print BOLD BLUE  "[WARN] ";
+            }
+        }elsif($flag =~ /ok/i){
+            if($nocolor){
+                print "[ OK ] ";
+            }else{
+                print BOLD GREEN "[ OK ] ";
+            }
+        }elsif($flag =~ /debug/i){
```
daniceexi: I am not sure how it looks like if we add ``[debug]`` as leading word for debug message?

xCAT-probe/xcatprobe
```diff
@@ -0,0 +1,258 @@
...
+}
+
+#######################################
+# main
+#######################################
+my @tmpargv=@ARGV;
+my @supportopt=("-v","-h","-l","-n");
+my $pluginname;
+my $optnum=0; 
+foreach my $attr (@tmpargv){
```
daniceexi: I think you should only handle the first two arguments:

* if the first arg is "-v","-h","-l","-n", the second arg is **subcmd**; otherwise the first arg is **subcmd**

or a string

xCAT-probe/subcmds/xcatmn
whowutwut: Is it difficult to add support for long option names?  I've now become a fan of the long options because the single options seem to limit what we can do and sometimes the letter doesn't match.  

No need to change now, but some idea for future support. Then i would say  use `--debug`` or ``--trace``

xCAT-probe/subcmds/xcatmn
```diff
@@ -44,432 +45,507 @@ sub returncmdoutput{
...
+        print "$::USAGE";
+    }
+    exit 0;
 }
 
-if($test){
-   ProbeUtils->send_msg("$output",3,"After xcat installation, use this command to check if xcat has been installed correctly and is ready for use. Before using this command        , please install tftp, nslookup and wget commands ahead. The platform supported are redhat, sles and ubuntu.");
-   exit 0;
+if ($test) {
+    probe_utils->send_msg("$output", "o", "After xcat installation, use this command to check if xcat has been installed correctly and is ready for use. Before using this command        , please install tftp, nslookup and wget commands ahead. The platform supported are redhat, sles and ubuntu.");
```
daniceexi: Is the multiple spaces after ``command        `` is correct?

xCAT-probe/subcmds/xcatmn
```diff
@@ -44,432 +45,507 @@ sub returncmdoutput{
...
+    probe_utils->send_msg("$output", "f", "Option -n is required");
+    probe_utils->send_msg("$output", "d", "$::USAGE");
+    exit 1;
+}
 
-ProbeUtils->send_msg("$output",0,"Short hostname of this server is $hostname") if($verbose);
-ProbeUtils->send_msg("$output",0,"The address of hostname $hostname is $hostipaddr")  if($verbose);
-ProbeUtils->send_msg("$output",0,"The other names of $hostname is $othername") if($verbose && ($othername ne ""));
+my $msg = "NIC $installnic exists on current server";
+my $nics = `ip addr show 2>&1 | grep -E '^[1-9]' | cut -d: -f2 |grep $installnic`;
```
daniceexi: We should use the ``ip addr <eth0>`` to show the configuration for **eth0** instead of ``grep``. grep cannot distinguish the ``eth1`` and ``eth11``.

xCAT-probe/lib/perl/probe_utils.pm
```diff
@@ -0,0 +1,341 @@
...
+        Format output message depending on probe framework requirement
+        Format is [<flag>] : <message>
+        The valid <flag> are debug, warning, failed and ok
+    Arguments:
+        output: where should output the message 
+        num:  the number of <flag>
+              0: debug
+              1: warning
+              2: failed
+              3: ok
```
xuweibj: The flag is not same with below.

xCAT-probe/subcmds/xcatmn
```diff
@@ -0,0 +1,551 @@
...
+
+
+#-------------------------------------
+# main process
+#-------------------------------------
+if (
+    !GetOptions("--help|h" => \$help,
+        "t"   => \$test,
+        "V"   => \$verbose,
+        "f=s" => \$output,
```
xuweibj: "f" is not supported.

daniceexi:
I have reviewed the change and agree to merge. Thanks @hu-weihua

PR #1220, opened 2016-05-27
add command switchprobe to retrieve information from switches and/or …

zet809:
…check whether it is ready for MAC searching

### DIFF
--- a/perl-xCAT/xCAT/MacMap.pm
+++ b/perl-xCAT/xCAT/MacMap.pm
@@ -236,6 +236,109 @@ sub scan_qbridge_vlans {
 #TODO: try to scan
     }
 }
+
+#--------------------------------------------------------------------------------
+
+=head3   dump_mac_info
+    Descriptions:
+        Retrieve information (switchport and the mac addresses got for that port) for the specified switch or all switches if no specified.
+    Arguments:
+        $req: the xcat request hash 
+        $callback: the function to output information
+    Returns:
+        The hash variable store retrieved inforamtions
+    Usage example:
+        my $macmap = xCAT::MacMap->new();
+        my $switch_data = $macmap->dump_mac_info($req, $callback);
+        foreach my $switch (keys %$switch_data) {...}
+=cut
+
+#--------------------------------------------------------------------------------
+
+sub dump_mac_info {
+    my $self = shift;
+    my $req = shift;
+    my $callback = shift;
+    my $noderange = undef;
+    if (defined($req->{node})) {
+        $noderange = $req->{node};
+    }
+    my %ret = ();
+    $self->{collect_mac_info} = 1;
+    if (defined($req->{opt}->{verbose})) {
+        $self->{show_verbose_info} = 1;
+        $self->{callback} = $callback;
+    }
+    my $community = "public";
+    my $dump_all_switches = 0;
+    my %switches_to_dump = ();
+    if (!defined($noderange)) {
+        $dump_all_switches = 1;
+    }
+    else {
+        foreach (@$noderange) {
+            $switches_to_dump{$_} = 1;
+        }
+    }
+    my $switchestab = xCAT::Table->new('switches', -create=>0);
+    my @switchesents = $switchestab->getAllNodeAttribs([qw(switch snmpversion username password privacy auth)]);
+    $self->fill_switchparms(community=>$community,switchesents=>\@switchesents);
+    my $switchtab = xCAT::Table->new('switch', -create=>0);
+    my @entries = ();
+    if ($switchtab) {
+        @entries = $switchtab->getAllNodeAttribs(['node','switch','port']);
+    }
+    #Build hash of switch port names per switch
+    $self->{switches} = {};
+    foreach my $entry (@entries) {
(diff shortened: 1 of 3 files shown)


perl-xCAT/xCAT/MacMap.pm
```diff
@@ -236,6 +236,90 @@ sub scan_qbridge_vlans {
 #TODO: try to scan
     }
 }
+
```
daniceexi: Please add description for the subroutine.  Refer to the http://xcat-docs.readthedocs.io/en/stable/developers/guides/code/code_standard.html

perl-xCAT/xCAT/MacMap.pm
```diff
@@ -236,6 +236,90 @@ sub scan_qbridge_vlans {
...
+    if (!defined($noderange)) {
+        $dump_all_switches = 1;
+    }
+    else {
+        foreach (@$noderange) {
+            $switches_to_dump{$_} = 1;
+        }
+    }
+    my $switchestab = xCAT::Table->new('switches', -create=>0);
+    my @switchesents = $switchestab->getAllNodeAttribs([qw(switch snmpversion username password privacy auth)]);
```
daniceexi: I thought the getAllNodeAttribs could only be used to get NODE attributes. The key of the switches table is ``switch`` instead of ``node``.

perl-xCAT/xCAT/MacMap.pm
```diff
@@ -236,6 +236,90 @@ sub scan_qbridge_vlans {
...
+            }
+            if (!defined($self->{macinfo}->{$switch})) { 
+                $ret{$switch}->{ErrorStr} = "No information get";
+                foreach my $defportname (keys %{$self->{switches}->{$switch}}) {
+                    $ret{$switch}->{$defportname}->{Node} = $self->{switches}->{$switch}->{$defportname};
+                }
+            }   
+            elsif (defined($self->{macinfo}->{$switch}->{ErrorStr})) {
+                    $ret{$switch}->{ErrorStr} = $self->{macinfo}->{$switch}->{ErrorStr};
+                    if ($ret{$switch}->{ErrorStr} =~ /username|password/i) {
```
daniceexi: could you add a comment here?

perl-xCAT/xCAT/MacMap.pm
whowutwut: :+1:

xCAT-client/debian/xcat-client.links
```diff
@@ -125,6 +125,7 @@ opt/xcat/bin/xcatclientnnr opt/xcat/bin/rmosdistro
 opt/xcat/bin/xcatclientnnr opt/xcat/bin/rmzone
 opt/xcat/bin/xcatclientnnr opt/xcat/bin/slpdiscover
 opt/xcat/bin/xcatclient    opt/xcat/bin/xCATWorld
+opt/xcat/bin/xcatclient    opt/xcat/bin/switchprobe
```
whowutwut: Should we make this new command or should we have this function in ``xcatprobe switch``.....

whowutwut:
I would like to see in this pull request define some externals for this command and what specific User Case are we trying to solve here.  

`Usage: switchprobe [<noderange>] [-V|--verbose | -c|--check]`

Right now i'm not seeing the value that this command provides... except adding another command into xCAT...  How does the end user use it? I can get the same information by logging into the switch and doing `show mac-address-table interface port X` but this just displays a bit easier.  But why would a user want to do this? 

Should this first go in as a script subcmds for `xcatprobe`?

perl-xCAT/xCAT/MacMap.pm
```diff
@@ -405,11 +489,33 @@ sub handle_output {
   return $rc;
 }
 
+sub walkoid_verbose {
```
daniceexi: @zet809 The ``walkoid`` subroutine supports the namedargs as the third argument. I think we can support the **verbose=>1** in the ``walkoid`` instead of have to use if/else to handle the oid calling for each snmp access.

xCAT-server/lib/xcat/plugins/switch.pm
```diff
@@ -100,6 +101,90 @@ sub preprocess_request {
...
+              }
+              elsif (!defined($swhash->{$node})) {
+                  push @errswnode, $node;
+              }
+              else {
+                  push @switchnode, $node;
+              }
+          }
+          if (@errornode) {
+              $callback->({error=>["Command not support for nodes". join(",",@errornode)],errorcode=>1});
```
daniceexi: * I'd like to add 'the nodetype is not switch' in the message to make it to be more specific.

* do we need to add a ': ' after "Command not support for nodes"?

daniceexi:
@whowutwut The big difference between running ``switchprobe`` and ``manually run switch cli`` is the ``switchprobe`` really run the xcat code which do switch-based discovery. If this passed, we can make sure the switch-based discovery code logic works for certain switch.

For the cmd line, I talked with @zet809 , this command won't be shipped by default. Only ``xcatprobe`` command will create a command file for it. So customer cannot see this command in /opt/xcat/bin/ and we don't need to maintain the doc for this command.

zet809:
@daniceexi I have modified based on your comments, pls review.

daniceexi:
I have reviewed the change and agree to merge. Thanks @zet809

Issue #1219, opened 2016-05-27
[DEV] xcatd keeps wait forever if tftp process can not be stopped


chenglch:
I  encounter this problem when running xcat in docker if /sbin/stop was not hacked. As `in.tftpd` can not be stopped, xcat runs in a loop that can not be break and no any message. See the code detail

https://github.com/xcat2/xcat-core/blob/master/xCAT-server/lib/xcat/plugins/AAsn.pm#L1433

    my @tftpprocs=`ps axf|grep -v grep|grep in.tftpd`;
	while (@tftpprocs) {
		sleep 0.1;   # chenglch: the value of @tftpprocs do not change within this loop
	}
        system("$startcmd");

PR #1221, opened 2016-05-27
Fix error when stopping tftp process in AAsn.pm

chenglch:
The check logic that wait for the tftp process can not be
finished due to the empty loop. This patch aims to check
the status of process group in the loop.

close-issue: #1219

### DIFF
--- a/xCAT-server/lib/xcat/plugins/AAsn.pm
+++ b/xCAT-server/lib/xcat/plugins/AAsn.pm
@@ -1411,31 +1411,38 @@ sub enable_TFTPhpa
 
 
     if (-x "/usr/sbin/in.tftpd") {
-	system("killall in.tftpd"); #xinetd can leave behind blocking tftp servers even if it won't start new ones
-    if ($distro =~ /ubuntu.*/i){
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/AAsn.pm
```diff
@@ -1411,31 +1411,40 @@ sub enable_TFTPhpa
...
-        sleep 1;
-        my @checkproc=`ps axf|grep -v grep|grep in.tftpd`;
-        if (@checkproc){
-            system("stop tftpd-hpa");
+        system("killall in.tftpd"); #xinetd can leave behind blocking tftp servers even if it won't start new ones
+        if ($distro =~ /ubuntu.*/i){
+            sleep 1;
+            my @checkproc=`ps axf|grep -v grep|grep in.tftpd`;
+            if (@checkproc){
+                system("stop tftpd-hpa");
```
immarvin: how about change this line to ``xCAT::Utils->stopservice("tftpd-hpa"); ``

immarvin:
agree to merge

Issue #1224, opened 2016-05-27
[FVT]:xcattest returns Use of uninitialized value if the case fails and there is not stop label defined.


tingtli:
xCAT 2.12 on rhels7.2
Run a case and find there is an error at the end of the result.
```
Use of uninitialized value in pattern match (m//) at /opt/xcat/bin/xcattest line 871.
```
Take a look at the code,  if there is no stop label for a case but the case fails, there comes the error. See lin 871.
```
if($$case{stop} =~ /^yes$/){
```

hu-weihua:
@tingtli , I have fixed this issue by pull request #1226, Could you please verify it?  thanks

tingtli:
Verified with
Version 2.12.1 (git commit 993ee532787e55b21a3d9a77ad5c9f447d07b511, built Tue Jun 14 06:45:40 EDT 2016)
on rhels7.2

PR #1226, opened 2016-05-30
fix bug 1224 xcattest returns Use of uninitialized value if the case fails and there is not stop label defined

hu-weihua:
Fix issue #1224 , xcattest returns Use of uninitialized value if the case fails and there is not stop label defined


hu-weihua:
@tingtli, Could you please review the changing code? thanks

PR #1236, opened 2016-05-31
Update mktoolscenter to incorporate Lenovo tools

arif-ali:
* prioritise IBM tools over Lenovo, in the same way as iflash
* Only use unique values for machine times, don't need to do it more than that


daniceexi:
I have reviewed the change and agree to merge. Thanks @arif-ali

Issue #1223, opened 2016-05-27
[FVT]:xcattest -c does not work for command list


tingtli:
xCAT 2.12
xcattest does not work for command list.
```
c910f03c04k04:/ # xcattest -f linux.conf.template -c lsdef
xCAT automated test started at Fri May 27 04:01:15 2016
******************************
Reading Configure
******************************
OBJECT:c910f03c05k27,TYPE:node
    monserver = 10.3.4.4;
    vmnics = br10;
    currstate = install;
    serialport = 0;
    currchain = boot;
    profile = compute;
    vmcpus = 1;
    serialflow = hard;
    mgt = kvm;
    netboot = grub2;
    tftpserver = 10.3.4.4;
    statustime = 12-15-2015;
    serialspeed = 115200;
    primarynic = mac;
    vmnicnicmodel = virtio;
    mac = 42:dc:da:97:e5:ab;
    cons = kvm;
    groups = all;
    updatestatus = synced;
    vmstorage = dir:///var/lib/libvirt/images/;
    installnic = mac;
    updatestatustime = 08-01-2015;
    status = installing;
    ip = 10.3.5.27;
    nameservers = 10.3.4.4;
    conserver = 10.3.4.4;
    vmhost = c910f03c05;
    provmethod = sles12.1-ppc64le-install-compute;
    os = sles12.1;
    vmmemory = 4096;
    arch = ppc64le;
OBJECT:auto_test,TYPE:network
    mask = 10.0.0.0;
    gateway = 10.0.0.103;
    net = 10_0_0_0-255_0_0_0;
TABLE:passwd
    key = system
    username = root
    password = cluster


TABLE:site
    key = domain
    value = pok.stglabs.ibm.com


    key = forwarders
    value = 9.114.1.1


    key = nameservers
    value = 10.3.4.4


Varible:
    LATEST_VERSION = 2.12
    UBUNTU_MIGRATION2_CORE = https://xcat.org/files/xcat/xcat-core/2.11.x_Ubuntu/xcat-core/xcat-core-2.11.1-ubuntu.tar.bz2
    MaxLparID = 32
    MIGRATION2_DEP = http://xcat.org/files/xcat/xcat-dep/2.x_Linux/xcat-dep-2.11.1.tar.bz2
    MIGRATION1_DEP = http://xcat.org/files/xcat/xcat-dep/2.x_Linux/xcat-dep-2.10.tar.bz2
    CN = c910f03c05k27
[...]

PR #1237, opened 2016-06-01
Fix issue1223 : xcattest -c does not work for command list

hu-weihua:
Fix issue #1223  xcattest -c does not work for command list


hu-weihua:
@tingtli , Could you please review the code changing and merge it? thanks. Any comments are welcome.

daniceexi:

PR #1238, opened 2016-06-01
Discovery static check, check whether genesis files ready

xuweibj:
Just check genesis files. 
"check_genesis_file_ppc64" and "check_genesis_file_x86_64"

### DIFF
--- a/xCAT-probe/subcmds/discovery
+++ b/xCAT-probe/subcmds/discovery
@@ -1,0 +1,275 @@
+#! /usr/bin/perl
+# IBM(c) 2016 EPL license http://www.eclipse.org/legal/epl-v10.html
+
+BEGIN { $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : -d '/opt/xcat' ? '/opt/xcat' : '/usr'; }
+
+use lib "$::XCATROOT/probe/lib/perl";
+use probe_utils;
+use File::Basename;
+use Getopt::Long qw(:config no_ignore_case);
+
+my $program_name = basename("$0");
+my $help;
+my $test;
+my $output  = "stdout";
+my $verbose = 0;
+my $rst     = 0;
+
+$::USAGE = "Usage:
+    $program_name -h
+    $program_name -t
+    $program_name [-V]
+    $program_name -s
+
+Description:
+    Check discovery files or process. 
+
+Options:
+    -h : Get usage information of $program_name
+    -t : To verify if $program_name can work, reserve option for probe framework
+    -V : Output more information for debug
+    -s : Discovery static check, check whether genesis files are ready
+";
+
+sub check_genesis_file {
+    my $arch  = shift;
+    if (($arch ne "ppc64") and ($arch ne "x86_64")) {
+        probe_utils->send_msg("$output", "f", "Please input correct arch type");
+        return 1;
+    }
+ 
+    my $rst_f = 0;
+    probe_utils->send_msg("$output", "d", "Start to check genesis files for $arch...") if ($verbose);
+
+    my $os = probe_utils->get_os();
+    my $genesis_base;
+    my $genesis_scripts;
+
+    if ($arch eq "x86_64") {
+        $arch_tmp = "amd64";
+    } else {
+        $arch_tmp = $arch;
+    }
+
+    if ($os =~ "unknown") {
+        probe_utils->send_msg("$output", "d", "The OS is not supported.") if ($verbose);
+        return 1;
+    } elsif ($os =~ "ubuntu") {
+        $genesis_base = `dpkg -l | grep -i "xcat-genesis-base" | grep -i "$arch_tmp"`;
+        $genesis_scripts = `dpkg -l | grep -i "xcat-genesis-scripts" | grep -i "$arch_tmp"`;
+    } else {
+        $genesis_base = `rpm -qa | grep -i "xcat-genesis-base" | grep -i "$arch"`;
+        $genesis_scripts = `rpm -qa | grep -i "xcat-genesis-scripts" | grep -i "$arch"`;
+    }
+    unless ($genesis_base and $genesis_scripts) {
+        probe_utils->send_msg("$output", "d", "xCAT-genesis for $arch did not be installed.") if ($verbose);
+        return 1;
+    }
+
+    probe_utils->send_msg("$output", "d", "xCAT-genesis for $arch installed, start to check files...") if ($verbose);
+
+    my $tftpdir = `tabdump site | grep "tftpdir" | awk -F "," '{print \$2}'`;
+    chomp($tftpdir);
+    $tftpdir =~ s/"//g;
+    my $genesis_folder;
+    my @genesis_files;
+    my $genesis_line;
+    my $wget_rst;
+
+    if ($arch eq "ppc64") {
+        $genesis_folder = "$tftpdir/pxelinux.cfg/p";
+        unless (-d "$genesis_folder") {
+            probe_utils->send_msg("$output", "d", "There is no genesis file for $arch. Please run 'mknb ppc64' if you use ppc64/ppc64le machine.") if ($verbose);
+            return 1;
+        }
+
+        @genesis_files = glob("$genesis_folder/*");
+
+        foreach (@genesis_files) {
+            unless (open(FILE, $_)) {
+                probe_utils->send_msg("$output", "d", "Cannot open file $_.") if ($verbose);
+                $rst_f = 1;
+                next;
+            }
+
+            while ($genesis_line = <FILE>) {
+                chomp($genesis_line);
+                $genesis_line =~ s/^\s+|\s+$//g;
+
+                if ($genesis_line =~ /^initrd/) {
+                    @initrd_info = split(' ', $genesis_line);
+                    $initrd_path = $initrd_info[1];
+                    $wget_rst = system("wget -q --spider $initrd_path -T 0.5 -t 3");
+                    if ($wget_rst) {
+                        probe_utils->send_msg("$output", "d", "initrd cannot be downloaded from $initrd_path.") if ($verbose);
+                        $rst_f = 1;
+                    } else {
+                        probe_utils->send_msg("$output", "d", "Check initrd file: $initrd_path PASS.") if ($verbose);
+                    }
+                }
+
+                if ($genesis_line =~ /^kernel/) {
+                    @kernel_info = split(' ', $genesis_line);
+                    $kernel_path = $kernel_info[1];
+                    $wget_rst = system("wget -q --spider $kernel_path -T 0.5 -t 3");
+                    if ($wget_rst) {
+                        probe_utils->send_msg("$output", "d", "kernel cannot be downloaded from $kernel_path.") if ($verbose);
+                        $rst_f = 1;
+                    } else {
+                        probe_utils->send_msg("$output", "d", "Check kernel file: $kernel_path PASS.") if ($verbose);
+                    }
+                }
+            }
+        }
+    } else {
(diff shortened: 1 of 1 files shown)


xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,225 @@
...
+
+Options:
+    -h : Get usage information of $proname
+    -t : To verify if $proname can work, reserve option for probe framework
+    -V : Output more information for debug
+    -s : Discovery static check, check whether genesis files are ready
+";
+
+sub check_genesis_file_ppc64 {
+    probe_utils->send_msg("$output", "d", "Beginnig check genesis files for ppc64...") if ($verbose);
```
daniceexi: ``Beginnig`` is typo;
We may say: Start to check genesis ...

xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,225 @@
...
+        probe_utils->send_msg("$output", "w", "There is no genesis file for x86_64. Please run 'mknb x86_64' if you use x86_64 machine.");
+        return;
+    }
+
+    my $rst_c        = "o";
+    my @gene_files_x = glob("$gene_folder_x/*");
+    foreach (@gene_files_x) {
+        if ($_ =~ /uefi$/) {
+            my @file_split = split('/', $_);
+            my $file_name  = pop(@file_split);
```
hu-weihua: Maybe perl function ``basename`` can obtain file name more easily.  just  ``my $file_name=basename($_)``

xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,225 @@
...
+            my @ip_range   = split('\.', $file_name);
+            my $host_ip = `ifconfig -a | grep inet | grep -v inet6 | awk '{print \$2}' | grep $ip_range[0]`;
```
hu-weihua: Maybe ``ifconfig -a | awk -F" " '/inet / {print \$2}' `` is a better option. save two ``grep`` (save two ``fork`` in system level)

xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,225 @@
...
+            chomp($host_ip);
+
+            unless ($host_ip) {
+                probe_utils->send_msg("$output", "f", "There is no ip for range '$ip_range[0].0.0.0'.");
+                $rst_c = "f";
+                next;
+            }
+
```
hu-weihua: If we hope to test if there is a ip configured on current server belongs to the network ``ip_range``, above code logic can't achieve what we expected.   `` $ip_range[0]`` just a number,  it means nothing.
There is a function in ``probe_utils.pm`` named  ``is_ip_belong_to_net``, we can use this fuction to test if one ip belongs to a network.

xuweibj:
Hi @hu-weihua , I modified, please help review again. Thank you.

xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    my $gene_folder_x = "/tftpboot/xcat/xnba/nets";
+    unless (-d "$gene_folder_x") {
+        probe_utils->send_msg("$output", "d", "There is no genesis file for x86_64. Please run 'mknb x86_64' if you use x86_64 machine.") if ($verbose);
+        return 1;
+    }
+
+    my $host_ip_str  = `ifconfig -a | awk -F" " '/inet / {print \$2}'`;
+    my $netmask_str  = `ifconfig -a | awk -F" " '/inet / {print \$4}'`;
```
hu-weihua: In rhels6, the output of ``ifconfig -a | awk -F" " '/inet / {print \$2}'`` maybe can't satisfy what we expected.  Below is a example of the output from rhels6

    -bash-4.1$ ifconfig -a |awk -F" " '/inet / {print $2}'
    addr:172.29.97.51
    addr:9.114.120.101
    addr:10.0.0.101
    addr:127.0.0.1

There is ``addr:``.  Maybe using ``ip addr`` instead of  ``ifconfig -a`` is a option.  I am not very sure about that, we can do more test in different distro

xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,272 @@
...
+# IBM(c) 2016 EPL license http://www.eclipse.org/legal/epl-v10.html
+
+BEGIN { $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : -d '/opt/xcat' ? '/opt/xcat' : '/usr'; }
+
+use lib "$::XCATROOT/probe/lib/perl";
+use probe_utils;
+use File::Basename;
+use Getopt::Long qw(:config no_ignore_case);
+
+my $proname = basename("$0");
```
whowutwut: What is "proname"?  

Should it be "progname" or "program_name" for readability.  I don't necessarily think we need to shorten the variable names if it helps readability of the code.

xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,272 @@
...
+    -V : Output more information for debug
+    -s : Discovery static check, check whether genesis files are ready
+";
+
+sub check_genesis_file_ppc64 {
+    my $rst_p = 0;
+    probe_utils->send_msg("$output", "d", "Start to check genesis files for ppc64...") if ($verbose);
+
+    my $os = probe_utils->get_os();
+    my $gene_base;
```
whowutwut: I would recommend not abbreviating genesis_base and genesis_scripts... 
but If we really don't want to type out readable variable names, just use `base` and `scripts` as the variables.

xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,272 @@
...
+    Check discovery files or process. 
+
+Options:
+    -h : Get usage information of $proname
+    -t : To verify if $proname can work, reserve option for probe framework
+    -V : Output more information for debug
+    -s : Discovery static check, check whether genesis files are ready
+";
+
+sub check_genesis_file_ppc64 {
```
whowutwut: There's a lot of common code here.  I think it would be better if we created more generic functions that could be reused instead of having 1 function to check ppc64 and one function to check x86 based. 

For example, `sub check_genesis_files(type)` where type is `ppc64` or `x86_64`

xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,272 @@
...
+    } else {
+        $gene_base = `rpm -qa | grep -i "xcat-genesis-base" | grep -i "ppc64"`;
+        $gene_scripts = `rpm -qa | grep -i "xcat-genesis-scripts" | grep -i "ppc64"`;
+    }
+    unless ($gene_base and $gene_scripts) {
+        probe_utils->send_msg("$output", "d", "xCAT-genesis for ppc64 did not be installed.") if ($verbose);
+        return 1;
+    }
+
+    my $gene_folder_p = "/tftpboot/pxelinux.cfg/p";
```
whowutwut: This could be another function that would return according to a variable passed in, for example: 

```
sub get_genesis_folder($type) { 
   if ($type =~ "ppc64") {
       return "/tftpboot/pxelinux.cfg/p";
   } elsif ($type ~= "x86_64") { 
       return "/tftpboot/xcat/xnba/nets";
   } else {
       ERROR 
   }
}
```

xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,272 @@
...
+        return 1;
+    } elsif ($os =~ "ubuntu") {
+        $gene_base = `dpkg -l | grep -i "xcat-genesis-base" | grep -i "ppc64"`;
+        $gene_scripts = `dpkg -l | grep -i "xcat-genesis-scripts" | grep -i "ppc64"`;
+    } else {
+        $gene_base = `rpm -qa | grep -i "xcat-genesis-base" | grep -i "ppc64"`;
+        $gene_scripts = `rpm -qa | grep -i "xcat-genesis-scripts" | grep -i "ppc64"`;
+    }
+    unless ($gene_base and $gene_scripts) {
+        probe_utils->send_msg("$output", "d", "xCAT-genesis for ppc64 did not be installed.") if ($verbose);
```
whowutwut: Here's a question that I keep asking myself @daniceexi  is there a situation where the genesis base does not get installed? I have never seen this situation but I think it comes up a lot in discussion...

xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,272 @@
...
+    }
+    exit 0;
+}
+
+if ($test) {
+    probe_utils->send_msg("$output", "o", "Discovery Check.");
+    exit 0;
+}
+
+if ($static) {
```
whowutwut: Here we are checking both ppc64 and x86_64, if the user is NOT doing any mixed cluster discovery, the results will not make sense.   I think we should detect the architecture that the probe command is being run from, maybe that should be a utility that is added to the library... then print out a message accordingly. 

For example:
```
if runnign on ppc64: 
    check_genesis_files_ppc64() 
    if good and ppc64 { 
        // print some message saying looks good
    }
    elsif failed and x86_64 {
        // print some warning msg, if you want to discovery x86_64 machines, need to do ...
    }
    else {
        // failure
    }
}
similar thing for x86_64
```

whowutwut:
I am questioning why we need to have this discovery option for  xcatprobe.  It seems like it's checking for the  /tftpboot  directory files that are created when `mknb` is executed.  But I still don't know when this would happen that mkbn is not there....  so I just don't want to add probe functions for situations that normally does not happen.  IF the user deletes `/tftpboot` directory, that's not something we should look for.. is it??  Would just running mkbn restore ALL the files needed or they would need to re-install RPMS.

xCAT-probe/subcmds/discovery
hu-weihua: Thanks @whowutwut , this is a good suggestion to handle mixed cluster.  The main job of @xuweibj  is to implement the function of checking genesis, so I will add your this comment to discovery main framework.

xCAT-probe/subcmds/discovery
```diff
@@ -0,0 +1,280 @@
...
+    my $rst_p = 0;
+    probe_utils->send_msg("$output", "d", "Start to check genesis files for ppc64...") if ($verbose);
+
+    if (check_genesis_packages("ppc64")) { 
+        return 1; 
+    } else {
+        probe_utils->send_msg("$output", "d", "xCAT-genesis for 'ppc64' installed, start to check files...") if ($verbose);
+    }
+
+    my $genesis_folder_p = "/tftpboot/pxelinux.cfg/p";
```
hu-weihua: The ``tftpdir`` is a attribute which user can set in ``site`` table, so we can't suppose it is ``/tftpboot/`` in all the time. We need to get ``tftpdir`` from ``site`` table first.

hu-weihua:
Hi @whowutwut, for why we need to check genesis files during static discovery probe, I think the major purpose is not to check if these genesis files exist, the major purpose is to make sure these files can be download during running time. we do simulate the download process in ``check_genesis_file`` function. Yes, you are right, if we just need to check if these files exist, this is ``mknb`` 's job. But ``mknb`` can't guarantee all files can be downloaded at running time.

hu-weihua:
@whowutwut , we need to keep move of this work item, so if you have any more comments about this item, you can comment for later related pull request. thanks.

hu-weihua:
I have reviewed this pull request, I agree to merge.

whowutwut:
@xuweibj @hu-weihua  I ran this function and wondering why we use the term "static"... do we have any plan for what externals are going to be supported with this command? 

Another question I have is maybe we should install both the x86_64 and ppc64 genesis packages by default. I don't think we are saving anything by not doing this, then we don't even have the need to check this...

daniceexi:
I agree to install genesis for both x86_64 and ppc64le

PR #1239, opened 2016-06-01
Change sleep time from 0.1 to 1 as float value is not support by default

chenglch:


daniceexi:

Issue #1168, opened 2016-05-20
[FVT] Error message found in CN:/var/log/xcat/xcat.log after CN is booted.


caomengmeng:
Find below error in CN:/var/log/xcat/xcat.log after install sles12.1 successfully with latest daily build:
```
  <error>Unable to read private ECDSA key from /etc/xcat/hostkeys</error>
  <error>Unable to read private ECDSA key from /etc/xcat/hostkeys</error>
```
Test ENV:
```
[root@c910f result]# rpm -qa | grep -i xCAT
xCAT-server-2.12-snap201605190710.noarch
xCAT-test-2.12-snap201605190711.noarch
syslinux-xcat-3.86-2.noarch
xCAT-genesis-base-x86_64-2.12-snap201605051534.noarch
ipmitool-xcat-1.8.15-1.x86_64
xCAT-client-2.12-snap201605190710.noarch
xCAT-buildkit-2.12-snap201605190711.noarch
xCAT-2.12-snap201605190711.x86_64
elilo-xcat-3.14-4.noarch
conserver-xcat-8.1.16-10.x86_64
grub2-xcat-2.02-0.16.el7.snap201506090204.noarch
perl-xCAT-2.12-snap201605190710.noarch
xCAT-genesis-scripts-x86_64-2.12-snap201605190711.noarch
[root@c910f result]# cat /etc/*release
Red Hat Enterprise Linux Server release 6.7 (Santiago)
Red Hat Enterprise Linux Server release 6.7 (Santiago)
```

daniceexi:
This is not a problem.
The ``ECDSA key`` is not supported if the xCAT MN is a a little old. So in your scenario, the rh6.7 MN does not generate the ``ECDSA`` key by default.

daniceexi:
The possible thing we need to fix is whether the error message can be improved.
@caomengmeng could you paste the error message here?

caomengmeng:
@daniceexi , Ｄo you mean the message in MN?
The message in CN is already in this issue, please see the beginning of first comments.
```
  <error>Unable to read private ECDSA key from /etc/xcat/hostkeys</error>
  <error>Unable to read private ECDSA key from /etc/xcat/hostkeys</error>
```

gurevichmark:
@daniceexi I had to fix something similar for the genesis boot in #888. Same error was being displayed.
The issue was the command `ssh-keygen` inside the `doxcat` script.
[...]

PR #1243, opened 2016-06-01
Discard grep output and display error message

gurevichmark:
Fix for issue #1168 
- Discard output of `grep` command when checking for error from `getcredentials.awk`.
- Use `-s` option of existing `logger` statement to display the text of the error on console.


daniceexi:

Issue #1036, opened 2016-05-04
[FVT]2.12 switchdiscover could not get switch mac using snmp for Mellanox switch


junxiawang:
env:rhels
build:latest build
lsdef - Version 2.12 (git commit 6e2f230ae49477612dc94d533afca67962d8e33f, built Tue May  3 09:30:57 EDT 2016

How to reproduce:
```
[root@c910f04x30 ~]# switchdiscover  --range  10.0.0.131 -s snmp
Discovering switches using snmpwalk.....
ip          	name           	vendor                                  	mac         
------------	------------   	------------                            	------------
10.0.0.131  	c910ibsw01     	Mellanox SX6036,MLNX-OS,SWvSX_3.3.5006  	nomac_nmap_0
[root@c910f04x30 ~]# switchdiscover  --range 10.0.0.131
Discovering switches using nmap for 10.0.0.131. It may take long time...
ip          	name           	vendor                                  	mac         
------------	------------   	------------                            	------------
10.0.0.131  	switch-10-0-0-131	Mellanox Technologies                   	00:02:C9:63:39:16
--------------------------------> in fact I could find mac using nmap
```

whowutwut:
Is nmap more consistent in getting the correct information back or snmp?

cxhong:
Well, I couldn't find which scan method is more consistent. Sometimes nmap couldn't find host name but snmp could.  Same as MAC address in this case, snmp couldn't, but nmap could.  I think it's relate to switch how it configured.  Users have to decided which method is working better for them. 
Or we can redesign the switchdiscover command, combine nmap and snmp scan method, returned whatever we can from those two commands.

daniceexi:
Generally mac address is not necessary for the management of a discovered switch, so it's fine to keep current code for 2.12.

But mac address is useful is we want to use switch-based discovery to discover a switch location. For how to get the mac address, I am thinking whether it easier that just getting it from arp or ping?

cxhong:
we can get mac address from snmp scan.  User may need to try both scan to find better method.

PR #1244, opened 2016-06-01
Use arp command to get switch mac address

cxhong:
The switchdiscover command couldn't find Mellox switch's Mac address via snmpwalk,  we are change it to use arp -n .  this is for issue #1036

### DIFF
--- a/perl-xCAT/xCAT/Utils.pm
+++ b/perl-xCAT/xCAT/Utils.pm
@@ -4711,5 +4711,37 @@ sub get_nmapversion {
     return $nmap_version;
 }
 
+#--------------------------------------------------------------------------------
+=head3  get_macbyarp
+    Get Mac address by arp -n
+    Returns:
+       mac:  Mac address
+=cut
+#--------------------------------------------------------------------------------
+
+sub get_macbyarp {
+    my $arptable;
+    my $mac;
+    my $ip = shift;
+    if ($ip =~ /xCAT::Utils/)
+    {
+        $ip = shift;
+    }
+    if ( -x "/usr/sbin/arp") {
+        $arptable = `/usr/sbin/arp -n`;
+    }
+    else{
+        $arptable = `/sbin/arp -n`;
+    }
+    my @arpents = split /\n/,$arptable;
+    foreach  (@arpents) {
+        if (m/^($ip)\s+\S+\s+(\S+)\s/) {
+            $mac=$2;
+            last;
+        }
+    }
+
+    return $mac;
+}
 
 1;
--- a/perl-xCAT/xCAT/Utils.pm
+++ b/perl-xCAT/xCAT/Utils.pm
@@ -4711,37 +4711,4 @@ sub get_nmapversion {
     return $nmap_version;
 }
 
-#--------------------------------------------------------------------------------
-=head3  get_macbyarp
-    Get Mac address by arp -n
-    Returns:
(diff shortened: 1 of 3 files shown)


cxhong:
@daniceexi  can you review this changes?

perl-xCAT/xCAT/Utils.pm
```diff
@@ -4711,5 +4711,37 @@ sub get_nmapversion {
     return $nmap_version;
 }
 
+#--------------------------------------------------------------------------------
```
daniceexi: @cxhong There's already a subroutine fro get mac from arp

~~~~
./xCAT-server/lib/perl/xCAT/SvrUtils.pm:=head3  get_mac_by_arp
~~~~

cxhong:
@daniceexi 
the subroutine get_mac_by_arp() in the SvrUtils.pm will call pping first
````
 my @output = xCAT::Utils->runcmd("/opt/xcat/bin/pping $noderange", -1);
``````
and it will failed because the switch is not defined in the xcat database yet and pping subroutine will return output as "unreachable"

So, add a simple get_macbyarp subroutine in the Utils.pm works better for this case.

**Comment by daniceexi** [comment]
@cxhong it's not a good idea to add more than one subroutine to do similar thing but in different utils files. Especially that user cannot figure them out easier that which one should be used through the ``description`` for the subroutines.

I suggest to add a new argument 'noping' for ``SvrUtils:: get_mac_by_arp`` to indicate that don't run the ``pping`` against the node or ip, but just run arp to get the mac.

**Comment by cxhong** [comment]
@daniceexi , I modified the code with:
1) used Svrutils::get_mac_by_arp subroutine
2) add message if snmp port is disabled.

File: xCAT-server/lib/xcat/plugins/switchdiscover.pm [hunk 2]
```diff
@@ -855,19 +860,35 @@ sub snmp_scan {
...
+        }
+
         my $vendor = get_snmpvendorinfo($request, $ip);
         if ($vendor) {
-            my $mac = xCAT::Utils->get_macbyarp($ip);
+            my $display = "";
+            my $nopping = "nopping";
+            my $output = xCAT::SvrUtils->get_mac_by_arp([$ip], $display, $nopping);
+            my $mac;
+            foreach my $node (keys %{$output}) {
```
daniceexi: if $output is a hash, you can use ``$output->{$ip}`` to get the value directly instead of a loop.

daniceexi: The indent should be 4 spaces.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -826,13 +827,17 @@ sub snmp_scan {
     my $ranges = get_ip_ranges($request);
 
 
-    #use nmap to find if snmp port is enabled  
-    # only open port will be scan
+    ##########################################################
+    #use nmap to parse the ip range and possible output from the command:
```
daniceexi: Do we need to add the output example for nmap 5.10 and earlier?

cxhong:
@daniceexi  Thanks,  code modified.

cxhong:
@daniceexi , can you review this again?  Thanks

daniceexi:
I have reviewed the change and agree to merge. Thanks. @cxhong

PR #1250, opened 2016-06-03
according to switchdiscover code modify

junxiawang:
@tingtli  modify switchdiscover testcase according to switchdiscover code change


junxiawang:
meng meng has been modified the case

PR #1251, opened 2016-06-03
modify statelite image

junxiawang:
@tingtli  please help review for statelite image modify


junxiawang:
@tingtli add statelite image definition creation and remove

Issue #1233, opened 2016-05-31
[FVT] It takes more than 6 hours to complete the rnetboot command on sles11.4 ppc64


caomengmeng:
Test ENV:

```
c910f02c01p:/opt/xcat/share/xcat/tools/autotest # lsxcatd -a
Version 2.12.1 (git commit ff6b02e84853a13437ace7f4a3471d12bef2a5e9, built Mon May 30 06:45:56 EDT 2016)
This is a Management Node
cfgloc=mysql:dbname=xcatdb;host=10.2.1.14|xcatadmin
dbengine=mysql
dbname=xcatdb
dbhost=10.2.1.14
dbadmin=xcatadmin
c910f02c01p:/opt/xcat/share/xcat/tools/autotest # cat /etc/*release
LSB_VERSION="core-2.0-noarch:core-3.2-noarch:core-4.0-noarch:core-2.0-ppc64:core-3.2-ppc64:core-4.0-ppc64"
NAME="SLES"
VERSION="11.4"
VERSION_ID="11.4"
PRETTY_NAME="SUSE Linux Enterprise Server 11 SP4"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:11:4"
SUSE Linux Enterprise Server 11 (ppc64)
VERSION = 11
PATCHLEVEL = 4
```
After rnetboot command, it takes "22374 sec" to complete. See below:
```
RUN:if [[ "ppc64" =~ "ppc64" ]]; then rnetboot c910f02c01p08;elif [[ "ppc64" =~ "x86_64" ]];then rpower c910f02c01p08 boot; fi

[if [[ "ppc64" =~ "ppc64" ]]; then rnetboot c910f02c01p08;elif [[ "ppc64" =~ "x86_64" ]];then rpower c910f02c01p08 boot; fi] Running Time:22374 sec
RETURN: rc = 0
OUTPUT:
c910f02c01p08: Success
CHECK:rc == 0   [Pass]
```

Although rnetboot successfully, SN status is still powering off:

```
c910f02c01p:/opt/xcat/share/xcat/tools/autotest # lsdef -l c910f02c01p|grep status
    status=powering-off
    statustime=05-30-2016 21:27:24
```

caomengmeng:
reproduced with latest daily build:
```
c910f02c01p14:/var/log/xcat # lsxcatd -a
[...]

PR #1253, opened 2016-06-03
issue 1233: try to rekill tftp and add more debug info when failed to…

daniceexi:
… kill tftpd


daniceexi:
fix for #1233

immarvin:
agree to merge

PR #1256, opened 2016-06-06
Fix syntax error when checking the existence of tftp process

chenglch:


daniceexi:

PR #1273, opened 2016-06-07
correct the logic to determine the ENV{XCATMASTER}

immarvin:
the logic to determine the $ENV{XCATMASTER} confirm to the following priority(from high to low):
  1. the "xcatmaster" attribute of the node
  2. the site.master
  3. the ip address of the mn facing the compute node


daniceexi:
I have reviewed the change and agree to merge

daniceexi:
cherry-picked to 2.12-pcm too.

PR #1278, opened 2016-06-07
[xCAT Jenkins] Log Analyzer with email report

neo954:
This code is used to create an email report to send to the team regarding the Jenkins testing framework.  It tracks the success rate to help us analyze the stability of the testing suite.


xCAT-server/share/xcat/tools/jenkins/testreport/send-report.sh
```diff
@@ -0,0 +1,341 @@
...
+
+# The main part
+
+for c in mysql tail sed grep
+do
+	! type "${c}" >/dev/null 2>&1 &&
+		echo "Command \"${c}\" not found" >&2 && exit 1
+done
+
+Email report
```
whowutwut: Is this line suppose to be a comment?

whowutwut: Can we create this email report list by checking for a file and reading the list of users from that file.  This way we can add/remove members of the list without modifying this piece of code

xCAT-server/share/xcat/tools/jenkins/testreport/send-report.sh
```diff
@@ -0,0 +1,341 @@
...
+$report_setCc      "GONG Jie"         gongjie@linux.vnet.ibm.com
+$report_setReplyTo "GONG Jie"         gongjie@linux.vnet.ibm.com
+$report_setFrom    "xCATjk Mail Bot"  root@c910f04x31.pok.stglabs.ibm.com
+
+DateTime="$(date -R)"
+
+MYSQL_COMMAND=("mysql" "-h" "${MYSQL_HOST}" -u "${MYSQL_USER}" -p"${MYSQL_PASS}" "${MYSQL_DB}")
+Subject="$("${MYSQL_COMMAND[@]}" <<<"SELECT CONCAT('Passed: ', SUM(Passed), ' Failed: ', SUM(Failed), ' No run: ', SUM(\`No run\`)) AS Summary FROM LatestDailyReport;" | tail -n 1)"
+
+$report_setSubject "[xCATjk] ${Subject}"
```
whowutwut: I'd suggest we have the subject be: 

`[xCAT Jenkins Automation Testing - mm-dd-yyyy]` where mm-dd-yyyy is the date

whowutwut:
Not something that needs to be changed in this pull request but thinking ahead for future... Instead of having emails sent, could we have a very small status email, then a URL that points to a webserver that hosts these status pages.  

Then we could simply have style sheets to control the display, and a history of the records will be in our GPFS share, also reduce the number of emails that we get.

xCAT-server/share/xcat/tools/jenkins/testreport/send-report.sh
```diff
@@ -0,0 +1,341 @@
...
+
+$report_setHTML <<-EOF
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
```
whowutwut: @neo954  First off, this email report looks very good.   Thank you. 

I think it's not that easy to make changes because the text is duplicated in lots of places. Is there any way to use variables to control all the duplicate text?   

Suggest the following small changes:
"No run" -> "Not Run"
"Test run" -> "Test Run"
"Pass rate" -> "Pass Rate"

Suggest changing to "7-day results"
Suggest changing to "30-day results"

xCAT-server/share/xcat/tools/jenkins/testreport/email.sh
```diff
@@ -0,0 +1,398 @@
+#!/bin/bash
+
+#
+# Author:   GONG Jie <gongjie@linux.vnet.ibm.com>
+# Create:   2016-05-27
+# Update:   2016-06-07
+# Version:  0.99
```
whowutwut: I don't think we need to track this version number since it's tracked in Git.

xCAT-server/share/xcat/tools/jenkins/testreport/email.sh
```diff
@@ -0,0 +1,398 @@
...
+	local mailAttachmentContentTypes="${this}_mailAttachmentContentTypes"
+	local mailAttachmentMessages="${this}_mailAttachmentMessages"
+	local mailAttachmentNames="${this}_mailAttachmentNames"
+
+	eval "${mailAttachmentContentTypes}+=(\"${inContentType}\")"
+	eval "${mailAttachmentMessages}+=(\"${inMessage}\")"
+	eval "${mailAttachmentNames}+=(\"${inName}\")"
+}
+
+function Email_addAttachmentFile()
```
whowutwut: Is this function being used? 

My only point here is that we should focus on things that we actually are using and not increase the scope of the work.  If we never use it, it's more code to maintain and takes time to implement.

xCAT-server/share/xcat/tools/jenkins/testreport/send-report.sh
neo954: @whowutwut, this line is not a comment, the line 36 is a function call of ``Email``.

xCAT-server/share/xcat/tools/jenkins/testreport/email.sh
neo954: @whowutwut, this function is not used currently. But in case we need it, I put it here. We used to attach failed test logs in the email reports.

xCAT-server/share/xcat/tools/jenkins/testreport/send-report.sh
whowutwut: :) Thanks. sorry about that.

whowutwut:
I'm okay with these changes, I'll let @tingtli  merge when she's ready.

tingtli:
I am ok with the script. Just one comment about the code format: Do we need to apply the 4 blanks indent?

neo954:
A bunch of commits were pushed.

neo954:
Now, the daily email report becomes a MySQL procedure.

This change makes the email report consistency. Since in previous implement, the email report was created with multiple SQL statement, and I did not apply any read lock to the tables. This behavior may cause inconsistent result of the report, since the data may change during the multiple database reading. And the function call of ``NOW()`` was performed a bunch of times during the email report generation. This may also cause inconsistency, as the time passed by during multiple database reading.

 Of course, this change also makes the script of ``send-report.sh`` much simpler than before.

neo954:
@caomengmeng, please help on review and merge these codes.

neo954:
This pull request was opened here for more than two months, @tingtli, @caomengmeng, @daniceexi, @whowutwut please help on merge all these patches.

tingtli:
I agree to merge. @whowutwut @daniceexi  what's your opinion?

whowutwut:
I have no issues to merge it.

whowutwut:
@neo954 pushed a commit to fix the NULL, I will merge this now. 

Does the Jenkins test framework do a daily pull on the test framework? Maybe not... since I think @neo954  fixed the problem manually...

@tingtli @neo954  Please take a look at this: https://help.github.com/categories/automation/, I think we should enable OAUTH tokens to get the changes to the test machine rather than manual intervention.

PR #1283, opened 2016-06-08
bmcsetup requires some changes to work around some issues with OpenPower BMC

whowutwut:
After review of the bmcsetup file that is executed in the genesis kernel during hardware discovery, a few things need to be changed in the bmcsetup file to work around some of the behavioral differences in the OpenPower BMC implementation.  

- Cleaned up the spacing of the code/formatting
- Added comments on the ipmitool raw commands to help understand the actions being performed
- Instead of blinking the light at the end of bmcsetup, use a different ipmitool command to light the LED for 5 minutes. 
- The mc reset cold for OpenPower BMC needs to query the status to ensure that the bmc is really in a ready state before continuing on 
- After making changes to networking commands, sleep 30 for OpenPower BMC (workaround)

Creating this pull request to start review and discussion with the team

### DIFF
--- a/xCAT-genesis-scripts/bin/bmcsetup
+++ b/xCAT-genesis-scripts/bin/bmcsetup
@@ -27,7 +27,7 @@ TIMEOUT=15
 #
 function cold_reset_bmc() { 
     if [ -z $XPROD ]; then 
-        echo "FATAL ERROR - XPROD must be set before calling cold_reset_bmc()"
+        logger -s -t $log_label -p local4.crit "FATAL ERROR - XPROD must be set before calling ${FUNCNAME[0]}"
         exit 1
     fi
 
@@ -40,7 +40,6 @@ function cold_reset_bmc() {
             sleep 30
         fi
         logger -s -t $log_label -p local4.info "Resetting BMC ..."
-        echo "Resetting BMC ..."
         ipmitool mc reset cold
 
         logger -s -t $log_label -p local4.info "Waiting for the BMC to appear ..."
@@ -59,7 +58,7 @@ function cold_reset_bmc() {
                 tries=$(($tries+1))
             done
             TOTAL_SEC=$((${SLEEP_INTERVAL} * ${MAX_ITERATION}))
-            echo "ERROR, After waiting ${TOTAL_SEC} seconds, the BMC is not in a ready state."
+            logger -s -t $log_label -p local4.error "ERROR, After waiting ${TOTAL_SEC} seconds, the BMC is not in a ready state."
         else
             # for Non OpenPower servers, just sleep for some set time.
             sleep 15 
@@ -83,7 +82,7 @@ function cold_reset_bmc() {
 #
 function snooze() { 
     if [ -z $XPROD ]; then 
-        echo "FATAL ERROR - XPROD must be set before calling snooze()"
+        logger -s -t $log_label -p local4.crit "FATAL ERROR - XPROD must be set before calling ${FUNCNAME[0]}"
         exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -221,228 +288,225 @@ done
...
+    CURRENTUSER=`ipmitool -d $idev user list $LANCHAN|grep ^$USERSLOT|awk '{print $2}'`
+    DISABLEUSERS=`echo 1 2 3 4|sed -e s/$USERSLOT//`
+    for user in $DISABLEUSERS; do
+        # Disable the non-specified user 
+        while ! ipmitool -d $idev user disable $user; do
+            sleep 1
+            let TRIES=TRIES+1
+            if [ $TRIES -gt $TIMEOUT ]; then break; fi
+        done
+        TRIES=0
```
whowutwut: @daniceexi  I think this line should be deleted. I think after formatting the code better, we see that this setting of TRIES=0 is not needed.

daniceexi: @whowutwut There's a ``for`` loop outside the ``while`` loop, this '`TRIES=`` is used to initialize the ``TRIES`` when running the next ``for`` loop.

daniceexi: I think this commit is used for tidy the format, so I did not review this commit.

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -497,10 +497,16 @@ while [ $idev -gt 0 ]; do
     fi
 
     echo "Lighting Identify Light"
-    while :
-        # Identify the server by turning on the LED light
-        do ipmitool -d $idev raw 0 4 10 > /dev/null
-        sleep 7
-    done &
+    if [ "$XPROD" = "43707" ]; then
```
daniceexi: @whowutwut Could someone in POK really take a look the LED reaction of the server (both x86 and OpenPower) when we use the following two ways to operate the server?

~~~~
ipmitool chassis identify 300
~~~~

OR
~~~~
while :
+            # Identify the server by turning on the LED light
+            do ipmitool -d $idev raw 0 4 10 > /dev/null
+            sleep 7
+        done
~~~~

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -12,6 +12,59 @@
...
 #
+
+
+# Cold reset the BMC for certain servers
+#    Product ID: 309   - x3755 M4 (8722)
+#    Product ID: 43707 - IBM Power S822LC and S812LC
+#
+# Otherwise the BMC will not respond to ping after running the ipmitool commands in this script
+#
+function cold_reset_bmc() {
```
daniceexi: @whowutwut In the version which changed by Er Tao, the code block for ``309`` and ``43707`` were separated since there were too much difference.

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -12,6 +12,59 @@
...
+    if [ "$PROD" = "309" -o "$PROD" = "43707" ] ; then
+        if [ "$PROD" = "43707" ]; then
+            # OpenPower SPECIFIC, the OpenPower machines with AMI BMC should NOT need a 
+            # reset after applying ipmitool commands.  However, it seems there is a problem with 
+            # the BMC where after 15 seconds, it stops responding.  To work around, sleep 30
+            # seconds before issuing the reset of the BMC.
+            sleep 30
+        fi
+        echo "Resetting BMC ..."
+        ipmitool mc reset cold
```
daniceexi: Looks like the ``cold reset`` for Firestone may not necessary from Uma. And Uma mentioned the major issue here is bmcetup tries to access BMC before it's ready.

But totally remove the ``cold reset`` for Firestone may cause several issues:
1. does not work for certain old version of firmware
2. after setting network, BMC have not a fixed time to take the setting and ready to use.

So ``cold reset`` is good here at least for us to control that make BMC is ready for using.
2.

xCAT-genesis-scripts/bin/bmcsetup
whowutwut: Yes, I'm going to try it out when I test the code changes, and see what it's doing..

xCAT-genesis-scripts/bin/bmcsetup
whowutwut: Maybe this TRIES=0 should be up at line 318

xCAT-genesis-scripts/bin/bmcsetup
whowutwut: @daniceexi  Right, FW team says we should not need it....  but from real world testing, there is inconsistency whether the BMC will stay up after "auto-apply" settings, so to ensure that it does come back up, we will do the mc reset cold.  The first sleep 30 is to allow the bmc to autoapply before we cold reset it.

xCAT-genesis-scripts/bin/bmcsetup
whowutwut: I wanted to put this into a function because ideally when the BMC issues are fixed, we can change this to be common code again...

whowutwut:
I've made some additional changes. I think the changes are complete, but I need to start testing this before we merge this pull request.  Please continue to review the changes if possible.  @daniceexi  @zet809   I know there is a lot of differences.

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -14,6 +14,87 @@
...
+                fi
+                tries=$(($tries+1))
+            done
+            TOTAL_SEC=$((${SLEEP_INTERVAL} * ${MAX_ITERATION}))
+            echo "ERROR, After waiting ${TOTAL_SEC} seconds, the BMC is not in a ready state."
+        else
+            # for Non OpenPower servers, just sleep for some set time.
+            sleep 15 
+        fi 
+
```
zet809: The code section below shall be put after line 64. Otherwise it will be run after line 62.

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -220,8 +313,9 @@ fi
 let idev=0
 for b in $BMCVLAN; do
     TRIES=0
+    # Set VLAN for the current channel 
     while ! ipmitool -d $idev lan set $LANCHAN vlan id $b; do
-        sleep 1
+        snooze
```
zet809: It seems when setting BMC ip source, ip address, netmask, gateway and vlan, they are all will sleep 30 seconds. But after everything done, sleep again when resetting BMC. 
Does those sleeping here is necessary?

xCAT-genesis-scripts/bin/bmcsetup
whowutwut: Thanks, yes right! I made a commit to fix this

xCAT-genesis-scripts/bin/bmcsetup
whowutwut: Per the firmware team, yes, for now.  But I am trying to push for a behavior change to more mimic xSeries... 

However,  I want to change this code to have OpenPower and non OpenPower be different code path.  For OpenPower, just run all the commands one after another and sleep 30 one time. 

I don't think we need to look for multiple BMCs in the OpenPower scenario, do we?

whowutwut:
@zet809  I think I've completed the changes, In the last commit, I put all the OpenPower commands together and then just sleep 30...  I hope to find some time tomorrow to test this on the Firestone machines up in the lab.

whowutwut:
After discussion with @daniceexi I undid the last commit and simply added a snooze after the network command complete.  I think now the changes should be completed

daniceexi:
OK, I will merge this change first and @zet809 will run some basic test.

PR #1288, opened 2016-06-08
Remove only xCAT entries from rsyslog.conf file

gurevichmark:
Mark the start and end of the xCAT section in rsyslog.conf file. 
Then, when needed, only remove lines between those 2 markers.


xCAT/postscripts/syslog
```diff
@@ -182,7 +187,8 @@ config_rsyslog_V8()
     # to prevent duplicate logs after "syslog" script is run 
     sed -i '/^[^#].*\/var\/log\/messages/s/^/#/' $conf_file
 
-    echo "# xCAT settings" >> $conf_file
+    # Mark the start of xCAT section
+    echo "# $xCATSettingsSTART" >> $conf_file
     echo "#*.debug   /var/log/messages"  >> $conf_file
```
whowutwut: @neo954  Do you recall why we echo out this statement commented out?

xCAT/postscripts/syslog
```diff
@@ -182,7 +187,8 @@ config_rsyslog_V8()
     # to prevent duplicate logs after "syslog" script is run 
     sed -i '/^[^#].*\/var\/log\/messages/s/^/#/' $conf_file
 
-    echo "# xCAT settings" >> $conf_file
+    # Mark the start of xCAT section
+    echo "# $xCATSettingsSTART" >> $conf_file
```
whowutwut: Should we add a comment in here informing users that changes between the START and END tags will be overwritten by xCAT?

gurevichmark: Added text to let the user know about entries being replaced.

whowutwut:
I'm ok with the changes

gurevichmark:
@whowutwut @daniceexi Do you guys think this needs to be cherry picked into 2.11 ?

whowutwut:
Cherry picked to 2.11 branch:
commit ccb5723c3bae4b7d43b5c9d28872eade806c050b
commit 031892a587266424aa4f780b5d9e9252c4eacb92

Issue #932, opened 2016-04-14
[DEV] some vm commands do not exit if errors are detected.


gurevichmark:
Some vm commands do not exit if errors are detected. For example mkvm without specifying -s option will display an error message and hang. CTRL-C is needed to break out.
Most likely "die" is used to exit from the Perl module instead of returning to the caller with "return".

whowutwut:
Moving the KVM virtualization support/cleanup/enhancements to 2.13 release

samveen:
Another example:
```[samveen@xcatmn1 ~]# mkvm prometheus1 -m kvm1 -s 10G --mem 1024 --cpus 1
Error: kvm plugin bug, pid 57609, process description: 'xCATd SSL: mkvm to prometheus1 for samveen@localhost: kvm instance' with error 'Unable to find any persistent disks at dir:///data/vms/ at /opt/xcat/lib/perl/xCAT_plugin/kvm.pm line 505.
' while trying to fulfill request for the following nodes: prometheus1
(... command doesn't exit ...)
```

whowutwut:
moving this back to 2.12.1 to see if we can fix it before 2.13.

samveen:
@gurevichmark Yeah, your observation about the use of `die(...)` is bang on target:
```
[samveen@catmn1 ~]# tail -n +505 /opt/xcat/lib/perl/xCAT_plugin/kvm.pm |head -n1 
                    die "Unable to find any persistent disks at ".$disk_parts[0];
```

daniceexi:
Looks like ``Sys::Virt`` is not safe to run directly. We should use ``eval`` to run and check the result from ``$@``;

~~~~
eval {  manipulate kvm };

if ($@) {
   # handle the error;
}

whowutwut:
@daniceexi  👍   We need to gracefully return instead of die otherwise the commands will hang and ctrl-c needs to be used to get out of the plugin...

gurevichmark:
Noticed that if XCATBYPASS is set, the `die` exits gracefully, but if XCATBYPASS is not set, the kvm.pm plugin is called through the xcatd daemon. In that case, `die` hangs.
I do not see a value of exiting using a `die` when some error is detected. I think we should just print an error message and exit with `exit 1`. Unlike `die`, exiting with `exit` would not print a line number where error has occurred. But who would care about a line number if we print a good error message ?

daniceexi:
[...]

PR #1289, opened 2016-06-09
Replace some die calls with exit and add undef for global variable

gurevichmark:
This pull request fixes the problem reported by issue #932 .

1. Replace some `die` statements with printing of the message to the screen and `exit`. This will display a more user friendly message in cases where the problem could be the user error.
2. In cases where `die` is more appropriate, add `undef $hypconn` before dying. This will free a global variable `$hyconn` and exit without the hang.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -235,7 +235,6 @@ sub get_storage_pool_by_url {
     if ($@) {
         my $error = $@;
         unless ($error =~ /vgcreate.*exit status 3/ or $error =~ /pvcreate.*exit status 5/) {
-            undef $hypconn;
             die $@;
         }
     }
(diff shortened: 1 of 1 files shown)


whowutwut:
On the places where you still keep the die around, does the program exit but now does not have the "Plugin bug" message.. or it just does not hang anymore?

gurevichmark:
The places where I left `die` appear to be less of user error and more of something we did not expect to happen. In those cases the error will still be "Plugin bug", it just will not hang.

daniceexi:
@gurevichmark 
I don't have time to investigate when the kvm.pm started to use the ``CAT::SvrUtils::sendmsg`` to display message, but it was NOT the correct way. Since all the ``CAT::SvrUtils::sendmsg`` will call the same ``$callback`` to pass message to xcatclient, if there are multiple kvmhost processes are forked, there might be conflict.

The correct way is to use the return like following to pass message back.
~~~~
1079         return (1, "Unable to find current location of $node");
~~~~

And please check if using ``return``, do we still need to ``undef $hypconn;``?

gurevichmark:
Changed pull request to use `return` instead of the `exit` to return error message and exit without the hang. As the result, caller of the subroutine had to change to detect the error condition and do its own `return`. I also removed the `undef $hypconn` statements that I added in the previous merge, because I do not know how to test all those code paths.

Therefore this pull request only fixes the problem reported by the issue #932. If other issues about `die` hangs from kvm.pm get reported, they will be fixed under a different pull request.

gurevichmark:
@daniceexi Can you please review these latest changes.

daniceexi:

PR #1291, opened 2016-06-10
Removing the unused xCAT-nbroot* files 

whowutwut:
The xCAT-nbroot has been replaced by xCAT-genesis kernel. 
This pull request is cleaning up old files that are unused


whowutwut:
@daniceexi @gurevichmark @zet809  I created this pull request to remove the nbroot directory from the source tree.  I don't think we ship any references to it and it causes confusion when we are searching through the code.    please help review.

gurevichmark:
I have not noticed any problems.

daniceexi:
The change looks good for me.

PR #1297, opened 2016-06-13
Add uuid into /opt/xcat/xcatinfo

cxhong:
Add uuid to /opt/xcat/xcatinfo into the rootimg dir during packimage, so customer can check uuid in the compute node to verify if osimage matches with the one on the MN.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/packimage.pm
+++ b/xCAT-server/lib/xcat/plugins/packimage.pm
@@ -181,8 +181,12 @@ sub process_request {
     system("mkdir -p $rootimg_dir/xcatpost");
     system("cp -r $installroot/postscripts/* $rootimg_dir/xcatpost/");
 
-    #put the image name and timestamp into diskless image when it is packed.
+    #put the image name, uuid and timestamp into diskless image when it is packed.
     `echo IMAGENAME="'$imagename'" > $rootimg_dir/opt/xcat/xcatinfo`;
+
+    my $uuid = `uuidgen`;
+    chomp $uuid;
+    `echo IMAGEUUID="'$uuid'" >> $rootimg_dir/opt/xcat/xcatinfo`;
     
     my $timestamp = `date`;
     chomp $timestamp;
--- a/xCAT-server/share/xcat/install/scripts/post.debian
+++ b/xCAT-server/share/xcat/install/scripts/post.debian
@@ -89,7 +89,12 @@ do
                 mkdir -p /opt/xcat
                 touch /opt/xcat/xcatinfo
             fi
-            echo "XCATSERVER=$i" > /opt/xcat/xcatinfo
+            grep 'XCATSERVER' /opt/xcat/xcatinfo 2>&1 > /dev/null
+            if [ $? -eq 0 ]; then
+                sed -i "s/XCATSERVER=.*/XCATSERVER=$i/" /opt/xcat/xcatinfo
+            else
+                echo "XCATSERVER=$i" >> /opt/xcat/xcatinfo
+            fi
 	    break
 	fi
 	RAND=$(perl -e 'print int(rand(5)). "\n"')
--- a/xCAT-server/share/xcat/install/scripts/post.xcat
+++ b/xCAT-server/share/xcat/install/scripts/post.xcat
@@ -47,8 +47,21 @@ if [ ! -f /opt/xcat/xcatinfo ]; then
     mkdir -p /opt/xcat
     touch /opt/xcat/xcatinfo
 fi
-echo "XCATSERVER=$MASTER_IP" > /opt/xcat/xcatinfo
-echo "INSTALLDIR=$INSTALLDIR" >> /opt/xcat/xcatinfo
+
+grep 'XCATSERVER' /opt/xcat/xcatinfo 2>&1 > /dev/null
+if [ $? -eq 0 ]; then
+    sed -i "s/XCATSERVER=.*/XCATSERVER=$MASTER_IP/" /opt/xcat/xcatinfo
+else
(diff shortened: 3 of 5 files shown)


whowutwut:
@cxhong can you show an example of the new file `/opt/xcat/xcatinfo` ?

cxhong:
Here is a example on a VM:
``````
 # cat /opt/xcat/xcatinfo
IMAGENAME='sles12-ppc64le-netboot-compute'
IMAGEUUID='0672ba43-1f47-4094-afb0-a75c4d6ecf12'
TIMESTAMP='Mon Jun 13 13:53:05 EDT 2016'
NODE=fs2vm109
XCATSERVER='10.3.31.101'
USEFLOWCONTROL=NO
``````

**Comment by immarvin** [comment]
hi @cxhong , since this feature will only affect diskless and statelite, why the "post.xcat", "post.debian" and "xcatinstallpost"  need to be modified?  these files are for diskful installation

**Comment by cxhong** [comment]
I changed those files because they used "> /opt/xcat/xcatinfo " and will create new file, and any old info will be removed. We just want to update that file, right? Or do you think we should not keep the old infos in the /opt/xcat/xcatinfo, every time we run into xcat.post, xcat.debian and xcatinstallpost , the file needs to recreate?

**Comment by whowutwut** [comment]
One comment is that the above output of `xcatinfo` is not easy to search for.  

I think the main object here is to quickly identify whether the diskless images are identical in a large cluster.  If we have the 3 ID fields in different lines, it makes it difficult to have  single query....  Then if we want to add more to this string, it requires all the scripts to grep for the new keyword.  

I would suggest the following format...

```
IMAGEID=sles12-ppc64le-netboot-compute|0672ba43-1f47-4094-afb0-a75c4d6ecf12|Mon Jun 13 13:53:05 EDT 2016
NODE=fs2vm109
XCATSERVER='10.3.31.101'
USEFLOWCONTROL=NO
```

This gives the admin much more flexibility to have a single xdsh query to the compute nodes and decide which information in the ID string is important to them. 

To look for the uuid for the image: 
```
[vhu@victorvm7 id]$ cat xcatinfo2 | grep IMAGEID | cut -d= -f2 | cut -d'|' -f2
0672ba43-1f47-4094-afb0-a75c4d6ecf12
```

To look for the image name:
```
[vhu@victorvm7 id]$ cat xcatinfo2 | grep IMAGEID | cut -d= -f2 | cut -d'|' -f1
sles12-ppc64le-netboot-compute
```

daniceexi:
@whowutwut Three lines have better format to edit and read. If necessary, user can use xdsh to cat all the files and use 'grep' to search what he wanted. Especially, if user wants to identify an osimage, only the ``UUID`` is necessary.

cxhong:
Admin/User can use xcatprob that Mark will implement it .

whowutwut:
@daniceexi  ok, I'm ok with it.

daniceexi:
Thanks. The change looks good. I'll merge it.

immarvin:
hi @cxhong , post.xcat and post.debian will only be run in the post installation section of the diskful installation,  the "/opt/xcat/xcatinfo" does not exist when these scripts are run.  

The script "xcatinstallpost" will only be run on 1st reboot after installation, the "/opt/xcat/xcatinfo"  file exists  but the line "REBOOT=TRUE" does not exist.

in post.xcat and post.debian, when the following code slice is run
````
 +grep 'XCATSERVER' /opt/xcat/xcatinfo 2>&1 > /dev/null
+if [ $? -eq 0 ]; then
+    sed -i "s/XCATSERVER=.*/XCATSERVER=$MASTER_IP/" /opt/xcat/xcatinfo
+else
+    echo "XCATSERVER=$MASTER_IP" >> /opt/xcat/xcatinfo
+fi
````
since /opt/xcat/xcatinfo does not exist,   an error message like 
````
# grep 'REBOOT' /opt/xcat/xcatinfo 2>&1 > /dev/null 
grep: /opt/xcat/xcatinfo: No such file or directory
````
will be left in the /var/log/xcat/xcat.log, which might be considered as an error.

 ``grep 'XCATSERVER' /opt/xcat/xcatinfo 2>&1 > /dev/null`` should be ``grep 'XCATSERVER' /opt/xcat/xcatinfo  >/dev/null 2>&1``

I suggest not to modify post.xcat, post.debian and xcatinstallpost


I do not think we should modify them.

cxhong:
@immarvin ,  file /opt/xcat/xcatinfo doest not exist in the post.xcat and post.debian , but it will created  before we do **echo** to this file.
''''''
 #save the master to /opt/xcat/xcatinfo file
            if [ ! -f /opt/xcat/xcatinfo ]; then
                mkdir -p /opt/xcat
                touch /opt/xcat/xcatinfo
            fi
            echo "XCATSERVER=$i" >> /opt/xcat/xcatinfo
''''''''

for xcatinstallpost,  the file is exists,  it will not hit the ````file is not exist``` message as you mentioned.

immarvin:
hi @cxhong , 

``grep 'XCATSERVER' /opt/xcat/xcatinfo 2>&1 > /dev/null`` will not redirect the STDERR to /dev/null as expect, it should be ``grep 'XCATSERVER' /opt/xcat/xcatinfo >/dev/null 2>&1``.

whowutwut:
@immarvin  Good point that you make .  I think there's 2 issues here...

1. The way the command is written `2>&1 > /dev/null` does not work as expected because stderr is not redirected to `/dev/null` . 
2. Should we be redirecting stderr to `/dev/null`? 

For those messages that REALLY want stderr to redirect to `/dev/null`, then we should code like you suggest ` > /dev/null 2>&1` but I think really we DO not want to mask stderr.  The case where we will see the error message is if we have a coding issue.  I would think that we would want the stderr messages to appear so that we can easily fix the code problems.

PR #1299, opened 2016-06-14
fix  "xcat-server can not be upgraded #1295"

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1295

the command,
````
apt-get upgrade xcat
````
the argument "xcat" is meaningless, in fact all the packages with upgrading candidates will be upgraded, this might not be what user want to do.

it need to be replaced with ``apt-get --only-upgrade install xcat*``


daniceexi:

daniceexi:
@neo954 Please refer to this change when implement the ``installxcat`` script.

Issue #1298, opened 2016-06-14
[DEV] dohyp subroutine in kvm can not work with runxcmd 


chenglch:
I encounter this problem when I prepare to enhance getadapter for kvm which call the runxcmd subroutine to run rpower command.

        main::__ANON__('Can\'t use string ("0") as an ARRAY ref while "strict refs" i...') called at    /opt/xcat/lib/perl/xCAT/Utils.pm line 1383
        xCAT::Utils::runxcmd_output('HASH(0x2ed8780)') called at /opt/xcat/lib/perl/xCAT_plugin/kvm.pm line 3878
        xCAT_plugin::kvm::forward_data('CODE(0x10216b0)', 'IO::Select=ARRAY(0x2ed8288)', 'HASH(0x2ed83c0)') called at /opt/xcat/lib/perl/xCAT_plugin/kvm.pm line 3788
        xCAT_plugin::kvm::process_request('HASH(0x2ea9dd0)', 'CODE(0x10216b0)', 'CODE(0x23d3ce0)') called at /opt/xcat/sbin/xcatd line 2099

https://github.com/xcat2/xcat-core/blob/master/xCAT-server/lib/xcat/plugins/kvm.pm#L3990

            $text =~ s/^\s+//;
            $text =~ s/\s+$//;
            $output{node}->[0]->{errorcode} = $rc;
            $output{node}->[0]->{name}->[0] = $node;

But the subroutine which process the callback message is 
https://github.com/xcat2/xcat-core/blob/master/perl-xCAT/xCAT/Utils.pm#L1383

    sub runxcmd_output {
      .........
     if (defined($node->{errorcode}))
        {
            if (ref(\($node->{errorcode}->[0])) eq 'SCALAR')
            {
                $::RUNCMD_RC |=  $node->{errorcode}->[0];
            }
        }
       .....

Maybe change the below code in kvm.pm

    $output{node}->[0]->{errorcode} = $rc;  -->   $output{node}->[0]->{errorcode} = [$rc]; 

could solve this issue

PR #1301, opened 2016-06-14
Change the ref type of errorcode in xcmdoutput to array

chenglch:
Currently runxcmd can not work with kvm when running rpower command as
`runxcmd_output` subroutine just apply array type for `node->{errorcode}`.

close-issue: #1298


daniceexi:

Issue #1217, opened 2016-05-26
[Customer] xCAT 2.7.10 command problems on AIX 7.1.3 with openssl 1.0.1.515 with Efix IV83169m9a


billwaj:
xCAT commands rpower, rinv, rvitals fail on xCAT 2.7.10 and AIX 7.1.3 with openssl 1.0.1.515 efix IV83169m9a on Power7 management node and targeting a Power7 CEC:

```
Can't load '/usr/opt/perl5/lib/site_perl/5.10.1/aix-thread-multi/auto/Crypt/SSLeay/SSLeay.so' for module Crypt::SSLeay:         0509-130 Symbol resolution failed for /usr/opt/perl5/lib/site_perl/5.10.1/aix-thread-multi/auto/Crypt/SSLeay/SSLeay.so because:
        0509-136   Symbol SSLv2_client_method (number 59) is not exported from
                   dependent module /usr/lib/libssl.a(libssl.so.0.9.8).
        0509-022 Cannot load module /usr/opt/perl5/lib/site_perl/5.10.1/aix-thread-multi/auto/Crypt/SSLeay/SSLeay.so.
        0509-026 System error: Cannot run a file that does not have a valid format.
        0509-192 Examine .loader section symbols with the
                 'dump -Tv' command. at /usr/opt/perl5/lib/5.10.1/aix-thread-multi/DynaLoader.pm line 200, <$readpipe> line 735.
 at /usr/opt/perl5/lib/site_perl/5.10.1/aix-thread-multi/Net/SSL.pm line 16
Compilation failed in require at /usr/opt/perl5/lib/site_perl/5.10.1/aix-thread-multi/Net/SSL.pm line 16, <$readpipe> line 735.
Compilation failed in require at /opt/xcat/lib/perl/xCAT_plugin/fsp.pm line 58, <$readpipe> line 735.
```

Similarly, the renergy command fails with above error also.

billwaj:
### ###

zet809:
hi, @billwaj, what is the status of this issue right now? Can we close it? Thx!

PR #1306, opened 2016-06-14
Removed require Net::SSL so that LWP.pm will default to use IO::Socket::SSL

billwaj:
This pull request is to fix issue #1217 where Net::SSL will fail to load SSLeay.so because libssl.a no longer supports SSLv2.  The fix is to not use Net::SSL and default to use IO::Socket::SSL which works.

### DIFF
--- a/perl-xCAT/xCAT/PPCcfg.pm
+++ b/perl-xCAT/xCAT/PPCcfg.pm
@@ -1352,8 +1352,8 @@ sub connect {
              return( "Unable to redirect STDERR: $!" );
         }
     }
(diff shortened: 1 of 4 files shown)


daniceexi:
In these files, xCAT requires ``require Net::SSL`` for web access against the ASMI. But now we have deprecated all the ASMI direct access features, so these ``require Net::SSL`` sentences can be removed.

daniceexi:
I have reviewed the change and agree to merge. Thanks @billwaj .

daniceexi:
I have merged this commit to 2.9.

Issue #1106, opened 2016-05-13
[FVT]2.12:switchdiscover return wrong information using snmp


junxiawang:
build:
[root@c910f02c01p13 tmp]# lsdef -v
lsdef - Version 2.12 (git commit 56cfba70babd342f63b17932b115f21040245033, built Thu May 12 09:30:47 EDT 2016)
env:
rhels7.2
How to reproduce:
```
[root@c910f02c01p13 tmp]# switchdiscover --range 10.2.3.0/24,192.168.3.0/24,11.5.6.7 -s snmp
Discovering switches using snmpwalk.....
 No switch found 
[root@c910f02c01p13 tmp]# switchdiscover --range 10.2.3.0/24,192.168.3.0/24,11.5.6.7
Discovering switches using nmap for 10.2.3.0/24 192.168.3.0/24 11.5.6.7. It may take long time...
ip          	name                	vendor                                            	mac         
------------	------------        	------------                                      	------------
10.2.3.1    	c910f02c03v01       	IBM                                               	00:21:5E:34:4E:2C

-----------------------------> could not switch info using -s snmp 
```

junxiawang:
[root@c910f02c01p13 ~]# switchdiscover --range 10.5.24.1  -s snmp
Discovering switches using snmpwalk.....
Could not process this command: /usr/bin/nmap -P0 -v -sU -p 161 -oA snmp_scan 10.5.24.1 | grep 'open port 161' 
 No switch found 
[root@c910f02c01p13 ~]#  switchdiscover --range 10.5.24.1
Discovering switches using nmap for 10.5.24.1. It may take long time...
ip          	name                	vendor                                            	mac         
------------	------------        	------------                                      	------------
10.5.24.1   	switch-10-5-24-1    	Juniper Networks                                  	5C:5E:AB:64:CA:7F
[root@c910f02c01p13 ~]# switchdiscover --range 10.5.24.1  -s snmp
Discovering switches using snmpwalk.....
Could not process this command: /usr/bin/nmap -P0 -v -sU -p 161 -oA snmp_scan 10.5.24.1 | grep 'open port 161' 
 No switch found 
[root@c910f02c01p13 ~]#

junxiawang:
10.2.3.1        c910f02c03v01           IBM
is viso server not switch

daniceexi:
@cxhong The issue cause by the change in the commit 4488c24a300001dc7f9abbc476474a581982595a.

[...]

Issue #1038, opened 2016-05-04
switchdiscover without any options scans too many networks


whowutwut:
On my management node, there's 3 networks defined
```
2: enP4p1s0f0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP qlen 1000
    inet 10.3.29.2/8 brd 10.255.255.255 scope global enP4p1s0f0
    inet6 fe80::6eae:8bff:fe02:4390/64 scope link 
3: enP4p1s0f1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP qlen 1000
    inet 50.3.29.1/8 brd 50.255.255.255 scope global enP4p1s0f1
    inet6 fe80::6eae:8bff:fe02:4391/64 scope link 
4: enP4p1s0f2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP qlen 1000
    inet 192.168.3.29/16 brd 192.168.255.255 scope global enP4p1s0f2
    inet6 fe80::6eae:8bff:fe02:4392/64 scope link 
```

However, in xCAT, I don't want xCAT to do anything with the 10 network , so I've disabled it in the networks table

```
[root@fs1 etc]# tabdump networks
#netname,net,mask,mgtifname,gateway,dhcpserver,tftpserver,nameservers,ntpservers,logservers,dynamicrange,staticrange,staticrangeincrement,nodehostname,ddnsdomain,vlanid,domain,comments,disable
"10_0_0_0-255_0_0_0","10.0.0.0","255.0.0.0","enP4p1s0f0","10.0.0.101",,"10.3.29.2",,,,,,,,,,,,"1"
"50_0_0_0-255_0_0_0","50.0.0.0","255.0.0.0","enP4p1s0f1","<xcatmaster>",,"50.3.29.1",,,,,,,,,,,,
"192_168_0_0-255_255_0_0","192.168.0.0","255.255.0.0","enP4p1s0f2","<xcatmaster>",,"192.168.3.29",,,,"192.168.5.150-192.168.5.170",,,,,,,,
```

but when I run `switchdiscover` without any options to get the help menu, it instead tried to scan all my networks. 

```
[root@fs1 etc]# switchdiscover
Discovering switches using nmap for 50.0.0.0/8 192.168.0.0/16 10.0.0.0/8. It may take long time...
You can modify the --range parameters to cut down the time.

^C
```

[...]

PR #1307, opened 2016-06-14
Add new subroutine and filter out disable networks for switchdiscover

cxhong:
This PR covers issue #1106 and issue #1038 
1) Process osscan-guess to find vendor info  if vendor info showed as "IBM" (or other than global_switch_type defined in the code)
2) created new subroutine nmap_osguess() so we can run nmap osscan-guess for multiple ips instead of one at a time.
3) only run switchdiscover for not disable networks if there is no input from --range options.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/switchdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/switchdiscover.pm
@@ -14,6 +14,7 @@ use xCAT::NodeRange;
 use xCAT::NetworkUtils;
 use xCAT::Utils;
 use xCAT::SvrUtils;
+use xCAT::Table;
 use XML::Simple;
 no strict;
 use Data::Dumper;
@@ -33,12 +34,13 @@ my %global_switch_type = (
     Juniper => "Juniper",
     juniper => "Juniper",
     Cisco => "Cisco",
+    cisco => "Cisco",
     BNT => "BNT",
+    Blade => "BNT",
     Mellanox => "Mellanox",
     mellanox => "Mellanox",
     MLNX => "Mellanox",
-    MELLAN => "Mellanox",
-    IBM  => "BNT"
+    MELLAN => "Mellanox"
 );
 
 
@@ -683,6 +685,7 @@ sub nmap_scan {
     my $switches;
     my $found;
     my $counter=0;
+    my $osguess_ips=[];
     if ($result_ref) {
         if (exists($result_ref->{host})) {
             my $host_ref = $result_ref->{host};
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -716,77 +719,115 @@ sub nmap_scan {
...
+                    if ( ( $line =~ /switch/) || ( $line =~ /router/) ) {
+                        $isswitch=1;
+                    } else {
+                        last;
+                    }  
+                }
+                my $search_string = join '|', keys(%global_switch_type);
+                if ($line =~ /Running/) {
+                    if ($line =~ /($search_string)/){
+                        $vendor = $1;
```
daniceexi: Do we need to set ``$isswitch=1`` here?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -716,77 +719,115 @@ sub nmap_scan {
...
+                if ($line =~ /Aggressive OS/) {
+                    if ($line =~ /($search_string)/){
+                        $vendor = $1;
+                        $isswitch=1;
+                    }
+                }
+                
+            }
+            if ($isswitch == 1) {
+                $switches->{$mac}->{ip} = $ip;
```
daniceexi: what if ``ip`` or ``vendor`` is blank?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -1164,13 +1202,42 @@ sub get_ip_ranges {
...
+    # my_nets will return all the networks defined in the networks 
+    # table, we need to exclude disable networks here.
     my $nets = xCAT::NetworkUtils->my_nets();
     my $ranges=[];
+    my $nettab = xCAT::Table->new('networks');
+    my $disable_nets=[];
+    if ($nettab) {
+        my $netents = $nettab->getAllEntries("all");
+        foreach (@$netents) {
+            if ($_->{disable} =~ /1|YES|Yes|yes|Y|n/) {
```
daniceexi: why check the ``|n/``  in line ``if ($_->{disable} =~ /1|YES|Yes|yes|Y|n/) {``? ``n`` means NO.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -1164,13 +1202,42 @@ sub get_ip_ranges {
...
+                push(@$disable_nets, $_->{'net'});
+                if (exists($globalopt{verbose}))    {
+                    send_msg($request, 1, "network is disabled for  $_->{'net'}, $_->{'mask'}");
+                }
+            }
+        }
+    }
```
daniceexi: I am concerning the algorithm that exclude the 'disabled' network in the ``networks`` table.

Generally, user only put the useful network in the ``networks`` table. So I suggest that we only scan the networks that enabled in the ``networks`` table.

Could you refer to the ``lsdef`` command?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: somehow this output is a little different depends version of nmap, OS or switch
``````
Device type:
Running (JUST GUESSING):
Aggressive OS guesses:
`````
for above three "keywords" line, sometime has two of three, sometime has one of three, the results are inconsistent.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: ip address will be always here.  if vendor is blank, but Device type showed as switch, we will consider this as switch.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: oh, right, i will change that

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: not sure how can I check if network is disabled via other method.

daniceexi: In fact, the call ``$nettab->getAllEntries()`` without argument will only return the enabled networks.

My question is if certain MN has 10 nics, but only defined the install_network and service_network in the networks table, do we want to discover against the 10 nics or just the two networks which have been defined in the networks table? My idea is we just discover the enabled networks which defined in the networks table.

cxhong:
@daniceexi , I agree we should get enabled networks from networks table.  Modify the code, can you review it again?

daniceexi:
I have reviewed the code change and agree to merge. Thanks @cxhong

PR #1313, opened 2016-06-15
issue 1191: do retry when update dns record replies NOTAUTH

daniceexi:


whowutwut:
@bybai  Can you provide a patch file for this change for xCAT 2.9.1? 
FYI @cxhong

bybai:
Hi @whowutwut ,
I do not find upstream/2.9.1 from git. 
Another thing, the 'NOAUTH' issue from ESS can work well now. It is usage issue, not caused by code. I do not know they want to change their code or not.

zet809:
@bybai What @whowutwut meaning is the upstream/2.9 branch. It shall be 2.9.4 right now.

whowutwut:
@bybai Thanks, I see that in the email exchanges.  @zet809  ESS is using xCAT 2.9.1 version, so we cannot just apply the patch to 2.9.4 and then give them a patch file.. unless there are no conflicts..  But for now, I think we don't need to do anything here.

Issue #1304, opened 2016-06-14
confignics -s failed, make the computer node can not ping 


wwzhubj:
I use updatenode  node "confignics -s" to set my installnic to static mode. But updatenode hanged and the computer node can not ping and ssh. 
when rcons it, we see in the config file for installnic, the netmask is 0.0.0.0, which is wrong. 

In my env:  route table is : 
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         v2199gw9-114-39 0.0.0.0         UG    0      0        0 eth0
9.114.39.0      *               255.255.255.0   U     0      0        0 eth0
10.128.128.0    *               255.255.255.0   U     0      0        0 eth1
link-local      *               255.255.0.0     U     1000   0        0 eth1

but in xCAT postscript configeth  476 line, will set the "0.0.0.0" as the netmask into network config file. 
str_inst_mask=`route |grep  ${inst_ip_pre}|awk '{print $3}'|head -1`

The issue can be fixed by change it like below:
str_inst_mask=`route |grep -E ^${inst_ip_pre}|awk '{print $3}'|head -1`

daniceexi:
In @wwzhubj 's env, the  ``${inst_ip_pre} = 9.114.39.0``, when run the ``route |grep ${inst_ip_pre}|awk '{print $3}'|head -1``, the first line ``default v2199gw9-114-39 0.0.0.0 UG 0 0 0 eth0`` is matched, then it get the incorrect value. I suggest to use the ``^`` to match the route entry from beginning.

daniceexi:
The fix has been verified by @wwzhubj, so close it.

PR #1315, opened 2016-06-16
issue 1304: fix the issue confignics -s cannot get the correct netmas…

daniceexi:
…k for installnic


daniceexi:
This is for issue #1304

immarvin:
agree to merge

Issue #1204, opened 2016-05-24
[Customer] /xcatpost not updated


dombrowa:
I have found the following problem.
 When running a postscript which is supposed to source a file containing bash functions it fails on some nodes
fstgb019: ./custom/fstgb-rhel7.2-sf-c/postscripts.test/update-mellanox: line 27: source: ./custom/fstgb-rhel7.2-sf-c/postscripts.test/functions: is a directory
For unknown reason some nodes show a directory functions instead of a file named functions - This caused failure to execute the postscript.

```
[root@dccxcat2 ~]# xdsh fstgb-compute ls -ald /xcatpost/custom/fstgb-rhel7.2-sf-c/postscripts.test/functions
fstgb009: drwxr-xr-x 2 root root 6 Dec 11 16:35 /xcatpost/custom/fstgb-rhel7.2-sf-c/postscripts.test/functions
fstgb007: drwxr-xr-x 2 root root 6 Dec 11 16:34 /xcatpost/custom/fstgb-rhel7.2-sf-c/postscripts.test/functions
fstgb008: drwxr-xr-x 2 root root 6 Dec 11 16:35 /xcatpost/custom/fstgb-rhel7.2-sf-c/postscripts.test/functions
fstgb001: -rwxr-xr-x 1 root root 1768 Dec 12 08:26 /xcatpost/custom/fstgb-rhel7.2-sf-c/postscripts.test/functions
fstgb019: drwxr-xr-x 2 root root 6 Dec 11 16:38 /xcatpost/custom/fstgb-rhel7.2-sf-c/postscripts.test/functions
fstgb010: drwxr-xr-x 2 root root 6 Dec 11 16:35 /xcatpost/custom/fstgb-rhel7.2-sf-c/postscripts.test/functions
fstgb015: drwxr-xr-x 2 root root 6 Dec 11 16:35 /xcatpost/custom/fstgb-rhel7.2-sf-c/postscripts.test/functions
fstgb011: drwxr-xr-x 2 root root 6 Dec 11 16:35 /xcatpost/custom/fstgb-rhel7.2-sf-c/postscripts.test/functions
fstgb012: drwxr-xr-x 2 root root 6 Dec 11 16:35 /xcatpost/custom/fstgb-rhel7.2-sf-c/postscripts.test/functions
fstgb018: drwxr-xr-x 2 root root 6 Dec 11 16:35 /xcatpost/custom/fstgb-rhel7.2-sf-c/postscripts.test/functions
[...]

PR #1316, opened 2016-06-16
Fix issue 1204, rm /xcatpost folder before wget postscripts

xuweibj:
#1204 

When updatenode, delete /xcatpost folder and files under it before wget postscripts from MN.


daniceexi:
the change looks good for me.

Issue #1308, opened 2016-06-15
[customer] lsdef and Rest API issues with regex on nicks


arif-ali:
Just have seen an issue with the nickips, when we have regex

```
"virt","|eno1!10.28.9.($1+10),ipmi0!10.28.25.($1+10),bond0!10.28.41.($1+10)|","bond0!-10g",,,,,,"bond0!MTU=9000","bond0!ens1f0|ens1f1",,,
```

if you look at the outputs below, we have |eno1 as the interface name, which doesn't quite work out. Not sure the best resolution to this though. But some idea on resolving this would be great

```
[root@admin001 ~]# lsdef -t group virt
Object name: virt
    ip=|10.28.9.($1+10)|
    members=virt000,virt001,virt002,virt003,virt004,virt005,virt006,virt007
    nicdevices.bond0=ens1f0|ens1f1
    nicextraparams.bond0=MTU=9000
    nichostnamesuffixes.bond0=-10g
    nicips.|eno1=10.28.9.($1+10)
    nicips.ipmi0=10.28.25.($1+10)
    nicips.bond0=10.28.41.($1+10)|
    postbootscripts=configbond bond0 ens1f0@ens1f1,ocf/setup_extnw -n virt_10g
    provmethod=base-image.virt
[root@admin001 ~]# curl -k "https://admin001.<snip>/xcatws/groups/virt?userName=root&userPW=<snip>&pretty=1"
{
   "virt":{
      "provmethod":"base-image.virt",
      "nicips.bond0":"10.28.41.($1+10)|",
      "ip":"|10.28.9.($1+10)|",
      "members":"virt000,virt001,virt002,virt003,virt004,virt005,virt006,virt007",
      "nicextraparams.bond0":"MTU=9000",
      "nicips.ipmi0":"10.28.25.($1+10)",
      "nicips.|eno1":"10.28.9.($1+10)",
      "postbootscripts":"configbond bond0 ens1f0@ens1f1,ocf/setup_extnw -n virt_10g",
      "nicdevices.bond0":"ens1f0|ens1f1",
      "nichostnamesuffixes.bond0":"-10g"
   }
}
```

daniceexi:
@arif-ali 
One thing to confirm is that the function works fine, but the ``lsdef/chdef`` have problem to list/change the group for nics table, am I right?

A potential solution is:
1. In the ``lsdef``, remove the ``^|`` and ``|$`` before displaying
[...]

PR #1317, opened 2016-06-16
issue1308 lsdef and Rest API issues with regex on nicks

bybai:
issue #1308   lsdef and Rest API issues with regex on nicks


daniceexi:
@bybai We also have to handle the following case. My idea is if the value has ``|`` at beginning and end, it means the ``regular expression``. 

* If no | in middle the value, remove the leading and end |;
* if has | in middle, remove the all the string between the first and second |.

Also people add enough comment to mention this is just a workaround to make the lsdef output looks better.

~~~~
"reg","|test(\d+)|eth0!192.186.0.($1)|"
~~~~

daniceexi:

PR #1318, opened 2016-06-16
go-xcat - A tool to install xCAT automatically

neo954:
Usage: go-xcat [OPTION]... [ACTION]
    Install xCAT automatically

    Options:
    Mandatory arguments to long options are mandatory for short options too.
      -h, --help                    display this help and exit
      --xcat-core=[URL]             use a different URL or path for the xcat-core
                                    repository
      --xcat-dep=[URL]              use a different URL or path for the xcat-dep
                                    repository
      -x, --xcat-version=[VERSION]  specify the version of xCAT; cannot use with
                                    --xcat-core
      -y, --yes                     answer yes for all questions

    Actions:
      install                       installs all the latest versions of xcat-core
                                    and xcat-dep packages from the repository

    Examples:
      go-xcat
      go-xcat install
      go-xcat --yes install
      go-xcat -x 2.12 -y install
      go-xcat --xcat-version=devel install
      go-xcat --xcat-core=/path/to/xcat-core.repo install
      go-xcat --xcat-core=/path/to/xcat-core install
      go-xcat --xcat-core=/path/to/xcat-core.tar install
      go-xcat --xcat-core=/path/to/xcat-core.tar.Z install
      go-xcat --xcat-core=/path/to/xcat-core.tar.gz install
      go-xcat --xcat-core=/path/to/xcat-core.tar.bz2 install
      go-xcat --xcat-core=/path/to/xcat-core.tar.xz install
      go-xcat --xcat-core=http://xcat.org/path/to/xcat-core.repo install
      go-xcat --xcat-core=http://xcat.org/path/to/xcat-core install
      go-xcat --xcat-core=http://xcat.org/path/to/xcat-core.tar.bz2 install
      go-xcat --xcat-core=/path/to/xcat-core.repo \
              --xcat-dep=/path/to/xcat-dep.repo install
      go-xcat --xcat-core=/path/to/xcat-core \
              --xcat-dep=/path/to/xcat-dep install
      go-xcat --xcat-core=/path/to/xcat-core.tar.bz2 \
              --xcat-dep=/path/to/xcat-dep.tar.bz2 install
      go-xcat --xcat-core=http://xcat.org/path/to/xcat-core.repo \
              --xcat-dep=http://xcat.org/path/to/xcat-dep.repo install
      go-xcat --xcat-core=http://xcat.org/path/to/xcat-core \
              --xcat-dep=http://xcat.org/path/to/xcat-dep install
      go-xcat --xcat-core=http://xcat.org/path/to/xcat-core.tar.bz2 \
              --xcat-dep=http://xcat.org/path/to/xcat-dep.tar.bz2 install

    xCAT (Extreme Cloud/Cluster Administration Toolkit): <http://xcat.org/>
    Full documentation at: <http://xcat-docs.readthedocs.io/en/stable/>

### DIFF
--- a/xCAT-server/share/xcat/tools/go-xcat
+++ b/xCAT-server/share/xcat/tools/go-xcat
@@ -8,21 +8,22 @@ function usage()
 	local script="${0##*/}"
 
 	while read ; do echo "${REPLY}" ; done <<-EOF
-	Usage: ${script} [OPTION]...
+	Usage: ${script} [OPTION]... [ACTION]
 	Install xCAT automatically
 
 	Options:
 	  --help                        display this help and exit
-	  --install                     installs all the latest versions from the
-	                                repository
 	  --xcat-version=[VERSION]      specify the version of xCAT
 	  -y, --yes                     answer yes for all questions
 
+	Actions:
+	  install                       installs all the latest versions from the
+	                                repository
 	Examples:
 	  ${script}
-	  ${script} --install
-	  ${script} --install --yes
-	  ${script} --install --xcat-version=2.12 --yes
+	  ${script} install
+	  ${script} --yes install
+	  ${script} --xcat-version=2.12 --yes install
 	EOF
 }
 
@@ -256,8 +257,8 @@ function check_linux_distro()
 {
 	local distro="$(source /etc/os-release >/dev/null 2>&1 &&
 		echo "${ID}")"
-	[[ -z "${distro}" ]] && [[ -f /etc/redhat-release ]] && distro="rhel"
-	[[ -z "${distro}" ]] && [[ -f /etc/SuSE-release ]] && distro="sles"
+	[[ -z "${distro}" && -f /etc/redhat-release ]] && distro="rhel"
+	[[ -z "${distro}" && -f /etc/SuSE-release ]] && distro="sles"
 	echo "${distro}"
 }
 
@@ -265,9 +266,9 @@ function check_linux_version()
 {
 	local ver="$(source /etc/os-release >/dev/null 2>&1 &&
 		echo "${VERSION_ID}")"
-	[[ -z "${ver}" ]] && [[ -f /etc/redhat-release ]] &&
+	[[ -z "${ver}" && -f /etc/redhat-release ]] &&
 		ver="$(awk '{ print $(NF - 1) }' /etc/redhat-release)"
-	[[ -z "${ver}" ]] && [[ -f /etc/SuSE-release ]] &&
+	[[ -z "${ver}" && -f /etc/SuSE-release ]] &&
 		ver="$(awk '/VERSION/ { print $NF }' /etc/SuSE-release)"
 	echo "${ver}"
 }
@@ -292,7 +293,7 @@ function function_dispatch()
 # $@    package names
 function check_package_version_rpm()
 {
-	! type rpm >/dev/null 2>&1 && return 255
+	type rpm >/dev/null 2>&1 || return 255
 	local ver=""
 	while read ver
 	do
@@ -302,14 +303,14 @@ function check_package_version_rpm()
 		else
 			echo "${ver}"
 		fi
-	done < <(rpm -q --qf '%{version}-%{release}\n' "$@")
+	done < <(rpm -q --qf '%{version}-%{release}\n' "$@" 2>/dev/null)
 	return 0
 }
 
 # $@    package names
 function check_package_version_deb()
 {
-	! type dpkg-query >/dev/null 2>&1 && return 255
+	type dpkg-query >/dev/null 2>&1 || return 255
 	local name=""
 	local ver=""
 	while read name ver
@@ -344,7 +345,7 @@ function check_package_version()
 # $@    package names
 function check_repo_version_yum()
 {
-	! type repoquery >/dev/null 2>&1 && return 255
+	type repoquery >/dev/null 2>&1 || return 255
 	local -a name=()
 	local -a ver=()
 	while read name ver
@@ -373,7 +374,7 @@ function check_repo_version_yum()
 # $@    package names
 function check_repo_version_zypper()
 {
-	! type zypper >/dev/null 2>&1 && return 255
+	type zypper >/dev/null 2>&1 || return 255
 	local -a name=()
 	local -a ver=()
 	while read name ver
@@ -403,7 +404,7 @@ function check_repo_version_zypper()
 # $@    package names
 function check_repo_version_apt()
 {
-	! type apt-cache >/dev/null 2>&1 && return 255
+	type apt-cache >/dev/null 2>&1 || return 255
 	local name=""
 	local ver=""
 	while read name ver
@@ -439,7 +440,7 @@ function check_repo_version()
 # $1    repo_id
 function get_package_list_yum()
 {
-	! type repoquery >/dev/null 2>&1 && return 255
+	type repoquery >/dev/null 2>&1 || return 255
 	local repo_id="$1"
 	[[ -z "${repo_id}" ]] && return 1
 	repoquery -qa "--repoid=${repo_id}" --qf "%{name}" 2>/dev/null
@@ -448,7 +449,7 @@ function get_package_list_yum()
 # $1    repo_id
 function get_package_list_zypper()
 {
-	! type zypper >/dev/null 2>&1 && return 255
+	type zypper >/dev/null 2>&1 || return 255
 	local repo_id="$1"
 	[[ -z "${repo_id}" ]] && return 1
 	zypper --no-gpg-checks -n search -r "${repo_id}" 2>/dev/null |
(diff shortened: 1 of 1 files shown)


xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -0,0 +1,1119 @@
...
+}
+
+# $1    version
+#       can be "2.10", "2.11", "2.12" or "latest"
+function add_xcat_core_repo_yum_or_zypper()
+{
+	! type rpm >/dev/null 2>&1 && return 255
+	local ver="$1"
+	[[ -z "${ver}" ]] && ver="latest"
+	local online_repo_file="http://xcat.org/files/xcat/repos/yum/${ver}/xcat-core/xCAT-core.repo"
```
whowutwut: Suggest that the `http://xcat.org` be a global variable in the event that URL changes in the future.

whowutwut:
@neo954 can this script be executed in a xCAT postscript?

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -0,0 +1,946 @@
...
+internal_setup
+
+# Check operating system
+function check_os()
+{
+	case "${OSTYPE}" in
+	"aix"*)     # AIX
+		echo "aix"
+		;;
+	"darwin"*)  # OS X
```
daniceexi: I am not sure whether we should have a this code to handle the unsupported OS even it's possible that certain day xCAT might support it.

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -0,0 +1,946 @@
...
+	local ver=""
+	while read ver
+	do
+		if [[ -z "${ver}" || "${ver}" =~ not\ installed ]]
+		then
+			echo "(not installed)"
+		else
+			echo "${ver}"
+		fi
+	done < <(rpm -q --qf '%{version}-%{release}\n' "$@")
```
daniceexi: Do you want to add the 2>/dev/null for the rpm command?

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -0,0 +1,946 @@
...
+	function_dispatch "${FUNCNAME}" "$@"
+}
+
+# $1    base url
+# $2    repo_id
+function add_repo_by_url_yum_or_zypper()
+{
+	local base_url="$1"
+	local repo_id="$2"
+	local tmp_repo_file="${TMP_DIR}/tmp_repo_file"
```
daniceexi: You should add a check to make sure the system support yum or zypper

~~~~
! type zypper && ! type yum >/dev/null 2>&1 && return 255
~~~~

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -0,0 +1,946 @@
...
+}
+
+function add_xcat_core_repo()
+{
+	function_dispatch "${FUNCNAME}" "$@"
+}
+
+function remove_xcat_core_repo_yum()
+{
+	rm -f /etc/yum.repos.d/xCAT-core.repo
```
daniceexi: why don't call ``remove_repo`` here?

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -0,0 +1,946 @@
...
+}
+
+function add_xcat_dep_repo_yum_or_zypper()
+{
+	! type rpm >/dev/null 2>&1 && return 255
+	local distro="${GO_XCAT_LINUX_DISTRO}"
+	case "${distro}" in
```
daniceexi: for ``fedora,centos,ol,SL``, we also should use the ``rh`` repo.

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -0,0 +1,946 @@
...
+	"fedora") ;;
+	"rhel") distro="rh" ;;
+	"sles") ;;
```
daniceexi: for ``sles and suse``, we should use ``sles`` repo

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -0,0 +1,946 @@
+#!/bin/bash
+
```
daniceexi: Several general comments:

* the name of the script is ``go-xcat``, that's cool. But I'd like it to be more straight forward. How bout ``instxcat`` or ``installxcat``, or ``xcatinst``, ``xcatinstall``?
* we should use `4 spaces`` to replace ``tab`` in the code format
* I'd like the ``install`` to be the default action. That means if no action is specified, use ``install``;
* how about make the second parameter to be ``action`` like: ``go-xcat install/update -y``
* I want to make a check after the install/update to make sure all the xcat-core packages have consistent version.

* update xCAT is a very common action, could you add it?

xCAT-server/share/xcat/tools/go-xcat
neo954: @daniceexi, this kind of check is handle by the lower level function it calls.

xCAT-server/share/xcat/tools/go-xcat
neo954: I don't have a ``fedora``/``centos``/``ol``/``SL`` environment currently. Once I have time and have chance to take a shot at these environments, I will add support for these Linux distributions.

xCAT-server/share/xcat/tools/go-xcat
neo954: @daniceexi, Four-space indent is good. But it will make HERE DOCUMENT looks messy. Since ``bash`` HERE DOCUMENT ``<<-`` operator does not work against four-space indent. In that way. the only way left is use HERE DOCUMENT ``<<`` operator. And it will make all the HERE DOCUMENT contents lost their indent. And this style looks messy.

neo954: @daniceexi, ``go-xcat install`` is good. And since currently, the ``install`` operation also handle ``update`` very well. I think there is no need to add a separate ``update`` operation.

I even imagine if we provide a feature of uninstalling xCAT for this script in the future. We can use ``go-xcat away`` to make it running. Okay, this is a joke.

xCAT-server/share/xcat/tools/go-xcat
neo954: @daniceexi, will do.

xCAT-server/share/xcat/tools/go-xcat
neo954: @daniceexi, this is a obsoleted function during reconstruction. This function will be removed entirely.

xCAT-server/share/xcat/tools/go-xcat
neo954: @daniceexi, some kind of checking was already added to this script. Please see function ``perform_smoke_test``. We need a way to figure out a way to check for consistent versions. Since currently, we have a timecode instead of git commit number in our version number. Which is not consistent. I suggest we use git commit number in our version number.

neo954:
@whowutwut, this script can be executed in a xCAT postscript. There is no reason why it cannot.

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -8,21 +8,22 @@ function usage()
...
 
 	Options:
 	  --help                        display this help and exit
-	  --install                     installs all the latest versions from the
-	                                repository
 	  --xcat-version=[VERSION]      specify the version of xCAT
 	  -y, --yes                     answer yes for all questions
 
+	Actions:
+	  install                       installs all the latest versions from the
```
daniceexi: could it be: install and update all the xcat-core and xcat-dep packages.

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -8,21 +8,22 @@ function usage()
...
 	while read ; do echo "${REPLY}" ; done <<-EOF
-	Usage: ${script} [OPTION]...
+	Usage: ${script} [OPTION]... [ACTION]
 	Install xCAT automatically
 
 	Options:
 	  --help                        display this help and exit
-	  --install                     installs all the latest versions from the
-	                                repository
 	  --xcat-version=[VERSION]      specify the version of xCAT
```
daniceexi: I'd like to have a short name for all the option. `--help`` could be ``-h``;
how about make a short name for option ``--xcat-version`` like ``-v`` or ``-x``

daniceexi:
I have reviewed the change and agree to merge. Thanks for @neo954 's great work.

whowutwut:
@neo954 Sorry for this late comment.   I still struggle with the chosen name for this script :-)  because it doesn't easily inform the user the purpose of the script....

why not something like `install-xcat`  where the user knows exactly the purpose of the script. 

or something more general where we could create a larger utility in the future... something like: `xcat-helper --install`

PR #1322, opened 2016-06-16
Altered statelite behavior w.r.t. /etc/mtab on RHEL7 distros

jtfrey:
In RHEL7/CentOS7, /etc/mtab has been changed to be a symlink to /proc/self/mount and the OS no longer updates /etc/mtab.  If /etc/mtab is replaced with a file under the /.statelite tmpfs, it remains blank and commands like "df" will fail.

Detecting when /etc/mtab is a symlink to /proc/self/mount and skipping that file's move to the tmpfs is the solution implemented here.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/statelite.pm
+++ b/xCAT-server/lib/xcat/plugins/statelite.pm
@@ -797,6 +797,18 @@ sub liteItem {
         # the /etc/mtab should be handled every time even the parent /etc/ has been handled.
         # if adding /etc/ to litefile, only tmpfs should be used. 
         if ($entry[1] eq "/etc/mtab") {
+            #
+            # In RHEL7 /etc/mtab is a symlink to /proc/self/mounts and not a regular file.
+            # If that's the case, then we skip forcing /etc/mtab into the .statelite path
+            # and symlink'ing /etc/mtab to that file.  Otherwise, since the OS isn't updating
+            # /etc/mtab commands like "df" will not be happy.
+            #
+            my $ret=`readlink $rootimg_dir/etc/mtab`;
+            chomp($ret);
+            if ($? == 0 and ($ret eq '/proc/self/mounts') ) {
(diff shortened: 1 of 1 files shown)


daniceexi:
@jtfrey Thanks for your contribution. xCAT only can accept your change after your signed the CLA: http://xcat-docs.readthedocs.io/en/latest/developers/license/contributors.html

xCAT-server/lib/xcat/plugins/statelite.pm
```diff
@@ -797,6 +797,18 @@ sub liteItem {
         # the /etc/mtab should be handled every time even the parent /etc/ has been handled.
         # if adding /etc/ to litefile, only tmpfs should be used. 
         if ($entry[1] eq "/etc/mtab") {
```
daniceexi: Looks like you are still using the ``link`` as the method to handle the lite files. Could you try to use ``tmpfs`` and to see whether it works. https://sourceforge.net/p/xcat/wiki/XCAT_Linux_Statelite/

jtfrey:
I'm not explicitly using "link" — that's what xcat does on its own when I run liteimg.

The fix I added here is to prevent /etc/mtab from ever being affected if in the osimage it's a symlink to /proc/self/mounts.  Whether it's affected via "link" or "tmpfs" is moot, since it just shouldn't be touched in that circumstance.

jtfrey:
(Oh, and I submitted my CLA yesterday afternoon shortly before adding this pull request.)

whowutwut:
@jtfrey  Thanks! I see the CLA.

jtfrey:
FYI:

[https://github.com/systemd/systemd/blob/v227/NEWS#L9-L14](https://github.com/systemd/systemd/blob/v227/NEWS#L9-L14)

daniceexi:
I have reviewed the change and agree to merge. Thanks @jtfrey

PR #1331, opened 2016-06-17
Ignore 'File exists' error from storage pool build

gurevichmark:
It is safe to ignore "File exists" error from `$poolobj->build()`
This error is returned when directory, even an empty one, exists when trying to setup a new storage pool.
The storage pool gets built even if this error is returned, so it is safe to continue and create a storage pool at that location. 
The files in the storage pool directory get preserved, if they existed there prior to storage pool creation.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -234,7 +234,10 @@ sub get_storage_pool_by_url {
     eval { $poolobj->build(); };
     if ($@) {
         my $error = $@;
-        unless ($error =~ /vgcreate.*exit status 3/ or $error =~ /pvcreate.*exit status 5/) {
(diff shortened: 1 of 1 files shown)


whowutwut:
I'm ok to merge

daniceexi:
@gurevichmark Sorry for the later comment, what if the origin disk size was 10G, but the new mkvm command requires 20G?

gurevichmark:
@daniceexi It still works. This fix deals with a storage pool which is just a container for volumes. 

I tested the following steps:

1. `mkvm mgmkvm1 -s 15G`
2. The following volume gets created for the vm:
image
3. `rmvm -p mgmkvm1`
4. The volume `mgmkvm1.sda.qcow2` is removed
5.  `mkvm mgmkvm1 -s 20G`
6. The following volume gets created for the vm:
image

daniceexi:
@gurevichmark OK, thanks.

Issue #1169, opened 2016-05-20
[Customer] The postscript syncfiles does not work when enabled the xcatdebugmode for rh7.2


daniceexi:
A customer found the syncfiles postscript did not work for rh7.2 OS deployment when ``xcatdebugmode` was enabled.

The root cause should be the ``sshd`` was started by anaconda, then the ``restart of sshd`` in the remoteshell postscript failed.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1334/files

daniceexi:
This has been verified on ``Red Hat Enterprise Linux Server release 7.2 (Maipo)``.

PR #1334, opened 2016-06-20
fix issue  [Customer] The postscript syncfiles does not work when enabled the xcatdebugmode for rh7.2 #1169;process the (sshd) service management in chrooted environment

immarvin:
fix issue  [Customer] The postscript syncfiles does not work when enabled the xcatdebugmode for rh7.2 #1169;
process the (sshd) service management in chrooted environment


daniceexi:

PR #1339, opened 2016-06-20
modify docker setup case according to issue 1336

junxiawang:
@tingtli  modify docker host set up case according to issue 1336


junxiawang:
The code is for pkg test.
perl /opt/xcat/share/xcat/tools/autotest/testcase/pkg/pkgtest.pl -s to use to add package
perl /opt/xcat/share/xcat/tools/autotest/testcase/pkg/pkgtest.pl -c to use to clear the environment
perl /opt/xcat/share/xcat/tools/autotest/testcase/pkg/pkgtest.pl -r to use to remove packages
perl /opt/xcat/share/xcat/tools/autotest/testcase/pkg/pkgtest.pl -i to use to add packages using incldue

PR #1340, opened 2016-06-20
Check if the postscript is only for diskless

mavscjb:
When adding the postscript to the osimage, check whether the script is only used for diskless.If only used for diskless, then do nothing for diskful osimage


daniceexi:
@bybai could you review this pull request?

PR #1341, opened 2016-06-20
fix sleep error when stopping tftpd process

chenglch:


daniceexi:

Issue #1343, opened 2016-06-20
xCAT rmdef command does not count up correctly in summary message


whowutwut:
I have a script creating 50 node definitions and removing those 50 node definitions but in the `rmdef` using a regular expression noderange.    In my naming scheme, There are 18 nodes per frame, so creating 50 nodes means that there's 4 nodes missing on the 3rd frame. 

f1n[01-18]  = 18
f2n[01-18]  = 18 + 18 = 36
f3n[01-14]  = 36 + 14 = 50

On the rmdef, I just did `rmdef -t node -o f[1-3]n[01-18]`. 

The error message that comes out is OK because the 4 nodes are not in the xCAT DB, but the summary count is 54 instead of 50 and is misleading. 

```
# cat xcatnodes.add | mkdef -z
50 object definitions have been created or modified.
# rmdef -t node f[1-3]n[01-18]
Error: Could not find an object named 'f3n17' of type 'node'.
Error: Could not find an object named 'f3n15' of type 'node'.
Error: Could not find an object named 'f3n18' of type 'node'.
Error: Could not find an object named 'f3n16' of type 'node'.
54 object definitions have been removed.  <--- this should be 50
```

whowutwut:
Verified on build 6/29
```
Version 2.12.1 (git commit 777164c7dea2c13807f3f847ab6499c2a06f9849, 
built Wed Jun 29 06:45:46 EDT 2016)
```

Adding in 100 nodes
```
[root@fs1 discovery]# cat xcatnodes.add  | mkdef -z 
100 object definitions have been created or modified.
```

Removing using a node range that does not contain the last 8 nodes.
```
[root@fs1 discovery]# rmdef -t node -o f[1-6]n[01-18]
Error: Could not find an object named 'f6n11' of type 'node'.
Error: Could not find an object named 'f6n14' of type 'node'.
Error: Could not find an object named 'f6n16' of type 'node'.
Error: Could not find an object named 'f6n17' of type 'node'.
Error: Could not find an object named 'f6n18' of type 'node'.
Error: Could not find an object named 'f6n15' of type 'node'.
Error: Could not find an object named 'f6n12' of type 'node'.
Error: Could not find an object named 'f6n13' of type 'node'.
100 object definitions have been removed.
```

The 100 count matches what was actually done against xCAT database,  Looks good. Thanks

PR #1345, opened 2016-06-20
Correct the count of removed objects

gurevichmark:
Display correct count of removed objects by subtracting the number of objects not removed from the total number of attempted objects to remove.
Fixed issue #1343

### DIFF
--- a/xCAT-server/lib/xcat/plugins/DBobjectdefs.pm
+++ b/xCAT-server/lib/xcat/plugins/DBobjectdefs.pm
@@ -3818,6 +3818,7 @@ sub defrm
     #    the memberlist nodes must be updated.
 
     my $numobjects = 0;
+    my $numobjects_not_removed = 0;
     my %objTypeLists;
     foreach my $obj (keys %objhash)
     {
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/DBobjectdefs.pm
```diff
@@ -3831,6 +3832,7 @@ sub defrm
             my $rsp;
             $rsp->{data}->[0] = "Could not find an object named \'$obj\' of type \'$objtype\'.";
             xCAT::MsgUtils->message("E", $rsp, $::callback);
+            $numobjects_not_removed++;
             next;
         }
         $numobjects++;
```
whowutwut: @gurevichmark  Is it better to keep track of another object, or simply to maybe move this incrementation down after the error cases are checked?

gurevichmark: Yes you are right. The solution can be simplified and it also handles -V (verbose) option printing the list of deleted objects.
New commit checked in.

whowutwut:
Thanks, will merge

Issue #1342, opened 2016-06-20
setupntp need enhance to support sles12


chudegao:
On sles12, the ntpd startup configuration file is /etc/sysconfig/ntp rather than /etc/sysconfig/ntpd. To make the hardware clock update automatically need update this configuration.  Othrewise, reboot CN the os time maybe different with hardware clock.

vi /etc/sysconfig/ntp

~~~~
## Type:           boolean
## Default:        "yes"
#
# Force time synchronization befor start ntpd
#
NTPD_FORCE_SYNC_ON_STARTUP="no"         ===> update to 'yes'

## Type:           boolean
## Default:        "no"
#
# Force time synchronization of hwclock befor start ntpd.
# This works only if NTPD_FORCE_SYNC_ON_STARTUP is set
# to yes.
#
NTPD_FORCE_SYNC_HWCLOCK_ON_STARTUP="yes"
~~~~

chudegao:
@bybai , PCM need this fix in current release. Is it possible make it available this week? I can provide a test environment for you. Thanks.

PR #1350, opened 2016-06-21
fix 1342 setupntp need enhance to support sles12

bybai:
fix issue #1342


daniceexi:
I am good with the change. @immarvin please review it.

@tingtli could we confirm why our automation did not find this problem?

immarvin:
agree to merge it

whowutwut:
Adding testcase requested label to track why auto test didn't catch this

Issue #1351, opened 2016-06-21
[FVT]:go-xcat needs manually input when --yes is placed after install 


tingtli:
xCAT 2.12.1
when I input "./go-xcat install --yes", it still ask user to input the answers for the question. "./go-xcat -- yes install " works without user's manual input.

```
[root@c910f02c01p21 /]# ./go-xcat install --yes
Operating system:   linux
Architecture:       ppc64
Linux Distribution: rhel
Version:            7.1

Reading repositories ...... done

xCAT is going to be installed.
Continue? [y/n] 
```

```
[root@c910f02c01p21 /]# ./go-xcat --yes install
Operating system:   linux
Architecture:       ppc64
Linux Distribution: rhel
Version:            7.1

Reading repositories ...... done
Failed to set locale, defaulting to C
Loaded plugins: product-id, subscription-manager
This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.
Package conserver-xcat-8.1.16-10.ppc64 already installed and latest version
Package elilo-xcat-3.14-4.noarch already installed and latest version
Package grub2-xcat-2.02-0.16.el7.snap201506090204.noarch already installed and latest version
Package ipmitool-xcat-1.8.15-1.ppc64 already installed and latest version
Package syslinux-xcat-3.86-2.noarch already installed and latest version
Package xnba-undi-1.0.3-131028.noarch already installed and latest version
Package yaboot-xcat-1.3.17-rc1.noarch already installed and latest version
Resolving Dependencies
--> Running transaction check
---> Package perl-xCAT.noarch 4:2.12-snap201605200310 will be installed
---> Package xCAT.ppc64 0:2.12-snap201605200311 will be installed
---> Package xCAT-buildkit.noarch 4:2.12-snap201605200311 will be installed
```

neo954:
Patch in pull request #1354

tingtli:
@neo954 
Please also take a look at -x option
"./go-xcat -x 2.12 -y install" does not install xCAT
"./go-xcat -y install -x 2.12" works

neo954:
@tingtli This is another problem. Please take the patch. And open another issue for the new problem.
Please attach all the detail outputs generated by ``go-xcat``. And please attach the error logs``go-xcat`` generated if you have any.

neo954:
@tingtli, the patch was merged last night. Please get the latest ``go-xcat`` from the git repo.

tingtli:
Verified 
```
[...]

PR #1354, opened 2016-06-21
[go-xcat] Fix bug #1351 of command line argument processing

neo954:


daniceexi:
The fix looks OK to merge.

Issue #980, opened 2016-04-27
The node status won't be updated to Powering-on after the running of [rpower on]


daniceexi:
In the latest xCAT code, when running **rpower on** to perform an OS deployment, the node **status** attribute won't be updated until the installer is running. So that means in the scenario the original status is **powering-off**, then user run **rpower on** to start the OS deployment. If the OS deployment process failed to start, the node still keeps as **powering-off** even the node has been in power-on state.

I'd like node **status** can be updated to **powering-on** immediately after running **rpower on** command.

immarvin:
hi @penguhyang , please take a look at this feature

daniceexi:
@xuweibj has done the basic code change for mgt=ipmi/blade/hmc/kvm, will change for mgt=fsp in next release. And also @xuweibj will improve the performance of rpower.

daniceexi:
The change has been verified.

PR #1358, opened 2016-06-22
Fix issue 980, add status powering-on

xuweibj:
#980 
Just change kvm.pm to check. If it is ok, will change other files.
If set command "power node on", the status will be "powering-on".
If set command "power node reset" after nodeset when the machine is on, the status will be powering-on--> installing--> booting--> booted.


immarvin:
i am ok with the change,agree to merge

PR #1366, opened 2016-06-22
Fix some issues with the HW discovery script after testing on OpenPower

whowutwut:
After running the testing on OpenPower machines, found some issues in the various hardware discovery scripts.

### DIFF
--- a/xCAT-genesis-scripts/bin/bmcsetup
+++ b/xCAT-genesis-scripts/bin/bmcsetup
@@ -37,12 +37,13 @@ function cold_reset_bmc() {
             # reset after applying ipmitool commands.  However, it seems there is a problem with 
             # the BMC where after 15 seconds, it stops responding.  To work around, sleep 30
             # seconds before issuing the reset of the BMC.
-            sleep 30
+            snooze
+        else
+            logger -s -t $log_label -p local4.info "Resetting BMC ..."
+            ipmitool mc reset cold
+            logger -s -t $log_label -p local4.info "Waiting for the BMC to appear ..."
         fi
-        logger -s -t $log_label -p local4.info "Resetting BMC ..."
-        ipmitool mc reset cold
 
-        logger -s -t $log_label -p local4.info "Waiting for the BMC to appear ..."
         if [ "$XPROD" = "43707" ]; then
             # OpenPower SPECIFIC, check the BMC with the following raw command to
             # make sure that the bmc is really in a "ready" state before continuing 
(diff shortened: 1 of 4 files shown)


xCAT-genesis-scripts/bin/doxcat
```diff
@@ -274,7 +285,7 @@ if [ -e "/dev/rtc" ]; then
     hwclock --systohc
 fi
 HOST_ARCH=`uname -m`
-if echo $HOST_ARCH | grep "ppc64"; then
+if [ `echo $HOST_ARCH | grep "bac" > /dev/null; echo $?` == 0 ]; then
```
zet809: @whowutwut Why bac here?

whowutwut: good catch, that is a typo.

whowutwut:
I made some updates to the pull request to fix some of the messages and remove some of the changes to the `doxcat` that was causing issues

whowutwut:
@zet809  This seems to be working on the firestone machine we have in the discovery environment at least for configuration of the BMC, should we merge this in for 2.12?

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -37,12 +37,13 @@ function cold_reset_bmc() {
             # reset after applying ipmitool commands.  However, it seems there is a problem with 
             # the BMC where after 15 seconds, it stops responding.  To work around, sleep 30
             # seconds before issuing the reset of the BMC.
-            sleep 30
+            snooze
+        else
+            logger -s -t $log_label -p local4.info "Resetting BMC ..."
```
zet809: So it means no "mc reset cold" is needed for both Firestone and Garrison?

xCAT-genesis-scripts/bin/getipmi
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+log_label="xcat.genesis.getipmi"
```
zet809: What is the purpose of add this variable? I didn't search out where  it is used.

xCAT-genesis-scripts/bin/bmcsetup
whowutwut: yes, it seems that way in the current testing, we do not need to issue `mc reset cold` . Maybe something was fixed in the firmware?

xCAT-genesis-scripts/bin/getipmi
whowutwut: There's no current purpose for this.  Since I touched the file, I just set up the framework incase we want to have syslog log messages in this script.  If we don't want this, I can undo it.

zet809:
@whowutwut @pdlun92 The modification looks good to me, I agree to merge it.

PR #1368, opened 2016-06-23
replace the control files with pcm specific control file

tingtli:


daniceexi:
The change looks ok to merge.

whowutwut:
By doing this, the changes are lost when we create 2.13-pcm branch.

daniceexi:
@whowutwut This is a temporary fix, the final solution should be make a script to generate pcm control file from the xcat ones.

Issue #1369, opened 2016-06-23
[FVT]lsdef/mkdef commands "Error: Unsupported request" error on rh6.8 ppc64


caomengmeng:
Description: using latest daily build to install rh6.8 on ppc64, after xcat is installed successfully, the lsdef/makedef commands return error message. See below for details:

Test ENV:
```
[root@c910f02c08p08 ~]# cat /etc/*release
Red Hat Enterprise Linux Server release 6.8 (Santiago)
Red Hat Enterprise Linux Server release 6.8 (Santiago)
[root@c910f02c08p08 ~]# rpm -qa | grep -i xCAT
conserver-xcat-8.1.16-10.ppc64
xCAT-buildkit-2.12.1-snap201606220646.noarch
perl-xCAT-2.12.1-snap201606220645.noarch
xCAT-server-2.12.1-snap201606220645.noarch
xCAT-genesis-scripts-ppc64-2.12.1-snap201606220646.noarch
ipmitool-xcat-1.8.15-1.ppc64
grub2-xcat-2.02-0.16.el7.snap201506090204.noarch
xCAT-2.12.1-snap201606220646.ppc64
xCAT-genesis-base-ppc64-2.12-snap201605051443.noarch
xCAT-client-2.12.1-snap201606220645.noarch
xCAT-test-2.12.1-snap201606220646.noarch
```
Details:

```
[root@c910f02c08p08 ~]# service xcatd restart
Restarting xcatd Error loading module /opt/xcat/lib/perl/xCAT_plugin/DBobjectdefs.pm  ...skipping
                                                           [  OK  ]
[root@c910f02c08p08 ~]# lsdef -v
Error: Unsupported request
[root@c910f02c08p08 ~]# mkdef -v
Error: Unsupported request 
```

immarvin:
btw, there is a warning in:
````
[root@c910f02c08p08 ~]# perl -I /opt/xcat/lib/perl/ -c /opt/xcat/lib/perl/xCAT_plugin/DBobjectdefs.pm 
Variable "$objfiledata" will not stay shared at /opt/xcat/lib/perl/xCAT_plugin/DBobjectdefs.pm line 678.
```` 

the 2 issues are caused by the syntax restrict in perl v5.10.1 

[root@c910f02c08p08 ~]# perl -v

This is perl, v5.10.1 (*) built for ppc64-linux-thread-multi

Copyright 1987-2009, Larry Wall

[...]

PR #1370, opened 2016-06-23
fix issue  [FVT]lsdef/mkdef commands <Error: Unsupported request error> on rh6.8 ppc64 #1369 @github

immarvin:
fix issue  [FVT]lsdef/mkdef commands "Error: Unsupported request" error on rh6.8 ppc64 #1369 
refine the code to apply to the syntax in perl 5.10.1:
1. ``map`` subroutine does not support anonymous hash 
2. use anonymous subroutine to get freshly bound to the lexical environment at runtime


daniceexi:

PR #1375, opened 2016-06-23
[go-xcat] Fix a bug on Ubuntu. On amd64, apt is configured for bi-arch, while xcat repository is not bi-arch

neo954:


daniceexi:

Issue #1379, opened 2016-06-23
Add xCAT Fedora 24 support


neo954:
```
# ./go-xcat --yes install
Operating system:   linux
Architecture:       x86_64
Linux Distribution: fedora
Version:            24
go-xcat: fedora: unsupported Linux distro

Reading repositories ...... done
Yum command has been deprecated, redirecting to '/usr/bin/dnf --nogpgcheck -y install perl-xCAT xCAT xCAT-buildkit xCAT-client xCAT-genesis-scripts-ppc64 xCAT-genesis-scripts-x86_64 xCAT-server conserver-xcat elilo-xcat grub2-xcat ipmitool-xcat syslinux-xcat xCAT-genesis-base-ppc64 xCAT-genesis-base-x86_64 xnba-undi yaboot-xcat'.
See 'man dnf' and 'man yum2dnf' for more information.
To transfer transaction metadata from yum to DNF, run:
'dnf install python-dnf-plugins-extras-migrate && dnf-2 migrate'

Last metadata expiration check: 0:00:01 ago on Thu Jun 23 21:53:16 2016.
Error: nothing provides /etc/xinetd.d/tftp needed by xCAT-2.12-snap201605200311.x86_64.
nothing provides libcrypto.so.6()(64bit) needed by conserver-xcat-8.1.16-10.x86_64
(try to add '--allowerasing' to command line to replace conflicting packages)

xCAT Core Packages
==================

Package Name                Installed                      In Repository
------------                ---------                      -------------
perl-xCAT                   (not installed)                2.12-snap201605200310
xCAT                        (not installed)                2.12-snap201605200311
xCAT-buildkit               (not installed)                2.12-snap201605200311
xCAT-client                 (not installed)                2.12-snap201605200310
xCAT-confluent              (not installed)                2.12-snap201605200311
xCAT-genesis-scripts-ppc64  (not installed)                2.12-snap201605200311
xCAT-genesis-scripts-x86_64 (not installed)                2.12-snap201605200311
xCAT-server                 (not installed)                2.12-snap201605200310
xCATsn                      (not installed)                2.12-snap201605200311
xCAT-SoftLayer              (not installed)                2.12-snap201605200311
xCAT-test                   (not installed)                2.12-snap201605200311
xCAT-vlan                   (not installed)                2.12-snap201605200311

xCAT Dependency Packages
========================

Package Name                Installed                      In Repository
------------                ---------                      -------------
[...]

PR #1380, opened 2016-06-23
[go-xcat] Add fedora 24 dnf support. Now, go-xcat can work with fedora 24

neo954:


neo954:
One issue #1379 is opened for xCAT fedora 24 support.

daniceexi:

Issue #1326, opened 2016-06-17
rmkit -f essl failed to remove the essl kit. 


wwzhubj:
My osimage is : 
lsdef -t osimage rh72diskfull1621
Object name: rh72diskfull1621
    exlist=/install/osimages/rh72diskfull1621/kits/KIT_COMPONENTS.exlist
    imagetype=linux
    kitcomponents=essl-license-5.5.0-0-rhels-7.2-ppc64le,xlf.license-compute-15.1.3-0-rhels-7.2-ppc64le,essl-computenode-5.5.0-0-rhels-7.2-ppc64le,xlc.compiler-compute-13.1.3-0-rhels-7.2-ppc64le,xlf.rte-compute-15.1.3-0-rhels-7.2-ppc64le,xlc.license-compute-13.1.3-0-rhels-7.2-ppc64le,xlf.compiler-compute-15.1.3-0-rhels-7.2-ppc64le,xlc.rte-compute-13.1.3-0-rhels-7.2-ppc64le,essl-loginnode-5.5.0-0-rhels-7.2-ppc64le,pessl-loginnode-5.2.0-0-rhels-7.2-ppc64le,pessl-license-5.2.0-0-rhels-7.2-ppc64le,pessl-computenode-5.2.0-0-rhels-7.2-ppc64le
    osarch=ppc64le
...
------------------------------------------------------------------------------------------------------------------------
[root@c712ems4 test_nogpu]# rmkit -f essl
Removing kit essl-5.5.0-0-ppc64le
Error: Command failed: rmkitcomp. Error message: kitcomponent pperte-compute basename does not existRemoving kitcomponent essl-computenode-5.5.0-0-rhels-7.2-ppc64le from osimage rh72diskfull1621kitcomponents essl-computenode-5.5.0-0-rhels-7.2-ppc64le were removed from osimage rh72diskfull1621 successfully.

Error: Failed to remove kit component essl-computenode-5.5.0-0-rhels-7.2-ppc64le from rh72diskfull1621

bybai:
Hi @wwzhubj , For test kit, rmkit can works, so I need your environment to reproduce this issue:
[root@c910f05c01bc06 testkit]# addkit testkit-1.0-1.tar.bz2
Adding Kit testkit-1.0-1
Kit testkit-1.0-1 was successfully added.
[...]

PR #1385, opened 2016-06-24
issue1326 rmkit -f essl failed to remove the essl kit

bybai:
#1326


daniceexi:
I am OK with the change.

PR #1387, opened 2016-06-24
Update failed runcmdinstaller case

caomengmeng:
@tingtli 
Update runcmdinstaller -h test case according to xCAT code changed
Please help to review 
Thanks.


tingtli:
We need to write a new case for runcmdinstaller according to their code change. I've create a task in our RTC.

Issue #1361, opened 2016-06-22
setroute failed when routenames= "v2094gw9-114-75-254"


wwzhubj:
setroute failed when routenames= "v2094gw9-114-75-254". 
when the routenames set with v2094gw, setroute works well.

wwzhubj:
the other issue is : 
After installation,  setroute can not work.
after comment GATEWAY=192.168.130.1 in /etc/sysconfig/network, reboot the node, the route table changed. 
[root@c712f7n04-pvt ~]# cat /etc/sysconfig/network
# Created by anaconda
#GATEWAY=192.168.130.1

wwzhubj:
My env: 

[root@c712ems4 ~]# lsxcatd -v
Version 2.12 (git commit 5df15f39216da9c64759d61b19fbabad1674727d, built Wed May 18 07:30:41 EDT 2016)
[root@c712ems4 ~]# uname -a
Linux c712ems4 3.10.0-327.13.1.el7.openpower3.ppc64le #1 SMP Thu Apr 28 13:22:48 EDT 2016 ppc64le ppc64le ppc64le GNU/Linux

daniceexi:
For the second issue that the default gateway set by the ``setroute`` postscript did not work was caused by the ``default`` gateway (from the install network) has been set by anaconda in the ``/etc/sysconfig/network`` during the OS deployment. Then when the preferred gateway (from the second network) to be enabled from route configuration file, the system said there was a conflict. 

The correct configuration is if user wants to set a default gateway, the gateway attribute for the install network should be removed.

~~~~
chdef -t network <install_network> gateway=
~~~~

wwzhubj:
What the fix is xCAT can not support this naming way. it will report error message when using the route name including '-'.

PR #1388, opened 2016-06-24
issue 1361: check the export sentence in the mypostscript to make sur…

daniceexi:
…e no - is used in the variable name


daniceexi:
fix issue #1361

immarvin:
agree to merge

PR #1391, opened 2016-06-24
pcm needs version 2.12 for PCM 4.2.2

tingtli:


daniceexi:
The change is OK for merge.

PR #1392, opened 2016-06-24
[go-xcat] Add better support for earlier Fedora Linux release. And add CentOS Linux support.

neo954:


daniceexi:

Issue #1185, opened 2016-05-23
[Customer] bootloader is not grabbing the addkcmdline from osimage


arif-ali:
I have updated my kickstart template to use #KICKSTARTBOOTLOADER#, but the values that have been added to the addkcmdline are not being added to this parameter. Any ideas on this would be great

Version of xCAT below
```
# lsxcatd -a
Version 2.12 (git commit 588076d19fdf219157690ce65299e92ca42f1398, built Fri May 20 03:10:38 EDT 2016)
This is a Management Node
dbengine=SQLite
```

daniceexi:
@arif-ali In the 2.12, we are using the keyword ``R:<kernel param>`` to add specific kernel parameter to the kickstart bootloader attribute. The ``#KICKSTARTBOOTLOADER#`` won't be used any more since we need to customize the ``bootloader`` parameters in the pre-script.

Could you give me more information of your requirement?

immarvin:
please try ``chdef -t osimage -o <osimage name> -p addkcmdline="R::<persistent kcmdline>"`` 

````
[root@c910f03c05k21 ~]# tabdump -d linuximage
...
addkcmdline:    User specified arguments to be passed to the kernel.  The user arguments are appended to xCAT.s default kernel arguments. For the kernel options need to be persistent after installation, specify them with prefix "R::".  This attribute is ignored if linuximage.boottarget is set.
````

#KICKSTARTBOOTLOADER# for redhat 7 has been obsoleted in xcat 2.12  
https://github.com/xcat2/xcat-core/pull/1118/files

arif-ali:
Thanks, I have figured it out.

I updated the addkcmdline, and then added #KICKSTARTBOOTLOADER# into my custom partition file. I presume, this is the recommended way?

whowutwut:
@immarvin  can you double check if our documentation reflected that change in xCAT 2.12?

daniceexi:
@arif-ali In the 2.12, the ``#KICKSTARTBOOTLOADER#`` has been commented off, so user don't need to add it back manually. Is there any problem that why you added it back?

arif-ali:
@daniceexi I have my own partitionfile, and therefore wouldn't be picked up in /tmp/partitioning, as I am not using the /tmp/partitioning. Or is there a new way of specifying the partition?

arif-ali:
[...]

PR #1393, opened 2016-06-24
fix issue  [Customer] bootloader is not grabbing the addkcmdline from osimage #1185

immarvin:
fix issue:https://github.com/xcat2/xcat-core/issues/1185

1. rename /tmp/partitioning to /tmp/partitionfile to keep consistent across automatically generated partition,user-specified partitionfile, user-specified partition script
2. support persistent kernel options addkcmdline="R::" in automatically generated partition,user-specified partitionfile, user-specified partition script


daniceexi:

Issue #450, opened 2015-11-20
[Customer］xCAT 2.11: failed to nodeset <node> offline (back to default/discovery boot config)


mfoliveira:
xCAT: xCAT-2.11-snap201511121931.ppc64le
OS: RHEL 7.2 on S822LC

Node is discovered, and defined in dhcpd.leases.
And the boot config file exists.

	# sed '/p8r1n2/,/}/p' -n /var/lib/dhcpd/dhcpd.leases
	host p8r1n2 {
	  dynamic;
	  hardware ethernet 98:be:94:59:f0:f2;
	  fixed-address 10.1.1.2;
		supersede server.ddns-hostname = "p8r1n2";
		supersede host-name = "p8r1n2";
		supersede conf-file = "http://10.1.0.1/tftpboot/petitboot/p8r1n2";
	}

	# cat /tftpboot/petitboot/p8r1n2 
	#offline
	default xCAT
	label xCAT
		kernel http://10.1.0.1:80/tftpboot/
		initrd http://10.1.0.1:80/tftpboot/

Now try to cleanup it's boot config back to none/default/discovery:

	# nodeset p8r1n2 offline
	p8r1n2: offline

	# nodeset p8r1n2 stat
	p8r1n2: offline

It didn't work.
The configuration didn't change in dhcpd.leases / boot config file.

	# sed '/p8r1n2/,/}/p' -n /var/lib/dhcpd/dhcpd.leases
	host p8r1n2 {
	  dynamic;
	  hardware ethernet 98:be:94:59:f0:f2;
	  fixed-address 10.1.1.2;
		supersede server.ddns-hostname = "p8r1n2";
		supersede host-name = "p8r1n2";
		supersede conf-file = "http://10.1.0.1/tftpboot/petitboot/p8r1n2";
	}

	# cat /tftpboot/petitboot/p8r1n2 
	#offline
	default xCAT
	label xCAT
		kernel http://10.1.0.1:80/tftpboot/
		initrd http://10.1.0.1:80/tftpboot/

It works correctly if you remove that section from dhcpd.leases, and restart the dhcp daemon.

	# sed '/p8r1n2/,/}/d' -i /var/lib/dhcpd/dhcpd.leases

	# sed '/p8r1n2/,/}/p' -n /var/lib/dhcpd/dhcpd.leases
	# 

	# systemctl restart dhcpd

daniceexi:
[...]

PR #1394, opened 2016-06-24
Nodeset offline for ppc64

gurevichmark:
This pull request fixes problem reported by issue #450.
The perl modules for handling nodeset offline for ppc64 did not remove boot configuration files in /tftpboot.
The perl modules for handling x86 were not changed. They appear to correctly remove the boot configuration files.

This pull request does the following:
1. Updates man page for **nodeset** command to remove the `iscsiboot` option
2. Removes duplicate `getstate` function definition from **petitboot.pm**
3. Removes appropriate boot configuration files from /tftpboot directory when **nodeset \<node\> offline** is issued for ppc64le virtual and ppc64le non-virtual
4. Removes dhcp entries from dhcp.leases file for nodes specified in **nodeset \<node\> offline**
5. Changes the status returned by **nodeset \<node\> stat** from `boot` to `offline` if no boot configuration file is found.

Unit test was performed against ppc64le VM and ppc64le non-VM systems. Single machine and multiples.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/grub2.pm
+++ b/xCAT-server/lib/xcat/plugins/grub2.pm
@@ -23,7 +23,7 @@ my $globaltftpdir = xCAT::TableUtils->getTftpDir();
 
 
 my %usage = (
-"nodeset" => "Usage: nodeset <noderange> [shell|boot|runcmd=bmcsetup|iscsiboot|osimage[=<imagename>]|offline|shutdown|stat]",
+"nodeset" => "Usage: nodeset <noderange> [shell|boot|runcmd=bmcsetup|osimage[=<imagename>]|offline|shutdown|stat]",
 );
 
 sub handled_commands {
@@ -78,7 +78,8 @@ sub getstate {
             chomp($headline);
             return $headline;
         } else {
-            return "boot";
+            # There is no boot configuration file, node must be offline
+            return "offline";
         }
     } else {
         return "discover";
@@ -181,18 +182,20 @@ sub setstate {
     my $nodemac;
     my %client_nethash = xCAT::DBobjUtils->getNetwkInfo([$node]);
 
-    open($pcfg, '>', $tftpdir . "/boot/grub2/" . $node);
     my $cref = $chainhash{$node}->[0]; #$chaintab->getNodeAttribs($node,['currstate']);
-    if ($cref->{currstate}) {
+
+    # remove the old boot configuration file and create a new one, but only if not offline directive
+    unlink($tftpdir . "/boot/grub2/" . $node);
+    if ($cref and $cref->{currstate} ne "offline") {
+        open($pcfg, '>', $tftpdir . "/boot/grub2/" . $node);
         print $pcfg "#" . $cref->{currstate} . "\n";
+
+        if (($::XCATSITEVALS{xcatdebugmode} eq "1") or ($::XCATSITEVALS{xcatdebugmode} eq "2")) {
+            print $pcfg "set debug=all\n";
+        }
+
+        print $pcfg "set timeout=5\n";
     }
-
-
-    if (($::XCATSITEVALS{xcatdebugmode} eq "1") or ($::XCATSITEVALS{xcatdebugmode} eq "2")) {
-        print $pcfg "set debug=all\n";
-    }
-
-    print $pcfg "set timeout=5\n";
 
     $normalnodes{$node} = 1;   #Assume a normal netboot (well, normal dhcp,
                                #which is normally with a valid 'filename' field,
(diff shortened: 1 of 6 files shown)


xCAT-server/lib/xcat/plugins/petitboot.pm
```diff
@@ -535,6 +542,14 @@ sub process_request {
           }
       }
   }
+
```
daniceexi: Please use 4 spaces as indent

xCAT-server/lib/xcat/plugins/petitboot.pm
```diff
@@ -188,8 +192,8 @@ sub setstate {
     #        arg=>['-s','filename = \"xcat/nonexistant_file_to_intentionally_break_netboot_for_localboot_to_work\";']},$callback);
     #print $pcfg "bye\n";
     close($pcfg);
-  } elsif ($kern and $kern->{kernel}) {
-    #It's time to set yaboot for this node to boot the kernel..
+  } elsif ($kern and $kern->{kernel} and $cref and $cref->{currstate} ne "offline") {
+    #It's time to set yaboot for this node to boot the kernel, but only if not offline directive
```
daniceexi: this comment is only for your change, the ``yaboot`` in comment message should be replaced to ``petitboot``

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -713,6 +724,17 @@ sub process_request {
         }
     }
 
+    if ($args[0] eq 'offline') {
+        # If nodeset directive was offline we need to remove the architecture file link and remove dhcp entries
+        foreach my $osimage (keys %osimagenodehash) {
```
daniceexi: for performance consideration, we should put all the nodes in an array and then call makedhcp -d at once.

xCAT-server/lib/xcat/plugins/petitboot.pm
```diff
@@ -535,6 +542,14 @@ sub process_request {
           }
       }
   }
+
+  if ($args[0] eq 'offline') {
+    # If nodeset directive was offline we need to remove dhcp entries
+    foreach my $node (@normalnodeset) {
+      $sub_req->({ command => ['makedhcp'],arg=>['-d'],
```
daniceexi: for performance consideration, we should put all the nodes in an array and then call makedhcp -d at once.

daniceexi: do we need to run ``makedhcp -d`` for netboot=xnba?

daniceexi: A common question, does the 'nodeset <node> offline' can make the the <node> get into the discovery in the next boot?

gurevichmark: @daniceexi I think we should add `makedhcp -d` to netboot=xnba. But I did not want to change code that was not specific for the issue #450.
If you agree, I can add code to call `makedhcp -d` to xnba.pm.
Should I:
a. Use the same pull request as this code for ppc64 ?
b. Use a separate pull request and reference the same issue #450 
b. Open a separate issue for xnba and fix it under a separate pull request ?

daniceexi: @gurevichmark I'd like to make changes for xnba in this pull request to ease to track. But should be in different commits.

Since we don't need to run discovery for ppc64, I prefer to leave ppc64 as is.

gurevichmark: @daniceexi Ok. I will make a similar change for **xnba**. What about **pxe** ? Should I make a similar change there ?
Also, can you please explain what you meant by: *Since we don't need to run discovery for ppc64, I prefer to leave ppc64 as is.* ? Did you mean that **yaboot** or **grub2** does not need to be changed ?

daniceexi:
@gurevichmark sorry to give the incorrect information. I meant that we don't need to take care of removing dhcp entry from dhcpd.leases for ppc64. But I don't think it was correct, please ignore this message.

I think we should make all the netboot methods have the consistent behaviour for ``nodeset offline``:
* remove configuration file
* remove bootloader file for the node if necessary
* remove dhcp entry from dhcpd.lease

daniceexi:
Looks like we cannot finish this pull request in 2.12.1, so move it to 2.12.2

gurevichmark:
Second commit added to fix code review suggestions:
1. Indentation
2. Single call to `makedhcp -d` with a list of nodes instead of one by one.

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -724,15 +724,16 @@ sub process_request {
         }
     }
 
+    my @makedhcp_nodes;
```
daniceexi: could you put this variable ``@makedhcp_nodes`` inside the L728 block?

I suggest the variable name could be 'rmdhcp_nodes'

gurevichmark:
@daniceexi Thank you for you suggestion. I checked in another commit with the suggested changes for `grub2.pm` and also simplified code for `xnba.pm`

daniceexi:

gurevichmark:
@daniceexi Thank you for reviewing the code changes.

* Should `pxe.pm` also be changed to remove dhcp entries from `dhcpd.lease` when `nodeset offline` is issued ?
* Should I fix the formatting to `xnba.pm`, `grub2.pm` and `petitboot.pm` using `perltidy` ?

daniceexi:
I don't think there's too much people are still using netboot=pxe. So we don't need to change it.

Yes, please run perltidy against these files and make a new new pull request.

In fact, I am thinking to run perltidy against all the xCAT files in the this release beginning phase.

Issue #1395, opened 2016-06-26
[FVT]:go-xcat failed if the local repo file is specified to /etc/apt/source.list.d/xcat-core.list and  /etc/apt/source.list.d/xcat-dep.list


tingtli:
xCAT 2.12.1
If the repo file is specified to  --xcat-core=/etc/apt/sources.list.d/xcat-core.list --xcat-dep=/etc/apt/sources.list.d/xcat-dep.list . 
```
root@c910f03c17k24:/# ./go-xcat --xcat-core=/etc/apt/sources.list.d/xcat-core.list --xcat-dep=/etc/apt/sources.list.d/xcat-dep.list -y install
Operating system:   linux
Architecture:       ppc64le
Linux Distribution: ubuntu
Version:            16.04

Reading repositories ...... failed
cp: '/etc/apt/sources.list.d/xcat-core.list' and '/etc/apt/sources.list.d/xcat-core.list' are the same file
```

neo954:
@tingtli, What is your expect behavior of the script?
IMHO, a reasonable error message is already printed, and nothing more can be done.

An administrator should understand what he/she is doing. I do not think this is a bug.

tingtli:
@neo954  You provide local repo file to be as the repo file. I specify the repo file, i think it should work. But why it says the conflict?  I think the input "/etc/apt/sources.list.d/xcat-core.list" is reasonable input.
it should not give error message but should install xcat.

neo954:
Local repo file support is not intend to be used by the end user. And xCAT does not provide any Ubuntu repo file at all. Thus, the local repo file support on Ubuntu is removed.

tingtli:
@neo954 

```
root@c910f03c17k24:/# ./go-xcat --xcat-core=/etc/apt/sources.list.d/xcat-core.list --xcat-dep=/etc/apt/sources.list.d/xcat-dep.list
Operating system:   linux
Architecture:       ppc64le
Linux Distribution: ubuntu
Version:            16.04

Reading repositories ...... failed
go-xcat: /etc/apt/sources.list.d/xcat-core.list: unknown archive file
go-xcat: extract xcat-core archive file failed
```

tingtli:
@neo954 Do you have any updates for this bug?

neo954:
Specify local repo files on Ubuntu is unsupported.

neo954:
Will close this one.

PR #1400, opened 2016-06-27
[go-xcat] Fix issue #1395.

neo954:
Fix issue #1395. Remove the support of local repoitory file on Debian/Ubuntu


daniceexi:
@neo954 The error was the apt-get install failed when specify a .tar.bz2 file. but did not see you have fix for that.

daniceexi:

PR #1332, opened 2016-06-17
xCAT probe subcommand to check consistency of os images on compute nodes

gurevichmark:
xcatprobe subcommand to check consistency of os images on compute nodes.

Two verification steps are provided:

`-c` verifies that all pingable diskless compute nodes are running identical versions of osimage. This is done by comparing osimagename and osimageuuid values from `/opt/xcat/xcatinfo` file on each node.

`-d` verifies that each pingable diskless compute node is running the same osimage as defined for it on MN. This is done by comparing osimagename and osimageuuid values from `/opt/xcat/xcatinfo` file on the node and the  osimagename and osimageuuid values from `<rootimagedir>/rootimg/opt/xcat/xcatinfo` file on MN.

PR #1406, opened 2016-06-27
xcatprobe image improvements based on code review suggestions

gurevichmark:
This pull request implements enhancements suggested by @daniceexi in pull request #1332

### DIFF
--- a/xCAT-probe/subcmds/image
+++ b/xCAT-probe/subcmds/image
@@ -71,9 +71,6 @@ unless (defined($CONSISTENCY_CHECK) || defined($DEFINITION_CHECK)) {
     exit 1;
 }
 
-my @nodes = `nodels`;
-@nodes = grep(s/^[\s]+|[\s]*$//g, @nodes);
-
 my @pingable_nodes;
 my @diskless_nodes;
 my $na = "N/A";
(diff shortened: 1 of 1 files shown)


xCAT-probe/subcmds/image
```diff
@@ -78,77 +78,134 @@ my @pingable_nodes;
...
+my %node_defined_image_uuid_hash;
+my %node_defined_image_name_hash;
+
 # First, extract diskless nodes
 foreach (@nodes) {
-    my $lsdef_provmethod = `lsdef $_ -i provmethod -c`;
-    if ($lsdef_provmethod =~ /netboot/) {
-        push(@diskless_nodes, $_);
-        probe_utils->send_msg("$output", "d", "$_ is diskless");
+    # Get osimage name for the node
```
daniceexi: If I pass in 100 nodes and all nodes use same osimage, in your code logic, it will call ``lsdef <node>`` 100 times and ``lsdef <image>`` 200 times.

The proper logic should be: run ``lsdef`` against 100 nodes and then figure out the osimage name. Then call ``lsdef`` against this osimage to get both the 'provmethod' and 'rootimgdir'. That means two call of ``lsdef`` command will be enough.

gurevichmark: Excellent suggestion. I modified the code under a new commit.

daniceexi:
@gurevichmark Your change was good for image management. But you also should change the L88 that don't call lsdef against each node, call one lsdef against all nodes instead and parse the output.

gurevichmark:
@daniceexi Thank you for the suggestion. Fixed L88.
There should be no more command calls from inside of any loop :-)

daniceexi:

PR #1407, opened 2016-06-28
save the content of the partition file to xcat.log to inspect the partition scheme after installation

immarvin:
Sometimes, we need to inspect the partition scheme in the installed system
The partition scheme can be found in /var/log/xcat/xcat.log in the installed system


daniceexi:
The change looks good for me.

PR #1408, opened 2016-06-28
[FVT]Update bundle files

caomengmeng:
@junxiawang 
Please help to review code
1. sles12.1 x86, change two cases order
2. delete sles12 ppc64 for not support just as @tingtli remainder 
3. sles12 ppc64le bundle files are not correct cases, so update all bundle files according to sles12.1 ppc64le bundle file.


tingtli:
@caomengmeng I don't think sles12 support ppc64.  Please remove sles12 ppc64 bundle
Please add the hierarchy cases in ppc64le bundle

Issue #1008, opened 2016-04-29
[FVT]:restart xcatd get "Error happened when receiving data from DB access process Magic number checking on storable file failed at /usr/lib64/perl5/vendor_perl/Storable.pm line 399, at /opt/xcat/lib/perl/xCAT/Table.pm line 127"


tingtli:
rhels7.2 ppc64le  xcat 2.12

restart xcatd and get the following log message.
Apr 29 03:15:44 c910f03c05k23 xcat[29091]: INFO xcatd: 'Command log writer' process 29091 is terminated by TERM or INT signal
Apr 29 03:15:44 c910f03c05k23 xcat[29088]: xcatd udp service 29088 quiescing
Apr 29 03:15:44 c910f03c05k23 xcat[29089]: xcatd install monitor 29089 quiescing
Apr 29 03:15:44 c910f03c05k23 xcat[29088]: xcatd: Error happened when receiving data from DB access process Magic number checking on storable file failed at /usr/lib64/perl5/vendor_perl/Storable.pm line 399, at /opt/xcat/lib/perl/xCAT/Table.pm line 127.
Apr 29 03:15:46 c910f03c05k23 xcat[29180]: INFO xcatd: command log process 29180 start

chenglch:
This is a original defect that db process exist before udp process which runs populate_site_hash to fetch data from db process at regular intervals.

tingtli:
@chenglch @daniceexi With today's build, i still get this error. But this is not always reproduced.
Jun 27 08:18:24 c910f03c04k05 xcat[6572]: xcatd: Error happened when receiving data from DB access process Magic number checking on storable file failed at /usr/lib64/perl5/vendor_perl/Storable.pm line 399, at /opt/xcat/lib/perl/xCAT/Table.pm line 131.

daniceexi:
Even we have fixed some issues related to this problem, but we still found something more need be fixed. Move it to next release.

tingtli:
verified on rhels7.2 and ubutu14.04.4. There is no error message when restart xcatd. 
root@c910f04x27v02:~# lsxcatd -v
Version 2.12.2 (git commit d3eaa578d09462577ccd43039af8ac75be3a1606, built Wed Jul 27 06:16:57 EDT 2016)

mteguhsat:
Hello, I'm having this error in xCAT 2.14. 
Anyone can advice for troubleshooting and resolve this ?

cxhong:
[...]

PR #1410, opened 2016-06-28
Wait the completion of monitor and cmdlog process

chenglch:
close-issue: #1008 #537


daniceexi:
The change looks good.

PR #1424, opened 2016-06-29
add logs in the deployment phase

penguhyang:
send msg to MN after the node deployment successfully.


daniceexi:

PR #1426, opened 2016-06-29
[go-xcat] Print more friendly error message, when command `repoquery' is not found

neo954:


tingtli:
I agree to merge

immarvin:
agree to merge

PR #1429, opened 2016-06-29
add powering-on status for PPC-HMC

xuweibj:
Normal: 
1. rpower node on ---> powering-on
2. rpower node off ---> powering-off
3. rpower node reset (current status is on) ---> powering-on
4. rpower node boot ---> powering-on
OS provision:
rnetboot node (after return SUCCESS) ---> powering-on --> installing --> booting --> booted

### DIFF
--- a/perl-xCAT/xCAT/PPCpower.pm
+++ b/perl-xCAT/xCAT/PPCpower.pm
@@ -8,6 +8,9 @@ use xCAT::Usage;
 use xCAT::MsgUtils;
 use xCAT::FSPpower;
 
+use xCAT::GlobalDef;
+use xCAT_monitoring::monitorctrl;
+
 ##########################################################################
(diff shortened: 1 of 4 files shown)


perl-xCAT/xCAT/PPCpower.pm
```diff
@@ -351,12 +402,35 @@ sub powercmd {
                  }
               }
               while (my ($name,$d) = each(%$hash) ) {
+
```
daniceexi: I think the status change for this block of code ``if (defined($request->{opt}->{m})) {`` is not necessary since this code just want to redo. redo won't change the status which has been set before.

same for the powcmd_boot

xuweibj:
Hi @daniceexi , I have modified, please help to review again. Thanks.

daniceexi:

PR #1430, opened 2016-06-29
add status powering-on for ipmi mgt

xuweibj:
Normal:
1. rpower node on ---> powering-on
2. rpower node off ---> powering-off
3. rpower node reset (current status is on) ---> powering-on
4. rpower node reset (current status is off) ---> powering-off
5. rpower node boot ---> powering-on
OS provision:
rpower node reset/boot/(off to on) ---> powering-on --> installing --> booting --> booted


daniceexi:

Issue #1419, opened 2016-06-28
Ubuntu Xenial: xcat-server needs to depend on openssh-server


torkelsson:
Trying to install xcat on Ubuntu Xenial (using core-snap) I ran into the problem that xcat-server is uninstallable because the xcatd init-script has a Required-Start: sshd, which means that the package installation of xcat-server fails unless openssh-server is already installed. However the xcat-server package does not depend on openssh-server.

whowutwut:
@immarvin @penguhyang can you take a look at this issue?

immarvin:
hi @whowutwut . I will try this in my env

immarvin:
recreated:
the log:
````

Setting up libhttp-async-perl (0.30-1) ...
Setting up libxml-parser-perl (2.44-1build1) ...
Setting up libxml-sax-expat-perl (0.40-2) ...
update-perl-sax-parsers: Registering Perl SAX parser XML::SAX::Expat with priority 50...
update-perl-sax-parsers: Updating overall Perl SAX parser modules info file...
Replacing config file /etc/perl/XML/SAX/ParserDetails.ini with new version
Setting up liblwp-protocol-https-perl (6.06-2) ...
Setting up libsoap-lite-perl (1.19-1) ...
Setting up libxmlrpc-lite-perl (0.717-1) ...
Setting up xcat-server (2.12-snap201605200311) ...
insserv: Service sshd has to be enabled to start service xcatd
insserv: exiting now!
update-rc.d: error: insserv rejected the script header
dpkg: error processing package xcat-server (--configure):
 subprocess installed post-installation script returned error exit status 1
Processing triggers for libc-bin (2.23-0ubuntu3) ...
Processing triggers for ureadahead (0.100.0-19) ...
Processing triggers for systemd (229-4ubuntu4) ...
Errors were encountered while processing:
 xcat-server
E: Sub-process /usr/bin/dpkg returned an error code (1)
````

immarvin:
````

root@c910f03fsp03v120:~# update-rc.d xcatd defaults
insserv: Service sshd has to be enabled to start service xcatd
insserv: exiting now!
update-rc.d: error: insserv rejected the script header
````

immarvin:
hi @whowutwut , I talked with @daniceexi today, we do not see it's necessary for xcatd service to depend on sshd service. 

I will remove sshd from the xcatd sysvinit file

immarvin:
[...]

PR #1432, opened 2016-06-29
fix issue  Ubuntu Xenial: xcat-server needs to depend on openssh-server #1419 ; remove sshd from xcatd sysvinit file

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/1419


daniceexi:
Base on our understanding that xcatd does not depend on the sshd, so remove the dependency of ``sshd`` when start the ``xcatd`` should be OK.

Issue #1439, opened 2016-06-30
[customer]confignics fails on CentOS 7.2


bybai:
This issue is from xcat-user.

I am having some issues with confignics not setting my install nic to a static address. I believe there are bugs with the script. It is failing with the following error:
 
analytics-n2: xcatdsklspost: downloaded postscripts successfully
analytics-n2: Fri Jun 24 14:50:50 CDT 2016 Running postscript: confignics
analytics-n2: confignics on analytics-n2: config install nic:1, remove: 0, iba ports: 
analytics-n2: ib0!172.40.21.2
analytics-n2: eno1!172.21.4.2
analytics-n2: confignics on analytics-n2: call 'configeth -s eno1
analytics-n2: configeth on analytics-n2: os type: redhat
analytics-n2: grep: 172.21.4.0: No such file or directory
analytics-n2: configeth on analytics-n2: config install nic, can not find information from dhcp lease file, return.
analytics-n2: confignics on analytics-n2: call 'configeth -s eno1
analytics-n2: configeth on analytics-n2: os type: redhat
analytics-n2: grep: 172.21.4.0: No such file or directory
analytics-n2: configeth on analytics-n2: config install nic, can not find information from dhcp lease file, return.
analytics-n2: confignics on analytics-n2: executed script: configib for nics: ib0, ports: 
analytics-n2: Postscript: confignics exited with code 0
 
 
One thing I noticed is the following line is looking for a file with the word "leases", when it's "lease" on centos 7:
str_lease_file=`ls /var/lib/dhclient/*$str_inst_nic* | grep leases`
Changing that one line yielded an almost-working static configuration. It got all the right IP information in there, but I ended up with 3 mac addresses in the HWADDR line.
 
 
Here's my lsdef and IP information of the node:
Object name: analytics-n2
    addkcmdline=nomodeset
    arch=x86_64
    currchain=boot
    currstate=boot
    groups=all,centos-analytics-df
    initrd=xcat/osimage/centos-analytics-df/initrd.img
    kcmdline=quiet inst.repo=http://172.21.0.1:80/install/centos7.2/x86_64/ inst.ks=http://172.21.0.1:80/install/autoinst/analytics-n2 ip=dhcp 
    kernel=xcat/osimage/centos-analytics-df/vmlinuz
    mac=00:21:5e:e8:8e:e0
    mgt=ipmi
    netboot=xnba
    nicips.ib0=172.40.21.2
    nicips.eno1=172.21.4.2
[...]

PR #1441, opened 2016-06-30
fix 1439 confignics fails on CentOS 7.2

bybai:
#1439


daniceexi:

PR #1442, opened 2016-06-30
fix #1427 record the status change of CN

### LINKED ISSUE #1427: [DEV] record the status change of CN in computes.log file is helpful when debug OS installation.
Currently, the CN installation information can be forwarded to MN without the CN status change.
It's can be helpful for user if they know the CN status and the CN status change time.
For example:
```
...
<node> status: booting
<node> statustime: 06-29-2016 05:51:19
...
<node> status: installing
<node> statustime: 06-29-2016 05:53:08
...
```
Also, if user use ``rpower`` command to change the node status. It should be record in computes.log file in MN.

penguhyang:
fix #1427
record the status change of CN into the ``cluster.log`` file on MN using templates of syslog.
If the templates of syslog don't support the OS version, then the message will write into the ``syslog`` or ``messages`` file...

### DIFF
--- a/xCAT-server/lib/xcat/monitoring/monitorctrl.pm
+++ b/xCAT-server/lib/xcat/monitoring/monitorctrl.pm
@@ -7,6 +7,7 @@ BEGIN
 }
 use lib "$::XCATROOT/lib/perl";
 use strict;
+use Sys::Syslog qw(:standard :extended);
 use xCAT::NodeRange;
 use xCAT::Table;
 use xCAT::MsgUtils;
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/monitoring/monitorctrl.pm
```diff
@@ -592,85 +593,96 @@ sub getNodeStatusFromNodesetState {
...
+        $mon + 1, $mday, $year + 1900,
+        $hour, $min, $sec);
+    if ($nltab) {
+        foreach (keys %status_hash) {
+            my $nodes = $status_hash{$_};
+            if (@$nodes > 0) {
+                $updates{'status'} = $_;
+                $updates{'statustime'} = $currtime;
+                my $nodestate = "@$nodes status: $updates{'status'} statustime: $updates{'statustime'}";
+                openlog('xcat', 'ndelay', 'local0');
```
immarvin: how about not using openlog/closelog. does ``syslog('local4|info', 'xcat:%s', $nodestate);``

penguhyang: It's not recommend by perl doc. http://perldoc.perl.org/Sys/Syslog.html
```
THE RULES OF SYS::SYSLOG
The Sixth Rule of Sys::Syslog is: No syslog before openlog .
```

immarvin:
seems ok for me, agree to merge

Issue #1371, opened 2016-06-23
[fvt]2.12.1 xcatprobe switch - could not get valid information while using switch-macmap to check whether the switch is OK to retrieve MAC address mapping


junxiawang:
env:linux os
build:c910f04x30v01:/opt/xcat/bin # lsdef -v
lsdef - Version 2.12.1 (git commit fddd78e1ec6a9d59bc921bef24e300e8d94f69e4, built Sun Jun 19 06:45:32 EDT 2016)
How to reproduce:
```
c910f04x30v01:/opt/xcat/bin # 

c910f04x30v01:/opt/xcat/bin # ./xcatprobe switch-macmap -c  -V
Smartmatch is experimental at ./xcatprobe line 64.
Smartmatch is experimental at ./xcatprobe line 173.
       <INFO>switch-10-4-25-1: Start to get information
       <INFO>switch-10-4-25-1: Generate SNMP session with parameter: 
       'UseNumeric' => '1'
       'SecLevel' => 'authNoPriv'
       'AuthPass' => 'passw0rd'
       'Version' => '3'
       'SecName' => 'admin'
       'AuthProto' => 'SHA'
       'DestHost' => 'switch-10-4-25-1'
       <INFO>switch-10-4-25-1: Finish to get information
[FAIL] switch-10-4-25-1 -  Failed to communicate with switch-10-4-25-1 through SNMP
c910f04x30v01:/opt/xcat/bin # tabdump switches
#switch,snmpversion,username,password,privacy,auth,linkports,sshusername,sshpassword,protocol,switchtype,comments,disable
"switch-10-4-25-1","3","admin","passw0rd",,"sha",,,,,"Juniper",,
```
The root cause for the command failed is I do not set the /etc/hosts correctly but I could not get valid information using this command

junxiawang:
@zet809 
```
in # ./xcatprobe switch-macmap -c
switch-10-4-25-1 -  Unknown user name                                                                             [FAIL]
c910f04x30v01:/opt/xcat/bin # tabdump switches
#switch,snmpversion,username,password,privacy,auth,linkports,sshusername,sshpassword,protocol,switchtype,comments,disable
"switch-10-4-25-1",,"admin","passw0rd",,"sha",,,,,"Juniper",,
````
------------------>only no snmpversion but message outputs unknow user name

**Comment by junxiawang** [comment]
@zet809 
```
c910f04x30v01:/opt/xcat/bin # tabedit switches
[...]

PR #1443, opened 2016-06-30
Fix issue 1371: xcatprobe switch - could not get valid information wh…

zet809:
…ile using switch-macmap
For issue #1371


daniceexi:

PR #1444, opened 2016-06-30
bug 1155 Initial merge of UI code; to test the process

zVMopenstack:
Change-Id: I4c594869f8919c8367174ab7610327f7f3af4e59

### DIFF
--- a/xCAT-UI/css/style.css
+++ b/xCAT-UI/css/style.css
@@ -1,439 +1,439 @@
 /*--- Dialogs ---*/
 .ui-dialog input {
-	border: solid 1px #BDBDBD;
-	font: 12px sans-serif;
+    border: solid 1px #BDBDBD;
+    font: 12px sans-serif;
 }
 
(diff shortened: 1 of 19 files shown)


daniceexi:
@zVMopenstack The pull request looks good. I did not review the detail in the files, but the commits are good for merge.

BTW, we recommend to use a separated branch for pull request instead of master branch ``zVMopenstack:master``. Using master branch would impact the later pull request. 

http://xcat-docs.readthedocs.io/en/stable/developers/github/create_branches.html

zVMopenstack:
OK, 
What we have been doing locally was to create a new branch on our 
workstation and then when pushing it up to GerritHub for review we tell it 
to use master.

Do you want the zVMopenstack to create a branch for each pull request, or 
just have one branch created off master for all the pull requests? 

Chuck Brazie 
brazie@us.ibm.com

daniceexi:
@chuckbrazie If the code has been reviewed in the GerritHub, the pull request for the github should be only used for code checkin, am I right? If yes, it would be OK to just use one branch. But I don't have experience that what will happen when you add commits in a branch which have been merged by a pull request. You could take a try.

Issue #1440, opened 2016-06-30
[FVT]Please update the "chdef template" return message


caomengmeng:
As the templates supported by xCAT are not allowed to be changed, so when try to change them, please just return clearly message just like the rmdef.
For details, see below: 

```
c910f03fsp03v02:~ # lsdef --template cec
Object name: cec
    groups=cec,all
    hcp=MANDATORY:The hardware control point(HMC) for the node
    hwtype=cec
    mgt=hmc
    mtm=MANDATORY:The machine type and model number of the node
    nodetype=ppc
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    serial=MANDATORY:The serial number of the node
    usercomment=the cec definition template
c910f03fsp03v02:~ # chdef -t node -o cec group=test
Error: 'group' is not a valid attribute name for an object type of 'node'.
Error: Skipping to the next attribute.
Error: Attribute 'groups' is not specified for node 'cec', skipping to the next node.
Error: One or more errors occured when attempting to create or modify xCAT
object definitions.
c910f03fsp03v02:~ # rmdef cec
Error: Could not find an object named 'cec' of type 'node'.
No objects have been removed from the xCAT database.
```

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1447

caomengmeng:
@immarvin 
Only cec-template works, others are not so .
After run chdef, a new node is created.
And I recheck the man page for chdef, ''' The command will also create a new definition if one doesn't exist.```, so if we allowed user to  define the same name node with template, so maybe we don't need to change the scripts.

```
[root@c910f02c008 ppc64]# lsdef --template
cec-template  (node)
hmc-template  (node)
ppc64le-template  (node)
ppc64lekvmguest-template  (node)
switch-template  (node)
x86_64-template  (node)
x86_64kvmguest-template  (node)
[root@c910f02c008 ppc64]# chdef -t node -o cec-template groups=test
Error: No object names were provided.
[root@c910f02c008 ppc64]# chdef -t node -o hmc-template groups=test
1 object definitions have been created or modified.
New object definitions 'hmc-template' have been created.
[root@c910f02c008 ppc64]# chdef -t node -o ppc64le-template  groups=test
1 object definitions have been created or modified.
New object definitions 'ppc64le-template' have been created.
[root@c910f02c008 ppc64]# lsdef
c910f02c04p09  (node)
c910f02c04p16  (node)
[...]

PR #1447, opened 2016-07-01
fix issue [FVT]Please update the 'chdef template' return message #1440

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/1440


daniceexi:

Issue #1437, opened 2016-06-30
[FVT]It is no unified for template, node, osimage while running lsdef


caomengmeng:
1. Do we need to make it clearly: the template is a specified node while using the lsdef command to show them? For example:
```
c910f03fsp03v02:~ # lsdef --template
cec  (node template)
```

This is what we see now:
It may make some confusion, what is the different actions for the node template and node if they are all node in our xCAT product

```
c910f03fsp03v02:~ # lsdef --template
cec  (node)
hmc  (node)
ppc64le  (node)
ppc64lekvmguest  (node)
switch  (node)
x86_64  (node)
x86_64kvmguest  (node)
c910f03fsp03v02:~ # lsdef -t node
c910f03c17k39  (node)
c910f03fsp03v03  (node)
c910f03fsp03v04  (node)
switch-10-4-25-1  (node)
switch-10-5-24-1  (node)
c910f03fsp03v02:~ # lsdef -t osimage
sles12.1-ppc64le-install-compute  (osimage)
sles12.1-ppc64le-install-service  (osimage)
sles12.1-ppc64le-netboot-compute  (osimage)
```

caomengmeng:
If there is another node named "cec", which is the same name with template, it will be like that:
Two node named "cec"
```
c910f03fsp03v02:~ # lsdef
c910f03c17k39  (node)
c910f03fsp03v03  (node)
c910f03fsp03v04  (node)
cec  (node)
switch-10-4-25-1  (node)
switch-10-5-24-1  (node)
c910f03fsp03v02:~ # lsdef --template
cec  (node)
hmc  (node)
ppc64le  (node)
ppc64lekvmguest  (node)
switch  (node)
x86_64  (node)
x86_64kvmguest  (node)

```

immarvin:
For your question 1:
````
    Do we need to make it clearly: the template is a specified node while using the lsdef command to show them? For example:

c910f03fsp03v02:~ # lsdef --template
cec  (node template)
````
Since templates are used to create object definitions, the type of templates should be consistent with object types.  There is no such ``node template`` type in xCAT schema.

immarvin:
It is truly confusion for ``mkdef --template`` against object and template with the same name, currently the logic is to that the existing node takes the 1st priority, then the template. To avoid this, maybe we  can consider separate ``mkdef --template`` to:
 ``mkdef --from``  to create object definitions based on the existing object
``mkdef --from-template``  to create object definitions based on the template

[...]

PR #1448, opened 2016-07-01
fix issue  [FVT]It is no unified for template, node, osimage while running lsdef #1437

immarvin:
rename the object templates to avoid conflicts between the names of objects and object templates


daniceexi:
I am good with the change.

PR #1451, opened 2016-07-01
Package xCAT probe into xCAT 

hu-weihua:
This pull request offer the xCAT probe packaging code. include both rpm and debian


hu-weihua:
@tingtli, I did test in my environment, but I am not sure if the change will effect the whole build framework, so could you review and give me comments about this part. thanks.

xCAT-probe/debian/changelog
```diff
@@ -0,0 +1,5 @@
+xcat-probe (2.7) maverick; urgency=low
+
+  * initial build
+
+ -- tonetto <tonetto@linux.vnet.ibm.com>  Thu, 22 Sep 2011 19:42:58 -0300
```
daniceexi: what does this email address ``tonetto@linux.vnet.ibm.com`` mean?

xCAT-probe/xCAT-probe.spec
```diff
@@ -0,0 +1,66 @@
...
+
+cp xcatprobe $RPM_BUILD_ROOT/%{prefix}/bin
+chmod 755 $RPM_BUILD_ROOT/%{prefix}/bin/*
+
+cp -r subcmds  $RPM_BUILD_ROOT/%{prefix}/probe/
+chmod 755 $RPM_BUILD_ROOT/%{prefix}/probe/subcmds/*
+
+cp -r lib $RPM_BUILD_ROOT/%{prefix}/probe/
+chmod 644 $RPM_BUILD_ROOT/%{prefix}/probe/lib/perl/*
+chmod 644 $RPM_BUILD_ROOT/%{prefix}/probe/lib/perl/xCAT/*
```
daniceexi: Could we use the ``chmod -R``?

build-ubunturepo
```diff
@@ -242,6 +242,12 @@ then
                     cd $file
                     dch -v $pkg_version -b -c debian/changelog $build_string
                     if [ "$target_arch" = "all" ]; then
+                        if [ $file_low = "xcat-probe" ]; then
+                            CURDIR=$(pwd)
+                            mkdir -p ${CURDIR}/xCAT-probe/lib/perl/xCAT/
+                            cp -f ${CURDIR}/perl-xCAT/xCAT/NetworkUtils.pm ${CURDIR}/xCAT-probe/lib/perl/xCAT/
```
daniceexi: Please add a comment for why we need copy these two files.

xCAT-probe/xCAT-probe.spec
```diff
@@ -0,0 +1,66 @@
+Summary: xCAT automated test tool
```
daniceexi: Change the summary to : **xCAT issue probing tool**

xCAT-probe/xCAT-probe.spec
```diff
@@ -0,0 +1,66 @@
...
+
+# AIX will build with an arch of "ppc"
+%ifos linux
+BuildArch: noarch
+%endif
+
+Provides: xCAT-probe = %{version}
+
+%description
+Provides a toolkits to help probe all the possible issues in xCAT
```
daniceexi: suggestion for the description:
~~~~
xCAT-probe provides a toolkits to probe all the potential issues for xCAT cluster
~~~~

xCAT-probe/xCAT-probe.spec
whowutwut: I agree to change the summary..  suggest "xCAT diagnostic tool"

xCAT-probe/xCAT-probe.spec
whowutwut: Suggest changing to "...probe potential issues with the xCAT cluster"

xCAT-probe/xCAT-probe.spec
```diff
@@ -0,0 +1,66 @@
...
+License: EPL
+Group: Applications/System
+Source: xCAT-probe-%(cat Version).tar.gz
+Packager: IBM Corp.
+Vendor: IBM Corp.
+Distribution: %{?_distribution:%{_distribution}}%{!?_distribution:%{_vendor}}
+Prefix: /opt/xcat
+BuildRoot: /var/tmp/%{name}-%{version}-%{release}-root
+
+# AIX will build with an arch of "ppc"
```
whowutwut: Why do we need this here to support AIX? Just incase?

xCAT-probe/xCAT-probe.spec
whowutwut: I think there's a way to specify file and directory attributes instead of doing this chmod ... Can't we also set the attributes correctly in the git repo for the file? Scripts that need to be executed should have those permissions set. Lib files should not be executable

xCAT-probe/xCAT-probe.spec
```diff
@@ -0,0 +1,66 @@
...
+rm -rf $RPM_BUILD_ROOT
+
+%files
+%defattr(-,root,root)
+#%doc LICENSE.html
+# Just package everything that has been copied into RPM_BUILD_ROOT
+%{prefix}
+
+%changelog
+* Tue Sep 14 2010 - Airong Zheng <zhengar@us.ibm.com>
```
whowutwut: Who is this and why is change log specifying this name/email?

xCAT-probe/debian/copyright
```diff
@@ -0,0 +1,88 @@
+Eclipse Public License - v 1.0
```
whowutwut: Not required to be solved in this pull request but thinking we should move these copyright files to some single place that the various packages can pull from

xCAT-probe/debian/changelog
```diff
@@ -0,0 +1,5 @@
+xcat-probe (2.7) maverick; urgency=low
```
whowutwut: What is this version and "maverick"?

makerpm
```diff
@@ -40,6 +40,12 @@ function makenoarch {
 				fi
 			fi 
 		fi
+                if [ $RPMNAME = "xCAT-probe" ]; then
+                    CURDIR=$(pwd)
+                    mkdir -p ${CURDIR}/xCAT-probe/lib/perl/xCAT/
+                    cp -f ${CURDIR}/perl-xCAT/xCAT/NetworkUtils.pm ${CURDIR}/xCAT-probe/lib/perl/xCAT/
```
whowutwut: Can this be handled with a symbolic link instead of a copy?

build-ubunturepo
whowutwut: Yes agree. Can we use a symbolic link?

whowutwut: Or have the code add the xcat lib directory to use ?

xCAT-probe/xCAT-probe.spec
hu-weihua: @whowutwut , you are right, we can set the attributes correctly in the git repo. What I thought was the files in git repo are maintained by many people, we suppose everyone operate git carefully, but accidents can't avoid all the time. So these sentences can be treated as double insurance.

build-ubunturepo
hu-weihua: @whowutwut , I tried to use symbolic link, it was my first option. But unfortunately, it didn't work. because when we run ``makerpm xCAT-prob``, all xCAT codes are Inaccessible because all xCAT probe code have been packaged to a tar file and this tar file is all the code ``makerpm`` can operated. That why I need to copy all xCAT code to ``xCAT-probe`` directory before running ``makerpm`` .

For your question ``have the code add the xcat lib directory to use``.  Depending on original design, xCAT probe should work independently, even without xCAT. So we hope the probe code can be self-contained.

xCAT-probe/debian/copyright
hu-weihua: @whowutwut, I agree with you, if we have such single copyright file, I will move this one.

xCAT-probe/debian/changelog
```diff
@@ -1,5 +1,5 @@
-xcat-probe (2.7) maverick; urgency=low
+xcat-probe (2.12) xcat; urgency=low
```
tingtli: xcat-probe (2.12) unstable; urgency=low
The third word is distribution(s), generally, it is unstable.

xCAT-probe/xCAT-probe.spec
```diff
@@ -12,15 +12,14 @@ Distribution: %{?_distribution:%{_distribution}}%{!?_distribution:%{_vendor}}
...
-# AIX will build with an arch of "ppc"
 %ifos linux
 BuildArch: noarch
 %endif
 
 Provides: xCAT-probe = %{version}
 
 %description
-Provides a toolkits to help probe all the possible issues in xCAT 
+CAT-probe provides a toolkits to probe potential issues with the xCAT cluster.
```
tingtli: xCAT-probe

xCAT-probe/debian/changelog
```diff
@@ -1,5 +1,5 @@
...
 
   * initial build
```
tingtli: * Initial build

xCAT-probe/xCAT-probe.spec
```diff
@@ -58,8 +57,8 @@ rm -rf $RPM_BUILD_ROOT
 %{prefix}
 
 %changelog
-* Tue Sep 14 2010 - Airong Zheng <zhengar@us.ibm.com>
-- "Create xCAT autotest package"
+* Fri Jul 1 2016 - huweihu <huweihua@cn.ibm.com>
```
tingtli: Fri Jul 1 2016 - huweihua <huweihua@cn.ibm.com>

tingtli:
I agree to merge

build-ubunturepo
whowutwut: @hu-weihua if we can make xcatprobe self contained it will be good. Currently it's not able to. We took the source code and dropped it into a directory in xCAT MN node and could not execute because the /opt/xcat/probe is hard code. Then the sub commands are also using XCATROOT.  If we can change this to use the pwd of the directory and relative path where the probe code is.... Will be much better.

xCAT-probe/xCAT-probe.spec
whowutwut: I understand your point but I don't agree with it because it hides to the developers what we are doing and causes mixed permission in the source tree where  files are not the correct permission.   Also if we really intended a file to not have executed permission it will be changed.  I think unit testing and better code review process now should help  catch these things or attempt to catch it 😃

Issue #1245, opened 2016-06-03
[Customer] When update xCAT, certain user cfg in rsyslog.conf is removed


daniceexi:
When run xCAT package update, the ``syslog`` script is run to configure/update the ``rsyslog.conf``. But for each run, ``syslog`` script always clean up the original configure section like following from the ``rsyslog.conf``. That's OK for xCAT since the ``syslog`` script will regenerate the configure for xCAT. 

The issue is ``syslog`` script deletes all the lines from ``# xCAT settings`` to end of the file for ``rsyslog.conf``. Then if user added anything after ``# xCAT settings``, they could be removed.

~~~~
# xCAT settings
#*.debug  /var/log/messages
~~~~

gurevichmark:
Taking this issue as Xiaopeng suggested.
My proposal is to mark the beginning and the end of the xCAT section in syslog.conf file:
```
# xCAT settings begin

our syslog entries

# xCAT settings end
```

daniceexi:
@gurevichmark sounds good.

whowutwut:
We may need to cherry pick this one into 2.11 branch when it's ready.

gurevichmark:
Fixed by pull request #1288

gurevichmark:
Victor cherry picked this into 2.11 branch

daniceexi:
@gurevichmark The fix has a shortage that if you migrated from a old version of xcat, when run syslog script against, the old configuration won'd be removed but the new cfg section will be added following.

I think we should refine this fix that remove the old configuration when adding the new one.
~~~~
# xCAT settings
~~~~

gurevichmark:
@daniceexi Of course you are right. Fixed under a new pull request #1453

daniceexi:
Add an enhancement with commit d1e6336

daniceexi:
This issue has been verified.

gurevichmark:
Cherry picked changes made by pull request #1253 into 2.11 branch using commit 96d7fb4075dd95b3c3c66941ae7c5e213c6aa2c3

PR #1453, opened 2016-07-01
Remove old version xCAT settings from rsyslog.conf file

gurevichmark:
Remove old version `# xCAT settings` tag and following line from rsyslog.conf file. Related to issue #1245


daniceexi:

Issue #921, opened 2016-04-13
[Customer]initrd stuck in loop and fails to get rootimg.gz for stateless image.


delhub:
While trying to provision a centos 6.6 stateless node, I ran into an issue where the node would proceed just fine with downloading the kernel and initrd images, but hung at getting the rootimg.gz.  After a long process of hacking the initrd-stateless.gz image to determine the root cause, I finally arrived at the /sbin/xcatroot call of /tmp/updateflag.

updateflag is an awk script found at /opt/xcat/share/xcat/netboot/rh/dracut/xcat-updateflag.  I noticed, however, that there are two versions of this script found in the  /opt/xcat/share/xcat/netboot subdirectories.  I found these via md5sum $(locate updateflag | grep opt).  By replacing that script with the alternate script, I was then able to get past the hang and the node booted fine.

Which of these scripts is the true script?  If the original script should have worked, why didn't it?  Is the problem elsewhere?

delhub:
I failed to point out that the issue seems to be isolated to RH and (possibly?) SLES OS'es.  So, it's possible the fix was realized at some earlier point in time, but the change got missed for the RH directories.

 md5sum $(locate updateflag | grep opt)
75f73c60ee564ce8b4c26cb00f28befc  /opt/xcat/share/xcat/netboot/SL/dracut/xcat-updateflag
75f73c60ee564ce8b4c26cb00f28befc  /opt/xcat/share/xcat/netboot/fedora/dracut/xcat-updateflag
75f73c60ee564ce8b4c26cb00f28befc  /opt/xcat/share/xcat/netboot/fedora/dracut_009/xcat-updateflag
75f73c60ee564ce8b4c26cb00f28befc  /opt/xcat/share/xcat/netboot/genesis/x86_64/fs/bin/updateflag.awk
75f73c60ee564ce8b4c26cb00f28befc  /opt/xcat/share/xcat/netboot/rh/dracut/xcat-updateflag
f045edad352dff88086516208603bbf3  /opt/xcat/share/xcat/netboot/rh/dracut/xcat-updateflag.sav
f045edad352dff88086516208603bbf3  /opt/xcat/share/xcat/netboot/rh/dracut_033/xcat-updateflag
f045edad352dff88086516208603bbf3  /opt/xcat/share/xcat/netboot/sles/dracut_033/xcat-updateflag

immarvin:
hi, 
[...]

PR #1454, opened 2016-07-03
fix issue  [Customer]initrd stuck in loop and fails to get rootimg.gz for stateless image. #921 @github;use /install/postscripts/updateflag.awk as the script to report status in diskless and diskful provision

immarvin:
fix https://github.com/xcat2/xcat-core/issues/921

use /install/postscripts/updateflag.awk as the script to report status in diskless and diskful provision


xCAT-server/share/xcat/install/scripts/pre.sles
```diff
@@ -19,32 +19,15 @@ fi
...
-
-        close(ns)
-
-        exit 0
-}
+(
+cat << 'EOF'
+#INCLUDE:#TABLE:site:key=installdir:value#/postscripts/updateflag.awk#
 EOF
+) > /tmp/bar.awk
```
daniceexi: could we rename bar.awk to updateflag to make it to be consistent with other OS.

immarvin:
hi @daniceexi , code updated

daniceexi:

PR #1456, opened 2016-07-04
add discid for centos 6.8 x86_64

zet809:


daniceexi:
The change looks good for checking in.

PR #1460, opened 2016-07-04
Fix issue caused by pull request #1400. Add remote URL as Ubuntu repo properly

neo954:


daniceexi:

Issue #1462, opened 2016-07-04
bmcsetup hang after  commencing transmit of discovery packets


wwzhubj:
I follower the process to use switch discovery , but bmcsetup hang after message: 
<166>Jul  4 01:32:32 xcat.genesis.dodiscovery: Network configuration complete, commencing transmit of discovery packets

```

[   14.618170] tg3 0001:01:00.2 enP1p1s0f2: renamed from eth2
 Petitboot (dev.20160629)                     8335-GTB         1005DCA
 ──────────────────────────────────────────────────────────────────────────────
  [Disk: sda2 / 65dfa5b4-f8f1-4c38-a61e-dd664f860090]
    Red Hat Enterprise Linux Server (0-rescue-d0fc706b57c14009a97a08583318abc2)
    Red Hat Enterprise Linux Server (3.10.0-327.18.2.el7.openpower4.ppc64le) 7.
  [Network: enP5p7s0f0 / 70:e2:84:14:0c:de]
    xCAT Genesis (192.168.130.1)
    xCAT Genesis (192.168.130.1)

  System information
  System configuration
  Language
  Rescan devices
  Retrieve config from URL
 *Exit to shell
───────────────────────────────────────────────────────────────
 Enter=accept, e=edit, n=new, x=exit, l=language, h=help
The system is going down NOW!t
Sent SIGTERM to all processes
Sent SIGKILL to all processes
[28416350838,3] OPAL: Trying a CPU re-init with flags: 0x1
[28549058054,3] OPAL: Trying a CPU re-init with flags: 0x1
 -> smp_release_cpus()
spinning_secondaries = 159
 <- smp_release_cpus()
 <- setup_system()
[    1.541174] OFfb vga: enable CONFIG_FB_LITTLE_ENDIAN to support this framebuffer
starting version 222
Done
<163>Jul  4 05:31:10 xcat.genesis.doxcat: BOOTIF missing, can't detect boot nic
ssh-keygen: generating new host keys: RSA DSA ECDSA ED25519
<166>Jul  4 05:31:11 xcat.genesis.doxcat: Generating private key...
<166>Jul  4 05:31:11 xcat.genesis.doxcat: Done
<166>Jul  4 05:31:11 xcat.genesis.doxcat: Setting IP via DHCP...
<166>Jul  4 05:31:11 xcat.genesis.doxcat: Acquiring network addresses..
logger: socket /dev/log: No such file or directory
[...]

PR #1465, opened 2016-07-04
fix issue 1462: bmcsetup hang after commencing transmit of discovery …

zet809:
…packets
For issue #1462 
The root reason is the hardware discovery doesn't completed, then the next task bmcsetup won't run.
The pull request is to fix the issue that hardware discovery process doesn't finished.


daniceexi:

PR #1470, opened 2016-07-04
fix rinstall bug for #1461

### LINKED ISSUE #1461: [FVT] `rinstall' did not check if `nodeset' runs successfully, but continue do the net booting.
The test environment is RHEL 7.2 runs on PowerVM. And xCAT is the latest build.

    # lsxcatd -v
    Version 2.12.1 (git commit 705c119d0f3166f0532cd401475666c3d5e11d38, built Sun Jul  3 06:45:35 EDT 2016)

The test scenario is quite simple. I stopped the DHCP server intensionally. Thus, ``nodeset`` will fail. Based on the screen output and exit code below, ``rinstall`` is failed as well at last. But while I was monitoring the virtual console output of the LPAR ``c910f02c03p26'', ``rinstall`` still try to do the net boot even if ``nodeset'' is failed. This is time wasting and can be avoid.

    # service dhcpd stop
    Redirecting to /bin/systemctl stop  dhcpd.service
    # rinstall c910f02c03p26 osimage=master_1603a
    Provision node(s): c910f02c03p26
    c910f02c03p26: install rhels7.2-ppc64-master
    dhcp server is not running.  please start the dhcp server.
    dhcp server is not running.  please start the dhcp server.
    Error: failed to run 'nodeset' against the following nodes: c910f02c03p26
    [root@c910f02c03p27 ~]# echo $?
    1

penguhyang:
fix #1461 
add verification about the dhcpd service.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/rinstall.pm
+++ b/xCAT-server/lib/xcat/plugins/rinstall.pm
@@ -321,15 +321,13 @@ sub rinstall {
 
     $rc = $::RUNCMD_RC;
     my $rsp = {};
-    if ($rc == 0) {
-        if ($VERBOSE) {
-            my @cmd = "Run command: nodeset @nodes @parameter";
-            push @{$rsp->{data}}, @cmd;
-            push @{$rsp->{data}}, @$res;
-            xCAT::MsgUtils->message("I", $rsp, $callback);
-        }
+    if ($VERBOSE) {
+        my @cmd = "Run command: nodeset @nodes @parameter";
+        push @{$rsp->{data}}, @cmd;
+        push @{$rsp->{data}}, @$res;
+        xCAT::MsgUtils->message("I", $rsp, $callback);
     }
-    else {
+    unless ($rc == 0) {
         # We got an error with the nodeset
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/rinstall.pm
```diff
@@ -342,10 +340,17 @@ sub rinstall {
                 $nodes{$successnode} = 0;
                 push @successnodes, $successnode;
             }
+            if ($line =~ /dhcp server is not running/) {
+                my $rsp = {};
+                $rsp->{error}->[0] = "dhcp server is not running.  please start the dhcp server.";
```
neo954: No need to repeat the error messages from ``nodeset``. Since ``nodeset`` already sent out its error message. Just print out some common error message, like ``Fatal error`` or something.

penguhyang: fixed.

daniceexi:
I have reviewed  the change and agree to merge. But @penguhyang need to fix some message issue later.

Issue #1471, opened 2016-07-04
[FVT]Please make the priority clearly as node name can be the same with template


caomengmeng:
As the node name can be the same with a template, please make sure the priority. 
It is still the template priority high than the node, we discussed this before and it is the node priority high than template, so how do we final decision?

```
[root@c910f04x27v05 autotest]# lsdef --template cec-template
Object name: cec-template
    groups=cec,all
    hcp=MANDATORY:The hardware control point(HMC) for the node
    hwtype=cec
    mgt=hmc
    mtm=MANDATORY:The machine type and model number of the node
    nodetype=ppc
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    serial=MANDATORY:The serial number of the node
    usercomment=the cec definition template
[root@c910f04x27v05 autotest]# mkdef -t node -o cec-template --template cec-template serial=test mtm=test hcp=test groups=test_template_priority
1 object definitions have been created or modified.
[root@c910f04x27v05 autotest]# lsdef cec-template
Object name: cec-template
    groups=test_template_priority
    hcp=test
    hwtype=cec
    mgt=hmc
    mtm=test
    nodetype=ppc
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    serial=test
    usercomment=the cec definition template
[root@c910f04x27v05 autotest]# mkdef -t node -o auto_test_cec_node --template cec-template serial=test2 mtm=test2 hcp=test2
1 object definitions have been created or modified.
[root@c910f04x27v05 autotest]#
[root@c910f04x27v05 autotest]# lsdef auto_test_cec_node
Object name: auto_test_cec_node
    groups=cec,all
    hcp=test2
    hwtype=cec
    mgt=hmc
    mtm=test2
    nodetype=ppc
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    serial=test2
    usercomment=the cec definition template
```

See another extreme example below:
```
[root@c910f04x27v05 autotest]# mkdef -t node -o switch-template --template hmc-template ip=test username=test password=test
1 object definitions have been created or modified.
[root@c910f04x27v05 autotest]# lsdef
auto_test_cec_node  (node)
c910f04x18v015  (node)
c910f04x27v06  (node)
c910f04x27v07  (node)
cec-template  (node)
switch-10-4-25-1  (node)
switch-10-5-24-1  (node)
switch-template  (node)
[...]

PR #1473, opened 2016-07-05
fix issue  [FVT]Please make the priority clearly as node name can be the same with template #1471

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1471

the updated doc:http://10.3.5.21/xcat-doc/html/guides/admin-guides/references/man1/mkdef.1.html


daniceexi:

Issue #1468, opened 2016-07-04
[DEV]Receive origin status when rpower node, cover current status


xuweibj:
```
Jul  4 03:08:58 c910f03c05k04 xcat: c910f03c17k17 status: powering-off statustime: 07-04-2016 03:08:58
Jul  4 03:08:58 c910f03c05k04 xcat: c910f03c17k17 status: booted statustime: 07-04-2016 03:08:58
```

xuweibj:
Verified.

PR #1474, opened 2016-07-05
Fix issue 1468, modified to check the value of array element

xuweibj:
#1468 
Now the hash of response array element is just like this. So modified to check the errorcode value of array element.
```
$VAR1 = {
          'node' => [
                      {
                        'errorcode' => [
                                                    0
                                            ],
                        'name' => [
                                    'c910f03c17k17'
                                  ],
                        'data' => [
                                    {
                                      'contents' => [
                                                      'off'
                                                    ]
                                    }
                                  ]
                      }
                    ]
        };
```


daniceexi:
The change looks good for merge.

PR #1477, opened 2016-07-05
Make xcat-porbe as xCAT's dependency to install xcat-porbe during xCAT installation

hu-weihua:
In order to install xcat-porbe during xCAT installation, now make xcat-probe as xCAT's dependency.


daniceexi:

PR #1481, opened 2016-07-05
Fix issue 1371: xcatprobe switch-macmap - could not get valid informa…

zet809:
…tion
For issue #1371 
2 more modifications:
1. create the link switchprobe if not exist whenever running './switch-macmap' 
2. deal with the error msg output by switchprobe

### DIFF
--- a/perl-xCAT/xCAT/MacMap.pm
+++ b/perl-xCAT/xCAT/MacMap.pm
@@ -570,10 +570,14 @@ sub getsnmpsession {
     my $snmpver='1';
     my $swent = $self->{switchparmhash}->{$switch};
 
-  if ($swent) {
-      $snmpver=$swent->{snmpversion};
-      $community=$swent->{password};
-  }
+    if ($swent) {
+        if ($swent->{snmpversion}) {
+            $snmpver=$swent->{snmpversion};
+        }
+        if ($swent->{password}) {
+            $community=$swent->{password};
+        }
+    }
   my $switch_ip = xCAT::NetworkUtils->getipaddr($switch);
   unless ($switch_ip) {
       return ({"ErrorStr"=>"Can not resolve IP address for $switch"});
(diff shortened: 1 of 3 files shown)


perl-xCAT/xCAT/MacMap.pm
```diff
@@ -570,10 +570,12 @@ sub getsnmpsession {
     my $snmpver='1';
     my $swent = $self->{switchparmhash}->{$switch};
 
-  if ($swent) {
-      $snmpver=$swent->{snmpversion};
-      $community=$swent->{password};
-  }
+    if ($swent->{snmpversion}) {
```
daniceexi: I think you should check the validity of $swent first.

zet809: Yes, modified, pls help to review again.

daniceexi:

xCAT-probe/subcmds/switch-macmap
```diff
@@ -53,13 +53,15 @@ if ($help) {
     probe_utils->send_msg("$output", "d", $::USAGE);
     exit 0;
 }
+
```
hu-weihua: 1. Line 45-48 seems to check the validity of input option, I think line 36-42 can cover this. 
2. If the rules of ``noderange`` of xCAT probe commands can comply with the rules of xCAT itself, it's easier for customer to use.

xCAT-probe/subcmds/switch-macmap
```diff
@@ -92,21 +94,25 @@ if (-f $error_file) {
 }
 my $fd;
 open($fd, "<", "$normal_file");
-my %fails = ();
+my @fails = ();
```
hu-weihua: line 78-29:  if we run ``switch-macmap`` twice at same time. that means two progress will operate ``/tmp/result_normal`` and ``/tmp/result_error``  at same time, that maybe result in unexpected problem. suggest change these two files to ``/tmp/result_normal.$$`` and ``/tmp/result_error.$$``.

Issue #1327, opened 2016-06-17
after pessl kitcomponents are removed from osimage, the INCLUDE env file for pessl are not cleaned up. 


wwzhubj:
Issue: After pessl kitcomponents are removed from osimage, the INCLUDE files related with pessl are not cleaned up. 

<[root@c712ems4 pessl]# rmkitcomp -i rh72diskfull1621 pessl-loginnode-5.3.0-0-rhels-7.2-ppc64le,pessl-license-5.3.0-0-rhels-7.2-ppc64le
Removing kitcomponent pessl-loginnode-5.3.0-0-rhels-7.2-ppc64le from osimage rh72diskfull1621
Removing kitcomponent pessl-license-5.3.0-0-rhels-7.2-ppc64le from osimage rh72diskfull1621
kitcomponents pessl-loginnode-5.3.0-0-rhels-7.2-ppc64le,pessl-license-5.3.0-0-rhels-7.2-ppc64le were removed from osimage rh72diskfull1621 successfully

[root@c712ems4 pessl]# cat /install/osimages/rh72diskfull1621/kits/KIT_DEPLOY_PARAMS.otherpkgs.pkglist
#INCLUDE:/install/kits/essl-5.5.0-0-ppc64le/other_files/essl-5.5.0-0-ppc64le_essl.env#
#INCLUDE:/install/kits/xlf-15.1.3-0-ppc64le/other_files/xlf-15.1.3-0-ppc64le_xlf.env#
#INCLUDE:/install/kits/xlc-13.1.3-0-ppc64le/other_files/xlc-13.1.3-0-ppc64le_xlc.env#
#INCLUDE:/install/kits/pessl-5.3.0-0-ppc64le/other_files/pessl-5.3.0-0-ppc64le_pessl.env#>

wwzhubj:
[root@c712ems4 pessl]# lsxcatd -v
Version 2.12 (git commit 5df15f39216da9c64759d61b19fbabad1674727d, built Wed May 18 07:30:41 EDT 2016)

bybai:
Hi @wwzhubj , I cannot reproduce it.

wwzhubj:
How to recreate it: 
1. addkit pessl-5.2.0-0-ppc64le.tar.bz2  
2.

wwzhubj:
[...]

PR #1486, opened 2016-07-06
fix 1327 after pessl kitcomponents are removed from osimage, the INCL…

bybai:
…UDE env file for pessl are not cleaned up.

#1327


daniceexi:

Issue #1479, opened 2016-07-05
[fvt]2.12.2 xcatprobe osdeploy still show osdeply information even though just rpower node boot


junxiawang:
env:linux 
build:root@c910f03c05k05:~# lsxcatd  -v
Version 2.12.1 (git commit ccf065943098e3c2162411e6e51e37c999fbd09f, built Mon Jul  4 22:52:11 EDT 2016)
How to reproduce:
rpower all boot
xcatprobe osdeploy -n all
------------------->Here just rpower boot no osdeply but still shows:
```
 Recv from c910f03c05k09 : xcat:  ready                                                                            
Recv from c910f03c05k09 : xcat:  done                                                                             
Recv from c910f03c05k09 : xcat: status is booted                                                                  
Node c910f03c05k09 has finished it's os provision process                                                         [ OK ]
All nodes need to monitor have finished os provision process                                                      [ OK ]

=============================================================
=  The summary of os provision:
=============================================================

```

junxiawang:
@xuweibj @hu-weihua 
discussed with wihua move this bug to 2.12.3

hu-weihua:
@junxiawang , could you verify again? thanks

PR #1488, opened 2016-07-06
Fix issue I1479, judge if the node do os provision

xuweibj:
#1479 
1. If get status of installing, set the  $monitor_nodes{$node}{"status"} "installing".
2. if get status of booted, check whether $monitor_nodes{$node}{"status"} is "installing", if yes, it's finished it's os provision. If not, the node did not do os provision.

### DIFF
--- a/xCAT-probe/subcmds/osdeploy
+++ b/xCAT-probe/subcmds/osdeploy
@@ -58,7 +58,7 @@ sub check_noderange{
             $nodecheckrst{$currentnode}{"error"} = "Could not find node definition";
             $rst = 1;
         } elsif ($_ =~ /^\s*Object name: (\w+)/i) {
-            $monitor_nodes{$1} = 0; 
+            $monitor_nodes{$1}{"rst"} = 0; 
             $currentnode = $1;
             $ip = "NOIP";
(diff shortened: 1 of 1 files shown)


xCAT-probe/subcmds/osdeploy
```diff
@@ -230,14 +230,27 @@ sub handle_cluster_msg {
...
+                probe_utils->send_msg("$output", "f", "$record");
+                push(@{ $rawdata{ $node }{"history"} }, $record);
+            }
+            $monitor_nodes{$node}{"rst"} = 1 if (defined($monitor_nodes{$node}));
         } elsif (exists($rawdata{$node}) and ($status eq "failed")) {
-            $monitor_nodes{$node} = 1 if (defined($monitor_nodes{$node}));
+            $monitor_nodes{$node}{"rst"} = 1 if (defined($monitor_nodes{$node}));
             probe_utils->send_msg("$output", "f", "Node $node has finished it's os provision process");
+            push(@{ $rawdata{ $node }{"history"} }, "Node $node os provision failed");
+        } elsif (exists($rawdata{$node}) and ($status eq "installing")) {
```
zet809: Shall not output the flag "installing"?

PR #1492, opened 2016-07-07
fix bug, update ipnodemap

xuweibj:
Update $ipnodemap, if get dhcpack.


daniceexi:

Issue #1493, opened 2016-07-07
[FVT]:sles11.4 x86_64 provision loops and does not finish


tingtli:
xCAT 2.12.1 sles11.4 x86_64
Provision sles11.4 on x86_64 and find that the provision does not end.

```
c910f04x30v01:/var/log/xcat # lsdef -l c910f04x30v11
Object name: c910f04x30v11
    arch=x86_64
    cons=kvm
    currchain=boot
    currstate=install sles11.4-x86_64-compute
    groups=all
    initrd=xcat/osimage/sles11.4-x86_64-install-compute/initrd
    ip=10.4.30.11
    kcmdline=quiet autoyast=http://!myipfn!:80/install/autoinst/c910f04x30v11 install=http://!myipfn!:80/install/sles11.4/x86_64/1 netdevice=42:f2:0a:04:1e:0b  console=tty0 console=ttyS0,115200n8r
    kernel=xcat/osimage/sles11.4-x86_64-install-compute/linux
    mac=42:f2:0a:04:1e:0b
    mgt=kvm
    netboot=xnba
    os=sles11.4
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    profile=compute
    provmethod=sles11.4-x86_64-install-compute
    serialflow=hard
    serialport=0
    serialspeed=115200
    status=powering-on
    statustime=07-07-2016 06:02:39
    vmcpus=2
    vmhost=c910f04x30
    vmmemory=4096
    vmnics=virbr1
    vmstorage=dir:///var/lib/libvirt/images/
```
The console shows it finishes the installation but again to download the installation images and initrd

```
xCAT Network Boot Agent
gPXE 1.0.3-131028 (d603e) -- Open Source Network Boot Firmware -- http://ipxe.or
Features: HTTP HTTPS iSCSI DNS TFTP bzImage ELF PXE PXEXT
net0: 42:f2:0a:04:1e:0b using undionly on UNDI-PCI00:03.0 (open)
  [Link:up, TX:0 TXE:0 RX:0 RXE:0]
DHCP (net0 42:f2:0a:04:1e:0b)... ok
net0: 10.4.30.11/255.0.0.0 gw 10.0.0.101
Next server: 10.4.30.2
Filename: http://10.4.30.2/tftpboot/xcat/xnba/nodes/c910f04x30v11
http://10.4.30.2/tftpboot/xcat/xnba/nodes/c910f04x30v11... ok
[...]

PR #1497, opened 2016-07-07
issue 1493, we made a change that include the updateflag from /instal…

daniceexi:
…l/postscripts/updateflag.awk. But the shebang in /install/postscripts/updateflag.awk points to the /usr/bin/awk which does not exist in sles11.x installer. This fix is to make a link to the installer shipped awk.


daniceexi:
fix issue #1493

neo954:
I suggest update the shebang to ``#!/usr/bin/env awk`` instead of create a symbolic link of ``/usr/bin/awk``.

neo954:
Okay, will merge this one, anyway.

PR #1499, opened 2016-07-07
Fix the issue that discovery request is cut when sending to xcatd

zet809:


daniceexi:

PR #1503, opened 2016-07-07
xCAT-server: Fix exclude list for RH7 - Vim >= 7.4

samveen:
Exclude lists are outdated. RH7 has Vim-minimal-7.4-*, not Vim-minimal-7.2-*


immarvin:
since xcat2.12.1 will be released, move to next release

immarvin:
agree to merge

PR #1505, opened 2016-07-08
Only try to kill ssl listener process at first

chenglch:
Previously service xcatd stop or systemctl stop xcatd.service
will send TERM signal to all the processes of xcatd on some
platform which support this feature.

Close-issue: #1008 #537


daniceexi:

Issue #1509, opened 2016-07-08
Updating xCAT on OS that uses systemctl has a warning message 


whowutwut:
When updating xCAT on RHEL 7.2 where systemctl is being used, there's a warning message

```
Running transaction
  Updating   : 4:xCAT-client-2.12.1-snap201607070618.noarch                                                       1/14 
  Updating   : 4:xCAT-server-2.12.1-snap201607070618.noarch                                                       2/14 
  Updating   : 4:perl-xCAT-2.12.1-snap201607070618.noarch                                                         3/14 
  Updating   : 4:xCAT-probe-2.12.1-snap201607070618.noarch                                                        4/14 
  Updating   : 4:xCAT-buildkit-2.12.1-snap201607070618.noarch                                                     5/14 
  Updating   : 1:xCAT-genesis-scripts-ppc64-2.12.1-snap201607070618.noarch                                        6/14 
If you are installing/updating xCAT-genesis-base separately, not as part of installing/updating all of xCAT, run 'mknb <arch>' manually
  Updating   : xCAT-2.12.1-snap201607070618.ppc64le                                                               7/14 
Restarting xcatd (via systemctl):  Warning: xcatd.service changed on disk. Run 'systemctl daemon-reload' to reload units.
[  OK  ]
Running '/opt/xcat/sbin/mknb ppc64', triggered by the installation/update of xCAT-genesis-scripts-ppc64 ...
Creating genesis.fs.ppc64.gz in /tftpboot/xcat
The 'mknb ppc64' command completed successfully.

```

Opening this issue to track whether or not this is a concern that we should address.

daniceexi:
quick solution is to run ``systemctl daemon-reload`` in the xcatconfig when install or update xCAT

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1510

whowutwut:
@immarvin  I just did a `yum update '*xCAT*'` on a 2.12.1 build (July 5) -> 2.12.2 (July 25 build) and still see the warning msg... 

```
Restarting xcatd (via systemctl):  Warning: xcatd.service changed on disk. Run 'systemctl daemon-reload' to reload units.
[  OK  ]
```

immarvin:
hi @whowutwut , the fix was overwritten by the PR https://github.com/xcat2/xcat-core/pull/1537, I think we'd better avoid the things like this in the following commit and review process

immarvin:
[...]

PR #1510, opened 2016-07-11
fix issue  Updating xCAT on OS that uses systemctl has a warning message #1509 ; run systemctl daemon-reload on xcat upgrade

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1509


daniceexi:

PR #1515, opened 2016-07-11
[go-xcat] Print out more error messages when reading repository failed.

neo954:


daniceexi:

PR #1517, opened 2016-07-11
restapirun to run system command part.

bybai:
1， Add plugin to execute system command part for rest api.
2, add getifaces in xcatws.cgi to get network interfaces.
Example:
]# curl -X GET -k 'https://xcatui/xcatws/services/getifaces?userName=root&userPW=cluster'
["eth0","lo"]

### DIFF
--- a/xCAT-server/lib/xcat/plugins/runsyscmd.pm
+++ b/xCAT-server/lib/xcat/plugins/runsyscmd.pm
@@ -14,32 +14,19 @@ BEGIN
 }
 use lib "$::XCATROOT/lib/perl";
 
-use IO::Socket;
-use Thread qw(yield);
-use POSIX "WNOHANG";
-use Storable qw(store_fd fd_retrieve);
 use strict;
-use warnings "all";
 use Getopt::Long;
-use xCAT::Table;
 use xCAT::Utils;
 use xCAT::MsgUtils;
 use Getopt::Long;
-use Data::Dumper;
-use File::Basename;
-use File::Path;
-use Cwd;
 
-my $nmap_path;
 
 my $debianflag = 0;
 my $tempstring = xCAT::Utils->osver();
 if ( $tempstring =~ /debian/ || $tempstring =~ /ubuntu/ ){
     $debianflag = 1;
 }
-my $parent_fd;
-my $bmc_user;
-my $bmc_pass;
+
 #-------------------------------------------------------
 
 =head3  handled_commands
@@ -74,15 +61,6 @@ sub process_request
     my $request_command = shift;
     $::CALLBACK = $callback;
 
-    unless(defined($request->{arg})){
-        runsyscmd_usage();
-        return 2; 
-    }
-    @ARGV = @{$request->{arg}};
-    if($#ARGV eq -1){
-            return 2;
-    }
-
 
     my $command  = $request->{command}->[0];
     my $rc;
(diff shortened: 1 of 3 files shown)


xCAT-server/lib/xcat/plugins/runsyscmd.pm
```diff
@@ -0,0 +1,293 @@
+# IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
+#-------------------------------------------------------
+
+=head1
+  xCAT plugin package to handle system command
+=cut
+
+#-------------------------------------------------------
+package xCAT_plugin::runsyscmd;
+
```
daniceexi: Please add comments for the purpose of this plugin

xCAT-server/lib/xcat/plugins/runsyscmd.pm
```diff
@@ -0,0 +1,293 @@
...
+use xCAT::MsgUtils;
+use Getopt::Long;
+use Data::Dumper;
+use File::Basename;
+use File::Path;
+use Cwd;
+
+my $nmap_path;
+
+my $debianflag = 0;
```
daniceexi: I did not see you are using this $debianflag variable.

xCAT-server/lib/xcat/plugins/runsyscmd.pm
```diff
@@ -0,0 +1,293 @@
...
+
+Return list of commands handled by this plugin
+
+=cut
+
+#-------------------------------------------------------
+
+sub handled_commands
+{
+    return {
```
daniceexi: please fix the indent

xCAT-server/lib/xcat/plugins/runsyscmd.pm
```diff
@@ -0,0 +1,293 @@
...
+    my $request_command = shift;
+    $::CALLBACK = $callback;
+
+    unless(defined($request->{arg})){
+        runsyscmd_usage();
+        return 2; 
+    }
+    @ARGV = @{$request->{arg}};
+    if($#ARGV eq -1){
+            return 2;
```
daniceexi: indent

whowutwut:
@bybai  Is this a new external command for users?  Why would we need to run this?

chenglch:
I don't think this is a general interface that we want. We need a plugin for rest api just because our cgi script is running with apache user who do not have permission to access some files( credential files) or execute some operation with privilege. Running command is a scenario but I think is not all, plus I don't think call the shell command is good way unless we have to.

bybai:
@chenglch , runsyscmd is to run system command or other command. I will use this in xcatws.cgi to realize some system command restapi, such as get network adapters. Currently, just "ls" command can run after runsyscmd.

chenglch:
@bybai , as I said on last comment , running command is just a part of work for rest api required by xcat web GUI. Your implementation is not a general interface to communicate with CGI, I have list the usage above. In addition, you should update xcatws.cgi to  see how it works on restapi level, you can make`ls` command as an example.

bybai:
@chenglch , you can regard this command is a common xcat command.  
If you want to add rest api in xcatws.cgi, you should write your own handler in it.  
Here is an example for get network adapters in xcatws.cgi to use:
runsyscmd "ls /sys/class/net"

I think the command is simple enough for xcatws.cgi to call.

chenglch:
@bybai, As a base framework, you should add basic or abstract implementation to let the next developer know how to use your framework. From my point view, the rest api part is not complete. 

Running shell command do not solve every thing, it is just a part of the framework to work with cgi, a more general method is needed. I have list the reason above.   
Take your `ls /sys/class/net` command as a example Perl have its own library to list directory or read the file, it is more efficient and support multi-platform we only call the shell command when we have to.

bybai:
this pull request is abandoned.

PR #1518, opened 2016-07-11
Noderange and error checking

gurevichmark:
Add noderange to image probe and add some error checking to display useful messages.

### DIFF
--- a/xCAT-probe/subcmds/image
+++ b/xCAT-probe/subcmds/image
@@ -22,7 +22,7 @@ my $rst     = 0;
 $::USAGE = "Usage:
     $program_name -h
     $program_name -T
-    $program_name {-c|-d} [-V]
+    $program_name {-c|-d} [-V] [noderange]
 
 Description:
     Use this command to check if diskless, pingable compute nodes have the same images installed as defines in xCAT DB. 
@@ -61,7 +61,7 @@ if ($help) {
 }
 
 if ($test) {
-    probe_utils->send_msg("$output", "o", "Use this command to check if all compute nodes have the same images installed or if compute nodes are installed with the same image as defined on MN.");
+    probe_utils->send_msg("$output", "o", "Use this command to check if specified compute nodes have the same images installed or if compute nodes are installed with the same image as defined on MN.");
     exit 0;
 }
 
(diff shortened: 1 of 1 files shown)


xCAT-probe/subcmds/image
```diff
@@ -22,7 +22,7 @@ my $rst     = 0;
 $::USAGE = "Usage:
     $program_name -h
     $program_name -T
-    $program_name {-c|-d} [-V]
+    $program_name {-c|-d} [-V] [noderange]
```
daniceexi: Generally, for the xcatprobe subcommand, the ``noderange`` is optional, so we are using ``-n`` to specify the noderange instead of the first non-flag parameter.

So I prefer we also use ``-n noderange``

xCAT-probe/subcmds/image
```diff
@@ -80,12 +80,25 @@ my %node_defined_image_uuid_hash;
 my %node_defined_image_name_hash;
 my %osimage_defined_provmethod_hash;
 
-my $all_nodes_provmethod = `lsdef -i provmethod -c`;
+my $all_nodes_provmethod = `lsdef -i provmethod -c @ARGV`;
```
daniceexi: use ``@ARGV`` is not safe since it would take all the available arguments. If using ``-n noderange`` would fix this problem.

daniceexi:
@gurevichmark Beside the comments I added in the commit, I have a comment for ``-d``. 
For the ``-d``, the image subcmd will display a message for each node when the checking succeeded. My comment is we can ignore the success nodes or use one message to say **all the nodes passed the checking** if no failure. If there are failed node, we just display the failed ones.

xCAT-probe/subcmds/image
gurevichmark: Fixed in the new commit.

xCAT-probe/subcmds/image
gurevichmark: Fixed in the new commit.

gurevichmark:
@daniceexi I understand your suggestion, about only displaying the nodes with failure, but I am not sure if I agree. I think it is useful to display each item being probed, in this case, each node.

If some nodes failed and some were ok, with your suggestion, we would display each failed node, but then for successful ones, would we say "The rest of the probes nodes were ok" ?  Would the user wonder which nodes we tested that were ok ?

Also, looking at the example of the `xcatmn` probe, we list each sub process that we tested, even if they are all ok:

> 
[root@fs2vm100 subcmds]# ../xcatprobe xcatmn -n eth3
[ OK ] NIC eth3 exists on current server
[ OK ] Get ip address of NIC eth3
[ OK ] Sub process 'xcatd: SSL listener' is running
[ OK ] Sub process 'xcatd: DB Access' is running
[ OK ] Sub process 'xcatd: UDP listener' is running
[ OK ] Sub process 'xcatd: install monitor' is running
[ OK ] Sub process 'xcatd: Discovery worker' is running
[ OK ] Sub process 'xcatd: Command log writer' is running
[ OK ] xcatd is listening on port 3001
[ OK ] xcatd is listening on port 3002
<

daniceexi:
@gurevichmark I agree with all you said. My concern was what happens when I use ``xcatprobe image`` to probe thousands of nodes. If I have several nodes fail in thousands of nodes, how could I found the failed nodes easily from thousands lines of output?

My suggestion is we just have two type of output:

* success for all
[ok] all nodes passed the check

* one or more failed
[fail] node1 failed 
[fail] node2 failed

gurevichmark:
@daniceexi Ok, I understand. In the second case, where one or more nodes have failed, we would not display anything for nodes that were **OK** ?

daniceexi:
@gurevichmark Yes, it was my idea.

gurevichmark:
@daniceexi Made another commit to:
1. Display success message only if all nodes are OK.
2. Display success message for individual nodes if `VERBOSE` option is turned on.

daniceexi:

Issue #1495, opened 2016-07-07
[PCM] Provision Dell PowerEdge M605 compute node hangs while setup BMC


AsirXing:
I am tried to provision Dell PowerEdge M605 compute node with xCAT 2.12, but the provisioning is hang while setup BMC. Below are details for this issue.
**1. The xCAT version:**
...............
[root@db04b11 ~]# lsxcatd -a
Version 2.12 (git commit b33cb213ce79f91621cd65932c9798c735b74bc4, built Fri Jun 24 03:17:21 EDT 2016)
This is a Management Node
cfgloc=Pg:dbname=xcatdb;host=192.168.108.80|xcatadm
dbengine=Pg
dbname=xcatdb
dbhost=192.168.108.80
dbadmin=xcatadm
[root@db04b11 ~]#
...............

**2. The compute node hardware model: Dell PowerEdge M605**

**3. The outputs of ipmitool commands in genesis**
...............
[xCAT Genesis running on db04b12 /tmp]# ipmitool lan print 1
No data available
Get Device ID command failed
No data available
Get Channel Info command failed: Invalid command
Invalid channel: 1
[xCAT Genesis running on db04b12 /tmp]#
...............

**4. The processes tree in genesis** 
...............
  307 ?        S      0:00 screen -L -ln doxcat
  308 ?        Ss     0:00  \_ SCREEN -L -ln doxcat
  309 pts/1    Ss+    0:00      \_ /bin/sh /usr/bin/doxcat
  442 pts/1    S+     0:01          \_ /bin/sh /usr/bin/doxcat
24492 pts/1    S+     0:00          |   \_ sleep 5
  443 pts/1    S+     0:01          \_ /bin/sh /usr/bin/doxcat
24493 pts/1    S+     0:00          |   \_ sleep 5
 1102 pts/1    S+     0:00          \_ /bin/sh /usr/bin/doxcat
24435 pts/1    S+     0:00              \_ /bin/sh /usr/bin/doxcat
24436 pts/1    S+     0:00                  \_ ipmitool raw 0x3a 0x0a
...............

daniceexi:
@AsirXing There's a commit 195f80ee6e544d500545dd00e7b9357498ba872f for Dell server support. You could take a try. 

You can just copy the bmcsetup script from master to 2.12-pcm build and take a try.

zet809:
# ipmitool mc info
Device ID                 : 32
Device Revision           : 1
Firmware Revision         : 1.65
[...]

PR #1522, opened 2016-07-12
Fix issue 1495: Provision Dell PowerEdge M605 compute node hangs whil…

zet809:
…e setup BMC

For issue #1495


daniceexi:

PR #1526, opened 2016-07-13
Improve the performance for ipmi mgt(powering-on)

xuweibj:
Use a global hash ``%newnodestatus`` to save nodes' new status.
If there is node's status needed to be updated, update it.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -37,6 +37,7 @@ use Time::HiRes qw/time/;
 my $iem_support;
 my $vpdhash;
 my %allerrornodes=();
+my %newnodestatus=();
 my $global_sessdata;
 my %child_pids;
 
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -7950,20 +7935,8 @@ sub process_request {
...
-    	    if (exists($old{$stat})) {
-    		    my $pa=$old{$stat};
-        		push(@$pa, $node);
-    	    } else {
-          		$old{$stat}=[$node];
-	        }
-        } 
-        xCAT_monitoring::monitorctrl::setNodeStatusAttributes(\%old, 1);
+    if ($command eq 'rpower' and %newnodestatus) {
+        xCAT_monitoring::monitorctrl::setNodeStatusAttributes(\%newnodestatus, 1);
```
daniceexi: @xuweibj I think you should just change the status for **success node** and keep the old status for failed node.

daniceexi:

PR #1527, opened 2016-07-13
[doc] Fix a typo in xCAT 2.12.x Release Information

neo954:


daniceexi:
The change looks good.

whowutwut:
@neo954  Thanks! For these small changes, I think we can also directly modify the files from the GitHub UI rather than create pull requests.

daniceexi:
Yes, I agree.

PR #1533, opened 2016-07-14
Improve performance for hmc mgt

xuweibj:
Modified hash ``%newnodestatus`` value to array.


daniceexi:

PR #1534, opened 2016-07-14
Improve performance of blade (powering-on)

xuweibj:
Update all nodes' status, not for one node.


daniceexi:

PR #1535, opened 2016-07-14
add genesis testcase for autotest

junxiawang:
add genesis testcase for autotest


neo954:
Okay, looks good. Will merge it.

PR #1536, opened 2016-07-14
fix issue  geninitrd fails when pkgdir has more than one directory #1538 ; fix  geninitrd fails to identify a ppc64le kernel #1539

immarvin:
the ``geninitrd`` command failed with 
````
[root@smn ~]# geninitrd rhels7.2-ppc64le-stateful-compute_kernel_update
Error: Could not get the format of the initrd.
````


daniceexi:

PR #1541, opened 2016-07-15
Fix bug, probe discovery get node ip

xuweibj:
If there is no ip, set nodeip to null.


hu-weihua:

Issue #1090, opened 2016-05-11
conserver looses console output with ppc64le firestone (OpenPower S822LC) systems


ralphbellofatto:
We are running xcat with IBM openpower firestone systems.  

xcat is configured with
```
 lsdef -t site clustersite -i consoleondemand
consoleondemand=no
```

if we execute 
```
service conserver restart
rcons <nodeid>
```
or 
```
service conserver stop
service conserver start
rcons <nodeid>
```

rcons will continue to work until 1 minute after we issued the "service conserver stop", or the "service conserver restart" commands.

We managed to duplicate the problem without conserver with the following script.

```
/opt/xcat/bin/ipmitool-xcat -v -e !  -I lanplus -U ADMIN -P xxxx  -H fstgb002-bmc sol activat &
sleep 5;
pid=$!;
kill $pid
/opt/xcat/bin/ipmitool-xcat -v -e !  -I lanplus -U ADMIN -P xxxx -H fstgb002-bmc sol deactivate
/opt/xcat/bin/ipmitool-xcat -v -e !  -I lanplus -U ADMIN -P xxxx -H fstgb002-bmc sol activate
```
```
Wait one minute and we see:
IPMI Request Match NOT FOUND
```

We believe this is what is happening with the conserver stop/start sequence. and a close approximation of what happens with the conserver restart sequence. NOTE: the -v option is needed to see the Match NOT FOUND message.

we observe that the deactivate and activate sequence is part of the "/opt/xcat/share/xcat/cons/ipmi" file.
```
if ($iface eq "lanplus") {
    system "$ipmitool -I lanplus $inteloption $user $pass -H $bmc $solcom deactivate"; #Stop any active session
}
exec "$ipmitool -I $iface $inteloption $user $pass -H $bmc $solcom activate";
````

We have looked inside the conserver source code and observed it does a killproc on all the child processes for "service conserver stop" and does a kill HUP for the "service conserver restart".

[...]

PR #1542, opened 2016-07-18
modify xcat to depended on ipmitool-xcat-1.8.15-2 and higher

zet809:
Fix issue #1090


daniceexi:

whowutwut:
Will try and re-create the `xcat-dep` package today because it will break installing of xCAT without the newer ipmitool version.

Issue #1404, opened 2016-06-27
Include git commit number in xCAT daily autobuild


neo954:
Currently, xCAT has the version number plus a time tag in its package name. The package follows current scheme looks like ``xCAT-2.12.1-snap201606260648.ppc64.rpm``. I would like to have the git commit number be a part of the package name. Something like ``xCAT-2.12.1-snap201606260648-f203009.ppc64.rpm`` or similar would be fine.

neo954:
The main reason I want this feature is due to the requirement of xCAT Jenkins Test Report. The xCAT Jenkins Test Report is based on the log analyzing of the xCAT auto-daily-test. But from the test log. There is no way to tell which branch or which git commit number the build is based. Or even worse, there is no way to tell if the build for Red Had and build for Ubuntu are on the same git commit number.

daniceexi:
I prefer to add some messages by the automation framework to indicate the informations instead of doing that in the rpm itself.

daniceexi:
As we discussed before, we might add a file named 'building' in the tar.bz2 file to include the version number, last commit id, build time, build server info. And also the commit id can be added in the tar file name.

Then user can have following three ways to get the version and commit id for a build:
* before downloading the tar: tar file name
* before install the packages: tar file name or the 'buildinfo' file
* after install: lsxcatd -a

neo954:
Using a seperate file named ``building`` in the tarball may also acceptable. In the point of view of the Log Analyzer, it just need the git commit number appeared in the log file. Any place in the log file will do the work.

whowutwut:
Sorry for the delay, I'm making a change and will create a pull request to address this

neo954:
@tingtli, since this feature is already available. Please consider update our Jenkins test script, and include this git commit number in the xCAT Jenkins test log files.

whowutwut:
I think we can close out this issue

PR #1550, opened 2016-07-18
Add a buildinfo file to the xcat core tar.bz2 file to provide information about the build

whowutwut:
Address isssue #1404, I created a file that contains information similar to the following: 

```
VERSION=2.12.2
BUILD_TIME=Mon Jul 18 13:03:51 EDT 2016
BUILD_MACHINE=victorvm7.pok.ibm.com
COMMIT_ID=7efd570

```
It has the short commit id...


daniceexi:
The change looks good. 

I have discussed with @hu-weihua that we will change the Jenkins to add the ``commit id`` in the test log so that the report script can read the build ID from the log.

PR #1551, opened 2016-07-19
Improve performance of kvm (powering-on)

xuweibj:
Set node status for each hyp.


daniceexi:

PR #1552, opened 2016-07-19
add the notification that need to install newer mlxen driver for xcat…

zet809:
…-genesis-base-ppc64


daniceexi:

PR #1543, opened 2016-07-18
Modify xcat-genesis-scripts-ppc64 depend on xcat-genesis-base-ppc64 a…

zet809:
…nd xcat-genesis-scripts-x86-64 depend on xcat-genesis-base-x86-64 for ubuntu

Fix issue #1074

PR #1561, opened 2016-07-20
modify dependency xcat-genesis-scripts to xcat-genesis-scripts-ppc64 …

zet809:
…and xcat-genesis-scripts-x86-64 for xcatsn and xcat
Appended for #1543


immarvin:
merge first to make build for migration test

Issue #1562, opened 2016-07-20
[customer] the rhels7.2 diskful installation with kernel and kernel modules updated by"geninitrd" failed with "Pane is dead" inside anaconda


immarvin:
The installation failed with:
````
=====================

Starting installer, one moment...
libkmod: kmod_module_insert_module: could not find module by name='bnx2i'
libkmod: kmod_module_insert_module: could not find module by name='bnx2i'
libkmod: kmod_module_insert_module: could not find module by name='bnx2i'
libkmod: kmod_module_insert_module: could not find module by name='bnx2i'
anaconda 21.48.22.56-1 for Red Hat Enterprise Linux 7.2 started.
 * installation log files are stored in /tmp during the installation
 * shell is available on TTY2
  File "/sbin/anaconda", line 1156, in <module>
    setupDisplay(anaconda, opts, addon_paths)
  File "/sbin/anaconda", line 680, in setupDisplay
    anaconda.initInterface(addons)
  File "/usr/lib64/python2.7/site-packages/pyanaconda/anaconda.py", line 220, in initInterface
    self._intf = TextUserInterface(self.storage, self.payload,
  File "/usr/lib64/python2.7/site-packages/pyanaconda/anaconda.py", line 168, in storage
    self._storage.setDefaultFSType(self.instClass.defaultFS)
  File "/usr/lib/python2.7/site-packages/blivet/__init__.py", line 1751, in setDefaultFSType
    self._check_valid_fstype(newtype)
  File "/usr/lib/python2.7/site-packages/blivet/__init__.py", line 1726, in _check_valid_fstype
    raise ValueError("new value %s is not valid as a default fs type" % fmt)
ValueError: new value non-existent xfs filesystem is not valid as a default fs type
Exception AttributeError: "'NoneType' object has no attribute 'udev_unref'" in <bound method Context.__del__ of <pyudev.core.Context object at 0x3fff8b7e0c50>> ignored

Pane is dead
  File "/usr/lib/python2.7/site-packages/blivet/__init__.py", line 1726, in _check_valid_fstype
    raise ValueError("new value %s is not valid as a default fs type" % fmt)
ValueError: new value non-existent xfs filesystem is not valid as a default fs type
Exception AttributeError: "'NoneType' object has no attribute 'udev_unref'" in <bound method Context.__del__ of <pyudev.core.Context object at 0x3fff8b7e0c50>> ignored
Pane is dead
===================== 
````

immarvin:
[...]

PR #1563, opened 2016-07-20
fix issue  [customer] the rhels7.2 diskful installation with kernel and kernel modules updated by"geninitrd" failed with "Pane is dead" inside anaconda #1562; add a dracut hook in intrd to update the kernle modules inside the stage2 installer LiveOs

immarvin:
fix issue: https://github.com/xcat2/xcat-core/issues/1562

add a dracut hook in intrd to update the kernle modules inside the stage2 installer LiveOs


daniceexi:

PR #1564, opened 2016-07-20
modify genesis scripts spec for debian/control to be debian/control-a…

zet809:
…md64 and debian/control-ppc64elwq


immarvin:
agree to merge

PR #1565, opened 2016-07-20
perltidy all perl files

gurevichmark:
All .pm and .pl files were processed by this command:
`perltidy -w -i=4 -nt -io -nbbc -kbl=2 -aws -pt=2 -bbc -nolc`


daniceexi:
I did a quick review about 10,000 lines of the change and looks like the perltidy works well. So I merge the change first, if anyone found any format related issue, we could fix them later.

Issue #1575, opened 2016-07-22
[DEV] "makehosts"  generates the /etc/hosts entries with wrong domain name for node with "otherinterfaces" attribute


immarvin:
````
[root@c910f03c05k21 ~]# lsxcatd -v
Version 2.12.2 (git commit aa840eb1074fd7bb8557105d18eec632325114d8, built Thu Jul 21 06:16:00 EDT 2016)

[root@c910f03c05k21 ~]# tabdump networks
#netname,net,mask,mgtifname,gateway,dhcpserver,tftpserver,nameservers,ntpservers,logservers,dynamicrange,staticrange,staticrangeincrement,nodehostname,ddnsdomain,vlanid,domain,comments,disable
"10_0_0_0-255_0_0_0","10.0.0.0","255.0.0.0","enp0s1","10.3.5.10",,"10.3.5.21",,,,,,,,,,"pok.stglabs.ibm.com",,
"50_0_0_0-255_0_0_0","50.0.0.0","255.0.0.0","enp0s2","<xcatmaster>",,"50.5.1.122",,,,,,,,,,"clusters.com",,

[root@c910f03c05k21 makehosts c910f03c05k27
[root@c910f03c05k21 ~]# cat /etc/hosts
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
10.3.5.227 c910f03c05k27-pub c910f03c05k27-pub.pok.stglabs.ibm.com 
50.5.1.27 c910f03c05k27-pri c910f03c05k27-pri.pok.stglabs.ibm.com 
````

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1576

PR #1576, opened 2016-07-22
fix issue  [DEV] "makehosts" generates the /etc/hosts entries with wrong domain name for node with "otherinterfaces" attribute #1575

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1575


daniceexi:

PR #1577, opened 2016-07-22
Modify conflict xcat-genesis-scripts-x86-64 and xcat-genesis-scripts-…

zet809:
…ppc64 with xcat-genesis-scripts


daniceexi:
The change looks OK.

PR #1578, opened 2016-07-22
Add some error checking and dependency packages to xCAT-genesis-builder

whowutwut:
This is to help make it easier to create the genesis-base on newly installed machines when installing the xCAT-genesis-builder so most of the packages that are required to successfully build will be pulled in when doing `yum install xCAT-genesis-builder...`

### DIFF
--- a/xCAT-genesis-builder/xCAT-genesis-builder.spec
+++ b/xCAT-genesis-builder/xCAT-genesis-builder.spec
@@ -4,7 +4,7 @@ BuildArch: noarch
 Release: snap%(date +"%Y%m%d%H%M")
 Epoch: 1
 AutoReq: false
-Requires: ipmitool screen btrfs-progs lldpad rpm-build compat-libstdc++-33
+Requires: ipmitool screen btrfs-progs lldpad rpm-build compat-libstdc++-33 mstflint xfsprogs nc reiserfs-utils
 Prefix: /opt/xcat
 AutoProv: false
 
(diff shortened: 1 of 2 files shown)


xCAT-genesis-builder/xCAT-genesis-builder.spec
```diff
@@ -4,7 +4,7 @@ BuildArch: noarch
 Release: snap%(date +"%Y%m%d%H%M")
 Epoch: 1
 AutoReq: false
-Requires: ipmitool screen btrfs-progs lldpad rpm-build compat-libstdc++-33
+Requires: ipmitool screen btrfs-progs lldpad rpm-build compat-libstdc++-33 mstflint xfsprogs nc reiserfs-utils
```
zet809: Why the 4 more packages are needed?

whowutwut: Sorry for the delay, when running the `buildrpm` script in the genesis-builder to create `xCAT-genesis-base`, I encountered errors when `buildrpm` is run.  The output does not always fail, so in order to help ensure that it builds successfully, I added those packages.   I can try to find the errors if you are interested.

whowutwut:
@zet809  sorry my response was so delay, if we want to merge this  in 2.12.2 it's low risk since this is not part of xCAT code directly but is required to build xCAT-genesis-base, so leaving in 2.12.2.. if we want to hold off, can move to 2.12.3 , I'm ok with that too

zet809:
Based on our build xcat-genesis-base process wikipage, it seems the dependency packages are different for x86_64 and ppc64. I suggest we move this pull request to the next release to figure out all the additional pkgs, steps and then enhance the build process.

whowutwut:
@zet809  This doesn't affect the xCAT code path, but just helps with building the genesis rpm, can we merge and then open more issues the next time we look at this?

zet809:
@whowutwut I agree with you, and merge this request.

PR #1586, opened 2016-07-25
Remove the code sections that upload debian builds to xcat.org

whowutwut:
This code block isn't being used anymore and the upload is handled in our daily build scripts. 
Removing this code to keep things simple.


daniceexi:
@immarvin will be the backup for xCAT build procedure in China team when @tingtli  is not ready. 
@immarvin could you review this change?

immarvin:
agree to merge

PR #1588, opened 2016-07-26
Fix the overwrite error when dealing with the conflict

chenglch:


immarvin:
seems ok for me, agree to merge

PR #1592, opened 2016-07-26
Add trust_host option in rest api for makeconservercf

chenglch:


daniceexi:
@zet809  could you review this change?

xCAT-server/xCAT-wsapi/xcatws.cgi
```diff
@@ -225,6 +225,8 @@ my %URIdef = (
                 usage => "|Json Formatted DataBody: {action:on/off}.|$usagemsg{non_getreturn}|",
                 example => "|Enable the console capability for node1|PUT|/nodes/node1/console {\"action\":\"on\"}||",
                 cmd      => "makeconservercf",
+                example => "|Enable the console capability for node1|PUT|/nodes/node1/console {\"action\":\"on\", \"trust_host\": \"host\"}||",
+                cmd => "makeconservercf",
```
zet809: Duplicated with line 227

xCAT-server/xCAT-wsapi/xcatws.cgi
```diff
@@ -2197,7 +2197,12 @@ sub actionhdl {
         if ($paramhash->{'action'}) {
             my $option = $paramhash->{'action'};
             my %op_hash = ('on' => '', 'off' => '-d');
-            push @args, $op_hash{ $paramhash->{'action'} };
+            my $trust_host = $paramhash->{'trust_host'};
+            push @args, $op_hash{$paramhash->{'action'}};
```
zet809: Why not use the '$option' variable? It have get value from $paramhash->{'action'}.

zet809:
Looks good to me, 1 more thing need to mention is:
Since console is treated as resource of a node, when configuring trusthost, it need to work with a node at present. In fact, the trusthost is for a conserver, we need to add another console resource to deal with trusthost without attach it to a node.

daniceexi:
👍

Issue #1557, opened 2016-07-19
[Customer] Using chvm -a to add additional disks does not preserve storageformat configuraiton


whowutwut:
Customer brings up this issue using the xCAT 2.9.X release.   I'm opening to milestone in the 2.9 release to track it.  I think if we fix in master, we should roll this fix back to 2.9 or provide a patch to the customer. 

A vm node definition in xCAT is changed to use `vmstorageformat=raw` and re-created. The VM storage file is created correctly with type=raw.  However, when adding additional disks with the `chvm -a <size>` command, the additional disks created appear with the default `qcow2` format. 

Example:
```
[root@fs2vm100 ~]# lsdef victorvm1 -i vmstorageformat
Object name: victorvm1
    vmstorageformat=raw

[root@fs2 ~]# ls -ltr /var/lib/libvirt/images/victor*
-rw-------. 1 root root 10737418240 Jul 16 19:57 /var/lib/libvirt/images/victorvm1.sda.raw
```

Then creating additional disks
```
chvm victorvm1 -a 20G
chvm victorvm1 -a 10G
```

Results in the following files: 
```
[root@fs2 ~]# ls -ltr /var/lib/libvirt/images/victor*
-rw-------. 1 root root 10737418240 Jul 16 19:57 /var/lib/libvirt/images/victorvm1.sda.raw
-rw-------. 1 root root      197120 Jul 16 19:59 /var/lib/libvirt/images/victorvm1.sdb.qcow2 <---qcow2
-rw-------. 1 root root      197120 Jul 19 09:08 /var/lib/libvirt/images/victorvm1.sdc.qcow2 <--- qcow2
```

daniceexi:
I think this is a bug of xCAT that we don't handle the storage format in the chvm command. Refer to the following code in the kvm.pm. We really should pass in the storage format when call the get_filepath_by_url in the chvm() subroutine.

~~~~
1925             #ok, now I need a volume created to attach
1926             push @diskstoadd, get_filepath_by_url(url => $location, dev => $dev, create => $_);
~~~~

gurevichmark:
@daniceexi Thank you for pointing out the lines to in question. Before I implement the fix, there are 3 questions:

1. When creating vm with `mkvm` command, we first set the storage format to the value of `vmvirtflags imageformat=<value>` if specified. Then overwrite it with the value of `vmstorageformat=<value>` if specified. I assume `chvm` should do the same to figure out the value for the storage format ?

[...]

PR #1593, opened 2016-07-26
Use correct vmstorageformat for chvm

gurevichmark:
This pull request fixes the problem reported by issue #1557.
The following changes were made:

1. Inline code to check the setting of `vmstorageformat` attribute is pulled into a function.
2. Value of `vmstorageformat` attribute is passed into the `get_filepath_by_url()` from `chvm()` function.
3. Error is returned if `chvm -a` is called on the vm that does not have `vmstorage` attribute set.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -1226,6 +1226,25 @@ sub getpowstate {
     }
 }
 
+# Return storageformat definition
+sub getstorageformat {
+    my $cfginfo = shift;
+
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -1226,6 +1226,25 @@ sub getpowstate {
     }
 }
 
+# Return storageformati definition
```
whowutwut: Small typo here, would suggest fixing and `commit --amend`

whowutwut:
I reviewed and  will merge

PR #1594, opened 2016-07-26
Use correct vmstorageformat for chvm

gurevichmark:
This pull request recreates for 2.9 release the changes under #1593 for 2.12

Fix issue #1557

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -1081,6 +1081,25 @@ sub getpowstate {
     }
 }
 
+# Return storageformat definition
+sub getstorageformat {
+    my $cfginfo = shift;
+
+    my @flags = split /,/, $cfginfo->{virtflags};
+    my $format;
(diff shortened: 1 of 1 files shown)


whowutwut:
@daniceexi  The fix seems OK with me.   Is there any FVT verification that we can run on the 2.9 release before releasing to the customer?

daniceexi:
@whowutwut Currently we don't have test cases to cover the chvm against KVM. So we can do is to run it manually.

I added a story to implement the test cases for kvm:
https://jazz07.rchland.ibm.com:21443/jazz/web/projects/pc-management#action=com.ibm.team.workitem.viewWorkItem&id=115983

Regarding whether we need to involve FVT to verify this change, my idea is that the DUT by @gurevichmark should be enough since the change is small and looks safe. If involving FVT, it needs at least one day to finish.

daniceexi:
This change has been reviewed by @whowutwut and @daniceexi.

gurevichmark:
Tested on xCAT 2.9 MN with RH 7.1

PR #1598, opened 2016-07-27
Support --resize option for chvm

gurevichmark:
Implement --resize option for kvm VMs using chvm command

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -1807,7 +1807,7 @@ sub rmvm {
 sub chvm {
     shift;
     my @addsizes;
-    my %resize;
+    my $resize;
     my $cpucount;
     my @purge;
     my @derefdisks;
@@ -1831,7 +1831,7 @@ sub chvm {
             "eject"               => \$eject,
             "cpus|cpu=s"          => \$cpucount,
             "p=s"                 => \@purge,
-            "resize=s%"           => \%resize,
+            "resize=s"            => \$resize,
             "cpupin=s"            => \$pcpuset,
             "membind=s"           => \$numanodeset,
             "devpassthru=s"       => \$passthrudevices,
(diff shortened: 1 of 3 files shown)


xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -2096,6 +2096,53 @@ sub chvm {
...
+                    # Is this a disk we were looking for to resize ?
+                    if ($file =~ /\.$disk_to_resize\./) {
+                        my $vol = $hypconn->get_storage_volume_by_path($file);
+                        if ($vol) {
+                            # Always pass RESIZE_SHRINK flag to resize(). It is required when shrinking
+                            #the volume size and is ignored when growing volume size
+                            eval {
+                                $vol->resize($value, &Sys::Virt::StorageVol::RESIZE_SHRINK);
+                            };
+                            if (($@) && ($@ =~ /$shrinking_not_supported/)) {
```
immarvin: if the resize() failed with the message other than "qcow2 doesn't support shrinking images yet" , should user be aware of that the operation is failed?

gurevichmark: @immarvin You are right, of course. Fixed this under a new commit.

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -2096,6 +2096,58 @@ sub chvm {
...
+                $value = getUnits($value, "G", 1);
+                # Now search kvm_nodedata table to find the volume for this disk
+                my $myxml    = $parser->parse_string($vmxml);
+                my @alldisks = $myxml->findnodes("/domain/devices/disk/source");
+                foreach my $disk (@alldisks) {
+                    my $disktype = $disk->parentNode()->getAttribute("device");
+                    if ($disktype eq "cdrom") { next; }
+                    my $file = $disk->getAttribute("file");
+                    # Is this a disk we were looking for to resize ?
+                    if ($file =~ /\.$disk_to_resize\./) {
```
immarvin: it is not reliable enough to pick out the disk via the source file name, there is a subroutine ``get_disks_by_userspecs()`` in kvm.pm, maybe we should leverage that.  please ignore the "remove" words inside the comments in the subroutine.

gurevichmark: @immarvin Thank you for pointing out the `get_disks_by_userspecs()`, I would like to use it. But there is no explanation for what this function expects as parameters and what it returns. Can you help me understand how I should call this function and what it will return ?

immarvin: hi @gurevichmark , I added the description of the get_disks_by_userspecs in https://github.com/xcat2/xcat-core/pull/1607

you can also refer to the code slice below in kvm.pm:
````
2016         my @disklist = get_disks_by_userspecs(\@purge, $vmxml, 'returnmodde     dxml');
2017         my $moddedxml = shift @disklist;
2018         foreach (@disklist) {
2019             my $devxml = $_->[0];
2020             my $file   = $_->[1];
2021             $file =~ m!/([^/]*)/($node\..*)\z!;
2022 
2023             #first, detach the device.
2024             eval {
2025                 if ($currstate eq 'on') {
2026                     $dom->detach_device($devxml);
2027                     $vmxml = $dom->get_xml_description();
2028                 } else {
2029                     $vmxml = $moddedxml;
2030                 }
2031                 $updatetable->{kvm_nodedata}->{$node}->{xml} = $vmxml;
2032             };
````

gurevichmark: @immarvin Thank you for the explanation of what `get_disks_by_userspecs()` does. Currently it seems to return disk xml structure and value of  `<source file=xxx>`. For my processing, I also need the `<target dev=xxx>` value. 
I am thinking of adding that to what `get_disks_by_userspecs()`  returns. Then, after calling this function I can extract `<target dev=xxx>` value. Since there is just one place currently where `get_disks_by_userspecs()` is used, I think this change will not affect that call. Do you agree ?

immarvin: hi @gurevichmark ,  you need the target/dev since you want to get the list of the specified disks that exists in the domain, is it? I think it's ok to enhance the get_disks_by_userspecs subroutine according to your need. maybe the return structure can be something like:
 
````
(
  domainxml => $domainxml,
  devinfo  =>(
      $targetdevname => (
                                         "devxml" => $devxml,
                                         "srcfile"  => $srcfilepath, 
                                  ),
     ...
  ),
)
````
btw, the calling of get_disks_by_userspecs() in "chvm -p" and the description of get_disks_by_userspecs() should also be updated after your modification. 

thanks

gurevichmark: @immarvin I checked in another commit to handle finding a match for diskname more reliably. I decided to not change the `get_disks_by_userspecs()` function. Instead I just find the `dev` attribute which contains the disk name in the XML structure that is already being searched in the `resize disk` processing.

immarvin:
hi @gurevichmark , thanks for your modification. The code looks good and will work well. 

Currently, in the iteration of each specified disk name:
1.  parse the domain xml and get all the disks of the domain
2.  iterate the disk list of the domain and look for the specified disk
3. ...
 I think the code logic can be optimized to improve the performance.  The  code logic on "chvm -p" is similar to the "chvm --resize" in functionality, it might be an example to optimize the code logic.

gurevichmark:
@immarvin Yes, the processing for `chvm -p` is more efficient. But I am not sure if I can do the same for `chvm --resize`. The disk list passed for `chvm -p` is just `sdb,sdc,sde,...` but for `chvm --resize` it is `sdb=10G,sdc=500M, sde=5G`, so I think I need to go through each disk at a time in order to keep the resize size together with the disk name.

gurevichmark:
@immarvin Did you get a chance to review the latest changes to this pull request ?

immarvin:
merge thos pr first, can create another pr to optimize the code logic on perfomance considerations

Issue #1413, opened 2016-06-28
[FVT]:mysqlsetup -i failed to setup database on sles12


tingtli:
xcat 2.12.1 on sles12.
```
c910f03c04k04:/ # lsxcatd -v
Version 2.12.1 (git commit 93f20131dc8ebf4d8c4de47000bd631b7c7a9be7, built Thu Jun 23 09:30:44 EDT 2016)

After running mysqlsetup -i and input the root and xcatadmin password. The output shows the database is setup. But it seems it is not correctly setup.

c910f03c04k04:/ # /usr/bin/mysql -u root -p
Enter password:
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 144
Server version: 10.0.11-MariaDB openSUSE package

Copyright (c) 2000, 2014, Oracle, SkySQL Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [(none)]> exit
Bye
c910f03c04k04:/ # /usr/bin/mysql -u xcatadmin -p
Enter password:
ERROR 1045 (28000): Access denied for user 'xcatadmin'@'localhost' (using password: YES)
```

tingtli:
@cxhong Please take a look at this issue. The environment is 10.3.4.4. The password is cluster. Thanks!

cxhong:
the wildcard "%" seems didn't work here
`````
MariaDB [(none)]> SELECT host,user FROM mysql.user;
+---------------+-----------+
| host          | user      |
+---------------+-----------+
| %             | xcatadmin |
| 10.3.4.4      | xcatadmin |
| 127.0.0.1     | root      |
| ::1           | root      |
| c910f03c04k04 |           |
| c910f03c04k04 | root      |
| localhost     |           |
| localhost     | root      |
+---------------+-----------+
``````
Once I added 
````
 GRANT ALL on xcatdb.* To xcatadmin@localhost IDENTIFIED by 'cluster';
````

the xcatadmin user can access to database again.

I will look into more.

cxhong:
The mysql docs have this to say: (from  http://dev.mysql.com/doc/refman/5.1/en/adding-users.html):
````


    Two of the accounts have a user name of monty and a password of some_pass. Both accounts are superuser accounts with full privileges to do anything. The 'monty'@'localhost' account can be used only when connecting from the local host. The 'monty'@'%' account uses the '%' wildcard for the host part, so it can be used to connect from any host.

[...]

PR #1600, opened 2016-07-28
fix  [FVT]:mysqlsetup -i failed to setup database on sles12 #1413 ;refine the logic to determine hostname of the master

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1413


cxhong:
thanks.

PR #1601, opened 2016-07-28
Deal with the scenario that more than 1 nic connecting to the provisi…

zet809:
…oning network
The idea is:
If there are more than 1 nic connect to the provisioning network, creating hostname like $node-$nicname as alias of the node for those nics. Then when getting IP configure with DHCP, those nics can get the same ip, hostname and statement for the specified node.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/nodediscover.pm
+++ b/xCAT-server/lib/xcat/plugins/nodediscover.pm
@@ -285,15 +285,15 @@ sub process_request {
                     my $ipn = unpack("N", inet_aton($ip));
                     my $mask = 2**$netbits - 1 << (32 - $netbits);
                     my $netn = inet_ntoa(pack("N", $ipn & $mask));
-                    my $hosttag;
-                    if ($usednames_for_net{$netn}) {
-                        $hosttag = $usednames_for_net{$netn};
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/nodediscover.pm
```diff
@@ -283,10 +285,20 @@ sub process_request {
                     my $ipn = unpack("N", inet_aton($ip));
                     my $mask = 2**$netbits - 1 << (32 - $netbits);
                     my $netn = inet_ntoa(pack("N", $ipn & $mask));
-                    my $hosttag = gethosttag($node, $netn, @ifinfo[1], \%usednames);
+                    my $hosttag;
+                    if ($usednames_for_net{$netn}) {
+                        $hosttag = $usednames_for_net{$netn};
```
daniceexi: Should you put this part of code in the L300 that when more than one mac maps to one hostname, then you should do something for hostname?

daniceexi:
@zet809 Could you investigate how the networks.nodehostname attribute works in **gethosttag** subroutine?

daniceexi:
@immarvin You could also review this change.

PR #1605, opened 2016-07-29
refine cases to remove packages on ubuntu completely

tingtli:


neo954:
The patches looked good.

PR #1606, opened 2016-07-29
To have runxcmd to keep $::callback in case some module use it

zet809:


zet809:
@gurevichmark Hi, Mark, this pull request is used to fix your problem.

daniceexi:

Issue #1609, opened 2016-08-01
[DEV]xdsh cannot propagate Ctrl+c signal  to the remote targets


xuweibj:
Comand: ``xdsh c910f03c17k17 -s "tail -f /tmp/probe.test"``
Operation: Ctrl + c

The process on c910f03c17k17 did not be killed.

chenglch:
As the fix https://github.com/xcat2/xcat-core/pull/1610 has been revert at https://github.com/xcat2/xcat-core/pull/1632, this problem remain.

PR #1610, opened 2016-08-01
Propagate signal via ssh even though pipe broken can not be detected

chenglch:
Add `-tt` option to force tty allocation when running ssh command

close-issue: #1609

### DIFF
--- a/perl-xCAT/xCAT/SSH.pm
+++ b/perl-xCAT/xCAT/SSH.pm
@@ -75,6 +75,7 @@ sub remote_shell_command {
     if ($ssh_version eq 'OpenSSH') {
         push @command, '-o';
         push @command, 'BatchMode=yes';
+        push @command, '-tt';
 
         ($$config{'options'} !~ /-X/) && push @command, '-x';
     }


zet809:
Since xdsh can be used for RH, ubuntu, sles, switch, HMC, we need to verify whether this option works on those platforms.

daniceexi:
@chenglch Could you give some explanation of why 'ssh -tt' helps for this issue?

chenglch:
@daniceexi  I can't explain too much for the -tt option. Ansible also use this option by default if ssh connection is used. One thing I can confirm is that  the signal code can not be transferred to the sshd side without -t option.  

Some information from the man page of ssh
-t      Force pseudo-tty allocation.  This can be used to execute arbitrary screen-based programs on a remote machine, which can be very useful, e.g. when implementing menu services.
             **Multiple -t options force tty allocation, even if ssh has no local tty**.

Like @zet809 mentioned, I think this should be a test driven patch, I'm not sure if some platform do not support this option, or some scripts can not be executed with prompt(enable by -t) enabled.

daniceexi:
@chenglch Get it. Thanks.

daniceexi:
Then I think we can merge it and label this issue as test-required

PR #1608, opened 2016-07-29
Master client merges for two files from zVM

chuckbrazie:
These are the only two updates I found in zVM xCAT xCAT-client which need to be merged. Let me know if the merge I did is correct.

PR #1617, opened 2016-08-02
Fix indentation in xdsh

chuckbrazie:


immarvin:
hi @chuckbrazie , it seems that this PR has some conflict to merge, I am afraid that this is due to the merge of https://github.com/xcat2/xcat-core/pull/1608. You need to rebase your branch with the updated master and then update this PR

xCAT-client/bin/xdsh
```diff
@@ -573,9 +575,11 @@ sub parse_args_xdcp
             'F|File=s'              => \$options{'File'},
             'h|help'                => \$options{'help'},
             'i|rootimg=s'           => \$options{'rootimg'},
+            'ip=s'                  => \$options{'ip'},
```
immarvin: if this option is only available for ZVM, please add a comment to indicate this.

immarvin:
I reverted the PR https://github.com/xcat2/xcat-core/pull/1608, so there is no need to rebase the branch of this PR.

xCAT-client/bin/xdsh
immarvin: merge the PR first,    you can create another PR on the comments.

Issue #1621, opened 2016-08-03
[FVT] updatenode CN without return value for more than hours


caomengmeng:
Description: Failed to get updatenode CN return value after more than hours
Test ENV:
```
[root@c910f02c13 ~]# lsxcatd -a
Version 2.12.2 (git commit 5cd203cdf563ac402d5718b5fdbfbec788870eaf, built Tue Aug  2 06:15:54 EDT 2016)
This is a Management Node
dbengine=SQLite
[root@c910f0213 ~]# cat /etc/*release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.2 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.2"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.2 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.2:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.2
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.2"
Red Hat Enterprise Linux Server release 7.2 (Maipo)
Red Hat Enterprise Linux Server release 7.2 (Maipo)
```
Details:
```
[root@c910f02c13 ~]# updatenode c910f02c02p27
There were no syncfiles defined to process. File synchronization has completed.
Performing software maintenance operations. This could take a while, if there are packages to install.

c910f02c02p27: xcatdsklspost: updating VPD database

c910f02c02p27: xcatdsklspost: downloaded postscripts successfully

c910f02c02p27: Tue Aug  2 23:58:48 EDT 2016 Running postscript: ospkgs

c910f02c02p27: Postscript: ospkgs exited with code 0

c910f02c02p27: Tue Aug  2 23:58:50 EDT 2016 Running postscript: otherpkgs

c910f02c02p27: ./otherpkgs: no extra rpms to install

c910f02c02p27: Postscript: otherpkgs exited with code 0

c910f02c02p27: Tue Aug  2 23:58:50 EDT 2016 Running postscript: syscloneimgupdate

[...]

PR #1623, opened 2016-08-03
fix issue  [FVT] updatenode CN without return value for more than hours #1621

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1621
The following code slice in remoteshell 
````
    225 
    226 # if node supports ecdsa host key then download the replacement from the         MN/SN
    227 if ssh-keygen -t ecdsa -f /tmp/ecdsa_key -P "" 1>/dev/null ; then
  ````
hang at
````

[root@c910f02c02p27 xcatpost]# ssh-keygen -t ecdsa -f /tmp/ecdsa_key -P ''
Generating public/private ecdsa key pair.
/tmp/ecdsa_key already exists.
Overwrite (y/n)? ^C
````
the fix is to remove the  /tmp/ecdsa_key to avoid the command hang on waiting for input

### DIFF
--- a/xCAT/postscripts/remoteshell
+++ b/xCAT/postscripts/remoteshell
@@ -225,6 +225,9 @@ fi
 rm /tmp/ssh_rsa_hostkey
 
 # if node supports ecdsa host key then download the replacement from the MN/SN
+# remove the /tmp/ecdsa_key first, otherwise the "ssh-keygen" below might hang
+# at waiting for user confirmation to overwritten the existing file
+rm -rf /tmp/ecdsa_key >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: xCAT/postscripts/remoteshell [hunk 1]
```diff
@@ -225,6 +225,7 @@ fi
 rm /tmp/ssh_rsa_hostkey
 
 # if node supports ecdsa host key then download the replacement from the MN/SN
+rm -rf /tmp/ecdsa_key >/dev/null 2>&1
```
daniceexi: I could guess what's the root cause for this issue, but could you add a description message in the commit and pull request for this change?

daniceexi:

PR #1627, opened 2016-08-03
Cleanup boot configuration files when processing rmdef

### LINKED ISSUE #1524: Clean up xCAT created files when removing node definition objects
If nodes are removed from xCAT database and machines are renamed, xCAT leaves around files in `/tftpboot` for example.  It would be good if xCAT's `rmdef` command handles the clean up so the user doesn't need to worry about rogue files on the management node.

gurevichmark:
This pull request fixes #1524 
During `rmdef` processing of the node, `nodeset offline` is called to remove boot configuration files from /tftpboot directory.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/DBobjectdefs.pm
+++ b/xCAT-server/lib/xcat/plugins/DBobjectdefs.pm
@@ -55,6 +55,8 @@ $Getopt::Long::ignorecase = 0;
 #@::objfilelist;     # list of object names from the "-f" option
 #@::allobjnames;     # combined list
 
+my $doreq;
+
 #@::noderange;       # list of nodes derived from command line
 
 #------------------------------------------------------------------------------
@@ -137,6 +139,7 @@ sub process_request
 
     $::request  = shift;
     $::callback = shift;
+    $doreq = shift;
 
     my $ret;
     my $msg;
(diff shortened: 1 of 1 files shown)


daniceexi:
:+1: 

@gurevichmark did you test the change against noderange with node and groups mix?

gurevichmark:
Yes. I tested the following cases:
```
rmdef <vmname1>
rmdef <vmname1>,<vmname2>,<vmname3>
rmdef <vmname>[1-3]
rmdef <groupname>
rmdef <groupname>,<vmname4>,<vmname5>
```

daniceexi:
Then I am fine to merge. @whowutwut

xCAT-server/lib/xcat/plugins/DBobjectdefs.pm
```diff
@@ -4239,6 +4242,26 @@ sub defrm
         }
     }
 
+    # Call nodeset offline on each node to cleanup its boot configuration files from /tftpboot directory
+    my @allnodes;
```
immarvin: if "@allnodes" is only used in the code block:
````
if ($doreq) {
...
}
````

maybe it should be declared inside the block.

gurevichmark: I agree, that will be better. I updated the merge.

PR #1632, opened 2016-08-04
Revert "Propagate signal via ssh even though pipe broken can not be detected"

chenglch:
Reverts xcat2/xcat-core#1610

Some problem detected, will solve it at localhost at first.


viniciusferrao:
@chenglch it has been a decade, but do you remember which problems arose?

Issue #1611, opened 2016-08-01
[DEV]"genimage ubuntu16.04.1-ppc64el-netboot-compute" failed 


immarvin:
````
root@c910f03c17k27:~# lsxcatd -v
Version 2.12.2 (git commit 1493386f1705be3c8fc6c4819fbfd0be2497e01f, built Sat Jul 30 06:17:01 EDT 2016)
root@c910f03c17k27:~# lsdef -t osimage -o ubuntu16.04.1-ppc64el-netboot-compute Object name: ubuntu16.04.1-ppc64el-netboot-compute
    exlist=/opt/xcat/share/xcat/netboot/ubuntu/compute.exlist
    imagetype=linux
    osarch=ppc64el
    osname=Linux
    osvers=ubuntu16.04.1
    otherpkgdir=/install/post/otherpkgs/ubuntu16.04.1/ppc64el
    permission=755
    pkgdir=/install/ubuntu16.04.1/ppc64el
    pkglist=/opt/xcat/share/xcat/netboot/ubuntu/compute.ubuntu16.04.ppc64el.pkglist
    postinstall=/opt/xcat/share/xcat/netboot/ubuntu/compute.postinstall
    profile=compute
    provmethod=netboot
    rootimgdir=/install/netboot/ubuntu16.04.1/ppc64el/compute
    synclists=/install/custom/netboot/ubuntu/compute.synclist
root@c910f03c17k27:~# genimage ubuntu16.04.1-ppc64el-netboot-compute
...
 DEBIAN_FRONTEND=noninteractive chroot /install/netboot/ubuntu16.04.1/ppc64el/compute/rootimg apt-get -y  install bash nfs-common openssl isc-dhcp-client libc-bin openssh-server openssh-client wget vim ntp ntpdate rsync busybox-static gawk dnsutils xz-utils
Reading package lists...
Building dependency tree...
Reading state information...
bash is already the newest version (4.3-14ubuntu1.1).
isc-dhcp-client is already the newest version (4.3.3-5ubuntu12).
libc-bin is already the newest version (2.23-0ubuntu3).
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 vim : Depends: vim-common (= 2:7.4.1689-3ubuntu1.1) but 2:7.4.1689-3ubuntu1 is to be installed
[...]

Issue #472, opened 2015-11-26
[DEV] trimming diskless rootimg.gz with exlist should be added to documentation; sles12 exlist should be updated


immarvin:
The exclude list attribute ``linuximage.exlist`` enables user to trim the image after the rpms are installed into the rootimage, so that the rootimg.gz will  be as small as possible. xCAT ships a default exlist, the users can add or remove entries in the exlist file according to there need.  There should be a section in ``Docs » Admin Guide » Manage Clusters » IBM Power LE / OpenPOWER » Diskless Installation » Customize osimage (Optional)`` on this

the entries in the sles12 exlist file ``xCAT-server/share/xcat/netboot/sles/compute.sles12.ppc64le.exlist`` seems to be out-of-date:
[root@c910f03c01p03 sles]# cat compute.sles12.ppc64le.exlist
./boot*
./etc/bootsplash/themes/SLES/images*
./opt/sci/include*
./usr/include*
./usr/lib/locale*
==================================================>
./usr/lib/perl5/5.10.0/ppc64-linux-thread-multi/auto/Encode/CN*
./usr/lib/perl5/5.10.0/ppc64-linux-thread-multi/auto/Encode/JP*
./usr/lib/perl5/5.10.0/ppc64-linux-thread-multi/auto/Encode/TW*
./usr/lib/perl5/5.10.0/ppc64-linux-thread-multi/auto/Encode/KR*
./usr/lib/perl5/5.10.0/ppc64-linux-thread-multi/Encode/CN*
./usr/lib/perl5/5.10.0/ppc64-linux-thread-multi/Encode/JP*
./usr/lib/perl5/5.10.0/ppc64-linux-thread-multi/Encode/KR*
./usr/lib/perl5/5.10.0/ppc64-linux-thread-multi/Encode/TW*
<====================================================
./usr/lib64/gcc/powerpc64-suse-linux/4.3/include*
./usr/local/include*
./usr/local/man*
./usr/powerpc64-suse-linux/include*
./usr/share/X11/locale/*
+./usr/share/X11/locale/en_US.UTF-8*
+./usr/share/X11/locale/C*
./usr/share/cracklib*
./usr/share/doc*
./usr/share/doc/packages/cyrus-sasl/doc*
./usr/share/gnome*
./usr/share/i18n*
./usr/share/info*
./usr/share/locale/*
+./usr/share/locale/en_US*
+./usr/share/locale/C*
./usr/share/man*
./usr/share/omf*
[...]

PR #1636, opened 2016-08-04
add ubuntu16.04.1 netboot pkglist;modify the precedence of SvrUtils:get_file_name

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1611

1. add ubuntu16.04.1 diskless pkglist  file "compute.ubuntu16.04.1.pkglist", which is based on the one for ubuntu16.04,  "vim" is not included since its dependency in the ubuntu 16.04.1 iso is incomplete. 

2. modify the the logic of get_file_name subroutine in SvrUtils.pm, so that the pkglist with the name with full os release number takes precedence over the base os release number. after modification, the "compute.ubuntu16.04.1.pkglist" will be the better candidate than "compute.ubuntu16.04.ppc64le.pkglist" for ubuntu 16.04.1 ppc64le osimage

### DIFF
--- a/xCAT-server/lib/perl/xCAT/SvrUtils.pm
+++ b/xCAT-server/lib/perl/xCAT/SvrUtils.pm
@@ -460,20 +460,20 @@ sub get_file_name {
     if (-r "$searchpath/$profile.$os.$arch.$extension") {
         return "$searchpath/$profile.$os.$arch.$extension";
     }
-    elsif (-r "$searchpath/$profile.$osbase.$arch.$extension") {
-        return "$searchpath/$profile.$osbase.$arch.$extension";
-    }
-    elsif (-r "$searchpath/$profile.$genos.$arch.$extension") {
-        return "$searchpath/$profile.$genos.$arch.$extension";
-    }
     elsif (-r "$searchpath/$profile.$os.$extension") {
         return "$searchpath/$profile.$os.$extension";
     }
+    elsif (($genos) && (-r "$searchpath/$profile.$genos.$arch.$extension")) {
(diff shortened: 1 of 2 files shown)


xCAT-server/lib/perl/xCAT/SvrUtils.pm
```diff
@@ -460,15 +460,15 @@ sub get_file_name {
...
     elsif (-r "$searchpath/$profile.$osbase.$arch.$extension") {
         return "$searchpath/$profile.$osbase.$arch.$extension";
     }
     elsif (-r "$searchpath/$profile.$genos.$arch.$extension") {
         return "$searchpath/$profile.$genos.$arch.$extension";
     }
-    elsif (-r "$searchpath/$profile.$os.$extension") {
-        return "$searchpath/$profile.$os.$extension";
-    }
     elsif (-r "$searchpath/$profile.$osbase.$extension") {
```
daniceexi: If we adopt your change, do we should move this part to line 469 to be consistent?

immarvin:
hi @daniceexi , I do not think the line #472 should be moved to line#469.

for ubuntu 16.04.1 ppc64el diskless osimage, the precedence of the pkglist files under netboot/ubuntu should be:
````
compute.ubuntu16.04.1.ppc64el.pkglist
compute.ubuntu16.04.1.pkglist
compute.ubuntu16.04.ppc64el.pkglist
compute.ubuntu16.04.pkglist
compute.ppc64el.pkglist
compute.pkglist
````

daniceexi:

Issue #1638, opened 2016-08-05
[DEV] Genesis not wait long enough time to have NICs get IP address when STP is not closed


zet809:
For currect genesis base for ppc64, the NICs will be in DOWN state when genesis kernel is running. We have set it up within xcat genesis doxcat script.
The issue is, if not close STP on switch or set port fast, it will be about 50 seconds before the first package can be send out. 
Currently, within doxcat, we only wait for 20 seconds to check whether the NICs can get IP address through DHCP, it is not long enough. 
We need to modify it to be 60s or longer. 

The log below shows that, from starting dhclient to require IP address to the time that the DHCP request is send out successfully, it will cost about 40 seconds.
2016-08-05T06:24:43.784494+00:00 (none) xcat.genesis.doxcat: Setting IP via DHCP...     **==> Start dhclient to require dynamic IP address for all NICs**
2016-08-05T06:24:45.793430+00:00 (none) xcat.genesis.doxcat: Acquiring network addresses..
2016-08-05T06:24:45.812918+00:00 (none) dhclient[1290]: Can't bind to dhcp address: Cannot assign requested address
2016-08-05T06:24:45.812924+00:00 (none) dhclient[1290]: Please make sure there is no other dhcp server
...
2016-08-05T06:24:59.308137+00:00 (none) dhclient[1289]: DHCPDISCOVER on enP5p7s0f0 to 255.255.255.255 port 67 interval 10 (xid=0x52655a84)
2016-08-05T06:25:05.819030+00:00 (none) xcat.genesis.doxcat: still can not get bootnic, go into /bin/bash
2016-08-05T06:25:09.676646+00:00 (none) dhclient[1289]: DHCPDISCOVER on enP5p7s0f0 to 255.255.255.255 port 67 interval 12 (xid=0x52655a84)
2016-08-05T06:25:21.089198+00:00 (none) dhclient[1289]: DHCPDISCOVER on enP5p7s0f0 to 255.255.255.255 port 67 interval 13 (xid=0x52655a84)
2016-08-05T06:25:22.089597+00:00 (none) dhclient[1289]: DHCPREQUEST on enP5p7s0f0 to 255.255.255.255 port 67 (xid=0x52655a84)                                                                              **==> The DHCP request send out successfully at this time.**
2016-08-05T06:25:22.089616+00:00 (none) dhclient[1289]: DHCPOFFER from 192.168.3.29
[...]

PR #1640, opened 2016-08-05
Fix issue 1638: Genesis not wait long enough time to have NICs get IP…

zet809:
… address
Fix issue #1638

### DIFF
--- a/xCAT-genesis-scripts/bin/doxcat
+++ b/xCAT-genesis-scripts/bin/doxcat
@@ -217,15 +217,16 @@ else
 	                fi
                 fi
                 done
-                sleep 2
+                sleep 3
                 tries=$(($tries+1))
-                if [ $tries -ge 10 ]; then
+                # Wait for 60 seconds to make sure the STP is done for at least one port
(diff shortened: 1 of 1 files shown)


daniceexi:
@whowutwut could you confirm that the top of rack switch has NTP enabled?

gurevichmark:
Changes look good to me.

whowutwut:
Sorry, was merged before I got a chance to comment.  

One thought that I had here is whether we should keep the existing code and simply add another fail over test...  The behavior right now is that we simply do not get an IP address because we did not wait long enough.  

However, I think there are problems even after Genesis regarding port-fast, so I think maybe we should have left the existing code. If it fails to get an IP address, print a message to inform the user of possible options to resolve it.  I think if genesis gets the IP address in HW discovery,  it doesn't necessarily mean that the OS will be able to get the IP address unless some configuration is fixed on the switch side.

daniceexi:
@whowutwut Do you mean we can recreate this issue even when the **STP** on the edge switch is disabled. If so I think we need figure out why this happens first.

PR #1646, opened 2016-08-08
Modify usage, some probe commands do not support hierarchy

xuweibj:
Modified discovery, image and switch-macmap.
Currently, these commands do not support hierarchy.


hu-weihua:
I have reviewed the change and agree to merge. Thanks @xuweibj .  good job.

PR #1654, opened 2016-08-09
[go-xcat] Remove all the complicated examples

neo954:


daniceexi:
@neo954 I think we need to keep the flags ``--xcat-core`` and ``--xcat-dep`` so that user can specify any 'xcat-core' and 'xcat-dep' tar files by themselves. This is useful especially the user cannot access internet that he has to download the tar.bz2 files to local and then use go-xcat to install.

~~~~
${script} --xcat-core=http://xcat.org/path/to/xcat-core.tar.bz2 --xcat-dep=http://xcat.org/path/to/xcat-dep.tar.bz2 


${script} --xcat-core=/path/to/xcat-core.tar.bz2 --xcat-dep=/path/to/xcat-dep.tar.bz2  install
~~~~

daniceexi:

whowutwut:
:+1:  I like this, many of the examples were very complicated :)

PR #1656, opened 2016-08-09
update bundle files according to go-xcat change

tingtli:


whowutwut:
@tingtli  just some thought... Is it possible to have a bundle import the cases from another bundle? 

If i understand this change correctly, it looks like the go-xcat testcase name changed. 
If we have the ability to include a file that contains the testcase list, the bundle files could import the go-xcat testcase list. Then we would only have to change 1 file..

The bundles would be more easily kept in sync.

tingtli:
@whowutwut The testcase name is not changed. i have a total bucket which has  a lot of test cases. Since @neo954 modified the man page, I choose different test cases for the daily run. 
The idea to include a bundle file in another bundle file is good. This requires code change for xcattest to process the bundle file. We can put it as xcattest enhancement.

daniceexi:
@whowutwut @tingtli We also have discussed the requirement that includes test cases from another bundle file. I have opened a story 117903 to track it.

neo954:
I will merge this one.

PR #1658, opened 2016-08-10
Probe package, add ServiceNodeUtils.pm to probe lib

hu-weihua:
xCAT probe code which support hierarchy leverages xCAT code ServiceNodeUtils.pm, so add ServiceNodeUtils.pm to package script.


daniceexi:
The change looks good for me!

PR #1661, opened 2016-08-10
Remove xcat-genesis-scripts-* from xcatsn depended packages, the work…

zet809:
…ing directory /tftpboot/ on SN was mounted from MN

The xcat-genesis-base/scripts-* will install files under /opt/xcat/share/xcat/netboot/genesis/$arch/.
But the kernel and initrd will be put under /tftpboot/ when running mknb. 
For xCAT sn, the /tftpboot/ can be either mount from MN or rsync from MN. Don't need to install them separately.

### DIFF
--- a/xCATsn/debian/control
+++ b/xCATsn/debian/control
@@ -7,7 +7,7 @@ Standards-Version: 3.7.2
 
 Package: xcatsn
 Architecture: all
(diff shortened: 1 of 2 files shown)


immarvin:
The 2 packages are required in %post of xCATsn when site.sharedtftp=0
````
# Run mknb if xCAT-genesis-scripts-x86_64 changed
%ifos linux
# prevent running it during install into chroot image and only run it if xCAT-genesis-scripts-x86_64
# was recently updated.
if [ -f "/proc/cmdline" -a -f "/etc/xcat/genesis-scripts-updated" ]; then
  rm -f /etc/xcat/genesis-scripts-updated
  # On the SN do not run mknb if we are sharing /tftpboot with the MN
  SHAREDTFTP=`/opt/xcat/sbin/tabdump site | grep sharedtftp | cut -d'"' -f 4`
  if [ "$SHAREDTFTP" != "1" ]; then
      . /etc/profile.d/xcat.sh
      echo Running '"'mknb `uname -m`'"', triggered by the installation/update of xCAT-genesis-scripts-x86_64 ...
      mknb `uname -m`
    fi
fi
````

Issue #1589, opened 2016-07-26
go-xcat -help has problem. 


wwzhubj:
What I want is help message, but go-xcat runs the same way when with'-help' and without '-help'. 
```
[root@c712ems4 yum.repos.d]# go-xcat -help
Operating system:   linux
Architecture:       ppc64le
Linux Distribution: rhel
Version:            7.2

Reading repositories ...... done

xCAT Core Packages
==================

Package Name                Installed                      In Repository
------------                ---------                      -------------
perl-xCAT                   2.12.1-snap201607070618        2.12.1-snap201607070618
xCAT                        2.12.1-snap201607070618        2.12.1-snap201607070618
xCAT-buildkit               2.12.1-snap201607070618        2.12.1-snap201607070618
...
xCAT-genesis-base-ppc64     2.12-snap201605051443          2.12-snap201607221057
xCAT-genesis-base-x86_64    (not installed)                2.12-snap201607221057
xCAT-genesis-x86_64         (not installed)                2.7.7-snap201301100842
xnba-undi                   (not installed)                1.0.3-131028
yaboot-xcat                 (not installed)                1.3.17-rc1
```

daniceexi:
I also prefer the run ``go-xcat`` without any parameter, the default action is to display the help message.

neo954:
Will working on it.

whowutwut:
:+1:  I also agree with @daniceexi , I just started trying the `go-xcat` command and hit the same issues as described by @wwzhubj

I think `go-xcat` without any options should either:

1. install the latest GA of xCAT (update if xCAT is already installed and there is an avail update) 
2. display the help message 

I think 2 is better in this case to not complicate things

whowutwut:
Verified with the latest downloadable version of `go-xcat` 

```
# /tmp/go-xcat -help
go-xcat: invalid option -- '-help'
go-xcat: Try `/tmp/go-xcat --help' for more information
```

running without --help causes the usage message to appear.

Issue #1603, opened 2016-07-29
[FVT]:go xcat failed with  xcat-genesis-scripts installation failure


tingtli:
xCAT 2.12 on ubuntu 16.04 

```
root@c910f03c17k29:/# ./go-xcat --xcat-core=/xcat-core -y install
Operating system:   linux
Architecture:       ppc64le
Linux Distribution: ubuntu
Version:            16.04

Reading repositories ...... done
Reading package lists...
Building dependency tree...
Reading state information...
Package xcat-genesis-scripts is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  xcat-genesis-scripts-x86-64 xcat-genesis-scripts-ppc64

E: Package 'xcat-genesis-scripts' has no installation candidate

xCAT Core Packages
==================

Package Name                Installed                      In Repository
------------                ---------                      -------------
perl-xcat                   (not installed)                2.12-snap201607280616
xcat                        (not installed)                2.12-snap201607280616
xcat-buildkit               2.12-snap201607070619          2.12-snap201607280616
xcat-client                 (not installed)                2.12-snap201607280616
xcat-confluent              (not installed)                2.12-snap201607280616
xcat-genesis-scripts        (not installed)                (none)
xcat-genesis-scripts-ppc64  (not installed)                2.12-snap201607280616
xcat-genesis-scripts-x86-64 (not installed)                2.12-snap201607280616
xcat-probe                  (not installed)                2.12-snap201607280616
xcat-server                 (not installed)                2.12-snap201607280616
xcat-test                   (not installed)                2.12-snap201607280616
xcat-vlan                   (not installed)                2.12-snap201607280616
xcatsn                      (not installed)                2.12-snap201607280616

xCAT Dependency Packages
========================

Package Name                Installed                      In Repository
------------                ---------                      -------------
conserver-xcat              8.1.16-3                       8.1.16-3
elilo-xcat                  3.14-4                         3.14-4
grub2-xcat                  2.02-0.16.el7.snap201506090204 2.02-0.16.el7.snap201506090204
ipmitool-xcat               1.8.15-2                       1.8.15-2
syslinux-xcat               3.86-2                         3.86-2
xcat-genesis-base-ppc64     (not installed)                2.12-snap201607211138
xcat-genesis-base-x86-64    (not installed)                2.12-snap201607212311
xnba-undi                   1.0.3-7                        1.0.3-7


Boo-boo
[...]

PR #1663, opened 2016-08-10
[go-xcat] Version 1.0.2 Fix a couple of bugs.

neo954:


neo954:
This pull request fix issue #1589 and issue #1603.

daniceexi:
@whowutwut if you would like review this change, please do it today. Since we are very close to the end of cycle test, China team will try to make this change merged and tested tomorrow.

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -1309,6 +1380,10 @@ do
 		usage
 		exit 0
 		;;
+	"--long-help")
```
immarvin: I do not think this fix the issue https://github.com/xcat2/xcat-core/issues/1589.
The final decision is that when "go-xcat" is run without no option or invalid option, just as "-help" in the issue, the xcat-usage should be displayed.

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -60,6 +60,65 @@ function usage()
...
+	  check                         check the version of the installed packages
+	                                of xCAT and packages in the repository
+	EOF
+	println 2
+	while read -r ; do echo "${REPLY}" ; done <<-EOF
+	  smoke-test                    preform basic tests of the xCAT installation
+	EOF
+	println 11
+	while read -r ; do echo "${REPLY}" ; done <<-EOF
+	  ${script} --xcat-core=/path/to/xcat-core.repo install
```
immarvin: according to the decision and the subroutine ``usage()``. the valid usage should only include
````

Examples:
  go-xcat
  go-xcat install
  go-xcat update
  go-xcat --yes install
  go-xcat -x 2.12 -y install
  go-xcat --xcat-version=devel install
  go-xcat --xcat-core=/path/to/xcat-core.tar.bz2 \
          --xcat-dep=/path/to/xcat-dep.tar.bz2 install
  go-xcat --xcat-core=http://xcat.org/path/to/xcat-core.tar.bz2 \
          --xcat-dep=http://xcat.org/path/to/xcat-dep.tar.bz2 install
```` 

please remove the examples out of the usages above

neo954: The function ``verbose_usage()`` is just for my own reference. It will bother the end user. Since it need the end user to pass the secret command line argument ``--long-help``. Unless the user read the source, he/she will not find out.

neo954:
@immarvin, I made anther commit to fix the problem when user passed an improper command line argument.

immarvin:
agree to merge

Issue #1666, opened 2016-08-10
[DEV] Diskless boot failed for rhel7.3 because selinux disabled.


cxhong:
I am trying to boot up a diskless node with rhel7.3 image.  it failed at:
````
[   64.538308] systemd[1]: Failed to load SELinux policy.
[!!!!!!] Failed to load SELinux policy, freezing.
````
looking into rootimg dir,  the selinux is enabled.
````
[root@fs1 rootimg]# cat etc/selinux/config

# This file controls the state of SELinux on the system.
# SELINUX= can take one of these three values:
#     enforcing - SELinux security policy is enforced.
#     permissive - SELinux prints warnings instead of enforcing.
#     disabled - No SELinux policy is loaded.
SELINUX=enforcing
# SELINUXTYPE= can take one of three two values:
#     targeted - Targeted processes are protected,
#     minimum - Modification of targeted policy. Only selected processes are protected.
#     mls - Multi Level Security protection.
SELINUXTYPE=targeted
``````
xCAT disable the selinux during provision and cause boot hang with that error message

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1676

PR #1676, opened 2016-08-11
fix issue  [DEV] Diskless boot failed for rhel7.3 because selinux disabled. #1666 ; disable selinux in the rootimg

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1666

Redhat 7.3 will install selinux-policy package by default, in the /etc/selinux/config, selinux will be enabled by default, can be disabled on genimage


daniceexi:
It looks OK for this change. But what's the root cause the selinux hang?

whowutwut:
oops, I didn't see @daniceexi  comment above, so removed the pending in case there work here is not fully complete.

Issue #1356, opened 2016-06-21
pkglist for CUDA8 requires grub2-tools


markprez:
When running genimage on base RH72 with CUDA8 repo we get the errors:
```
  Installing : 1:xorg-x11-drv-nvidia-361.62-1.el7.ppc64le                 41/58
/var/tmp/rpm-tmp.YOiQYC: line 8: grub2-mkconfig: command not found
error opening /etc/yaboot.conf for read: No such file or directory
ls: cannot access /boot/vmlinuz-*.el7.ppc64le*: No such file or directory
```
Adding grub2-tools to the pkglist fixed this issue.

cxhong:
yeh, after we added grub2-tools to pkglist, the above message is fixed, but following issue is created:
````
/usr/sbin/grub2-probe: error: cannot find a device for / (is /dev mounted?).
error opening /boot/grub/grub.cfg for read: No such file or directory
`````
on our system, we had /boot/grub2/grub.cfg
````
# rpm -qf /boot/grub2/grub.cfg
grub2-2.02-0.29.el7.ppc64le
`````
so, even installed grub2  packages, those message will still show up.  

CUDA7.5 had same message during the genimage
````
/var/tmp/rpm-tmp.6hCJAI: line 8: grub2-mkconfig: command not found
error opening /etc/yaboot.conf for read: No such file or directory
`````
this message seems didn't cause anything, the image is generated successfully and diskless CUDA7.5 installation and verification worked fine.

cxhong:
For cuda8 on rhel7.3,  addiing grub2 and grub2-tools to package list will get rid of the error messages on genimage.

whowutwut:
Closing this issue, @markprez  please re-open or create another issue if the problem persists.

PR #1678, opened 2016-08-11
CUDA8 support on rhel7.2 and rhel7.3

cxhong:
for issue #1356,
1) add grub2 and grub2-tools to netboot package list
2) remove hard code of cuda 7.5 to cuda
3) get rid of cuda version on the config_cuda postscripts

### DIFF
--- a/xCAT-server/share/xcat/install/rh/cudaruntime.rhels7.ppc64le.pkglist
+++ b/xCAT-server/share/xcat/install/rh/cudaruntime.rhels7.ppc64le.pkglist
@@ -1,9 +1,13 @@
 #INCLUDE:compute.rhels7.pkglist#
 
-#For Cuda 7.5
+#For Cuda
 kernel-devel
 gcc
 pciutils
 dkms
+#To install cuda-runtime, need to specify the version
+#if want to install cuda-runtime-8-0, uncomment it,
+#then comment out cuda-runtime-7-5
+#cuda-runtime-8-0
 cuda-runtime-7-5
 
--- a/xCAT-server/share/xcat/install/rh/cudaruntime.rhels7.ppc64le.pkglist
+++ b/xCAT-server/share/xcat/install/rh/cudaruntime.rhels7.ppc64le.pkglist
@@ -6,8 +6,8 @@ gcc
 pciutils
 dkms
 #To install cuda-runtime, need to specify the version
-#if want to install cuda-runtime-8-0, uncomment it,
-#then comment out cuda-runtime-7-5
-#cuda-runtime-8-0
-cuda-runtime-7-5
+#if want to install cuda-runtime-7-5, uncomment it,
+#then comment out cuda-runtime-8-0
+#cuda-runtime-7-5
+cuda-runtime-8-0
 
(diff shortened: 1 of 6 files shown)


xCAT/postscripts/config_cuda
```diff
@@ -1,8 +1,7 @@
...
 
-# set the paths required for cuda7.5
-CUDA_VER="cuda-7.5"
-FILENAME="/etc/profile.d/xcat-${CUDA_VER}.sh"
+# set the paths required for cuda
+FILENAME="/etc/profile.d/xcat-cuda.sh"
 
-echo "export PATH=/usr/local/${CUDA_VER}/bin:\$PATH" > ${FILENAME}
-echo "export LD_LIBRARY_PATH=/usr/local/${CUDA_VER}/lib64:\$LD_LIBRARY_PATH" >> ${FILENAME}
+echo "export PATH=/usr/local/cuda/bin:\$PATH" > ${FILENAME}
```
daniceexi: Does the path ``/usr/local/cuda/bin`` work for both cuda-7.5 and cuda-8? If yes, I am curious why the path ``/usr/local/cuda-7.5/bin`` worked before for cuda-7.5?

daniceexi:
Recently, @bybai might not have time to review the change for cuda. 
@cxhong If you cannot receive the comment from @bybai in time, you can ping me or @whowutwut for review.

xCAT/postscripts/config_cuda
cxhong: cuda path is linked to cuda-7.5 or cuda-8
````
drwxr-xr-x 12 root root 4096 Aug 10 16:31 cuda-7.5
lrwxrwxrwx  1 root root    8 Aug 10 16:31 cuda -> cuda-7.5

`````

Comment (whowutwut) [inline xCAT/postscripts/config_cuda]: Can we add a variable for this for future protection...  something like... 

```
CUDA_PATH=/usr/local/cuda
echo "export PATH=\$CUDA_PATH/bin:\$PATH" > ${FILENAME}
echo "export LD_LIBRARY_PATH=\$CUDA_PATH/lib64:\$LD_LIBRARY_PATH" >> ${FILENAME}
```

File: xCAT-server/share/xcat/install/rh/cudaruntime.rhels7.ppc64le.pkglist [hunk 2]
```diff
@@ -1,9 +1,13 @@
...
-#For Cuda 7.5
+#For Cuda
 kernel-devel
 gcc
 pciutils
 dkms
+#To install cuda-runtime, need to specify the version
+#if want to install cuda-runtime-8-0, uncomment it,
+#then comment out cuda-runtime-7-5
+#cuda-runtime-8-0
```
whowutwut: My thought here is that maybe we should support CUDA8 by default and leave CUDA7 code in 2.11 and maybe 2.12. and 2.12.1.. from 2.12.2 onwards, we support CUDA8, so then just change line 12 to cuda-runtime-8-0 and delete lines 8-11...

daniceexi: I agree that we should keep the latest CUDA version for the current xCAT release.

whowutwut:
I have reviewed and will merge

Issue #1097, opened 2016-05-12
Nodeset for Ubuntu OS displays warning, but we should be more specific and only print on conditions where it might fail


whowutwut:
when provisioning Ubuntu OS, the following warning messages comes out: 

```
# nodeset fs4 osimage=ubuntu14.04.3-ppc64el-install-compute
Warning: rcons my not work since no serialport specified  <====
fs4: install ubuntu14.04.3-ppc64el-compute
```

We should identify the case where rcons does not work and only print the msg then. 
The message should also include the chdef command to help the user set the serial port attribute. 

Also "may" is spelled wrong ("my")

daniceexi:
@whowutwut 
@immarvin is working on an item that predefine a bunch of node templates. After the node template is supported, we can direct the user to refer to a template for the node definition.

whowutwut:
Is there any small design or anything to explain the concept. I'd be interested to see the use case. I've been keeping bare node definitions for some of the hardware I have to move them around or re-provision in the event that the mgmt node is lost. I'd like to see if we are address this problem with this new templates

Or basically, from the node template. How can I easily define the nodes.  I'd like to not save the info but it's very difficult to recreate...

immarvin:
hi @whowutwut , the basic idea is:
1.  package serveral node definition templates in xcat, includding “powerNV”, "systemx","kvm","lpar", etc. , a set of template nodes based on these templates will be created on xcat installation or ``xcatconfig -<some option>``.   
2. add a new option "-t <template node def/existing node def>" for "mkdef",   so that the node definition can be created based on the template nodes or existing nodes,  with the specified attributes.

I will send out the mini-design after investigation.

immarvin:
The message is refined in this PR https://github.com/xcat2/xcat-core/pull/1686

The --template option has been supported 
https://github.com/xcat2/xcat-core/pull/1254

PR #1686, opened 2016-08-12
fix issue  Nodeset for Ubuntu OS displays warning, but we should be more specific and only print on conditions where it might fail #1097

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1097

refine the warning message when node.serialport not specified


penguhyang:
agree to merge

PR #1690, opened 2016-08-12
add rhels7.3 bundle

junxiawang:
@tingtli  please help review


tingtli:
@junxiawang i think we should use rhels7.2_x86_64.bundle as prototype and remove the migration test. Right?

PR #1691, opened 2016-08-12
Fix bug1687 xcatprobe could not surpport xcatmn command on ubuntu

hu-weihua:
Fix bug1687 xcatprobe could not surpport xcatmn command on ubuntu
Remove 2 useless files


immarvin:
agree to merge

Issue #1689, opened 2016-08-12
[FVT]on rhels6.8, rootimg.tar.gz is only 20 bytes after run packimage -m tar  rhels6.8-ppc64-netboot-compute


tingtli:
rhels6.8 xCAT 2.12.2
```
[root@c910f02c01p22 compute]# packimage rhels6.8-ppc64-netboot-compute -m tar
Packing contents of /install/netboot/rhels6.8/ppc64/compute/rootimg
archive method:tar
compress method:gzip

[root@c910f02c01p22 compute]# ls -l
total 272472
-rw-------  1 root root  24737902 Jun 17 02:14 initrd-stateless.gz
-rw-------  1 root root  23000676 Jun 17 02:15 initrd-statelite.gz
-rw-r--r--  1 root root  17241480 Jun 17 02:14 kernel
dr-xr-xr-x 24 root root      4096 Aug 12 04:30 rootimg
-rw-r--r--  1 root root 214008614 Jun 17 02:07 rootimg.gz
-rw-r--r--  1 root root        20 Aug 12 04:30 rootimg.tar.gz
```

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1694

Issue #1688, opened 2016-08-12
[FVT]:diskless provision broken using rootimg.tar.gz created by packimage -m tar osimage 


tingtli:
ubuntu16.04.1. xCAT 2.12.2
This is the steps to create rootimg.tar.gz
```
root@c910f04x27v02:/install/netboot/ubuntu16.04.1/x86_64/compute# packimage -m tar ubuntu16.04.1-x86_64-netboot-compute
Packing contents of /install/netboot/ubuntu16.04.1/x86_64/compute/rootimg
archive method:tar
compress method:gzip

root@c910f04x27v02:/install/netboot/ubuntu16.04.1/x86_64/compute# nodeset c910f04x27v04 osimage=ubuntu16.04.1-x86_64-netboot-compute
c910f04x27v04: netboot ubuntu16.04.1-x86_64-compute
root@c910f04x27v02:/install/netboot/ubuntu16.04.1/x86_64/compute# rpower c910f04x27v04 reset
c910f04x27v04: reset
root@c910f04x27v02:/install/netboot/ubuntu16.04.1/x86_64/compute# ls
initrd-stateless.gz  initrd-statelite.gz  kernel  rootimg  rootimg.tar.gz
root@c910f04x27v02:/install/netboot/ubuntu16.04.1/x86_64/compute# ls -l
total 1598080
-rw-r--r--  1 root root   87128518 Aug 11 18:30 initrd-stateless.gz
-rw-r--r--  1 root root   87511832 Aug 11 18:29 initrd-statelite.gz
-rw-------  1 root root    7047504 Aug 11 18:28 kernel
drwxr-xr-x 23 root root       4096 Aug 12 04:03 rootimg
-rw-r--r--  1 root root 1453113967 Aug 12 04:08 rootimg.tar.gz
root@c910f04x27v02:/install/netboot/ubuntu16.04.1/x86_64/compute# file rootimg.tar.gz
rootimg.tar.gz: gzip compressed data, last modified: Fri Aug 12 08:03:38 2016, from Unix
```
Monitor the diskless provision screen and it gets broken here
```
cp: can't create '/sysroot/etc/resolv.conf': No such file or directory
mount: mounting /dev on /sysroot/dev failed: No such file or directory
mount: mounting /proc on /sysroot/proc failed: No such file or directory
mount: mounting /sys on /sysroot/sys failed: No such file or directory
switch_root: can't open '/dev/console': No such file or directory
[...]

PR #1694, opened 2016-08-13
[packimage -m -c]:  supress the verbose info of rootimg decompression; add tar "--no-recursion" option in packimage; correct the path of gnu tar to be "/bin/tar" in genimage

immarvin:
supress the verbose info of rootimg decompression during diskless boot to reduce the time of rootimg decompress; 
fix https://github.com/xcat2/xcat-core/issues/1688: 
1. add tar "--no-recursion" option in packimage, to prevent the duplicate compress of directory and the files under the directory.
2. the gnu tar in rootimg is located in "/bin/tar" instead of "/usr/bin/tar"
3. in the initrd, the full path to the gnu tar "/bin/tar" should be used, otherwise, tar in busybox will be used

fix https://github.com/xcat2/xcat-core/issues/1689:
1. the return value of the rootimg archive/compress should be checked and prompt user about the error if any
2. use "--use-compress-program" instead of piped "find ,...| tar ...|gzip/pigz/xz ..." to capture the error message during archive if any 

fix https://github.com/xcat2/xcat-core/issues/1695:
when run "packimage", remove all the existed compressed rootimg to make sure the  newly generated one will be used on "nodeset"

add "-m" and "-c" to "packimage" usage info of "-h|--help" and manpage
The updated doc: http://10.3.5.21/xcat-doc/html/guides/admin-guides/references/man1/packimage.1.html

### DIFF
--- a/docs/source/guides/admin-guides/references/man1/packimage.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/packimage.1.rst
@@ -23,7 +23,7 @@ SYNOPSIS
 
 \ **packimage  [-v| -**\ **-version]**\ 
 
-\ **packimage**\  \ *imagename*\ 
+\ **packimage**\  [\ **-m | -**\ **-method**\  \ *cpio|tar*\ ] [\ **-c | -**\ **-compress**\  \ *gzip|pigz|xz*\ ]  \ *imagename*\ 
 
 
 ***********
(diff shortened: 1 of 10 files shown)


daniceexi:
@penguhyang has reviewed this change.
I am OK with the change too.

Issue #1529, opened 2016-07-13
[Customer] nodeset failed with meaningless messages


daniceexi:
A customer encountered following error when run nodeset:
~~~~
[root@xcat ~]# nodeset s01n01 osimage
Error: xnba plugin bug, pid 24575, process description: 'xcatd SSL: nodeset to s01n01 for root@localhost: xnba instance: destiny instance: anaconda instance' with error 'Died at /opt/xcat/sbin/xcatd line 1876.
' while trying to fulfill request for the following nodes: s01n01
~~~~

The root cause was the node ``s01n01`` could not be resolved to an IP address. We does have a log like following in the /var/log/xcat/cluster.log.
~~~~
Jul 13 19:01:07 xcat xcat[25834]:  cannot resolve the network configuration of f01n01 at /opt/xcat/lib/perl/xCAT/Template.pm line 958.
~~~~

But I don't think it's enough. We should display a meaningful message in the CLI.

immarvin:
fix https://github.com/xcat2/xcat-core/pull/1696

daniceexi:
I am OK with the change.

@immarvin The code logic is OK, but the code format needs be refined to follow xCAT's code format standard. I'll merge it now catch the RC.

PR #1696, opened 2016-08-14
fix issue  [Customer] nodeset failed with meaningless messages #1529

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1529


daniceexi:
I am OK with the change.

@immarvin The code logic is OK, but the code format needs be refined to follow xCAT's code format standard. I'll merge it now catch the RC.

PR #1700, opened 2016-08-15
use dependency on mn to run test

tingtli:


tingtli:
@penguhyang  please review

penguhyang:
looks good, agree to merge

Issue #898, opened 2016-04-08
[FVT]: No hierarchy support for sles12 and sles12.1


tingtli:
When i setup test environment on Jenkins, I setup cluster to run regression on sles12 and sles12.1. 
I found there is no service image. And the old way to install mysql is not correct. Please take a look at the hierarchy support on sles12 and sles12.1.

cxhong:
@immarvin, @tingtli  I will take look this issue.

whowutwut:
we want to check the 2.11 branch to see if it also exist, we should fix since hierarchy is considered basic function for RHEL  & SLES

daniceexi:
This feature will be done in 2.12.1

whowutwut:
@hu-weihua  can we have someone verify on behalf of Ting Ting and close out?

hu-weihua:
@junxiawang , Could you help to take look at this issue? thank you very much.

junxiawang:
@whowutwut @daniceexi @cxhong 
sles12.1 on ppc64le could work for hierarchy
But sles12 on x86_64 could not work

cxhong:
since mysql is running on the system before we ran "mysqlsetup -i" , xcat will not run initialization again.
`````
c910f03c04k04:~ # mysqlsetup -i
MySQL is running.  Database initialization will not take place.
``````
so if password is not matched, it will exit with "Wrong MySQL root password".   

For debian and sles12.1,   after install mysql/mariadb server, the mysql will running, so we will stop mysql first, then run initialization.  

I don't think mysql is running after installation for sles12.  so, this is correct behavior for our current design.  

@daniceexi @immarvin ,  do we need to change the code to stop DB service for any OS if it is "-i" flag ?

immarvin:
hi @cxhong , I think we do not need to stop mysql in the "mysqlsetup" code for the linux distributions except debian and sles12.1 since the mysqlsetup manpage has  mentioned this. I still have 2 questions on this:
1. when the message "MySQL is running.  Database initialization will not take place." is raised, why does "mysqlsetup -i" not return? 
2. the message "MySQL is running.  Database initialization will not take place." is raised, why does "mysqlsetup -i"  is not instructive enough for user, how about modify it to be "MySQL is running. Please stop it first"?

cxhong:
[...]

PR #1702, opened 2016-08-15
mysqlsetup failed becuase /usr/bin/mysql_install_db is not available

cxhong:
for issue #898 ,  the mysqlsetup -i failed because /usr/bin/mysql_install_db is not available.
Add the error message if /usr/bin/mysql_install_db is not exists.
`````
# mysqlsetup -i
Input the alpha-numberic  password for xcatadmin in the MySQL database:
Input the password for root in the MySQL database:
 /usr/bin/mysql_install_db is not available, please install required packages
``````

### DIFF
--- a/xCAT-client/bin/mysqlsetup
+++ b/xCAT-client/bin/mysqlsetup
@@ -916,12 +916,13 @@ sub initmysqldb
     }
     else
     {
-        if (!(-e ("/usr/bin/mysql_install_db"))) {
-            xCAT::MsgUtils->message("E", "/usr/bin/mysql_install_db is not available, please install required packages");
+        my $sqlcmd = "/usr/bin/mysql_install_db";
+        if (!(-e ($sqlcmd))) {
+            xCAT::MsgUtils->message("E", "$sqlcmd is not available, please install required mysql/mariadb packages");
             exit(1);
         }
 
-        $cmd = "/usr/bin/mysql_install_db --user=mysql";
+        $cmd = "$sqlcmd --user=mysql";
     }
     xCAT::Utils->runcmd($cmd, 0);
     if ($::RUNCMD_RC != 0)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by cxhong** [comment]
@immarvin , can u review this?

File: xCAT-client/bin/mysqlsetup [hunk 1]
```diff
@@ -916,6 +916,11 @@ sub initmysqldb
     }
     else
     {
+        if (!(-e ("/usr/bin/mysql_install_db"))) {
```
whowutwut: Can that `/usr/bin/mysql_install_db` be a variable and reused in the error message below.

xCAT-client/bin/mysqlsetup
```diff
@@ -916,6 +916,11 @@ sub initmysqldb
...
+            xCAT::MsgUtils->message("E", "/usr/bin/mysql_install_db is not available, please install required packages");
```
whowutwut: Can we specify which package provides mysql_install_db?

whowutwut: is it difficult to check the required packages in the script, if the user satisfies the one package, will they hit another error in dependency, i think there's a few packages that are required for mysql/mariadb support...

xCAT-client/bin/mysqlsetup
```diff
@@ -916,7 +916,13 @@ sub initmysqldb
     }
     else
     {
-        $cmd = "/usr/bin/mysql_install_db --user=mysql";
+        my $sqlcmd = "/usr/bin/mysql_install_db";
+        if (!(-e ($sqlcmd))) {
```
immarvin: should the test to be "-x"?

cxhong: -e  File exists.
 -x  File is executable by effective uid/gid

I will change to -x since mysql_install_db is a command

immarvin:
agree to merge

PR #1704, opened 2016-08-16
Fix issue, xcat-core builed failed in ubuntu due to xcat probe code checking

hu-weihua:
Fix issue, xcat-core builed failed in ubuntu due to xcat probe code checking


zet809:

PR #1709, opened 2016-08-16
Add dependency to support rhels7.2+IB

hu-weihua:
Support rhels7.2+IB in ESP,  need one more dependency ``ethtool``. Add it into ``ib.rhels7.ppc64le.pkglist``


zet809:

PR #1710, opened 2016-08-16
Fix bug 1017: xcatprobe osdeploy -n node -r 24h could not get right deploy messages

hu-weihua:


zet809:
Looks good to me, I agree to merge

PR #1713, opened 2016-08-16
Provide Security Notice for Vulnerability created on 2016-08-15

whowutwut:
OpenSSL is vulnerable to a denial of service, caused by an out-of-bounds read in the TS_OBJ_print_bio function. A remote attacker could exploit this vulnerability using a specially crafted time-stamp file to cause the application to crash.


whowutwut:
Merge to close out PSIRT task

Issue #1714, opened 2016-08-17
Remove the Debian Support Message in the xcatprobe command.


daniceexi:
Since xCAT does not support latest Debian for a while, please remove the messages which related to `Debian` in the xcatprobe command.

hu-weihua:
@daniceexi , I have fixed this issue by pull request #1715 , Could you review and merge it? thanks.

whowutwut:
@daniceexi  Can you close this if resolved?

daniceexi:
I have verified the change.

PR #1715, opened 2016-08-17
Remove the Debian Support Message in the xcatprobe command.

hu-weihua:
Fix issue #1714 Remove the Debian Support Message in the xcatprobe command.


daniceexi:
The change looks OK.

Issue #1305, opened 2016-06-14
Install required genesis packages to support mixed arch "out of the box" 


whowutwut:
Created this issue based on discussions in #1238 

If we install the genesis packages on both platforms...
- x86_64 mgmt node will install ppc64 genesis packages
- ppc64(le) mgmt nodes will install x86_64 genesis packages

Then it will simplify the pre-requisites in a mixed arch environment where the xCAT mgmt node does not match the architecture of the compute nodes for doing discovery... This will reduce the complexity of our documentation for mixed arch support.

This issue is created to track the following work:

1. Make the changes to install the genesis packages by default , ppc64(le) systems will install x86_64 and vice versa. 
2. Clean up the documentation so that we no longer need to describe the process for installing the various rpm packages needed to support the mixed arch environment.

immarvin:
hi @zet809 , has this issue already been fixed? if yes, please set it to "pending", thanks

whowutwut:
@immarvin  @zet809  I set this to pending during our call because we discussed that the fix has been made.  However, in my testing, I don't see it completely done...  I think we should remove the pending status.. Is there a pull request that we can associate with this Issue? 

I see the genesis base installed for x86_64 and ppc64 
```
[root@fs1 ~]# rpm -qa | grep genesis
xCAT-genesis-scripts-x86_64-2.12.2-snap201608160911.noarch
xCAT-genesis-base-ppc64-2.12-snap201607221057.noarch
xCAT-genesis-base-x86_64-2.12-snap201607221057.noarch
xCAT-genesis-scripts-ppc64-2.12.2-snap201608160911.noarch
```

But I don't see anywhere that we run the mknb so that x86_64 is ready to go... 
This is confirmed by runnning `xcatprobe discovery`

```
[root@fs1 discovery.fs1]# xcatprobe discovery
Attribute 'master' in 'site' table is configured well                                                             [ OK ]
Genesis files for ppc64le are available                                                                           [ OK ]
Genesis files for x86 are not available                                                                           [WARN]  <=====
If don't plan to manage a x86 server, please ignore above warning                                                 [INFO]
DHCP dynamic range is configured well                                                                             [ OK ]
```

I think we should be adding some code to run mknb for both architectures.  Where is this done today?

whowutwut:
Looks good with todays build.

PR #1716, opened 2016-08-17
run mknb for both ppc64 and x86_64

zet809:
Fix issuse #1305


daniceexi:

PR #1718, opened 2016-08-17
add cases for the packimage -m -c option

tingtli:
@penguhyang  packimage cases, please help review. Thanks!


penguhyang:
agree to merge

PR #1721, opened 2016-08-17
[go-xcat] Fix typo

neo954:


whowutwut:
Should we be bumping the version up for go-xcat and uploading it to xcat.org on each commit?

PR #1722, opened 2016-08-17
Update rest api for rinstall and console

chenglch:
- Refactor console api, console will be looked as a service resource
- Add rest api to support rinstall


zet809:
A small concern is the using of <noderange> in the URL may cause problem, such as "/nodes/node1,node3-node5/status", I think we shall enhance its process in the future.

PR #1724, opened 2016-08-17
Protect umount_chroot from running with rootimage_dir undefined

mattaezell:
We had an issue with our kernel version, so xCAT called xdie.  There's a typo in the RHEL 7.2 xdie, so it called umount_chroot with a blank rootimage_dir, which tried to unmount the xCAT server's /proc, /sys, and delete its device files.  Fix the typo and add some logic to umount_chroot to prevent this in the future

### DIFF
--- a/xCAT-server/share/xcat/netboot/rh/genimage
+++ b/xCAT-server/share/xcat/netboot/rh/genimage
@@ -70,7 +70,7 @@ my $noupdate;
 
 sub xdie {
     system("rm -rf /tmp/xcatinitrd.$$");
-    umount_chroot($rootimage_dir);
+    umount_chroot($rootimg_dir);
     die @_;
 }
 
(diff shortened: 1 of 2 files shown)


hu-weihua:
@mattaezell , thank you very much for your contribution. it's helpful. We have already found other potential risks about handling rootimage_dir too. so we plan to enhance this part( including your problem)later.  Thank you again.

mattaezell:
We hit this today on our new GTB system.  I guess we will patch locally...

whowutwut:
@hu-weihua Is there any information you can provide as to how we plan to do to address this issue and how it differs from this pull request so @mattaezell  can work around the issues better.  

I'm going to put this as high priority so we implement the fix in the next release.

hu-weihua:
@whowutwut , We did hit another ``rootimg_dir`` problem which cause umount failure. We planed to fix all in ``genimage``,  including @mattaezell  code.  But unfortunately,  nobody take this task now. So I think we can merge @mattaezell 's pull request.  Sorry for delay this pull request.

adambertsch:
I note that rh/genimage had use strict at some point, but that it was commented out.  use strict would have likely caused this bug to be found and fixed prior to it ever making it in to the tree.  Would the xCAT team be open to either restoring use strict and making the needed updates to this file, or accepting a patch that does this?

immarvin:
hi @adambertsch , another PR on the strict guard for mount points under rootimg dir is created https://github.com/xcat2/xcat-core/pull/2036

immarvin:
hi @adambertsch , the #2036 has been merged

Issue #1717, opened 2016-08-17
[fvt]2.12.2 xcatprobe osdeploy -n could not monitor the node deploy process if message's log : dhcp[process id]


junxiawang:
env:ubuntu16.04.1
build:
root@c910f04x21v02:~# lsxcatd  -v
Version 2.12.2 (git commit 9298b5ff211500e22398abd7c35b3fe28038feff, built Tue Aug 16 09:12:17 EDT 2016)
root@c910f04x21v02:~# 
How to reproduce:
``` 
xcatprobe osdeploy -n c910f04x21v04
 

All pre_defined nodes are valid [ OK ]
The installation network interface is ens3 [INFO]
-------------------------------------------------------------
Start to capture every message during os provision process......
-------------------------------------------------------------

[c910f04x21v04] Node status is changed to powering-on
[c910f04x21v04] Node status is changed to installing
[c910f04x21v04] xcat Wed Aug 17 02:26:00 EDT 2016 Running postscript: syslog
[c910f04x21v04] xcat Wed Aug 17 02:26:00 EDT 2016 postscript syslog return with 0
[c910f04x21v04] xcat Wed Aug 17 02:26:01 EDT 2016 Running postscript: remoteshell
[c910f04x21v04] xcat Wed Aug 17 02:26:04 EDT 2016 postscript remoteshell return with 0
[c910f04x21v04] xcat Wed Aug 17 02:26:04 EDT 2016 Running postscript: syncfiles
[c910f04x21v04] xcat Wed Aug 17 02:26:04 EDT 2016 postscript syncfiles return with 0
[c910f04x21v04] Node status is changed to booting
[c910f04x21v04] xcat running /xcatpost/mypostscript.post
[c910f04x21v04] xcatpostinit1[1834]: Wed Aug 17 02:26:39 EDT 2016 Running postscript: otherpkgs
[c910f04x21v04] xcatpostinit1[1834]: Wed Aug 17 02:26:39 EDT 2016 Running postscript: otherpkgs
[c910f04x21v04] xcat Wed Aug 17 02:26:39 EDT 2016 Running postscript: otherpkgs
[c910f04x21v04] xcat: Running otherpkgs
[c910f04x21v04] xcatpostinit1[1834]: ./otherpkgs: no extra rpms to install
[c910f04x21v04] xcatpostinit1[1834]: ./otherpkgs: no extra rpms to install
[...]

PR #1726, opened 2016-08-18
Fix probe bug #1717 log parse error when the lable of log looks like "dhcp[process id]"

hu-weihua:


zet809:

Issue #1727, opened 2016-08-18
[fvt]2.12.2 xcatprobe discovery Unescaped left brace


junxiawang:
env: ubuntu 16.04.1
build:root@c910f03c17k27:~# lsxcatd -v
Version 2.12.2 (git commit 29eeef24abec7692194ef4c71b5cccfab89b01d1, built Wed Aug 17 06:21:09 EDT 2016)
root@c910f03c17k27:~# 
How to reproduce:
```
root@c910f03c17k27:~# xcatprobe discovery -n all
Unescaped left brace in regex is deprecated, passed through in regex; marked by <-- HERE in m/\${ <-- HERE next-server}/ at /opt/xcat/probe//subcmds/discovery line 207.
Option '-n' must used with '-m'                                                                                   [FAIL]
Usage:                                  
```

junxiawang:
verified

PR #1733, opened 2016-08-18
Fix bug #1727, probe sub_command discovery has syntax error in ubuntu

hu-weihua:


zet809:

whowutwut:
@hu-weihua  Please add the issue number in the description, or GitHub won't make the references to the issue..  if it's only in the subject of the pull request.

hu-weihua:
@whowutwut , I have added the issue number in the description. thanks for your comment

PR #1737, opened 2016-08-18
Add security notices from 2015 over to RTD

whowutwut:
Created the security notices that were on SF for 2015 into the xCAT doc so that we can remove the broken links in the GitHub Wiki. 

After 2 years, I think we should start deleting the old ones from the master branch and only keep a running 2 years.  The 2.12 branch would have 2015, but maybe the 2.13 release should remove 2015 and only keep 2016,2017.. etc..


whowutwut:
Documentation change so I will merge.

PR #1739, opened 2016-08-18
Add the URL for the 16.04.1 mini.iso

whowutwut:
I added the mini.iso for 16.04.1 into our xcat.org download page and also updated the documentation.


whowutwut:
@immarvin @zet809  I took the file from /xcat/iso, please verify that it's the correct mini.iso file to have on xcat.org.  @cxhong  we can merge this documentation change.

zet809:
@whowutwut I have verified the md5sum of kernel file vmlinux both in mini.iso you loaded to xcat.org and 16.06.1 ppc64el ISO, they are same. So the mini.iso upload to xcat.org is correct.

PR #1749, opened 2016-08-22
use the same way to promote error msg

### LINKED ISSUE #1706: [FVT]:use the same way to process the invalid archive and compress method for packimage
xCAT 2.12.2 
This is a low priority requirement. i found when use invalid archive and compress method. The code logic is different. If the return warning message is the same, that would be great!

```
[root@c910f02c01p13 ppc64]# packimage rhels7.2-ppc64-netboot-compute -m test
Packing contents of /install/netboot/rhels7.2/ppc64/compute/rootimg
archive method:test
compress method:gzip
Error: Invalid archive method 'test' requested

[root@c910f02c01p13 ppc64]# packimage rhels7.2-ppc64-netboot-compute -c test
Packing contents of /install/netboot/rhels7.2/ppc64/compute/rootimg
Error: Invalid compress method 'test' requested 
```

penguhyang:
use the same way to process the invalid archive and compress method for packimage
fix #1706

### DIFF
--- a/xCAT-server/lib/xcat/plugins/packimage.pm
+++ b/xCAT-server/lib/xcat/plugins/packimage.pm
@@ -417,7 +417,6 @@ sub process_request {
         $callback->({ error => ["$rootimg_dir does not exist, run genimage -o $osver -p $profile on a server with matching architecture"] });
         return 1;
     }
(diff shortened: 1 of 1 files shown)


immarvin:
agree to merge

Issue #1628, opened 2016-08-03
Add disk to powered on kvm fails.


gurevichmark:
Currently adding a disk to a kvm (`chvm -a`) works for a kvm that is **not** powered on. But fails for a kvm that is already powered on.

immarvin:
hi @gurevichmark , maybe you can try  ``virsh attach-disk <domain> <source> <target>``. the <source> is the full path of the qcow2 file on the kvm host. the <target> is the disk device name for the newly added disk on the kvm guest, you should make sure the specified <target> does not exist on the kvm guest.

gurevichmark:
@immarvin Thank you for the suggestion, I tried `virsh attach-disk <domain> <source> <target>` and the results are the same as with `virsh attach-device <xml file>`. 
```
[root@fs2 ~]# virsh attach-disk mgXgeneratedvm1 /var/lib/libvirt/images/mgXgeneratedvm1.sdb.qcow2 sdb
error: Failed to attach disk
error: operation failed: target sdb already exists

[root@fs2 ~]# virsh attach-disk mgXgeneratedvm1 /var/lib/libvirt/images/mgXgeneratedvm1.sdd.qcow2 sdd
error: Failed to attach disk
error: Cannot access storage file '/var/lib/libvirt/images/mgXgeneratedvm1.sdd.qcow2' (as uid:107, gid:107): No such file or directory

[root@fs2 ~]#
```

I also tried running it with a flag `--live`, same result.

immarvin:
hi @gurevichmark , would you please run``virsh dumpxml mgXgeneratedvm1 `` on the kvm host or "fdisk -l" inside the kvm guest, is there already a disk named "sdb" in the kvm guest?

For "/var/lib/libvirt/images/mgXgeneratedvm1.sdd.qcow2", how was this qcow2 file produced?

gurevichmark:
I think the problem might be with the `target` we assign to the disk to be attached. It seems that when a vm is created target `sda` is for a disk, and `sdb` is for cdrom. However when `chvm -a <size>` is issued on this vm, we attempt to create a disk with `sdb` target, even though `sdb` is already assigned to the cdrom. 
I think the solution should be to fix `chvm -a` processing to skip the target assigned to cdrom.

gurevichmark:
The root cause seems to me is `get_multiple_paths_by_url()` in kvm.pm.
[...]

PR #1751, opened 2016-08-22
Do not reuse cdrom device label when adding new disks to KVM

gurevichmark:
This pull request fixes issue #1628 

Without this fix, cdrom is assigned **sdb** label, then, when disk is added to the vm with `chvm -a` command it also gets **sdb** label. When vm is powered off it is not a problem, but when vm is up or being powered on, an error is reported.

The solution, implemented by this pull request, is to take into account the device label used by cdrom and not reuse it when adding a disk to vm with `chvm -a` command.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -1901,8 +1901,8 @@ sub chvm {
     # in xml definition of the VM.
     # We add cdrom entry to useddisks hash to make sure the device name used by cdrom is not 
     # selected for the new disk about to be added (chvm -a)
-    my $cdrom_name = get_cdrom_device_name($vmxml);
-    if ($cdrom_name) {
+    my @cdrom_names = get_cdrom_device_names($vmxml);
+    foreach my $cdrom_name (@cdrom_names) {
         $useddisks{$cdrom_name} = 1;
     }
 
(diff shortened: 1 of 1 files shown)


daniceexi:
@immarvin Could review this pull request?

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -4093,4 +4106,24 @@ sub dohyp {
...
+    foreach my $disknode (@alldisks) {
+         my $devicetype = $disknode->getAttribute("device");
+         # Check if it is cdrom
+         if ($devicetype eq "cdrom") {
+             # Get name of the cdrom
+             $device_name = $disknode->findnodes('./target')->[0]->getAttribute('dev');
+             last; # only one cdrom device is expected
+         }
+    }
+    return $device_name;
```
immarvin: The return value should better be an array in case there  are multiple cdrom devices in the kvm guest.  This might be rare, but we need to make code robust.

immarvin:
hi @gurevichmark , the code looks good, I have one comment on the return value of the newly added subroutine, please see details in the comment in the code diff

gurevichmark:
@immarvin Second commit for this pull request implements the change to return an array of cdrom device names.

immarvin:
agree to merge
thanks @gurevichmark

PR #1754, opened 2016-08-24
update cases since -o -p and -a is  not supported anymore

tingtli:
@penguhyang please help review. Thanks!


penguhyang:
agree to merge.

PR #1756, opened 2016-08-24
increase the time since we add some time-consuming test cases

tingtli:
@hu-weihua please review and merge.

### DIFF
--- a/xCAT-server/share/xcat/tools/jenkins/xcatjktest
+++ b/xCAT-server/share/xcat/tools/jenkins/xcatjktest
@@ -1278,7 +1278,7 @@ while(! $sub_process_rt) {
     }
 
     if(time() - $regstarttime > 36000) {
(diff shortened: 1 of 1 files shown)


xCAT-server/share/xcat/tools/jenkins/xcatjktest
```diff
@@ -1277,7 +1277,7 @@ while(! $sub_process_rt) {
         }
     }
 
-    if(time() - $regstarttime > 28800) {
+    if(time() - $regstarttime > 36000) {
         send_msg(1, "[timing] 8 hours is expired");
```
hu-weihua: The message should be changed to ``10 hours``

hu-weihua:
I have reviewed the code change and agree to merge. thanks @tingtli

daniceexi:
As we discussed that we should NOT use fixed wait, but a loop check instead.

PR #1757, opened 2016-08-25
give warnings and return when use -o -p -a for packimage

### LINKED ISSUE #1755: [FVT]since -o -a and -p are not supported in packimage, please give warnings and return when use it.
We have decide not to support -a -o and -p in packimage, but when i use it, i found it could pack the image and return normally. but , when i use nodeset to set the image with the node, it will gives error message.  see following example,
```
packimage -o rhels6.8 -p compute -a ppc64 -m cpio -c gzip
Packing contents of /install/netboot/rhels6.8/ppc64/compute/rootimg
archive method:cpio
compress method:gzip

nodeset c910f02c02p28 osimage=rhels6.8-ppc64-netboot-compute
Error: stop configuration because Unable to find the os shipped yaboot file.
c910f02c02p28: netboot rhels6.8-ppc64-compute
```

I think since -a -o and -p are not supported anymore, please give warning and exit when user uses it.

penguhyang:
fix #1755 
Check options before continue.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/packimage.pm
+++ b/xCAT-server/lib/xcat/plugins/packimage.pm
@@ -108,6 +108,10 @@ sub process_request {
         "help|h"      => \$help,
         "version|v"   => \$version
     );
+    if ($arch or $osver or $profile) {
+        $callback->({ error => ["-o, -p and -a options are obsoleted, please use 'packimage <osimage name>' instead."], errorcode => [1] });
+        return 1;
+    }
     if ($version) {
         my $version = xCAT::Utils->Version();
         $callback->({ info => [$version] });


xCAT-server/lib/xcat/plugins/packimage.pm
```diff
@@ -108,6 +108,10 @@ sub process_request {
         "help|h"      => \$help,
         "version|v"   => \$version
     );
+    if ($arch or $osver or $profile) {
+        $callback->({ error => ["-o, -p and -a options has been discarded, please specify imagename instead."], errorcode => [1] });
```
immarvin: the message should be "-o, -p and -a options are obsoleted, please use 'package <osimage name>' instead."

penguhyang: fixed.

immarvin:
agree to merge

PR #1763, opened 2016-08-25
fix package return code is 0 when there are errors

### LINKED ISSUE #1761: [DEV]"packimage" return code is 0 when there are errors
````
c910f02c08p08:~ # packimage -c pigz sles11.4-ppc64-netboot-compute;echo $?
Error: Command pigz does not exist, please make sure it is installed.
0
````

There are many places in packimage.pm
````
431             my $ispigz = system("bash -c 'type -p pigz' >/dev/null 2>&1");
432             unless ($ispigz == 0) {
433                 $callback->({ error => ["Command pigz does not exist, please make sure it is installed."] });
434                 return 1;
435             }
436             $suffix = "gz";
437         } elsif ($compress eq 'xz') {
438             my $isxz = system("bash -c 'type -p xz' >/dev/null 2>&1");
439             unless ($isxz == 0) {
440                 $callback->({ error => ["Command xz does not exist, please make sure it is installed."] });
441                 return 1;
442             }
````

it should be like this, otherwise, error code can not be passed as the return code of the command:
````
444         } else {
445             $callback->({ error => ["Invalid compress method '$compress' requested"], errorcode => [1] });
446             return 1;
447         }
````

please fix all the errors in packimage.pm like this

penguhyang:
fix #1761

### DIFF
--- a/xCAT-server/lib/xcat/plugins/packimage.pm
+++ b/xCAT-server/lib/xcat/plugins/packimage.pm
@@ -418,7 +418,7 @@ sub process_request {
     my $temppath;
     my $oldmask;
     unless (-d $rootimg_dir) {
-        $callback->({ error => ["$rootimg_dir does not exist, run genimage -o $osver -p $profile on a server with matching architecture"] });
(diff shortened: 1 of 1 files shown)


immarvin:
agree to merge

Issue #1767, opened 2016-08-25
genimage deleted /dev/* on the management node...


ralphbellofatto:
we ran genimage on the following osimage stanza:

```
rhels72_compute:
    objtype=osimage
    exlist=/install/custom/rhel72_compute/compute.exlist
    imagetype=linux
    osarch=ppc64le
    osdistroname=rhels7.2-ppc64le
    osname=Linux
    osvers=rhels7.2
    otherpkgdir=/install/post/otherpkgs/rhels7.2/ppc64le
    pkgdir=/install/rhels7.2/ppc64le
    pkglist=/install/custom/rhel72_compute/compute.rhels7.pkglist
    postinstall=/install/postscripts/custom/rhel72_compute/compute.postinstall
    profile=compute
    provmethod=netboot
    rootimgdir=/install/netboot/rhels7.2/ppc64le/rhel72_compute
```

then ran genimage rhels72_compute:

right before it stopped it printed:

```
    Complete!
    umount: /sys: umount failed: No such file or directory
    rm: cannot remove '/dev/hugepages': Device or resource busy
    rm: cannot remove '/dev/mqueue': Device or resource busy   
    rm: cannot remove '/dev/pts/10': Operation not permitted   
    rm: cannot remove '/dev/pts/9': Operation not permitted    
    rm: cannot remove '/dev/pts/8': Operation not permitted    
    rm: cannot remove '/dev/pts/12': Operation not permitted   
    rm: cannot remove '/dev/pts/13': Operation not permitted   
    rm: cannot remove '/dev/pts/11': Operation not permitted   
    rm: cannot remove '/dev/pts/14': Operation not permitted   
    rm: cannot remove '/dev/pts/7': Operation not permitted    
    rm: cannot remove '/dev/pts/6': Operation not permitted    
    rm: cannot remove '/dev/pts/5': Operation not permitted    
    rm: cannot remove '/dev/pts/4': Operation not permitted    
    rm: cannot remove '/dev/pts/1': Operation not permitted    
    rm: cannot remove '/dev/pts/0': Operation not permitted    
    rm: cannot remove '/dev/pts/16': Operation not permitted   
    rm: cannot remove '/dev/pts/2': Operation not permitted    
    rm: cannot remove '/dev/pts/3': Operation not permitted    
    rm: cannot remove '/dev/pts/ptmx': Operation not permitted 
    rm: cannot remove '/dev/shm': Device or resource busy      

[...]

PR #1769, opened 2016-08-26
fix issue enimage deleted /dev/* on the management node... #1767

immarvin:
correct a typo in rh/genimage
fix issue https://github.com/xcat2/xcat-core/issues/1767


penguhyang:
agree to merge

Issue #1748, opened 2016-08-19
go-xcat does not show prompt from yum and fails on RHEL 7 x86_64 


whowutwut:
Using the latest go-xcat today for the 2.12.2 GA, i find some issues on x86_64 rhel 7

```
[root@c910f04x42vm106 ~]# /tmp/go-xcat install
Operating system:   linux
Architecture:       x86_64
Linux Distribution: rhel
Version:            7.0

Reading repositories ...... done

xCAT is going to be installed.
Continue? [y/n] y
....
....

 zip                 x86_64 3.0-10.el7         local-rhels7-x86_64-x86_64 260 k

Transaction Summary
================================================================================
Install  16 Packages (+103 Dependent packages)

Total download size: 106 M
Installed size: 352 M

```

Right here, it's trying to prompt for user response, but I do not see any prompt. 

When I hit enter, go-xcat fails... 

```
Is this ok [y/d/N]: Exiting on user command
Your transaction was saved, rerun it with:
 yum load-transaction /tmp/yum_save_tx.2016-08-19.16-09.Wobqmi.yumtx
Command "repoquery" not found.
go-xcat: Above listed required command(s) not found.
Command "repoquery" not found.
go-xcat: Above listed required command(s) not found.

xCAT Core Packages
==================

Package Name                Installed                      In Repository                 
------------                ---------                      -------------                 
Command "repoquery" not found.
go-xcat: Above listed required command(s) not found.
perl-xCAT                   (not installed)                                              
xCAT                        (not installed)                                              
xCAT-buildkit               (not installed)                                              
xCAT-client                 (not installed)                                              
xCAT-confluent              (not installed)                                              
xCAT-genesis-scripts-ppc64  (not installed)                                              
xCAT-genesis-scripts-x86_64 (not installed)                                              
xCAT-server                 (not installed)                                              
xCATsn                      (not installed)                                              
xCAT-SoftLayer              (not installed)                                              
xCAT-test                   (not installed)                                              
xCAT-vlan                   (not installed)                                              

xCAT Dependency Packages
========================

Package Name                Installed                      In Repository                 
------------                ---------                      -------------                 
Command "repoquery" not found.
go-xcat: Above listed required command(s) not found.


Boo-boo
=======

Something went wrong. :(

Please check log file `/tmp/go-xcat.log' for more details.
```

whowutwut:
[...]

Issue #1641, opened 2016-08-05
go-xcat update will continue even if required repoquery command is not installed


whowutwut:
Using `go-xcat` to update a development version of xCAT... 
```
# /tmp/go-xcat --xcat-version=devel update
....
....
Complete!
Command "repoquery" not found.  <=======================
go-xcat: Above listed required command(s) not found.
Command "repoquery" not found. <=======================
go-xcat: Above listed required command(s) not found.

xCAT Core Packages
==================

Package Name                Installed                      In Repository                 
------------                ---------                      -------------                 
Command "repoquery" not found.    <=======================
go-xcat: Above listed required command(s) not found.
perl-xCAT                   2.12.2-snap201608040616                                      
xCAT                        2.12.2-snap201608040616                                      
xCAT-buildkit               2.12.2-snap201608040616                                      
xCAT-client                 2.12.2-snap201608040616                                      
xCAT-confluent              (not installed)                                              
xCAT-genesis-scripts-ppc64  2.12.2-snap201608040616                                      
xCAT-genesis-scripts-x86_64 (not installed)                                              
xCAT-server                 2.12.2-snap201608040616                                      
xCATsn                      (not installed)                                              
xCAT-SoftLayer              (not installed)                                              
xCAT-test                   (not installed)                                              
xCAT-vlan                   (not installed)            
...
```

we see messages that repoquery is not found.  

I think we should pre-check for required commands, or we do not run go-xcat.  We don't want the user to see errors when running the go-xcat.

tingtli:
update is a second version for the go-xcat, we need to add cases to cover this option

neo954:
It worked as design.

The command ``repoquery`` is provided by the RPM package ``yum-utils`` on RHEL. And package ``yum-utils`` will be installed even if you choose "minimal install" during the operating system installation. Thus, there is no way to get a living RHEL system without the command ``repoquery``.

[...]

PR #1775, opened 2016-08-26
[go-xcat] Fix github issue #1641, issue #1748

neo954:
[go-xcat] Fix github issue #1641, issue #1748

### DIFF
--- a/xCAT-server/share/xcat/tools/go-xcat
+++ b/xCAT-server/share/xcat/tools/go-xcat
@@ -2,7 +2,7 @@
 #
 # go-xcat - Install xCAT automatically.
 #
-# Version 1.0.2
+# Version 1.0.3
 #
 # Copyright (C) 2016 International Business Machines
 # Eclipse Public License, Version 1.0 (EPL-1.0)
@@ -483,7 +483,8 @@ function check_repo_version_dnf()
 function check_repo_version_yum()
 {
 	type yum >/dev/null 2>&1 || return 255
-	check_exec_or_exit repoquery
+	check_executes repoquery
+	exit_if_bad "$?" "Install the 'yum-utils' package and rerun."
 	local -a name=()
 	local -a ver=()
 	while read -r name ver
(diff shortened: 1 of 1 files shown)


whowutwut:
@neo954  This one resolves Issue #1748 but does not resolve #1641 .   It still continues if `repoquery` command is not found.  I'm not sure why in the code it doesn't exit the script...  because in a test program, it works...   I tested the version of `go-xcat` in this pull request and the behavior is not correct...

```
[root@fs2vm111 ~]# ./go-xcat --xcat-version=devel update
Operating system:   linux
Architecture:       ppc64le
Linux Distribution: rhel
Version:            7.3

Reading repositories ...... done

xCAT is going to be updated.
Continue? [y/n] y
Command "repoquery" not found.
go-xcat: Install the 'yum-utils' package and rerun.  <---- SHOULD EXIT HERE, but continues...
Package conserver-xcat-8.1.16-10.ppc64le already installed and latest version
Package elilo-xcat-3.14-4.noarch already installed and latest version
Package grub2-xcat-2.02-0.16.el7.snap201506090204.noarch already installed and latest version
Package ipmitool-xcat-1.8.15-3.ppc64le already installed and latest version
Package syslinux-xcat-3.86-2.noarch already installed and latest version
```

In my test program:
```
check_executes repoquery
exit_if_bad "$?" "hello"

echo "Going on!"
```

Running this: 
```
[root@fs2vm111 ~]# ./test.sh 
Command "repoquery" not found.
1
test.sh: hello
```

We exit before seeing the echo "Going on!"

neo954:
Thanks for the comment, @whowutwut. Even if `repoquery` is absent, I intend to let the `go-xcat update` run.

neo954:
The `repoquery' is run in a subshell. Another patch is submitted. I finally find a way to handle the error properly.

neo954:
@whowutwut, please help on merge the patches.

whowutwut:
@gurevichmark  can you review this and merge if ok?

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -1600,6 +1594,7 @@ case "${GO_XCAT_ACTION}" in
 		xCAT has been updated!
 		======================
 
+		Congratulations!
```
gurevichmark: I would suggest not using "Congratulations!",
perhaps better to just change the line above from
"xCAT has been updated!" to "xCAT has been successfully updated."

neo954: That is thoughtful. I will change the code.

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -1603,10 +1603,7 @@ case "${GO_XCAT_ACTION}" in
 	"update")
 		while read -r ; do echo "${REPLY}" ; done <<-EOF
 
-		xCAT has been updated!
-		======================
-
-		Congratulations!
+		xCAT has been successfully updated!
```
daniceexi: I am trying the new comment approaches.

daniceexi: so this comment will be a reply to my first comment

Issue #1634, opened 2016-08-04
[FVT]restore option is listed in "xcattest -h" but no man page description


caomengmeng:
Description: restore option is listed in "xcattest -h" but no man page description, unify the two.

```
[root@c910f03 bundle]# xcattest -?
Usage:xcattest - Run xcat test cases.
  xcattest [-?|-h]
  xcattest [-f configure file] [-b case bundle list] [--restore]
  xcattest [-f configure file] [-t case list] [--restore]
  xcattest [-f configure file] [-c cmd list] [-l] [--restore]

[root@c910f03 bundle]# xcattest -h
Usage:xcattest - Run xcat test cases.
  xcattest [-?|-h]
  xcattest [-f configure file] [-b case bundle list] [--restore]
  xcattest [-f configure file] [-t case list] [--restore]
  xcattest [-f configure file] [-c cmd list] [-l] [--restore]
```

whowutwut:
@caomengmeng  please assign this to a milestone.  If it's just a documentation change, maybe we can finish and close it out in 2.12.2

caomengmeng:
No need to open this option, so just delete the related code in usage, and will fixed by: https://github.com/xcat2/xcat-core/pull/1833

whowutwut:
I removed the pending status because the pull request that will address this issue is not yet merged.

caomengmeng:
As this restore function is not opened to users, so just delete in usage, this will make the usage and man page unified.
```
[root@c910f02c01p13 lsdef]# xcattest -h
Usage:Run xcat test cases:
  xcattest [-?|-h]
  xcattest [-f configure file] [-b case bundle files]
  xcattest [-f configure file] [-t cases list]
  xcattest [-f configure file] [-c cmds list]
Show xcat test cases, bundle files, commands lists:
  xcattest [-f configure file] [-b case bundle files] [-l]
  xcattest [-f configure file] [-t cases list] [-l]
  xcattest [-f configure file] [-c cmds list] [-l]
  xcattest [-s command|bundle]
```

so close it.

PR #1776, opened 2016-08-26
Update xcattest and complete install xCAT case

caomengmeng:
In order to divide the install xCAT from the xcatjktest auto-test framework, just change the xcattest file and complete the install xCAT case.
Meanwhile, improve the xcattest command:
1. optional define
2. optional input error messages: config/bundle file error and bundle file list show.
3. delete the case files output when unexpected.
 
@tingtli @hu-weihua

### DIFF
--- a/xCAT-test/autotest/testcase/install_xCAT/case0
+++ b/xCAT-test/autotest/testcase/install_xCAT/case0
@@ -1,5 +1,5 @@
-start:install_xCAT_new
-description:install xCAT with go-xcat tool
+start:install_xCAT_on_rhels_sles
+description:install xCAT with go-xcat tool in a fresh environment for rhels and sles
 os:Linux
 cmd:if grep "Red Hat" /etc/*release >/dev/null; then yum install -y yum-utils bzip2; fi
 check:rc==0
@@ -19,3 +19,27 @@ check:rc==0
 check:output=~running
 cmd:rm -rf /xcat-core.tar.bz2 /xcat-dep.tar.bz2
 end
+
+start:install_xCAT_on_ubuntu
+description:install xCAT with go-xcat tool in a fresh environment for ubuntu
+os:Linux
+cmd:arc_all=`uname -a`; code=`lsb_release -sc`;if [[ $arc_all =~ "ppc64le" ]]; then arch="ppc64el";else arch="x86_64";fi; cp "/opt/xcat/share/xcat/tools/autotest/testcase/go-xcat/$code-$arch.sources.list" "/etc/apt/sources.list" 
+cmd:apt-get clean;apt-get update
+check:rc==0
+cmd:cp /core-*-snap.tar.bz2 /xcat-core.tar.bz2
+check:rc==0
+cmd:cp /xcat-dep*.tar.bz2 /xcat-dep.tar.bz2
+check:rc==0
+cmd:./go-xcat --xcat-core=/xcat-core.tar.bz2 --xcat-dep=/xcat-dep.tar.bz2  -y install;
+check:rc==0
+cmd:source "/etc/profile.d/xcat.sh"
+check:rc==0
+cmd:lsxcatd -v
+check:rc==0
+check:output=~Version
+cmd:service xcatd status
+check:rc==0
+check:output=~running
+cmd:rm -rf /xcat-core.tar.bz2 /xcat-dep.tar.bz2
+end
+
--- a/xCAT-test/autotest/testcase/install_xCAT/case0
+++ b/xCAT-test/autotest/testcase/install_xCAT/case0
@@ -1,0 +1,21 @@
+start:install_xCAT_new
+description:install xCAT with go-xcat tool
+os:Linux
+cmd:if grep "Red Hat" /etc/*release >/dev/null; then yum install -y yum-utils bzip2; fi
+check:rc==0
+cmd:cp /core-*-snap.tar.bz2 /xcat-core.tar.bz2
+check:rc==0
+cmd:cp /xcat-dep*.tar.bz2 /xcat-dep.tar.bz2
+check:rc==0
+cmd:cd /; ./go-xcat --xcat-core=/xcat-core.tar.bz2 --xcat-dep=/xcat-dep.tar.bz2  -y install;
+check:rc==0
+cmd:source "/etc/profile.d/xcat.sh"
+check:rc==0
+cmd:lsxcatd -v
+check:rc==0
+check:output=~Version
+cmd:service xcatd status
+check:rc==0
+check:output=~running
+cmd:rm -rf /xcat-core.tar.bz2 /xcat-dep.tar.bz2
+end
--- a/xCAT-test/xcattest
+++ b/xCAT-test/xcattest
@@ -14,7 +14,7 @@ BEGIN
 use lib "$::XCATROOT/lib/perl";
 my $rootdir     = "$::XCATROOT/share/xcat/tools/autotest";
 my $needhelp    = 0;
-my $configfile  = "$rootdir/default.conf";
+my $config_info  = "$rootdir/default.conf";
 my $bundle_list = undef;
 my $case_list   = undef;
 my $cmd_list    = undef;
@@ -24,7 +24,7 @@ my $ret         = 0;
 my $string1     = undef;
 if (
     !GetOptions("h|?" => \$needhelp,
-        "f=s"     => \$configfile,
+        "f=s"     => \$config_info,
         "b=s"     => \$bundle_list,
         "t=s"     => \$case_list,
         "c=s"     => \$cmd_list,
@@ -36,10 +36,46 @@ if (
     exit 1;
 }
 
+unless(defined($bundle_list) || defined($case_list) || defined($cmd_list) || defined($needshow)){
+    &usage;
+    exit 1;
+}
 if ($needhelp)
 {
     &usage;
     exit 0;
+}
+
+#check config file and system label
+my $configfile  = undef;
+my $inital_info = undef;
+if ($config_info){   
+    if ($config_info =~ /:/){
+        my @config_info = split /:/, $config_info;
+        $configfile = $config_info[0];
+        $inital_info = $config_info[1];
+        if ($inital_info ne "system"){
+            print "Failed: $inital_info in $config_info is not supported!\n";
+            exit 1;
+        }
+    }else{
+        $configfile = $config_info;
+    }
+}
+if (!open(FILE, "$configfile")) {
+    print "can't open config file: $configfile\n";
+    exit 1;
+}
+
+#check bundle files
+if ($bundle_list){
+     my @bundles = split /,/, $bundle_list;
+     foreach my $bundle (@bundles){
+        if (!open(FILE, "<$rootdir/bundle/$bundle")) {
+            print "can't open bundle file: $rootdir/bundle/$bundle\n";
+            exit 1;
+        }
+    }
 }
 
 #load case to $cases
@@ -51,7 +87,6 @@ if ($needhelp)
 #    hcp:hmc/mm/bmc/fsp         string
 #    cmd:                       array
 #    check:                     array
-
 my @cases = ();
 if ($needshow) {
     &loadcase;
@@ -150,10 +185,6 @@ sub getConfig
     log_this("******************************");
     log_this("Reading Configure");
     log_this("******************************");
-    if (!(-e $configfile)) {
-        log_this("Warning: The xCAT test Configure file doesn't exist!");
-        return 0;
-    }
 
     my $type     = undef;    #Script_Prev,Script_Post,Table,Object,System,Custom
     my $sub_type = undef;    # The string after $type_
@@ -175,7 +206,6 @@ sub getConfig
 
     my $mgt_name = undef;
 
-    open(FILE, "$configfile") or die "can't to open $configfile";
     while (my $line = <FILE>) {
         $line = &trim($line);
         next if (length($line) == 0);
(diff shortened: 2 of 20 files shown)


xCAT-test/xcattest
```diff
@@ -36,12 +38,52 @@ if (
...
+     my @bundles = split /,/, $bundle_list;
+     foreach my $bundle (@bundles){
+        if (!open(FILE, "< $bundle_dir/$bundle")) {
+            print "Can't open bundle file: $bundle_dir/$bundle\n";
+            &show_bundle_files;
+            exit 1;
+        }
+    }
+}
+
```
hu-weihua: I don't think we can add code in this area, this area is out of control from log function, that means if we run ``xcattest`` in the background, we will get nothing in the log and don't know what happened.

xCAT-test/xcattest
```diff
@@ -36,12 +38,52 @@ if (
...
+            exit 1;
+        }
+    }else{
+        $configfile = $config_info;
+    }
+}
+if (!open(FILE, "$configfile")) {
+    print "Can't open config file: $configfile\n";
+    exit 1;
+}
```
hu-weihua: [Comment for line 57-73]Our goal is change the way we load config file, I think this change should be done in ``getConfig`` function. Coding like this will broken the structure of while script, it isn't a better option.

xCAT-test/autotest/testcase/install_xCAT/case0
```diff
@@ -0,0 +1,21 @@
+start:install_xCAT_new
+description:install xCAT with go-xcat tool
+os:Linux
+cmd:if grep "Red Hat" /etc/*release >/dev/null; then yum install -y yum-utils bzip2; fi
+check:rc==0
+cmd:cp /core-*-snap.tar.bz2 /xcat-core.tar.bz2
+check:rc==0
+cmd:cp /xcat-dep*.tar.bz2 /xcat-dep.tar.bz2
+check:rc==0
+cmd:cd /; ./go-xcat --xcat-core=/xcat-core.tar.bz2 --xcat-dep=/xcat-dep.tar.bz2  -y install;
```
tingtli: so the xcatjktest should copy go-xcat to mn's / directory

xCAT-test/xcattest
```diff
@@ -36,12 +38,52 @@ if (
...
+
+#check bundle files
+if ($bundle_list){
+     my @bundles = split /,/, $bundle_list;
+     foreach my $bundle (@bundles){
+        if (!open(FILE, "< $bundle_dir/$bundle")) {
+            print "Can't open bundle file: $bundle_dir/$bundle\n";
+            &show_bundle_files;
+            exit 1;
+        }
```
hu-weihua: If all bundle file are valid, we will get a lot of opening file descriptors and don't use any more.  Further more, this area is global area, the ``FILE`` file descriptor will valid in whole script, it's  very dangerous and maybe trigger problem unexpected.

xCAT-test/autotest/testcase/install_xCAT/case0
```diff
@@ -0,0 +1,21 @@
...
+check:rc==0
+cmd:source "/etc/profile.d/xcat.sh"
+check:rc==0
+cmd:lsxcatd -v
+check:rc==0
+check:output=~Version
+cmd:service xcatd status
+check:rc==0
+check:output=~running
```
tingtli: I don't think all the system will return running.
see some output on sles12.1

xCAT-test/xcattest
hu-weihua: @caomengmeng , Could you please pay more attention to  logger component in ``xcattest`` script, it is very important.

xCAT-test/xcattest
```diff
@@ -945,10 +972,19 @@ sub usage
...
     print "  xcattest [-?|-h]\n";
-    print "  xcattest [-f configure file] [-b case bundle list] [--restore]\n";
-    print "  xcattest [-f configure file] [-t case list] [--restore]\n";
-    print "  xcattest [-f configure file] [-c cmd list] [-l] [--restore]\n";
+    print "  xcattest [-f configure file] [-b case bundle list]\n";
+    print "  xcattest [-f configure file] [-t case list]\n";
+    print "  xcattest [-f configure file] [-c cmd list]\n";
+    print "  xcattest [-f configure file] [-b case bundle list] [-l]\n";
+    print "  xcattest [-f configure file] [-t case list] [-l]\n";
+    print "  xcattest [-f configure file] [-c cmd list] [-l]\n";
```
tingtli: Please add the configfile:System in the usage.

xCAT-test/xcattest
```diff
@@ -36,12 +38,52 @@ if (
...
 
+#check config file and System label
+#this System label means only the [System] variable will be loaded
+my $configfile  = undef;
+my $inital_info = undef;
+if ($config_info){   
+    if ($config_info =~ /:/){
+        my @config_info = split /:/, $config_info;
+        $configfile = $config_info[0];
+        $inital_info = $config_info[1];
```
hu-weihua: Line 58-61 could be refined as

    if ($config_info =~ /(\w+):(\w+)/){
          $configfile =$1;
          $inital_info = $2;

xCAT-test/xcattest
```diff
@@ -84,9 +125,11 @@ if ($ret != 0) {
     goto EXIT;
 }
 
-$ret = &init;
-if ($ret != 0) {
-    goto EXIT;
+if ($inital_info ne "System"){
```
hu-weihua: Maybe we can let ``inital_info`` as a output attribute of  ``getConfig`` function. Let all config file parse are handled by ``getConfig`` function.

xCAT-test/xcattest
```diff
@@ -150,10 +193,6 @@ sub getConfig
     log_this("******************************");
     log_this("Reading Configure");
     log_this("******************************");
-    if (!(-e $configfile)) {
-        log_this("Warning: The xCAT test Configure file doesn't exist!");
-        return 0;
-    }
```
hu-weihua: Why we delete line 153-156?

xCAT-test/xcattest
```diff
@@ -172,11 +211,28 @@ sub getConfig
     my $value = undef;
     my $c     = 0;
     my $cmd   = undef;
-
     my $mgt_name = undef;
+    
+    open(FILE, "$configfile") or die "Can't open $configfile";    
+    #Only load System information
```
hu-weihua: It's not good option to use  ``die`` in a function.  @caomengmeng , could you please refine this code?

xCAT-test/xcattest
```diff
@@ -501,10 +558,7 @@ sub loadcase
     if ($bundle_list) {
         my @bundles = split /,/, $bundle_list;
         foreach my $bundle (@bundles) {
-            if (!open(FILE, "<$rootdir/bundle/$bundle")) {
-                log_this("can't open $rootdir/bundle/$bundle");
-                return 1;
-            }
+            open(FILE, "<$bundle_dir/$bundle") or die "Can't open bundle file: $bundle";
```
hu-weihua: Just as above comment, ``die`` is not a good option in a function, I think original code logic is better.

xCAT-test/xcattest
```diff
@@ -953,10 +1017,7 @@ sub reordercases {
...
```
hu-weihua: Just as above comment, die is not a good option in a function, I think original code logic is better.

xCAT-test/xcattest
```diff
@@ -919,13 +973,23 @@ sub usage
 {
     print "Usage:xcattest - Run xcat test cases.\n";
     print "  xcattest [-?|-h]\n";
-    print "  xcattest [-f configure file] [-b case bundle list] [--restore]\n";
-    print "  xcattest [-f configure file] [-t case list] [--restore]\n";
-    print "  xcattest [-f configure file] [-c cmd list] [-l] [--restore]\n";
+    print "  xcattest [-f configure file] [-b case bundle list]\n";
+    print "  xcattest [-f configure file] [-t case list]\n";
+    print "  xcattest [-f configure file] [-c cmd list]\n";
+    print "  xcattest [-f configure file] [-b case bundle list] [-l]\n";
```
hu-weihua: After going through the whole code, I couldn't find where we implement this feature.

tingtli:
@caomengmeng  A general comment is we need to log the errors and don't use much print in the code since we don't monitor the execution of the program. Please see loggers in the sub functions

xCAT-test/xcattest
```diff
@@ -919,13 +973,23 @@ sub usage
...
+    print "  xcattest [-f configure file] [-t case list] [-l]\n";
+    print "  xcattest [-f configure file] [-c cmd list] [-l]\n";
```
hu-weihua: After going through the code, I think option ``c`` didn't work any more, it isn't the problem triggered by this pull request, but it is a problem now.

xCAT-test/xcattest
```diff
@@ -36,12 +38,52 @@ if (
...
+#check bundle files
+if ($bundle_list){
+     my @bundles = split /,/, $bundle_list;
+     foreach my $bundle (@bundles){
+        if (!open(FILE, "< $bundle_dir/$bundle")) {
+            print "Can't open bundle file: $bundle_dir/$bundle\n";
+            &show_bundle_files;
+            exit 1;
+        }
+    }
```
hu-weihua: Depending on the conclusion in the meeting, I think we need a way to show all valid bundle file. It dosen't means we only show bundle file list when customer input wrong bundle file. we need a command option to let customer to obtain these information

xCAT-test/autotest/testcase/install_xCAT/case0
hu-weihua: I have same question with @tingtli, where we get ``go-xcat`` script

xCAT-test/autotest/testcase/install_xCAT/case0
```diff
@@ -0,0 +1,24 @@
...
+description:install xCAT with go-xcat tool in a fresh environment
+os:Linux
+cmd:if grep "Red Hat" /etc/*release >/dev/null; then yum install -y yum-utils bzip2; fi
+check:rc==0
+cmd:if grep Ubuntu /etc/*release; then arc_all=`uname -a`; code=`lsb_release -sc`;if [[ $arc_all =~ "ppc64le" ]]; then arch="ppc64el";else arch="x86_64";fi; cp "/opt/xcat/share/xcat/tools/autotest/testcase/go-xcat/$code-$arch.sources.list" "/etc/apt/sources.list"; fi
+cmd:if grep Ubuntu /etc/*release;then apt-get clean && apt-get update; fi
+check:rc==0
+cmd:cp /core-*-snap.tar.bz2 /xcat-core.tar.bz2
+check:rc==0
+cmd:cp /xcat-dep*.tar.bz2 /xcat-dep.tar.bz2
```
hu-weihua: If we mount the disk which save all xcat build into MN by ``xcatjktest``, maybe we should change the way we obtain ``xcat-core`` and ``xcat-dep``

xCAT-test/xcattest
hu-weihua: It seams we lost ``[--restore]`` option, it's a feature in old code

xCAT-test/autotest/testcase/install_xCAT/case0
```diff
@@ -0,0 +1,24 @@
+start:install_xCAT
+description:install xCAT with go-xcat tool in a fresh environment
+os:Linux
```
hu-weihua: It's better for us to let ``install_xCAT`` as a stop case.  ``xcattest`` always return "0", not matter if there is failed case, except there is stop case. If we failed to install xcat, ``xcatjktest`` should not run other cases any more.  it's the only way to let ``xcatjktest`` to know ``install_xCAT`` failed

caomengmeng: Will add a stop label

xCAT-test/autotest/testcase/install_xCAT/case0
caomengmeng: @tingtli  @hu-weihua , I just forget who, but I discussed this before with one of you,  what we agreed:  xcatjktest will copy the go-xcat to the "/" and we will use a stable go-xcat version.
Does anyone remember this?
If not, how do you suggest we deal with the go-xcat thing?

xCAT-test/xcattest
caomengmeng: @hu-weihua  That is what we discussed in daily meeting, some sub functions use the lsdef command in the init method. That is the way to avoid it. For another thing, if the init should be run even only load file, it is strange.

xCAT-test/xcattest
caomengmeng: @hu-weihua  will add a close here

xCAT-test/xcattest
caomengmeng: Only when user input the wrong file name, this will works.
I don't think we need a special command.

xCAT-test/xcattest
caomengmeng: @hu-weihua  -c works well, it run many cases directly and I tried it before.
--restore I am not sure it is an open function to users or not.
And I already submit an issue to track it before: 
https://github.com/xcat2/xcat-core/issues/1634

xCAT-test/xcattest
caomengmeng: @hu-weihua Because it is duplicated after this change.

caomengmeng:
Wrong action with rebase, so just close it and open a new one

gurevichmark:
Should not directly modify xcattest.1.rst file. Instead xcattest.1.pod file should be modified. The .rst files get generated from .pod files.

PR #1777, opened 2016-08-26
Fix the issue with bmcdiscover on older FSP and Tuletta machines where bmcuser is not needed 

whowutwut:
Seems like for discovery of FSP and TUL servers, there is no bmcuser name so adding the code from commit 3f54b456d2f2479489386b9291fa1867dd22c32d and d82e14b56656482279d58e38295a9900e82ba056 to set the default user breaks that function.  

Reverting back to the old ways

Discussion by @zet809  in Pull Request #1066 

For OpenPower machines, the behavior now looks like this: 
```
[root@c910f02c05p03 ~]# bmcdiscover --range 50.5.33.10 -p passw0rd
Error: BMC username is incorrect for 50.5.33.10
[root@c910f02c05p03 ~]# bmcdiscover --range 50.5.33.10 -p passw0rd -u ADMIN
50.5.33.10,8335-GTA,1210532,ADMIN,passw0rd
```

Also blocked the use of `-i` and `--range`
```
[root@c910f02c05p03 ~]#  bmcdiscover --range 50.5.33.10 -u ADMIN -p passw0rd --check
Error: The 'check' and 'range' option cannot be used together.
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -204,24 +204,19 @@ sub bmcdiscovery_processargs {
         return 1;
     }
 
-    #
-    # Get the default bmc account from passwd table
-    #
-    ($bmc_user, $bmc_pass) = bmcaccount_from_passwd();
-
-    # overwrite the default user/pass with what is passed in
-    if ($::opt_U) {
-        $bmc_user = $::opt_U;
-    }
-    if ($::opt_P) {
-        $bmc_pass = $::opt_P;
-    }
-
     #########################################
     # Option -s -r should be together
     ######################################
     if (defined($::opt_R))
     {
+        # Option -c should not be used with -r 
+        if (defined($::opt_C)) {
+            my $msg = "The 'check' and 'range' option cannot be used together.";
+            my $rsp = {};
+            push @{ $rsp->{data} }, "$msg";
+            xCAT::MsgUtils->message("E", $rsp, $::CALLBACK);
+            return 2;
+        }
         ######################################
         # check if there is nmap or not
         ######################################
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -240,6 +235,22 @@ sub bmcdiscovery_processargs {
...
+        ($bmc_user, $bmc_pass) = bmcaccount_from_passwd();
+        # overwrite the default password if one is provided
+        if ($::opt_P) {
+            $bmc_pass = $::opt_P;
+            if (!$::opt_U) {
+                $bmc_user = '';
+            } else { 
+                $bmc_user = $::opt_U;
+            }
+        }
```
zet809: I think the code below can overwrite bmcusername easily. Then you don't need to add notification in the manpage.
if ($::opt_U) {
    $bmc_user = $::opt_U;
}
if ($::opt_P) {
    $bmc_pass = $::opt_P;
    if (!$::opt_U) {
        $bmc_user = '';
    }
}

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -240,6 +235,24 @@ sub bmcdiscovery_processargs {
...
+        ($bmc_user, $bmc_pass) = bmcaccount_from_passwd();
+        # overwrite the default password if one is provided
+        if ($::opt_U) {
+            $bmc_user = $::opt_U;
+        } else {
+            # If password is provided, but no user, set the user to blank
+            # Support older FSP and Tuletta machines
+            $bmc_user = '';
+        }
+        if ($::opt_P) {
```
whowutwut: @zet809  I undid the man page change and made this change.. 
However, even with this. do you think we should make a note in the man page.  

For example, if we have 200 nodes Open Power, 180 of them using the default user and password set in the ipmi table.  This will work.  But for the 20 nodes , the password is differnt, but the user is the same,  Now we have to provide both -u and -p options to reset the password.  but I guess there's nothing we can do about this.

zet809: @whowutwut You are right that user have to specify -u even username is same as default one.
As it is very hard to judge there is no username supported or use the default username, I think we can add a keyword like "*null*" to indicate username is empty when running ipmitool. And need to update it into manpage and document. Sorry for having you to redo this.

zet809:
@whowutwut Hi, Victor, I think we can merge the modification first and if there is more concern, we can fix them then.

PR #1779, opened 2016-08-30
Add security advisory for SWEET32

whowutwut:
Add security information for the Sweet32 vulnerability


whowutwut:
Merged documentation changes for security advisory.

Issue #1645, opened 2016-08-08
[FVT]:rinv does not have manpage for openpower


tingtli:
I didn't find the openpower section in the manpage of rinv. Please update that. Thanks!

whowutwut:
@tingtli  @neo954  Are there specific things that need to be addressed for OpenPower servers that is different on `rinv`?

tingtli:
For the manpage of rinv, we have several sections for each hardware platform. e.g
BMC/MPA specific:,  PPC (with HMC) specific: etc.  
I think we need to add a section for "OpenPOWER BMC specific:" to let the users know which options would be used for rinv on OpenPOWER.
See manpage of rflash.

whowutwut:
@zet809 I think we currently do not support any options to rinv for OpenPower..

whowutwut:
Using 10/14 build, man pages are there for Open Power
```
RINV.1(1)                                                    User Contributed Perl Documentation                                                   RINV.1(1)

Name
       rinv - Remote hardware inventory

...
   OpenPOWER server specific:
       rinv noderange {model|serial|deviceid|uuid|guid|vpd|mprom|firm|all}
...
...
```

tingtli:
@zet809 @whowutwut  Same problem with rvitals, do you need to add manpage for OpernPOWER

whowutwut:
@tingtli yea but we should open another issue to track that. This one is for rinv and didn't mention rvitals initially. And this issue is tracked in 2.12.3 and we cannot fix in that release

tingtli:
@whowutwut Ok, i will open another one to trace this #1996

PR #1784, opened 2016-08-30
support options model|serial|deviceid|uuid|guid|vpd|mprom|firm for ri…

zet809:
…nv for OpenPOWER servers

For issue #1645

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -1574,10 +1574,13 @@ sub isfpc {
     return 1
 }
 
-sub isfirestone {
+sub isopenpower {
     my $sessdata = shift;
-    if ($sessdata->{prod_id} == 43707) {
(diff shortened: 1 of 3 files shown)


xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1574,11 +1574,14 @@ sub isfpc {
     return 1
 }
 
-sub isfirestone {
+sub isopenpower {
     my $sessdata = shift;
-    if ($sessdata->{prod_id} == 43707) {
+    if ($sessdata->{prod_id} == 43707 and $sessdata->{mfg_id} == 0) {
```
whowutwut: Did we check that Garrison machines are the same prod_id and mgf_id?

zet809: Yes, as well as habanero. There all have the same product ID(43707) and Manufacturer ID(0)

whowutwut:
I have reviewed and will merge. Thanks @zet809

Issue #1591, opened 2016-07-26
Genesis loading hang in firestone node if ntp time offset is too large


DanielXLee:
Sometimes, genesis loading hang at the script `/usr/bin/doxcat` in Firestone node, and we found the progress stayed at the below while statement for a long time due the large `ntp` offset in node:

----------------------------

```
ntpd -g -x
# rv 0 state does not work with the new ntp versions
while [ "`ntpq -c \"rv 0 offset\" | grep \"offset=\" | awk -F \"=\" '{print $2}' | awk -F \".\" '{print $1}' | sed s/-//`" -ge 1000 ]; do
    sleep 1
done
```
--------------------
```
[xCAT Genesis running on io155 /]# ntpq -c "rv 0 offset"
offset=1604.980
```
--------------------------

In fact, the unit of value of offset is milliseconds, and 1000 ms is too little to identify ntp service. By default, the threshold is 1000s(1000000 ms) for ntpd exiting with a message to the system log if the offset exceeds the panic threshold
So suggest to change the limitation of offset in script `/usr/bin/doxcat`

neo954:
More then 1000ms time offset in a NTP network environment is quiet abnormal. In regular gigabit ethernet network environment, the typical time offset between NTP server and NTP client is less than 5ms. Even in wide area network environment, the offset is less than 50ms.

wjianf:
In IBM OpenPower Firestone node provisioning, the value of offset is in 50 to 1000 ms when xCAT genesis loading in most time, but sometimes, the value is greater than 1000 ms, and the adjustment is very slow. So the provisioning process sames like hang at the scripts doxcat.

zet809:
Move to next release to reconsider this issue.

neo954:
This is the real problem. Cannot access the hardware clock. :(

    [xCAT Genesis running on c910f03c25 /]# hwclock
    hwclock: Cannot access the Hardware Clock via any known method.
    hwclock: Use the --debug option to see the details of our search for an access method.

neo954:
This is the correct shape on another Firestone machine runs RHEL 7.3 Beta. It seems we really need the kernel module `rtc_opal` in genesis for Firestone machine.

    [root@c910f05c35 ~]# hwclock
    Wed 24 Aug 2016 03:41:42 AM EDT  -0.693350 seconds
    [root@c910f05c35 ~]# ls -l /dev/rtc*
[...]

PR #1789, opened 2016-08-31
Fix github issue #1591

neo954:
This patch partly fixed issue #1591.

### DIFF
--- a/xCAT-genesis-scripts/bin/doxcat
+++ b/xCAT-genesis-scripts/bin/doxcat
@@ -274,14 +274,17 @@ logger -s -t $log_label -p local4.info "Acquired IPv4 address on $bootnic"
 
 ip addr show dev $bootnic|grep -v 'scope link'|grep -v 'dynamic'|grep -v  inet6|grep inet|awk '{print $2}'
 ntpd -g -x
+
+if [ -e "/dev/rtc" ]; then
+    ( sleep 8 ; hwclock --systohc ) </dev/null >/dev/null 2>&1 &
+    disown
+fi
+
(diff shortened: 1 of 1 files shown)


zet809:

xCAT-genesis-scripts/bin/doxcat
```diff
@@ -274,14 +274,17 @@ logger -s -t $log_label -p local4.info "Acquired IPv4 address on $bootnic"
 
 ip addr show dev $bootnic|grep -v 'scope link'|grep -v 'dynamic'|grep -v  inet6|grep inet|awk '{print $2}'
 ntpd -g -x
+
+if [ -e "/dev/rtc" ]; then
+    ( sleep 8 ; hwclock --systohc ) </dev/null >/dev/null 2>&1 &
```
whowutwut: Why is this sleep 8 needed?

neo954: @whowutwut,

That is due to `ntpd` is run as a daemon. If we run `ntpd -g -q -x` here. We can safely wait the `ntpd` exit, and do the `hwclock` things after that. Whatever `ntpd -g -q -x` did a time sync successful or not.

But as the original design, `ntpd` here is run as a daemon. In this case, the command `ntpd -g -x` will exit immediately after it become a daemon. After that, it will then do the configuration parsing, initial time sync. It will usually take around one or two seconds for the initial time sync. To make it safe, I made the sleep time eight seconds. And I do not want to waste time on sleep here, so I made the `hwclock` things run as a child process and running in the background.

Issue #1785, opened 2016-08-30
Lack of Linux kernel module rtc-opal.ko in genesis-base ppc64


neo954:
In the genesis environment running on Firestone machine. The hardware clock is not accessable. Due to the lack of Linux kernel module `rtc-opal.ko`.

```
[xCAT Genesis running on c910f03c25 /]# hwclock
hwclock: Cannot access the Hardware Clock via any known method.
hwclock: Use the --debug option to see the details of our search for an access method.
```

neo954:
Issue #1591 is depend on this issue.

neo954:
I will reopen this one. Since genesis need to be rebuilt.

zet809:
The xcat-genesis-base-ppc64 had been uploaded to GSA.

zet809:
@neo954 Hi, gongjie, pls help to verify.

whowutwut:
@neo954  Can you please verify this and close out since we are close to the 2.13 release...

neo954:
Verified with xCAT-genesis-base-ppc64-2.12-snap201610250931.noarch. The problem was fixed. I will close this issue.

PR #1790, opened 2016-08-31
add hardware clock driver when building genesis

zet809:
For issue #1785


neo954:
Looks good. Will merge this one.

Issue #1783, opened 2016-08-30
[Feature request]non-interactive option for "/opt/xcat/share/xcat/scripts/setup-local-client.sh"


samveen:
When the command  `/opt/xcat/share/xcat/scripts/setup-local-client.sh` is run, the `openssl req` command prompts for `[y/n]` twice:
```
[root@xcatmn1 ~]# /opt/xcat/share/xcat/scripts/setup-local-client.sh samveen
...
Certificate is to be certified until Aug 25 05:36:37 2036 GMT (7300 days)
Sign the certificate? [y/n]:y


1 out of 1 certificate requests certified, commit? [y/n]y
...
```
`openssl req` accepts a `-batch` option so that such prompts are no longer shown. 
1. Can this script be changed to add this to the openssl command?
OR
2. If this user interaction is required, can an option  `-b` (or `--batch` ) be added to allow for non-interactive operation?

If this looks like a valid requests, please let me know and I'll submit a MR for the selected option(I've already implemented option 1 at my end).

**_Edit_**: From an automation viewpoint, the following also works(making this request redundant):
```
[root@xcatmn1 ~]# yes y| /opt/xcat/share/xcat/scripts/setup-local-client.sh samveen
```

immarvin:
hi @samveen , thanks for your proposal.  We prefer option 2:
````
2. If this user interaction is required, can an option -b (or --batch ) be added to allow for non-interactive operation?
````
The option should better be "-f|--force" to keep consistent with the existing xCAT command options.

If you are ok with this, please feel free to create a PR.

thanks

samveen:
@immarvin I'll get it done in a couple of days. (Feel free to add  me in the assigned list)

whowutwut:
@samveen  I just added you to the xcat contributor team, without this we cannot assign you in the "asignee list"  If you accept the invite, we can assign you this issue.  :)

samveen:
@whowutwut Victor, accepted with great pleasure :)

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1793

PR #1793, opened 2016-09-01
Add force option to `setup-local-client.sh`

samveen:
Fix for #1783 

I've added `-f|--force` option to setup-local-client.sh for non-interactive usage. 

Caveat: I've used bash-2.04 compliant syntax. If this assumption about the shell is incorrect, please let me know.

Additional minor fixes:
- Indentation was set to 2 spaces: fixed non-compliant lines
- Fixed multiple spaces intra-line
- Redirected output of `cd -` as it leads to confusion in the output.

### DIFF
--- a/xCAT-server/share/xcat/scripts/setup-local-client.sh
+++ b/xCAT-server/share/xcat/scripts/setup-local-client.sh
@@ -15,15 +15,30 @@ umask 0077 #nothing make by this script should be readable by group or others
 if [ -z "$XCATDIR" ]; then
   XCATDIR=/etc/xcat
 fi
+# if `-f`|`--force` option is supplied, set a variable and remove option from paraameters
+# This allows use of the old code without modification
+FORCE=0
+for v in "$@"; do
+  case "$v" in
+    "-f"|"--force")
+      FORCE=1
+      continue
+      ;;
+  esac
+  ARGS[${#ARGS[@]}]="$v"
+done
+if [ ${#ARGS[@]} -gt 0 ]; then
+  set "${ARGS[@]}"
+fi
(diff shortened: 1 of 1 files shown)


xCAT-server/share/xcat/scripts/setup-local-client.sh
```diff
@@ -15,15 +15,31 @@ umask 0077 #nothing make by this script should be readable by group or others
 if [ -z "$XCATDIR" ]; then
   XCATDIR=/etc/xcat
 fi
+# if `-f`|`--force` option is supplied, set a variable and remove option from paraameters
+# This allows use of the old code without modification
+FORCE=0
+declare -a ARGS
```
immarvin: this script might be run in "ksh" on aix, the "declare" is not supported
````
declare:  not found.
````

samveen: @immarvin Can you confirm ksh version? I'll alter the script to match shell syntax intersection.

immarvin: hi @samveen ,
````
# oslevel -r
7100-03
# echo $SHELL
/usr/bin/ksh
# lslpp -l | grep ksh
  bos.rte.shell             7.1.3.15  COMMITTED  Shells (bsh, ksh, csh)
  bos.rte.shell             7.1.3.15  COMMITTED  Shells (bsh, ksh, csh)
````

samveen:
@immarvin I've just removed the `declare -a` which wasn't strictly necessary. Please check again.

immarvin:
agree to merge

PR #1799, opened 2016-09-01
Add vlan tag to node and group object attributes

gurevichmark:
Add `vlantag` attribute to the display of `node` and `group` objects. This tag comes from `ipmi` table `taggedvlan` column.
Also update man pages describing `group` and `node` objects.


docs/source/guides/admin-guides/references/man7/group.7.rst
```diff
@@ -1146,6 +1146,12 @@ group Attributes:
  
 
 
+\ **vlantag**\  (ipmi.taggedvlan)
```
whowutwut: Wondering if we should  keep the same keyword "taggedvlan" or change to vlantag?

docs/source/guides/admin-guides/references/man7/group.7.rst
```diff
@@ -1146,6 +1146,12 @@ group Attributes:
...
+ 
+ Have bmcsetup place the BMC on the specified vlan tag on a shared netwirk interface.  Some network devices may be incompatible with this option
```
whowutwut: Network is spelled incorrectly. Why some network devices may be incompatible?

Suggest:

Bmcsetup will configure the network interface of the BMC to be tagged to the VLAN specified.

whowutwut:
So it would be "chdef node vlantag=123"

docs/source/guides/admin-guides/references/man7/group.7.rst
daniceexi: The attribute name can be different with the table column, we should use the understandable one. I prefer the ``vlantag``.

whowutwut: Sounds good

gurevichmark:
Yes, "chdef node vlantag=123" to change the vlantag attribute.

docs/source/guides/admin-guides/references/man7/group.7.rst
gurevichmark: This description of the taggedvlan column of ipmi table was there before my pull request.
But I agree that that your suggestion is better. Used another commit to change it.

whowutwut:
I have reviewed the changes, will merge.

docs/source/guides/admin-guides/references/man7/group.7.rst
arif-ali: Should this not be like `bmcvlantag`, as you could have tagged interfaces on different interfaces? this way it would not be confusing?

gurevichmark: Good suggestion, changed to `bmcvlantag` using pull request #1820

PR #1802, opened 2016-09-02
Checking nodetype to distinguish between node/switch discovery

cxhong:
finding this on fs1,  the top of rack switch frame45_sw1 is connect to core switch switch-10-5-23-1, and compute node c39 is connect to top of rack switch.  
`````
[root@fs1 xCAT]# switchprobe switch-10-5-23-1  | grep frame45_sw1
switch-10-5-23-1  Ethernet45  a8:97:dc:02:92:00  frame45_sw1
switch-10-5-23-1  Ethernet45  98:be:94:59:ee:a7  frame45_sw1
switch-10-5-23-1  Ethernet45  70:e2:84:14:02:34  frame45_sw1
switch-10-5-23-1  Ethernet45  a8:97:dc:02:92:2f  frame45_sw1
switch-10-5-23-1  Ethernet45  70:e2:84:14:02:5f  frame45_sw1
[root@fs1 xCAT]# switchprobe frame45_sw1 | grep c910f05c39
frame45_sw1  Ethernet3   98:be:94:59:ee:a7  c910f05c39
````````
when we ran switch-based node discovery for c39, find_mac subroutine will find frame45_sw1 as compute node based on mac address  98:be:94:59:ee:a7 on port 45 of switch-10-5-23-1.  that cause switch unreachable any more and c39 never discovered.   

to fix this:
for all the switch, the nodetype has to set as "switch",  find_mac subroutine will add a new flag to indicate if this is node discovery or switch discovery, then based on nodetype, we will collect (switch-port, node) mapping.

this PR also add code for issue #1798

### DIFF
--- a/perl-xCAT/xCAT/MacMap.pm
+++ b/perl-xCAT/xCAT/MacMap.pm
@@ -439,9 +439,6 @@ sub refresh_table {
     my @switchentries = $self->{switchestab}->getAllNodeAttribs([qw(switch snmpversion username password privacy auth)]);
     my $community = "public";
 
-    #$self->{sitetab} = xCAT::Table->new('site');
-    #my $tmp = $self->{sitetab}->getAttribs({key=>'snmpc'},'value');
-    #if ($tmp and $tmp->{value}) { $community = $tmp->{value} }
     my @snmpcs = xCAT::TableUtils->get_site_attribute("snmpc");
     my $tmp    = $snmpcs[0];
     if (defined($tmp)) { $community = $tmp }
@@ -476,18 +473,23 @@ sub refresh_table {
     #Build hash of switch port names per switch
     $self->{switches} = {};
     
-    #get nodetype from nodetype
+    #get nodetype from nodetype table, build a temp nodetype hash
+    my %typehash;
+    my @typeentries;
     my $ntable = xCAT::Table->new('nodetype');
-    my @typeentries = $ntable->getAllNodeAttribs(['node', 'nodetype']);
-    foreach my $ntnode (@typeentries) {
-        if ($ntnode->{nodetype} eq "switch") {
-            $self->{switches}->{$ntnode->{node} }->{nodetype}  = $ntnode->{nodetype};
-            xCAT::MsgUtils->message("S", "refresh_table: $ntnode->{node} is $self->{switches}->{$ntnode->{node} }->{nodetype}");
-        }
+    if ($ntable) {
+        @typeentries = $ntable->getAllNodeAttribs(['node', 'nodetype']);
     }
+    for my $typeentry (@typeentries) {
+        $typehash{ $typeentry->{node} } = $typeentry->{nodetype};
+    }
+
     foreach my $entry (@entries) {
-        if ( (($discover_switch) and ($self->{switches}->{$entry->{node}}->{nodetype} ne "switch"))
-            or ( !($discover_switch) and ($self->{switches}->{$entry->{node}}->{nodetype} eq "switch")) ){
+        # if we are doing switch discovery and the node is not a switch, skip
+        # if we are NOT doing switch discovery, and the node is a switch, skip
+        my $ntype = $typehash{$entry->{node}};
+        if ( (($discover_switch) and ( $ntype ne "switch"))
+            or ( !($discover_switch) and ( $ntype eq "switch")) ){
             xCAT::MsgUtils->message("S", "refresh_table: skip $entry->{node} and $entry->{switch}");
             next;
         }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: perl-xCAT/xCAT/MacMap.pm [hunk 1]
```diff
@@ -469,7 +475,22 @@ sub refresh_table {
 
     #Build hash of switch port names per switch
     $self->{switches} = {};
+    
+    #get nodetype from nodetype
+    my $ntable = xCAT::Table->new('nodetype');
+    my @typeentries = $ntable->getAllNodeAttribs(['node', 'nodetype']);
+    foreach my $ntnode (@typeentries) {
+        if ($ntnode->{nodetype} eq "switch") {
+            $self->{switches}->{$ntnode->{node} }->{nodetype}  = $ntnode->{nodetype};
```
zet809: The nodetype is only used here, I don't think we shall keep them in $self->{switches} which will be keep all the time.

perl-xCAT/xCAT/MacMap.pm
```diff
@@ -469,7 +475,22 @@ sub refresh_table {
...
+            xCAT::MsgUtils->message("S", "refresh_table: $ntnode->{node} is $self->{switches}->{$ntnode->{node} }->{nodetype}");
+        }
+    }
     foreach my $entry (@entries) {
+        if ( (($discover_switch) and ($self->{switches}->{$entry->{node}}->{nodetype} ne "switch"))
+            or ( !($discover_switch) and ($self->{switches}->{$entry->{node}}->{nodetype} eq "switch")) ){
+            xCAT::MsgUtils->message("S", "refresh_table: skip $entry->{node} and $entry->{switch}");
+            next;
+        }
         if (defined($entry->{switch}) and $entry->{switch} ne "" and defined($entry->{port}) and $entry->{port} ne "") {
```
zet809: There will be an issue here:
For switch-based node discovery, the switch-port which connect to a switch node will also be matched. Just as the scenario we encountered in our fs1 cluster.

So, my idea is:
1. read node,nodetype mapping to a variable.
2. within this loop, if $discover_switch defined, only the switch node is pushed. If not, only the non-switch node is pushed.

perl-xCAT/xCAT/MacMap.pm
```diff
@@ -469,7 +475,22 @@ sub refresh_table {
...
+    my $ntable = xCAT::Table->new('nodetype');
+    my @typeentries = $ntable->getAllNodeAttribs(['node', 'nodetype']);
+    foreach my $ntnode (@typeentries) {
+        if ($ntnode->{nodetype} eq "switch") {
+            $self->{switches}->{$ntnode->{node} }->{nodetype}  = $ntnode->{nodetype};
+            xCAT::MsgUtils->message("S", "refresh_table: $ntnode->{node} is $self->{switches}->{$ntnode->{node} }->{nodetype}");
+        }
+    }
     foreach my $entry (@entries) {
+        if ( (($discover_switch) and ($self->{switches}->{$entry->{node}}->{nodetype} ne "switch"))
```
whowutwut: I think we can change some logic here to avoid adding the `nodetype=switch` to the hash entry.  In the above loop at line 484, we are adding in ONLY switches into this hash table.  So we can test if an entry exists in the hash, that would be a switch. The danger is that any code that does not test for the existence of the entry will then add an entry into the hash....

perl-xCAT/xCAT/MacMap.pm
```diff
@@ -469,7 +472,27 @@ sub refresh_table {
 
     #Build hash of switch port names per switch
     $self->{switches} = {};
+    
+    #get nodetype from nodetype table, build a temp nodetype hash
+    my %typehash;
```
zet809: The hash return format of getAllNodeAttribs can be used like this:
my $typehash = $ntable->getAllNodeAttribs(['node','nodetype'], 1);
The $typehash will be like this:
'switch-10-5-23-1' => [
                                  {
                                    'node' => 'switch-10-5-23-1',
                                    'nodetype' => 'switch'
                                  }
                                ],

perl-xCAT/xCAT/MacMap.pm
```diff
@@ -469,7 +472,27 @@ sub refresh_table {
...
+        @typeentries = $ntable->getAllNodeAttribs(['node', 'nodetype']);
+    }
+    for my $typeentry (@typeentries) {
+        $typehash{ $typeentry->{node} } = $typeentry->{nodetype};
+    }
+
     foreach my $entry (@entries) {
+        # if we are doing switch discovery and the node is not a switch, skip
+        # if we are NOT doing switch discovery, and the node is a switch, skip
+        my $ntype = $typehash{$entry->{node}};
```
zet809: Need to check whether $ntype exist

cxhong: Can't check that since most time predefined node didn't have nodetype attribute defined.

zet809:
@cxhong Looks good to me, agree to merge. Thx a lot!

PR #1806, opened 2016-09-07
support vlan for rspconfig

penguhyang:
get and set the vlan info from bmc.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -929,6 +929,19 @@ sub setnetinfo {
             $mask[$_] = $mask[$_] + 0;
         }
         @cmd = (0x01, $channel_number, 12, @mask);
+    } elsif ($subcommand =~ m/vlan/) {
+        unless ( int($argument) == $argument ) {
+            $callback->({ errorcode => [1], error => ["The input $argument is invalid, please input an integer"] });
+            return;
+        }
+        unless (($argument>=1) && ($argument<=4096)) {
+            $callback->({ errorcode => [1], error => ["The input $argument is invalid, valid value [1-4096]"] });
+            return;
+        }
+        my @vlanparam = ();
+        $vlanparam[0] = ($argument & 0xff);
+        $vlanparam[1] = 0x80 | (($argument & 0xf00) >> 8);
+        @cmd = (0x01, $channel_number, 0x14, @vlanparam);
     } elsif ($subcommand =~ m/ip/ and $argument =~ m/dhcp/) {
         @cmd = (0x01, $channel_number, 0x4, 0x2);
     } elsif ($subcommand =~ m/ip/) {
@@ -1058,6 +1071,9 @@ sub getnetinfo {
     elsif ($subcommand =~ m/^snmpdest(\d+)/) {
         @cmd = (0x02, $channel_number, 0x13, $1, 0x00);
     }
+    elsif ($subcommand eq "vlan") {
+        @cmd = (0x02, $channel_number, 0x14, 0x00, 0x00);
+    }
     elsif ($subcommand eq "ip") {
         @cmd = (0x02, $channel_number, 0x03, 0x00, 0x00);
     }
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -929,6 +929,9 @@ sub setnetinfo {
             $mask[$_] = $mask[$_] + 0;
         }
         @cmd = (0x01, $channel_number, 12, @mask);
+    } elsif ($subcommand =~ m/vlan/) {
+        my $vlanid = $argument;
+        @cmd = (0x01, $channel_number, 0x14, $vlanid);
```
zet809: Why don't you use $argument directly?
I think the data type shall be check, a number between 1 to 4096.

penguhyang: fixed.

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -929,6 +929,14 @@ sub setnetinfo {
             $mask[$_] = $mask[$_] + 0;
         }
         @cmd = (0x01, $channel_number, 12, @mask);
+    } elsif ($subcommand =~ m/vlan/) {
+        unless ( int($argument) == $argument ) {
+            xCAT::SvrUtils::sendmsg("VLAN ID is invalid, please input a integer", $callback, $sessdata->{node}, %allerrornodes);
+        }
```
zet809: shall it return if error encountered?

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -929,6 +929,14 @@ sub setnetinfo {
...
+        unless (($argument>=1) && ($argument<=4096)) {
+            xCAT::SvrUtils::sendmsg("VLAN ID is invalid, valid value [1-4096]", $callback, $sessdata->{node}, %allerrornodes);
+        }
+        @cmd = (0x01, $channel_number, 0x14, $argument);
```
zet809: The vlanid larger than 256 can not be put in 1 byte. You shall follow the specification to send data correctly.

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1125,6 +1136,14 @@ sub getnetinfo_response {
                 $returnd[6],
                 $returnd[7],
                 $returnd[8]), $callback, $sessdata->{node}, %allerrornodes);
+    } elsif ($subcommand eq "vlan") {
+        if (($returnd[3] >> 7) & 0x01) {
+            xCAT::SvrUtils::sendmsg(sprintf("$format %d" . $bmcifo,
+                "BMC VLAN ID enabled:",
+                (($returnd[3] & 15) << 8) | $returnd[2]), $callback, $sessdata->{node}, %allerrornodes);
```
zet809: I think we always use 0x0f rather than 15 when operating bytes.
And will you pls add some comments here what is the meaning for $returnd[2] and $returnd[3]

xCAT-server/lib/xcat/plugins/ipmi.pm
penguhyang: fixed.

xCAT-server/lib/xcat/plugins/ipmi.pm
penguhyang: fixed.

xCAT-server/lib/xcat/plugins/ipmi.pm
penguhyang: fixed.

whowutwut:
After a code review its best to commit in another commit so that it's easier to see the changes made resulting in the review instead of amend.

whowutwut:
Should there be some associated man page changes for adding this function? Is only for open power based machine?

zet809:
Yes, the manpage of rspconfig will be updated. 
And the vlan functions can be used for all BMC based serer not only for open power.

zet809:

Issue #1801, opened 2016-09-02
creating a lot of new nodes during pre-disovery causes conserver a long time to acquire locks.


ralphbellofatto:
We Are attempting to use MTMS based discovery so we created a stanza file containing 80+ new nodes.

When we do this and then startup conserver, we notice that it takes a long for the console to acquire their lock before we can use them.

we have setup our site configuration with "consoleondemand=no" so we are logging all of the console traffic all of the time.

I have attached the stanza file with the new node definitions here:
[newnodes.stanza.txt](https://github.com/xcat2/xcat-core/files/451867/newnodes.stanza.txt)

when attempting to try to find out why all the console would show "Acquiring lock.... for a really long time we instrumented the code in:
/opt/xcat/share/xcat/cons/ipmi
to try to tell what is ho long the it takes to acquire the lock..

```
sub get_lock {
    unless (flock(LOCKHANDLE, LOCK_EX | LOCK_NB)) {
        $| = 1;
        print "$ARGV[0] Acquiring startup lock...";
        ($seconds, $microseconds) = gettimeofday;  $starttime=$seconds+($microseconds/1000000);

        flock(LOCKHANDLE, LOCK_EX) or die "Fatal Error securing startup lock";
        ($seconds, $microseconds) = gettimeofday;  $endtime=$seconds+($microseconds/1000000);
        $deltatime = $endtime-$starttime;
        print "done $deltatime MS\n";
    }
    truncate(LOCKHANDLE, 0);
    print LOCKHANDLE $$ . "\n";
}
```

When examining the code in the /var/log/consoles/*

we see acquire lock times between 30 and 4000 seconds.

```
[Thu Sep  1 23:45:16 2016]c460c043 Acquiring startup lock...done 33.6890850067139 MS
[Fri Sep  2 00:51:30 2016]c460c090 Acquiring startup lock...done 4661.84337496758 MS
```

And the problem persists and never stabilizes, making trying to use the system.

Here are the times to acquire the locks for nodes that have bmc's that we can connect to.

```
c460c001 Acquiring startup lock...done 265.741617918015 MS
c460c002 Acquiring startup lock...done 1489.7107770443 MS
c460c003 Acquiring startup lock...done 2173.9434659481 MS
c460c004 Acquiring startup lock...done 223.871228933334 MS
c460c005 Acquiring startup lock...done 95.7902700901031 MS
c460c007 Acquiring startup lock...done 280.591845035553 MS
```

[...]

PR #1811, opened 2016-09-08
Support yes and no for node object attribute consoleondemand

gurevichmark:
This pull request fixes the attribute value inconsistency of `consoleondemad` attribute in `nodehm` table. Reported by issue #1801 

1. Verifies that `consoleondemad` attribute in `site` table is set to "yes" or "no". Error is reported if something else.
2. Allows case insensitive setting for `consoleondemad` attribute in `site` table
3. Allows case insensitive setting for `consoleondemad` attribute in `nodehm` table to "yes", "no" in addition to existing 0, or 1


xCAT-server/lib/xcat/plugins/conserver.pm
```diff
@@ -575,10 +577,11 @@ sub donodeent {
             }
         }
         if (defined($cfgent->{consoleondemand})) {
-            if ($cfgent->{consoleondemand} && !$siteondemand) {
+            # consoleondemand attribute for node can be "1", "yes", "0" and "no"
+            if ((($cfgent->{consoleondemand} eq "1") || lc($cfgent->{consoleondemand}) eq "yes") && !$siteondemand) {
                 push @$content, "  options ondemand;\n";
             }
-            elsif (!$cfgent->{consoleondemand} && $siteondemand) {
+            elsif ((($cfgent->{consoleondemand} eq "0") || lc($cfgent->{consoleondemand}) eq "no") && $siteondemand) {
```
daniceexi: @gurevichmark 
I think the code logic here is a little strange. As mentioned in the definition of node.consoleondemand:
~~~~
consoleondemand:	This overrides the value from site.consoleondemand; (0=no, 1=yes). Default is the result from site.consoleondemand.
~~~~

The setting in the ``node.consoleondemand`` has higher priority. If it was set, we don't need to check the ``$siteondemand``.

And we need to change the man page for ``site.consoleondemand`` to support **yes|on**.

gurevichmark: 1. Yes, I agree that those lines are confusing, but I think it is done this way so as to not repeat the settings from `default` section of `/etc/conserver.cf`
`/etc/conserver.cf` file has a section for `default` and a section for each node.
For example, if `site.consoleondemand` is set to `yes`, we will put a line `option ondemand` into the `default` section. Then if `nodehm.consoleondemand` is also set to `yes`, the way the code is written today, we will not insert the same line into the individual node section of `/etc/conserver.cf` file.
If I change the code to only check the setting of `nodehm.consoleondemand`, we could repeat the same line from `default` section into the individual node section.
2. I made a second commit to update the `Schema.pm` file which is used to generate man pages containing `consoleondemand` attribute.

whowutwut: Shouldn't the node overwrite go into each node section of the `/etc/conserver.cf` file? maybe i'm not following this conv very well...

daniceexi:
Then I am fine with the change and agree to merge.

Issue #1805, opened 2016-09-07
Enable Kdump can not work on diskless image for rh7.3 snapshot1


wwzhubj:
I follow steps in  the doc: 
http://xcat-docs.readthedocs.io/en/stable/guides/admin-guides/manage_clusters/ppc64le/diskless/customize_image/enable_kdump.html
But the core files are not saved in the directory I set. 
[root@c712ems4 ~]# lsxcatd -v
Version 2.12.2 (git commit 6fff020a905c64cabe31896e653aebc942ee0694, built Thu Sep  1 06:18:21 EDT 2016)

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1812

Test case required for rh7.x bundle

immarvin:
ping @tingtli

daniceexi:
@immarvin for the fix which needs to be test or add new test case, add the label **component:test**

wwzhubj:
verified it on Sep 19 version, closed it.

hu-weihua:
Test case ``linux_diskless_kdump`` was added by pull request #4964

PR #1812, opened 2016-09-09
fix issue Enable Kdump can not work on diskless image for rh7.3 snapshot1 #1805;add support for kdump on rh7.x

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/1805

add support for kdump on redhat 7.x

verified on rhels7.[1,2,3beta2] on ppc64le

### DIFF
--- a/xCAT-server/share/xcat/netboot/rh/genimage
+++ b/xCAT-server/share/xcat/netboot/rh/genimage
@@ -1000,12 +1000,17 @@ sub mkinitrd_dracut {
 
         #update etc/dracut.conf
         open($DRACUTCONF, '>', "$rootimg_dir/etc/dracut.conf");
+        my $dracutmodulelist = "xcat nfs base network kernel-modules syslog ";
+
+        if (-d glob($dracutmoduledir . "[0-9]*systemd")) {
+            $dracutmodulelist .= " systemd ";
+        }
+
         if (-d glob($dracutmoduledir . "[0-9]*fadump")) {
-            print $DRACUTCONF qq{dracutmodules+="xcat nfs base network kernel-modules syslog fadump"\n};
+            $dracutmodulelist .= " fadump ";
         }
-        else {
-            print $DRACUTCONF qq{dracutmodules+="xcat nfs base network kernel-modules syslog"\n};
-        }
+
+        print $DRACUTCONF qq{dracutmodules+="$dracutmodulelist"\n};
         print $DRACUTCONF qq{add_drivers+="$add_drivers"\n};
         print $DRACUTCONF qq{filesystems+="nfs"\n};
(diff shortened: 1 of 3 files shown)


xCAT/postscripts/enablekdump
```diff
@@ -219,9 +219,29 @@ EOF
...
+               #with "command line overflow" since the kernel options are longer than 255 chars`
+               kdumpcmdline=
+               for i in $(cat /proc/cmdline)
+                   do 
+                       if [[ $i == console=* ]] || [[ $i == crashkernel=* ]]; then 
+                             kdumpcmdline="$kdumpcmdline $i "
+                       fi
+               done
+               sed -i "s#^[\t ]*KDUMP_COMMANDLINE=\"#KDUMP_COMMANDLINE=\"$kdumpcmdline#" /etc/sysconfig/kdump
+               sed -i "s#^[\t ]*KDUMP_COMMANDLINE_APPEND=\"#KDUMP_COMMANDLINE_APPEND=\"root=nfs:$KDIP:$KDPATH #" /etc/sysconfig/kdump
```
daniceexi: If the ``root=`` is a fake parameter for kdump, I prefer to make it looks like ``root=nfs:this_is_a_fake_param_for_rh7_workaround`` so that the later developer could know what it's for.

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -1000,12 +1000,17 @@ sub mkinitrd_dracut {
...
+        my $dracutmodulelist = "xcat nfs base network kernel-modules syslog ";
+
+        if (-d glob($dracutmoduledir . "[0-9]*systemd")) {
+            $dracutmodulelist .= " systemd ";
         }
-        else {
-            print $DRACUTCONF qq{dracutmodules+="xcat nfs base network kernel-modules syslog"\n};
+
+        if (-d glob($dracutmoduledir . "[0-9]*fadump")) {
+            $dracutmodulelist = "$dracutmodulelist fadump";
```
daniceexi: it should be consistent to use ``=.``.

xCAT/postscripts/enablekdump
immarvin: ``root=nfs:<server ip>:<directory>`` will be mounted and validated during the initramfs bootup, we can not specify it to be a arbitrary value, it must be a nfs shared directory with file/directory "proc" inside.  I use the kdump target directory here

xCAT-server/share/xcat/netboot/rh/genimage
immarvin: fixed

daniceexi:

PR #1814, opened 2016-09-09
Buildsanity

jjohnson42:


whowutwut:
@jjohnson42  I take a review.   Also @immarvin @daniceexi can you also take a look at this pull request?  If we are changing the version ,should we target this for major version number release (2.13) in dec... or change it in a minor release.   I'll build this in my sandbox to see the results.  Since it's on master, there's nothing that Jarrod needs to change, we just decide when to actually merge the pull request, (hopefully no conflict resolution is needed if we hold off)

makerpm
```diff
@@ -55,7 +55,7 @@ function makenoarch {
 
 		tar --exclude .svn -czf $RPMROOT/SOURCES/$RPMNAME-$VER.tar.gz $RPMNAME
 		rm -f $RPMROOT/SRPMS/$RPMNAME-$VER*rpm $RPMROOT/RPMS/noarch/$RPMNAME-$VER*rpm
-		rpmbuild $QUIET -ta $RPMROOT/SOURCES/$RPMNAME-$VER.tar.gz
+		rpmbuild $QUIET -ta $RPMROOT/SOURCES/$RPMNAME-$VER.tar.gz --define "version $VER" $RELEASE
```
arif-ali: Shouldn't this be `--define "version $VER" --define "release $RELEASE"`

xCAT-genesis-scripts/xCAT-genesis-scripts.spec
```diff
@@ -13,7 +13,8 @@ BuildArch: noarch
 %define __spec_install_post :
 %define debug_package %{nil}
 %define __prelink_undo_cmd %{nil}
-Release: snap%(date +"%Y%m%d%H%M")
+Version: %{?version:%{version}}%{!?version:%{version}}
```
immarvin: should this be ``Version: %{?version:%{version}}%{!?version:%(cat Version)}``

immarvin:
hi @whowutwut , 
I have several comments on this change:

1. ``git describe`` only shows annotated tags(tags created with "-a","-u" or "-s")

````
man git-describe
...

       By default (without --all or --tags) git describe only shows annotated tags. For more information about  creating annotated tags see the -a and -s options to git-tag(1).


man git-tag
...

       Tag objects (created with -a, -s, or -u) are called "annotated" tags; they contain a creation date, the tagger  name and e-mail, a tagging message, and an optional GnuPG signature. Whereas a "lightweight" tag is simply a name for an object (usually a commit object).
````

So as a requisite for "USEGITVER=1" , we should create tags with ``git tag -a/-s/-u`` on each of xCAT releases. 


2. since "USEGITVER" is a new ENV variable to control the build behaviour, we need some description on it at the beginning of "buildcore.sh", "build-ubunturepo.sh", just as "GITPULL", "UP"...

3.  Should build scripts of ubuntu deb packages be taken into consideration? if yes, we may need to modify the ubuntu build scripts as well.

whowutwut:
@immarvin , did you mean to direct that last comment to Jarrod @jjohnson42 ?

zet809:
Seems fixed with Pull Request #2195

zet809:
The issue had been fixed in #2195, so I close this one.

PR #1820, opened 2016-09-12
Change attribute name from vlantag to bmcvlantag

gurevichmark:
As suggested by @arif-ali , changing the name of the attribute from from `vlantag` to `bmcvlantag`.


docs/source/guides/admin-guides/references/man7/node.7.rst
```diff
@@ -133,6 +133,12 @@ node Attributes:
  
 
 
+\ **bmcvlantag**\  (ipmi.taggedvlan)
+ 
+ Bmcsetup will configure the network interface of the BMC to be tagged to the VLAN specified.
```
whowutwut: since this is a script, should we keep it lower case? `bmcsetup`

gurevichmark: Changed to "bmcsetup script".

whowutwut:
Thanks, will merge

Issue #1818, opened 2016-09-12
Configbond is not work


wjianf:
## Configured nic bonding with "configbond bond0 ib0@ib2 mode=1", but the bonding nic couldn't be brought up with the message as below:

[root@pcm421rhel71mn postscripts]# updatenode compute000 confignics
compute000: Mon Sep 12 05:08:39 EDT 2016 Running postscript: confignics
compute000: confignics on compute000: config install nic:0, remove: 0, iba ports:
compute000: eth2!192.168.12.3
compute000: ib1!192.168.199.2
compute000: bond0!192.168.200.2
compute000: bmc!9.111.251.142
compute000: bond0!configbond bond0 ib0@ib2 mode=1
compute000: confignics on compute000: processing custom scripts: configbond bond0 ib0@ib2 mode=1 for interface bond0
compute000: configbond: Create bond interface bond0 with slaves="ib0 ib2" opts="mode=1"
compute000: configbond: IP information for bond0: IP=192.168.200.2; network=192.168.200.0; netmask=255.255.255.0
## After adding 'TYPE=Bond' in bonding config files, the bonding can be brought up

[root@compute000 ~]# cat /etc/sysconfig/network-scripts/ifcfg-bond0
DEVICE=bond0
BOOTPROTO=none
IPADDR=192.168.200.2
NETMASK=255.255.255.0
ONBOOT=yes
USERCTL=no
BONDING_OPTS="mode=1"
[root@compute000 ~]# ifup bond0
Error: no device found for connection 'System bond0'.
[root@compute000 ~]# echo "TYPE=Bond" >> /etc/sysconfig/network-scripts/ifcfg-bond0
[root@compute000 ~]# ifup bond0
Connection successfully activated (master waiting for slaves) (D-Bus active path: /org/freedesktop/NetworkManager/ActiveConnection/14)
You have mail in /var/spool/mail/root
[root@compute000 ~]# ifconfig bond0
bond0: flags=5187<UP,BROADCAST,RUNNING,MASTER,MULTICAST>  mtu 1500
        inet 192.168.200.2  netmask 255.255.255.0  broadcast 192.168.200.255
        inet6 fe80::6833:76ff:fed4:57a5  prefixlen 64  scopeid 0x20<link>
        ether 6a:33:76:d4:57:a5  txqueuelen 0  (Ethernet)
        RX packets 0  bytes 0 (0.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
[...]

PR #1827, opened 2016-09-13
fix issue 1818 add type bond in configbond

bybai:
#1818 
https://github.com/xcat2/xcat-core/issues/1818

Add TYPE=Bond in master config file.


immarvin:
agree to merge

PR #1829, opened 2016-09-13
issue 1818 add type bond in configbond

bybai:
#1818 

Add TYPE=Bond in configbond.


immarvin:
agree to merge

immarvin:
since for PCM branch, we need their request to push new commits in.

PR #1834, opened 2016-09-13
Probe to catch duplicate MTM+SERIAL node definitions

gurevichmark:
Probe to catch duplicate node definitions that have the same mtm and serial numbers.
Also discovered nodes that do not have matching pre-defined nodes, reported as "Waiting to be discovered"


whowutwut:
@zet809 can you take a look at this also?.  I think we are trying to find a better name for the probe option. .

xCAT-probe/subcmds/uniq_definitions
```diff
@@ -0,0 +1,180 @@
...
+    $program_name -h
+    $program_name [-d] [-n noderange] [-V]
+
+Description:
+    Use this command to check if any node defintions in xCAT DB have duplicate model type and serial numbers. 
+
+Options:
+    -h : Get usage information of $program_name
+    -n : Range of nodes to check
+    -d : Remode duplicate model type and serial number node definition from xCAT DB
```
zet809: Shall remode be remove

xCAT-probe/subcmds/uniq_definitions
```diff
@@ -0,0 +1,180 @@
...
+
+if (scalar(@all_nodes_mtm_serial_lines) <= 0) {
+
+    # There were no nodes matching the noderange. Issue a warning and exit.
+    probe_utils->send_msg("$output", "w", "No nodes matching the noderange were found.");
+    exit 1;
+}
+
+# Build a hash of key="nodename" value="mtm+serial"
+foreach (@all_nodes_mtm_serial_lines) {
```
zet809: There is error lines also appearing in the array \@all_nodes_mtm_serial_lines like this, so all the error lines will be append to a hash entry with key equal 'Error'
# lsdef -i mtm,serial -c bb,dd,cc,frame45cn02
Error: Could not find an object named 'bb' of type 'node'.
Error: Could not find an object named 'cc' of type 'node'.
Error: Could not find an object named 'dd' of type 'node'.
frame45cn02: mtm=8335-GTB
frame45cn02: serial=100499A

xCAT-probe/subcmds/uniq_definitions
```diff
@@ -0,0 +1,180 @@
...
+    probe_utils->send_msg("$output", "d", "Processing $_.") if ($VERBOSE);
+    my ($node_name, $value) = split ":", $_;
+    if (exists($node_mtm_serial_hash{$node_name})) {
+        # already have an entry for this node with mtm, concat the serial
+        $value = $node_mtm_serial_hash{$node_name} . $value;
+    }
+    $node_mtm_serial_hash{$node_name} = $value;
```
zet809: Why not get all the $waiting_to_be_discovered nodes here? With key equal $mtms, and $value is nodename.
Then, remove it from the $node_mtm_serial_hash.
When in the while loop, only check whether the $mtms of a node also point to a waiting_for_discoved node, if yes delete it if required.

daniceexi:
A general comment that I think we should make this **node validation** check to be common besides the MTMS check. Maybe the subcmd name can be **nodecheck** and the MTMS check can be one of the sub function.

gurevichmark:
@daniceexi and @zet809 Thank you for your suggestions. I will carefully review your comments and will make the changes in the next few days.

xCAT-probe/subcmds/uniq_definitions
```diff
@@ -0,0 +1,180 @@
...
+my $any_dups = 0;
+while (($node_name, $mtm_serial) = each %node_mtm_serial_hash) {
+    # Check if hash already has the same key indicating another node definition has the same mtm+serial
+    if (exists($mtm_serial_node_hash{$mtm_serial})) {
+        if ($mtm_serial eq " mtm= serial=") {
+            # Exclude entries that do not have mtm+serial set
+            probe_utils->send_msg("$output", "d", "No mtm and no serial for node $node_name") if ($VERBOSE);
+            next;
+        }
+        probe_utils->send_msg("$output", "f", "Duplicate node definition for $mtm_serial : $node_name and $mtm_serial_node_hash{$mtm_serial}.");
```
whowutwut: Suggest changing this message to be 
`Duplicate found for machine with mtm=<>, serial=<>:  <def1> and <def2>.`

xCAT-probe/subcmds/uniq_definitions
```diff
@@ -0,0 +1,180 @@
...
+my $program_name = basename("$0");
+my $help;
+my $noderange = "";
+my $test;
+my $output  = "stdout";
+my $verbose = 0;
+my $rst     = 0;
+# Match pattern for discovered nodes: node-1234-ABCDE-123456
+#                                     node- d+ -[a-z]- d+
+my $discovered_node_pattern = '^node-\d+-[a-zA-Z]+-\d+';
```
whowutwut: I think the regular expression should be kept simple for now, maybe only detect `node-*'` definitions. 

We can enhance it in the future if there is issues opened against this very broad regex.  Really xCAT is using that `node-$mtm-$serial` format like @zet809  said above...   If a customer names their machines the same, maybe we can either recommend to not use that. or create a more uniq name... `_node_$mtm-$serial`  the underscores there is intentional to create a more unique temporary node name.  but maybe not do that extra work for now and see the reaction to the probe.

whowutwut:
:+1:  to @daniceexi  suggestion for nodecheck.  `xcatprobe nodecheck` seems good.

gurevichmark:
Pushed another commit to implement suggested changes:
1. Probe renamed to be `nodecheck`
2. Added flags and controls to allow for future expansion to probe other node attributes
3. Simplified processing for hash `waiting_to_be_discovered`
4. Changed the discovered node name pattern to be `node-*`
5. Discard std error output from `lsdef` command if invalid node names are specified in the noderange
6. Changed the output display to be:
```
[root@fs1 subcmds]# xcatprobe nodecheck -c duplicate_mtm
Duplicate found for machine with  mtm=8335-GTA serial=1210532 : frame45cn10 and node-8335-gta-1210532             [FAIL]
node-8335-gtb-2109d0a: Waiting to be discovered                                                                   [ OK ]
[root@fs1 subcmds]#
```

whowutwut:
@gurevichmark  Similar to the comment in #1874, can we possibly just have `xcatprobe nodecheck` be the main caller of this command.  We can do the underlying checks without the user specifying a particular `-c` option to the probe.  

I'd like the first pass of these probe commands to be very user friendly.  So `xcatprobe nodecheck` is much easier to remember than to have a `-c duplicate_mtm` option that is required.   We should just probe all things that could be potential problems with the node definitions.

gurevichmark:
@whowutwut Changes mode. All tests will be performed one after another.

gurevichmark:
@whowutwut Removed RCs for each probe step.

whowutwut:
Thanks Mark! I have reviewed and agree to merge.  We can keep building on this probe function to add in more checks to help make the discovery process easier to debug.

PR #1835, opened 2016-09-14
add syslinux-xcat and xnba-undi to the dependency of xCAT and xCATsn to support mixed cluster

immarvin:
add syslinux-xcat and xnba-undi to the dependency of xCAT and xCATsn to support mixed cluster


zet809:
I think the xCATsn/debian/control file also need to modify.

immarvin:
xCATsn/debian/control has already included syslinux-xcat as a dependency without arch

zet809:
Agree to merge.

Issue #1837, opened 2016-09-14
[Customer]a network configuration of interface named "bootnic" will be created on CN when the "site.managedaddressmode='static'"


immarvin:
from mail list:

>     >     >     Hi,
>     >     >
>     >     >     when deploying a compute node with SL7.2, xCAT generates
>     >     >     /etc/sysconfig/network-scripts/ifcfg-bootnic instead of
>     >     >     /etc/sysconfig/network-scripts/ifcfg-eth0. After the first reboot of the
>     >     >     machine, the network is not working because it looks for a device named
>     >     >     "bootnic" that does not exists. How can I tell xCAT to create
>     >     >     /etc/sysconfig/network-scripts/ifcfg-eth0 instead?
>     >     >
>     >     >     I have set the kernel parameters "net.ifnames=0 biosdevname=0" because
>     >     >     we need the old style network device names.
>     >     >
>     >     >
>     >     >     Markus

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1838

PR #1838, opened 2016-09-14
fix issue:[Customer]a network configuration of interface named bootnic will be created on CN when the site.managedaddressmode=static #1837

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/1837

do not rename the install nic to "bootnic" in diskful installation when site.managedaddressmode="static"


whowutwut:
@gurevichmark  can you also review when you have a chance

gurevichmark:
Looks good to me. Not merging to allow @daniceexi to review also.

russauld:
Doesn't this change remove the nicname completely? What if the user actually specified it in the networks or nics table?

immarvin:
hi @russauld , thanks for your comment, code updated

Issue #1803, opened 2016-09-02
node discovery won't keep nodeset parameters if genesis is still running.


ralphbellofatto:
We attempted use the MTMS method to do discovery.    We are generally following this process..

http://xcat-docs.readthedocs.io/en/stable/guides/admin-guides/manage_clusters/ppc64le/discovery/mtms/discovery_using_defined.html

When we get to the end of this process, the genesis image is still running on the system in a loop:

we will see this on the rcons:
```
<166>Sep  2 14:28:57 xcat.genesis.bmcsetup: Enabling the non-volatile channel access (1)
<166>Sep  2 14:28:57 xcat.genesis.bmcsetup: Enabling the non-volatile channel access (1): OK
<166>Sep  2 14:28:57 xcat.genesis.bmcsetup: Enabling the volatile channel access (1)        
<166>Sep  2 14:28:57 xcat.genesis.bmcsetup: Enabling the volatile channel access (1): OK    
<166>Sep  2 14:28:57 xcat.genesis.bmcsetup: Enabling ARP responses                          
<166>Sep  2 14:28:57 xcat.genesis.bmcsetup: Enabling ARP responses: OK                      
<166>Sep  2 14:28:57 xcat.genesis.bmcsetup: Enabling IPMI MD5 LAN access                    
<166>Sep  2 14:28:58 xcat.genesis.bmcsetup: Enabling IPMI MD5 LAN access: OK                
<166>Sep  2 14:28:58 xcat.genesis.bmcsetup: Enabling IPMI v 2.0 LAN access                  
<166>Sep  2 14:29:10 xcat.genesis.bmcsetup: ACCESS=cXaaaaaaaaaaXXX
<166>Sep  2 14:29:10 xcat.genesis.bmcsetup: Set the cipher_privileges for the channel
<166>Sep  2 14:29:11 xcat.genesis.bmcsetup: Set the cipher_privileges for the channel: OK
<166>Sep  2 14:29:11 xcat.genesis.bmcsetup: Enabling SOL for channel 1
<166>Sep  2 14:29:11 xcat.genesis.bmcsetup: Enabling SOL for channel 1: OK
<166>Sep  2 14:29:11 xcat.genesis.bmcsetup: Enabling SOL for ADMIN
<166>Sep  2 14:29:11 xcat.genesis.bmcsetup: Enabling SOL for ADMIN: OK
[...]

Issue #1788, opened 2016-08-30
Update the documentation to set the default behavior for node discovery to directly provision the OS


whowutwut:
The current documentation describes the process for running a chain command `runcmd=bmcsetup` to set up OpenPower servers.  However, the genesis kernel just sits there in a loop looking for the next chain command. 

This seems to be an area where there is much confusion regarding what xCAT is doing and whether discovery is complete or in some error state (Maybe a update to the msg should be made to help the user who is looking at the console, get an idea of what should be done next.)

For example, maybe the msg: 
`"Received request to retry in a bit, will call xCAT back in $delay seconds"` 

Should be changed 
`Received request=$dest, will call xCAT back in $delay seconds.  Discovery is complete, run nodeset on this node to provision an Operating System`

Also would suggest that we document the default process to include the `osimage=` attribute in the chain command instead of just leaving the genesis running in an infinite loop.

zet809:
@xuweibj Hi, xu wei, will you pls take a look at this defect? Thx!

whowutwut:
@zet809  @xuweibj  I actually took care of this by the pull request referenced above, so we can close this.

Issue #1422, opened 2016-06-28
after os installation, the system will enter into genesis, on garrison node. 


wwzhubj:
I will write this issue down, since it's difficult to recreate it. 
```
[root@c712ems4 ~]# lsxcatd -v
Version 2.12 (git commit 5df15f39216da9c64759d61b19fbabad1674727d, built Wed May 18 07:30:41 EDT 2016)
[root@c712ems4 ~]# uname -a
Linux c712ems4 3.10.0-327.13.1.el7.openpower3.ppc64le #1 SMP Thu Apr 28 13:22:48 EDT 2016 ppc64le ppc64le ppc64le GNU/Linux
```
```
[root@c712ems4 tools]# lsdef node-8335-gtb-100585a
Object name: node-8335-gtb-100585a
    bmc=192.168.130.214
    cons=ipmi
    groups=all
    hwtype=bmc
    mgt=ipmi
    mtm=8335-GTB
    nodetype=mp
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    serial=100585A
    statustime=06-28-2016 01:47:23

[root@c712ems4 tools]# ping 192.168.130.214
PING 192.168.130.214 (192.168.130.214) 56(84) bytes of data.
64 bytes from 192.168.130.214: icmp_seq=1 ttl=64 time=0.345 ms
^C
--- 192.168.130.214 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
```

```
[root@c712ems4 tools]# lsdef node-8335-gtb-100585a
Object name: node-8335-gtb-100585a
    bmc=192.168.130.214
    cons=ipmi
    groups=all
    hwtype=bmc
    mgt=ipmi
    mtm=8335-GTB
    nodetype=mp
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    serial=100585A
    status=booting
    statustime=06-28-2016 02:22:02
[root@c712ems4 tools]# ping 192.168.130.214
PING 192.168.130.214 (192.168.130.214) 56(84) bytes of data.
From 192.168.130.1 icmp_seq=1 Destination Host Unreachable
```

then I see the 100585a 's mac has been added into c712f9n01 . 
```
[root@c712ems4 tools]#  tabdump mac|grep 70:e2:84:14:0f:0c
"c712f9n01-pvt",,"70:e2:84:14:0f:0c",,

[...]

PR #1844, opened 2016-09-15
Update the documentation to recommend OS is installed when doing discovery

whowutwut:
I have been getting a lot of request and questions about the xCAT genesis kernel and why it's looping forever in the discovery process.  So to avoid this, I think the default recommendation when doing discovery and running `bmcsetup` is to go directly into provisioning the Operating system, rather than having the genesis kernel loop looks for the next destiny... until the admin does a `nodeset` to provision the OS.  

This might help alleviate some confusion..

Fix issue #1803 and #1422


perl-xCAT/xCAT/Schema.pm
```diff
@@ -358,7 +358,7 @@ passed as argument rather than by table value',
...
-            chain => 'A comma-delimited chain of actions to be performed automatically when this node is discovered. ("Discovered" means a node booted, but xCAT and DHCP did not recognize the MAC of this node. In this situation, xCAT initiates the discovery process, the last step of which is to run the operations listed in this chain attribute, one by one.) Valid values:  boot, runcmd=<cmd>, runimage=<URL>, shell, standby. (Default - same as no chain - it will do only the discovery.).  Example, for BMC machines use: runcmd=bmcsetup,shell.',
+            chain => 'A comma-delimited chain of actions to be performed automatically when this node is discovered for the first time.  (xCAT and the DHCP server does not recognize the MAC address of the node and xCAT initializes the discovery process.  The last step in this process is to run the operations listed in the chain attribute, one by one.  Valid values:  boot, runcmd=<cmd>, runimage=<URL>, shell, standby. Example, to have the genesis kernel pause, chain=shell.',
```
gurevichmark: 1. In the sentence:
`(xCAT and the DHCP server does not recognize the MAC address of the node and xCAT initializes the discovery process.`
- Missing closing ")" ?
- Should "does" be "do" ?
- should "and xCAT" be "when xCAT" ?
2. In the last sentence, should "Example," be "For example" or "Example:" ?

whowutwut: Thanks, I've made the changes in another commit

whowutwut:
@zet809  if you do not have any issues with the default documented behavior to recommend an osimage to be specified , please accept the pull request.

zet809:
Agree to merge.

whowutwut:
This also resolves #1788, I missed that one.

Issue #1752, opened 2016-08-23
bmc tmperary nodes are not deleted after the bmc setup has finished. 


wwzhubj:
issue: bmc temp nodes are not deleted after the bmc setup has finished. 
Yesterday  Dave discovered  5 JEMTs and entered into the xcat db (via  bmcdiscover -s nmap --range 192.168.130.100-140 -t -z -w )
which created the 5 temp nodes
node-8335-gtb-2109c4a 
node-8335-gtb-2109c6a  
node-8335-gtb-2109c8a 
node-8335-gtb-2109e9a  
node-8335-gtb-2109eaa  

After powered up these 5 temp nodes
and they are stucked in a loop waiting to turn the genesis discovery into the permanent node.
  ```
<166>Aug 22 15:17:48 xcat.genesis.dodiscovery: Beginning echo infomation to discovery packet file...
<166>Aug 22 15:17:50 xcat.genesis.dodiscovery: Discovery packet file is ready.
<166>Aug 22 15:17:50 xcat.genesis.dodiscovery: Sending the discovery packet to xCAT (192.168.130.1:3001)...
<166>Aug 22 15:17:50 xcat.genesis.dodiscovery: Sleeping 5 seconds...
<166>Aug 22 15:17:55 xcat.genesis.dodiscovery: Beginning echo infomation to discovery packet file...
<166>Aug 22 15:17:55 xcat.genesis.dodiscovery: Discovery packet file is ready.
<166>Aug 22 15:17:55 xcat.genesis.dodiscovery: Sending the discovery packet to xCAT (192.168.130.1:3001)...
<166>Aug 22 15:17:55 xcat.genesis.dodiscovery: Sleeping 5 seconds...
<166>Aug 22 15:18:00 xcat.genesis.dodiscovery: Beginning echo infomation to discovery packet file...
<166>Aug 22 15:18:02 xcat.genesis.dodiscovery: Discovery packet file is ready.
<166>Aug 22 15:18:02 xcat.genesis.dodiscovery: Sending the discovery packet to xCAT (192.168.130.1:3001)...
<166>Aug 22 15:18:02 xcat.genesis.dodiscovery: Sleeping 5 seconds...
```
```
After investigation, I found there is two conditions. 
1. for some predefined nodes , the switchport is set to "/1/0/32" , which caused the bmc setup can not find the matching nodes.  After I changed switchport  to the "1/0/32" , the bmc setup can continue, and set the node with more parameters. 
[...]

Issue #433, opened 2015-11-19
2.11 build failed with /xpod2man: pods/man1/renergy.1.pod unexpected =item directive in paragraph 309.  ignoring.


tingtli:
ERROR:root:./xpod2man: pods/man1/renergy.1.pod: unexpected =item directive in paragraph 309.  ignoring.
Traceback (most recent call last):
  File "/media/xcat/build/xcat2_autobuild/buildxcat.py", line 133, in main
    buildxCAT(args, logger)
  File "/media/xcat/build/xcat2_autobuild/buildxcat.py", line 55, in buildxCAT
    linuxRPM.buildxCAT(args,logger)
  File "/media/xcat/build/xcat2_autobuild/xcatlib/linux_rpm.py", line 193, in buildxCAT
    raise Exception(err)
Exception: ./xpod2man: pods/man1/renergy.1.pod: unexpected =item directive in paragraph 309.  ignoring.

CRITICAL:root:<== Build ERRORS
FATAL ERROR building branch master, cannot continue

zet809:
Fixed with pull request 434.

tingtli:
Verified and close

PR #431, opened 2015-11-19
add dns document

junxiawang:
add dns document ,@tingtli please help review  and give comments
Link:
http://10.4.30.1/install/html/advanced/dns/index.html

PR #1845, opened 2016-09-15
BMC temp nodes are not deleted after compute node discovered

cxhong:
The discover node definition  created by bmcdiscover command didn't get removed after compute node found and provisioned.  this is issue #1752.

From erTao's suggestion:  replace ``$doreq->({ command => ['rmdef'], arg => [$bmc_node] })`` to ``runcmd()``.   Tested code changes on fs1 environment, worked fine.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/nodediscover.pm
+++ b/xCAT-server/lib/xcat/plugins/nodediscover.pm
@@ -424,7 +424,14 @@ sub process_request {
         if (defined($request->{bmc_node}) and defined($request->{bmc_node}->[0])) {
             my $bmc_node = $request->{bmc_node}->[0];
             syslog("local4|info", "Found node corresponding to BMC=$bmc_node, removing it...");
-            $doreq->({ command => ['rmdef'], arg => [$bmc_node] });
+            my $rmcmd = "rmdef $bmc_node";
+            xCAT::Utils->runcmd($rmcmd, 0);
+            if ($::RUNCMD_RC != 0)
+            {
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/nodediscover.pm
```diff
@@ -424,7 +424,14 @@ sub process_request {
         if (defined($request->{bmc_node}) and defined($request->{bmc_node}->[0])) {
             my $bmc_node = $request->{bmc_node}->[0];
             syslog("local4|info", "Found node corresponding to BMC=$bmc_node, removing it...");
-            $doreq->({ command => ['rmdef'], arg => [$bmc_node] });
+            my $rmcmd = "rmdef $bmc_node";
+            xCAT::Utils->runcmd($rmcmd, 0);
+            if ($::RUNCMD_RC != 0)
```
zet809: Do you think we shall use 'lsdef' to check whether the temporary bmc node is really removed?

zet809:
@cxhong Thx very much for doing this.

xCAT-server/lib/xcat/plugins/nodediscover.pm
cxhong: I think return code message will cover that . line #431 and #433

zet809: OK, I agree

zet809:
Look good to me, I agree to merge. Open a new defect #1862 to trace finding the root cause of the doreq issue.

Issue #1786, opened 2016-08-30
By creating the documentation per tagged release, we have broken old links for xCAT docs


whowutwut:
In the 2.12.2 release, we tagged the release on GitHub. ReadTheDocs will detect it and we enabled that version of the documentation. 

The 2.12 release docs are ALL the same because we only have 2.12 branch in xCAT, so there's no need to have 2.12, 2.12.1, and 2.12.2 docs built on RTD

![image](https://cloud.githubusercontent.com/assets/10049070/18098181/1be4465e-6eaf-11e6-8a7e-b665029d9408.png)

However, by removing the 2.12 docs, we broke all the links for the release notes in the past for 2.12* releases.  

One possible solution here is to just undo the 2.12.2 and just enable 2.12 documentation

whowutwut:
http://xcat-docs.readthedocs.io/en/2.12/index.html <-- this is a broken link, for example

whowutwut:
For the meantime, since I need to provide links for the documentation, I re-enabled the 2.12 doc build.

daniceexi:
@whowutwut 
I did not realize that. My thought was since we have doc change for each minor release so I'd like to make each release has corresponding doc version. For example, even after the release of 2.12.2, the xCAT 2.12.1 still has a doc of 2.12.1 in RTD which is different with the 2.12.2.

What do you think that we enable all the minor versions for a branch like 2.12, 2.12.1, 2.12.2?

whowutwut:
@gurevichmark is looking at this. I think yesterday we decided it might just be better to link just to the documentation and not a specific build of the doc. If we enable each release we will have so many different versions and really that seems pointless since as always ask the customer to go to the latest mod release. And we will keep around the latest in the doc, 2.12.2 right now

daniceexi:
So you meant all the 2.12.x release should use latest doc of the 2.12 branch. I am fine with that if you are fine with the following two facts 

* There are doc differences between releases
* We do have the RTC doc for each release

gurevichmark:
To make release numbers more consistent, I have created a new release 2.12.0 and tagged it to the same commit number as 2.12_release. So now Read The Docs can display documentation for 2.12.2, 2.12.1, 2.12.0, etc. `stable` is still `2.12.2` and `latest` is still `master` as before. I also changed the release notes documentation links to point to the correct URL for that release.
[...]

PR #1847, opened 2016-09-16
Fix doc version in `README.rst`

samveen:
Due to the changes in the release version tag of the documentation from `2.12` to `2.12.0`, the links listed in the README.rst are broken. This MR fixes it. Reference: #1786

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,10 +23,10 @@ Developers and prospective contributors are encouraged to read the `Developers G
 In particular the `GitHub <http://xcat-docs.readthedocs.io/en/latest/developers/github/>`_ related subsection.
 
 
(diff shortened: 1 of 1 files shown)


whowutwut:
Thank you very much!

whowutwut:
@gurevichmark  This fixes the link and is showing the 2.12.0 build.  I'm wondering whether we want to advertise all the builds of the README. (2.12.1, 2.12.2, etc..) 

The downside is that we have to maintain updating these links (adding/removing)...  perhaps a better option is to look for another place we can put the badges where we don't need to continually maintain changes to the README file.

gurevichmark:
To me, keeping the badges in the README feels like an appropriate way to go, even if we have to update the file for each release.

Issue #1671, opened 2016-08-11
mkdef -z does not work when reading redirected input from a file


chrissamuel:
Noticed this for a while, but given it's still in 2.12.1 I thought I'd make a ticket for it:
```
[root@bruce-m tmp]# mkdef -z < rhel68login_updates
Error: No object names were provided.
```
compared to the successful: 
```
[root@bruce-m tmp]# cat rhel68login_updates | mkdef -z
1 object definitions have been created or modified.
```
It's rather counter-intuitive.

Any ideas?

cheers!
Chris

immarvin:
Currently, the mkdef only support the stanza file piped to STDIN,  for the stanza file redirected to STDIN, we can enhance it if required,  

````
 70 else
 71 {
 72     if (-p STDIN) {
 73         while (<STDIN>) { $data .= $_; }
 74         $cmdref->{stdin}->[0] = $data;
 75     }
 76 }
 77
````

samveen:
@chrissamuel  To revisit your original question on the `xcat-users` mailing list:
https://sourceforge.net/p/xcat/mailman/message/34544088/
and the associated caveats
https://sourceforge.net/p/xcat/mailman/message/34544204/

immarvin:
hi @chenglch , would you please take a look at this bug?

thanks

PR #1863, opened 2016-09-20
Add STDIN redirection support for stanza definition

chenglch:
close-issue: #1671


immarvin:
hi @chenglch , with your fix, all the def commands without i/o redirection or pipe will hang

immarvin:
looks good to me

PR #1867, opened 2016-09-20
update rspconfig man page for vlan

penguhyang:
update man page.


zet809:

Issue #1734, opened 2016-08-18
Update the help info for node.mac to know how to handle installnic bond


daniceexi:
xCAT supports to use `bond` for install nics. The discovery process can set the `node.mac` attribute with a special format for `bond` support, but we have not a description for `node.mac` for user to know how to set it manually.

PR #1868, opened 2016-09-20
fix issue 1734: Update the help info for node.mac to know how to hand…

zet809:
…le installnic bond
Fix issue #1734 

For a Compute Node that have multiple nic connected to the Management Network, those NICs may be configured as a bond interface after deployed host OS. 
xCAT need to provide a method to have those NICs get the same IP address and point to the same boot configuration file.
Currently, for the mac attributes set as 'mac1|mac2!hostname2', the meaning is, the mac1 have the default hostname equal to nodename, and mac2 will use hostname2.
If hostname2 is alias of the nodename, when running makedhcp, there will be 2 entries created within dhcp.lease file. One entry name is nodename, and the other one will be hostname2. Both of those 2 entries will have the same ip address and boot configuration file configured. 

This pull request is planed to explain how to configure mac1!hostname1|mac2!hostname2 pair.


daniceexi:
@zet809 Sorry, I cannot follow your message. So you need to refine it.

whowutwut:
@zet809  can you refine the description, I will assign to @cxhong  for review

whowutwut:
@zet809 please add some more description that will help to understand the purpose of this pull request.

zet809:
@daniceexi @cxhong Hi, xiao peng and casandra, I have updated the description of the mac attribute, pls help to review. Thx!

perl-xCAT/xCAT/Schema.pm
```diff
@@ -472,7 +472,7 @@ passed as argument rather than by table value',
...
+            mac => 'The mac address or addresses for which xCAT will manage static bindings for this node.  This may be simply a mac address, which would be bound to the node name (such as "01:02:03:04:05:0E").  This may also be a "|" delimited string of "mac address!hostname" format (such as "01:02:03:04:05:0E!node5|01:02:03:04:05:0F!node6-eth1"). If there are multiple nics connected to Management Network, in order to make sure the OS deployment finished successfully, the macs of those nics must be able to resolve to same IP address. First, users have to create alias for this node for each mac in this network through either 1. adding the alias into /etc/hosts for the node directly or 2. setting the alias to the "hostnames" attribute and then run "makehost" against the node. Then, configure the "mac" attribute of the node like "mac address1!node alias 1|mac address2|node alias 2". For the first mac address set in "mac" attribute, do not need to set a "node alias" string for it since the nodename of the node will be used for it automatically.',
```
daniceexi: If there are multiple nics connected to Management Network(usually for bond), in order to make sure the OS deployment finished successfully, the macs of those nics must be able to resolve to same IP address. First, users have to create alias for each mac in the Management Network through either: 1. adding the alias into /etc/hosts for the node directly or: 2. setting the alias to the "hostnames" attribute and then run "makehost" against the node. Then, configure the "mac" attribute of the node like "mac1!node|mac2!node-alias". For the first mac address (mac1 in the example) set in "mac" attribute, do not need to set a "node name" string for it since the nodename of the node will be used for it by default.'

daniceexi: I refine your change as I pasted above. You can refer it.

zet809:
@daniceexi Hi, Xiao peng, modified, pls help to review.

daniceexi:

Issue #1797, opened 2016-09-01
Using the --template option to create a new osimage definition for netboot does not create a new rootimgdir


whowutwut:
I tried the --template option to create a new osimage definition from a netboot (diskless) osimage, and the result does not create a new `rootimgdir` attribute.  It re-uses the image directory from the template.  

I think this is very dangerous for the new user to xCAT who is trying to duplicate diskless osimages and we should provide some protection.   If a osimgae is copied using this feature, running `genimage` will wipe out the template osimage. 

```
[root@fs1 ~]# mkdef -t osimage -o victor-beta-rhels7.3-netboot --template rhels7.3-beta-ppc64le-netboot-compute
1 object definitions have been created or modified.
```

The resulting osimage uses the `rootimgdir` from the template: 
```
[root@fs1 ~]# lsdef -t osimage -o victor-beta-rhels7.3-netboot -i rootimgdir
Object name: victor-beta-rhels7.3-netboot
    rootimgdir=/install/netboot/rhels7.3-beta/ppc64le/compute
```

Suggested solutions here: 
1) Use the name provided to create a new path for the `rootimgdir` 
2) Add a option to the mkdef command to allow the user to specify a "short name" that will be used.  If we have this option, then we block the mkdef command to not successfully complete if the template is a "netboot" osimage and the `--name` is not specified
3) Print a warning message to inform the user to create a new `rootimgdir` and chdef the osimage to use the new directory.

daniceexi:
@whowutwut 
It does have the issue as you mentioned. But the current implementation also has some goodness, 
1. From user, it's consistent for mkdef that mkdef won't do any specific operation. If we make mkdef handle the **rootimgdir** in a specific way, user might worry is there any other attributes have special handling?
2. For code, it's consistent that no special code logic for specific attribute.

whowutwut:
Sure, if we want to keep the code changes to a minimium, then maybe suggest a warning message if the template contains a `rootimgdir` attribute , such as: 
```
[root@fs1 ~]# mkdef -t osimage -o victor-beta-rhels7.3-netboot --template rhels7.3-beta-ppc64le-netboot-compute
WARNING: The rootimgdir attribute is copied from the template.   Change this value before running genimage to avoid overwriting the diskless image for 'rhels7.3-beta-ppc64le-netboot-compute' definition.
1 object definitions have been created or modified.

immarvin:
[...]

PR #1874, opened 2016-09-21
Probe for osimage verification

gurevichmark:
This pull request creates a new `osimagecheck` probe verify no 2 osimages have the same rootimgdir. 
It answers a problem raised by issue #1797

### DIFF
--- a/xCAT-probe/subcmds/osimagecheck
+++ b/xCAT-probe/subcmds/osimagecheck
@@ -1,0 +1,155 @@
+#! /usr/bin/perl
+# IBM(c) 2016 EPL license http://www.eclipse.org/legal/epl-v10.html
+
+BEGIN { $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : -d '/opt/xcat' ? '/opt/xcat' : '/usr'; }
+
+use lib "$::XCATROOT/probe/lib/perl";
+use probe_utils;
+use File::Basename;
+use Net::Ping;
+use Getopt::Long qw(:config no_ignore_case);
+
+use Data::Dumper;
+use warnings;
+
+my $program_name = basename("$0");
+my $help;
+my $noderange = "";
+my $test;
+my $output  = "stdout";
+my $verbose = 0;
+my $rst     = 0;
+
+$::USAGE = "Usage:
+    $program_name -h
+    $program_name -c {duplicate_rootimgdir|valid_osimage_attributes} [-V|--verbose]
+
+Description:
+    Use this command to check osimage defintions in xCAT DB. 
+
+Options:
+    -h : Get usage information of $program_name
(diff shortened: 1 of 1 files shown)


gurevichmark:
Output when duplicate rootimgdir found:
```
[root@fs1 subcmds]# xcatprobe osimagecheck -c duplicate_rootimgdir
Duplicate found for osimage with  rootimgdir=/install/netboot/ubuntu14.04.3/ppc64el/compute :                     [WARN]
    markg2 and markg
Duplicate found for osimage with  rootimgdir=/install/netboot/rhels7.3snapshot4/ppc64le/compute :                 [WARN]
    rhels7.3snapshot4-ppc64le-netboot-compute and markg3
Duplicate found for osimage with  rootimgdir=/install/netboot/ubuntu14.04.3/ppc64el/compute :                     [WARN]
    ubuntu14.04.3-ppc64el-netboot-compute and markg
[root@fs1 subcmds]#
```

Output when no duplicates found:
```
[root@fs1 subcmds]# xcatprobe osimagecheck -c duplicate_rootimgdir
No osimages with duplicate rootimgdir were found.                                                                 [ OK ]
[root@fs1 subcmds]#
```

whowutwut:
Can we eliminate the `-c duplicate_rootimgdir` flag so it just runs all the checks by default for now, until more function is added

gurevichmark:
@whowutwut Changes mode. All tests will be performed one after another.

xCAT-probe/subcmds/osimagecheck
```diff
@@ -0,0 +1,133 @@
...
+    # After processing all the expected flags and arguments,
+    # there is still left over stuff on the command line
+    probe_utils->send_msg("$output", "f", "Invalid flag or parameter: @ARGV");
+    probe_utils->send_msg("$output", "d", "$::USAGE");
+    exit 1;
+}
+
+# Run all osinage probe checks one after another
+$rc = check_for_duplicate_rootimgdir();
+$rc = check_for_valid_osimage_attributes();
```
whowutwut: This will overwrite a bad return code  and return 0.  Should we initialize $rc to 0, and only set it to non zero returns from the functions?

gurevichmark: Yea. I have set these return codes for some future use. Right now, I am not sure what is the proper way to set them. If first one fails, but second succeeds, is overall rc fail or success ? If first one fail, do we keep going or exit right there ? Do we even need exit rc for probe ?

gurevichmark:
@whowutwut Removed the RCs for each probe step.

whowutwut:
Let's merge this to get it into the 2.12.3 release.  The only other comments that I would have is 
* maybe we should define the sub routines above and have the "main" at the bottom
* there are some other return code variables that we could also eliminate the next time we touch this file

PR #1879, opened 2016-09-22
Update bundle files for sles and rhels with hierarchy cases

caomengmeng:
@tingtli 
please help review bundle files changes.


tingtli:
i agree to merge, let's take a look at tomorrow's test result.

PR #1883, opened 2016-09-23
Support Predefine switch has same hostname as discovery

cxhong:
#for issue 1878

If discovery switch hostname has same name with predefine switch name, the switchdiscovery --setup command will be failed and some of predefine switch's attributes will be removed.


cxhong:
@zet809 , can u review this?  Thanks.

zet809:
@cxhong I will merge the code first to not block the release candidate build. For the return code, we can open a defect to trace it in the next release.

Issue #1882, opened 2016-09-23
todays development version of xcat has a dramatically longer netboot time.


ralphbellofatto:
There is a dramatic increase in the time it takes to decompress the rootimg.cpio.gz with the development branch of xcat:

lsxcatd -v
Version 2.12.3 (git commit ee11ecddb7e3c323b11857826a072303ae5a7b20, built Thu Sep 22 06:16:00 EDT 2016)

we recently changed from the "latest" xcat core code to the development code to pick up some recent changes.

previous version:
```
	==> xCAT-latest/xcat-core.repo <==
	[xcat-core]
	name=xCAT 2 Core packages
	baseurl=http://xcat.org/files/xcat/repos/yum/latest/xcat-core
	enabled=1
	gpgcheck=1
	gpgkey=http://xcat.org/files/xcat/repos/yum/latest/xcat-core/repodata/repomd.xml.key
	
	==> xCAT-latest/xcat-dep.repo <==
	[xcat-dep]
	name=xCAT 2 depedencies
	baseurl=http://xcat.org/files/xcat/repos/yum/xcat-dep/rh7/ppc64le
	enabled=1
	gpgcheck=1
	gpgkey=http://xcat.org/files/xcat/repos/yum/xcat-dep/rh7/ppc64le/repodata/repomd.xml.key
```

and the new version:
[root@c460mgt01 yum.repos.d]# head xCAT-latest/*.repo
```
	==> xCAT-latest/xcat-core.repo <==
	[xcat-core]
	name=xCAT 2 Core packages
	baseurl=http://xcat.org/files/xcat/repos/yum/latest/xcat-core
	enabled=1
	gpgcheck=1
	gpgkey=http://xcat.org/files/xcat/repos/yum/latest/xcat-core/repodata/repomd.xml.key
	
	==> xCAT-latest/xcat-dep.repo <==
	[xcat-dep]
	name=xCAT 2 depedencies
	baseurl=http://xcat.org/files/xcat/repos/yum/xcat-dep/rh7/ppc64le
	enabled=1
	gpgcheck=1
	gpgkey=http://xcat.org/files/xcat/repos/yum/xcat-dep/rh7/ppc64le/repodata/repomd.xml.key
```

We noticed a dramatic increase in the boot time of our production network boot image, due to the time it takes to decompress the rootimg.cpio.gz file.

If we boot a minimal image based on rhels7.2-ppc64le-netboot-compute. we see the following in the console log:

```
[...]

PR #1884, opened 2016-09-23
fix todays development version of xcat has a dramatically longer netboot time. #1882

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1882:

add "-nv" option to suppress the verbose output of wget while downloading rootimg.gz inside the diskless initrd


neo954:
Simple patch. Looks good. Will merge this one.

daniceexi:
👍

Issue #1857, opened 2016-09-19
netboot postscripts have no knowledge about the osimage it is installing.


ralphbellofatto:
We have a need to bake somethings in the image created in the directory that come from the site table as well as the osimage definition itself.

for example we need to know the address of the management node as well as the ospkgdir and otherpkgdir specification so we can do some additional setup in the postinstall scripts to bake somethings into the image.

Currently there is nothing on the command line given to the postinstall scripts or its environment that tells us what is available.

Here is the current information we have available:

Paratemers to this script:
```
installroot=/install/netboot/rhels7.2/ppc64le/test-postscript-nb/rootimg
osver=rhels7.2
arch=ppc64le
profile=compute
workdir=
```

environment:
```
MANPATH=/opt/xcat/share/man:
DB2INSTANCE=xcatdb
PERL_BADLANG=0
LC_ALL=C
XCATROOT=/opt/xcat
PATH=/opt/xcat/bin:/opt/xcat/sbin:/opt/xcat/share/xcat/tools:/sbin:/usr/sbin:/bin:/usr/bin
PWD=/opt/xcat/share/xcat/netboot/rh
LANG=en_US.UTF-8
EXTSHM=ON
SHLVL=3
_=/bin/env
```

To duplicate this issue we did first copied the osimage rhels7.2-ppc64le-netboot-compute to  test-postscript-nb

And then did the following:
```
copy  postinstall script and display environment variables and parameters...


chdef -t osimage test-postscript-nb postinstall=/install/postscripts/custom/test-postscript-nb/compute.postinstall


[root@c460mgt01 tools]# cat >/install/postscripts/custom/test-postscript-nb <<EOF
#!/bin/sh
installroot=$1
osver=$2
arch=$3
profile=$4
workdir=$5

echo "=================================================="
echo installroot=$installroot
echo osver=$osver
echo arch=$arch
echo profile=$profile
echo workdir=$workdir
echo "environment:"
echo "--------------------------------------------------"
env

echo "=================================================="
EOF

chmod  +x /install/postscripts/custom/test-postscript-nb

genimage  test-postscript-nb 2>&1 | tee /tmp/ genimage. test-postscript-nb.log
```

immarvin:
hi @ralphbellofatto , we will try to pass all the specified attributes of the osimage definition via command line arguments or environment variables. 

[...]

PR #1885, opened 2016-09-23
add IMG_NAME into the environment variables to postinstall scripts

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1857


neo954:
This patch looks good. I will merge this one.

PR #1891, opened 2016-09-25
Fix xCAT probe rpm install failed problem

hu-weihua:
Fix xCAT probe rpm install failed problem

    --> Finished Dependency Resolution
    Error: Package: 4:xCAT-probe-2.12.3-snap201609230616.noarch (xcat-2-core)
           Requires: perl(probe_global_constant)
    You could try using --skip-broken to work around the problem
    You could try running: rpm -Va --nofiles --nodigest


immarvin:
agree to merge

daniceexi:
@hu-weihua @immarvin 

I am very appreciated for your responsibility to fix the automation issue in weekend. 👍

Issue #1730, opened 2016-08-18
[FVT]: sed: can't read /boot/grub/grub.conf in servicenode log file


tingtli:
I remember we ever had this issue in CN's log when we support rh7.2 and sles12.1 on flat cluster.   Please double check the code and give a patch for servicenode Thanks!

daniceexi:
The issue caused by the following piece of code. I don't think we have fix for this issue. 
Since this issue was there for a long time. We can move it to next release.
@immarvin when fix this issue, please add comments why we need change the grub.conf?

./xCAT-server/share/xcat/install/scripts/post.xcat
~~~~
436 else
437     sed -i 's/^serial/#serial/' /boot/grub/grub.conf
438     sed -i 's/^terminal/#terminal/' /boot/grub/grub.conf
439     if [ "$XCATDEBUGMODE" = "1" ] || [ "$XCATDEBUGMODE" = "2" ]; then
440        msgutil_r "$MASTER_IP" "debug" "/boot/grub/grub.conf updated" "/var/log/xcat/xcat.log"
441     fi
442 fi
~~~~

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1897/files

immarvin:
will add the reason for this 2 line later

tingtli:
I've verified on 2.12.4 on rhels7.2
verified and close this one.

PR #1897, opened 2016-09-27
fix issue [FVT]: sed: can't read /boot/grub/grub.conf in servicenode log file #1730

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/1730


penguhyang:
looks good, agree to merge.

Issue #1828, opened 2016-09-13
xCAT not start from hard disk after provsioned statefull node


bhbai:
Env: 

> lsxcatd -a
> Version 2.12 (git commit b33cb213ce79f91621cd65932c9798c735b74bc4, built Fri Jun 24 03:17:21 EDT 2016)
> This is a Management Node
> cfgloc=Pg:dbname=xcatdb;host=10.45.5.5|xcatadm
> dbengine=Pg
> dbname=xcatdb
> dbhost=10.45.5.5
> dbadmin=xcatadm
> [root@cit1055 ~]# lsb_release -a
> LSB Version:    :core-4.1-noarch:core-4.1-ppc64le
> Distributor ID: RedHatEnterpriseServer
> Description:    Red Hat Enterprise Linux Server release 7.2 (Maipo)
> Release:        7.2
> Codename:       Maipo
>  uname --a
> Linux  3.10.0-327.el7.ppc64le #1 SMP Thu Oct 29 17:31:13 EDT 2015 ppc64le ppc64le ppc64le GNU/Linux

Expected result: xCAT must/should boot from hard disk automatically while provision with statefull image.

Detail steps as following:
```
1. Generate discovered_stanzas:
#bmcdiscover --range 9.42.xx.xx -z

2. Add mtm/serial number in stanzas files(If Neccessery)

3. Import <nodecn>
#cat /tmp/cf_predefined_stanzas | mkdef -z -f 
#makehosts <nodecn>

3. Add some <nodecn> attributes
#chdef <nodecn> groups=__Managed,__NetworkProfile_network,__ImageProfile_rhels7.2-ppc64le-statefull-compute,__HardwareProfile_IBM_PowerNV,compute
#chdef <nodecn> bmcpassword=test
#<nodecn> power reset
#makedns -n

4.Scan until get <nodecn> mac
#lsdef <nodecn> | grep mac= | wc -l

5.Cleanup imported node
#nodepurge <nodecn>
#service dhcpd restart

6.Generate bmc format import file as following:
<nodecn>:
	mac=00:90:fa:b9:xx:xx
	nicips=bmc!9.42.xx.xx

7.Import nodecn
nodeimport file=/cf_hostinfo_file networkprofile=fsp_network imageprofile=rhels7.2-ppc64le-statefull-compute hostnameformat=gssio#NNN hardwareprofile=IBM_PowerNV

8.Add some <nodecn> attributes
#chdef <nodecn> bmcpassword=test
[...]

PR #1900, opened 2016-09-27
fix issue 1828: xCAT not start from hard disk after provsioned statef…

zet809:
…ull node

For issue #1828


daniceexi:
👍

bhbai:
After add postscripts/cleanup_nvm.sh in SCF, this issue not rise up again.

Issue #1813, opened 2016-09-09
renergy for Power 7 hardware does not work in xcat 2.12 on redhat 2.7


billwaj:
The xcat 2.12.2 renergy man page states that it supports older Power 7 hardware which should use the PPCenergy.pm code to ultimately use the xCAT-pEnergy rpm to obtain the specified values.  However, PPCenergy.pm is not used and energy.pm is used and has errors with some values.

Is older Power 7 and Power 6 hardware still supported in xCAT 2.12 ?

whowutwut:
@zet809  could you take a look at this?

daniceexi:
@billwaj The energy management for none-OPAL is supported in 2.12.2.  
In xCAT code, we need a way to use the machine type to figure out whether it's a OPAL server. So we have a list for all none-OPAL server. The ones which are not in the list are considered as OPAL.

So I guess you problem was you run renergy against a non-OPAL server which is not in the **non-OPAL** list.

Then what's the MTMS of your server?

billwaj:
The machine type and model of the power 7 I tested for was 9125-F2C.

daniceexi:
@zet809 could you check whether the 9125-F2c has been added in the power6 list. In certain subroutine in the Utils.pm.

zet809:
It seems the MTM 9125-F2C is not put in the power6 list. 
@billwaj Hi, bill, the hash below include the machine type/model for p6 and p7 servers, will you pls add MTM for your servers to have a try? If works, pls tell us all the MTM you'd like to add into the list and then we can check them in xcat. Thx!

# the MTM of P6 and P7 machine
my %MTM_P6P7 = (
    # P6 systems
    'xxxx-xxx' => 1,
    # P7 systems
    'xxxx-xxx' => 1,
);

zet809:
@billwaj Hi, Bill, any update?

billwaj:
Yes and thanks, adding 9125-F2C to %MTM_P6P7 in Utils.pm now successfully calls the PPCenergy plugin  which in turn calls xCAT_cim_client.


One problem I am now working through is that xCAT_cim_client is passed in the hostname of the specified FSP and not the ip address as was done in past xCAT release.  This is causing an error which I am now investigating.


________________________________

daniceexi:
@billwaj 
Besides ``9125-F2C``, are there any other server types you want to add in the P6P7 support list?

billwaj:
[...]

PR #1908, opened 2016-09-28
fix issue 1813, add 9125-f2c in mtm hash

xuweibj:
#1813 

9125-F2C is Power 7, add it in MTM_P6P7 hash.


zet809:
Seems OK to me, I agree to merge.

Issue #1910, opened 2016-09-28
[fvt]2.12.3 install_lsf need to modify the  lsb.hosts script


junxiawang:

junxiawang:
LSB_HOSTS_FILE="$LSF_TOP/conf/lsbatch/$LSF_CLUSTER_NAME/configdir/lsb.hosts"
  LSB_HOSTS_FILE_ORIG="$LSF_TOP/conf/lsbatch/$LSF_CLUSTER_NAME/configdir/lsb.hosts.orig"
 mv $LSB_HOSTS_FILE $LSB_HOSTS_FILE_ORIG

 PRIMARY_MASTER_NODE=`echo $LSF_MASTER_LIST |  awk '{print $1}'`
     echo "Begin Host" >> lsb.hosts
   ------------------------> need to point the file's directory

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1911/files

shensisi:
We saw a problem in xcat postscript install_lsf.   We used this postscript to install and configure LSF.  After LSF is up, we saw ERROR in LSF configuration file lsf.hosts.     The section header “Begin Host” was missing in lsf.hosts which caused LSF could not get the correct  MXJ (maximum number of jobs).

whowutwut:
@shensisi  does the fix that @immarvin  references resolve the issue that you saw?

PR #1911, opened 2016-09-28
fix [fvt]2.12.3 install_lsf need to modify the lsb.hosts script #1910

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/1910


zet809:
I agree to merge.

Issue #1906, opened 2016-09-28
[fvt]2.12.3switch based switch discovery could not support BNT switch


junxiawang:
env:fs1
build:latest build
```
[root@fs1 ~]# switchdiscover --range 192.168.5.41 -s snmp --setup
Discovering switches using snmpwalk for 192.168.5.41 .....
ip          	name                	vendor                                            	mac         
------------	------------        	------------                                      	------------
192.168.5.41	switch-192-168-5-41 	IBM Networking Operating System RackSwitch G8264-T	6c:ae:8b:d1:82:fe
the switch type Not support is not support yet

[root@fs1 ~]#  switchdiscover --range 192.168.5.163 -s snmp
Discovering switches using snmpwalk for 192.168.5.163 .....
 No switch found 
[root@fs1 ~]# 
```

PR #1917, opened 2016-09-28
Add switch matching string from vendor to switchtype

cxhong:
fix issue #1906
Different BNT switch model has different vendor information output.
Add "RackSwitch" to identify BNT switch for the model G8264-T and G8124


cxhong:
@zet809 can u review this?

zet809:

Issue #1881, opened 2016-09-22
Minsky boots genesis kernel when mutliple interfaces have dhcp dynamic ranges.


markprez:
xCAT's default response to a dhcp bootp request is to boot the genesis kernel.
If you have multiple interfaces service dhcp dynamic responses but only one setup to provision and run distributed commands to the nodes the default on the other network will be genesis boot.
To dedicate the shared BMC port to BMC traffic only, we need to prevent dhcp responses to the OS side of the shared connection (enP_f0).
If xCAT is setup to install on the enP_f1 OS network:
Object name: c712garrisonBMC
    gateway=10.128.130.1
   dynamicrange=10.128.130.100-10.128.130.140
    mask=255.255.255.0
    mgtifname=enP4p1s0f0
    net=10.128.130.0
    tftpserver=10.128.130.1
Object name: c712garrisonOS
    dynamicrange=192.168.130.100-192.168.130.140
    mask=255.255.255.0
    mgtifname=enP4p1s0f1
    net=192.168.130.0
    tftpserver=192.168.130.1
*
tabdump site | grep dhcpinterfaces
"dhcpinterfaces","enP4p1s0f0,enP4p1s0f1",,
*
BMC requests will be handled across the BMC network, but so will enP*f0 requests.
This is a requirement to prevent xCAT from trying to provision the node on the shared network.
Er Tao recommended removing /tftpboot/pxelinux.cfg/p/10.130.130.0_24 file before provisioning the nodes, this will prevent xCAT from sending the genesis kernel.

cxhong:
Ralph suggested we should modify the dhcpd.conf too .

daniceexi:
We may enhance the ``dhcpinterfaces`` to specify which nic cannot ``boot`` by adding a pending ``noboot`` keyword like following. If this is configured, we won't add the genesis load entry in the dhcpd.conf for this nic.

~~~~
"dhcpinterfaces","enP4p1s0f0!noboot,enP4p1s0f1",,
~~~~

ralphbellofatto:
That looks like a good way to express this.....

What is the time line for getting this implemented...

ralphbellofatto:
I have been talking to our engineers who are successfully setting up garisions with dual cables, I think xcat already handles this case.

The trick is to place each of the cables on connected to the garison's on a separate vlan, such that untagged traffic.  This way the untagged traffic gets directed to different ip addresses on the managemt node depending on what vlan was assigned to the switch.

[...]

PR #1930, opened 2016-09-30
Provide a 'noboot' option to disable genesis kernel on specific dhcp interfaces

zet809:
…s have dhcp dynamic ranges

Fix issue #1881 
The modification include 2 modifications:
1. in mknb.pm, rename the default petitboot configuration file for the noboot nic.
2. in dhcp.pm, ignore the noboot keywords within site.dhcpinterfaces attibute.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/mknb.pm
+++ b/xCAT-server/lib/xcat/plugins/mknb.pm
@@ -18,6 +18,7 @@ sub process_request {
     my $serialport;
     my $serialspeed;
     my $serialflow;
+    my %nobootnicips = ();
     my $initrd_file = undef;
     my $xcatdport   = 3001;
     my @entries     = xCAT::TableUtils->get_site_attribute("defserialport");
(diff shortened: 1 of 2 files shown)


whowutwut:
@zet809  Some general comments: 

* can you add the externals into the description for this pull request so the knowledge of the rest of the community. Ex: How do we enable it, how do we disable it? 

* We need some changes done for the dhcpinterfaces description in our documentation and associated man pages.  

* Have we tried this out for service nodes, does it propagate down correctly.

xCAT-server/lib/xcat/plugins/mknb.pm
```diff
@@ -207,7 +243,14 @@ sub process_request {
     my $dopxe = 0;
     foreach (keys %{$normnets}) {
         my $net = $_;
+        my $nicip = $normnets->{$net};
         $net =~ s/\//_/;
+        if (defined($nobootnicips{$nicip})) {
+            if ($arch =~ /ppc/ and -r "$tftpdir/pxelinux.cfg/p/$net") {
+                rename("$tftpdir/pxelinux.cfg/p/$net", "$tftpdir/pxelinux.cfg/p/noboot-$net");
```
whowutwut: If `:noboot` is set for a network, running `mknb` will disable this file by renaming.  If we remove the  `:noboot` attribute, a new file is created by `mknb` , reactivating it.  But the old file is not removed, so the `noboot-NETWORK` file is left around.  

Is it better to rename this file, or simply delete it?  Maybe deleting would avoid confusion...

whowutwut:
@zet809  I don't think we can merge this pull request just yet.  Was doing some testing with this change on the Minsky nodes with 2 wires connected, and running HW discovery.   With the change , the dhcpd.conf file does not contain anything about the network that is set with `:noboot` so that interface doesn't give out any DHCP IP addresses.  

It is possible that on the dev machine, we have some other code change that is not included in this pull request.   Need to reinstall the mgmt node clean to be sure.

xCAT-server/lib/xcat/plugins/mknb.pm
whowutwut: @zet809  Why do you choose to rename this instead of delete? 
The issue comes when the user removes the `:noboot` on the interface and rerun mkbn.  Now there are two files in the directory: `/tftpboot/pxelinux.cfg/p` 
```
noboot-50.0.0.0_8
50.0.0.0_8
```

I think this is confusing.

whowutwut:
Waiting for a response for the question about rename vs remove, otherwise I'm OK with this changes and will merge soon...   Retested and seems to behave as expected, no changes are made to the dhcpd.conf file.  Not sure what happened the first time.

zet809:
@whowutwut Hi, Victor, I have modified the code to use 'unlink' to remove the default petitboot configuration file for the "noboot" interface, pls help to review.

whowutwut:
Thanks @zet809  I think this is good for now, let's get it merged so we can get some teams to exercise it. 

I think we need to have some documentation in the hardware discovery section to expose this. Will see if I can do something about that today.

PR #1933, opened 2016-10-01
Add support for a default MTU on a network

mattaezell:
Add a MTU column to the networks table.  If set, this will configure the MTU in
DHCP so that nodes will have the correct MTU during provisioning.  This is
important for large systems to enable jumbo frames to reduce CPU usage when
copying diskless images over the network (before the confignics postscript can
run to set the desired MTU).


mattaezell:
I don't have an AIX system to test this on, so AIX support is best-guess.  I'm hoping someone on the xCAT team can validate that part.

docs/source/guides/admin-guides/references/man5/networks.5.rst
```diff
@@ -136,6 +136,12 @@ networks Attributes:
  
 
 
+\ **mtu**\ 
+ 
+ The default MTU on the network (ex. 1500 or 9000).
```
whowutwut: @mattaezell  The man pages are "auto-generated" using the `create_man_pages.py` script sitting in xcat-core directory.  The man pages are created from various individual .pod files and the Schema.pm file. 

So if you want this change, you should put this text at line 562 in Schema.pm and run the `create_man_pages.py` script, which will auto generate the networks.5.rst doc file.  Otherwise when someone else runs the script, it will get overwritten. 

Here's some documentation about it: http://xcat-docs.readthedocs.io/en/stable/developers/guides/code/builds.html?#man-pages

docs/source/guides/admin-guides/references/man7/network.7.rst
```diff
@@ -145,6 +145,12 @@ network Attributes:
  
 
 
+\ **mtu**\  (networks.mtu)
+ 
+ The default MTU for the network.
```
whowutwut: Same comment as with networks.5.rst, running the `create_man_pages.py` script will overwrite this comment.

whowutwut:
@mattaezell Thanks for this, we are reviewing the changes

mattaezell:
Thanks @whowutwut, I didn't know about the automatic man page generation.  I added a second commit that is the result of running create_man_pages.py, so (if this is merged) someone running create_man_pages.py won't see unanticipated changes.

whowutwut:
Thanks Matt,  I've reviewed, will merge the changes in.

PR #1945, opened 2016-10-07
Mellanox IB documentation clarification for OFED Installation Step

whowutwut:
The documented mellanox installation process is quite complicated and it's a very difficult section to follow. 

The changes in this pull request remove the references to the v1 version of the installation script. Since we have version control in the documentation now, we don't need to reference things that are obsolete. 

The other attempt made is to only document the recommended process with real examples that can be copy-pasted by the users. I know we have lots of ways to do the same thing, but we should not document every possibility but to make it very clear how to proceed with the recommended path.


whowutwut:
I'd like to get this pull request merged in ASAP so we can reference  the new documentation and exercise the steps to make sure they work well.

hu-weihua:
@whowutwut, thanks for all you have done about refining the IB document. I have reviewed the change and agree to merge.  thanks again.

Issue #1816, opened 2016-09-10
[CORAL-EA]: Mellanox IB installer cannot support multiple kernels on RHEL 7.x


whowutwut:
In the hack_uname function, it doesn't support multiple kernels correctly.  Introduced to fix Issue #316   The `uname -r` command looks in the `/lib/modules` directory but only echos 1 kernel back.  If there are more, it will pick the wrong one and fail when `uname -m` is called because it takes the output of `uname -r` and passes it to rpm.  

We will get a message like:
```
# uname -m
package kernel-3.10.0-327.el7.ppc64le is not installed
```

The other issue found by LBS team is that the line 150, ` ppc64el vs ppc64le`

neo954:
@whowutwut, if there are multiple linux kernel installed on the same machine, the `hack_uname()` will choose the latest one, which is the one has larger version number. There is no way to guess which is the right one. The administrator just know which is the right one in his/her mind. There is no way to guess it properly.

neo954:
@whowutwut, on a ppc64le system, `uname -m` always show `ppc64le`, no matter on RHEL or Ubuntu. There is no chance of it to show `ppc64el`. The function `hack_uname` just mimic the behavior of the real `uname` command.

hu-weihua:
Hi @whowutwut , how does ``hack_uname()`` work just as @neo954 explained above, we can't anticipate the right kernel version unless administrator tell us. If we must know some information from administrator before running IB installer, we have to offer a command option to administrator to input such information.  But it is a real good option?  xCAT really can support multiple kernel well?  I think xCAT install the latest kernel during provision too.

Could you tell me where  LSB team get ``ppc64el``? May I know their environment information?

whowutwut:
@neo954 @hu-weihua  If you look at our code for https://github.com/xcat2/xcat-core/blob/master/xCAT-server/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2 , at line 150, we have the following code...

```
function hack_uname()
{
...
		"-m")
			ARCH="\$(dpkg --print-architecture 2>/dev/null || rpm -q kernel-\$("\$0" -r) --qf '%{arch}' 2>/dev/null)"
			case "\$ARCH" in
			"amd64")
				ARCH="x86_64"
				;;
			"ppc64el")
			        ARCH="ppc64le"
				;;
...
...
}
```

Was that a typo in our code for the case statement?  `ppc64el`

whowutwut:
[...]

PR #1953, opened 2016-10-11
Update the hackuname function

mattaezell:
There are currently 2 different ways that uname is being faked;
combine both use cases.  Use the kernelver provided by the user
instead of attempting to detect it

### DIFF
--- a/xCAT-server/share/xcat/netboot/rh/genimage
+++ b/xCAT-server/share/xcat/netboot/rh/genimage
@@ -467,7 +467,7 @@ unless ($onlyinitrd) {
         #        }
 
         # Hack uname when deal otherpkgs
-        use_hackuname();
+        use_hackuname($arch, $kernelver);
         use_devnull();
 
 
@@ -713,31 +713,7 @@ unless ($imagename) {
 
 if ($postinstall_filename) {
 
-    #For Mellonax IB script. In diskless image, the uname -r not returning the rootimg level,
-    #because the "uname -r" only returns the version of the kernel in use
-    #create a temporary uname script. for every flag except for -r, it should just call the real
-    #uname with the same flags and return that info.
-    if (!(-e "$rootimg_dir/bin/orig_uname")) {
-        system("mv $rootimg_dir/bin/uname $rootimg_dir/bin/orig_uname");
-    }
-    my $tmpuname;
-    open($tmpuname, ">", "$rootimg_dir/bin/uname");
-    print $tmpuname <<EOS_UNAME;
-#!/bin/sh
-
-if [[ \$\# -eq 1 && \$1 == "-r" ]]; then
-    echo $kernelver
-    exit 0
-fi
-
-res=`/bin/orig_uname \$\*`
-echo \$res
-
-EOS_UNAME
-
-    close($tmpuname);
-    system("chmod +x $rootimg_dir/bin/uname");
-
+    use_hackuname($arch, $kernelver);
 
     $updates{'postinstall'} = $postinstall_filename if ($tempfile);
 
@@ -756,12 +732,14 @@ EOS_UNAME
     foreach my $postinstall (split /,/, $postinstall_filename) {
         if (!-x $postinstall) {
             print "postinstall script $postinstall is not executable\n";
+            unuse_hackuname();
             exit 1;
         }
 
         my $rc = system($postinstall, $rootimg_dir, $osver, $arch, $profile);
         if ($rc) {
             print "postinstall script $postinstall failed\n";
+            unuse_hackuname();
             exit 1;
         }
     }
@@ -769,9 +747,7 @@ EOS_UNAME
     #delete the osimage attributes from environment variable
     delete @ENV{qw(IMG_ARCH IMG_NAME IMG_OSVER IMG_PROFILE IMG_PKGLIST IMG_PKGDIR IMG_OTHERPKGLIST IMG_OTHERPKGDIR IMG_ROOTIMGDIR)};
 
-    # restore the orig uname
-    system("mv $rootimg_dir/bin/orig_uname $rootimg_dir/bin/uname");
-
+    unuse_hackuname();
 }
 
 # all the attributes have been gathered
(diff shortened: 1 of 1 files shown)


mattaezell:
We ran into an issue where a DKMS rpm being installed in otherpkgs couldn't build.  I found the following in the output:

> uname -m = package kernel-3.10.0-327.el7.ppc64le is not installed
> uname -r = 3.10.0-327.el7.ppc64le
> uname -s = Linux
> uname -v =

This code:
`cd /lib/modules && for d in * ; do : ; done && echo $d`
is not very reliable because it relies on alphanumeric sorting.  At minimum, this should use **sort -V**, but in reality we already know the kernel version.

whowutwut:
@mattaezell  Thanks!  I think this resolves the issues we hit with multiple kernels in Issue #1816 

@hu-weihua @neo954  Can you please review this change?

hu-weihua:
@mattaezell , thanks for your contribution. it is helpful for ``genimage``.

@whowutwut, for the issue #1816, I think these change maybe not enough.

 1. From ``genimage`` point of view, the attributes ``$arch`` and ``$kernelver `` both exist. But when ``mlnxofed_ib_install`` run, we don't  know the attribute ``$kernelver ``, at least, depending on current code, it is true.

 2. After discussing with @immarvin , we can add code to transmit ``$kernelver `` to ``mlnxofed_ib_install`` , but even if it is finished, it just can resolve multiple kernel problem in diskless scenario, ``mlnxofed_ib_install`` still can't get value of ``$kernelver `` in diskful scenario. because we don't have change to set value for ``kernelver ``. But if issue #1816  is just related to diskless scenario, it is another story.
 
 3. For the value of ``$kernelver `` in ``genimage``. depending on code, we accept customer's input first, if there isn't input, we still guess one. So it still has risk to guess wrong. 

 4.  For ``sort -V``,  when we designed ``hack_uname`` function, we hope it could run in more environments, such like one environment which is without ``sort`` command. So ``hack_uname`` just use the simplest syntax.  But if we have to do this, we can take it into account.

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -2316,18 +2298,30 @@ sub use_hackuname {
...
 #!/bin/bash
 
-case "\$1" in 
+ARCH=$arch
+KERNELVER=$kernelver
+
+case "\$1" in
 "-m")
-    ARCH=\"\$(dpkg --print-architecture 2>/dev/null || rpm -q kernel-\$(\"\$0\" -r) --qf '%{arch}' 2>/dev/null)\" 
+    if [ -z "$ARCH" ]; then
```
mattaezell: Yes, will update

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -2316,18 +2298,30 @@ sub use_hackuname {
...
+    fi
     case "\$ARCH" in 
         "amd64") ARCH="x86_64" ;;
         "ppc64el") ARCH="ppc64le" ;;
         esac
     echo "\$ARCH" ;;
 "-r")
-    cd /lib/modules && for d in * ; do : ; done && echo \$d ;; 
+    if [ -n "$KERNELVER" ]; then
+        echo $KERNELVER
```
mattaezell: Yes, will update

mattaezell:
Thanks @hu-weihua for the review.  For item #4, are you more comfortable using RPM commands?  One option would be to pick the last-installed kernel, which is *probably* what the admin wanted:
`KERNEL=$(rpm -q --last kernel)`
`echo ${KERNEL%% *}`

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -756,22 +732,22 @@ EOS_UNAME
...
         }
     }
 
     #delete the osimage attributes from environment variable
     delete @ENV{qw(IMG_ARCH IMG_NAME IMG_OSVER IMG_PROFILE IMG_PKGLIST IMG_PKGDIR IMG_OTHERPKGLIST IMG_OTHERPKGDIR IMG_ROOTIMGDIR)};
 
-    # restore the orig uname
-    system("mv $rootimg_dir/bin/orig_uname $rootimg_dir/bin/uname");
-
+    unuse_hackuname();
```
mattaezell: It already existed in this file.  It was line 2339 in the original and 2333 in the patched version.

xCAT-server/share/xcat/netboot/rh/genimage
mattaezell: Fixed in commit e71c03d

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -2316,18 +2298,30 @@ sub use_hackuname {
...
         "amd64") ARCH="x86_64" ;;
         "ppc64el") ARCH="ppc64le" ;;
         esac
     echo "\$ARCH" ;;
 "-r")
-    cd /lib/modules && for d in * ; do : ; done && echo \$d ;; 
+    if [ -n "$KERNELVER" ]; then
+        echo $KERNELVER
+    else
+      for d in \$(ls /lib/modules | sort -V) ; do : ; done && echo \$d
```
mattaezell: I left a comment above with an alternate solution that just uses rpm (already used in the script) that might work better.  The other option is just to add the "smarts" in genimage to set the `$kernelver` statically even if the admin didn't specify it.  We can assume the xCAT server has `sort` and `head`

xCAT-server/share/xcat/netboot/rh/genimage
neo954: You are right, it is an existing function. Sorry.

neo954:
I think we still need proper signal handler for `SIGINT` and `SIGTERM`.

hu-weihua:
After discussing in CDL team, we decide to merge this pull request.
@mattaezell , thank you very much for what you have done.

Issue #1948, opened 2016-10-10
some actions keep setting the ospkgsdir to a directory that does not exist.


ralphbellofatto:
we noticed that xcat keeps setting the ospkgsdir attribute on its own.

To duplicate the problem we do:

```
lsdef -t osimage rh72_op5_compute | tee rh72_01.lsdef
genimage rh72_op5_compute
lsdef -t osimage rh72_op5_compute | tee rh72_02.lsdef

diff rh72_01.lsdef rh72_02.lsdef
7a8
>     otherpkgdir=/install/post/otherpkgs/rhels7.2/ppc64le

```

immarvin:
"/install/post/otherpkgs/<os>/<arch>" is the default value of  "otherpkgdir" for an osimage definitions, which are defined on "copycds".  
This default value is a hint for user on where to place the additional 3rd party packages they want to install on compute node, xCAT won't create this directory, user need to create this directory by themselves(http://xcat-docs.readthedocs.io/en/stable/guides/admin-guides/manage_clusters/common/deployment/additionalpkg/nonubuntu_os_other_pkg.html). 
There used to be some discussion inside xCAT team, we finally decided to keep the "otherpkgdir" default value and not to create the actual path by default, which left as a step in the documentation.

ralphbellofatto:
Is there any way to get xcat to not continually insert this.  We removed it and every time we run genimage it comes back..

cxhong:
for this issue, otherpkgdir is inserted from genimage command.  users removed this attribute from their osimage, xCAT should not insert it again after genimage.

daniceexi:
@ralphbellofatto I'd like to know the background why you need to clean up this pah?

ralphbellofatto:
well,

1.  we don't use this path in our systems.
2. we keep our xcat lsdefs under source control and this change that we did not do, keeps appearing.
3.  I get tired of explaining why this path has to be in the system, especially when subsystems complain about referencing it.

whowutwut:
I agree with @ralphbellofatto and @cxhong  that we should not put this back into the attribute if the user removed it.   This is a difference in behavior between diskful and diskless, we are not replacing the `otherpkgdir` if the user removes it for diskful systems when we provision the node.     @cxhong , will create a pull request to remove the addition of the otherpkgdir.

PR #1954, opened 2016-10-11
Don't need to update otherpkgdir to default dir for genimage command

cxhong:
for issue #1948 .

genimage will not add otherpkgdir to default directory if there is no otherpkgdir specified in the osimage.

### DIFF
--- a/xCAT-server/share/xcat/netboot/debian/genimage
+++ b/xCAT-server/share/xcat/netboot/debian/genimage
@@ -119,7 +119,7 @@ $updates{'pkgdir'} = $srcdir if ($tempfile);
 if (!$srcdir_otherpkgs) {
     $srcdir_otherpkgs = "$installroot/post/otherpkgs/$osver/$arch";
 }
-$updates{'otherpkgdir'} = $srcdir_otherpkgs if ($tempfile);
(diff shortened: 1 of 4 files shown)


immarvin:
hi @cxhong ,in ``/opt/xcat/bin/genimage``, ``/opt/xcat/share/xcat/netboot/rh/genimage`` and ``/opt/xcat/lib/perl/xCAT_plugin/genimage.pm`` , the code logic to update the osimage and linux image table after "genimage" are there because we once supported "genimage -a/-p/-o..." to specify the attributes for osimage instead of osimage name.  Now "genimage -a/-p/-o..." has been obsoleted, currently, we only support "genimage <osimage name>", all the attributes will be taken from the osimage definition and there is no need to update the osimage definition on "genimage" any more.

I think the code logic to update the osimage definition on "genimage" should be removed. what's more, we also need to modify "genimage" code to prompt user as an error when  mandatory attribute can not be found, such as "pkglist","pkgdir"...

cxhong:
@immarvin Can we open another issue to keep track the genimage issues you mentioned above?
like to fix the otherpkgdir in this pull request first.

immarvin:
agree to merge this PR first.

Issue #448, opened 2015-11-20
[DEV]xcat code logic only can transfer 'INT' and 'TERM' to plugin, cannot transfer to the script triggered by plugin


hu-weihua:
If user hit "ctrl+c" in client side during running one command, xcatd will receive this signal and kill the specific process (i.e. plugin) which was forked by xcatd and take responsible to handle this command. sometime, xcat plugin will trigger other scripts, when this plugin terminate, these scripts won't be handled. so even these scripts have signal handle function, but due to they can't receive signal, so some unpredictable things will happen.

neo954:
I am going to look into this one.

neo954:
Yeah, a new patch against this issue is available.

neo954:
@hu-weihua, the patch has been merged days ago. Please verify if this issue is fixed.

hu-weihua:
I have verified this issue, the fix works. thanks @neo954 ! :+1:

neo954:
@hu-weihua, Cool.

PR #1961, opened 2016-10-13
Fix github issue #448. From now on, send SIGINT signal to the whole process group of the xCAT plug-in.

neo954:
Fix github issue #448. From now on, send SIGINT signal to the whole process group of the xCAT plug-in.

:-D

### DIFF
--- a/xCAT-server/sbin/xcatd
+++ b/xCAT-server/sbin/xcatd
@@ -1748,7 +1748,11 @@ sub plugin_command {
     # make the request handler process to take care all the plugin children
     $SIG{TERM} = $SIG{INT} = sub {
         foreach (keys %plugin_children) {
+            # Send signal to the whole process group
+            kill 'INT', -$_;
             kill 'INT', $_;
+            kill 'TERM', -$_;
+            kill 'TERM', $_;
         }
         $SIG{ALRM} = sub { xexit 0; };    # wait 1s for grace exit
         alarm(1);
@@ -1873,6 +1877,14 @@ sub plugin_command {
                     binmode($parfd, ':utf8');
                     binmode($pfd,   ':utf8');
                     $child = xCAT::Utils->xfork;
+                    unless (defined $child) {
+                        xCAT::MsgUtils->message("S", "Fork failed");
+                        die;
+                    }
+                    # Fix github issue #448. For the child process, setpgrp()
+                    # here. In this way, the parent can send SIGINT or SIGTERM
+                    # to all the process group.
+                    setpgrp unless $child;
                 } else {
                     if ($req_back) {
                         $req = dclone($req_back);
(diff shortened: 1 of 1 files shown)


neo954:
Use the command below to check the process tree and the process group id (SESS).

    ps axfo stat,euid,ruid,tty,tpgid,sess,pgrp,ppid,pid,pcpu,comm

neo954:
This patch will affect almost all xCAT client command, including but not limited to `genimage`, `copycds`, `xdsh`, `pping`, `ppping`, `rpower`.

The commands listed above are what I tested.

xCAT-server/sbin/xcatd
```diff
@@ -1748,7 +1748,11 @@ sub plugin_command {
     # make the request handler process to take care all the plugin children
     $SIG{TERM} = $SIG{INT} = sub {
         foreach (keys %plugin_children) {
+            # Send signal to the whole process group
+            kill 'INT', -$_;
             kill 'INT', $_;
+            kill 'TERM', -$_;
+            kill 'TERM', $_;
```
chenglch: As the signal for the process group is already sent, why need to send the signal to the single process again?

xCAT-server/sbin/xcatd
```diff
@@ -1873,6 +1877,14 @@ sub plugin_command {
                     binmode($parfd, ':utf8');
                     binmode($pfd,   ':utf8');
                     $child = xCAT::Utils->xfork;
+                    unless (defined $child) {
+                        xCAT::MsgUtils->message("S", "Fork failed");
+                        die;
+                    }
+                    # Fix github issue #448. For the child process, setsid()
```
chenglch: I think the purpose of `setsid` is to **create a new session** and become the process group leader among its child processes.  

Does `setpgrp` works in this situation ?

neo954: The main difference of `setsid()` and `setpgrp()` is `setsid()` detach the tty, and `setpgrp()` do the foreground and background process group things. But since `xcatd` is mostly already runs in a daemon environment, there is not too much difference.

Yes, `setpgrp()` will do the work. I will think through it much better.

xCAT-server/sbin/xcatd
neo954: The main propose of `kill 'INT', $_;` here is in case `setsid()` failed. Since I did not do any error handling there. In this case, `xcatd` will still send signal to the single process. Do as much as what we can do here.

neo954:
Changed to use `setpgrp()` instead of `setsid()`.

chenglch:
cool fix,  agree to merge

neo954:
Thanks, @chenglch.

daniceexi:
Good fix. 👍

PR #1962, opened 2016-10-13
osimagecheck probe enhancements

gurevichmark:
Enhancements to `osimagecheck` probe:
1. Display list of osimage names that point to the same rootimgdir, one per line
2. Display a warning if any osimage object has "template" value of "service.tmpl" or "compute.tmpl", templates that are deprecated and should not be used.

### DIFF
--- a/xCAT-probe/subcmds/osimagecheck
+++ b/xCAT-probe/subcmds/osimagecheck
@@ -9,7 +9,7 @@ use File::Basename;
 use Net::Ping;
 use Getopt::Long qw(:config no_ignore_case);
 
-#use Data::Dumper;
+use Data::Dumper;
 use warnings;
 
 my $program_name = basename("$0");
@@ -120,15 +120,15 @@ sub check_for_duplicate_rootimgdir {
             my @list_of_osimages = split ",", $rootimgdir_osimage_hash{$rootimgdirpath};
             if (scalar(@list_of_osimages) > 1) {
                 # More than one entry in the hash value, means we
-                # have more than on osimage point to the same rootimgdir
+                # have more than one osimage point to the same rootimgdir
                 my $print_list = join("\n  ",@list_of_osimages);
                 $rootimgdirpath=~s/^\s+//; #discard leading space
-                probe_utils->send_msg("$output", "w", "Same attribute value ($rootimgdirpath) used by : \n  $print_list");
+                probe_utils->send_msg("$output", "w", "Same attribute value ($rootimgdirpath) used by diskless osimages: \n  $print_list");
             }
         }
     }
     else {
-        probe_utils->send_msg("$output", "o", "No osimages with duplicate rootimgdir were found.");
+        probe_utils->send_msg("$output", "o", "No diskless osimages with duplicate rootimgdir were found.");
         $rc = 0;
     }
 
@@ -165,7 +165,7 @@ sub check_for_valid_osimage_attributes {
         }
         my ($junk, $template_file) = split "=", $template;
         if (($template_file =~ /compute.tmpl/) || ($template_file =~ /service.tmpl/)) {
-            probe_utils->send_msg("$output", "w", "$osimage_name contains depricted 'template' value \n  $template_file");
+            probe_utils->send_msg("$output", "w", "$osimage_name contains depricated 'template' value \n  $template_file");
         }
     }
 
--- a/xCAT-probe/subcmds/osimagecheck
+++ b/xCAT-probe/subcmds/osimagecheck
@@ -9,7 +9,7 @@ use File::Basename;
 use Net::Ping;
 use Getopt::Long qw(:config no_ignore_case);
 
-use Data::Dumper;
+#use Data::Dumper;
 use warnings;
 
 my $program_name = basename("$0");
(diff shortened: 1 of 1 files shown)


gurevichmark:
Example output:
```
[root@fs1 subcmds]# xcatprobe osimagecheck
No osimages with duplicate rootimgdir were found.                                                                 [ OK ]
```

```
[root@fs1 subcmds]# xcatprobe osimagecheck
Same attribute value (rootimgdir=/install/netboot/rhels7.3-snap3/ppc64le/compute) used by :                       [WARN]
  mg_osimage1
  mg_osimage2
  mg_osimage3
  rhels7.3-snap3-ppc64le-netboot-compute
Same attribute value (rootimgdir=/install/netboot/ubuntu14.04.3/ppc64el/compute) used by :                        [WARN]
  mg2_osimage
  ubuntu14.04.3-ppc64el-netboot-compute
ubuntu14.04.3-ppc64el-install-compute contains depricted 'template' value                                         [WARN]
  /opt/xcat/share/xcat/install/ubuntu/compute.tmpl
ubuntu14.04.3-ppc64el-install-service contains depricted 'template' value                                         [WARN]
  /opt/xcat/share/xcat/install/ubuntu/service.tmpl
ubuntu16.04-ppc64el-install-compute contains depricted 'template' value                                           [WARN]
  /opt/xcat/share/xcat/install/ubuntu/compute.tmpl
ubuntu16.04-ppc64el-install-service contains depricted 'template' value                                           [WARN]
  /opt/xcat/share/xcat/install/ubuntu/service.tmpl
[root@fs1 subcmds]#

```

xCAT-probe/subcmds/osimagecheck
```diff
@@ -127,5 +141,33 @@ sub check_for_valid_osimage_attributes {
...
+        probe_utils->send_msg("$output", "d", "Processing $_.") if ($VERBOSE);
+        my ($osimage_name, $template) = split ":", $_;
+        if ($template eq " template=") {
+            # Exclude entries that do not have template set
+            probe_utils->send_msg("$output", "d", "No template for osimage $osimage_name") if ($VERBOSE);
+            next;
+        }
+        my ($junk, $template_file) = split "=", $template;
+        if (($template_file =~ /compute.tmpl/) || ($template_file =~ /service.tmpl/)) {
+            probe_utils->send_msg("$output", "w", "$osimage_name contains depricted 'template' value \n  $template_file");
```
whowutwut: I think there's a spelling issue here, `depricted`

gurevichmark:
Suggested changes made under the second commit.

xCAT-probe/subcmds/osimagecheck
```diff
@@ -110,11 +111,24 @@ sub check_for_duplicate_rootimgdir {
...
+    if ($any_dups) {
+        # Loop through each entry in hash to print a list of osimgage objects that point to the same rootimgdir
+        foreach $rootimgdirpath (keys %rootimgdir_osimage_hash) {
+            my @list_of_osimages = split ",", $rootimgdir_osimage_hash{$rootimgdirpath};
+            if (scalar(@list_of_osimages) > 1) {
+                # More than one entry in the hash value, means we
+                # have more than one osimage point to the same rootimgdir
+                my $print_list = join("\n  ",@list_of_osimages);
+                $rootimgdirpath=~s/^\s+//; #discard leading space
+                probe_utils->send_msg("$output", "w", "Same attribute value ($rootimgdirpath) used by diskless osimages: \n  $print_list");
```
whowutwut: Sorry to being a little picky on this but the externals still not expressing the right level of concern in my opinion.  Looking at the externals, I don't think I would necessarily feel it's a potentially dangerous situation.  Since we are not abstracting out the variable, maybe we should just not use "attribute value"   How about ....

`Identical root image directory (rootimgdir) is configured for the following diskless osimages!`

xCAT-probe/subcmds/osimagecheck
```diff
@@ -127,5 +141,33 @@ sub check_for_valid_osimage_attributes {
...
+    foreach (@all_osimages_template_lines) {
+        probe_utils->send_msg("$output", "d", "Processing $_.") if ($VERBOSE);
+        my ($osimage_name, $template) = split ":", $_;
+        if ($template eq " template=") {
+            # Exclude entries that do not have template set
+            probe_utils->send_msg("$output", "d", "No template for osimage $osimage_name") if ($VERBOSE);
+            next;
+        }
+        my ($junk, $template_file) = split "=", $template;
+        if (($template_file =~ /compute.tmpl/) || ($template_file =~ /service.tmpl/)) {
```
whowutwut: It looks like there's either a bug in our code or something not right with the ubuntu templates, as it's always picking up the compute.tmpl files instead of their respective version file.   However, I think we should leave this and open an issue for the osimage

whowutwut: Actually, investigating further.... we only ship a single compute.tmpl file for Ubuntu
```
[root@fs1 tmp]# ls -ltr /opt/xcat/share/xcat/install/ubuntu/*.tmpl
lrwxrwxrwx 1 root root   12 Jan  1  2000 /opt/xcat/share/xcat/install/ubuntu/service.tmpl -> compute.tmpl
lrwxrwxrwx 1 root root   12 Jan  1  2000 /opt/xcat/share/xcat/install/ubuntu/kvm.tmpl -> compute.tmpl
lrwxrwxrwx 1 root root   12 Jan  1  2000 /opt/xcat/share/xcat/install/ubuntu/cudaruntime.tmpl -> compute.tmpl
lrwxrwxrwx 1 root root   12 Jan  1  2000 /opt/xcat/share/xcat/install/ubuntu/cudafull.tmpl -> compute.tmpl
-rw-r--r-- 1 root root 4826 Sep 20 04:43 /opt/xcat/share/xcat/install/ubuntu/compute.tmpl
```

So we actually can't check just for "compute.tmpl" in the string as that only applies to RHEL  images., not Ubuntu.  So I think we do need a code change here because this warning doesn't apply for Ubuntu images.   I think the same comment applies to `service.tmpl`, perhaps we should not check for that at this time.

gurevichmark:
Another commit to fix output message wording as suggested, and skip printing warnings for `compute.tmpl` and `service.tmpl` templates for Ubuntu osimages.

whowutwut:
Thanks, will merge

PR #1964, opened 2016-10-13
Allow pgsqlsetup(1) to specify additional IP addresses to listen on

khm:
On AIX, pgsqlsetup(1) only configures the management node IP in PostgreSQL's configuration.  On Linux, it only configures localhost and the management node IP.  This change creates a new option in pgsqlsetup(1) that allows the user to specify an additional address to be added in the listen_addresses line in PostgreSQL's configuration.  If the user specifies this option more than once, all specified addresses are added to the line.

This is useful in cases where the management node has more than one IP, such as in HA configurations wherein an address is shared between multiple management nodes.  In these cases, pgsqlsetup(1) may use the system's primary address, when it is preferred to listen on the shared address as well.  This is also useful in situations where the cluster is fragmented across different subnets, and the management node needs to talk to service nodes that are on disparate networks.

### DIFF
--- a/xCAT-client/pods/man1/pgsqlsetup.1.pod
+++ b/xCAT-client/pods/man1/pgsqlsetup.1.pod
@@ -9,7 +9,7 @@ B<pgsqlsetup> {B<-h> | B<--help>}
 
 B<pgsqlsetup> {B<-v> | B<--version>}
 
-B<pgsqlsetup> {B<-i> | B<--init>} [B<-N> | B<--nostart>] [B<--listen> I<address>] [B<-P> | B<--PCM>] [B<-o> | B<--odbc>] [B<-V> | B<--verbose>]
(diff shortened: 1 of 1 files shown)


whowutwut:
@khm Thanks for this pull request and your interest in the xCAT project.   In order for the xCAT team to consider your changes, you must first sign a CLA, instructions are documented here: http://xcat-docs.readthedocs.io/en/stable/developers/license/contributors.html

Or if you know you are covered under a Corporate CLA that was signed, can you please give us the corporation name and we can verify.  Thanks!

whowutwut:
@immarvin  have you reviewed this change? Taking off a release....  I think we still require a CLA in order to consider this.  If no response , we can close for now and wait for @khm  to reopen.

khm:
I'll close this for now, as I do not have a CLA and it will take some time to convince my organization.  Thanks for your consideration.

PR #1965, opened 2016-10-14
Clean up some of the log messages from bmcsetup that seems useless 

whowutwut:
- Remove the USER/PASS clear text being sent back to syslog
-  Remove the access display, doesn't look useful


zet809:

PR #1970, opened 2016-10-14
Small fixes to ib pkglist and mlnxofed_ib_install script

gurevichmark:
Two very small fixes to IB:
1. Remove duplicate entry for `libnl` from ib pkglist
2. Add missing character to the output of `mlnxofed_ib_install` script so that "Before" and "After" lines line up.


whowutwut:
Agreed to merge

Issue #1972, opened 2016-10-14
stateless boot node looses network connection at dhcp lease expiration time.


ralphbellofatto:
We noticed this first that when we booted a diskless image, that it would loose its network connectivity after 12 hours.

After some further isolation we determined that this 12 hours correlated to the dhcp leases time.

We followed the directions to define a node:
http://xcat-docs.readthedocs.io/en/stable/guides/admin-guides/manage_clusters/ppc64le/discovery/manually_define.html

and ended up with the following lsdef clause...

```
c460c037:
    objtype=node
    arch=ppc64le
    bmc=10.2.3.1
    bmcport=0
    bmcvlantag=2
    chain=runcmd=bmcsetup,standby
    cons=ipmi
    consoleondemand=0
    cpucount=160
    cputype=POWER8NVL (raw), altivec supported
    disksize=sda:2000GB,sdb:2000GB
    groups=all,compute,c460f03
    hostnames=c460f03u03
    initrd=xcat/osimage/rh72_op5_compute/initrd-stateless.gz
    installnic=mac
    ip=10.1.3.1
    kcmdline=imgurl=http://!myipfn!:80//install/netboot/rhels7.2/ppc64le/rh72_op5_compute/rootimg.cpio.gz XCAT=!myipfn!:3001 NODE=c460c037 FC=0 BOOTIF=70:e2:84:14:0d:e6 
    kernel=xcat/osimage/rh72_op5_compute/kernel
    mac=70:e2:84:14:0d:e6
    memory=523125MB
    mgt=ipmi
    mtm=8335-GTB
    netboot=petitboot
    nodetype=mp
    os=rhels7.2
    postbootscripts=confignics,setroute,setupntp
    primarynic=mac
    profile=compute
    provmethod=rh72_op5_compute
    rack=c460f03
    serial=100A49A
    supportedarchs=ppc64
    switch=c460tors03
    switchport=3
    unit=u03
```

To see the problem without waiting 12 hours and to isolate parts of our software loadout we pointed the node at the default netboot-compute image we got with copycds and then reduced the dhcpleases time to 5 minutes with the following procedure:

```
genimage rhels7.2-ppc64le-netboot-compute
packimage rhels7.2-ppc64le-netboot-compute
nodeset c460c037 osimage=rhels7.2-ppc64le-netboot-compute


# keep the leases time short to stimulate the problem.
chdef -t site clustersite dhcplease=300
makedhcp -n
makedhcp -a
[...]

Issue #104, opened 2015-08-20
[Customer]Makes xCAT support NetworkManager in CentOS7


onimsha:
From RHEL/CENTOS 7, NetworkManager was default, hence it makes some problems with xCAT:
1. setroute won't work if we add a routenames with net=0.0.0.0 (for setup default gw via setroute)
2. /etc/sysconfig/static-routes wont apply on boot

Workaround is disable NetworkManager, but is there any possibilities xCAT configure routes via Network Manager?

daniceexi:
1. We considered that using networkManager to manage all the network configuration. But from some customers that networkManager is not stable too. So we are looking that if there issues that cannot be handled without NetworkManager, we will migrate to NetworkManager.

2. The NetworkManager is disabled by xCAT after node deployment. Could you check the status of it, and how about reboot your node?

whowutwut:
Are you only trying to configure the default gateway for the machines OR trying to set up routing to different networks provided by multiple network interfaces? 
 i.e  eth0 -> 10.X network, eth1 -> 20.X network.. etc...

onimsha:
@daniceexi : Actually i wrote a postscript for disabling NetworkManager
@whowutwut : Both, i manage default gateway and static routes via "setroute" and noderes.routenames

daniceexi:
@immarvin 
Could you check why the NetworManager did not stop for CentOS7?

whowutwut:
We are looking at NetworkManager in the next release.

daniceexi:
We are working in the NetworkManager support for rh7. It will be ready in next release.

immarvin:
NetworkManager should have been stopped in the following commit:

````
commit 52061dd6623a8bf81dd9145653759428c1b89fd5
Author: ertaozh <ertaozh@cn.ibm.com>
Date:   Mon Apr 25 03:08:59 2016 -0400

    Fix issue 942:[PCM] CN IP loss after 12 hours due to the dhcp.lease timeout value 43200

diff --git a/xCAT-server/share/xcat/install/scripts/post.rhels7 b/xCAT-server/share/xcat/install/scripts/post.rhels7
index b34d7d0..3b36dfe 100644
--- a/xCAT-server/share/xcat/install/scripts/post.rhels7
+++ b/xCAT-server/share/xcat/install/scripts/post.rhels7
@@ -20,3 +20,11 @@ do
   #nic name change during the install and first_reboot
   sed -i '/HWADDR/d' $i
 done
+
[...]

PR #1977, opened 2016-10-17
fix issue #1972: Stateless boot node looses network connection at dhcp lease expiration time;add NetworkManager in to rh7 diskless osimage pkglist

immarvin:
fix issue:stateless boot node looses network connection at dhcp leaseexpiration time. #1972;
add NetworkManager in to rh7 diskless osimage pkglist

### DIFF
--- a/xCAT-server/share/xcat/netboot/rh/compute.rhels7.ppc64.pkglist
+++ b/xCAT-server/share/xcat/netboot/rh/compute.rhels7.ppc64.pkglist
@@ -30,3 +30,4 @@ xz
 grub2
 grub2-tools
 bzip2
+NetworkManager
--- a/xCAT-server/share/xcat/netboot/rh/compute.rhels7.x86_64.pkglist
+++ b/xCAT-server/share/xcat/netboot/rh/compute.rhels7.x86_64.pkglist
@@ -22,3 +22,4 @@ net-tools
 gzip
 tar
 xz
+NetworkManager


zet809:
@immarvin From my point of view, if no NetworkManager installed, the CN can not even get a dhcp IP at all rather than lose IP 6 hours later since there is no dhclient run. So the dhclient must run at beginning, and then there is other reason that dhclient process is killed.

immarvin:
hi @zet809 , the diskless node will get ip address on "service network start" without NetworkManager installed. "dhclient" is not run directly, it is triggered by either NetworkManager or network service. The difference between "NetworkManager" and "network" service is that dhclient process will be running as a child process of "NetworkManager" service, but will not be running after "network" service is started.

whowutwut:
I am ok with the changes in this pull request.

whowutwut:
Actually, on 2nd thought, if you look at Matt's comment https://github.com/xcat2/xcat-core/issues/1972#issuecomment-254380386, does it make sense to introduce NetworkManager into the compute nodes? It looks like some customers will change to static IP to reduce the dhcp from running on the compute nodes in production...

immarvin:
hi @whowutwut ,  "dhcp" is the default option for user, we'd better make it work anyway.  NetworkManager is just a new framework in redhat to manage network configuration instead of traditional "ifcfg-xxx" way. If customer want to configure static ip on diskless node, that can use the postscript "confignics -s".  

There is some problem introduced by NetworkManager, see https://github.com/xcat2/xcat-core/issues/104

PR #1991, opened 2016-10-19
fix list wrong adapters.

bybai:
In the old code, when hash value is changed, the array value is changed too, so the array just keep one same adapter information.
````
# /opt/xcat/bin/localrest list adapters
json
[{"ip":"127.0.0.1/8","name":"lo"},{"ip":"10.5.106.8/8","name":"eth0","mac":"42:5f:0a:05:6a:08"}]
````


immarvin:
agree to merge

Issue #1958, opened 2016-10-13
[FVT]Failed to run rmvm when setup  SN in x86_64 for rh7.3 and rh6.8


caomengmeng:
Using latest daily build to install SN, when run below command, case failed.

```
RUN:if [ "x86_64" != "ppc64"  -a  "kvm" != "ipmi" ];then if [[ "dir:///var/lib/libvirt/images/" =~ "phy" ]]; then rmvm c910f04x18v03 -f -p  &&  mkvm c910f04x18v03; else rmvm c910f04x18v03 -f -p  &&  mkvm c910f04x18v03 -s 20G; fi;fi

[if [ "x86_64" != "ppc64"  -a  "kvm" != "ipmi" ];then if [[ "dir:///var/lib/libvirt/images/" =~ "phy" ]]; then rmvm c910f04x18v03 -f -p  &&  mkvm c910f04x18v03; else rmvm c910f04x18v03 -f -p  &&  mkvm c910f04x18v03 -s 20G; fi;fi] Running Time:1 sec
RETURN rc = 1
OUTPUT:
c910f04x18v03: Error: Cannot remove guest vm, no such vm found
```

tingtli:
@caomengmeng  As i comments on defect ＃1872. we need to give a steps for how to reproduce this defect. Not only the logs in our test environment. 
Please reference to https://w3-connections.ibm.com/wikis/home?lang=en#!/wiki/xCAT/page/Defect%20process for the how to open a valid defect. Thanks!

gurevichmark:
I can see now that the code I put under pull request #1942 is being executed. That code was put in to fix issue #1747.
From what I can see here, when `rmvm` is called, the vm is not there anymore.
Now the question is how the vm was removed ?

tingtli:
I followed this defect and didn't find the reason why it was deleted before we execute the rmvm. In all the test cases, we always run rmvm and mkvm at the same time to make sure we have a fresh vm. My guess is since the vm is reused, it is probably that something went wrong(mkvm fails??). @gurevichmark Let me first find some trace and give you the reproduced steps.

whowutwut:
Is it possible that before the rmvm execution line, for the case where the storage `!= "phy"`, we list out the contents of `/var/lib/libvirt/images`?   Like: 

```
if [[ "__GETNODEATTR($$CN,vmstorage)__" =~ "/var/lib/libvirt/images" ]]; then
[...]

PR #1993, opened 2016-10-19
update for test vm

tingtli:


daniceexi:
This is a workaround for the rmvm issue in the automation. It looks ok for merge.

whowutwut:
I'm not sure why we do that instead of listing out the images dir? To see what it looks like before running any commands. Otherwise we are just guessing on what to run...

Also...should just remove the code instead of commenting out

whowutwut:
@tingtli can you include the issue number in the description that it is resolving?

whowutwut:
Adding reference to Issue #1958

tingtli:
@whowutwut  i make some some updates according to your comments. See PR #2033 
And for the hold operation which I mentioned in the mail, this is a xcatjktest operation which will pending on the failed case.

whowutwut:
Ok, thanks @tingtli ! I see it now

PR #1995, opened 2016-10-19
Message to tell if just dup definitions or discovery prob

gurevichmark:
When two node definitions found with the same mtm and serial number, display 2 different messages.
If one of the node names starts with "node-", display a message indicating a possible discovery issue. Otherwise display a message indication a generic problem of 2 definitions with the same mtm + serial.

Output as follows:

```
[root@fs1 subcmds]# xcatprobe nodecheck
node-mg has been discovered as frame45cn08 and can be removed using 'rmdef node-mg'                          [FAIL]
[root@fs1 subcmds]# 
```

```
[root@fs1 subcmds]# xcatprobe nodecheck
Duplicate node definition found for the same  mtm=8335-GTB serial=100465A : frame45cn08 and frame45mg        [FAIL]
[root@fs1 subcmds]# 
```

```
[root@fs1 subcmds]# xcatprobe nodecheck
No nodes with duplicate mtm and serial numbers were found.                                                   [ OK ]
[root@fs1 subcmds]#
```

### DIFF
--- a/xCAT-probe/subcmds/nodecheck
+++ b/xCAT-probe/subcmds/nodecheck
@@ -23,7 +23,7 @@ my $rst     = 0;
 # Match pattern for discovered nodes: node-1234-ABCDE-123456
 #                                     node- d+ -[a-z]- d+
 #my $discovered_node_pattern =      '^node-\d+-[a-zA-Z]+-\d+';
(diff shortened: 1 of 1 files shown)


whowutwut:
Thanks for making these changes!

PR #1997, opened 2016-10-20
fix rinv should return correct warning message when the option is not valid

### LINKED ISSUE #1987: [FVT]:rinv should return correct warning message when the option is not valid.
xCAT 2.12.4 on rhels7.2

```
[root@c910f03c05k23 /]# lsxcatd -v
Version 2.12.4 (git commit 6ec81e919be18e62f1f4b898a7336a7316bdad1f, built Tue Oct 18 06:15:54 EDT 2016)
[root@c910f03c05k23 /]# rinv p8euh mprom
p8euh: BMC Firmware: 7.40
[root@c910f03c05k23 /]# rinv p8euh maprom
== hangs here
```

p8euh is an Tulleta L machine.

On another node, Here is the output.

```
[root@fs1 ~]# rinv frame45cn02 mprom
frame45cn02: BMC Firmware: 2.13
[root@fs1 ~]# rinv frame45cn02 maprom
Error: Error parsing FRU data from BMC
```

We should return the warning that the parameter is not correct. But not go in the code to get the value from bmc.

penguhyang:
I think return the usage info to user can be helpful.
fix #1987

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -2460,6 +2460,7 @@ sub beacon_answer {
 
 sub inv {
     my $sessdata   = shift;
+    my $command = $sessdata->{command};
     my $subcommand = $sessdata->{subcommand};
 
     my $rc = 0;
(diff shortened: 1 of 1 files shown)


zet809:

PR #2004, opened 2016-10-20
Support kitcomponent can be add to higher version of OS

cxhong:
Request from ESSL,,  If osminor version is set for the kits,  it can be add to higher version of OS minor number instead of minor version has to be matched.

Test results:

```
# addkitcomp -i rhels7.3-ppc64le-netboot-compute xlc.license-compute-13.1.4-0-rhels-7.2-ppc64le
Assigning kit component xlc.license-compute-13.1.4-0-rhels-7.2-ppc64le to osimage rhels7.3-ppc64le-netboot-compute
Kit components xlc.license-compute-13.1.4-0-rhels-7.2-ppc64le were added to osimage rhels7.3-ppc64le-netboot-compute successfully

# addkitcomp -i rhels7.1-ppc64le-netboot-compute xlc.license-compute-13.1.4-0-rhels-7.2-ppc64le
Error: osimage rhels7.1-ppc64le-netboot-compute is not compatible with kit component xlc.license-compute-13.1.4-0-rhels-7.2-ppc64le with attribute minorversion

I ran genimage , looks like xlc.license-compute is successfully installed.

# chroot /install/netboot/rhels7.3/ppc64le/compute/rootimg
# rpm -qa | grep xlc
xlc.license-compute-13.1.4-0.noarch
xlc-license.13.1.4-13.1.4.0-160519.ppc64le
```

### DIFF
--- a/docs/source/advanced/kit/custom/build/createkit.rst
+++ b/docs/source/advanced/kit/custom/build/createkit.rst
@@ -56,7 +56,7 @@ The ``buildkit.conf`` file is a sample file that contains a description of all t
       kitlicense=ILAN           <== the default kit license string is "EPL"
       kitdeployparams=pe.env    <== pe.env has to define in the other_files dir.
 
-**kitrepo** --- This stanza defines a Kit Package Repository. There must be at least one kitrepo stanza in a kit build file.  If this kit need to support multiple OSes, user should create a separate repository for each OS.  Also, no two repositories can be defined with the same OS name, major/minor version, and arch.  ::
+**kitrepo** --- This stanza defines a Kit Package Repository. There must be at least one kitrepo stanza in a kit build file.  If this kit need to support multiple OSes, user should create a separate repository for each OS.  Also, no two repositories can be defined with the same OS name, major version, and arch.  ::
 
   kitrepo:
       kitrepoid=rhels6_x86_64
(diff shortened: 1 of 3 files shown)


bybai:
Hi @cxhong,
I think if we want to support >=, we should support = at the same time.
For example, user just want to support rh7.3, no other version, if we use lt,  we lose control of rh7.3.

If we want to support >= and = both, we should think about how to modify buildkit.conf.
What about your opinion.

cxhong:
@bybai , do we still want to support this after HongBo's email? he agreed your comment about this issue 
````
 customized osimage can use different osvers in osimage, it is flexible, while kit has static osbasename, osmajorversion and osminorversio in buildkit.conf, in such case, user always use -f as work around for addkitcomp. >= is not so valuable for osminorversion.    
``````

**Comment by bybai** [comment]
hi @cxhong and @whowutwut ,
Although hongbo wants to use -f, we can still support >= and == here. The requirements are clear.  -f is a workaround, but >= and == should be better way for os minorversion check.  What about your comments?

**Comment by whowutwut** [comment]
In our meeting today, we discussed the possibility of using the following: 

```
minor_version = 2  # minor version must match 2
minor_version = >2 # minor version can be >=2
minor_version = <2 # minor version can be <= 2
```
or  only support the >= using 
```
minor_version = 2+
```

bybai:
Hi @whowutwut and @cxhong,
 It should be the interface in buildkit.conf, so I think the following better:
minor_version = 2  # minor version must match 2
minor_version = >2 # minor version can be >=2
minor_version = <2 # minor version can be <= 2

usually， we use >,>=,=,=<, to describe dependencies version, if we want to simplify the version control , we can just use above.

whowutwut:
@cxhong should we move this to next release or first allow this pull request in and then add the discussion work in another pull request for next release?

cxhong:
i am working on this issue,  have a lot of hits.....  I will move to 2.13

cxhong:
@bybai , can u review this code?

Issue #1852, opened 2016-09-17
when using copycds -p, stateful install has yum point to the wrong repository.


ralphbellofatto:
we installed a new iso to a different directory with 
copycds isoname.iso -p /install/RHEL-7.2-20160512.0-Garrison

And we created a repository that points the pkgdir attribute to this directory.

And then provisioned a node with this image:  after the node boots we inspected the /etc/yum.repos.d file and found the following:'

[root@c460c072 yum.repos.d]# ls /etc/yum.repos.d
local-repository.repo  redhat.repo

the xcat postinstall environment properly points to the same place as the place we copied it:

```
[root@c460mgt01 lsdefs]# lsdef -t osimage test-sb-yum -i pkgdir
Object name: test-sb-yum
    pkgdir=/install/RHEL-7.2-20160512.0-Garrison/ppc64le
```

The environment that the postscripts get points to this image.
/xcatpost/mypostscript:OSPKGDIR='/install/RHEL-7.2-20160512.0-Garrison/ppc64le'

This sometimes works right and sometimes gets really messed up because each time we perform a
copycds -p '/install/RHEL-7.2-somenew-dir

a new link gets created in:
lrwxrwxrwx 1 root root 45 Sep 16 09:35 ppc64le -> /install/RHEL-some-new-dir

So the yum repos.d file now will point to the wrong place...

NOTE: When we have two or more pkgdir's then this seems to work correctly:

We also have a configuration where we have two pkgdirs as follows:

```
[root@c460mgt01 lsdefs]# lsdef -t osimage rh72_op5_login -i pkgdir
Object name: rh72_op5_login
    pkgdir=/install/RHEL-7.2-20160428.0-Garrison/ppc64le,/install/kernel-3.10.0-327.28.2.el7.openpower5
```

When we provision with this configuration we get:

```
/xcatpost/mypostscript:OSPKGDIR='/install/RHEL-7.2-20160428.0-Garrison/ppc64le,/install/kernel-3.10.0-327.28.2.el7.openpower5'
```

and

```
[root@c460mgt01 lsdefs]# xdsh c460login01 cat /etc/yum.repos.d/local*
c460login01: [local-rhels7.2-ppc64le-ppc64le]
c460login01: name=xCAT configured yum repository for rhels7.2/ppc64le/ppc64le
c460login01: baseurl=http://172.1.0.1/install/rhels7.2/ppc64le/
c460login01: enabled=1
c460login01: gpgcheck=0
```

[...]

PR #2009, opened 2016-10-21
Use path from copycds -p in local-repository.tmpl for RHEL

gurevichmark:
This pull request fixes issue #1852 and does the following:
1. For RHEL puts the path specified on `copycds -p` into **local-repository.tmpl** file for `baseurl` attribute.
2. For RHEL uses the value of site.installdir instead of hardcoded `/install` in **local-repository.tmpl** file for `baseurl` attribute.
3. Returns error message if path specified on `copycds -p` does not start with the path in site.installdir.
4. Fixes some formatting and grammar in docs


immarvin:
hi @gurevichmark , I think the root cause of the broken repo on the installed node is caused by the following code slice in subroutine ``copycd`` in ``anaconda.pm``.

````
2456     unless ($path =~ /^($defaultpath)/)
2457     {
2458         mkpath($defaultpath);
2459         if (-d $defaultpath)
2460         {
2461             rmtree($defaultpath);
2462         }
2463         else
2464         {
2465             unlink($defaultpath);
2466         }
2467
2468         my $hassymlink = eval { symlink("", ""); 1 };
2469         if ($hassymlink) {
2470             symlink($path, $defaultpath);
2471         } else
2472         {
2473             link($path, $defaultpath);
2474         }
2475
2476     }
````

The code in your PR is cute,  but since it still relies on the link created in ``copycds``, which will break the yum repo on the installed nodes in the cluster when a ``copycds -p`` is performed with a upgraded xCAT version. 

My advice is remove the code slice in anaconda::copycd above, and pass the real $path as a argument to subroutine localize_yumrepo(), and use $path instead of "$installpfx" in the following line in Yum.pm:
````
 25     find(\%options, $installpfx);
````
this involves the interface change of   localize_yumrepo(), but since localize_yumrepo is only called in anaconda.pm, I think it's ok to make this modification.

gurevichmark:
Closing this pull request. The code logic was made under #2050 and the doc changes were done under #2099

Issue #2013, opened 2016-10-24
[fvt]2.14 xcatprobe could not return right output if many nodes's  status return in same  line in cluster's log


junxiawang:
env:rhels 7.3 ppc64le
build:root@c910f03c05k09 ~]# lsxcatd -v
Version 2.12.4 (git commit b1874179c7250cacaaed1e9fa20d120bb5f9fad0, built Sun Oct 16 06:15:52 EDT 2016)
[root@c910f03c05k09 ~]# 
How to reproduce:

```
nodeset all1 osimage
rpower all1 boot
xcatprobe osdeploy -n all1
[c910f03c05k02] /xcatpost/mypostscript.post return                                                                
60 minutes have expired, stop monitoring                                                                          [INFO]
==================osdeploy_probe_report=================
There are 2 node provision failures                                                                               
c910f03c05k02 : Trigger target node reboot failed                                                                 [FAIL]
c910f03c05k0220 : Trigger target node reboot failed                                                               [FAIL]
According to cluster' log:
Oct 23 10:57:57 c910f03c05k09 xcat: c910f03c05k02 c910f03c05k0220 status: powering-on statustime: 10-23-2016 10:57:57s 

```

PR #2017, opened 2016-10-24
fix issue 2013, get nodes array to set status

xuweibj:
#2013


hu-weihua:
I have reviewed the change and agree to merge, thanks @xuweibj .

PR #2018, opened 2016-10-24
Ensure the requested kernel version and kernel packages are correct

mattaezell:
If a specific version of the kernel was requested, make sure any other
kernel-related packages requested in a package list get the same version.
Note, packages that get pulled in due to dependencies are NOT checked.
Also avoid updating the kernel packages when the requested version has already
been installed.

### DIFF
--- a/xCAT-server/share/xcat/netboot/rh/genimage
+++ b/xCAT-server/share/xcat/netboot/rh/genimage
@@ -367,20 +367,20 @@ unless ($onlyinitrd) {
             my $pa  = $pkg_hash{$pass}{$_};
             my @npa = ();
 
-            # replace the kernel package with the name has the specific version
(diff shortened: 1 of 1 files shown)


mattaezell:
We have a specific kernel that we need to run on our GTB/Garrison/Minsky systems, but we also want to be able to pull in security updates for other packages.  We specify a `kerneldir` and `kernelver`, and point our `pkgdir` at a local RHEL mirror.  When new kernel is released from RedHat and we update our local mirror, `genimage` pulls in the newer `kernel-headers` and `kernel-devel` so our `dkms` builds fail.  This patch seems to avoid that issue

immarvin:
agree to merge, thanks for your contribution @mattaezell

Issue #1922, opened 2016-09-29
[DEV]"updatenode" with "debugmode=1/2" does not output the debug trace of shell postscripts


immarvin:
````
c910f02c01p14:/install/postscripts # lsxcatd -v
Version 2.12.3 (git commit 8903bbaa637f7e44dde01c95e546c88f67019a0b, built Wed Sep 28 11:14:39 EDT 2016)
c910f02c01p14:/install/postscripts # tabdump site |grep xcatdebugmode
"xcatdebugmode","1",,
c910f02c01p14:/install/postscripts # updatenode c910f02c02p28 -P syslog -V
Running command on c910f02c01p14: /bin/hostname 2>&1

Running command on c910f02c01p14: ip -4 --oneline addr show |awk -F ' ' '{print $4}'|awk -F '/' '{print $1}' 2>&1

Running command on c910f02c01p14: /bin/hostname 2>&1

Running command on c910f02c01p14: ip -4 --oneline addr show |awk -F ' ' '{print $4}'|awk -F '/' '{print $1}' 2>&1

  c910f02c01p08: Internal call command: xdsh c910f02c02p28 --nodestatus -s -v -e /install/postscripts/xcatdsklspost 1 -m 10.2.1.8 'syslog' --tftp /tftpboot --installdir /install --nfsv4 no -c -V
Running command on c910f02c01p08: /bin/hostname 2>&1
Running command on c910f02c01p08: ip -4 --oneline addr show |awk -F ' ' '{print $4}'|awk -F '/' '{print $1}' 2>&1
Running command on c910f02c01p08: hostname 2>&1
Running command on c910f02c01p08: /opt/xcat/bin/pping c910f02c02p28 2>&1
c910f02c02p28: running /tmp/file1woCXW.dsh 1 -m 10.2.1.8 syslog --tftp /tftpboot --installdir /install --nfsv4 no -c -V
c910f02c02p28: xcatdsklspost : trying to download postscripts from http://10.2.1.8/install/postscripts/
c910f02c02p28: /tmp/file1woCXW.dsh : download_postscripts return successfully
c910f02c02p28: xcatdsklspost: downloaded postscripts successfully
c910f02c02p28: /tmp/file1woCXW.dsh :  trying to download http://10.2.1.8/tftpboot/mypostscripts/mypostscript.c910f02c02p28
c910f02c02p28: /tmp/file1woCXW.dsh download_mypostscript return successfully
[...]

PR #2022, opened 2016-10-25
Add test case to check the update syslog when xcatdebugmode opens.

caomengmeng:
@tingtli 
add test case to check if the execution logs are output, when xcatdebugmode value is one or two: 
https://github.com/xcat2/xcat-core/issues/1922
Test in rhels7.2 ppc64 and ubuntu16.04.1 ppc64le, works well.
please help to review.


tingtli:
@penguhyang  please help review. Thanks!

penguhyang:
Looks good, agree to merge.

PR #2023, opened 2016-10-25
Missing version when building xcat-genesis-base

zet809:
1. Add param version in xCAT-genesis-base.spec
2. Modify an spell error in buildrpm script when building mlx driver.


whowutwut:
I'm ok with the changes. will merge

whowutwut:
The versioning was changed, but for the xCAT-genesis base, we have been using the major version instead of the minor version, so 2.12 makes sense here

Issue #1940, opened 2016-10-05
running packimage prevents nodes from booting.


ralphbellofatto:
If we run the following command, which builds a 10GB diskless boot image and there was a previous valid image in that place, we are unable to boot the nodes that point to this image until the packimage completes.
```
packimage rh72_op5_compute
```

the node eventually boots up but reports this:
```
[Wed Oct  5 08:27:55 2016][   14.747784] dracut-initqueue[1470]: http://10.1.0.1//install/netboot/rhels7.2/ppc64le/rh72_op5_compute/rootimg.cpio.gz:                                                    
[Wed Oct  5 08:27:55 2016][   14.748376] dracut-initqueue[1470]: 2016-10-05 12:27:49 ERROR 403: Forbidden.                                                                                              
[Wed Oct  5 08:28:22 2016][   41.753150] dracut-initqueue[1470]: http://10.1.0.1//install/netboot/rhels7.2/ppc64le/rh72_op5_compute/rootimg.cpio.gz:                                                    
[Wed Oct  5 08:28:22 2016][   41.753823] dracut-initqueue[1470]: 2016-10-05 12:28:16 ERROR 403: Forbidden.                                                                                              
[Wed Oct  5 08:28:49 2016][   68.758139] dracut-initqueue[1470]: http://10.1.0.1//install/netboot/rhels7.2/ppc64le/rh72_op5_compute/rootimg.cpio.gz:                                                    
[Wed Oct  5 08:28:49 2016][   68.758699] dracut-initqueue[1470]: 2016-10-05 12:28:43 ERROR 403: Forbidden.                                                                                              
```

And there is nothing in the xcat node status to indicate that this is the hold up.

Since package takes about 10 minutes, this starts to cause operational problems when nodes are being rebooted.

I'm not sure what the expected behavior should be here.  There are a couple of possibilites.

- have pack image put the put the new compressed files into a temporary directory and when done, remove the original files and move the new files in its place.  This would have the advantage of cutting down the time that the booting node is waiting for the packimage to complete.
- detect the condition and change the lsdef status and the nodestat results during the boot sequence to indicate a delay is in progress and why.

- when performing the pack image, warn the operator that what they are doing will prevent nodes from booting if they are currently provisioned with that diskless image.  This might encourage using an a/b approach to image directories and properly direct the administrator to avoid using one that is currently in play.

[...]

PR #2027, opened 2016-10-25
Update the packimage man page, rootimg file not avail while running the command

whowutwut:
The issue brought up in #1940, this small change to the man page documentation to help explain the timing issue when running packimage on an osimage that is currently active in the cluster.  The compute nodes will encounter an ERROR when trying to pull the rootimg file until the packimage command completes.


immarvin:
agree to merge

PR #2032, opened 2016-10-26
fix nodediscoverls output limits serial length

### LINKED ISSUE #2002: `nodediscoverls` output limits `serial` length
The command `nodediscoverls` displays just the first 13 characters of the node serial in it's output. This truncation is not useful at all, given that it is common to find serial numbers longer than 13 digits (I've got servers with 15 character serials).

In such a situation, it might be better to change the output to display an ellipsis (`...`) followed by the last 10 characters of the serial, as that is where the variance lies when discovering new nodes, i.e. instead of `1234567890ABCDEF` being truncated to `1234567890ABC`, it might be truncated to `...7890ABCDEF`.

This issue follows through from Issue #78 and MR #79.

penguhyang:
fix #2002


immarvin:
agree to merge

Issue #1936, opened 2016-10-05
genimage ctl-c leaves mount points active


ralphbellofatto:
on our setup we ran:

```
genimage rh72-op5-garison-nb

ctl-c (break out genimage).
```

After the job was exited we got:

```
grep -i install /proc/mounts
/dev/mapper/xcat_data-install /install xfs rw,relatime,attr2,inode64,noquota 0 0
proc /install/netboot/rhels7.2/ppc64le/rh72-op5-garison-nb/rootimg/proc proc rw,relatime 0 0
sysfs /install/netboot/rhels7.2/ppc64le/rh72-op5-garison-nb/rootimg/sys sysfs rw,relatime 0 0
```

If any of the postinstall scripts add additional mount points, very bad things happen with when the rmimage is done.

For example, we had an incident where there was this mount point left active after a genimage ctl-c by a postinstall script

```
install  /install/netboot/rhels7.2/ppc64le/rh72-op5-garison-nb/rootimg/install install rw,relatime 0 0
```

Where the /install directory directory was mounted the rootimage directory.  When rmimage was done, then entire contents of the install /install were deleted.

Also, mount points being left can also render the system unable to re-create the directory again, when genimage is run again, the system returns a failure when genimage attempts to create the root directory.  This is due to stale file handles left behind in active mount points.  The only way we were able to recover from this was to reboot the xcat management node.

We believe that this could be fixed with a simple script in the rmimage program:

```
grep $installroot /proc/mounts | perl -nle '/(\S+) (\S+)/ && print "$2"' | while read -r dir
do
logInfo "$thisscript Unmounting $dir"
umount -rl $dir
done
```

where installroot is pointed to the rootimg directory of the osimage.  We use this script in an exit trap in our postinstall scripts to avoid this problem, but it turned out that ctl-c during the genimage can sometimes bypass this, with rather catastrophic resutls.

immarvin:
hi @ralphbellofatto , would you please provide the xCAT release you are using(``lsxcatd -v``)?

in fact we have signal handler in ``genimage`` in the code,  we need to recreate this issue in our env, this is a bug for genimage if the issue exists in latest xCAT release.

ralphbellofatto:
lsxcatd -v
Version 2.12.3 (git commit ee11ecddb7e3c323b11857826a072303ae5a7b20, built Thu Sep 22 06:16:00 EDT 2016)

immarvin:
[...]

PR #2036, opened 2016-10-26
add lock in rh/genimage to avoid multiple genimage processes on same rootimg directory;add guard code before umount/remove mount points under rootimg directory

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1936
fix https://github.com/xcat2/xcat-core/issues/1767
1. add lock in rh/genimage to avoid multiple genimage processes on same rootimg directory;
2. add guard code for umount/remove mount points under rootimg directory

### DIFF
--- a/xCAT-server/lib/xcat/plugins/packimage.pm
+++ b/xCAT-server/lib/xcat/plugins/packimage.pm
@@ -37,6 +37,8 @@ use File::Path;
 use xCAT::Utils;
 use xCAT::TableUtils;
 use xCAT::SvrUtils;
+use Digest::MD5 qw(md5_hex);
+
 Getopt::Long::Configure("bundling");
 Getopt::Long::Configure("pass_through");
 
@@ -97,7 +99,8 @@ sub process_request {
     my $provmethod;
     my $help;
     my $version;
-
+    my $lock;  
+ 
     GetOptions(
         "profile|p=s" => \$profile,
         "arch|a=s"    => \$arch,
@@ -198,6 +201,14 @@ sub process_request {
         $destdir = "$installroot/netboot/$osver/$arch/$profile";
     }
     $rootimg_dir = "$destdir/rootimg";
+
+    
+    my $retcode;
+    ($retcode,$lock)=xCAT::Utils->acquire_lock_imageop($rootimg_dir);
+    if($retcode){
+        $callback->({ error => ["$lock"], errorcode => [1]});
+        return 1;
+    }
 
     my $distname = $osver;
     until (-r "$::XCATROOT/share/xcat/netboot/$distname/" or not $distname) {
--- a/xCAT-server/lib/xcat/plugins/rmimage.pm
+++ b/xCAT-server/lib/xcat/plugins/rmimage.pm
@@ -10,9 +10,11 @@ BEGIN
 use lib "$::XCATROOT/lib/perl";
 use Getopt::Long;
 use File::Path;
+use Cwd qw(realpath);
 use xCAT::Utils;
 use xCAT::TableUtils;
 use xCAT::DBobjUtils;
+use Digest::MD5 qw(md5_hex);
 
 Getopt::Long::Configure("bundling");
 Getopt::Long::Configure("pass_through");
@@ -41,7 +43,8 @@ sub process_request {
     my $xcatdef;
     my $imagename;
     my $imagedir;
-
+    my $lock;
+    
     if (!xCAT::Utils->isLinux()) {
         $callback->({ error => ["The rmimage command is only supported on Linux."], errorcode => [1] });
         return;
@@ -175,8 +178,17 @@ sub process_request {
         $callback->({ error => ["Image directory $imagedir does not exist"], errorcode => [1] });
         return;
     }
+    
 
-    # Doing this extra check now because we now have a method and arch from either the node or the image name.
+    my $retcode;
+    ($retcode,$lock)=xCAT::Utils->acquire_lock_imageop("$imagedir/rootimg");
+    if($retcode){
+        $callback->({ error => ["$lock"], errorcode => [1]});
+        return 1;
+    }
+
+
+    # Doing this extra check now because we now have a method and arch from either the node or the image namee
     if (($method eq "sysclone") && ($arch ne "s390x")) {
 
         # Only supporting removing sysclone images for s390x at this time.
@@ -184,17 +196,28 @@ sub process_request {
         return;
     }
 
-    my @filestoremove = ("$imagedir/rootimg.gz", "$imagedir/kernel", "$imagedir/initrd-stateless.gz", "$imagedir/initrd-statelite.gz");
+    my @filestoremove = ("$imagedir/kernel", "$imagedir/initrd-stateless.gz", "$imagedir/initrd-statelite.gz");
+    my @rootimgtars=glob "$imagedir/rootimg.{tar,cpio}.{xz,gz}"; 
+    push @filestoremove,@rootimgtars;
+    #unmount all the mount points under rootimg directory 
+    #to avoid removing the directory/files on management node by mistake  
+    $realimagedir=realpath("$imagedir/rootimg");
+    my @mntptlist;
+    my $FILEHD=undef;
+    open($FILEHD,"<","/proc/mounts");
+    while(<$FILEHD>){
+        my @arr=split / /;
+        push @mntptlist,$arr[1] if(substr($arr[1],0,length($realimagedir)) eq $realimagedir);
+    }
+    close($FILEHD);
+    foreach my $mntpt (@mntptlist){
+        system("umount -l $mntpt >/dev/null 2>&1")
+    }
 
-    #some rpms like atftp mount the rootimg/proc to /proc, we need to make sure rootimg/proc is free of junk
-    `umount -l $imagedir/rootimg/proc 2>&1 1>/dev/null`;
-
-    # also umount the rootimg/sys
-    `umount -l $imagedir/rootimg/sys 2>&1 1>/dev/null`;
 
     # umount the rootimg/dev
-    my $devmount = `cat /proc/mounts |grep  "$imagedir/rootimg/dev"`;
-    if ($devmount) {
+    my $devmount = system("findmnt $imagedir/rootimg/dev/ >/dev/null 2>&1");
+    if (!$devmount) {
         xCAT::Utils->runcmd("umount -l  $imagedir/rootimg/dev");
         if ($?) {
             $callback->({ error => ["$imagedir/rootimg/dev mount on /dev, and can't umount. remove $imagename will lead to unpredictable result, please umount manualy before try again"], errorcode => [1] });
--- a/xCAT-server/share/xcat/netboot/rh/genimage
+++ b/xCAT-server/share/xcat/netboot/rh/genimage
@@ -19,6 +19,8 @@ use File::Temp qw/mkdtemp/;
 use FindBin;
 use lib "$FindBin::Bin/../imgutils";
 use imgutils;
+use xCAT::Utils;
+use Digest::MD5 qw(md5_hex);
 
 #use strict;
 Getopt::Long::Configure("bundling");
@@ -66,6 +68,8 @@ my $tempfile;
 my $prompt;
 my $ignorekernelchk;
 my $noupdate;
+my $lock;
+
 
 
 sub xdie {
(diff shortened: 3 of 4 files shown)


daniceexi:
@immarvin Good enhancement.

All of xcat commands do not support ``lock`` that avoid to be run in parallel. Could we make this ``lock`` mechanism code to be a general subroutines in the Utils.pm so that other cmd also can use it.

xCAT-server/lib/xcat/plugins/rmimage.pm
```diff
@@ -185,12 +186,20 @@ sub process_request {
     }
 
     my @filestoremove = ("$imagedir/rootimg.gz", "$imagedir/kernel", "$imagedir/initrd-stateless.gz", "$imagedir/initrd-statelite.gz");
+    
+    #unmount all the mount points under rootimg directory 
+    #to avoid removing the directory/files on management node by mistake  
+    $realimagedir=realpath($imagedir);
+    my @mntptlist=qx(cat /proc/mounts |awk -F' ' '{print \$2}'|grep "$realimagedir" 2>/dev/null);
```
neo954: Since we are writing perl code here. I suggest do not use external command here. This makes the code more complex and less efficiency. Please use perl to do things here.

That is something like the following ...

    open FILE, "<", "/proc/mounts" or die;
    while (<FILE>) {
        my @arr = split / /;
        print $arr[1] if (substr($arr[1], 0, length($realimagedir)) eq $realimagedir);
    }
    close FILE;

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -66,6 +66,38 @@ my $tempfile;
...
+#0: lock acquired
+#1: acquire lock failed due to some system error
+#2: acquire lock failed due to the lock is not release yet 
+sub lock {
+    my $lockinfo=undef;
+    my $ret=0;
+    open($lockinfo,">", $lockinfofn) or return 1;
+    print $lockinfo  "$$";
+    close($lockinfo);
+    if( -l $lockfile){
```
neo954: Should check if `$lockfile` is `undef` first. Or maybe, `$lockfile` should be passed as an argument of this sub-function.

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -66,6 +66,38 @@ my $tempfile;
...
+      open($lockinfo,"<",$lockfile ) or return 1;
+      $exps=<$lockinfo>;
+      close($lockinfo);
+      system qq(ps --pid $exps --format command h|grep -i genimage) or return 2;
```
neo954: Please do not use external command, but use `kill SIGZERO $exps;` to check if the process exist.

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -66,6 +66,38 @@ my $tempfile;
...
+      unlink($lockfile);
```
neo954: Why delete the file `$lockfile` here. This is not right.

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -66,6 +66,38 @@ my $tempfile;
...
+#1: acquire lock failed due to some system error
+#2: acquire lock failed due to the lock is not release yet 
+sub lock {
+    my $lockinfo=undef;
+    my $ret=0;
+    open($lockinfo,">", $lockinfofn) or return 1;
+    print $lockinfo  "$$";
+    close($lockinfo);
+    if( -l $lockfile){
+      open($lockinfo,"<",$lockfile ) or return 1;
```
neo954: If the file `$lockfile` does not exist, the program should move on, not `return  1` here.

xCAT-server/share/xcat/netboot/rh/genimage
neo954: And if the process `$exps` does not exist here, the program should on, not `return 2` here.

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -66,6 +66,38 @@ my $tempfile;
...
+      system qq(ps --pid $exps --format command h|grep -i genimage) or return 2;
+      unlink($lockfile);
+    }
+    if ($hassymlink) {
+        $ret=symlink($lockinfofn, $lockfile) ;
+       } else{
+        $ret=link($lockinfofn, $lockfile);
+    } 
+    $ret=($ret==0) ? 1:0;
+    return $ret;
```
neo954: In this case, `return not $ret;` will do the work.

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -112,17 +144,26 @@ sub mount_chroot {
...
+
+    $rootimage_dir=realpath($rootimage_dir);
+    if ($rootimage_dir eq "/") {
         print "\n\n\n\nWARNING: /proc and /sys may still be mounted in the rootimgdir\n\n\n\n";
         return 1;
     }
+    
+    #unmount all the mount points under rootimg directory
+    #to avoid removing the direcoty/files on managemment node by mistake
+    my @mntptlist=qx(cat /proc/mounts |awk -F' ' '{print \$2}'|grep "$rootimage_dir" 2>/dev/null);
```
neo954: Same as the comment above, please use perl codes to do the work.

xCAT-server/lib/xcat/plugins/rmimage.pm
neo954: Consider the `$realimgedir` is `/path/to/somewhere`. But in `/proc/mounts` there is a line `/some/other/dir/path/to/somewhere/foo/bar`. In this case, use `grep` is not really good enough. That is why I use `substr()` and string comparison in the example code above.

neo954: And please also consider if there are `/path/to/somewhere` and `/path/to/somewhere-else`, then, while you matching `/path/to/somewhere`, try not `umount`ing subdirectories under `/path/to/somewhere-else`

gurevichmark:
Since this pull request makes changes to the rmimage.pm, will this be the good place to also fix the problem that `rmimage` command does not remove the `rootimg.cpio.gz` file ? I believe during one of our conference call it was agreed that this command should do that.

xCAT-server/lib/xcat/plugins/rmimage.pm
```diff
@@ -185,12 +186,20 @@ sub process_request {
...
+    foreach my $mntpt (@mntptlist){
+        system("umount -l $mntpt >/dev/null 2>&1")
+    }
 
     #some rpms like atftp mount the rootimg/proc to /proc, we need to make sure rootimg/proc is free of junk
-    `umount -l $imagedir/rootimg/proc 2>&1 1>/dev/null`;
+    #`umount -l $imagedir/rootimg/proc 2>&1 1>/dev/null`;
```
whowutwut: Should delete these lines instead of comment out.

immarvin:
hi @whowutwut @gurevichmark @daniceexi @neo954 , all the comments have been processed

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -348,6 +366,18 @@ unless ($onlyinitrd) {
...
     }
 
+    my $mylockfile=realpath($rootimg_dir); 
+    my $mymd5=md5_hex($mylockfile);
+    $mylockfile=~s/\//./g;
+    my $mylockfile=$mylockfile."."."$mymd5";
+    my $NON_BLOCK=1;
+    $lock = xCAT::Utils->acquire_lock("$mylockfile", $NON_BLOCK);
+    unless ($lock){
+        my $pid=qx(lsof -Fp /var/run/lock/xcat/$mylockfile 2>/dev/null|tr p " ");
```
neo954: Package `lsof` is not installed by default. If you insist use `lsof` here, you have to put package `lsof` into the dependency list of package `xCAT-server`.

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -112,17 +116,30 @@ sub mount_chroot {
...
+    }
+    
     if (majversion($osver) > 6) {
-        system("umount -l $rootimage_dir/proc");
-        system("umount -l $rootimage_dir/sys");
-
-        #system("umount -l $rootimage_dir/dev");
-        system("rm -rf $rootimage_dir/dev/*");
+        #only remove the /dev in rootimg directory if it is not a mount point
+        system("findmnt $rootimage_dir/dev/ >/dev/null 2>&1 || rm -rf $rootimage_dir/dev/*");
```
neo954: Even if udev is used, remove everything under the `/dev` directory may cause problem. At lease `/dev/null`, `/dev/zero` should be there.

immarvin: the amount will occur on genimage exit or interrupted, no following operation will be performed on rootimg,  I think it's ok to remove the /dev/null and /dev/zero at this time

xCAT-server/lib/xcat/plugins/packimage.pm
```diff
@@ -199,6 +202,21 @@ sub process_request {
     }
     $rootimg_dir = "$destdir/rootimg";
 
+    my $mylockfile=Cwd::realpath($rootimg_dir);
+    my $mymd5=md5_hex($mylockfile);    
+    $mylockfile=~s/\//./g;
+    my $mylockfile=$mylockfile.".".$mymd5;
```
neo954: I still prefer let `$mylockfile = "../../.." . $destdir . "/.genimage.lock";`

xCAT-server/lib/xcat/plugins/rmimage.pm
```diff
@@ -175,6 +178,22 @@ sub process_request {
...
+    my $NON_BLOCK=1;
+    $lock = xCAT::Utils->acquire_lock("$mylockfile", $NON_BLOCK);
+    unless ($lock){
+        my $pidfd;
+        open($pidfd,"<","/var/run/lock/xcat/$mylockfile");
+        my $pid=<$pidfd>;
+        close($pidfd);
+        $callback->({ error => ["failed to acquire lock, seems there is another genimage/packimage/rmimage process $pid running on root image dir \"$rootimg_dir\""], errorcode => [1] });
+        return 1;
+    }
```
neo954: There are identical codes in `packimg.pm`, `rmimage.pm` and `genimage`. I think there is no point to write the same code three times.

I suggest made these codes a function, and reuse the same code in three different places.

xCAT-server/share/xcat/netboot/rh/genimage
```diff
@@ -348,6 +366,21 @@ unless ($onlyinitrd) {
...
+    $lock = xCAT::Utils->acquire_lock("$mylockfile", $NON_BLOCK);
+    unless ($lock){
+        my $pidfd;
+        open($pidfd,"<","/var/run/lock/xcat/$mylockfile");
+        my $pid=<$pidfd>;
+        close($pidfd);
+        print "genimage: failed to acquire lock, seems there is another genimage/packimage/rmimage process $pid running on root image dir \"$destdir\" ";
+        exit 1;
+    }
+
```
neo954: It is worth to port the "lock" feature to Debian/Ubuntu as well.

xCAT-server/lib/xcat/plugins/packimage.pm
immarvin: we  have talked about this offline, the `/var/run/lock/xcat/` is the conventional place of locks in xCAT

xCAT-server/share/xcat/netboot/rh/genimage
immarvin: will open another issue on this, do not spare time on the unit test on sles and ubuntu this PR

immarvin: traced wth the ticket https://github.com/xcat2/xcat-core/issues/2077

xCAT-server/lib/xcat/plugins/rmimage.pm
immarvin: good idea, but there is a problem that the lock will be released outside the namespace which  "$lock" in,  in the following code slice
````
$lock = xCAT::Utils->acquire_lock("$mylockfile", $NON_BLOCK);
````

with this problem, we'd better not to abstract this code slice as a subroutine.
I leave this as a issue https://github.com/xcat2/xcat-core/issues/2078

neo954:
I believe this patch is good enough to merge. Great works, @immarvin.

I will merge this one.

immarvin:
hi @daniceexi , there is already some lock subroutines(acquire_lock,....) in Utils.pm,  I leveraged the subroutine in this PR.

These are several issues in this subroutine, I fixed them in this  PR.

I added the lock for packimage, rummage and rh/genimage in this PR. the genimage on sles and ubuntu, imgimport, imgexport and liteimg  are left in ticket https://github.com/xcat2/xcat-core/issues/2077.

hi @gurevichmark , I have also fixed the issue that rmimage can not find rootimg.gz you mentioned in this PR

immarvin:
@neo954 raised a lot of valuable comments on this PR to help refine the code, and provided a lot of support on technical details and suggestion on debug, thanks a lot

daniceexi:
@immarvin Thanks 👍

daniceexi:
@neo954 Good work. Thanks 👍

PR #2038, opened 2016-10-26
Xcat client updates

chuckbrazie:


daniceexi:
@chuckbrazie I have a little concern about this change.  xCAT is trying to deprecate the '-o/-p' in the xCAT command like 'nodeset','genimage' but using the ``osimage`` object to specify the image attributes. Does zVM still use the ``-o/-p`` a lot?

@immarvin Please track this discussion.

PR #2042, opened 2016-10-27
update to add xCAT-probe to PCM build (#2041)

tingtli:
@immarvin Please help merge


immarvin:
agree to merge

tingtli:
@whowutwut  The pcm branch is requested by PCM team. They may add some fix in this branch for their purpose (maybe this time they didn't).   xiao peng told me that we could not use capital word in the branch name because there probably has some issues ... so I use pcm not PCM in the branch name.

whowutwut:
Ok thanks

Issue #2037, opened 2016-10-26
changing xcat from to postgres threw away node switch information.


ralphbellofatto:
While closing out this issue we found a problem with xcat:
https://gitlabhost.rtp.raleigh.ibm.com/4A8491897/research-92-cluster/issues/141

We migrated our xcat db from SQLite to postgres by followed the directions here:

http://xcat-docs.readthedocs.io/en/stable/advanced/hierarchy/databases/postgres_configure.html 

specifically we executed:

```
1) yum install postgresql*
yum install perl-DBD-Pg

2) pgsqlsetup -i -V  (-V will spit out verbose output, looking at trimming this down some..) 

That should automatically migrate the existing SQLite DB to PostgreSQL 

1:59:21 PM: Will look like this: 

[root@fs4 log]# lsxcatd -a
Version 2.12.4 (git commit 2bf96ee97df3e3a04ef72c42f2e756af22ecad5a, built Tue Oct 25 11:15:48 EDT 2016)
This is a Management Node
cfgloc=Pg:dbname=xcatdb;host=192.168.3.25|xcatadm
dbengine=Pg
dbname=xcatdb
dbhost=192.168.3.25
dbadmin=xcatadm

[root@fs4 log]# cat /etc/xcat/cfgloc 
Pg:dbname=xcatdb;host=192.168.3.25|xcatadm|XXXXX
```

When we were done, we compared the output of the lsdef -l command with what it was prior to doing the update and we found the following two fields are missing on every node:

```
switch=c460tors01
switchport=3
```

The following attachment contains the before(orig) and after(new) node definitions extracted by xcat.
[xcat-postgres-migrate.zip](https://github.com/xcat2/xcat-core/files/554187/xcat-postgres-migrate.zip)

ralphbellofatto:
NOTE:
We attempted to restore the above configuration with
cat node.lsdef | chdef -z

And it failed to update the switch ports...

We extracted the data by hand from the lsdef file and then constructed a script as follows:
```
chdef c460c001 switch=c460tors01 switchport=3
chdef c460c002 switch=c460tors01 switchport=4
chdef c460c003 switch=c460tors01 switchport=5
chdef c460c004 switch=c460tors01 switchport=6
chdef c460c005 switch=c460tors01 switchport=7
```

And then the data stuck...

ralphbellofatto:
reopening...

whowutwut:
[...]

Issue #2007, opened 2016-10-21
Setting "switch" attribute by chdef command is ineffective


xuweibj:
MN: 10.3.5.4

Environment:

```
# lsxcatd -a
Version 2.12.4 (git commit 2b7fa190053e67d9cbf3e838fe6544ff3206a9d8, built Thu Oct 20 06:16:09 EDT 2016)
This is a Management Node
cfgloc=Pg:dbname=xcatdb;host=10.3.5.4|xcatadm
dbengine=Pg
dbname=xcatdb
dbhost=10.3.5.4
dbadmin=xcatadm
```

Output as follows:

```
# chdef discoverytest3 switch=switchtest
1 object definitions have been created or modified.
# lsdef discoverytest3
Object name: discoverytest3
    groups=test
    mtm=test
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfilesk,setupntp
    serial=0001
```

But if set 'switch' and 'switchport' at the same time, it's OK.

```
# chdef discoverytest3 switch=switchtest switchport=7
1 object definitions have been created or modified.
# lsdef discoverytest3
Object name: discoverytest3
    groups=test
    mtm=test
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfilesk,setupntp
    serial=0001
    switch=switchtest
    switchport=7
```

immarvin:
is this a similar issue with https://github.com/xcat2/xcat-core/issues/2037 ?

whowutwut:
Interesting. This might be the same...

cxhong:
yeh,  I told Ralph about this issue,  we think this is similar

xuweibj:
Verified on `Version 2.12.4 (git commit 6450ef2ab6c290389f650f304b5951bd2659d2a6, built Sat Nov  5 06:15:51 EDT 2016)`, so close.

PR #2045, opened 2016-10-27
Set value to '' if value within composite primary keys not defined

chenglch:
As history reasons, null value is always set within the composite primary
keys, this patch is just a work aroud for the issue encountered on postgres.
Correct fix is to report error to the client side, but too much error handler
is missing in xcat code

close-issue: #2037
close-issue: #2007


whowutwut:
I am ok with this change,  I think it's ready to merge in right?

PR #2046, opened 2016-10-27
Add sslversion and ubuntu migration cases to bundle

caomengmeng:
@tingtli 
please help review the bundle files:
1. add sslversion test cases
2. add ubuntu test cases


tingtli:
We add more cases in bundle. I agree to merge

Issue #1792, opened 2016-09-01
copycds report no error when it is not complete successfully. 


wwzhubj:
I use : 
 copycds -p /install/rh73snapshot1 -n rhels73 /root/Downloads/RHEL-7.3-20160825.1-Server-ppc64le-dvd1.iso
Copying media to /install/rh73snapshot1

Normally, after it completed, lsdef -t osimage can show the template osimage created for snapshot1. but I can not find.  Which means the copycds haven't ran successfully. 
In this condition, the user should get some error / failure message. 
the version I use: 
[root@c712ems4 install]# lsxcatd -v
Version 2.12.2 (git commit e4a7359d31008d81c4c57e0857d5f09009a3269c, built Thu Aug 18 06:43:29 EDT 2016)

immarvin:
please use "-n rhels7.3"

wwzhubj:
yes, I know "-n rhels7.3" can work. But there are 2 issues: 
1. why  "rhels73 " will fail? 
2. when it fails, why the copycds shows no any error or failure message to user?

immarvin:
"-n" should only be used when the xCAT cannot recognise the iso, this is obviously not the case here. in fact, without "-n", xcat can recognise the iso and "copycds" works well. 

It is not a error for xcat, you just specified an arbitrary os name that xcat does not know what it is, so xcat cannot  create the osimage for it.

whowutwut:
@wwzhubj  I'm not too clear about the issue you encountered.  Can you please provide some proof in the issue of the osimages after running copycds?  I ran the same command and it succeeded for me on my test machine... 

```
[root@fs1 RH7.3beta]# copycds -p /install/rh73snapshot1 -n rhels73 ./RHEL-7.3-20160825.1-Server-ppc64le-dvd1.iso 
Copying media to /install/rh73snapshot1
Media copy operation successful

[root@fs1 ~]# lsdef -t osimage | grep rhels73
rhels73-ppc64le-install-compute  (osimage)
rhels73-ppc64le-install-service  (osimage)
rhels73-ppc64le-netboot-compute  (osimage)
```

Using "-n rhels73" is OK, we still are able to pick up the rhels7 as the default.   However, by not using the "7.3" convention, if xCAT ship a specific template for 7.3, you may default to the rhels7 template file which *may* cause issues down the road. 

```
[root@fs1 ~]# lsdef -t osimage -o rhels73-ppc64le-install-compute
[...]

PR #2048, opened 2016-10-27
Better message wording for osimagecheck probe

gurevichmark:
As suggested by #1792 fixing message wording. Now displays:

```
[root@fs1 subcmds]# xcatprobe osimagecheck
No diskless osimages with identical root image directories were found.                                       [ OK ]
Default 'template' value is being used by compute_1638i                                                      [WARN]
  If this image was generated by 'copycds -n <os version>', make sure the <os version> specified
  follows xCAT naming convention (for example 'rhels7.2')
[root@fs1 subcmds]#
```


immarvin:
agree to merge

Issue #1947, opened 2016-10-10
[fvt]2.12.4 switch based switch discovery could not set config on System RackSwitch G8124 and G8264-T


junxiawang:
env：fsl
build:latest build
How to reproduce:

```
Switch G8264-T:
[root@fs1 ~]# switchdiscover --range 192.168.5.41 -s snmp --setup
Discovering switches using snmpwalk for 192.168.5.41 .....
ip name vendor mac
------------ ------------ ------------ ------------
192.168.5.41 switch-192-168-5-41 IBM Networking Operating System RackSwitch G8264-T 6c:ae:8b:d1:82:fe
call to config BNT switches frame31sw1

output = change frame31sw1 to static ip address
switch ssh username or password is not define, add default one
Setup SNMP server for frame31sw1
Trying 192.168.5.163...

telnet: connect to address 192.168.5.163: No route to host
Failed expect command 3:Child PID 128582 exited with status 256


[root@fs1 ~]# switchdiscover --range 192.168.5.163 -s snmp
Discovering switches using snmpwalk for 192.168.5.163 .....
No switch found 
Switch G8124:
[root@fs1 ~]# makehosts switch-192-168-5-165
[root@fs1 ~]# chdef switch-192-168-5-165 switch=switch-10-5-23-1 switchport=32
1 object definitions have been created or modified.
[root@fs1 ~]# vi /etc/hosts
[root@fs1 ~]# switchdiscover --range 192.168.5.165 -s snmp --setup
Discovering switches using snmpwalk for 192.168.5.165 .....
ip name vendor mac
------------ ------------ ------------ ------------
192.168.5.165 switch-192-168-5-165 IBM Networking Operating System RackSwitch G8124 fc:cf:62:10:04:fe
call to config BNT switches switch-192-168-5-165

output = change switch-192-168-5-165 to static ip address
finish setup static ip address for switch-192-168-5-165
switch ssh username or password is not define, add default one
Setup SNMP server for switch-192-168-5-165
telnet: switch-192-168-5-165: Name or service not known
switch-192-168-5-165: Unknown host



```

zet809:
@cxhong Hi, Casandra, do you have plan to support those kinds of switch in this release? If not, will you pls move it to the next release?

whowutwut:
@junxiawang  This issue is pending, can you verify and if OK close out?

PR #2049, opened 2016-10-29
Configure different model of BNT switches

cxhong:
for issue #1947 

1) Add Configuration for BNT model G8264-T and G8124.
2) match discover switch to predefined switch as default option for switchdiscover command

### DIFF
--- a/xCAT-server/lib/xcat/plugins/switchdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/switchdiscover.pm
@@ -326,7 +326,7 @@ sub process_request {
     my $req      = shift;
     my $callback = shift;
     my $sub_req  = shift;
-  
+
     ###########################################
     # Build hash to pass around
     ###########################################
@@ -472,19 +472,20 @@ sub process_request {
             my $msg = sprintf $format, $ip, $name, $vendor, $mac;
             send_msg(\%request, 0, $msg);
         }
+        send_msg(\%request, 0,"\n");        
     }
 
+    my ($discoverswitch, $predefineswitch) = matchPredefineSwitch($result, \%request, $sub_req);
 
     # writes the data into xCAT db
     if (exists($globalopt{w})) {
         send_msg(\%request, 0, "Writing the data into xCAT DB....");
-        xCATdB($result, \%request, $sub_req);
+        xCATdB($discoverswitch, \%request, $sub_req);
     }
 
     if (exists($globalopt{setup})) {
-        switchsetup($result, \%request, $sub_req);
+        switchsetup($predefineswitch, \%request, $sub_req);
     }
-
 
     return;
 }
@@ -1199,7 +1200,6 @@ sub xCATdB {
             $mac=" ";
         }
 
-
         #################################################
         # use lsdef command to check if this switch is
         # already in the switch table
@@ -1369,13 +1369,9 @@ sub format_xml {
 }
 
 #--------------------------------------------------------------------------------
-=head3  switchsetup
+=head3 matchPredefineSwitch 
       find discovered switches with predefine switches
       for each discovered switches:
-      1) matching mac to a predefined node
-      2) if get predefined node, config the discovered switch, if failed, update
-         'otherinterface' attribute of predefined node
-      3) remove hosts record and node definition for the discovered switch
     Arguments:
       outhash: a hash containing the switches discovered
     Returns:
@@ -1383,14 +1379,13 @@ sub format_xml {
 =cut
 #--------------------------------------------------------------------------------
 
-sub switchsetup {
+sub matchPredefineSwitch {
     my $outhash = shift;
     my $request = shift;
     my $sub_req = shift;
-    my @switchnode = ();
+    my $discoverswitch;
+    my $configswitch;
     my $static_ip;
-    my $discover_switch;
-    my $nodes_to_config;
 
     #print Dumper($outhash);
     my $macmap = xCAT::MacMap->new();
@@ -1409,7 +1404,10 @@ sub switchsetup {
 
         my $node = $macmap->find_mac($mac,0,1);
         if (!$node) {
-            send_msg($request, 0, "NO predefined switch matched this switch $dswitch with ip address $ip and mac address $mac");
+            send_msg($request, 0, "Switch discovered: $dswitch ");
+            $discoverswitch->{$mac}->{ip} = $ip;
+            $discoverswitch->{$mac}->{vendor} = $vendor;
+            $discoverswitch->{$mac}->{name} = $dswitch;
             next;
         }
 
@@ -1417,14 +1415,29 @@ sub switchsetup {
         $static_ip = xCAT::NetworkUtils->getipaddr($node);
         my $stype = get_switchtype($vendor);
 
-        if (exists($globalopt{verbose}))    {
-            send_msg($request, 0, "Found Discovery switch $dswitch, $ip, $mac with predefine switch $node, $static_ip $stype switch\n" );
-        }
+        send_msg($request, 0, "Switch discovered and matched: $dswitch to $node" );
+
         xCAT::Utils->runxcmd({ command => ['chdef'], arg => ['-t','node','-o',$node,"otherinterfaces=$ip",'status=Matched',"mac=$mac","switchtype=$stype","usercomment=$vendor"] }, $sub_req, 0, 1);
 
-        push (@{$nodes_to_config->{$stype}}, $node);
+        push (@{$configswitch->{$stype}}, $node);
     }
 
+    return ($discoverswitch, $configswitch);
+}
+
+#--------------------------------------------------------------------------------
+=head3 switchsetup 
+      configure the switch 
+    Arguments:
+      outhash: a hash containing the switches need to configure
+    Returns:
+      result:
+=cut
+#--------------------------------------------------------------------------------
+sub switchsetup {
+    my $nodes_to_config = shift;
+    my $request = shift;
+    my $sub_req = shift;
     foreach my $mytype (keys %$nodes_to_config) {
         my $config_script = "$::XCATROOT/share/xcat/scripts/config".$mytype;
         if (-r -x $config_script) {
--- a/xCAT-server/lib/xcat/plugins/switchdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/switchdiscover.pm
@@ -771,7 +771,7 @@ sub nmap_scan {
     if ($osguess_ips) {
         my $guess_switches = nmap_osguess($request, $osguess_ips);
         foreach my $guess_mac ( keys %$guess_switches ) {
-            $switches->{$guess_mac}->{ip} = $guess_switches->{$guess_mac}->{ip};;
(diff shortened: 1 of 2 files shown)


cxhong:
@zet809 or @hu-weihua , Can u review this PR?

zet809:
@cxhong Sorry, cassandra, do you think we can move this Pull Request for the next release since I won't be able to finish reviewing in this week?

cxhong:
sure,  I will remove it.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -727,28 +737,42 @@ sub nmap_scan {
...
-                                $switches->{$mac}->{ip} = $ip;
-                                $switches->{$mac}->{vendor} = $vendor;
-                                $switches->{$mac}->{name} = $host->{hostname};
                                 $found = 1;
-                                if (exists($globalopt{verbose}))    {
-                                    send_msg($request, 0, "FOUND Switch: ip=$ip, mac=$mac, vendor=$vendor\n");
-                                }
                             }
                         } 
                         if ( ($found == 0) && ($type eq "mac") ) {
```
zet809: Shall not the $found initialized for each $host before line 724 ( foreach my $addr ( @$addr_ref ) { )?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -727,28 +737,42 @@ sub nmap_scan {
...
-                        } # end nmap osscan command
+                            my $search_string = join '|', keys(%global_mac_identity);
+                            if ($mac =~ /($search_string)/i) {
+                                my $key = $1;
+                                $vendor = $global_mac_identity{lc $key};
+                                $found = 1;
+                            }
+
+                            # still not found, used nmap osscan command
+                            if ( $found == 0) {
```
zet809: Why do the $found check here?
Why not put it to be else of if in line 756?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -727,28 +737,42 @@ sub nmap_scan {
...
                 }
             } #end for each host
         }
     }
 
-    my $guess_switches = nmap_osguess($request, $osguess_ips);
-    foreach my $guess_mac ( keys %$guess_switches ) {
-        $switches->{$guess_mac}->{ip} = $guess_switches->{$guess_mac}->{ip};;
-        $switches->{$guess_mac}->{vendor} = $guess_switches->{$guess_mac}->{vendor};
+    if (!$osguess_ips) {
```
zet809: I think it shall be if ($osguess_ips), right?

xCAT-server/share/xcat/scripts/configBNT
```diff
@@ -118,12 +118,28 @@ sub config_ip {
     my @config_switches;
     my @discover_switches;
     my $nodetab = xCAT::Table->new('hosts');
-    my $nodehash = $nodetab->getNodesAttribs(\@nodes,['ip','otherinterfaces']);
+    my $nodehash = $nodetab->getNodesAttribs(\@nodes,['ip','otherinterfaces','mac']);
```
zet809: There is no mac column in hosts table.

xCAT-server/share/xcat/scripts/configBNT
```diff
@@ -118,12 +118,28 @@ sub config_ip {
...
+    }
+    #get mac address for the switches
+    my $mactab = xCAT::Table->new("mac");
+    my $machash = $mactab->getNodesAttribs(\@nodes,['mac']);
+
     foreach my $switch (@nodes) {
         print "change $switch to static ip address\n";
+
+        #makesure host is in the /etc/hosts
+        $cmd = "makehosts $switch";
```
zet809: Why not run "makehosts \@nodes" before the loop?

xCAT-server/share/xcat/scripts/configBNT
```diff
@@ -161,16 +179,45 @@ sub config_ip {
         $cmd = "makehosts $dswitch";
         $rc= xCAT::Utils->runcmd($cmd, 0);
 
-        # verify if xdsh works
-        $cmd = "xdsh $dswitch --devicetype EthSwitch::BNT 'enable;configure terminal;exit' ";
+        #get netmask
```
zet809: The $dswitch is empty in line 168, and then get a new string in line 175, why?

xCAT-server/share/xcat/scripts/configBNT
```diff
@@ -118,12 +118,28 @@ sub config_ip {
...
+    my $nodehash = $nodetab->getNodesAttribs(\@nodes,['ip','otherinterfaces','mac']);
+    # get netmask from network table
+    my $nettab = xCAT::Table->new("networks");
+    my @nets;
+    if ($nettab) {
+        @nets = $nettab->getAllAttribs('net','mask');
+    }
+    #get mac address for the switches
+    my $mactab = xCAT::Table->new("mac");
+    my $machash = $mactab->getNodesAttribs(\@nodes,['mac']);
```
zet809: Every configuring sub function need the mac address to judge the switch type, why not put it out of the sub functions and get the switch type before go into them?

xCAT-server/share/xcat/scripts/configBNT
```diff
@@ -161,16 +179,45 @@ sub config_ip {
         $cmd = "makehosts $dswitch";
         $rc= xCAT::Utils->runcmd($cmd, 0);
 
-        # verify if xdsh works
-        $cmd = "xdsh $dswitch --devicetype EthSwitch::BNT 'enable;configure terminal;exit' ";
+        #get netmask
+        my $mask;
+        foreach my $net (@nets) {
+            if (xCAT::NetworkUtils::isInSameSubnet( $net->{'net'}, $static_ip, $net->{'mask'}, 0)) {
+                $mask=$net->{'mask'};
```
zet809: If a mask got, why not exit the loop?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: we only support few vendor info and also relate to "mac" string in the hash.  Can't put into else statement.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: will make change

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: yeh, should be

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -1386,22 +1404,40 @@ sub switchsetup {
 
         my $node = $macmap->find_mac($mac,0,1);
         if (!$node) {
-            send_msg($request, 0, "NO predefined switch matched this switch $dswitch with ip address $ip and mac address $mac");
+            send_msg($request, 0, "Switch discovered: $dswitch ");
+            $discoverswitch->{$mac}->{ip} = $ip;
+            $discoverswitch->{$mac}->{vendor} = $vendor;
+            $discoverswitch->{$mac}->{name} = $dswitch;
```
cxhong: no, not yet, didn't write to xCATdb yet at that time

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -1386,22 +1404,40 @@ sub switchsetup {
...
             next;
         }
 
         # get predefine node ip address
         $static_ip = xCAT::NetworkUtils->getipaddr($node);
```
cxhong: right

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -1386,22 +1404,40 @@ sub switchsetup {
...
         my $stype = get_switchtype($vendor);
 
-        if (exists($globalopt{verbose}))    {
-            send_msg($request, 0, "Found Discovery switch $dswitch, $ip, $mac with predefine switch $node, $static_ip $stype switch\n" );
-        }
+        send_msg($request, 0, "Switch discovered and matched: $dswitch to $node" );
+
         xCAT::Utils->runxcmd({ command => ['chdef'], arg => ['-t','node','-o',$node,"otherinterfaces=$ip",'status=Matched',"mac=$mac","switchtype=$stype","usercomment=$vendor"] }, $sub_req, 0, 1);
```
cxhong: well, as we discussed, not all discoveryswitch are wrote to xcatdb,  if discoveryswitch found a match with predefine switch, it only updated predefine node,  and that discoveryswitch will not be write to xCATdb

xCAT-server/share/xcat/scripts/configBNT
cxhong: right, I used mac table to get mac address

xCAT-server/share/xcat/scripts/configBNT
zet809: @cxhong Hi, Casandra, you hadn't answer my question here.

xCAT-server/share/xcat/scripts/configBNT
zet809: I see, but why the IP can be resolved to a hostname, but we create a new one like switch-$ip_str, if the switch-$ip_string is same as the hostname got in line 164, we in fact do nothing, right?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: I like to keep here for debug purpose also.

xCAT-server/share/xcat/scripts/configBNT
cxhong: will change it

xCAT-server/share/xcat/scripts/configBNT
```diff
@@ -147,7 +166,7 @@ sub config_ip {
 
         # if hostnames are same, created different one for discovery name
         if ($dswitch eq $switch) {
```
zet809: For ($dswitch eq $switch), there must be error configuration in xcatdb, why do we create a new object? Shall not output an error message?

cxhong: For the normal (good) case, the $dswitch should not be in the xCATdb and hostname should not be define in the /etc/hosts.  the discovery ip address is defined in the otherinterface attribute, I used that ip address to define template $dswitch to use for xdsh command.  it will be removed after.   but in some of error case or user error,  may found $dswitch in xcatdb or /etc/hosts,   if $dswitch has same name as $switch, the $switch definition (like username, password, etc)will be modified, to avoid this,  that's why  discovery name and predefine name should not be same.  that's one of issue junxia found during her test.

xCAT-server/share/xcat/scripts/configBNT
zet809: ?

zet809:
The section that matching pre-defined switch node looks good to me, but the configuration script still need to be enhanced, so merge this pull request but keep the defect #1947 in open status.

Issue #2051, opened 2016-10-30
[fvt]xcatprobe discovery could not return right monitor information if node range invalid 


junxiawang:
env:fs4
build:[root@fs4 ~]# lsxcatd -v
Version 2.12.4 (git commit 05fdc34e19d3a6349635d3d0ae75d573c5f6f813, built Wed Oct 26 06:15:59 EDT 2016)
How to reproduce:

```
[root@fs4 junxiaw]# xcatprobe discovery -n 50.23.17.1 -m switch
All pre_defined nodes are valid                                                                                   [ OK ]
Attribute 'master' in 'site' table is configured well                                                             [ OK ]
DHCP dynamic range is configured well                                                                             [ OK ]
Genesis files are avaliable                                                                                       [ OK ]

-------------------------------------------------------------
                               ___
        ____    _  _____   _.-|   |          |\__/,|   (`\
 __  __/ ___|  / \|_   _| {   |   |          |x x  |__ _) )
 \ \/ / |     / _ \ | |    "-.|___|        _.( T   )  `  /
  >  <| |___ / ___ \| |    .--'-`-.     _((_ `^--' /_<  \
 /_/\_\\____/_/   \_\_|  .+|______|__.-||__)`-'(((/  (((/
-------------------------------------------------------------

Start capturing every message during discovery process......

Error: Invalid nodes and/or groups in noderange: 50.23.17.1
All nodes specified to monitor, have finished discovery process                                                   [ OK ]
==================discovery_probe_report=================
All nodes matched successfully                            
```

PR #2052, opened 2016-10-31
fix issue 2051&2054, update noderange check and do not print log from /install/

xuweibj:
#2051 #2054 #2060


hu-weihua:
I have reviewed the code and agree to merge, thanks @xuweibj .

whowutwut:
Just FYI, it would be nice in the Pull request description to describe the changes in more details and/or show the externals of the behavior after the change.  It also helps the reviewer understand the goal of the change.

PR #2056, opened 2016-10-31
postscripts: add --showduplicates to yum list

thiell:
"yum list" alone does NOT check the older packages available in the
yum repo. If we want to install a specific package version, the
--showduplicates option is required. As an example, with this patch,
we can now add a specific kernel package to otherpkgs.pkglist that is
older that the one available in the distro.

### DIFF
--- a/xCAT/postscripts/otherpkgs
+++ b/xCAT/postscripts/otherpkgs
@@ -715,7 +715,7 @@ while [ $op_index -le $OTHERPKGS_INDEX ]; do
 	    fi
 	        if [ $hasyum -eq 1 ]; then 
 	            #use yum
-		    result=`yum list $fn 2>&1`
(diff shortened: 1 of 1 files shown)


immarvin:
@thiell  Thanks for this pull request and your interest in the xCAT project. In order for the xCAT team to consider your changes, you must first sign a CLA, instructions are documented here: http://xcat-docs.readthedocs.io/en/stable/developers/license/contributors.html

Or if you know you are covered under a Corporate CLA that was signed, can you please give us the corporation name and we can verify. Thanks!

whowutwut:
Thanks for sending the CLA! I've added you to the contributor team for xcat2.

thiell:
@immarvin, @whowutwut: thanks! let me know what you think about this PR.

immarvin:
I am ok with the change, thanks for your contribution @thiell

PR #2064, opened 2016-11-02
move rsyslogd restart code from dhclient-script to doxcat

xuweibj:
To get log after time synchronization，move rsyslogd restart to doxcat.
The code after ntp start.

Fix issue #1788 , modified log to "Received request=$dest, will call xCAT back in $delay seconds. Discovery is complete, run nodeset on this node to provision an Operating System.

### DIFF
--- a/xCAT-genesis-scripts/bin/doxcat
+++ b/xCAT-genesis-scripts/bin/doxcat
@@ -335,8 +335,8 @@ while :; do
 		/bin/getcert $XCATMASTER:$XCATPORT
 		destiny=''
 		dest=''
-		DISCOVERYDONE=1
-		export DISCOVERYDONE
+		XCAT_DISCOVERY_WAS_DONE=1
+		export XCAT_DISCOVERY_WAS_DONE
 	elif [ "$dest" = shell ]; then
 		logger -s -t $log_label -p local4.info "Dropping to debug shell(exit to run next destiny)..."
 		destiny=''
@@ -400,7 +400,7 @@ while :; do
 		delay=$((30+$RANDOM%270))
 		while [ $delay -gt 0 ]; do
 			if [ $((delay%10)) == 0 ]; then 
-				if [ $DISCOVERYDONE == 1 ]; then
+				if [ "$XCAT_DISCOVERY_WAS_DONE" == 1 ]; then
 					logger -s -t $log_label -p local4.info "Received request=$dest, will call xCAT back in $delay seconds. Discovery is complete, run nodeset on this node to provision an Operating System"
 				else
 					logger -s -t $log_label -p local4.info "Received request to retry in a bit, will call xCAT back in $delay seconds"
@@ -409,8 +409,8 @@ while :; do
 			delay=$((delay-1))
 			sleep 1
 		done
-                destiny=''
-                dest=''
+		destiny=''
+		dest=''
 		echo "Retrying                                                                                   ";
 	elif [ "$dest" = shutdown ]; then
 		logger -s -t $log_label -p local4.info "Poweroff..."
--- a/xCAT-genesis-scripts/bin/doxcat
+++ b/xCAT-genesis-scripts/bin/doxcat
@@ -169,6 +169,10 @@ export XCATMASTER
 
 logger -s -t $log_label -p local4.info "XCATMASTER is $XCATMASTER, XCATPORT is $XCATPORT"
 
+head -n -1 /etc/rsyslog.conf > /etc/rsyslog.conf.new
+cp /etc/rsyslog.conf.new /etc/rsyslog.conf
+echo "*.* @$XCATMASTER" >> /etc/rsyslog.conf
+
 if [[ -n $hostip && -n $netmask && -n $gateway && -n $bootnic ]]; then
 	# doing static ip
 	# the device was determined above from the bootif mac, and put in bootnic
@@ -253,21 +257,6 @@ else
         fi
 fi
 
-count_f=0
-while [ $count_f -le 8 ]; do
-    rsyslogd_ps=`ps -ef | grep 'rsyslogd' | grep -v 'grep'`
-    if [ "$rsyslogd_ps" ]; then
-        break
-    else
-        ((count_f++))
-        sleep 0.5
-    fi
-done
-
-if [ $count_f -gt 8 ]; then
-    echo "rsyslogd maybe off"
-fi
-
 openssl genrsa -out /etc/xcat/certkey.pem 4096 > /dev/null 2>&1 &
 
 logger -s -t $log_label -p local4.info "Acquired IPv4 address on $bootnic"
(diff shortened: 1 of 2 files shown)


xCAT-genesis-scripts/bin/doxcat
```diff
@@ -282,9 +271,19 @@ fi
 
 # rv 0 state does not work with the new ntp versions
 while [ "`ntpq -c 'rv 0 offset' | awk -F '=' '/offset=/ { print $2 }' | awk -F '.' '{ print $1 }' | sed s/-//`" -ge 1000 ]; do 
+    echo `ntpq -c 'rv 0 offset`
     sleep 1
 done
 
+kill `cat /var/run/syslogd.pid`
+sleep 3
+RSYSLOGD_VERSION=`rsyslogd -v | grep "rsyslogd" | cut -d" " -f2 | cut -d"." -f1`
```
neo954: It is to complex to use `grep` and two `cut`s to do the work. Since we already use `awk` in this script. Just use `awk` to do the work.

    rsyslogd -v | awk '/rsyslogd/ { split($2, a, "."); print a[1]; }'

xCAT-genesis-scripts/bin/doxcat
```diff
@@ -282,9 +271,19 @@ fi
 
 # rv 0 state does not work with the new ntp versions
 while [ "`ntpq -c 'rv 0 offset' | awk -F '=' '/offset=/ { print $2 }' | awk -F '.' '{ print $1 }' | sed s/-//`" -ge 1000 ]; do 
+    echo `ntpq -c 'rv 0 offset`
     sleep 1
 done
 
+kill `cat /var/run/syslogd.pid`
+sleep 3
```
neo954: Just sleep three seconds here is a shoddy code style. Please rewrite it as the code you removed before. Check if the process is still running. Make sure the process is killed. And then move on.

neo954: Instead of use `ps` and `grep` to do the checking. It is  a better method to use `kill -0` to check if a process is still running.

    read -r RSYSLOG_PID </var/run/syslogd.pid 2>/dev/null
    kill "$RSYSLOG_PID" 2>/dev/null
    while kill -0 "$RSYSLOG_PID" 2>/dev/null
    do
        sleep 0.5
    done
    unset RSYSLOG_PID

xCAT-genesis-scripts/bin/doxcat
```diff
@@ -282,9 +271,19 @@ fi
...
+RSYSLOGD_VERSION=`rsyslogd -v | grep "rsyslogd" | cut -d" " -f2 | cut -d"." -f1`
+if [ $RSYSLOGD_VERSION -ge 8 ]; then
```
neo954: Add double quotes with the shell variable. In this way, if `$RSYSLOGD_VERSION` is an empty string. You will not get a syntax error.

Use `"$RSYSLOGD_VERSION"` instead of `$RSYSLOGD_VERSION`.

xCAT-genesis-scripts/bin/doxcat
```diff
@@ -331,6 +330,7 @@ while :; do
 		/bin/getcert $XCATMASTER:$XCATPORT
 		destiny=''
 		dest=''
+		touch /tmp/xcatgenesisdiscovery
```
neo954: I think just log an extra log here with `logger` is more simple and straight forward. In this way, we can avoid all the `if .. else ... fi` in line 397 to 401.

xCAT-genesis-scripts/bin/doxcat
```diff
@@ -282,9 +271,19 @@ fi
 
 # rv 0 state does not work with the new ntp versions
 while [ "`ntpq -c 'rv 0 offset' | awk -F '=' '/offset=/ { print $2 }' | awk -F '.' '{ print $1 }' | sed s/-//`" -ge 1000 ]; do 
+    echo `ntpq -c 'rv 0 offset`
```
neo954: BTW, you lost a single quote in this line. And, ``echo `ntpq -c 'rv 0 offset'` `` is all the same with `ntpq -c 'rv 0 offset'`, but waste a `pipe()` system call.

xuweibj:
Hi @neo954 , I have modified. please help to review again. Thanks!

xCAT-genesis-scripts/bin/doxcat
```diff
@@ -391,16 +397,20 @@ while :; do
                 destiny=''
                 dest='' 
 	elif [ "$dest" = standby ]; then
-		destiny=''
-		dest=''
 		delay=$((30+$RANDOM%270))
 		while [ $delay -gt 0 ]; do
 			if [ $((delay%10)) == 0 ]; then 
-				logger -s -t $log_label -p local4.info "Received request to retry in a bit, will call xCAT back in $delay seconds"
+				if [ $DISCOVERYDONE == 1 ]; then
```
neo954: Please use double quotes along with the variable. In this way, you will not get syntax if `$DISCOVERYDONE` is empty.

    [ "$DISCOVERYDONE" == 1 ]

xCAT-genesis-scripts/bin/doxcat
```diff
@@ -331,6 +335,8 @@ while :; do
 		/bin/getcert $XCATMASTER:$XCATPORT
 		destiny=''
 		dest=''
+		DISCOVERYDONE=1
+		export DISCOVERYDONE
```
neo954: I suggest named this environment variable with a `XCAT_` prefix. I suggest use something like `XCAT_DISCOVERY_WAS_DONE`.

In this way, we can reduce the chance of pollute the namespace of all the environment variables.

xCAT-genesis-scripts/bin/doxcat
```diff
@@ -391,16 +397,20 @@ while :; do
...
                 dest='' 
 	elif [ "$dest" = standby ]; then
-		destiny=''
-		dest=''
 		delay=$((30+$RANDOM%270))
 		while [ $delay -gt 0 ]; do
 			if [ $((delay%10)) == 0 ]; then 
-				logger -s -t $log_label -p local4.info "Received request to retry in a bit, will call xCAT back in $delay seconds"
+				if [ $DISCOVERYDONE == 1 ]; then
+					logger -s -t $log_label -p local4.info "Received request=$dest, will call xCAT back in $delay seconds. Discovery is complete, run nodeset on this node to provision an Operating System"
```
neo954: Please quote the shell variable. Otherwise, when `$log_label` is empty, you will short a command line argument.

xCAT-genesis-scripts/bin/doxcat
```diff
@@ -391,16 +397,20 @@ while :; do
...
+				else
+					logger -s -t $log_label -p local4.info "Received request to retry in a bit, will call xCAT back in $delay seconds"
```
neo954: Please quote the shell variable.

xCAT-genesis-scripts/bin/doxcat
```diff
@@ -391,16 +397,20 @@ while :; do
...
+				fi
 			fi
 			delay=$((delay-1))
 			sleep 1
 		done
+                destiny=''
+                dest=''
```
neo954: It seems all the other lines in this script are using `<TAB>` to indent. Please keep the code style identical.

Line 412 and 412 are using `<SPACE>` to indent.

neo954:
I will merge this one.

daniceexi:
Good work.

PR #2067, opened 2016-11-02
update dhcpd conf file when upgrade xcat

### LINKED ISSUE #1994: Invalid option in dhcp.conf, breaking dhcpd service on RHEL6
In commit ee1a1574c361ce85b30c04b118c12fe611607bf9 from Sept 2014, the following line was added to dhcp.pm:

```
push @netent,
          "    } else if option client-architecture = 00:0e { #OPAL-v3\n ";
        push @netent, "      option conf-file = \"http://$tftp/tftpboot/pxelinux.cfg/p/" . $net . "_" . $maskbits . "\";\n";
        push @netent,
```

The trouble is that "option conf-file" is invalid on RHEL6.7. The resultant dhcpd.conf file results in a dhcpd service that will not start.

```
dhcpd: /etc/dhcp/dhcpd.conf line 848: no option named conf-file in space dhcp
dhcpd:        option conf-file =
dhcpd:                ^
dhcpd: /etc/dhcp/dhcpd.conf line 849: expecting a parameter or declaration
dhcpd:     } else
dhcpd:      ^
dhcpd: Configuration file errors encountered -- exiting
```

It would appear that the service is parsing the entire file, despite the effort to wrap the code in an if check for client-architecture.
The offending line can be commented out manually and the file will otherwise function properly. Subsequent calls to "makedhcp" will not change the section back since it is already present for the given interface.

I'm surprised that no one else has had this problem given that this code is two years old.

penguhyang:
fix #1994

### DIFF
--- a/xCAT-server/sbin/xcatconfig
+++ b/xCAT-server/sbin/xcatconfig
@@ -1464,6 +1464,13 @@ sub initDB
                 xCAT::MsgUtils->message('E', "Could not add default value for site.xcatsslversion.");
             }
         }
+
+        # Set dhcpd.conf be updated when update xcat
+        my $cmds = "XCATBYPASS=Y $::XCATROOT/sbin/makedhcp -n 2>/dev/null";
+        xCAT::Utils->runcmd("$cmds", -1);
+        if ($::RUNCMD_RC != 0) {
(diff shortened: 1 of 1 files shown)


immarvin:
agree to merge

PR #2069, opened 2016-11-02
New template of probe command dispatch in hierarchy environment

hu-weihua:
What are this pull request do:
1. Extract command dispatch and reply collection part into a independent file, work as a class
2. Offer a probe sub_command template for customer, customer can use directly. this template  has finished the basic work what every sub_command have to do.

This template can run independently and output like below

    # xcatprobe hierarchy_template -n test[1-15]
    <mn>: I reveive node range is test9,test1,test2,test3,test4,test5                                                 [ OK ]
    <mn>: first checking point                                                                                        [ OK ]
    <mn>: second checking point                                                                                       [FAIL]
    <c910f02c04p04>: I reveive node range is test10,test11,test12,test13,test14,test15,test6,test7,test8              [ OK ]
    <c910f02c04p04>: first checking point                                                                             [ OK ]           
    <c910f02c04p04>: second checking point                                                                            [FAIL]
    ======================do summary=====================
    [mn]
        [ok]     : I reveive node range is test9,test1,test2,test3,test4,test5
        [ok]     : first checking point
        [failed] : second checking point
    [c910f02c04p04]
        [ok]     : I reveive node range is test10,test11,test12,test13,test14,test15,test6,test7,test8
        [ok]     : first checking point
        [failed] : second checking point

### DIFF
--- a/xCAT-probe/subcmds/code_template
+++ b/xCAT-probe/subcmds/code_template
@@ -1,6 +1,26 @@
 #!/usr/bin/perl
 # IBM(c) 2016 EPL license http://www.eclipse.org/legal/epl-v10.html
 
+#--------------------------------------------------------
+#This is a template for developing a new probe sub_command. Especially in hierarchical environment.
+#This template mainly implement the sub_comamd dispatch in hierarchical structure and basic framework of a new sub_comamd.
+#Developer only need to focus on main probe job (by implement do_main_job function) and friendly output (by implement summary_all_jobs_output function) for user.
+#This template can also be used in flat structure. but if developer think it's too heavy in flat, it's fine to develop sub command directly.
+#But in hierarchical structure, we strongly recommand using this template.
+#
+#The main dispatch policy are:
+#1. if there isn't noderange input from commmand line and  there are service nodes defined in current MN,
+#   dispatch exact same command input from STDIN to all SNs and current MN. if there isn't service nodes defined,
+#   just hanld command input from STDIN in current MN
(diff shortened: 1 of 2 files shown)


xCAT-probe/subcmds/code_template
```diff
@@ -78,23 +55,19 @@ Options:
...
 # Recommand to use probe_utils->send_msg() to handle message you plan to print out
 # A simple example has been written in funciton.
 #------------------------------------
 sub do_main_job {
-    my $outputtarget = shift;
-    $outputtarget = "stdout" if (!$outputtarget);
     my $rst = 0;
 
-    probe_utils->send_msg($outputtarget, "o", "I reveive node range is $noderange");
+    probe_utils->send_msg("$output", "o", "I reveive node range is $noderange");
```
whowutwut: I don't think we should use "I" in the message.  
Revive also should be "Received"? 

Suggest:
"Received node range: "

whowutwut:
What is the first checking point and second checking point mean?

hu-weihua:
@whowutwut ， thanks for your comments.

This is a template, also is a basic framework of how ``xcat probe`` to support hierarchy. 
For example, if a developer want to develop a new probe tool which will be used in hierarchy environment. he/she doesn't need to put effort to code for dispatching command to SNs and collect replay from SNs, this framework has done this part, he/she just needs to finish the main probe job of the new tool, just like this new tool just only work on  MN.

So ``first checking point and second checking point`` means the probe jobs which the new probe tool will do. Maybe this expression isn't easy to understand. How about using ``probe job 1 and probe job 2``. what do you think?

xuweibj:
I agree to merge.

PR #2000, opened 2016-10-20
enhance xcatprobe discovery sub command

xuweibj:
This version only support monitor discovery process.

Add some bmc, mtms, and switch check items in pre_defind_node.

Output as follows (with ``-V``):

```
# xcatprobe -w discovery -n frame23cn18 -m mtms -V
All pre_defined nodes are valid                                                                                   [ OK ]
Attribute 'master' in 'site' table is configured well                                                             [ OK ]
DHCP dynamic range is configured well                                                                             [ OK ]
Genesis files are avaliable                                                                                       [ OK ]

-------------------------------------------------------------
                               ___
        ____    _  _____   _.-|   |          |\__/,|   (`\
 __  __/ ___|  / \|_   _| {   |   |          |x x  |__ _) )
 \ \/ / |     / _ \ | |    "-.|___|        _.( T   )  `  /
  >  <| |___ / ___ \| |    .--'-`-.     _((_ `^--' /_<  \
 /_/\_\\____/_/   \_\_|  .+|______|__.-||__)`-'(((/  (((/
-------------------------------------------------------------

Start capturing every message during discovery process......

[70:e2:84:14:0f:0a] Receive DHCPDISCOVER from 70:e2:84:14:0f:0a via enP4p1s0f2
[70:e2:84:14:0f:09] Receive DHCPDISCOVER from 70:e2:84:14:0f:09 via enP4p1s0f3
[70:e2:84:14:0f:0a] Send DHCPOFFER on 192.168.5.176 back to 70:e2:84:14:0f:0a via enP4p1s0f2
[70:e2:84:14:0f:09] Send DHCPOFFER on 50.0.0.8 back to 70:e2:84:14:0f:09 via enP4p1s0f3
[70:e2:84:14:0f:0a] Receive DHCPREQUEST from 70:e2:84:14:0f:0a for 192.168.5.176 (192.168.3.25) via enP4p1s0f2
[70:e2:84:14:0f:0a] Send DHCPACK on 192.168.5.176 back to 70:e2:84:14:0f:0a via enP4p1s0f2
[70:e2:84:14:0f:09] Receive DHCPREQUEST from 70:e2:84:14:0f:09 for 50.0.0.8 (50.3.25.1) via enP4p1s0f3
[70:e2:84:14:0f:09] Send DHCPACK on 50.0.0.8 back to 70:e2:84:14:0f:09 via enP4p1s0f3
[...]

PR #2070, opened 2016-11-02
Add replay function for discovery

xuweibj:
add replay function ----option "-r" can be use.
If can not get node ip, add error msg.
modified output msg based on #2000 comments.


hu-weihua:
I have reviewed the code change and agree to merge, thanks @xuweibj .

PR #2076, opened 2016-11-03
quote date command in string

### LINKED ISSUE #2075: Quote or delete `date` string in xcatinstallpost
In change d2cee73 on Jul 7 2015, the file `xcat-core/xCAT/postscripts/xcatinstallpost` was updated as follows on line 46:
```
msgutil_r "$MASTER_IP" "err" `date`" xcatinstallpost: Network not configured, please check..." "/var/log/xcat/xcat.log"
```
The output from `date` contains spaces, and when it's not quoted, the rest of the message above does not get logged. You end up with "Wed" in the log file and nothing else (if you run the code on Wednesday). 

At a minimum, the output from `date` should be quoted but honestly I don't think you even need it. The logger already logs the date and time in the log file.  Removing the date output string from the message would suffice.

penguhyang:
fix #2075

### DIFF
--- a/xCAT/postscripts/xcatinstallpost
+++ b/xCAT/postscripts/xcatinstallpost
@@ -43,7 +43,7 @@ while true; do
 
     if [ $RETRY -eq 90 ];then
        #timeout, complain and exit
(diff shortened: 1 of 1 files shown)


immarvin:
hi @penguhyang ,is it possible to add some delimiter such as ":" between the `date` and the message following?

immarvin:
agree to merge

PR #2079, opened 2016-11-03
Use sleep subroutine instead of exec(sleep)

chenglch:
exec function call will take over all the process, lead to no chance to
continue the work within the script, this patch use sleep function call from
perl library instead.

Close-issue: #1801

### DIFF
--- a/xCAT-server/share/xcat/cons/ipmi
+++ b/xCAT-server/share/xcat/cons/ipmi
@@ -1,50 +1,52 @@
 #!/usr/bin/env perl
 # IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
 use Fcntl qw(:DEFAULT :flock);
-
-sub get_lock {
-    unless (flock(LOCKHANDLE, LOCK_EX | LOCK_NB)) {
-        $| = 1;
-        print "Acquiring startup lock...";
-        flock(LOCKHANDLE, LOCK_EX) or die "Fatal Error securing startup lock";
-        print "done\n";
-    }
-    truncate(LOCKHANDLE, 0);
-    print LOCKHANDLE $$ . "\n";
-}
-
-sub release_lock {
-    truncate(LOCKHANDLE, 0);
-    flock(LOCKHANDLE, LOCK_UN);
-}
-
-BEGIN
-{
-    use Time::HiRes qw(sleep);
-    use File::Path;
-    use Fcntl qw(:DEFAULT :flock);
-    $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : '/opt/xcat';
-    umask 0077;
-    mkpath("/tmp/xcat/");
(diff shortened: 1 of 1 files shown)


xCAT-server/share/xcat/cons/ipmi
```diff
@@ -1,50 +1,52 @@
...
 use Fcntl qw(:DEFAULT :flock);
+use Time::HiRes qw(sleep);
+use File::Path;
+BEGIN {
+    $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : -d '/opt/xcat' ? '/opt/xcat' : '/usr';
+}
+my $sleepint = int(rand(10)); #Stagger start to avoid overwhelming conserver/xCATd
+my ($lockfd, $bmc);
+my $username = 'USERID';
+my $password = 'PASSW0RD';
```
zet809: We shall use the default account 'ADMIN/admin'

xCAT-server/share/xcat/cons/ipmi
```diff
@@ -1,50 +1,52 @@
...
+        exit 1;
+    }
+    unless (flock($lockfd, LOCK_EX)) {
+        print "Unable to lock file ".CONSOLE_LOCK_FILE."\n";
+        close($lockfd);
+        sleep(15);
+        exit 1;
+    }
+    print "done\n";
+    unless (syswrite($lockfd, $$, length($$))) {
```
zet809: Will every rcons write something in the lock file, and the lock file will be larger and larger?

xCAT-server/share/xcat/cons/ipmi
```diff
@@ -63,16 +65,22 @@ my $cmdref = {
     arg       => ["text"],
     noderange => [ $ARGV[0] ]
 };
+acquire_lock();
+# avoid of congestion
+sleep(0.1);
+release_lock();
```
zet809: What is the purpose of get the lock and then release it?
Nothing shall do before releasing the lock?
Same question for line 79-81.

xCAT-server/share/xcat/cons/ipmi
chenglch: This file is opened in create and write mode, not append mode, so the lock file will not be larger and larger.

xCAT-server/share/xcat/cons/ipmi
chenglch: By default xcatd can fork 64 child processes to handle the request which means it has the concurrent processing capacity, I just use sleep here to control the request frequency sent to to the xcatd. If put the `submit_request` in the lock scope, xcatd can only handle the client request one by one.

daniceexi:
@chenglch I'd like to have the root cause of #1801, and why your change could fix it.

chenglch:
@daniceexi , Original implementation use the exec("sleep") which will take place of the process launched by the conserver. When the sleep process ends, conserver will launched a new ipmi script process.   I use non-block lock here instead.

zet809:
Looks OK to me, I agree to merge it. Thx!

PR #2086, opened 2016-11-04
update to add more debug information before rmvm is executed

tingtli:
update to add more debug information before rmvm is executed


immarvin:
since this PR is for debugging, merge it first so that the debug information can be obtained by next monday

PR #2088, opened 2016-11-04
Add extra nics to link-up detection search list

russauld:
When a bonded NIC configuration is to be used as the primary interface
for a node, often that interface cannot be used to PXE boot the node.
In this case, an different NIC is used as the "installnic".
It's possible that after the node reboots, that the "installnic"
is not configured with an address. The IP address can be moved to the
bonded connection after provisioning.
The additional interfaces used by the node are listed in "NICIPS"
from the "nics" table. This code change parses "NICIPS" for adapter
names and adds them to the INSTALLNIC variable so that all defined
interfaces for this node are checked in the link-up check.
This should still fix "defect 4616"
https://sourceforge.net/p/xcat/bugs/4616/
The original code change was intended to delay the execution of the
xcatinstallpost script until a real network interface came up (i.e. not usb0)

### DIFF
--- a/xCAT/postscripts/xcatinstallpost
+++ b/xCAT/postscripts/xcatinstallpost
@@ -20,10 +20,21 @@ fi
 SLI=$(awk 'BEGIN{srand(); printf("%d\n",rand()*10)}')
 sleep $SLI
 
-
+# Get the name of the installnic
 MACADDR=`grep MACADDRESS= /xcatpost/mypostscript.post |awk -F = '{print $2}'|sed s/\'//g`
 INSTALLNIC=`ip -o link|grep -i $MACADDR|awk  '{print $2}'|sed s/://`
 
+# Also get the any interface names set in NICIPS (nics table)
+NICIPS=$(grep NICIPS= /xcatpost/mypostscript.post |awk -F = '{print $2}'|sed s/\'//g)
+NICS=""
+OIFS=$IFS
+IFS=$','
+for n in $NICIPS; do
+  NAME=$(echo $n | awk -F \! '{print $1}')
+  NICS="$NICS $NAME"
+done
+IFS=$OIFS
+INSTALLNIC="$INSTALLNIC $NICS"
 
 #the nics have not been configured when running the PBS sometimes, need to make sure... 
 RETRY=0


daniceexi:
@immarvin could you work with @bybai for this pull request?

bybai:
hi @russauld ,
1, I have a question in your pull request:
although NICIPS nics are added into INSTALLNIC, if  one of the INSTALLNIC is link-up, the left others will not be checked in the code:
```
RETRY=0
 NETUP=0
 while true; do
     for nic in $INSTALLNIC ;do
         ip -4 --oneline addr show dev $nic |grep inet >/dev/null && NETUP=1 && break
     done

     [ $NETUP -ne 0 ] && break;

     RETRY=$[ $RETRY + 1 ]

     if [ $RETRY -eq 90 ];then
        msgutil_r "$MASTER_IP" "err" `date`" xcatinstallpost: Network not configured, please check..." "/var/log/xcat/xcat.log"
        exit 1
     fi

     sleep 2
 done
```

So how are all defined
interfaces for this node  checked in the link-up check?

2, Do you want to make sure all the nics for bond and bond interface itsself are link-up or not?

Thanks

russauld:
You are correct, not all interfaces are checked for link-up. If any of the interfaces in INSTALLNIC are up, the code will exit the loop and proceed. There is no reason to ensure that all of the interfaces are up, and in fact, they may not be. 
If you examine the code change that was implemented to fix defect 4616, you can see that the code used to check that any interface was up and proceed. That was too broad and resulted in a problem on RHEL7.
The fix ensured that the installation interface was up, and that is too narrow.
This change makes a broader list of interfaces comprised of the installation interface and any additional interfaces explicitly listed for the node in the nics table.
I feel that this is approach establishes a middle ground between the old approach and the new approach.

Perhaps a better approach would be to test for a route to the MASTER or SERVICENODE. Ensuring that at least one interface is up may not guarantee that a route to the MASTER node is available.
But forcing the installation interface to be available to complete the postbootscripts is not reasonable.

bybai:
Hi @russauld ,
I agreed with you that this is approach establishes a middle ground between the old approach and the new approach and one interface is up may not guarantee that a route to the MASTER node is available.
Hi @immarvin ,
What is your opinion?

immarvin:
This is the background for this PR

From: "Russell Auld" <russauld@comcast.net>
To: "'xCAT Users Mailing list'" <xcat-user@lists.sourceforge.net>
Cc:
Subject: [xcat-user] INSTALLNIC
Date: Thu, Nov 3, 2016 10:32 AM
 
I have a situation where postbootscripts do not run. I traced the problem back to “/opt/xcat/xcatinstallpost” script which is called by the “/etc/init.d/xcatpostinit1” service after the first boot.

The trouble is with this section of code:

````
MACADDR=`grep MACADDRESS= /xcatpost/mypostscript.post |awk -F = '{print $2}'|sed s/\'//g`

INSTALLNIC=`ip -o link|grep -i $MACADDR|awk  '{print $2}'|sed s/://`

while true; do

    #scan the nics with the specified mac address

    #there will be multiple nic names for a mac address when the network bridge exists

    for nic in $INSTALLNIC ;do

        #check whether the nic is configured and linkup

        ip -4 --oneline addr show dev $nic |grep inet >/dev/null && NETUP=1 && break

    done

   

    #nic is configured,terminate scan...

    [ $NETUP -ne 0 ] && break;   

    RETRY=$[ $RETRY + 1 ]

    if [ $RETRY -eq 90 ];then

       #timeout, complain and exit

       msgutil_r "$MASTER_IP" "err" `date`" xcatinstallpost: Network not configured, please check..." "/var/log/xcat/xcat.log"

       exit 1

    fi

   

    #sleep sometime before the next scan 

    sleep 2

done
````
 

The issue that I have is that the NIC that I use to install the machines is not the primary NIC that is used when the node reboots.

That’s because the NIC used during the PXE/Anaconda stage cannot be part of a bonded connection, so we have to run a separate single link for imaging. When the node runs normally, we want to use the bonded connections. In other words, the node would have three Ethernet connections - two 10Gb connections (which would be bonded) and one 1Gb standard Ethernet connection.

Normally I set the “installnic” to “mac” in the noderes table and I set “mac” to the mac of the nic that is used to PXE the device.

The code above insists that the link associated with the MAC used to PXE the node be up, and will fail if not. That’s not reasonable after the node reboots.

This method worked well with xCAT v2.8.5. Now I’m using v2.12.1 and this method doesn’t work.

Is there a better way to handle this issue with bonded Ethernet and PXE booting?

This is what the man page for ‘noderes’ says about ‘installnic’:

The network adapter on the node that will be used for OS deployment, the installnic can be set to the network adapter name or the mac address or the keyword “mac” which means that the network interface specified by the mac address in the mac table will be used. If not set, primarynic will be used. If primarynic is not set too, the keyword “mac” will be used as default.

immarvin:
hi @russauld , I have 2 questions :

1. why the installnic is not "up" on the 1st reboot after provision? since its dhcp lease generated on "nodeset" is still available on the management node
2. is it possible that the bonded network has no route to the management node?

russauld:
The installnic is not up due to a postscript that we run on the node. The script disables the installnic interface and configures the bond0 interface. Thus when the node reboots after Anaconda, the bond0 interface comes up and the installnic interface does not.
We don't use a pool of DHCP IP addresses - we know ahead of time what the IP address will be for each node. Furthermore, we use site.managedaddressmode=static so that the IP address, netmask, gateway, etc. are specified on the kernel command line.
Since we cannot PXE boot from a bond interface, we must PXE from another interface. That means that the MAC for the node in xCAT is the installnic and not the bond connection.
Since we don't want to reserve another IP address for the installnic interface and we don't use a pool of DHCP addresses for node discovery, we have a postscript that manipulates the ifcfg-eth2 and ifcfg-bond0 scripts so that the proper interface comes up when the machine boots.

immarvin:
hi @russauld , thanks for the info. The intent of the code logic in the fix for https://sourceforge.net/p/xcat/bugs/4616/ was to make sure and wait for the network access between provisioning node and management node( or service node) been well configured and activated. We do not care which NIC is up, in fact, a NIC up can not make sure this, considering the scenario that there is no route between the NIC linked up and the MN/SN.  

So your fix might fix the problem you encountered, but it does not solve the problem. I created a PR on this https://github.com/xcat2/xcat-core/pull/2190

russauld:
I agree, the best thing is to ensure that there is a route to the master node.
I think your code change will suffice.

immarvin:
hi @russauld , since you are ok with the PR merged, I will close this PR. If you have other question on this, you can reopen this or open another ticket.

thanks

Issue #1770, opened 2016-08-26
[DEV] After update MN, SN SSL connect failed


xuweibj:
SN: c910f03c17k17

```
[root@c910f03c05k04]# nodeset c910f03c17k20 osimage=sles12-ppc64le-install-compute
c910f03c17k20: install sles12-ppc64le-compute
Error: Unable to dispatch hierarchical sub-command to c910f03c17k17:3001.  Error: Connection failure: SSL connect attempt failed with unknown error error:0407006A:rsa routines:RSA_padding_check_PKCS1_type_1:block type is not 01 error:04067072:rsa routines:RSA_EAY_PUBLIC_DECRYPT:padding check failed error:0D0C5006:asn1 encoding routines:ASN1_item_verify:EVP lib error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed at /opt/xcat/lib/perl/xCAT/Client.pm line 278. 
```

immarvin:
can be workaround by "update node SN -P servicenode" after MN updating. Maybe we should add this into xCAT Doc

whowutwut:
Yes, I think we should document the process that should be followed in this scenario.  Is there any more information about how to re-create this?

xuweibj:
Hi @whowutwut , my steps as follows:
1. on MN, used `nodeset` command install SN package on my SN.
2. run `yum update '*xCAT*'` on MN, install a newer version 
3. run any xCAT command will see this error message

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2094
hi @whowutwut , I added a message on xCAT upgrade to make user aware of this

immarvin:
hi @tingtli , could FVT help to recreate this issue ? thanks

tingtli:
@junxiawang  will help verify this bug.

immarvin:
cannot be recreated
unless you run ``xcatconfig -c``

immarvin:
close this issue, please reopen it when you find it

PR #2094, opened 2016-11-07
fix issue [DEV] After update MN, SN SSL connect failed #1770; Invalid option in dhcp.conf, breaking dhcpd service on RHEL6 #1994

immarvin:
fix https://github.com/xcat2/xcat-core/issues/1994:

add the tips on upgrading xCAT on xcat management node in hierarchy cluster in xCAT doc and ``xcatconfig -u`` message

fix https://github.com/xcat2/xcat-core/issues/1770:
run "makedhcp -n" on xcat upgrade

### DIFF
--- a/docs/source/guides/install-guides/apt/update_xcat.rst
+++ b/docs/source/guides/install-guides/apt/update_xcat.rst
@@ -5,4 +5,4 @@ If at a later date you want to update xCAT, first, update the software repositor
     apt-get update
     apt-get --only-upgrade install xcat*
 
-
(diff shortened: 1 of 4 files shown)


daniceexi:
Regarding the fix for #1770, I think we did not figure out the root cause that why the 'makedhcp' changed part of the dhcpd.conf (make it inconsistent). 
@immarvin Please investigate more. My suggestion is we should skip our command to generate inconsistent dhcpd.conf.

Issue #2081, opened 2016-11-03
[fvt]2.12.4 xcatprobe discovery could not probe the errors if ip=NULL or invalid


junxiawang:
env:fs4
build:~]# lsxcatd -v
Version 2.12.4 (git commit ed4e645558076d38c9bb65db5c2155025d721c10, built Thu Oct 27 06:15:34 EDT 2016)
How to reproduce:
```
[root@fs4 ~]# chdef frame23cn17 ip=
1 object definitions have been created or modified.
[root@fs4 ~]# xcatprobe discovery -n frame23cn17 -m mtms
All pre_defined nodes are valid                                                                                   [ OK ]
Attribute 'master' in 'site' table is configured                                                                  [ OK ]
DHCP dynamic range is configured                                                                                  [ OK ]
Genesis files are avaliable         
---------------------->
the predefine node'iip =Null or invalid 
but using xcatprobe discovery could not probe the errors.
```

xuweibj:
#2015 fix a bug that if no ip definition, but the node can be resolved. Set the ip get from resolved as the node's ip.

PR #2105, opened 2016-11-08
if no ip definition get ip from getipaddr function

xuweibj:
#2081


hu-weihua:

PR #2107, opened 2016-11-08
add cumulus switch definition template

immarvin:
add cumulus switch definition template

### DIFF
--- a/xCAT/templates/objects/node/cumulusswitch.stanza
+++ b/xCAT/templates/objects/node/cumulusswitch.stanza
@@ -1,0 +1,14 @@
+# <the template for PowerLE NV node definition>
+
+cumulusswitch:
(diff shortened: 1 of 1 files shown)


xCAT/templates/objects/node/cumulusswitch.stanza
```diff
@@ -0,0 +1,14 @@
+# <the template for PowerLE NV node definition>
+
+cumulusswitch:
+    objtype=node
+    arch="OPTIONAL: the arch of the switch management service processor, such as armv71"
+    groups=switch
+    ip="MANDATORY:the ip address of the managaement interface"
```
neo954: Wrong spelling of the word `management'.

neo954:
It seems good. I will merge this one.

PR #2111, opened 2016-11-08
ddns.pm: fix condition to append domainname to hostname

kcgthb:
Fix an issue where the domainname was not appended to the hostname when the domainname was contained in a substring of the hostname.
Update the regular expression to make sure it only check for the domainname at the end of the hostname string.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ddns.pm
+++ b/xCAT-server/lib/xcat/plugins/ddns.pm
@@ -493,7 +493,7 @@ sub process_request {
             # for only the sake of comparison, ensure consistant dot suffix
             unless ($canonical =~ /\.\z/) { $canonical .= '.' }
             foreach my $alias (@aliases) {
-                unless ($alias =~ /$domain/) {
+                unless ($alias =~ /\.$domain\z/) {
                     $alias .= "." . $domain;
                 }
                 unless ($alias =~ /\.\z/) {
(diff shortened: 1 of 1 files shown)


bybai:
hi @kcgthb , could you give an issue example which you fixed ?

kcgthb:
Hi @bybai 

Yes. For instance, if you have the following network:
```
Object name: ib
    domain=ib
    mask=255.255.255.0
    mgtifname=ib0
    nameservers=10.0.2.1
    net=10.0.2.0
```
and the following host definition:
```
Object name: host1
    nichostnamesuffixes.ib0=-ib0
    nicips.ib0=10.0.2.41
    nicnetworks.ib0=ib
    nictypes.ib0=infiniband
```

In that case,the nic hostname suffix in the `ib` network is `-ib0`, so the hostname will be `host1-ib0` in that net. 

Since the domain name (`ib`) is a substring of the hostname (`host1-ib0`) the regexp [there](https://github.com/xcat2/xcat-core/blob/0ad16c3a16128a09f0429bf956ad9e55f1959354/xCAT-server/lib/xcat/plugins/ddns.pm#L496) and [there](https://github.com/xcat2/xcat-core/blob/0ad16c3a16128a09f0429bf956ad9e55f1959354/xCAT-server/lib/xcat/plugins/ddns.pm#L1420) will match, and the script will consider that the hostname already contains the domain name. 

As a consequence, it will not append the domainname to the hostname, and `makedns` will fail trying to update a record which is not in the correct zone.

The fix is to restrict the regexp to check for the domain name string in the hostname only at the end of the string, and after a dot.

I hope this clarifies things.

bybai:
Hi @immarvin ,
Could you have more comments?    
I am OK with the change.
Thanks

immarvin:
I am ok with the change, thanks for your contribution @kcgthb

Issue #1939, opened 2016-10-05
linux orderly shutdown not noticed by xcat.


ralphbellofatto:
We use xdsh xxx reboot to do an orderly shutdown of our nodes.

when we do this,  xCAT fails to notice the change in status on the node.

To observe this we use a loop to watch the status of the node as is rebooted.
on one terminal we issue:

```
while true; do 
  date | tr -d '\n'; echo -n " "; 
  lsdef c460c002 -i status -c; 
  sleep 5;
done                                                                              
```

On another terminal we issue 
date ; xdsh c460c002 reboot;

Where we see:

```
Wed Oct  5 08:05:58 EDT 2016
c460c002: Connection to c460c002 closed by remote host.
```

Watching the boot sequence on the first terminal we see:

Wed Oct  5 08:05:29 EDT 2016 c460c002: status=booted  
...
Wed Oct  5 08:06:49 EDT 2016 c460c002: status=booted  
...
Wed Oct  5 08:08:58 EDT 2016 c460c002: status=netbooting  
...
Wed Oct  5 08:12:45 EDT 2016 c460c002: status=failed
...

The shutdown appears to be completly unnoticed by xCAT until the node reaches out and requests it's boot image.

This causes problems for us when we attempt to do a reboot and then the shutdown hangs, there is nothing in the xcat status that tell that we are in this state.

tingtli:
I remember we have a powering-off state when we do some test but not sure why there is no in ralph's environment. Is it because the time  framework? @immarvin

ralphbellofatto:
hmm...

observations:
```
[root@c460mgt01 otherpkgs]# rpower c460c002 off
c460c002: off
[root@c460mgt01 otherpkgs]# nodestat c460c002
c460c002: noping

[root@c460mgt01 otherpkgs]# lsdef c460c002 -i status
Object name: c460c002
    status=powering-off

```

so, the rpower off does get noticed, but not by nodestat..

however
```
rpower c460c002 on
# wait for boot to complete...

[root@c460mgt01 otherpkgs]# nodestat c460c002
c460c002: sshd
[root@c460mgt01 otherpkgs]# lsdef c460c002 -i status
Object name: c460c002
    status=booted

[root@c460mgt01 otherpkgs]# xdsh c460c002 reboot

[root@c460mgt01 otherpkgs]# nodestat c460c002
c460c002: noping
[root@c460mgt01 otherpkgs]# lsdef c460c002 -i status
Object name: c460c002
[...]

PR #2113, opened 2016-11-09
use the relative path to enable xcatpostinit1.service;remove previousxcatpostinit1 sysvinit file and systemd service unit file

immarvin:
fix the following updates from Ralph on issue https://github.com/xcat2/xcat-core/issues/1939


daniceexi:

PR #2114, opened 2016-11-09
fix issue xcat object naming is too restrictive after sourceforge issue 4675 fix #2100

### LINKED ISSUE #2100: xcat object naming is too restrictive after sourceforge issue 4675 fix
sourceforge 4675 bug limits changing characters used in an object name to a-z, A-Z, 0-9. "-" and "_". However P775 bpa and fsp objects get created with a dynamic ip address and then changed to a static ip address.  The 4675 fix does not allow an ip address to be used for the object name because of the "." in the ip address.

Can DBobjectdefs.pm line following:
    # check the new object name specified with -n
be changed from:
    if ($::command eq "chdef" && $::opt_n && ($::opt_n !~ /^[a-zA-Z0-9-_]+$/)) {
to:
    if ($::command eq "chdef" && $::opt_n && ($::opt_n !~ /^[a-zA-Z0-9-_\.]+$/)) {

To allow for the "." character?

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2100

### DIFF
--- a/xCAT-server/lib/xcat/plugins/DBobjectdefs.pm
+++ b/xCAT-server/lib/xcat/plugins/DBobjectdefs.pm
@@ -21,6 +21,7 @@ use Getopt::Long;
 use xCAT::MsgUtils;
 use xCAT::Utils;
 use xCAT::SvrUtils;
+use xCAT::NetworkUtils;
 use File::Find;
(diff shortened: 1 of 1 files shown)


daniceexi:

Issue #2117, opened 2016-11-09
[fvt]2.12.4 xcatprobe discovery -r could not probe switch based discovery or mtms based


junxiawang:
env:fs4
build:[root@fs4 ~]# lsxcatd -v
Version 2.12.4 (git commit d57001530b6d58a3ee0e52ede1f794e065da135f, built Mon Nov  7 22:37:04 EST 2016)
[root@fs4 ~]# 
How to reproduce:
```
1.node1  group wjx is predefined mtms based discovery
2.node2 group wjx is predefined switch based discovery
3.all nodes discovered successfull,but node1 using mtms node2 using switch 
4.[root@fs4 ~]# xcatprobe discovery -n wjx -m switch -r 3h
All pre_defined nodes are valid                                                                                   [ OK ]
Attribute 'master' in 'site' table is configured                                                                  [ OK ]
DHCP dynamic range is configured                                                                                  [ OK ]
Genesis files are avaliable                                                                                       [ OK ]
Starting to scan logs which are later than 'Tue Nov  8 22:26:18 2016', please waiting for a while.............    
==================discovery_probe_report=================
All nodes matched successfully                                                                                    [ OK ]
[root@fs4 ~]# 
[root@fs4 ~]# xcatprobe discovery -n wjx -m switch -r 30m
All pre_defined nodes are valid                                                                                   [ OK ]
Attribute 'master' in 'site' table is configured                                                                  [ OK ]
DHCP dynamic range is configured                                                                                  [ OK ]
Genesis files are avaliable                                                                                       [ OK ]
Starting to scan logs which are later than 'Wed Nov  9 00:55:50 2016', please waiting for a while.............    
==================discovery_probe_report=================
Discovered 1 node(s) successfully, 1 node(s) failed.
Unmatched node(s):
frame23cn17
Unmatched MAC(s):
[0c:c4:7a:d2:cf:f8] : stop at stage 'download_bootloder'                                                          [FAIL]
[70:e2:84:14:0f:09] : stop at stage 'download_bootloder'                                                          [FAIL]

```

PR #2120, opened 2016-11-09
fix issue 2117, distinguish discovery type

xuweibj:
#2117 

If found node by using different type with ``-m`` specified, give a conclusion that discovery successfully but different type.


hu-weihua:
I have reviewed the change and agree to merge, thanks @xuweibj .

Issue #2110, opened 2016-11-08
xcatprobe xcatmn requires tftp tool , add dependency to xcatprobe rpm


whowutwut:
If the xcatprobe xcatmn option requires tftp in order to do the verification.

```
[root@fs4 autotest]# xcatprobe xcatmn -i enP4p1s0f2
<mn>: All xCAT deamons are running                                                                                [ OK ]
<mn>: xcatd can receive command request                                                                           [ OK ]
<mn>: 'site' table is configured correctly                                                                        [ OK ]
<mn>: Provision network is configured correctly                                                                   [ OK ]
<mn>: 'passwd' table is configured correctly                                                                      [ OK ]
<mn>: Important directory is configured correctly                                                                 [ OK ]
<mn>: SELinux is disabled on current server                                                                       [ OK ]
<mn>: HTTP service works well                                                                                     [ OK ]
<mn>: TFTP service works well                                                                                     [FAIL]
<mn>: TFTP check need 'tftp' tool, please install 'tftp' tool and try again                                       
<mn>: DHCP service works well                                                                                     [ OK ]
<mn>: Firewall is closed on current server                                                                        [ OK ]
<mn>: The disk space is enough for xCAT to work                                                                   [ OK ]
<mn>: The IP of master is a static IP address                                                                     [ OK ]
<mn>: The dhcpd.leases file is less than 100M                                                                     [ OK ]
======================do summary=====================                                                             
[MN]: Check on MN FAILED.                                                                                         [FAIL]
    TFTP service works well                                                                                       [FAIL]
        TFTP check need 'tftp' tool, please install 'tftp' tool and try again                                     
```

I suggest that we put a requirement on this tool in the xcatprobe rpm so that the user does not need to do additional work.  

If not, at least we should print those as WARN, instead of failures.  Because we cannot validate without the ``tftp`` tool.

hu-weihua:
@whowutwut , I add ``tftp``, ``nslookup`` and ``wget`` as ``xcatmn``'s   requirement by PR #2122 , Could you review this pull request?

whowutwut:
I have verified using the following version

```
Version 2.13.1 (git commit c7eb8b8025d9b8c815b3630461294e5af9803747, built Thu Dec 15 06:15:40 EST 2016)
```

There are dependencies now: 
```
# rpm -qi --requires xCAT-probe-2.13.1-snap201612141342.noarch |
....
/usr/bin/nslookup
/usr/bin/perl
[...]

Issue #1725, opened 2016-08-17
xcatprobe xcatmn usage errors in help menu


whowutwut:
Few comments on xcatprobe: 
- The `-i` option is required , so it should not be surrounded by `[ ]` brackets which indicate optional argumnets... 
- replace `xcat` with `xCAT` to be consistent with the naming of the product.  

```
[root@c910f04x42vm20 ~]# xcatprobe xcatmn
Option -i is required                                                                                             [FAIL]
Usage:                                                                                                            
    xcatmn -h
    xcatmn [-i <install_nic>] [-V]  <---- indicates optional

Description:
    After xcat installation, use this command to check if xcat has been installed correctly and is ready for use.
    For hierarchical cluster, only the provision network on the same network as the management node is supported. If ...

Options:
    -h : Get usage information of xcatmn
    -V : Output more information for debug
    -i : Required. Specify the network interface name of provision network on management node

```

whowutwut:
I'm ok to move this to 2.12.3..

hu-weihua:
I fixed this bug by PR #2071 .

whowutwut:
@hu-weihua  I ran this on the RC1 build today, it seems like the help message changes but nothing blocks if the `-i` option is not specified.. 

```
[root@fs4 yum.repos.d]# xcatprobe xcatmn 
<mn>: All xCAT deamons are running                                                                                [ OK ]
...
```

whowutwut:
In addition, when I run the command with `-i` specified.. .seems like the command hangs forever. 

```
[root@fs4 yum.repos.d]# xcatprobe xcatmn -i enP4p1s0f2
<mn>: All xCAT deamons are running                                                                                [ OK ]
<mn>: xcatd can receive command request                                                                           [ OK ]
<mn>: 'site' table is configured correctly                                                                        [ OK ]
<mn>: Provision network is configured correctly                                                                   [ OK ]
<mn>: 'passwd' table is configured correctly                                                                      [ OK ]
<mn>: Important directory is configured correctly                                                                 [ OK ]
<mn>: SELinux is disabled on current server                                                                       [ OK ]
<mn>: HTTP service works well                                                                                     [ OK ]
<mn>: TFTP service works well                                                                                     [FAIL]

<waiting here for some time..never come back> 
^C

^C

^C^Z
[1]+  Stopped                 xcatprobe xcatmn -i enP4p1s0f2
[root@fs4 yum.repos.d]# ps -ef | grep xcatprobe
[...]

PR #2122, opened 2016-11-09
Fix xcatmn's bugs #2110  and  #1725

hu-weihua:
This pull request fix below 2 bugs:

1  If the  ``-i`` option is not specified for ``xcatmn``,  ``xcatmn`` should not run because option ``i`` is required for ``xcatmn``. Refer to issue #1725 

2 ``xcatmn`` need 3 tools (wget, tftp, nslookup) to do probe, so add these tools as xCAT-probe package's dependency. Refer to issue #2110

### DIFF
--- a/xCAT-probe/debian/control
+++ b/xCAT-probe/debian/control
@@ -8,4 +8,8 @@ Standards-Version: 3.7.2
 Package: xcat-probe
 Architecture: all
 Depends: ${perl:Depends}
+
+#Sub-command 'xcatmn' need tools tftp, nslookup, wget. 
+Suggests: wget, dnsutils, tftp-hpa
+
 Description: Provides a toolkits to help probe all the possible issues in xCAT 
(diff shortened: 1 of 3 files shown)


hu-weihua:
Due to change the scripts of builder is risk for the xcat release, so decide to split this pull request to 2. Close this  one .

xCAT-probe/debian/control
```diff
@@ -8,4 +8,8 @@ Standards-Version: 3.7.2
 Package: xcat-probe
 Architecture: all
 Depends: ${perl:Depends}
+
+#Sub-command 'xcatmn' need tools tftp, nslookup, wget. 
+Suggests: wget, dnsutils, tftp-hpa
+
 Description: Provides a toolkits to help probe all the possible issues in xCAT
```
neo954: `... a toolkit ...`

Consistency of singular vs. plural usage problem

PR #2126, opened 2016-11-09
[DO NOT MERGE] Fix issue #2110 xcatprobe xcatmn requires tftp tool , add dependency to xcatprobe rpm

hu-weihua:
Fix bug #2110 .

### DIFF
--- a/xCAT-probe/xCAT-probe.spec
+++ b/xCAT-probe/xCAT-probe.spec
@@ -16,6 +16,13 @@ BuildRoot: /var/tmp/%{name}-%{version}-%{release}-root
 BuildArch: noarch
 %endif
 
+%ifos linux
+#Below tools are required by sub-command 'xcatmn' 
+Requires: /usr/bin/nslookup
+Requires: /usr/bin/tftp
+Requires: /usr/bin/wget
+%endif
+
 Provides: xCAT-probe = %{version}
 
 %description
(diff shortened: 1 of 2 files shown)


hu-weihua:
@whowutwut , this pull request fix bug ``xcatprobe xcatmn requires tftp tool , add dependency to xcatprobe rpm``. but due to it's too close to release, so it's better not to modify the builder scripts. 
We think this bug can be really fixed in next sprint. What do you think?

xCAT-probe/xCAT-probe.spec
```diff
@@ -16,6 +16,13 @@ BuildRoot: /var/tmp/%{name}-%{version}-%{release}-root
 BuildArch: noarch
 %endif
 
+%ifos linux
+#Below tools are required by sub-command 'xcatmn' 
+Requires: /usr/bin/nslookup
```
whowutwut: I think we can provide the rpm package name as a requirement....instead of the path to the binary. 

nslookup is provided by bind-utils, I think the following would do the same thing in a single line
`Requires: bind-utils wget tftp`

whowutwut:
Sure, let's wait until 2.13 release to make this change.

xCAT-probe/xCAT-probe.spec
hu-weihua: I consulted @immarvin, if it is possible for one command to be included in different package in different distro and version.  His answer is in the affirmative. So he recommend me to use ``/usr/bin/nslookup`` instead of ``bind-utils``.  I think maybe it's a better option.

whowutwut: ok, sure. Sounds good.

whowutwut:
I'm ok with the changes to be merge when ready

neo954:
I will merge this one.

Issue #2131, opened 2016-11-10
[FVT]mkdef with osimage returns object name " invalid


tingtli:
xcat 2.12.4 on rhels and sles

```
c910f04x18v02:~ # mkdef -t osimage -o sles12.1-x86_64-statelite-compute -u profile=compute provmethod=statelite osvers=sles12.1
Error: The object name '' is invalid, please refer to "man xcatdb" for the valid "xCAT Object Name Format"
c910f04x18v02:~ # echo $?
1
c910f04x18v02:~ # lsxcatd -v
Version 2.12.4 (git commit 2d6354d0a87ba18cc763f3d1c80df09ab940ded2, built Wed Nov  9 06:35:54 EST 2016)
```

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2134/files

PR #2134, opened 2016-11-10
fix [FVT]mkdef with osimage returns object name  invalid #2131

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2131


daniceexi:

Issue #2115, opened 2016-11-09
[FVT]: diskless provision is in netbooting status after the privision is finished


tingtli:
xCAT 2.12.4 both on rhels7.3 and sles11.4
Steps to reproduce
1. Install xCAT 2.12.4
2. Define CN 
3. Do diskless provision on CN
4. After the provision, check CN's status

```
[root@c910f02c08p05 autotest]# lsxcatd -v
Version 2.12.4 (git commit 1b5160257787eb027ecc5311ec9a3778f903657b, built Mon Nov  7 06:16:10 EST 2016)
[root@c910f02c08p05 autotest]# lsdef -l c910f02c08p07
Object name: c910f02c08p07
    arch=ppc64
    cons=hmc
    currchain=boot
    currstate=netboot rhels7.3-ppc64-compute
    groups=all
    hcp=c910hmc01
    hidden=0
    hwtype=lpar
    id=7
    initrd=xcat/osimage/rhels7.3-ppc64-netboot-compute/initrd-stateless.gz
    kcmdline=imgurl=http://c910f02c08p05:80//install/netboot/rhels7.3/ppc64/compute/rootimg.cpio.gz XCAT=c910f02c08p05:3001 NODE=c910f02c08p07 FC=0 BOOTIF=4a:c8:fd:4a:c0:03
    kernel=xcat/osimage/rhels7.3-ppc64-netboot-compute/kernel
    mac=4a:c8:fd:4a:c0:03
    mgt=hmc
    monserver=c910f02c08p05
    netboot=grub2
    nfsserver=c910f02c08p05
    nodetype=ppc
    os=rhels7.3
    parent=c910f02fsp08
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    pprofile=c910f02c08p07
    profile=compute
    provmethod=rhels7.3-ppc64-netboot-compute
    status=netbooting        <== here status should be booted
    statustime=11-08-2016 21:40:35
    tftpserver=c910f02c08p05
    updatestatus=synced
    updatestatustime=11-07-2016 14:26:05
    xcatmaster=c910f02c08p05
```
The node is in problem state, please debug on this node. Thanks!

immarvin:
duplicate with https://github.com/xcat2/xcat-core/issues/1939
fixed in https://github.com/xcat2/xcat-core/pull/2113

can FVT verification this ? thanks

immarvin:
[...]

PR #2137, opened 2016-11-10
create /opt/xcat under rootimage dir incase it is not there

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2115


daniceexi:
I have reviewed this change.

Issue #2130, opened 2016-11-10
[CUSTOMER] the old diskless image built by xCAT can not be used after xcat 2.12.2 


immarvin:
this is a problem from mail list

From: peter CZ1 Peng [mailto:pengcz1@lenovo.com]
Sent: Wednesday, November 09, 2016 2:20 AM
To: xcat-user@lists.sourceforge.net
Subject: [xcat-user] xcat build old diskless image can not use in the new xcat 2.12.2
hi ,all
  I meet a new issue when I ungrade my xcat from 2.11.1 to 2.12.2
  after upgrade ,and I try to preload my image (nodeset node01 osimage=rhels7.2-x86_64-netboot-compute ),it have something change

````
[root@mgt compute]# cat /tftpboot/xcat/xnba/nodes/node01.elilo
default="xCAT"
delay=0
image=/tftpboot/xcat/osimage/rhels7.2-x86_64-netboot-compute/kernel
  label="xCAT"
 initrd=/tftpboot/xcat/osimage/rhels7.2-x86_64-netboot-compute/initrd-stateless.gz
  append="imgurl=http://%N:80//install/netboot/rhels7.2/x86_64/compute/rootimg.cpio.gz XCAT=%N:3001 NODE=node01 FC=no ifname=eno2:40:f2:e9:c8:d4:fc netdev=eno2  console=tty0 console=ttyS0,115200n8r BOOTIF=%B"
[root@mgt compute]# cat /tftpboot/xcat/xnba/nodes/node01.
node01.elilo  node01.uefi
[root@mgt compute]# cat /tftpboot/xcat/xnba/nodes/node01.uefi
#!gpxe

chain http://${next-server}/tftpboot/xcat/elilo-x64.efi -C /tftpboot/xcat/xnba/nodes/node01.elilo

[root@mgt compute]#
````

when I build my rootimg in xcat 2.11.1 ,it is a default format with the name rootimg.gz, and now I copy rootimg.gz to rootimg.cpio.gz ,but it can not work as show as below:

````
Getting http://172.20.0.1:80//install/netboot/rhels7.2/x86_64/compute/rootimg.cpio.gz...
--2016-11-09 23:10:12--  http://172.20.0.1//install/netboot/rhels7.2/x86_64/compute/rootimg.cpio.gz
Connecting to 172.20.0.1:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 445923212 (425M) [application/x-gzip]
[...]

PR #2138, opened 2016-11-10
fix issue [CUSTOMER] the old diskless image built by xCAT can not be used after xcat 2.12.2 #2130

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2130


daniceexi:

PR #2144, opened 2016-11-10
Disable ac97 for ppc64le platforms for KVM

whowutwut:
Problems seem to happen when we power on RHEV KVMs when we request the ac97 driver. 
This just disables the support for ppc64le for now...


zet809:

PR #2149, opened 2016-11-14
fix probe discovery usage bug

xuweibj:
If use ``xcatprobe discovery -h``, will show "-m : The method of discovery, the valid values are .". So modified code.


hu-weihua:

PR #2150, opened 2016-11-14
modified probe discovery usage

xuweibj:


xCAT-probe/subcmds/discovery
```diff
@@ -46,7 +46,7 @@ $::USAGE = "Usage:
 
 Description:
     Probe the discovery process, including pre-check for required configuration, 'Realtime monitor' and 'Replay history' of discovery process.
```
whowutwut: Suggest changing this line also.... 

Probe the discovery process, including pre-check for required configuration.  If any pre-checks fail, '$program_name' will exit.

xCAT-probe/subcmds/discovery
```diff
@@ -46,7 +46,7 @@ $::USAGE = "Usage:
 
 Description:
     Probe the discovery process, including pre-check for required configuration, 'Realtime monitor' and 'Replay history' of discovery process. 
-    If all pre-check items failed, $program_name will exit.  
+    If any pre-check item fails, $program_name will exit.  
     Realtime monitor: This is a default. This tool with monitor discovery state of the node. Trigger 'Realtime monitor' before rebooting target node to do discovery.
```
whowutwut: Also suggest changing section to be: 

The default behavior of xcatprobe discovery is to monitor the discovery process in real time.  Use the -r option for "replay" mode which probes the discovery based on information stored in various log files.

whowutwut: Below in the options, we should format this so that the : lines up...

```
    -h           : Get usage information of discovery.
    -V           : Output more information for debug.
    -m           : The method of discovery, the valid values are .
    -n           : The range of predefined nodes, must be used with option -m.
    --noprecheck : skip pre-checking discovery to validate correct configuration. 
   ... 
```

whowutwut:
I added some comments in the general help since you are making some changes here. . I might have some more once you make the above changes.  THanks @xuweibj !

xuweibj:
Hi @whowutwut , thank you for your comments. I modified depending on comments, please help review again.

whowutwut:
Thanks!  I will merge

Issue #2157, opened 2016-11-15
"xCAT Object Name Format" error when trying to name osimages starting with "rhels7.3"


mew2057:
When I create an image with the mkdef -z directive and the following name "rhels7.3-diskless-test" I receive the following error message:

> Error: The object name 'rhels7.3-diskless-test' is invalid, please refer to "man xcatdb" for the valid "xCAT Object Name Format"

The following other image names were considered valid using the same process:

> diskless  (osimage)
diskless-test  (osimage)
diskless91012  (osimage)
diskless9101214161820  (osimage)

It appears that the if the rhels7.3 token is present this error will persist.

pdlun92:
Experimenting further with `mkdef -z` yields following results:
`rhels7.2-GA-ppc64le-RHEV4-install-compute` `rhels7.3-test` and `rhels7.3` return with the error

`rhels-GA-ppc64le-RHEV4-install-compute` `rhels72-GA-ppc64le-RHEV4-install-compute` `rhels7-GA-ppc64le-RHEV4-install-compute` and `rhels-test` are accepted

mkdef seems to be only command affected, as chdef from a name accepted by `mkdef -z` to one that isn't will successfully change the name.  Example:
`rhels7.3-test` is not accepted by mkdef -z
`rhels-test` IS accepted by mkdef -z
then
```[root@fs4 osimage]# chdef -t osimage rhels-test -n rhels7.3-test
Changed the object name from rhels-test to rhels7.3-test.
[root@fs4 osimage]# lsdef -t osimage -o rhels7.3-test
Object name: rhels7.3-test
```

furthermore, running `mkdef` with `-u` option will run successfully with `rhels7.3-test` as name while `mkdef -z` will NOT
```[root@fs4 ~]# mkdef rhels7.3-test -u profile=compute provmethod=statelite
1 object definitions have been created or modified.
```

In short, it appears that the stanza `-z` option for mkdef osimage is unable to correctly parse `.`

pdlun92:
Traced this back to #2134 

Added print statements to indicate what `$objtype` was being passed to/evaluated in `sub isobjnamevalid` 

```
[root@fs4 osimage]# export XCATBYPASS=TRUE
[root@fs4 osimage]# cat test.stanza | mkdef -z
objtype =  BEFORE
objtype = node AFTER
Error: The object name 'rhels7.3-test' is invalid, please refer to "man xcatdb" for the valid "xCAT Object Name Format"
```

[...]

PR #2160, opened 2016-11-16
fix issue "xCAT Object Name Format" error when trying to name osimages starting with "rhels7.3" #2157

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2157


immarvin:
fixed in https://github.com/xcat2/xcat-core/issues/2157

pdlun92:
Looks good to me.  

Unit test with two different stanza files, one named "rhels7.3-test" with objtype=osimage and another named "node.test" with objtype=node.  osimage stanza will create, node stanza will give error, which is expected and intended behavior.

Agree to merge

immarvin:
thanks @pdlun92

Issue #2132, opened 2016-11-10
"makedhcp -n" failed to add subnet info to dhcpd.conf


cxhong:
makedhcp command failed to add subnet info to dhcpd.conf 

`````
# makedhcp -n
Renamed existing dhcp configuration file to  /etc/dhcp/dhcpd.conf.xcatbak

Error: Could not add the subnet 2002:903:15f:2320::/64 for interface eth0 into /etc/dhcp/dhcpd.conf.
Please verify the xCAT database matches networks defined on this system.
```````

dump the network table:
`````
# tabudump networks
#netname,net,mask,mgtifname,gateway,dhcpserver,tftpserver,nameservers,ntpservers,logservers,dynamicrange,staticrange,staticrangeincrement,nodehostname,ddnsdomain,vlanid,domain,comments,disable
"public","9.3.232.0","255.255.254.0","eth0","9.3.232.1",,"9.3.232.60",,,,,,,,,,,,
"provision","10.10.1.0","255.255.255.0","eth2","<xcatmaster>",,"10.10.1.10",,,,"10.10.1.200-10.10.1.232",,,,,,"cluster.xcat",,
"bmc","10.10.3.0","255.255.255.0","eth3","<xcatmaster>",,"10.10.3.12",,,,,,,,,,"cluster.xcat",,
"docker","172.17.0.0","255.255.0.0","docker0","<xcatmaster>",,"172.17.0.1",,,,,,,,,,,,
"ib0ipv41","20.0.0.0","255.255.255.0","ib0",,,,,,,,,,,,,,,
"ib0ipv42","30.0.0.0","255.255.255.0","ib0",,,,,,,,,,,,,,,
"ib0ipv61","1:2::/64","/64","ib0","1:2::3",,,,,,,,,,,,,,
"ib0ipv62","2:2::/64","/64","ib0","2:2::3",,,,,,,,,,,,,,
``````````

The subnet of 2002:903:15f:2320::/64  is from eth0 interface.
`````
# ifconfig eth0
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 9.3.232.60  netmask 255.255.254.0  broadcast 9.3.233.255
        inet6 2002:903:15f:2320:a242:3fff:fe31:1d6a  prefixlen 64  scopeid 0x0<global>
        inet6 fe80::a242:3fff:fe31:1d6a  prefixlen 64  scopeid 0x20<link>
[...]

PR #2162, opened 2016-11-16
Filter out global local ipv6 address for makedhcp command

cxhong:
This is for issue #2132 .  From ipv6 documentation:
``````
The following prefixes have been reserved for special use:

2002::/16
Indicates that a 6to4 routing prefix follows.

fe80::/10
Indicates that a link-local address follows.

ff00::/8
Indicates that a multicast address follows.
```````

So, I think it's safe for us to filter out prefix 2002 for ipv6 global local address, as we did for prefix fe80 for ipv6 link local address.


bybai:
Hi @cxhong and @immarvin ,
I agreed that it's safe for us to filter out prefix 2002 for ipv6 global local address just like we did for prefix fe80.
I am OK for this fix.

immarvin:
agree to merge

PR #2164, opened 2016-11-16
Added sample scripts for the various node roles in csm. Each postscri…

mew2057:
…pt will scp the latest version of the config file appropriate for it and then start the daemon. The master script will perform some degree of automatic configuration.

Added the sample csm postscripts discussed.

### DIFF
--- a/xCAT-csm/install/postscripts/csm_post
+++ b/xCAT-csm/install/postscripts/csm_post
@@ -1,1 +1,0 @@
-#example postscript for CSM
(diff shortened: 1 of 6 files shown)


whowutwut:
@mew2057  Did we want to put this in a `csm` subdirectory under `/install/postscripts`.  You should be able to build the rpm locally under xcat-core using `./makerpm xCAT-csm`

mew2057:
It might not be the right move to put it in a sub directory if the post scripts have the `ibm-csm-*` naming convention. I believe it was determined that they should just be in   `/install/postscripts`. I've updated the spec file to reflect this change.

whowutwut:
Ok, Will merge this so @mew2057  can start testing

Issue #1890, opened 2016-09-23
cryptmethod support on linux


khm:
Currently, xCAT passwd table does not support any sort of cryptmethod except md5 on linux.  glibc and musl both support sha256 ($5) and sha512 ($6) crypt methods, so there's no good reason to continue defaulting to md5.

Please consider at least introducing support for other cryptmethods, and preferably making the default choice stronger.

whowutwut:
Thanks @khm   I also agree that this is something we should look into.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2168

whowutwut:
@tingtli  If @khm  cannot verify this for the 2.13.0 release, can you verify and close out?

khm:
Sorry; didn't know you were waiting on my feedback.  This looks excellent.  Thanks so much for the quick implementation!

PR #2168, opened 2016-11-18
Load the cryptmethod in the passwd table to generate the password

chenglch:
sha256 and sha512 crypt methods are supported from glibc 2.7 version.
This patch enable the crypt feature on linux system and encrypt the
password with sha256 by default.

implement: #1890


immarvin:
agree to merge

daniceexi:
👍

PR #2171, opened 2016-11-18
Add security bulletin for removal of hard coded password

whowutwut:
Prepare the security bulletin for the vulnerability for having hard coded HMC/FSP password used by Service teams.  Target to merge this Nov 30, 2016.  The 2.9.x release needs to be built and released at the same date.


whowutwut:
@zet809  We can merge this and it will be part of the "latest" URL in the documentation.  We can reference that in the release notes.  @tingtli  I had created some release notes URL here in the documentation, we should make sure the release notes is matched to that URL

docs/source/security/2016/20161130_hard_code_password.rst
```diff
@@ -0,0 +1,60 @@
...
++-------------------------+-----------------------------------+---------------------------------------+
+| **2.11.x**              | Update to **2.12.3**, or higher   | `2.12.3 Release Notes <https://       |
+|                         |                                   | github.com/xcat2/xcat-core/wiki       |
+|                         |                                   | /XCAT_2.12.3_Release_Notes>`_         |
++-------------------------+-----------------------------------+---------------------------------------+
+| **2.10.x**              | Update to **2.12.3**, or higher   | `2.12.3 Release Notes <https://       |
+|                         |                                   | github.com/xcat2/xcat-core/wiki       |
+|                         |                                   | /XCAT_2.12.3_Release_Notes>`_         |
++-------------------------+-----------------------------------+---------------------------------------+
+| **2.9.x**               | Update to **2.9.4**, or higher    | `2.9.4 Release Notes <https://        |
```
zet809: @whowutwut Hi, Victor, in fact for 2.7.10 and 2.9.4, they are just for AIX OS. If the OS of MN is non-AIX, we shall advise the user to update xCAT to 2.12.3 or higher.

whowutwut:
Thanks @zet809  I updated the documentation. 

I think we should not merge this in until we decide on the version for the 2.7 and 2.9 release.  I think we are leaning towards 2.7.10 and 2.9.4

docs/source/security/2016/20161130_hard_code_password.rst
```diff
@@ -0,0 +1,63 @@
...
++-------------------------+-----------------------------------------------+---------------------------------------+
+| **2.11.x**              | Update to **2.12.3**, or higher               | `2.12.3 Release Notes <https://       |
+|                         |                                               | github.com/xcat2/xcat-core/wiki       |
+|                         |                                               | /XCAT_2.12.3_Release_Notes>`_         |
++-------------------------+-----------------------------------------------+---------------------------------------+
+| **2.10.x**              | Update to **2.12.3**, or higher               | `2.12.3 Release Notes <https://       |
+|                         |                                               | github.com/xcat2/xcat-core/wiki       |
+|                         |                                               | /XCAT_2.12.3_Release_Notes>`_         |
++-------------------------+-----------------------------------------------+---------------------------------------+
+| **2.9.x**               | Update to:                                    | `2.9.4 Release Notes <https://        |
```
zet809: I think for xCAT version older than 2.9.x, we advise the user to update xCAT to 2.12.3 and higher for Redhat. And update to 2.9.4 for AIX.
So, we can delete the rows for 2.8.*, 2.7.* and 2.6.* and lower.

whowutwut:
@zet809  The pull request is ready to be merged.

zet809:
I think we can merge this after 2.13 release. @whowutwut What is your idea?

whowutwut:
I think we should merge this when we are ready with the 2.9.4 release and reference this page in the release notes for 2.9.4...This  is not tied to the 2.13 release but if we do not merge soon it won't be in the "stable" documentation which I think should be the link used to reference this security bulletin. Shouldn't we be ready with the 2.9.4 release soon?

whowutwut:
I think perhaps we also want to put this security bulletin in the 2.9 documentation.. I will create another pull request for that.

zet809:

Issue #2174, opened 2016-11-19
[DEV] several issues on generating  grub.cfg-xxx when netboot="grub2" 


immarvin:
1. when node.installnic="xx-xx-xx-xx-xx-xx", there is no /tftpboot/boot/grub2/grub.cfg-01-xx-xx-xx-xx-xx-xx genrated.

2. for the grub config file  " /tftpboot/boot/grub2/grub.cfg-<node ip in hex>" , the characters in <node ip in hex> should be in upper case
````
[root@c910f03c05k21 ~]# xcatprobe osdeploy -n c910f03c05k27
The install NIC in current server is eth0                                                                         [INFO]
eth1
All nodes to be deployed are valid                                                                                [ OK ]
-------------------------------------------------------------
Start capturing every message during OS provision process....
-------------------------------------------------------------

[c910f03c05k27] Use command rpower to reboot node c910f03c05k27
[c910f03c05k27] Node status is changed to powering-on
[c910f03c05k27] Via TFTP download /boot/grub2/grub2-c910f03c05k27
[c910f03c05k27] Via TFTP download /boot/grub2/powerpc-ieee1275/normal.mod
[c910f03c05k27] Via TFTP download /boot/grub2/powerpc-ieee1275/terminal.mod
[c910f03c05k27] Via TFTP download /boot/grub2/powerpc-ieee1275/crypto.mod
[c910f03c05k27] Via TFTP download /boot/grub2/powerpc-ieee1275/gettext.mod
[c910f03c05k27] Via TFTP download /boot/grub2/powerpc-ieee1275/gzio.mod
...
[c910f03c05k27] Via TFTP download //boot/grub2/grub.cfg-0A03051B
...
````

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2175

Issue #1502, opened 2016-07-07
kcmdline for centos netboot images set incorrectly


samveen:
I'm using xCAT-2.9.1. I set up a netboot osimage by cloning and modifying the basic `*-netboot-compute` image. I'm using it to boot a node `spare1`:

```
[root@mn1 ~]# nodeset spare1 osimage=centos7.1-x86_64-netboot-compute-custom
spare1: netboot centos7.1-x86_64-compute
[root@mn1 ~]# rsetboot spare1 net
spare1: Network
[root@mn1 ~]# rpower spare1 boot
spare1: reset
```

When I boot the node off this image, dracut always fails (into a debug shell) with an error: `"Invalid arguments for ifname="`.

On looking at my node definition:

```
[root@mn1 ~]# lsdef -t node spare1 -i kcmdline
Object name: spare1
    kcmdline=imgurl=http://!myipfn!:80//install/netboot/centos7.1/x86_64/compute-custom/rootimg.gz XCAT=!myipfn!:3001 NODE=spare1 FC=0 ifname=eno1: netdev=eno1  console=tty0 console=ttyS1,115200
```

I find that the `ifname` param is set to just `eno1:`, while it should be set to `eno1:XX:XX:XX:XX:XX:XX`

I have confirmed this in the initrd code:

```
[root@mn1 initrd_exploded]# tail  usr/lib/net-lib.sh -n +459 |head -n 25
parse_ifname_opts() {
    local IFS=:
    set $1

    case $# in
        7)
            ifname_if=$1
            # udev requires MAC addresses to be lower case
            ifname_mac=$(echo $2:$3:$4:$5:$6:$7 | sed 'y/ABCDEF/abcdef/')
            ;;
        *)
            die "Invalid arguments for ifname="
            ;;
    esac

    case $ifname_if in
        eth[0-9]|eth[0-9][0-9]|eth[0-9][0-9][0-9]|eth[0-9][0-9][0-9][0-9])
            warn "ifname=$ifname_if uses the kernel name space for interfaces"
            warn "This can fail for multiple network interfaces and is discouraged!"
            warn "Please use a custom name like \"netboot\" or \"bluesocket\""
            warn "or use biosdevname and no ifname= at all."
            ;;
    esac

}
```

To get my nodes booting correctly, I'm forced to do the following workaround: I'm manually updating the xnba configuration for the node at `/tftpboot/xcat/xnba/nodes/spare1` to contain the correct value for `ifname`.

[...]

PR #2175, opened 2016-11-19
fix several issues on redhat kernel cmdline  and grub2 config file generating

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2174

fix issue https://github.com/xcat2/xcat-core/issues/1502:
according to dracut documentation https://www.kernel.org/pub/linux/utils/boot/dracut/dracut.html#_troubleshooting :
````
ifname=<interface>:<MAC>
Assign network device name <interface> (ie "bootnet") to the NIC with MAC <MAC>.

Warning
Do not use the default kernel naming scheme for the interface name, as it can conflict with the kernel names. So, don’t use "eth[0-9]+" for the interface name. Better name it "bootnet" or "bluesocket".

....

bootdev=<interface>
specify network interface to use routing and netroot information from. Required if multiple ip= lines are used.
````
"ifname" is used to assign the name for the nic with specified mac address, it should not be in kernel cmdline for provision.
"bootdev" should be the nic name instead of mac address

### DIFF
--- a/xCAT-server/lib/xcat/plugins/grub2.pm
+++ b/xCAT-server/lib/xcat/plugins/grub2.pm
@@ -302,7 +302,6 @@ sub setstate {
             $macstring = $ment->{mac};
             my @macs = split(/\|/, $ment->{mac});
             foreach (@macs) {
-                $nodemac = $_;
                 if (/!(.*)/) {
                     my $ipaddr = xCAT::NetworkUtils->getipaddr($1);
                     if ($ipaddr) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: xCAT-server/lib/xcat/plugins/grub2.pm [hunk 1]
```diff
@@ -317,15 +316,27 @@ sub setstate {
...
         if ($cref and $cref->{currstate} ne "offline") {
             link($tftpdir . "/boot/grub2/" . $node, $tftpdir . "/boot/grub2/" . $pname);
         }
     }
-    if ($macstring) {
+    
+    my $myinstallnic;
+    my $nrent=$nrhash{$node}->[0];
+    if($nrent and $nrent->{installnic}){
+        my $myinstallnic=$nrent->{installnic};
```
daniceexi: The $myinstallnic has been defined before.

immarvin:
hi @daniceexi ,thanks for the comments, code updated

daniceexi:

PR #2178, opened 2016-11-21
fix 2025 Re-enable the creation of statelite when copycds on RHEL 7

bybai:
After copycds, osimage definition rhels*-*-statelite-compute is created.


immarvin:
I am ok with the PR
thanks @bybai

PR #2187, opened 2016-11-22
Fix issue for genesis ppc64, wait 2+number_of_nics seconds for all NICS to come up

zet809:
Fix issue for genesis ppc64, wait 2+number_of_nics seconds for all NICS to come UP

To fix the issue that the MAC address of install NICS doesn't appear in the mac attribute of the matching node.

### DIFF
--- a/xCAT-genesis-builder/xcat-cmdline.sh
+++ b/xCAT-genesis-builder/xcat-cmdline.sh
@@ -51,7 +51,7 @@ if [ $ARCH = "ppc64le" ]; then
 fi
 
 if [ $ARCH == 'ppc64' ]; then
-
+    waittime=2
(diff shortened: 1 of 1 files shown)


whowutwut:
I have reviewed the changes and agree to merge.  Thanks @zet809 
Do we know why even with certain machines that are the same ,we get the different times to bring up the NICs?

zet809:
@whowutwut Hi, Victor, not sure about that, maybe because the physical device itself.
Besides this pull request, a new genesis-base for ppc64 is also needed, since the xcat-cmdline.sh is included in the genesis-base package. I have uploaded a latest one onto GSA for both RH and ubuntu.

adambertsch:
I see.  Please close this issue with my thanks.

Issue #2188, opened 2016-11-22
[fvt] nodeset node diskless image failed for could not find the diskless image


junxiawang:
Env:sles ubuntu
Build:c910f02c01p14:~ # lsxcatd  -v
Version 2.13 (git commit 6653dfdc6acc292cf7ccd8e053263009a6ea960a, built Mon Nov 21 06:15:37 EST 2016)
c910f02c01p14:~ #

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2189

PR #2189, opened 2016-11-22
fix [fvt] nodeset node diskless image failed for could not find the diskless image #2188

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2188


bybai:
I am OK to merge it. Thx

PR #2190, opened 2016-11-23
check whether the network access between MN/CN and the node is ready with ping instead of checking whether the installnic is ip

immarvin:
background: https://github.com/xcat2/xcat-core/pull/2088


bybai:
I am OK

PR #2191, opened 2016-11-23
Modify building failure in ubuntu

hu-weihua:


xCAT-probe/debian/control
```diff
@@ -7,9 +7,5 @@ Standards-Version: 3.7.2
 
 Package: xcat-probe
 Architecture: all
-Depends: ${perl:Depends}
-
-#Sub-command 'xcatmn' need tools tftp, nslookup, wget. 
-Suggests: wget, dnsutils, tftp-hpa
-
+Depends: wget, dnsutils, tftp-hpa, ${perl:Depends}
```
neo954: I still suggest use `Recommends` for package `wget`, `dnsutils` and `tftp-hpa`. Since based on Debian Reference Book, see URL below, it is not a `absolute dependency' semantics here. Package `xcat-probe` can run without `wget`, `dnsutils` and `tftp-hpa` if the administrator knows who he/she is doing.

https://www.debian.org/doc/manuals/debian-reference/ch02.en.html#_package_dependencies

hu-weihua: Yes, ubuntu has more detailed label for ``Package dependencies``, but redhat only has ``Requires``, that equals to ``depends`` of ubuntu.  So in order to maintain unity, I choose ``Depends`` in ubuntu.

neo954:

Issue #1547, opened 2016-07-18
Please update xCAT.spec and xCATsn.spec to have proper package dependency


neo954:
Since xCAT requires all the xCAT packages are on the same version level. xCAT should put all these requirement in its RPM spec file.

A proper fix of this bug should allow the end user to upgrade xCAT with a simple command.

```
 yum install xCAT
```

neo954:
In this way, xCAT will not obfuscate rookie user after all. See issue #1523 as an example of how xCAT could obfuscate rookie user.

neo954:
The bug is partly covered by the script ``go-xcat``. But I think even though we have ``go-xcat``, this is a bug of xCAT itself, and should be fixed properly.

whowutwut:
:+1:  Nice @neo954  Yes, it would make things a lot easier to not use Wildcards and have the packages do exactly as they should

daniceexi:
Good idea. Now I think we need a way to set the dependency version from the **Version** file instead of changing the .spec files for each Version change.

zet809:
Will deal with this issue in 2.12.3.

whowutwut:
Since this is a big change in behavior, should we deal with this and target a 2.13 release? Maybe we should have the changes in separate branch off of `xcat2/xcat-core` and we can test it separately, then if we feel it's good to go, we can merge it into master, instead of just directly merging into master?

zet809:
The pull request #1659 and #1385 can fix this issue.

neo954:
Verified with xCAT version 2.12.3-snap201609290802. This is the xCAT 2.12.3 GA version.

neo954:
Failed to close this bug. re-open this one.

neo954:
I reviewed pull request #1659 and #1385. I do not see any line related to this issue.

neo954:
Verified with xCAT 2.12.3 GA version. And failed. This problem still exist.

zet809:
We didn't add version number  for depended packages, will deal with it later.

zet809:
Move to next release to do more investigation.

neo954:
Since @zet809 did not spend time on this issue. I tried to fix this issue by myself.

daniceexi:
@neo954 That's good. But we may not put this change in the 2.13 release since it's a little risky.

neo954:
@daniceexi That is fine. Just merge this pull request after xCAT 2.13 GA.

whowutwut:
[...]

PR #2195, opened 2016-11-23
Make all the rpm packages have the exactly same version and release numbers

neo954:
This pull request try to fix issue #1547.


immarvin:
please do not merge this before xCAT 2.13 is released

neo954:
Hello @jjohnson42,

I read your pull request #1814. It is great if you have time and can take a look at the patches I made.

whowutwut:
@neo954  In your description for this PR , can you should the before and after result of the changes made in this pull request regarding the naming of the xCAT rpms? 

Also, do we have a plan for the naming convention for the xCAT related RPMs in xcat-dep.  Do we know whether we can use the new naming with the old xcat-deps... we don't want to break backwards compatibility. 

I see you target-ed this for 2.14...  I think that is good.  There's no ETA for 2.14 and we probably should start defining 2.13.x release and figure out when 2.14 would be released.

neo954:
@whowutwut, 

The naming of all the xCAT RPMs will remain as before. That is something like `xCAT-2.13-snap201612010059.rpm` But all the xCAT RPMs will have the same time tag for the `snap201612010059` part.

Plus, all the dependency packages in `xCAT-core` RPMs will need to have the exactly same version number. Please see issue #1547, for the background story.

For example, package `xCAT` version `2.13-snap201612010059` requires `xCAT-client` version `2.13-snap201612010059`, `xCAT-server` version `2.13-snap201612010059`, `xCAT-probe` version `2.13-snap201612010059`, `xCAT-genesis-scripts-x86_64` version `2.13-snap201612010059` and `xCAT-genesis-scripts-ppc64` version `2.13-snap201612010059`.

In the past, there is no version number requirement, or there is version number requirement, but it is something like `>= 2.12`.

xCAT use to have some kind of cross version compatibility. That means, user can mixed different version of `xCAT-server` and `xCAT-client. But currently, xCAT does not provide cross version compatibility any more. That is due to, when we code xCAT, we did not put cross version compatibility in our mind. Thus xCAT should not mislead the end user by faulty RPM spec files. This is the key point of this pull request, let `rpm` do what it should do, check dependency by the version numbers.

GONG Jie

whowutwut:
@neo954 Thanks for that, I understand now.  I've also seen in the past that depending on the time stamp during the build, that is what the rpms has in it's name.

neo954:
@whowutwut, @zet809 and @immarvin, If you have time, please review the patch.

whowutwut:
Accidentally removed the other reviewers.  I put them back.  

I didn't review the code changes too closely but I did check out the changes to make sure the result is what we are expecting.  On my linux machine I ran a large extract process to slow down the build to span multiple minutes.   

```
[vhu@victorvm7 xcat-core]$ time ./buildcore.sh 
Building /home/vhu/rpmbuild/RPMS/noarch/perl-xCAT-2.13.1-snap*.noarch.rpm ...
Building /home/vhu/rpmbuild/RPMS/noarch/xCAT-client-2.13.1-snap*.noarch.rpm ...
Building /home/vhu/rpmbuild/RPMS/noarch/xCAT-server-2.13.1-snap*.noarch.rpm ...
...
Creating /home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/core-rpms-snap.tar.bz2 ...

real	2m56.877s
user	0m21.288s
sys	0m10.422s
```

The resulting RPMs contain the same snap date, which is what we are expecting.  
```
[vhu@victorvm7 xcat-core]$ ls -1 /home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/*.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/perl-xCAT-2.13.1-snap201701031526.noarch.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-2.13.1-snap201701031526.ppc64le.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-2.13.1-snap201701031526.ppc64.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-2.13.1-snap201701031526.s390x.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-2.13.1-snap201701031526.x86_64.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-buildkit-2.13.1-snap201701031526.noarch.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-client-2.13.1-snap201701031526.noarch.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-confluent-2.13.1-snap201701031526.noarch.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-csm-2.13.1-snap201701031526.noarch.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-genesis-scripts-ppc64-2.13.1-snap201701031526.noarch.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-genesis-scripts-x86_64-2.13.1-snap201701031526.noarch.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-probe-2.13.1-snap201701031526.noarch.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-server-2.13.1-snap201701031526.noarch.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCATsn-2.13.1-snap201701031526.ppc64le.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCATsn-2.13.1-snap201701031526.ppc64.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCATsn-2.13.1-snap201701031526.s390x.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCATsn-2.13.1-snap201701031526.x86_64.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-SoftLayer-2.13.1-snap201701031526.noarch.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-test-2.13.1-snap201701031526.noarch.rpm
/home/vhu/xcatbuild/neo954-issue1547-rpm-packaging/xcat-core/xCAT-vlan-2.13.1-snap201701031526.noarch.rpm
```
I'm OK with these changes.

daniceexi:
The change looks good for me too.

@zet809 and @immarvin please review and merge the change ASAP.

buildcore.sh
```diff
@@ -490,18 +492,18 @@ EOF
 #!/bin/sh
 cd `dirname $0`
 REPOFILE=`basename xCAT-*.repo`
-if [[ $REPOFILE == "xCAT-*.repo" ]]; then 
+if [ $REPOFILE = "xCAT-*.repo" ]; then
```
immarvin: should this be ``[[ ]]``?
````
[root@c910f03c05k21 xcat-core]# aa="xcat-mm.repo"
[root@c910f03c05k21 xcat-core]# echo $aa
xcat-mm.repo
[root@c910f03c05k21 xcat-core]# if [ $aa = "xcat-*.repo" ];then echo pp; fi
[root@c910f03c05k21 xcat-core]# if [[ $aa = xcat-*.repo ]];then echo pp; fi
pp
````

File: xCATsn/xCATsn.spec [hunk 2]
```diff
@@ -13,11 +13,15 @@ BuildRoot: /var/tmp/%{name}-%{version}-%{release}-root
...
 Source1: xcat.conf
 Source2: license.tar.gz
-Source3: xCATSN 
+Source3: xCATSN
 Source5: templates.tar.gz
 Source6: xcat.conf.apach24
-Provides: xCATsn = %{version}
-Requires: xCAT-server xCAT-client perl-DBD-SQLite xCAT-genesis-scripts-x86_64 xCAT-genesis-scripts-ppc64 xCAT-probe >= 2.12.2 
+Requires: perl-DBD-SQLite
+Requires: xCAT-client = 4:%{version}-%{release}
```
immarvin: ``4`` is ``%{Epoch}``? why is it hard coded?

xCATsn/xCATsn.spec
```diff
@@ -13,11 +13,15 @@ BuildRoot: /var/tmp/%{name}-%{version}-%{release}-root
...
+Requires: xCAT-server = 4:%{version}-%{release}
+Requires: xCAT-probe  = 4:%{version}-%{release}
+Requires: xCAT-genesis-scripts-x86_64 = 1:%{version}-%{release}
```
immarvin: should the Epoch of  xCAT-genesis-scripts be consistent with other xCAT-* packages?

immarvin:
hi @neo954 , you added a file "Release" including the build timestamp in the xcat-core code base,  I can see 2 possible problems:
1. unlike the "Version" file (which will be modified manually on each release), this "Release" file will be modified on each build  with ``build core.sh`,  it has to be pushed back to the code base each time, otherwise, a code diff in file "Release" will always be left in the building branch. Request merging "Release" on each build is not feasible,  considering the local build performed by xCAT contributors, should he/she push the "Release" to upstream?

2. sometimes,  the contributor may run "makerpm" to build individual package, he/she has to write the timestamp manually. With this PR merged, individual packages built with "makerpm" might not be installed/upgraded due to the strict dependency, should we prohibit running "makerpm" with a message? 
 
 I think the timestamp should better be passed in as a variable instead of a file in code base, just like what we do for "pcm","notpcm" or "fsm" in "xCAT-server/xCAT-server.spec".

xCATsn/xCATsn.spec
neo954: The %{Epoch} is "4". This is the historical code.

xCATsn/xCATsn.spec
neo954: The %{Epoch} of xCAT-genesis-scripts-x86_64 is "1". This is historical code.

neo954:
@immarvin, 
1. The file `Release` will not be committed to the git repo, unless user did `git add Release`. And it is obviously improper, and should be avoid.
2. Yep, individually build package will have the problem you mentioned and may need to be installed with `rpm -i --nodeps`, if the user did not update file `Release` and did not build all the packages. But developer should know what he/she is doing, and know what happened behind the screen.

buildcore.sh
neo954: That is my mistake, pattern matching is required here. I will change it back.

buildcore.sh
```diff
@@ -490,18 +492,18 @@ EOF
 #!/bin/sh
 cd `dirname $0`
 REPOFILE=`basename xCAT-*.repo`
-if [[ $REPOFILE == "xCAT-*.repo" ]]; then 
+if [[ $REPOFILE == "xCAT-*.repo" ]]; then
```
immarvin: i do not think this will work
should be ``if [[ $REPOFILE = xcat-*.repo ]]; then``

xCATsn/xCATsn.spec
immarvin: i suggest to modify 4 to ``%{Epoch}``

immarvin:
hi @neo954 , thanks for your expl.
for 1. a new file "Release" is indeed in your PR, I insist  not to leave the file "Release" in xcat-core on each build, it will always leave a untracked file or unmerged file in xcat-core, which is quite confusing

buildcore.sh
```diff
@@ -402,48 +403,49 @@ fi
...
 # get gpg keys in place
 if [ "$OSNAME" != "AIX" ]; then
     if [ -z "$RPMSIGN" -o "$RPMSIGN" == "1" ]; then
-    mkdir -p $HOME/.gnupg
-    for i in pubring.gpg secring.gpg trustdb.gpg; do
-        if [ ! -f $HOME/.gnupg/$i ] || [ `wc -c $HOME/.gnupg/$i|cut -f 1 -d' '` == 0 ]; then
-            rm -f $HOME/.gnupg/$i
-            ${WGET_CMD} -P $HOME/.gnupg $GSA/keys/$i
-            chmod 600 $HOME/.gnupg/$i
+        mkdir -p $HOME/.gnupg
```
whowutwut: Thank you for formatting the code and cleaning up the file! :+1:   In the future,   If doing cleanup, it might be good to put the clean up into a single commit, then the changes in other commits, this will help make it easier for the reviewers to see code changes vs formatting changes...

buildcore.sh
whowutwut: This change may break as we figure out what is missing.  The `go-xcat` script started creating lower case repo files.  There was some discussion about this before and I think we should sync up the .repo files to be consistent.  If we want to change to all lowercase, this would be the time.. but we will probably find places where it needs to be corrected manually.  One place is in the xcat-deps location on GSA.   I'd like to NOT re-create a xcat-deps package until we get closer to the GA time to NOT break existing CS16A customers using the public xcat-dep repo.

perl-xCAT/perl-xCAT.spec
```diff
@@ -1,7 +1,7 @@
 Summary: xCAT perl libraries
 Name: perl-xCAT
 Version: %{?version:%{version}}%{!?version:%(cat Version)}
-Release: %{?release:%{release}}%{!?release:snap%(date +"%Y%m%d%H%M")}
+Release: %{?release:%{release}}%{!?release:%(cat Release)}
```
whowutwut: Does this still work with Jarrod's change for Lenovo builds that stopped using the `snapXXXXXX` format?  We don't want to break this.

xCAT-buildkit/xCAT-buildkit.spec
```diff
@@ -20,15 +20,13 @@ AutoReqProv: no
 # also need to fix Requires for AIX
 %ifos linux
 BuildArch: noarch
-#Requires: 
+#Requires:
```
whowutwut: Let's delete this line since it's unused.

xCAT-buildkit/xCAT-buildkit.spec
```diff
@@ -20,15 +20,13 @@ AutoReqProv: no
...
 %endif
 
 # No dependencies on any other xCAT rpms
 # so that this rpm can be installed in a separate build server
 Requires: /usr/bin/rpmbuild
 
-Provides: xCAT-buildkit = %{epoch}:%{version}
```
whowutwut: Is it safe to remove this? is the xCAT Kits spec file looking for this?  @cxhong  @bybai   comments?

xCAT-genesis-scripts/xCAT-genesis-scripts.spec
```diff
@@ -13,23 +13,21 @@ BuildArch: noarch
...
 AutoProv: false
 Obsoletes: xCAT-genesis-%{tarch}
 Provides: xCAT-genesis-%{tarch}
 
-
-
 Name:	 %{name}
 Group: System/Utilities
 License: EPL
 Vendor: IBM Corp
```
whowutwut: If you want to add a . after Corp.   A little picky but since we touched this file... :-)

xCATsn/xCATsn.spec
```diff
@@ -38,7 +42,7 @@ Requires: /usr/bin/ssh
 %ifnarch s390x
 Requires: /etc/xinetd.d/tftp
 # yaboot-xcat is pulled in so any MN can manage ppc nodes
-#Requires: yaboot-xcat 
+#Requires: yaboot-xcat
```
whowutwut: Suggest removed the unused line

Release
```diff
@@ -0,0 +1 @@
+snap000000000000
```
neo954: This file is necessary. If the user intend to do `rpmbuild` for a individual package manually, he/she will need this file. At least, there is a default minimal value in the file `Release`.

immarvin: how about add a logic in spec, if this file does not exist, use the ``date +"%Y%m%d%H%M"`` ? it is not reasonable to build based on a timestamp in the outdated ``Release`` file, requesting user to generate a ``Release`` file manually before ``rpmbuild`` is not reasonable either.

another problem is, with this PR merged, is building individual package with``rpmbuild`` still make sense? since the dependency is so strict, we have to build all of its dependency packages with the same timestamp, in this case, building all the packages with "buildcore.sh" is the right choice.

buildcore.sh
immarvin: hi @whowutwut , lower case "xcat-" is my typo, the problem is that double quotes ``"``  can not be used in pattern matching inside ``[[]]``here

immarvin:
I am ok with the change and have little concern about the unclean git directory after "buildcore.sh", maybe we can mention this in Doc http://xcat-docs.readthedocs.io/en/stable/developers/guides/code/builds.html afterwords.

I agree to merge this so that we can start verification ASAP.

Release
whowutwut: @neo954  Then we can fix `makerpm` to generate the file.  

In my sandbox for your code changes, I create the `.gitignore` and remove the Release file

```
[vhu@victorvm7 xcat-core]$ git checkout neo954-issue1547-rpm-packaging
M	.gitignore
[vhu@victorvm7 xcat-core]$ rm Release 
```

The `makerpm` fails as expected....
```
[vhu@victorvm7 xcat-core]$ ./makerpm xCAT-vlan
cat: Release: No such file or directory
```

Modify `makerpm` to create a new Release number each time...
```
[vhu@victorvm7 xcat-core]$ git diff makerpm
diff --git a/makerpm b/makerpm
index c80ed0e..01a5f0f 100755
--- a/makerpm
+++ b/makerpm
@@ -232,6 +232,9 @@ else
        QUIET="--quiet"
 fi
 
+XCAT_RELEASE="snap$(date '+%Y%m%d%H%M')"
+echo "$XCAT_RELEASE" > Release
+
 if [ "$OSNAME" = "AIX" ]; then
        RPMROOT=/opt/freeware/src/packages
 else   # linux
```

Run again:

```
[vhu@victorvm7 xcat-core]$ ./makerpm xCAT-vlan
Building /home/vhu/rpmbuild/RPMS/noarch/xCAT-vlan-2.13.1-snap*.noarch.rpm ...
```

This seems to be in line with what we want.  When we run buildcore , the release date is set to something consistent.  If the user builds manually, we want to have a version that changes with each build so that rpm -U still works....

buildcore.sh
whowutwut: OK thanks

perl-xCAT/perl-xCAT.spec
whowutwut: I validated that this seems to still work if we set XCATVER in the `makerpm` file.

xCAT-IBMhpc/xCAT-IBMhpc.spec
```diff
@@ -20,7 +20,7 @@ AutoReqProv: no
 # also need to fix Requires for AIX
 %ifos linux
 BuildArch: noarch
-#Requires: 
+#Requires:
```
whowutwut: This is unused, can we delete this. 

I did a quick search, since you did some clean up, can you remove all of the following commented out Requires lines. 
```
[vhu@victorvm7 xcat-core]$ git grep "#Requires:" 
perl-xCAT/perl-xCAT.spec:#Requires: perl-SOAP-Lite
xCAT-IBMhpc/xCAT-IBMhpc.spec:#Requires:
xCAT-buildkit/share/xcat/kits/kit_template/source_packages/sample/pkg2/pkg2.spec:#Requires:
xCAT-buildkit/share/xcat/kits/kit_template/source_packages/sample/pkg3/pkg3.spec:#Requires:
xCAT-buildkit/xCAT-buildkit.spec:#Requires:
xCAT/xCAT.spec:#Requires: yaboot-xcat
xCATsn/xCATsn.spec:#Requires: yaboot-xcat
```

whowutwut:
> I am ok with the change and have little concern about the unclean git directory after "buildcore.sh", maybe we can mention this in Doc http://xcat-docs.readthedocs.io/en/stable/developers/guides/code/builds.html afterwords.

@immarvin  I think the `.gitignore` proposed changes should leave the directory clean here, right?  any other files you are seeing?

Release
neo954: @whowutwut, But in the way of not ship `Release` file in the git repository, the end user cannot make a single rpm manually. When I say manually, I mean the end user run `rpmbuild` against the spec file directly. And this is not what I want. I do not want to limit the end user on what he or she can do. The end user can do whatever he or she want. Checkout the code from git repository, run `rpmbuild` directly after that. Not rely on an xCAT shipped shell script to create a `Release` file for him/her. I mean, maintain the basic procedure of hand made packaging, as the old fashion way.

xCATsn/xCATsn.spec
neo954: @immarvin, I do not second your motion. Since the `%{epoch}` will not change from time to time. And additional, we do not have a identical `%{epoch}` number for each xCAT rpm package.

Actually, the only chance we need to change `%{epoch}`, is when we abandon the xCAT version number scheme we currently used. And it is not going to happen very soon. When it really happen, at that time, we may have to largely rewrite all the spec files. In that case, `%{epoch}` may change to `5`.

Thus, the hard coded `4` here is not a big deal. Take it easy.

neo954: @immarvin, I do not second your motion. Since the `%{epoch}` will not change from time to time. And additional, we do not have a identical `%{epoch}` number for each xCAT rpm package.

Actually, the only chance we need to change `%{epoch}`, is when we abandon the xCAT version number scheme we currently used. And it is not going to happen very soon. When it really happen, at that time, we may have to largely rewrite all the spec files. In that case, `%{epoch}` may change to `5`.

Thus, the hard coded `4` here is not a big deal. Take it easy.

Release
whowutwut: @neo954  I understand your point and I do agree with your intention to allow `rpmbuild` to continue to work.  But it doesn't work once we add this Release file.  This is because the generation of the snap date is not in the spec file like @immarvin  said... 

So I restored the Release file in my sandbox and created xCAT-client rpm, result is this: 
```
xCAT-client-2.13.1-snap000000000000.noarch.rpm
```

So now with that numbering, I cannot update the rpm: 

```
[root@fs2vm100 vhu]# rpm -Uhv xCAT-client-2.13.1-snap000000000000.noarch.rpm 
Preparing...                          ################################# [100%]
	package xCAT-client-4:2.13.1-snap201612270615.noarch (which is newer than xCAT-client-4:2.13.1-snap000000000000.noarch) is already installed
```

So to install, I have to first remove it... and I cannot remove the xCAT-client because of the dependencies:
```
[root@fs2vm100 vhu]# rpm -e xCAT-client
error: Failed dependencies:
	xCAT-client >= 4:2.13.1 is needed by (installed) xCAT-server-4:2.13.1-snap201612270615.noarch
	xCAT-client is needed by (installed) xCAT-2.13.1-snap201612270616.ppc64le
```

So the use case you describe doesn't work anymore.  


I still think we need to
* add the generation of the snap date into `makerpm` 
* remove the Release file in this commit
* add entry into `.gitignore` 

We can look into other solutions in the future to allow `rpmbuild` to operate again.

neo954: @whowutwut, After the end user hand made an individual package, and try to run `rpm -U` against it, for a single rpm package updating, `rpm` will prevent it. As you mentioned in your previous comment. This is worked as designed.

The basic idea of this pull request is, in a working xCAT environment, require all the packages installed from xcat-core to have the exactly same version number. Otherwise `rpm` will complain a dependencies error. Thus, your attempt of trying to install a single hand made rpm among with all the other existing xcat-core packages is definitely against the rule. The design goal of this pull request is prevent such an attempt, not allow it.

If the end user do know what he or she is doing and understand what is happening behind the screen and all the side effects, he or she can do such thing with `rpm -i --nodeps` or `rpm -U --nodeps`. The `--nodeps` command line argument of `rpm` indicate the end user knows he or she will suffer rpm dependency hell and is willing to solve it manually. And after that, the eye-catching `snap000000000000` will become a reminder, indicate the end user there is something abnormal.

One-step back, if the end user hand made every package in `xcat-core`, then, all of them will have `snap000000000000` part in their version number. In that case, everything will be fine.

whowutwut: Hi @neo954, I just don't see that as a valid use case for xCAT.  This means that in order to install their own version of xCAT, they need to do the following:
1. Uninstall xCAT 2.13.0 (snap20151215....)
2. run `rpmbuild` on all the individual spec files and move the rpms to the mgmt node.
3. Install their version of xCAT with `snap0000000000`.  If using `yum` they need to create repo with the individual rpms... 

At this point, the user would probably create a different version completely to differentiate their forked copy with our version of xCAT. 

> And after that, the eye-catching snap000000000000 will become a reminder, indicate the end user there is something abnormal.

I don't think it should be abnormal, to fork a copy and make changes in open source.  If this is done, we should not treat it as "abnormal" 

The problem still exists that we have to leave the git repo unclean each time we run `buildcore.sh`.  This is the purpose of `.gitignore`.  To tell git to not track files that are generated on the fly. https://git-scm.com/docs/gitignore

Do you agree that we should at least generate the `Releases` file in the `makerpm` command?  Regardless of if we remove this file, we still need to make that change to allow makerpm to continue to create rpms that increment in snap date.

immarvin:
hi @whowutwut , ``. gitignore `` is ok for "Release" file in this PR. but in the PR https://github.com/xcat2/xcat-core/pull/2198 , the debian/control files will be modified: 
````
+                        sed -i -e "s#>= 2.13-snap000000000000#= ${pkg_version}#g" ${CURDIR}/debian/control

````
I do not think the control files should be ignored

File: Release (hunk 11 above)
Comment (neo954) [inline Release]: @whowutwut, generate the `Release` file in the `makerpm` script is not correct. In that case, `makerpm` will regenerate `Release` for each xcat-core rpm package. Course each xcat-core rpm package have different rpm release number. And this is the case we want to avoid at first.

Comment (whowutwut) [inline Release]: @neo954  you're right..   `buildcore.sh` calls `makerpm`.  humm.. this is a problem I think...  

In my dev work, I use both the `buildcore.sh` and the `makerpm` all the time. . `buildcore.sh` if I want to create a fully new xCAT installable package and install a clean version of my changes... and `makerpm` when I change only a few files within the same package and want to test out the new rpm.  In both situations, I'm able to execute `rpm -U` or `yum update` command against the RPM because anything that is built has a number that is higher than a previously build rpm.   With this pull request, it breaks `makerpm`

So I see why you want to keep the Release file checked in, because `makerpm` will not work otherwise.  But I think this whole change breaks the usefulness of `makerpm`.   

Here's where this solution and also Jarrods has it's issues...   
* This one will freeze the build snap number to the last time that the `Release` file is created by `buildcore.sh`  If it's not run, `makerpm` will fail ( because I still think we should NOT ship `Release` file because it's a generated file...
* Jarrod's solution using the commit ID also doesn't work in a development  and UT environment because I need to commit my changes in my branch in the git repo before the release number gets incremented. (if using rpm -U)

So now i'm not sure if we should even do this... let's talk  about this on Mondays call..  can we get input from the team members about whether they favor this PR if it breaks `makerpm`?  In some past products, we didn't have this problem because we could `rpm -e` to uninstall everything  and then install the same version again cleanly... .here's it's not that easy for xCAT...

**Comment by daniceexi** [comment]
My ideas about this change, just for reference.

* Make all the rpms have same release number and make each rpm has the accurate version for required dep rpms are good. It could skip the possible mismatch of installed xCAT rpms.
* Generally we uses the xcatbuild.sh, so satisfy it should be the priority.
* If someone wants to use 'makerpm', then he/she needs to install the rpm through 'rpm --nodeps'

**Comment by zet809** [comment]
My idea is:
For the spec file, it use 'Release' file first, if not exist, it use 'snap'+date.
For the 'Release' file, both buildcore.sh and makerpm can generate it. The difference is:
    Within makerpm, if it detect there is no 'Release' file exist, it need to generate one and remove it before it return.
For @jjohnson42 's concern, he used "--define 'release 1'" for the rpmbuild. To match my idea, the '1' can be write into the 'Release' file.
What is your idea?

File: Release (hunk 11 above)
Comment (neo954) [inline Release]: @whowutwut, If we are going to switch the xCAT version number scheme to Jarrod's solution. That will be totally a new story. That is not what we are going to discuss in this pull request. The prerequisite of this pull request is we still maintain current xCAT version number scheme.

I can see the point of not shipping file `Release` in the git repository. But at the moment, I do not see a solution which can make everything right without the `Release` file. If you have any, please nominate one.

**Comment by neo954** [comment]
A new commit is made. The original copy of file `Release` will be saved in script `buildcore.sh`, and it will be restored after the build process. See `buildcore.sh` line 193 through 212.

    internal_backup
    trap internal_cleanup 0

**Comment by whowutwut** [comment]
@neo954, As we discussed on the phone....  I think the following changes would solve all the cases. 

1. Release file stays as `snap0000000`...
2. `buildcore.sh` executes and saves it, create new, then at the end restores it....  (You just made this change) 
3. `makerpm` changes (modified a little from @zet809 suggestion on the phone)
    a) check if  `snap0000000...` is in the Release file.... if NO, then use the content...  if YES, then back up, create new one, and restore after `makerpm` is executed. 


Then `rpmbuild` stills works and should generate a `snap0000` release file name. 
`buildcore.sh` and `makerpm` will both work and the behavior is preserved, at the end of executing all these options, the git repo is clean.

File: buildcore.sh [hunk 13]
```diff
@@ -190,6 +190,27 @@ function setversionvars {
     echo "$XCAT_RELEASE" >Release
 }
 
+function internal_backup()
+{
+    # Create a backup for file `Release'
+    if [ ! -f Release.save.998 ]
```
whowutwut: Since we are referencing this release file in multiple places, maybe should make it into a variable. 
```
RELEASE_FILE="Release"
RELEASE_FILE_SAVE=${RELEASE_FILE}.tmp
```
Not sure why we use `save.998`, maybe just `.tmp` since it will show the user that it's a temporary file.

jjohnson42:
How about my proposol:
-Move away from snap entirely, the datecode is too ambiguous anyway
-Avoid having any 'snap' like output for tagged versions
-Use git rather than a 'Version' or 'Release' file to determine state
-When using a non-tagged version, use 'git describe --tags' as a basis for the version/release.
    -Use convention like 'POST<number>' for the 'number of commits since release, so that tags with lowercase names would supersede others.

This is what I'm doing with the Lenovo builds, and have been trying to have as part of buildcore behavior when USEGITVER=1.

jjohnson42:
For reference, https://github.com/jjohnson42/xcat-core/tree/lenovobuild2.13 reperesnts USEGITVER=1 and how it automatically decides between a 'POST' build and a 'release' build and largely ignores Version or similar files when making rpm versions.

whowutwut:
Thanks @jjohnson42  I agree with you.   This PR is an attempt to first sync up the `snapXXXXXXXXXXXXXX` numbers to be the same within a single release of xCAT.  

I did look at the changes you had made a with the buildsanity branch and using the `git describe` command and I wanted to implement those changes back then but hit a wall on this issue:  On the master branch, I couldn't get the version to bump up.... 
```
[vhu@victorvm7 xcat-core]$ git checkout master
Switched to branch 'master'
[vhu@victorvm7 xcat-core]$ git describe --tags
2.12.0-1531-g3bdf3b6
```

Today, I spent some time to figure out why... turns out that the expected behavior does happen when we switch to the release branches, but not on the master.  The tags are on the release branches, and NOT on the master branch.  The last tag we had is 2.12.0... (so it's actually correct)

```
[vhu@victorvm7 xcat-core]$ git checkout 2.13
Switched to branch '2.13'
g[vhu@victorvm7 xcat-core]$ git describe --tags
2.13.0-6-g372efba
```

So it looks like before we can make this kind of change we need to re-think the way we are releasing, perhaps we should release on the master branch and have a development branch that merges back into master... instead of what we are doing today.

xCAT-buildkit/xCAT-buildkit.spec
bybai: Hi @whowutwut ,
buildkit testcases are running daily, we can have a try to remove this, and autotest can test to see if there is problem.

buildcore.sh
neo954: Okay, as you wish, these variables were added.

neo954:
@whowutwut, It is great @jjoknson42 put his comments here. I would like to change a new xCAT version number scheme as well. But that will pull us into another endless argument about what version number scheme should be used. Since we did not make any decision yet. Let us just move on.

The basic idea of this pull request is solve the dependency problem xCAT met, but keep the changes minimal. we already spent more than six weeks on discussing this pull request. During the discussion, I believe many of us did not think through the problem we met aloud. At this stage, I would like to keep this pull request at its current shape. And expect it can be merged as soon as possible. In this way, we can have a correct rpm dependencies in xCAT 2.13.1.

For the new xCAT version number scheme. Let us discuss it in a different place. It may cost another six weeks or more to put everyone of us on the same page.

buildcore.sh
```diff
@@ -186,10 +186,36 @@ function setversionvars {
     BUILD_TIME=`date`
     BUILD_MACHINE=`hostname`
     COMMIT_ID=`git rev-parse --short HEAD`
+    XCAT_RELEASE="snap$(date '+%Y%m%d%H%M')"
+    echo "$XCAT_RELEASE" >Release
```
immarvin: should this line use ``${RELEASE_FILE}``? @neo954

zet809:
@whowutwut and @jjohnson42 I think we can merge this pull request. For the package name changing topic, we can open another defect to trace. Thx!

buildcore.sh
```diff
@@ -190,6 +190,31 @@ function setversionvars {
...
+{
+    # Create a backup for file `Release'
+    if [ ! -f "${RELEASE_FILE_SAVE}" ]
+    then
+        mv "${RELEASE_FILE}" "${RELEASE_FILE_SAVE}"
+        cp "${RELEASE_FILE_SAVE}" "${RELEASE_FILE}"
+    fi
+}
+
+function internal_cleanup()
```
whowutwut: Doing some UT on this today, this code does not work because the PWD is not correct to find the `Release.save.998` file. .. will look into creating a PR to fix this...

neo954: Okay, thanks @whowutwut.

jjohnson42:
Let me know about opening up for version change.

It may make sense to tag master to mark the beginning of a new major, e.g. tag it '2.14', and tag releases always with three digits (2.14.0).  Of course then we'd probably want to change the branch naming strategy so that it doesn't conflict, but we already have that fun in play.  The script could put a 'PRE' when the version only has one dot, and a 'POST' when it has two dots.  Sorry to clutter this PR more, just wanted to jot down a few thoughts.

The only non-obvious thing I think is using uppercase for dev build tag, to stick with the 'oldest' class of characters by rpm reckoning in case we want a text string in some build family in the future.

Issue #2104, opened 2016-11-08
[CentOS6.8 and RHEL6.8] User will see ERROR message during running updatenode and in xcat.log on compute node.


SStar1314:
On CentOS6.8 and RHEL6.8, user will see ERROR messages when running updatnode command. See below detail info:
[root@ip181mn ~]# updatenode pcmstdcn0
File synchronization has completed for nodes.
Performing software maintenance operations. This could take a while, if there are packages to install.
....
pcmstdcn0:
pcmstdcn0: xcat: **Error: Unable to read private ECDSA key from /etc/xcat/hostkeys**    ----------------------------------------------  See ERROR here.
...

My investigation for this bug:
1. During installation, it will run "xcatconfig -m" command, this command will execute "genSSHNodeHostKey" function(line 910).
2. "genSSHNodeHostKey" function need to **check if file "/etc/ssh/ssh_host_ecdsa_key" is exist**, if true, then generate file "/etc/xcat/hostkeys/ssh_host_ecdsa_key" file and copy those files  to "/install/postscripts/hostkeys".
3. "updatenode" command will make compute node to execute "remoteshell" script to get ssh key files from MN, but in "remoteshell" script, **it will check if cn support "ecdsa" by executing "ssh-keygen -t ecdsa -f /tmp/ssh_ecdsa_hostkey -P "" &>/dev/null" command**, if true, then cn will try to get ecdsa key file from mn.
4. When **MN don't exist** "/etc/ssh/ssh_host_ecdsa_key" file, so it will not generate "/etc/xcat/hostkeys/ssh_host_ecdsa_key" file. While **cn support ecdsa**, it will want to get ecdsa key file from mn. So error happens.

Seems on centos/rhel6.8, it will not generate ecdsa/rsa/dsa key by default.
[root@ip181mn ~]# cat /etc/ssh/sshd_config | grep HostKey
#HostKey for protocol version 1
#HostKey /etc/ssh/ssh_host_key
#HostKeys for protocol version 2
#HostKey /etc/ssh/ssh_host_rsa_key
#HostKey /etc/ssh/ssh_host_dsa_key
[root@ip181mn ~]#

I think the handle method of ecdsa key in remoteshell or xcatconfig script is not consistent.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2196

SStar1314:
Close this issue for now, need to verify this on xCAT 2.13.

PR #2196, opened 2016-11-24
fix issue [CentOS6.8 and RHEL6.8] User will see ERROR message during running updatenode and in xcat.log on compute node. #2104

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2104
on centos6.8, the opens-server support "ecdsa", but there is no "/etc/ssh/ssh_host_ecdsa_key" created on ssh server start by default. correct the logic to determine the ecdsa support in xcatconfig with the same logic in "remoteshell"


daniceexi:

Issue #1548, opened 2016-07-18
Please update the control file of package xcat and xcatsn to have proper package dependency


neo954:
Since xCAT requires all the xCAT packages are on the same version level. xCAT should put all these requirement in its RPM spec file.

A proper fix of this bug should allow the end user to upgrade xCAT with a simple command.

```
apt-get install xcat
```

neo954:
Basically, this is the same bug as #1547 but for Ubuntu.

neo954:
The bug is partly covered by the script ``go-xcat``. But I think even though we have ``go-xcat``, this is a bug of xCAT itself, and should be fixed properly.

zet809:
Will deal with this issue in 2.12.3.

zet809:
The pull request #1659 and #1385 can fix this issue.

neo954:
I reviewed pull request #1659 and #1385. I do not see any line related to this issue.

neo954:
Verified with xCAT 2.12.3 GA version. And failed. This problem still exist.

zet809:
Need to add version number for depended packages.

zet809:
Move to next release to do more investigation.

neo954:
Since @zet809 did not spend time on this issue. I tried to fix this issue by myself.

tingtli:
From the end user's perspective, i verified this bug using apt-get install xcat.  It  could installed all xCAT packages. I think we can close this one
```
root@c910f03c17k29:~# dpkg -l |grep -i xcat
ii  conserver-xcat                     8.1.16-3                         ppc64el      Serial console server daemon/client
ii  elilo-xcat                         3.14-4                           all          Open source implementation of Service Location Protocol V2
ii  grub2-xcat                         2.02-0.16.el7.snap201506090204   all          grub2 resources generated by grub2-mknetdir
ii  ipmitool-xcat                      1.8.17-1                         ppc64el      utility for IPMI control with kernel driver or LAN interface
ii  perl-xcat                          2.13.2-snap201701170616          all          xCAT perl libraries
ii  syslinux-xcat                      3.86-2                           all          SYSLINUX modules in /opt/xcat/share/xcat/netboot/syslinux, available for network booting
ii  xcat                               2.13.2-snap201701170616          ppc64el      Metapackage for a common, default xCAT setup
ii  xcat-buildkit                      2.13.2-snap201701170616          all          Provides build tools and sample files to build an xCAT kit
[...]

PR #2198, opened 2016-11-24
Make all the deb packages have the exactly same version and release numbers

neo954:
This pull request try to fix issue #1548.


neo954:
In these patches, I heavily update the packaging of xCAT Debian packages, to make them contemporary.

whowutwut:
@neo954  Can you provide the result of this change in regards to how it changes the package names for Debian?  Since this seems like a drastic naming convention change, I moved this to the next major milestone.   Thoughts?

neo954:
I solved the code conflict manually. Now, the patch is clean.

neo954:
@whowutwut, This patch is basically the same thing as pull request #2195, but for Debian/Ubuntu packaging.

neo954:
@whowutwut, @zet809 and @immarvin, Now, xCAT 2.13 is released. If you have time, please review the patch.

xCAT-IBMhpc/debian/control
```diff
@@ -2,14 +2,14 @@ Source: xcat-ibmhpc
...
+Standards-Version: 3.9.4
 
 Package: xcat-ibmhpc
 Architecture: all
-Depends: ${perl:Depends}, perl-xcat (>= 2.8.0), xcat-client (>= 2.8.0)
-Description: Install and configuration utilities for IBM HPC products in an xCAT cluster
- xCAT-IBMhpc provides sample installation and configuration scripts for running 
- the IBM HPC software stack in an xCAT cluser.  Support for the following IBM 
- products is provided:  GPFS, LoadLeveler, Parallel Environment, ESSL and Parallel 
+Depends: ${perl:Depends}, perl-xcat (>= 2.13-snap000000000000), xcat-client (>= 2.13-snap000000000000)
```
whowutwut: Not familiar with what this does, can you explain...  does it just ignore any value in "0000..."?

neo954: Nope. Since Debian build system does not support including files or executing external command in the `control` file. Thus, `(>= 2.13-snap000000000000)` is just hard coded here.

If the end user make a manual build for an individual package, it will leave there. And work as is.

If package building is run with our build script, the `(>= 2.13-snap000000000000)` part in the `control` file will be replaced to the actual version number and time tag in the build script, that is something like `(= 2.13-snap201701051325)`.

This is the explain of how it works.

neo954: See line 251 of script `build-ubunturepo`.

immarvin:
I am ok with the change and have little concern about the unclean git directory after "buildcore.sh", maybe we can mention this in Doc http://xcat-docs.readthedocs.io/en/stable/developers/guides/code/builds.html afterwords.

I agree to merge this so that we can start verification ASAP.

whowutwut:
@immarvin  yes, I agree with you, I think we should really utilize the `gitignore` file if we do have things we do not want to commit back to the repo so that the git directory is clean.

zet809:
@neo954 Looks good to me, I agree to merge. Thx very much!

Issue #1849, opened 2016-09-16
rflash does not display any output or progress


markprez:
```
Version 2.12.2 (git commit 6fff020a905c64cabe31896e653aebc942ee0694, built Thu Sep  1 06:18:21 EDT 2016)
[root@c712ems4 ppc64le]# rpm -qa | grep xcat | grep ipmi
ipmitool-xcat-1.8.17-0.ppc64le
[root@c712ems4 ppc64le]# ipmitool-xcat -V
ipmitool-xcat version 1.8.17
```

rflash against Garrison node does not display any output or indicate progress. 
Also, I am not sure if the rflash is will continue if the session is disconnected. If not, it would be good to have this. On some earlier hardware, rflash would write to the screen and  a log (/var/log/xcat/rflash/<hostname> or something similar) and continue if the console disconnected.  Status could then be obtained from the log when reconnected.
Would like some output to the screen indicating progress and a log file.

whowutwut:
@pdlun92  Please work with @zet809  to see if there's a good way to provide some status during the update of the firmware.

pdlun92:
Moving to milestone 2.12.4

Thinking this will have to be added as an option to `rflash` as default behavior could become very messy/ineffective with scale up

pdlun92:
Talked to @markprez about this today.  I think the end goal here is to make sure that rflash will continue running even through terminal disconnects and other similar interruptions, as if it was running through a `screen` session, allowing admin/user to issue command and check back in on progress as they please

chenglch:
I think running command in `screen` is acceptable. Do we need to record more information like `verbose` debug message from the ipmitool command?  The original output of `ipmitool hpm update` is tty-based message and not accurate, it's hard to make use of this result.

markprez:
I usually run: (nohup rflash  Pf2,Pf3 8335GTB_820.1634.20160917a.hpm 2>&1 | tee -a 1634i.install.f1 & ) which will keep running if my connection drops... but if I forget screen or to start it in the background with nohup I am stuck at the terminal until it completes. Firmware flash is a dangerous process to have exit abnormally. On a large system, this will cause undo inconvenience. Once a system is stable, flashing will most likely be infrequent. You are asking an admin to remember to run it from screen or have the confidence to nohup a process they do not often do.
In my latest run of 36 nodes, 3 nodes failed, without the tee of the output, this may have scrolled by:
nohup: ignoring input
[...]

PR #2203, opened 2016-11-25
Redirect progress message to rflash log files when updating firmware

chenglch:
Previously rflash process will be terminated when the tty session is closed
which is a very dangerous operation  may bricked the BMC. This patch will
ignore the `int`, `term` and `hup` signal when rflash is running and redirect
the output of ipmitool command to `/var/log/xcat/rflash/<node.log>` which make
the progress trackable. In addition,  -V option is provided to print more
verbose message of ipmitool when `hpm upgrade` command is running.

This patch also update the node status in the `nodelist` table to record
the rflash status.

fix-issue: #1849


zet809:

Issue #2183, opened 2016-11-22
[fvt]2.13 mypostscript syntax error in subroutine run_ps  on sles11.4


junxiawang:
Env:sles11.4
c910f02c01p14:~ # uname -a
Linux c910f02c01p14 3.0.101-63-ppc64 #1 SMP Tue Jun 23 16:02:31 UTC 2015 (4b89d0c) ppc64 ppc64 ppc64 GNU/Linux

Build:c910f02c01p14:~ # lsxcatd -v
Version 2.13 (git commit 6653dfdc6acc292cf7ccd8e053263009a6ea960a, built Mon Nov 21 06:15:37 EST 2016)

How to reproduce:
```
1.Run updatenode node confignics
   [updatenode c910f02c02p28 -P confignics] Running Time:3 sec
RETURN: rc = 0
OUTPUT:
c910f02c02p28: xcatdsklspost: downloaded postscripts successfully
c910f02c02p28: //xcatpost/mypostscript: line 27: syntax error near unexpected token `('
c910f02c02p28: //xcatpost/mypostscript: line 27: `        bash -x ./$@ 2>&1 | tee -a $logfile | tee >(logger -t xcat -p debug)'
c910f02c02p28: Running of postscripts has completed.
CHECK:rc == 0   [Pass]

2.In   mypostscript code 27:
 20  if [ -f $1 ]; then
 21   echo "`date` Running postscript: $*"
 22   msgutil_r "$MASTER_IP" "info" "`date` Running postscript: $*" "$logfile"
 23   if [ "$XCATDEBUGMODE" = "1" ] || [ "$XCATDEBUGMODE" = "2" ]; then
 24      local compt=$(file $1)
 25      local reg="shell script"
 26      if [[ "$compt" =~ $reg ]]; then
 27         bash -x ./$@ 2>&1
 28         ret_local=$?
 29      else
 30         ./$@ 2>&1 | tee -a $logfile |  logger -t xcat -p debug
 31         ret_local=${PIPESTATUS[0]}
 32      fi

   
```

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2207

whowutwut:
@junxiawang  please verify and close out this issue asap.

PR #2207, opened 2016-11-28
fix issue [fvt]2.13 mypostscript syntax error in subroutine run_ps on sles11.4 #2183

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2183
add she-bang to mypostscript, so that it will be invoked with ``bash``, otherwise it will be invoked with "sh", and some bash specific features will not be supported.
````
~man bash
...
       If  bash is invoked with the name sh, it tries to mimic the startup behavior of historical versions of sh as closely as
       possible, while conforming to the POSIX standard as well.  When invoked as an interactive login shell, or a  non-inter‐
       active  shell with the --login option, it first attempts to read and execute commands from /etc/profile and ~/.profile,
       in that order.  The --noprofile option may be used to inhibit this behavior.  When invoked as an interactive shell with
       the  name  sh,  bash looks for the variable ENV, expands its value if it is defined, and uses the expanded value as the
       name of a file to read and execute.  Since a shell invoked as sh does not attempt to read and execute commands from any
       other  startup  files,  the  --rcfile  option has no effect.  A non-interactive shell invoked with the name sh does not
       attempt to read any other startup files.  When invoked as sh, bash enters posix mode after the startup files are read.

       When bash is started in posix mode, as with the --posix command line option, it follows the POSIX standard for  startup
       files.  In this mode, interactive shells expand the ENV variable and commands are read and executed from the file whose
       name is the expanded value.  No other startup files are read.
,,,
````

### DIFF
--- a/xCAT/postscripts/xcatdsklspost
+++ b/xCAT/postscripts/xcatdsklspost
@@ -827,7 +827,7 @@ if [ $NODE_DEPLOYMENT -eq 1 ] || [ "$MODE" = "4" ] || [ "$MODE" = "6" ]; then
 else
     TMP=`sed "/# postscripts-start-here/,/# postscripts-end-here/ s/\(.*\)/run_ps postscript \1/;s/run_ps postscript\s*#/#/;s/run_ps postscript\s*$//" /$xcatpost/mypostscript`
 fi
-echo "
+echo "#!/bin/bash
 . /xcatpost/xcatlib.sh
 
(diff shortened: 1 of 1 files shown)


penguhyang:
Looks good to me. 👍

hu-weihua:
I have reviewed and agree to merge, thanks @immarvin .

Issue #2204, opened 2016-11-25
[fvt]diskfull provison node status  always booting


junxiawang:
Env:
ubuntu 14.04 ppc64le 
ubuntu 16.04 ppc64le
ubuntu 14.04.3 ppc64le
ubutu 16.04 x86-64
ubuntu 16.04 p8le
sles11.4 ppc64
Build:c910f02c01p14:~ # lsxcatd -v
Version 2.13 (git commit 49e6f4ba89a280099caa222cc6f46ee3afb23ae1, built Thu Nov 24 06:15:37 EST 2016)
c910f02c01p14:~ # 

How to reproduce:
```
1.nodeset node diskfull install
2.after provison the node status alway booting.
    lsdef -l c910f02c08p07 | grep status 
    ------------> always booting here
3.xdsh c910f02c08p07 date
   
    c910f02c08p07: Thu Nov 24 12:59:20 EST 2016
    ------------------------> successfull here
4.according to xcat.log on c910f02c08p07:

System start/stop links for /etc/init.d/ntp already exist.
Postscript: setupntp exited with code 0
Thu Nov 24 20:31:17 EST 2016 postscript setupntp return with 0
Generating grub configuration file ...
Warning: Setting GRUB_TIMEOUT to a non-zero value when GRUB_HIDDEN_TIMEOUT is set is no longer supported.
Found linux image: /boot/vmlinuz-3.19.0-74-generic
Found initrd image: /boot/initrd.img-3.19.0-74-generic
done
Generating grub configuration file ...
Warning: Setting GRUB_TIMEOUT to a non-zero value when GRUB_HIDDEN_TIMEOUT is set is no longer supported.
Found linux image: /boot/vmlinuz-3.19.0-74-generic
Found initrd image: /boot/initrd.img-3.19.0-74-generic
done
x86_64
Generating grub configuration file ...
Warning: Setting GRUB_TIMEOUT to a non-zero value when GRUB_HIDDEN_TIMEOUT is set is no longer supported.
Found linux image: /boot/vmlinuz-3.19.0-74-generic
Found initrd image: /boot/initrd.img-3.19.0-74-generic
done
running /xcatpost/mypostscript.post
Thu Nov 24 20:32:25 EST 2016 Running postscript: otherpkgs
./otherpkgs: no extra rpms to install
Thu Nov 24 20:32:25 EST 2016 postscript otherpkgs return with 0
/xcatpost/mypostscript.post return
 ```

[...]

PR #2212, opened 2016-11-29
fix issue [fvt]diskfull provison node status always booting #2204

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2204


daniceexi:

PR #2215, opened 2016-11-29
change the shebang of xcatpostinit1 to bash

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2216
 in case that sh is not linked to bash
xcatpostinit1 used some bash features,


daniceexi:

PR #2221, opened 2016-11-29
Have snmpmon set the community string

jjohnson42:
Some BMCs do not default to 'public' for alerts.
Change this to do public for now, since other parts
of the code are still hard baked.


whowutwut:
Since #2223 was reviewed by @daniceexi  and merged in, going to also merge this into 2.12 branch.  Currently there is no plan at this time to create a 2.12.5 release.

PR #2223, opened 2016-11-29
Have snmpmon set the community string

jjohnson42:
Some BMCs do not default to 'public' for alerts.
Change this to do public for now, since other parts
of the code are still hard baked.


whowutwut:
@daniceexi @zet809  Jarrod would like this to be integrated into the code for their upcoming release.., let's see if we can review and merge for 2.13...

daniceexi:
I am OK with the change.

Issue #2225, opened 2016-11-29
[CUSTOMER] Stale rootimg.gz file left in the rootimg dir caused node boot failed.


cxhong:
This issue found from PMR 08785,999,000 - unable to provision diskless nodes.

On the customer system,  the stale rootimg.gz left in the rootimg dir:
`````
# ls -ltr /install/netboot/rhels7.2/ppc64le/cn
total 1909852
-rw-r--r--  1 root root 911616542 Sep 15 14:27 rootimg.gz
-rwxr-xr-x  1 root root  18484632 Nov 29 09:09 kernel
-rw-------  1 root root  39652770 Nov 29 09:14 initrd-stateless.gz
-rw-------  1 root root  36876313 Nov 29 09:15 initrd-statelite.gz
dr-xr-xr-x 20 root root      4096 Nov 29 09:17 rootimg
-rw-r--r--  1 root root 949043824 Nov 29 09:19 rootimg.cpio.gz
`````
nodeset command picked up rootimg.gz instead of rootimg.cpio.gz and  failed to reprovision node.

I create a test osimage, copy rootimg.cpio.gz to rootimg.gz and ran rmimage command.
the rmimage didn't remove rootimg.gz
````
# ls -ltr /install/netboot/rhels7.3/ppc64le/compute-netboot
total 562680
-rwxr-xr-x  1 root root  19008920 Nov 29 11:26 kernel
-rw-------  1 root root  36963784 Nov 29 11:26 initrd-stateless.gz
-rw-------  1 root root  34264846 Nov 29 11:27 initrd-statelite.gz
dr-xr-xr-x 20 root root       274 Nov 29 11:31 rootimg
-rw-r--r--  1 root root 242967845 Nov 29 11:31 rootimg.cpio.gz
-rw-r--r--  1 root root 242967845 Nov 29 14:20 rootimg.gz

# rmimage netboot-compute
Removing directory /install/netboot/rhels7.3/ppc64le/compute-netboot/rootimg
Removing file /install/netboot/rhels7.3/ppc64le/compute-netboot/kernel
Removing file /install/netboot/rhels7.3/ppc64le/compute-netboot/initrd-stateless.gz
Removing file /install/netboot/rhels7.3/ppc64le/compute-netboot/initrd-statelite.gz
Removing file /install/netboot/rhels7.3/ppc64le/compute-netboot/rootimg.cpio.gz
Removing directory /tftpboot/xcat/osimage/netboot-compute
[...]

PR #2228, opened 2016-11-30
fix issue [CUSTOMER] Stale rootimg.gz file left in the rootimg dir caused node boot failed. #2225

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2225


cxhong:
Thanks, @immarvin

whowutwut:
@immarvin  In the old code in `packimage.pm` , there are unlink calls to `rootimg.gz` file, do we know why this didn't remove the sale file on the customer site? 

I think they ran rmimage, genimage, packimage...

PR #2240, opened 2016-12-01
fix statelite persistent dir use short name

bybai:
After fixed, will use short name as following:

[root@bybc0607 ~]# df
Filesystem                       1K-blocks     Used Available Use% Mounted on
rootfs                             2477816   600496   1877320  25% /
tmpfs                              2477816    16672   2461144   1% /run
rw                                 2477816    56792   2421024   3% /tmp
bybc0602:/nodedata/bybc0607       27769856 14007296  13762560  51% /.statelite/persistent
rw                                 2477816        4   2477812   1% /.sllocal/log


bybai:
fix issue #2241

immarvin:
agree to merge

PR #2243, opened 2016-12-01
Become process group leader when updating firmware

chenglch:
As exec is invoked, child process can not ignore the signal sent
to the process group, use setpgrp to set the pgid to itself.

close-issue: #1849


zet809:
Looks good to me, thx

PR #2246, opened 2016-12-01
Add the security bulletin for the removal of the service stream PWs for 2.9 branch

whowutwut:
Prepare the security bulletin for the vulnerability for having hard coded HMC/FSP password used by Service teams. Target to merge this Nov 30, 2016. The 2.9.x release needs to be built and released at the same date.


zet809:
Thx, Victor.

PR #2250, opened 2016-12-02
Make changes to xCAT-genesis-build process to create version dependencies in xCAT-genesis-base

whowutwut:
Fixes Issues #2239 

Includes the following changes: 

- Bump up the version to 2.13.0
- Edit the makerpm to replace version of xCAT-genesis-base.spec
  file so there's no need to manually set the VERSION.  We
  do need a better way for the RELEASE variable
- Require xCAT-genesis-base >= 2.13.0 for the rebuild of genesis-base
- Clean up the builder rpm to include directories to clean up correctly
  on rpm removal

With this change the builder-rpm is still created with the same naming: 
```
$ ./makerpm xCAT-genesis-builder
Building /home/vhu/rpmbuild/RPMS/noarch/xCAT-genesis-builder-2.13.0-snap*.noarch.rpm ...
```

But when we create the genesis-base, the result is the following: 
```
[root@c910f02c05p05 vhu]# /opt/xcat/share/xcat/netboot/genesis/builder/buildrpm -y
cp: omitting directory ‘/opt/xcat/share/xcat/netboot/genesis/builder/debian’
Creating the initramfs in /tmp/xcatgenesis.20848.rfs using dracut ...
Disabling early microcode, because kernel does not support it. CONFIG_MICROCODE_[AMD|INTEL]_EARLY!=y

Expanding the initramfs into /tmp/xcatgenesis.20848/opt/xcat/share/xcat/netboot/genesis/ppc64/fs ...
...
...
Wrote: /root/rpmbuild/SRPMS/xCAT-genesis-base-ppc64-2.13.0-0.src.rpm
Wrote: /root/rpmbuild/RPMS/noarch/xCAT-genesis-base-ppc64-2.13.0-0.noarch.rpm
Executing(%clean): /bin/sh -e /var/tmp/rpm-tmp.BE3mUu
+ umask 022
+ cd /root/rpmbuild/BUILD
+ /usr/bin/rm -rf /root/rpmbuild/BUILDROOT/xCAT-genesis-base-ppc64-2.13.0-0.ppc64
+ exit 0

```

Result: 
```
[root@c910f02c05p05 vhu]# rpm -qip --provides /root/rpmbuild/RPMS/noarch/xCAT-genesis-base-ppc64-2.13.0-0.noarch.rpm
Name        : xCAT-genesis-base-ppc64
Version     : 2.13.0
Release     : 0
Architecture: noarch
Install Date: (not installed)
Group       : System/Utilities
Size        : 172959487
[...]

PR #2260, opened 2016-12-05
Revert the naming of xCAT-genesis-base back to using snap<date> and use Epoch instead 

whowutwut:
The changes in #2250 attempted to change the versioning of xCAT-genesis-base to something more standard, removing the `snap<date>` naming from the RPM.  This was so that xCAT-genesis-scripts could require a certain version of the xCAT-genesis-base so that it ensures that the end user has configured the correct xcat-dep repository if we made changes that need xcat-core and xcat-dep to be in sync.  

What I was unaware of is the high use of `Epoch` in the xCAT rpms.  My guess is that the irregular naming convention for the xCAT rpms caused for this `Epoch` keyword to control which package is "higher priority" 

```
$ git grep "Epoch"  | grep ".spec" | wc  -l
18
```

So in testing the changes from #2250, the behvior is not as I wanted it to be.  I knew that `yum update "*xCAT*"` would not pull in a new xCAT-genesis-base version because of the `Epoch` already set in the older RPM.   However, I thought that doing `yum update xCAT-genesis-base` would force the update, but it turns out that it does not.

So rather than fight this Epoch, I'm going to just bump it up `Epoch: 2` and go back to the snap naming convention.  

Then the expected behavior happens. 
```
[root@fs4 ppc64le]# yum update xCAT-genesis-base*
Examining xCAT-genesis-base-ppc64-2.13.0-snap201612051402.noarch.rpm: 2:xCAT-genesis-base-ppc64-2.13.0-snap201612051402.noarch
Marking xCAT-genesis-base-ppc64-2.13.0-snap201612051402.noarch.rpm as an update to 1:xCAT-genesis-base-ppc64-2.12-snap201610250931.noarch
Examining xCAT-genesis-base-x86_64-2.13.0-snap201612051408.noarch.rpm: 2:xCAT-genesis-base-x86_64-2.13.0-snap201612051408.noarch
Marking xCAT-genesis-base-x86_64-2.13.0-snap201612051408.noarch.rpm as an update to 1:xCAT-genesis-base-x86_64-2.12-snap201607221057.noarch
Resolving Dependencies
--> Running transaction check
---> Package xCAT-genesis-base-ppc64.noarch 1:2.12-snap201610250931 will be updated
---> Package xCAT-genesis-base-ppc64.noarch 2:2.13.0-snap201612051402 will be an update
---> Package xCAT-genesis-base-x86_64.noarch 1:2.12-snap201607221057 will be updated
---> Package xCAT-genesis-base-x86_64.noarch 2:2.13.0-snap201612051408 will be an update
--> Finished Dependency Resolution

Dependencies Resolved

====================================================================================================================================================
 Package                        Arch         Version                           Repository                                                      Size
====================================================================================================================================================
Updating:
 xCAT-genesis-base-ppc64        noarch       2:2.13.0-snap201612051402         /xCAT-genesis-base-ppc64-2.13.0-snap201612051402.noarch        165 M
 xCAT-genesis-base-x86_64       noarch       2:2.13.0-snap201612051408         /xCAT-genesis-base-x86_64-2.13.0-snap201612051408.noarch        71 M

Transaction Summary
====================================================================================================================================================```

Still keeping the 2.13.0 "three number" naming for the xCAT-genesis-base so that we can match up the minor release that a new version was created for.

### DIFF
--- a/makerpm
+++ b/makerpm
@@ -165,9 +165,7 @@ function makegenesis {
 	SPEC_FILE="xCAT-genesis-base.spec"
 	RPMNAME="$1"
 	cd `dirname $0`/$DIR
-	GEN_BASE_REL=0
 	sed -i s/%%REPLACE_CURRENT_VERSION%%/${VER}/g ${SPEC_FILE}
-	sed -i s/%%REPLACE_RELEASE%%/${GEN_BASE_REL}/g ${SPEC_FILE}
 	tar --exclude .svn -cjf $RPMROOT/SOURCES/$RPMNAME.tar.bz2 .
 	# undo the changes from the SED command above so it's not tracked by Git
         git checkout ${SPEC_FILE}
--- a/xCAT-genesis-builder/xCAT-genesis-base.spec
+++ b/xCAT-genesis-builder/xCAT-genesis-base.spec
@@ -1,5 +1,4 @@
 %define version     %%REPLACE_CURRENT_VERSION%%
-%define release     %%REPLACE_RELEASE%%
 Version: %{?version:%{version}}%{!?version:%(cat Version)}
 Release: %{?release:%{release}}%{!?release:snap%(date +"%Y%m%d%H%M")}
 %ifarch i386 i586 i686 x86
@@ -18,6 +17,7 @@ BuildArch: noarch
 %define __prelink_undo_cmd %{nil}
 # To fix the issue error: Arch dependent binaries in noarch package, the following line is needed on Fedora 23 ppc64
 %define _binaries_in_noarch_packages_terminate_build   0
+Epoch: 2
 AutoReq: false
 Prefix: /opt/xcat
 AutoProv: false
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by whowutwut** [comment]
I know I'm bypassing process but I want to create a xcat-core and xcat-dep that closely resembles the GA version so that we can start testing the RC builds, I'm going to merge this.  I've already created a version of xCAT-genesis-base from the changes in this PR.

**Comment by whowutwut** [comment]
Just for completeness , I did some investigation here today.  It turns out that there was no need to bump `Epoch: 2` .  If I had left `Epoch: 1` and restored the naming to 2.13.0-snap<date> it would have been successful in updating from 2.12-snap<date> format.  

I think in my dev time, I changed the number to 2.13.0-0 and that version was unable to tell that it was > 2.12-snap<date> so I was not able to upgrade as expected.  

```
---> Package xCAT-genesis-base-ppc64.noarch 1:2.12-snap201610250931 will be updated
---> Package xCAT-genesis-base-ppc64.noarch 1:2.13.0-snap201612061320 will be an update
```

@daniceexi @zet809 
However, i think that it's OK to leave the Epoch: 2 for 2.13 and leave the numbering as is since we would require a re-build of xCAT-genesis-base and repackage of xcat-dep tar.  Also the current one is available for 24 hours now :-)

zet809:
OK, I agree. Thx a lot!

PR #2270, opened 2016-12-06
modify testcases using "makedhcp -q " to validate the "makedhcp"  res…

junxiawang:
@hu-weihua  modify testcases using "makedhcp -q " to validate the "makedhcp" result check  for issue 2265


hu-weihua:

Issue #2263, opened 2016-12-06
[DEV] go-xcat test case in ubuntu16.04.1 still obtain apt.key from SourceForge


immarvin:
xCAT doc mentioned:
````
The xCAT GPG Public Key must be added for apt to verify the xCAT packages

wget -O - "http://xcat.org/files/xcat/repos/apt/apt.key" | apt-key add -

````

however, we are still downloading apt.key from SF, which is not stable enough.
````
==========
[if grep Ubuntu /etc/*release;then xdsh c910f03fsp03v04 "scp -r c910f03fsp03v02:/etc/resolv.conf /etc/resolv.conf" && xdsh c910f03fsp03v04 "wget -O - http://sourceforge.net/projects/xcat/files/ubuntu/apt.key/download | apt-key add -"; fi] Running Time:45 sec
RETURN rc = 1
OUTPUT:
/etc/lsb-release:DISTRIB_ID=Ubuntu
/etc/lsb-release:DISTRIB_DESCRIPTION="Ubuntu 16.04.1 LTS"
/etc/os-release:NAME="Ubuntu"
/etc/os-release:PRETTY_NAME="Ubuntu 16.04.1 LTS"
c910f03fsp03v04: URL transformed to HTTPS due to an HSTS policy
c910f03fsp03v04: --2016-12-05 13:16:27--  https://sourceforge.net/projects/xcat/files/ubuntu/apt.key/download
c910f03fsp03v04: Resolving sourceforge.net (sourceforge.net)... 216.34.181.60
c910f03fsp03v04: Connecting to sourceforge.net (sourceforge.net)|216.34.181.60|:443... connected.
c910f03fsp03v04: HTTP request sent, awaiting response... 302 Found
c910f03fsp03v04: Location: http://downloads.sourceforge.net/project/xcat/ubuntu/apt.key?r=&ts=1480961374&use_mirror=superb-sea2 [following]
c910f03fsp03v04: --2016-12-05 13:16:27--  http://downloads.sourceforge.net/project/xcat/ubuntu/apt.key?r=&ts=1480961374&use_mirror=superb-sea2
c910f03fsp03v04: Resolving downloads.sourceforge.net (downloads.sourceforge.net)... 216.34.181.59
c910f03fsp03v04: Connecting to downloads.sourceforge.net (downloads.sourceforge.net)|216.34.181.59|:80... connected.
c910f03fsp03v04: HTTP request sent, awaiting response... 302 Found
[...]

PR #2271, opened 2016-12-06
fix issue [DEV] go-xcat test case in ubuntu16.04.1 still obtain apt.key from SourceForge #2263

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2263


caomengmeng:
Verified in ubunt16.04.1, merge it

PR #2272, opened 2016-12-06
modify testcase for rmvm issues:2235

junxiawang:
@zet809  modify testcase for rmvm issues


zet809:
According to @junxiawang , the test case xCAT-test/autotest/testcase/installation/reg_linux_SN_installation_hierarchy  had been replaced by xCAT-test/autotest/testcase/installation/SN_setup_case, need to deal with the case after the 2.13 release.

Issue #2262, opened 2016-12-06
[DEV] invalid condition to validate the result in "runcmdinstaller" test case 


immarvin:
````
RUN:runcmdinstaller c910f02c08p07 "ls -al /tmp"

[runcmdinstaller c910f02c08p07 "ls -al /tmp"] Running Time:1 sec
RETURN rc = 0

OUTPUT:
total 897
drwxrwxrwt.  9 root root   520 Dec  5 19:27 .
dr-xr-xr-x. 14 root root  1024 Dec  5 19:27 ..
drwxrwxrwt.  2 root root    40 Dec  5 19:27 .ICE-unix
drwxrwxrwt.  2 root root    40 Dec  5 19:27 .Test-unix
drwxrwxrwt.  2 root root    40 Dec  5 19:27 .X11-unix
drwxrwxrwt.  2 root root    40 Dec  5 19:27 .XIM-unix
drwxrwxrwt.  2 root root    40 Dec  5 19:27 .font-unix
-rw-r--r--.  1 root root  2287 Dec  5 19:27 anaconda.log
-rwxr-xr-x.  1 root root   294 Dec  5 19:27 baz.py
-rwxr-xr-x.  1 root root  2113 Dec  5 19:27 foo.py
-rw-r--r--.  1 root root 10480 Dec  5 19:27 ifcfg.log
-rwx------.  1 root root 14635 Dec  5 19:27 ks-script-AXq1pe
-rw-r--r--.  1 root root     0 Dec  5 19:27 ks-script-AXq1pe.log
-rw-------.  1 root root   183 Dec  5 19:27 libuser.kSwoNm
-rw-------.  1 root root   235 Dec  5 19:27 libuser.u0F7k2
-rw-r--r--.  1 root root     0 Dec  5 19:27 packaging.log
-rw-r--r--.  1 root root   350 Dec  5 19:27 partitionfile
-rw-r--r--.  1 root root 15942 Dec  5 19:27 pre-install.log
-rw-r--r--.  1 root root  1789 Dec  5 19:27 program.log
-rw-r--r--.  1 root root    55 Dec  5 19:27 repos
-rw-r--r--.  1 root root     0 Dec  5 19:27 sensitive-info.log
-rw-r--r--.  1 root root   832 Dec  5 19:27 storage.log
-rw-------.  1 root root 39769 Dec  5 19:27 syslog
[...]

PR #2273, opened 2016-12-06
[FVT]Update runcmdinstaller_command case to check the root path

caomengmeng:
To fix the runcmdinstaller_command case for below error:
https://github.com/xcat2/xcat-core/issues/2262
@immarvin  please help to review it.


immarvin:
agree to merge

whowutwut:
@caomengmeng @immarvin Was this change created directly on the xcat-core project?

immarvin:
hi @whowutwut , yes, the PR goes into xcat-core/xCAT-test

PR #2278, opened 2016-12-07
Clean up the errors in doxcat and add messages for tracing issues  [ DO NOT MERGE ] 

whowutwut:
We have seen some reports from customers that hardware discovery takes up to 15 minutes in some cases in the genesis kernel.  It looks like it has something to do with NTP but the actual reason is not known.  

```
[Tue Nov 29 11:18:03 2016]ppc64
```

Between the IP address and "ppc64" it's around 16 minutes.  

The changes in this pull request is to clean up the error messages that come out during genesis `doxcat` script runs and have better logging messages to pinpoint the issue.

### DIFF
--- a/xCAT-genesis-scripts/bin/doxcat
+++ b/xCAT-genesis-scripts/bin/doxcat
@@ -5,24 +5,37 @@
 #
 log_label="xcat.genesis.doxcat"
 
-# Start rsyslogd and log into a local file specified in /etc/rsyslog.conf
-# Later, once xCAT MN is known, dhclient-script will change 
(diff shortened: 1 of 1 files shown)


whowutwut:
@zet809  I created this pull request so that it can be reviewed.  I'm debating if we should add these changes since we are re-spinning the xCAT-genesis-base for 2.13.0 release.  I still need to UT these changes as I didn't get around to it today, but will do it first thing tomorrow.

whowutwut:
The issue is fixed in https://github.com/xcat2/xcat-core/pull/2307

PR #2279, opened 2016-12-07
modify genesis testcase for nodeset shell case issue

junxiawang:
@zet809 please help review
modify genesis test case for nodeset_shell case,


zet809:
I agree to merge.

Issue #2282, opened 2016-12-07
[DEV] "makentp -a" always return 1 


immarvin:
````
[root@c910f03c17k27 ~]# makentp -a
configuring management node: c910f03c17k27.
configuring service nodes: c910f03c17k28
c910f03c17k28: vpdupdate is not supported on the pSeries KVM Guest platform
c910f03c17k28: xcatdsklspost: updating VPD database
c910f03c17k28: xcatdsklspost: downloaded postscripts successfully
c910f03c17k28: Wed Dec  7 02:05:16 EST 2016 Running postscript: setupntp
c910f03c17k28: active
c910f03c17k28: systemctl stop ntpd
c910f03c17k28:  7 Dec 02:05:32 ntpdate[15237]: adjust time server 10.3.17.27 offset -0.000020 sec
c910f03c17k28: SYNC_HWCLOCK=yes
c910f03c17k28: postscript: setupntp exited with code 0
c910f03c17k28: Running of postscripts has completed.


[root@c910f03c17k27 ~]# echo $?
1

[root@c910f03c17k27 ~]# lsxcatd -v
Version 2.13.0 (git commit 99da40c22a1b442a60eb346ee337442d155df2fb, built Tue Dec  6 05:00:44 EST 2016)
[root@c910f03c17k27 ~]#
````

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2283

PR #2283, opened 2016-12-07
fix issue [DEV] 'makentp -a' always return 1 #2282

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2282


bybai:
The old code direct use "1" as return value as ``send_msg(\%request, 1, "$msg");``
I agreed with the code change use actual return value instead of 1.

daniceexi:

PR #2285, opened 2016-12-07
SN ntp service setup

junxiawang:
@tingtli ,@immarin modify sn's ntp service setup for testcase.
Please help review.


immarvin:
agree to merge

Issue #2287, opened 2016-12-07
[DEV]genimage failed due to incorrect lock file path


immarvin:
````
[root@c910f04x14v02 ~]# genimage rhels6.8-x86_64-netboot-compute
Generating image:
cd /opt/xcat/share/xcat/netboot/rh; ./genimage -a x86_64 -o rhels6.8 -p compute --permission 755 --srcdir "/install/rhels6.8/x86_64" --pkglist /opt/xcat/share/xcat/netboot/rh/compute.rhels6.x86_64.pkglist --otherpkgdir "/install/post/otherpkgs/rhels6.8/x86_64" --postinstall /opt/xcat/share/xcat/netboot/rh/compute.rhels6.x86_64.postinstall --rootimgdir /install/netboot/rhels6.8/x86_64/compute --tempfile /tmp/xcat_genimage.9415 rhels6.8-x86_64-netboot-compute
readline() on closed filehandle $pidfd at /opt/xcat/lib/perl/xCAT/Utils.pm line 4822.
Use of uninitialized value $pid in concatenation (.) or string at /opt/xcat/lib/perl/xCAT/Utils.pm line 4824.
failed to acquire lock, seems there is another genimage/packimage/rmimage process  running on root image dir "/install/netboot/rhels6.8/x86_64/compute/rootimg"
Error: Command failed: cd /opt/xcat/share/xcat/netboot/rh; ./genimage -a x86_64 -o rhels6.8 -p compute --permission 755 --srcdir "/install/rhels6.8/x86_64" --pkglist /opt/xcat/share/xcat/netboot/rh/compute.rhels6.x86_64.pkglist --otherpkgdir "/install/post/otherpkgs/rhels6.8/x86_64" --postinstall /opt/xcat/share/xcat/netboot/rh/compute.rhels6.x86_64.postinstall --rootimgdir /install/netboot/rhels6.8/x86_64/compute --tempfile /tmp/xcat_genimage.9415 rhels6.8-x86_64-netboot-compute 2>&1. Error message: readline() on closed filehandle $pidfd at /opt/xcat/lib/perl/xCAT/Utils.pm line 4822.
Use of uninitialized value $pid in concatenation (.) or string at /opt/xcat/lib/perl/xCAT/Utils.pm line 4824.
failed to acquire lock, seems there is another genimage/packimage/rmimage process  running on root image dir "/install/netboot/rhels6.8/x86_64/compute/rootimg".

````

PR #2288, opened 2016-12-07
correct the path of genimage lock

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2287


daniceexi:

whowutwut:
Was this introduced by a recent change?

immarvin:
hi @whowutwut , this is introduced in https://github.com/xcat2/xcat-core/pull/2036/files. it's my fault during code check in

PR #2292, opened 2016-12-08
Add messages to trace the ntpd commands in genesis

whowutwut:
We have seen some reports from customers that hardware discovery takes up to 15 minutes in some cases in the genesis kernel. It looks like it has something to do with NTP but the actual reason is not known.
```
[Tue Nov 29 11:18:03 2016]ppc64
```
Between the IP address and "ppc64" it's around 16 minutes.

To help us trace the time that it takes, adding some more detailed logger messages between the ntpd calls


whowutwut:
@daniceexi  As discussed on the phone, this pull request change is safer than what I did in #2278.  It's also affect the xCAT-genesis-scripts, not xCAT-genesis-base, so it's a little better.  

Regardless of whether we merge for 2.13. or wait for 2.13.1, I will try and run discovery tomorrow with this to see if we can see any delay in time from the ntpd code path.

daniceexi:
The change looks safe. So I'll merge it to master branch. If we have to respin the build before the GA, this change will be added.

Issue #2297, opened 2016-12-09
[fvt]sles12.2 mysqlsetup -i could not work


junxiawang:
env:sles12.2
build:
c910f04x30v24:~ # lsdef -v
lsdef - Version 2.13.0 (git commit baf1b4b73da49a0cb4dd71873dfe6d9d70a4794d, built Wed Dec  7 04:35:35 EST 2016)

How to reproduce:
```
c910f04x30v24:~ # lsdef -v
lsdef - Version 2.13.0 (git commit baf1b4b73da49a0cb4dd71873dfe6d9d70a4794d, built Wed Dec  7 04:35:35 EST 2016)

c910f04x30v24:~ # mysqlsetup -i
Input the alpha-numberic  password for xcatadmin in the MySQL database: 
Input the password for root in the MySQL database: 
 Could not start the mysql daemon, in time allocated ( 2minutes)
```
actually, the mysqld is running
````
c910f03c17k22:~ # ps aux|grep mysql|grep -v grep
mysql    21937  1.0  1.6 1486912 104192 ?      Ssl  03:13   0:02 /usr/sbin/mysqld --defaults-file=/etc/my.cnf --user=mysql
````

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2298

immarvin:
hi @junxiawang , please verify the issue

PR #2298, opened 2016-12-09
fix issue [fvt]sles12.2 mysqlsetup -i could not work #2297

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2297

### DIFF
--- a/xCAT-client/bin/mysqlsetup
+++ b/xCAT-client/bin/mysqlsetup
@@ -165,12 +165,8 @@ if ($::linuxos == "sles12") {
     close($relfile);
     chomp(@lines);
 
-    # if == 12.1
-    if (grep /VERSION_ID="12.1"/, @lines) {
-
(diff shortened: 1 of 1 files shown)


hu-weihua:
I have reviewed the code change and agree to merge, thanks @immarvin.

xCAT-client/bin/mysqlsetup
```diff
@@ -165,12 +165,8 @@ if ($::linuxos == "sles12") {
     close($relfile);
     chomp(@lines);
 
-    # if == 12.1
-    if (grep /VERSION_ID="12.1"/, @lines) {
-
-        #$::linuxos == "sles12.1";
-        $sp1flag = 1;
-    }
+    # for sles12.x
```
daniceexi: The fix is a little strange. 
Looks like this '$sp1flag' is working for all the sles12, should we change the variable name to be a reasonable one?

Issue #2299, opened 2016-12-09
[fvt]xcatprobe xcatmn do not clear the /etc/hosts if no usefull cn node to check dhcp in the environment


junxiawang:
env:all os
build:[root@c712f1n05 subcmds]# lsxcatd -v
Version 2.13.0 (git commit 99da40c22a1b442a60eb346ee337442d155df2fb, built Tue Dec  6 05:00:44 EST 2016)
[root@c712f1n05 subcmds]# 
How to reproduce:
```
1.if no cn for dhcp probe
2.xcatprobe xcatmn -i eth0
3.cat /etc/hosts:
9.114.39.15 xcatmntest xcatmntest.pok.stglabs.ibm.com
-------------------> this line will not be clear after xcatprobe command
```

hu-weihua:
@junxiawang , this issue have been fixed by pull request 2300, Cloud you verify it? thanks.

PR #2300, opened 2016-12-09
Modify issue2299: xcatprobe xcatmn do not clear the /etc/hosts if no usefull cn node to check dhcp in the environment

hu-weihua:
Modify issue #2299: xcatprobe xcatmn do not clear the /etc/hosts if no usefull cn node to check dhcp in the environment


xuweibj:
Agree to merge.

PR #2305, opened 2016-12-13
Global symbol "$csw" requires explicit package name at configBNT

cxhong:
fix syntax error


whowutwut:
I have reviewed and will merge

PR #2311, opened 2016-12-14
add sles12.2 bundle files

caomengmeng:
@junxiawang please help to review.thanks.


whowutwut:
@caomengmeng (FYI @junxiawang, we should block this PR) You should be creating pull requests from the forked copy, instead of direct clone to xcat2/xcat-core.  Please check `git remote -v` in your clone and make sure the origin is your forked copy and the upstream is `xcat2/xcat-core`

whowutwut:
@caomengmeng  Please also add milestones to the pull request... and labels.

Issue #2302, opened 2016-12-12
makeconserver -d should display an error if no noderange specified


gurevichmark:
The man page for `makeconserver -d` shows that *noderange* is a required parameter, but the code allows running `makeconserver -d` and no error is displayed. 
This is confusing, as the user might think that **all** node entries were removed from `/etc/conserver.cf` file.

The fix to this issue could be either to display an error when *noderange* is not specified, or change `makeconserver -d` to remove all node entries from the `/etc/conserver.cf` file.

gurevichmark:
Since it appears that `conserver.pm` already has code to remove all the xCAT entries from /etc/conserver.cf file, the pull request #2321 will enable that. This way `makeconservercf -d` will remove all the xCAT node entries from /etc/conserver.cf file if no *noderange* is specified.

PR #2321, opened 2016-12-15
makeconservercf -d to remove all xCAT nodes

gurevichmark:
This pull request fixes the problem reported by #2302 
1. Added `$delmode` flag when calling `donodeent()`. This call is done in 2 places, first when *noderange* is specified, second when *noderange* is not specified. First call to `donodeent()` was already passing `$delmode`, this pull request adds `$delmode` flag to the second call of  `donodeent()`.
2. Modified man page for `makeconservercf` to make it more clear what happens when *noderange* is not specified.


whowutwut:
@chenglch can you also take a look at these changes?

PR #2324, opened 2016-12-16
modify ubuntu migration case

junxiawang:
@immarvin please help review


immarvin:
'.*xcat*. ' is not a valid regular expression,shouls it be '.*xcat.*'?

immarvin:
agree to merge

Issue #2325, opened 2016-12-16
[CUSTOMER] updateflag.awk fail with timeout or be blocked with "Retrying flag update"  when provisioning a bunch of node 


immarvin:
the is reported from the customer,  when they reprovision 516 nodes concurrently, some nodes delayed a significant time to update its status, and several nodes failed to update their status to "booted". in the xcat.log, they observed this:
````
updateflag.awk: Retrying flag update
````

**Comment by immarvin** [comment]
fixed in https://github.com/xcat2/xcat-core/pull/2326

**Comment by immarvin** [comment]
another PR related to this issue https://github.com/xcat2/xcat-core/pull/2318

**Comment by Obihoernchen** [comment]
Is it feasible (without causing other problems) to cherry-pick both PRs for 2.12.3?

**Comment by immarvin** [comment]
hi @Obihoernchen , the PR https://github.com/xcat2/xcat-core/pull/2326 have some problems during review.
It is ok to cherry-pick the PR https://github.com/xcat2/xcat-core/pull/2318 for 2.12.3

**Comment by Obihoernchen** [comment]
I did some more research on this because it was way too annoying to have those issues every boot.

First of all I've patched xcatd of all our 2.12.3 service xcat nodes with https://github.com/xcat2/xcat-core/pull/2318/commits/3d3f367f69d18e09e849101d7110e16536cc2030

Now I don't see the `xcatd: possible BUG encountered by xCAT DB worker SIGPIPE xcatd: [...]` error anymore.
But it doesn't seem to fix anything else except this log message.

------------------------------

If you boot more than 5 nodes simultaneously, 50% of them will fail to boot.
The named daemon will go to (and stay at) 100% cpu load on the corresponding xcat service node.
Actually named gets ddosed by all compute nodes with boot issues.

The reason for this is the `/tmp/updateflag` script in `initrd-stateless.gz`.
The script is created during the genimage process: https://github.com/xcat2/xcat-core/blob/master/xCAT-server/share/xcat/netboot/sles/genimage#L1751
​
The problem with this script is, that it has no loop exit condition and no retry delay at all.
So if xcat is unable to update the nodestatus it will run the following code indefinitely: 
```awk
        ns = "/inet/tcp/0/" xcatdhost "/" xcatiport
        print "xCAT_xcatd" |& ns
        while(1) {
                ns |& getline
                if(\$0 == "ready")
                        print ARGV[3] |& ns
                if(\$0 == "done")
                        break
        }

```
[...]

PR #2326, opened 2016-12-16
fix [CUSTOMER] updateflag.awk fail with timeout or be blocked with "Retrying flag update" when provisioning a bunch of node #2325

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2325 :

   add a message queue and a worker process to process the status update message from the compute nodes asynchronously, to prevent block or timeout of updateflag.awk, this might happen when many nodes reports their status concurrently

### DIFF
--- a/xCAT-server/sbin/xcatd
+++ b/xCAT-server/sbin/xcatd
@@ -29,7 +29,7 @@ my $sslctl;
 my $udpctl;
 my $pid_UDP;
 my $pid_MON;
-
+my $pid_queueworker;
 # ----used for command log start---------
 my $cmdlog_svrpid;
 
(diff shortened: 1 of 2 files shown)


immarvin:
hi @chenglch , code has been updated

the queue worker process will be children of install monitor service process

xCAT-server/sbin/xcatd
```diff
@@ -461,22 +488,16 @@ sub do_installm_service {
...
-                        my %request = (
-                            command => ['updatenodestat'],
-                            node    => [$node],
-                            arg     => ["$newstat"],
-                        );
-
-                        # node should be blocked, race condition may occur otherwise
-                        #my $pid=xCAT::Utils->xfork();
-                        #unless ($pid) { # fork off the nodeset and potential slowness
-                        plugin_command(\%request, undef, \&build_response);
```
chenglch: Just a little curious about the original implementation. `plugin_command` will execute multiple times if multiple `node stat` is given in `@tmpa`?

PR #2331, opened 2016-12-19
add xcatprobe discovery time info when monitor

xuweibj:
Show time info when monitoring.


hu-weihua:

Issue #2327, opened 2016-12-16
[CUSTOMER] - ntp code in doxcat causing discovery delays up to 15 minutes 


whowutwut:
Customer reported this issue during hardware discovery of 50+ nodes where a few seem to take 15 minutes before actually getting discovered. 

The logs provided are in the `doxcat` code 

```
[Tue Nov 29 11:02:05 2016]cp: cannot stat '/usr/share/zoneinfo/posix/America/Los_Angeles': No such file or directory
[Tue Nov 29 11:02:07 2016]<166>Nov 29 11:02:06 xcat.genesis.doxcat: Acquired IPv4 address on enP5p7s0f1
[Tue Nov 29 11:02:07 2016]192.168.64.174/24
[Tue Nov 29 11:18:03 2016]ppc64   <-- 15 minutes later
```

Looking at the message around this, the only possible code path where we would wait is in the ntp code where we are attempting to sync the clock in the genesis kernel, then sync the hwclock on the server so it will persist on the next reboot. 

Creating this issue to track the discussions .

zet809:
I think we can close this defect first, if there is still issues after customer verified with our fix, reopen it.

whowutwut:
@zet809   Just want to mention is that the changes made doesn't actually do anything with regards to reducing the 15 minute time required to sync the clock. It only makes it more transparent to the user what the code is actually doing during this wait time.

zet809:
@whowutwut Yes, you are right. There are some logs added in the doxcat script, then if the issue happened again, we will know whether the genesis hung for the ntp reason.

whowutwut:
@zet809 @neo954 So with the JEMT Garrison under fs4, I think I hit this problem again trying to test out HW discovery to validate some defects for 2.13.2...   This is very frustrating and I still don't understand why we don't simply run a `ntpd -g -x -q` to attempt to sync the clocks, if failed, it continues.  

Few things I do not like about the new spinning shell....  I am no longer able to copy and paste with the terminal that I have because the spinning icon resets rcons.  So I will provide a screen capture... 

Notice the time differences....  Not sure why that happens. 

![image](https://cloud.githubusercontent.com/assets/10049070/23276371/fe51d6d8-f9d7-11e6-9a15-2c2131ca7afb.png)

I'm going to try and complete the discovery, and I'm not sure what happens once it completes, and whether or not we can re-create this...  To help rerun discovery, you can do the following: 

[...]

PR #2314, opened 2016-12-14
[CUSTOMER] Attempt to force sync time with ntpd instead of starting the ntp daemon to reduce discovery time

whowutwut:
Attempt to resolve #2327

See #2307 for details about this pull request.  It was reverted because `ntpdate` is deprecated, using ntpd command instead, but do not start the daemon, simply call ntpd and exit.   The customer would need to have ntp running to force sync the clocks, then force the hardware clock to sync with the system time. 

Output from the code changes
```
node-8335-gtb-100470a:[Wed Dec 14 10:56:46 2016]<166>Dec 12 14:40:24 xcat.genesis.doxcat: Attempting to sync time with using ntpd...
node-8335-gtb-100470a:[Wed Dec 14 10:56:47 2016]ntpd: time set +159381.409771s
node-8335-gtb-2109e4a:[Wed Dec 14 10:57:11 2016]<166>Dec 12 14:40:49 xcat.genesis.doxcat: Attempting to sync time with using ntpd...
node-8335-gtb-2109e4a:[Wed Dec 14 10:57:12 2016]ntpd: time set +159381.830421s
```

PR #2333, opened 2016-12-20
Fix the code of waiting for ntpd to synchronize in genessis

neo954:


zet809:
@whowutwut The pull request looks good to me, Pls help to review.
We may have a new genesis-base for customer to verify. And then we may need to build genesis-base based on CentOS for both ppc64le and x86_64.

whowutwut:
@neo954  can you add a description to the pull request and also can you show some unit test to make sure that the date before and after are in sync. We can set the date back to show that it's waiting before continuing. How long does the npt-wait take to sync up? Should be 1-2 seconds? What happens if the nptd does not sync. Then will it wait 15 min again until the next sync?

@zet809 ok I think we can build a new genesis but I DO NOT think we should run the build-deps script until we get closer to the 2.13.1 GA to not upload it.  otherwise we can potentially mess with the GA version

neo954:
@whowutwut,

The command `ntp-wait` is a tool designed for use in shell script. It wait for NTP server to synchronize. It is a 1st party tool, shipped in the ntp source tarball. You may see `man ntp-wait` for more details. And `ntp-wait` is a piece of perl code. You may also read the source by `cat /usr/sbin/ntp-wait`.

Different Linux disto shipped `ntp-wait` in different way. In RHEL, `ntp-wait` is packed in a separate package `ntp-perl`. In SLES and Ubuntu, `ntp-wait` is a part of package `ntp`.

zet809:
@neo954 and @whowutwut I have built a genesis-base to include ntp-wait and put it onto GSA, pls verify.

xCAT-genesis-scripts/bin/doxcat
```diff
@@ -266,19 +266,19 @@ ip addr show dev $bootnic|grep -v 'scope link'|grep -v 'dynamic'|grep -v  inet6|
 logger -s -t $log_label -p local4.info "Starting ntpd..." 
 ntpd -g -x
 
+logger -s -t $log_label -p local4.info "Waiting for ntpd to synchronize..."
+ntp-wait -n 1999 -s 1 -v
```
whowutwut: @neo954  The default for `ntp-wait` is `-n 1000` and `-s 6`, so 6000 seconds or 100 minutes.  So the concern here is that we are shorting the delay interval from 6 seconds to 1...  Does this command do anything that would impact the ntp server or simply checks locally on this machine whether or not ntp has been snychronized. 

Regardless, this doesn't at all address the issue of bringing down the 15 minute time... are we stating that there's essentially nothing we _should_ do to reduce this 15 min refresh configuration?

whowutwut:
@neo954  you should be able to edit the initial description on this PR to add more context. 

@neo954 @zet809  Regarding the code changes, I don't have any issues with the code change, but I still don't really like this solution because it doesn't address the initial problem reported in #2327 with the delay caused by the 15 minute interval between syncing of the clock from ntpd.    We still can't reliable re-create this in our own environment repeatably. 

Could we do any of the following, in addition to the code change in this PR... 
- Shorten the refresh time for `ntpd` from 15 minutes to something configurable that we could set in the xcat site table?  IT would default to 15 minutes, but we could set it down to 1 or 30 seconds
- Have any of the options to `ntp-wait` be configurable

This way if we do find that we have some problems with this code, we can adjust the parameters without the need to build a new genesis base and re-package or make manually changes to the genesis scripts...

xCAT-genesis-scripts/bin/doxcat
neo954: @whowutwut, The default `num-tries` is 100, and `interval` is 6. Red Hat changed the `num-tries` to 1000. And the implementation details of `ntp-wait` changed slightly across different `ntp` version. If you read the source and man page of `ntp-wait` across Red Hat, SLES and Ubuntu, you will find that. Thus, the main idea of the specified values used here, in the command line is to avoid the difference between different version of `ntp-wait`. Although it is used in the `genessis` environment, we can still get the benefit when we rebuild `genesis-base` from a different Linux distro.

And yes, `ntp-wait` simply query the `ntpd` on the same host. So there is no extra workload on the remote ntp server running in the xCAT management node. One second interval may save a little time.

neo954: Normally, `ntpd` cost tens of seconds to get synchronized with an upstream server. But, with the default configuration of xCAT generated `ntp.conf`, there is no upstream server and a local clock with stratum 10. It may cause `ntpd` cost up to 15 minute to get synchronized. It is just the natural of `ntpd`.

While the `ntpd` on the xCAT management node is unsynchronized, the `ntpd` on the xCAT compute node has to wait its upstream server to get synchronized first, then the `ntpd` on the xCAT compute node can get synchronized after that.

I do not think we can do anything about it, or make it fast. It is the natural of ntp. The only thing we can do here, is evaluate, leave the whole cluster unsynchronized, or wait 15 minutes, which is better. I believe leave the whole cluster unsynchronized is the worse case.

whowutwut:
OK, I think this is OK....

Is there a way to pass down some site configuration down to genesis kernel?  we can incorporate this change but if there is some real problem that we hit where many machines are taking 15 minutes to sync, the site can choose to use this method or set a configuration value to take the risk and just do a `ntpq` to force sync the clock in #2314.

whowutwut:
We discussed this in our interlock call so I'll merge the changes.  Thanks @neo954 ! 
@zet809  Do we need to re-create the genesis-base rpms for the 2.13.1 GA?  i didn't check the ones you created earlier in this thread.

Issue #2336, opened 2016-12-21
[FVT]:xcatprobe does not give dhcp information when mac attribute has several values.


tingtli:
xCAT 2.13.0
mac attribute is specified as follows.

```
[root@c910f03c05k23 ~]# lsdef -l c910f03c05k07
Object name: c910f03c05k07
    arch=ppc64le
    cons=kvm
    currchain=boot
    currstate=install sles12.1-ppc64le-compute
    groups=all
    initrd=xcat/osimage/sles12.1-ppc64le-install-compute/initrd
    installnic=eth0
    kcmdline=quiet autoyast=http://!myipfn!:80/install/autoinst/c910f03c05k07 install=http://!myipfn!:80/install/sles12.1/ppc64le/1 netdevice=eth0  Loghost=!myipfn! console=tty0 console=hvc0,115200 dhcptimeout=150
    kernel=xcat/osimage/sles12.1-ppc64le-install-compute/linux
    mac=42:08:0a:03:05:07!c910f03c05k07-pri|42:08:0a:03:05:11!c910f03c05k07-pub
    mgt=kvm
    netboot=grub2
    nichostnamesuffixes.eth1=-pub
    nichostnamesuffixes.eth0=-pri
    nicips.eth1=50.3.5.7
    nicips.eth0=10.3.5.7
    os=sles12.1
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    profile=compute
    provmethod=sles12.1-ppc64le-install-compute
    serialport=0
    serialspeed=115200
    status=powering-on
    statustime=12-21-2016 01:31:39
    updatestatus=synced
    updatestatustime=11-24-2015 04:24:31
    vmcpus=1
    vmhost=c910f03c05
    vmmemory=4096
    vmnicnicmodel=virtio
    vmnics=br10
    vmstorage=dir:///var/lib/libvirt/images/
[root@c910f03c05k23 ~]# nodeset c910f03c05k07 osimage=sles12.1-ppc64le-install-compute
c910f03c05k07: install sles12.1-ppc64le-compute
[root@c910f03c05k23 ~]# rpower c910f03c05k07 boot
c910f03c05k07: on reset
```
On another screen, check the status
```
[root@c910f03c05k23 subcmds]# xcatprobe -w osdeploy -n c910f03c05k07
The install NIC in current server is eth0                                                           [INFO]
All nodes to be deployed are valid                                                                  [ OK ]
[...]

PR #2339, opened 2016-12-21
fix issue 2336, get correct mac address

xuweibj:
#2336 

If mac is defined as below:
```
mac=42:08:0a:03:05:07!c910f03c05k07-pri|42:08:0a:03:05:11!c910f03c05k07-pub
```

get correct info.


hu-weihua:

Issue #2296, opened 2016-12-09
[DEV]bmcdiscovery return incorrect mtm and serial number for B&S servers


zet809:
The fru 0 of B&S server looks like this:
```
 Chassis Type          : Other
 Chassis Part Number   : 8001-12C
 Chassis Serial        : 130124A
 Board Mfg Date        : Sun Dec 31 19:00:00 1995
 Board Mfg             : IBM
 Board Product         : NONE
 Board Serial          : OM168S010286
 Board Part Number     : P8DTU
 Product Manufacturer  : IBM
 Product Name          : P8DTU-IBM-1
 Product Part Number   : SSP-6018U-TNR4T-IB001
 Product Version       : NONE
 Product Serial        : S239290X6A20331
 Product Asset Tag     : NONE
```
Our bmcdiscovery wrongly get Product Part Number and Product Serial as the mtm and serial:
```
bmcdiscover -s nmap --range 50.6.24.1 -z
node-ssp-6018u-tnr4t-ib001-s239290x6a20331:
	objtype=node
	groups=all
	bmc=50.6.24.1
	cons=ipmi
	mgt=ipmi
	mtm=SSP-6018U-TNR4T-IB001
	serial=S239290X6A20331
```

zet809:
@xuweibj Hi, xu wei, will you pls take a look at this issue? Thx!

xuweibj:
Verified, so close.

PR #2340, opened 2016-12-21
fix issue 2296, get correct mtms for B&S

xuweibj:
#2296 

If use command ``ipmitool-xcat fru print 0``, check "chassis part number" & "chassis serial" first. If has, set mtm as "chassis part number" and serial as "chassis serial".
If no, check "product part number" and "product serial".


zet809:
Looks good to me, thx a lot!

Issue #2337, opened 2016-12-21
[FVT]"/tftpboot/boot/grub2/grub.cfg-<MAC>" is not consistent with  "/tftpboot/boot/grub2/grub.cfg-<IP>" and "/tftpboot/boot/grub2/<node>" on "nodeset" 


tingtli:
xCAT 2.13.0 on rhels
How to reproduce
1. use xcatdebugmode default value "1". The provision works normally
2.  set xcatdebugmode to "2" in site table and the provision stops at following screen.
```
Loading Installation System (3/5) -      100%
Loading Installation System (4/5) -      100%
Loading Installation System (5/5) -      100%
starting syslogd (logging to /var/log/messages)... ok
starting klogd... ok
starting nscd... ok
Checking for missing server keys in /etc/ssh
ssh-keygen: generating new host keys: RSA1 RSA DSA ECDSA ED25519
Starting SSH daemon... ok
IP addresses:
  10.3.5.7

***  login using 'ssh -X root@10.3.5.7'  ***
***  run 'yast.ssh' to start the installation  ***
```
3. Set xcatdebugmode to "1" and run nodeset, then boot cn to reinstall the node, but the screen is still same with when debugmode is 2

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2341

immarvin:
"/tftpboot/boot/grub2/grub.cfg-<MAC>" is not consistent with  "/tftpboot/boot/grub2/grub.cfg-<IP>" and "/tftpboot/boot/grub2/<node>" on "nodeset" 

steps to reproduce:
````
chdef c910f03c05k07 mac="42:08:0a:03:05:07!c910f03c05k07-pri|42:08:0a:03:05:11!c910f03c05k07-pub"
chdef c910f03c05k07 installnic="42:08:0a:03:05:07" 
nodeset c910f03c05k07 osimage
chdef c910f03c05k07 installnic=eth0 
nodeset c910f03c05k07 osimage
diff /tftpboot/boot/grub2/grub.cfg-01-42-08-0a-03-05-07 /tftpboot/boot/grub2/c910f03c05k07
````

PR #2341, opened 2016-12-21
fix [FVT]set xcatdebugmode=1 does not work after xcatdebugmode is set to 2 #2337

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2337

 remove all the existing grub.cfg-xxx files while running ``nodeset``


immarvin:
updated

hu-weihua:
I have reviewed the change and agree to merge, thanks @immarvin .

Issue #2024, opened 2016-10-25
Time required to run nodeset


mattaezell:
We've noticed that running `nodeset` to set an `osimage` on multiple nodes can take several seconds.

Single node:

```
# time nodeset summitdev-c0r0n01 osimage=summitdev-compute-rhels7.2 > /dev/null
real    0m1.402s
user    0m0.103s
sys     0m0.003s
```

18 Nodes:

```
# time nodeset c0r0 osimage=summitdev-compute-rhels7.2 > /dev/null
real    0m5.713s
user    0m0.110s
sys     0m0.004s
```

36 Nodes:

```
# time nodeset compute osimage=summitdev-compute-rhels7.2 > /dev/null
real    0m10.648s
user    0m0.118s
sys     0m0.004s
```

I'm worried that when we scale up to the full CORAL system this will take way too long.  What can we do to improve this?

daniceexi:
@chenglch could you take a look why the time cost is in linear increase?

immarvin:
``nodeset <node> osimage=xxx`` will run "makedhcp -a <node>", the time cost versus node number of ``makedhcp`` confirms to the linear increase since  ``makedhcp  -a <node>`` adds the dhcp lease per node in the following steps:
1) resolve the hostnames of the node to ip address
2) add dhcp leases for the node  with omshell

daniceexi:
@immarvin If the call of ``makedhcp`` could not be skipped, we could try to add an option for **nodeset** cmd to avoid of running ``makedhcp`` if user knows that my change does not impact the dhcp cfg.

mattaezell:
I think resolving the hostnames to IPs are pretty quick because we have them in the hosts file.  It looks like about half the time is in `mknetboot`.  There's a really long `foreach my $node (@nodes)` loop, that I wonder if some of the common elements could be pulled outside the loop to save some time.
The other half of the time is in dhcp; it looks like the `dhcp` plugin is called once per node.  Would it be possible to instead call it once with the noderange provided to `nodeset`?  Would that provide opportunities for the `dhcp` plugin to streamline things - like open a single `omshell` and pipe in multiple commands?

chenglch:
<img width="1197" alt="nodeset" src="https://cloud.githubusercontent.com/assets/7165199/20514026/a9fa3424-b0c4-11e6-8f35-45d7d955aab2.png">

[...]

PR #2351, opened 2016-12-23
Do not update bootparams table when running nodeset

chenglch:
With the nytprof tool, the buttleneck is the time to update the bootparams
table.To optimize the performance on large scale nodes, this patch tranfer
the bootparams hash through variable reference.

posible impact: getdestiny can not get the information about bootparams.

partial-issue: #2024

### DIFF
--- a/xCAT-server/lib/xcat/plugins/anaconda.pm
+++ b/xCAT-server/lib/xcat/plugins/anaconda.pm
@@ -172,6 +172,7 @@ sub mknetboot
     if ($req->{command}->[0] =~ 'mkstatelite') {
         $statelite = "true";
     }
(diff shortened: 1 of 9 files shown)


immarvin:
agree to merge to start verification ASAP

immarvin:
hi @whowutwut , would you please add the auto daily build of this branch in the build server?

Obihoernchen:
Now new nodes don't have the following entries
```
kcmdline
kernel?
initrd
```
only addkcmdline.

Are kcmdline, kernel and initrd obsolete now?

chenglch:
@Obihoernchen , yes, they will not be updated.

PR #2356, opened 2016-12-29
update xcat control file

SStar1314:
The require deb package(xcat-genesis-scripts) will make xcat failed to install on ubuntu16.04. The reason is the deb package name is already changed.


SStar1314:
Hi, pls accept this check-in. Thanks.

PR #2357, opened 2016-12-29
add makentp case to synchronize time with external ntp server

tingtli:
@neo954 please review. Thanks!

### DIFF
--- a/xCAT-test/autotest/testcase/makentp/cases0
+++ b/xCAT-test/autotest/testcase/makentp/cases0
@@ -114,11 +114,8 @@ check:output=~ntp servers are: $$extntpserversip
 check:output=~stopping ntpserver
 check:output=~starting ntpserver
 check:output=~enabling ntpserver
-cmd:if [ -f "/usr/sbin/ntp-wait" ];then ntp-wait; else sleep 1200; fi
+cmd:if [[ "__GETNODEATTR($$CN,os)__" =~ "ubuntu" ]] || [[ "__GETNODEATTR($$CN,os)__" = "sles11.4" ]];then service ntp status;else service ntpd status; fi
 check:rc==0
-cmd:ntpq -p
-check:output=~*$$extntpserversip
-cmd:grep $$extntpserversip /etc/ntpd.conf
+cmd:a=0;while ! `ntpq -n -c peers | awk '{print \$1}'|grep "*" >/dev/null`; do sleep 30;((a++));if [ $a -gt 30 ];then exit 1;fi done
 check:rc==0
 end
-
--- a/xCAT-test/autotest/testcase/makentp/cases0
+++ b/xCAT-test/autotest/testcase/makentp/cases0
@@ -116,6 +116,6 @@ check:output=~starting ntpserver
 check:output=~enabling ntpserver
 cmd:if [[ "__GETNODEATTR($$CN,os)__" =~ "ubuntu" ]] || [[ "__GETNODEATTR($$CN,os)__" = "sles11.4" ]];then service ntp status;else service ntpd status; fi
 check:rc==0
-cmd:a=0;while ! `ntpq -n -c peers | awk '{print \$1}'|grep "*" >/dev/null`; do sleep 30;((a++));if [ $a -gt 30 ];then exit 1;fi done
+cmd:a=0;until ntpq -n -c peers | tail -n +2 | grep -q '^*'; do sleep 30;((a++));if [ $a -gt 30 ];then exit 1;fi done 
 check:rc==0
 end
--- a/xCAT-test/autotest/testcase/makentp/cases0
+++ b/xCAT-test/autotest/testcase/makentp/cases0
@@ -100,3 +100,25 @@ check:rc==0
 cmd:rm -rf /tmp/postscript
 end
 
+
+start:makentp_initialize
+descriptiop:start ntp service on management node and synchronize time with external ntp server
+cmd:chtab key=extntpservers site.value="$$extntpserversip"
+check:rc==0
+cmd:makentp -V
+check:rc==0
+check:output=~configuring management node: $$MN
+check:output=~checking if nptd is installed
+check:output=~backing up the ntp configuration file /etc/ntp.conf
+check:output=~ntp servers are: $$extntpserversip
+check:output=~stopping ntpserver
+check:output=~starting ntpserver
+check:output=~enabling ntpserver
+cmd:if [ -f "/usr/sbin/ntp-wait" ];then ntp-wait; else sleep 1200; fi
+check:rc==0
+cmd:ntpq -p
+check:output=~*$$extntpserversip
+cmd:grep $$extntpserversip /etc/ntpd.conf
+check:rc==0
+end
+


xCAT-test/autotest/testcase/makentp/cases0
```diff
@@ -100,3 +100,25 @@ check:rc==0
...
+check:output=~configuring management node: $$MN
+check:output=~checking if nptd is installed
+check:output=~backing up the ntp configuration file /etc/ntp.conf
+check:output=~ntp servers are: $$extntpserversip
+check:output=~stopping ntpserver
+check:output=~starting ntpserver
+check:output=~enabling ntpserver
+cmd:if [ -f "/usr/sbin/ntp-wait" ];then ntp-wait; else sleep 1200; fi
+check:rc==0
+cmd:ntpq -p
```
neo954: No need to check the output of `ntpq` anymore, since `ntp-wait` was already used.

xCAT-test/autotest/testcase/makentp/cases0
```diff
@@ -100,3 +100,25 @@ check:rc==0
...
+check:output=~*$$extntpserversip
```
neo954: Since `$extntpserversip` is a comma separated list of IP addresses, you cannot check the output just like this.

xCAT-test/autotest/testcase/makentp/cases0
```diff
@@ -100,3 +100,25 @@ check:rc==0
...
+cmd:grep $$extntpserversip /etc/ntpd.conf
```
neo954: Since $extntpserversip is a comma separated list of IP addresses, you cannot do the checking of file `/etc/ntp.conf` just like this.

xCAT-test/autotest/testcase/makentp/cases0
```diff
@@ -100,3 +100,25 @@ check:rc==0
...
+check:rc==0
+cmd:makentp -V
+check:rc==0
+check:output=~configuring management node: $$MN
+check:output=~checking if nptd is installed
+check:output=~backing up the ntp configuration file /etc/ntp.conf
+check:output=~ntp servers are: $$extntpserversip
+check:output=~stopping ntpserver
+check:output=~starting ntpserver
+check:output=~enabling ntpserver
```
neo954: I believe checking the exit code of `makentp` is enough. No need to check all the detail outputs, in case we make some change to the script `makentp` some time in the future.

neo954: Need to check if the daemon `ntpd` is running.

xCAT-test/autotest/testcase/makentp/cases0
```diff
@@ -100,3 +100,25 @@ check:rc==0
 cmd:rm -rf /tmp/postscript
 end
 
+
+start:makentp_initialize
+descriptiop:start ntp service on management node and synchronize time with external ntp server
+cmd:chtab key=extntpservers site.value="$$extntpserversip"
```
neo954: I prefer use `chdef` to change the attribute.

neo954: And where does the variable `$extntpserversip` come from?

xCAT-test/autotest/testcase/makentp/cases0
```diff
@@ -100,3 +100,25 @@ check:rc==0
...
+cmd:makentp -V
+check:rc==0
+check:output=~configuring management node: $$MN
+check:output=~checking if nptd is installed
+check:output=~backing up the ntp configuration file /etc/ntp.conf
+check:output=~ntp servers are: $$extntpserversip
+check:output=~stopping ntpserver
+check:output=~starting ntpserver
+check:output=~enabling ntpserver
+cmd:if [ -f "/usr/sbin/ntp-wait" ];then ntp-wait; else sleep 1200; fi
```
neo954: By default, `ntp-wait` will wait for maximum 600 seconds on SLES and Ubuntu, or 6000 seconds on Red Hat.

Without `ntp-wait`, just waiting 1200 seconds here is nonsense. You can check the output of `ntpq -n -c 'rv 0'` here, just like what `ntp-wait` does.

xCAT-test/autotest/testcase/makentp/cases0
tingtli: It comes from configuration files default.conf

xCAT-test/autotest/testcase/makentp/cases0
tingtli: I think we can put the output here. If we make updates to the makentp command, we can update .

xCAT-test/autotest/testcase/makentp/cases0
tingtli: I think for this case, our goal is to synchronize mn with external server and setup ntp service on the mn. For the complicated configuration, let's add another case.

neo954: Just check the output of `ntpq -n -c peers' from line 3 to the end, and see if there is a line begin with the asterisk character, `*`. Other checking is irrationality.

xCAT-test/autotest/testcase/makentp/cases0
```diff
@@ -100,3 +100,22 @@ check:rc==0
 cmd:rm -rf /tmp/postscript
 end
 
+
+start:makentp_initialize
+descriptiop:start ntp service on management node and synchronize time with external ntp server
```
hu-weihua: I think this case should be a ``stop`` case.

xCAT-test/autotest/testcase/makentp/cases0
```diff
@@ -100,3 +100,22 @@ check:rc==0
...
+check:output=~configuring management node: $$MN
+check:output=~checking if nptd is installed
+check:output=~backing up the ntp configuration file /etc/ntp.conf
+check:output=~ntp servers are: $$extntpserversip
+check:output=~stopping ntpserver
+check:output=~starting ntpserver
+check:output=~enabling ntpserver
+cmd:if [[ "__GETNODEATTR($$CN,os)__" =~ "ubuntu" ]] || [[ "__GETNODEATTR($$CN,os)__" = "sles11.4" ]];then service ntp status;else service ntpd status; fi
+check:rc==0
+cmd:a=0;while ! `ntpq -n -c peers | awk '{print \$1}'|grep "*" >/dev/null`; do sleep 30;((a++));if [ $a -gt 30 ];then exit 1;fi done
```
neo954: Do the check from the 3rd line to the end, and see if there is a asterisk character in the first column of the rest lines.
You may use `ntpq -n -c peers | tail -n +2 | grep -q '^*'` and check the `rc`.
Or you may use `ntpq -n -c peers | awk '/^*/ { if ( NR > 2 ) print ; }'` and check if there are any output.

neo954: @tingtli, your code is not working at all
``while ! `cmd ...`; do ... done`` simply will not work. The output of `ntpq -n -c peers | awk '{print \$1}'|grep "*" >/dev/null` is a mess. It is not a command. And you put the outputs between backquotes. It simply will not run.

neo954:
Cool. I will merge this one.

PR #2359, opened 2017-01-03
remove setupntp for sles12.1 ntpd doesn't synchronize to local clock …

tingtli:
…anymore in version 4.2.8p4. see#2248

### DIFF
--- a/xCAT-test/autotest/testcase/installation/Full_installation_flat_p8_le
+++ b/xCAT-test/autotest/testcase/installation/Full_installation_flat_p8_le
@@ -25,7 +25,8 @@ cmd:makeconservercf $$CN
 check:rc==0
 cmd:cat /etc/conserver.cf | grep $$CN
 check:output=~$$CN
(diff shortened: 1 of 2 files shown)


immarvin:
agree to merge

whowutwut:
@tingtli  @immarvin   I just caught this....  this pull request branch is against 2.13, this is incorrect.  we should not be creating changes against 2.13 branch but against master branch, then we merge master -> release when we are creating the RC builds.

whowutwut:
I think it might be OK because there seems to be no conflicts with merging master to 2.13, but your changes might NOT be in the master branch....

tingtli:
@whowutwut  This commit is also in master branch. After I check in the code in master branch, i cherry pick this commit to 2.13 branch. The commit number is 8ec1af2f0c07e09cad8644708b04a18d9c099e5d.

The reason I need to check this change to 2.13 is because FVT also run some tests against 2.13 branch to fix some test framework stability problems.

Issue #2281, opened 2016-12-07
[DEV]"updatenode -P" does not return the correct return code 


immarvin:
````
[root@c910f03c17k27 ~]# cat /install/postscripts/setupntp1
exit 1
[root@c910f03c17k27 ~]# updatenode c910f03c17k28 -P setupntp1
c910f03c17k28: vpdupdate is not supported on the pSeries KVM Guest platform
c910f03c17k28: xcatdsklspost: updating VPD database
c910f03c17k28: xcatdsklspost: downloaded postscripts successfully
c910f03c17k28: Wed Dec  7 01:54:47 EST 2016 Running postscript: setupntp1
c910f03c17k28: postscript: setupntp1 exited with code 1
c910f03c17k28: Running of postscripts has completed.
[root@c910f03c17k27 ~]# echo $?
0
[root@c910f03c17k27 ~]# lsxcatd -v
Version 2.13.0 (git commit 99da40c22a1b442a60eb346ee337442d155df2fb, built Tue Dec  6 05:00:44 EST 2016)
[root@c910f03c17k27 ~]#
````

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2360

PR #2360, opened 2017-01-03
return error code when updatenode command completed

chenglch:
close-issue: #2281


immarvin:
i am ok with the change

PR #2363, opened 2017-01-04
update sles and redhat migration test case latest version check method

caomengmeng:
@tingtli 
Please help to review.


tingtli:
I agree to merge

whowutwut:
@tingtli  @caomengmeng I just caught this....  this pull request branch is against 2.13, this is incorrect.  we should not be creating changes against 2.13 branch but against master branch, then we merge master -> release when we are creating the RC builds.

whowutwut:
I think it might be OK because there seems to be no conflicts with merging master to 2.13, but your changes might NOT be in the master branch....

Issue #1567, opened 2016-07-21
[DEV] Enhance switch base hardware discovery output error message 


zet809:
In customer's environment, there is a input error when setting switch and switchport attributes for a predefined node(one '=' more write in switchport attribute), then it cause our switch-based hardware discovery failed to find a matching node.
So, the idea is, we can output a message like "Didn't find matching node, the mac can be found at switch1 and port-xxxx", then it will be much helpful to know what happened.

zet809:
Talking with @daniceexi shows that, if we want to return more helpful message, the performance of MacMap.pm will be a problem. Maybe we can add an option "--cache" for "xcatprobe switch-macmap", and call it within "xcatprobe discovery".
@whowutwut Do you have any idea?

zet809:
Move to next release to rethink the solution.

zet809:
Seems it can be included in xcatprobe discovery subcommand later.

zet809:
Add @hu-weihua and @xuweibj in this discussion.

zet809:
Split this issue into 2 parts:
1. for the parameter error, can be included in xcatprobe discovery subcommand
2. for the not matching MAC address, may need to work with xcatprobe switch-macmap in Ansible process.
So, in this defect, we will focus on the 1st issue, and @xuweibj  will you pls open another defect to trace the 2nd issue? Thx a lot!

zet809:
I think we can close this defect, there is another defect opened to trace the 2nd scenario that "for the not matching MAC address, may need to work with xcatprobe switch-macmap in Ansible process."

PR #2365, opened 2017-01-04
Fix issue 1567, check attribute switch & switchport

xuweibj:
#1567 
If attribute switch or switchport is not began with number or letter, show error message.

### DIFF
--- a/xCAT-probe/subcmds/discovery
+++ b/xCAT-probe/subcmds/discovery
@@ -315,10 +315,20 @@ sub check_pre_defined_node {
         my $keystring;
         foreach my $node (keys %nodecheckrst) {
             {
+                last if ($nodecheckrst{$node}{"nodetype"} eq "switch");
+
+                my @error_attribute = ();
                 $keystring = "";
-                if (!(exists($nodecheckrst{$node}{"switch"}) && exists($nodecheckrst{$node}{"switchport"}))) {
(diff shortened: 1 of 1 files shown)


xCAT-probe/subcmds/discovery
```diff
@@ -320,6 +320,12 @@ sub check_pre_defined_node {
                     last if ($nodecheckrst{$node}{"nodetype"} eq "switch");
                     $keystring = "Atrribute 'switch' or 'switchport' isn't defined for '$discovery_type' type discovery";
                     last;
+                } elsif ($nodecheckrst{$node}{"switch"} !~ /^\w/) {
```
zet809: A problem maybe, if both switch and switchport attributes are wrongly configured, only error against switch attribute can be reported.

xCAT-probe/subcmds/discovery
```diff
@@ -315,10 +315,20 @@ sub check_pre_defined_node {
         my $keystring;
         foreach my $node (keys %nodecheckrst) {
             {
+                last if ($nodecheckrst{$node}{"nodetype"} eq "switch");
+
+                my @error_arr = ();
```
zet809: error_attr will be more make sense.

zet809:
@xuweibj I agree to merge, thx a lot.

PR #2369, opened 2017-01-05
fix 2090 delete underlying files for ro statelite mounts

### LINKED ISSUE #2090: delete underlying files for ro statelite mounts
This depends on the suggestions in #2089.

When one specifies ro children under a tmpfs parent, e.g.:

```
"ALL","/var/","tmpfs",,
"ALL","/var/cache/","ro",,
"ALL","/var/lib/rpm/","ro",,
"ALL","/var/lib/texmf/","ro",,
```
the contents of the child in the tmpfs mount are only removed if the path in tmpfs doesn't match the expected file type (https://github.com/xcat2/xcat-core/blob/master/xCAT-server/share/xcat/netboot/add-on/statelite/rc.statelite#L462):

```
            if [ "${CHAREND}" = "/" ]; then # it is one directory
                if [ ! -d ${STRPATH} ]; then
                    /bin/rm -rf ${STRPATH}
                    /bin/mkdir ${STRPATH}
                fi
            else
                if [ ! -f ${STRPATH} ]; then
                    /bin/rm -rf ${STRPATH}
                    /bin/touch ${STRPATH}
                fi
            fi
```

by removing the file type tests around the `rm`, `mkdir`, and `touch` we can free up the space that would be occupied by those files/directory trees in tmpfs which depending on the amount of RAM available and the size of those trees could yield a significant increase in usable RAM.

-Aaron

bybai:
fix #2090 

After statelite compute node is up, STRPATH will be overwriten by mount bind.  If the old STRPATH contains files, it occupies the RAMm, deleting these 2 dir can free RAM.

### DIFF
--- a/xCAT-server/share/xcat/netboot/add-on/statelite/rc.statelite
+++ b/xCAT-server/share/xcat/netboot/add-on/statelite/rc.statelite
@@ -460,15 +460,11 @@ ProcessType () {
             CHAREND=`echo ${STRPATH} | /usr/bin/cut -c${STRLEN}`            
 
             if [ "${CHAREND}" = "/" ]; then # it is one directory
-                if [ ! -d ${STRPATH} ]; then
                     /bin/rm -rf ${STRPATH}
                     /bin/mkdir ${STRPATH}
-                fi
             else
-                if [ ! -f ${STRPATH} ]; then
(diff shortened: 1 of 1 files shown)


immarvin:
agree to merge

Issue #2371, opened 2017-01-06
[FVT]:enhancement: cumulus switch could not be configured correctly if it uses the default onie switch type


tingtli:
Following Cumulus documentation. Use following command to create a switch definition
```
[root@fs4 ~]# mkdef edgecoresw1 --template onieswitch arch=armv71 ip=192.168.5.191 mac=8C:EA:1B:12:CA:40
1 object definitions have been created or modified.

[root@fs4 ~]# lsdef -l edgecoresw1
Object name: edgecoresw1
    arch=armv7
    groups=switch
    ip=192.168.23.1
    mac=8C:EA:1B:12:CA:40
    mgt=switch
    netboot=onie      
    nodetype=switch
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    provmethod=/install/custom/sw/edgecore/cumulus-linux-3.1.0-bcm-armel-1471981017.dc7e2adzfb43f6b.bin
    switchtype=onie                     <== here default switchtype is onie
    usercomment=the template for cumulus switch definition

```
Use script configucumulus to config the ssh for root and get the switch type error

```
[root@fs4 ~]# /opt/xcat/share/xcat/scripts/configcumulus --switches edgecoresw1 --ssh
The edgecoresw1 is not cumulus switch, will not config
No Valid Switch to process
[root@fs4 ~]# lsdef -l edgecoresw1
Object name: edgecoresw1
    arch=armv7
    groups=switch
    ip=192.168.23.1
    mac=8C:EA:1B:12:CA:40
    mgt=switch
    netboot=onie
    nodetype=switch
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    provmethod=/install/custom/sw/edgecore/cumulus-linux-3.1.0-bcm-armel-1471981017.dc7e2adzfb43f6b.bin
    switchtype=onie
    usercomment=the template for cumulus switch definition
```

Check the script, it should match cumulus as the switchtype
```
     75     foreach my $fsw (@filternodes)  {
     76         if (($switches_hash->{$fsw}->[0]->{switchtype}) =~ /cumulus/) {
     77             push @nodes, $fsw;
     78         } else {
     79             xCAT::MsgUtils->message("E","The $fsw is not cumulus switch, will not config");
     80         }
     81     }
```
Modify the switch type to cumulus and the ssh could be configured.
```
[root@fs4 ~]# chdef edgecoresw1 switchtype=cumulus
[...]

PR #2375, opened 2017-01-06
fix issue https://github.com/xcat2/xcat-core/issues/2371

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2371


tingtli:
i agree to merge

PR #2379, opened 2017-01-09
Running dist-upgrade for not upgraded packages on diskless Ubuntu images

cxhong:
For issue #71 ,  on Ubuntu genimage, add dist-upgrade after running apt-get update and upgrade in case some packages didn't get upgraded.


immarvin:
agree to merge

Issue #2373, opened 2017-01-06
[FVT]:enhancement: give correct postscript and postbootscript when the template is  onieswitch


tingtli:
Use --template to create a switch template and get following definition
```
[root@fs4 install]# mkdef testswitch --template onieswitch
1 object definitions have been created or modified.

[root@fs4 install]# lsdef -l testswitch
Object name: testswitch
    groups=switch
    mgt=switch
    netboot=onie
    nodetype=switch
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    switchtype=onie
    usercomment=the template for cumulus switch definition
```
Since we don't support switch to use otherpkgs and syslog,remoteshell and syncfiles at the moment. How about make the value blank. then if users use the definition, they could not wondering why they need to remove it.

immarvin:
fix in https://github.com/xcat2/xcat-core/pull/2381

tingtli:
Fixed and verified on xCAT 2.13.1 GA release

PR #2381, opened 2017-01-10
support cumulus os in syslog postscript

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2373

support cumulus in syslog
exit with "not support" message in other default postscripts


tingtli:
A general comments is should we also put the warning message in xcat.log?

immarvin:
updated

tingtli:
I agree to merge.

Issue #2384, opened 2017-01-10
[fvt]2.13.1 xcatprobe xcatmn need modify firefall and diskspace check's output 


junxiawang:
env:rhels7.3 ppc64le
build:
 lsxcatd -v
Version 2.13.1 (git commit bce3ec37e0762a1f9dbe8828c10ef7dac80c51a4, built Fri Jan  6 06:16:18 EST 2017)
How to reproduce:
```
 xcatprobe xcatmn -i eth0
...
[mn]: Checking NTP service is configured...                                                                       [ OK ]
[mn]: Checking firewall is disabled                                                                               [ OK ]
[mn]: Checking minimum disk space for xCAT ['/var' needs 1GB;'/install' needs 10GB;'/tmp' needs 1GB]              [ OK ]
----------------> Need to add three points for the end of firewall and diskpace probe's output 
...

```

hu-weihua:
@junxiawang , Could you verify it? thanks

PR #2386, opened 2017-01-10
fix issue#2384 2.13.1 xcatprobe xcatmn need modify firefall and diskspace check's output

hu-weihua:
Fix issue #2384 2.13.1 xcatprobe xcatmn need modify firefall and diskspace check's output"


xuweibj:
I have reviewed the code and agree to merge.

PR #2155, opened 2016-11-15
refine the cumulus support documentation

immarvin:
1. refine the cumulus support documentation
2. add a object template for cumulus switch

See http://10.3.5.21/xcat-doc/html/advanced/networks/edgecore_switches/index.html

PR #2395, opened 2017-01-10
Change switchtype from cumulus to onie

cxhong:
The switch type changed from cumulus to onie on PR #2155,  corresponding config file needs to change also because it based on switch type.
PR #2155  and issue #2371


immarvin:
agree to merge

Issue #2392, opened 2017-01-10
Unsupported character in bootparams.addkcmdline


Phhere:
It is not possible to use the character "," in bootparams.addkcmdline. It alsways gets replaces by a space in elilo files.
I need it for UV2000 to set some console parameters like 
`earlyprintk=ttyS0,115200n8` console=ttyS0,115200n`

Affected Version:
Version 2.11.1 (git commit b6e4c8ca1847bf574d7e86687d74a7f2576d428a, built Thu Apr 14 09:32:27 EDT 2016)

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2397

immarvin:
hi @Phhere ,would you please verify the fix? thanks

whowutwut:
@tingtli  Can you help find someone in FVT to validate this fix so we can close the issue?

tingtli:
@junxiawang Please help verify this bug. Thanks!

PR #2397, opened 2017-01-11
fix issue Unsupported character in bootparams.addkcmdline #2392

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2392


hu-weihua:

PR #2400, opened 2017-01-11
[FVT]Update installation cases with rinstall command

caomengmeng:
@tingtli 
Please help to review code


tingtli:
@caomengmeng  i am fine with the change. One comment is we also need to update the migration cases, please search all the cases and make sure we updates.

caomengmeng:
@tingtli
please help to review files with latest update for rinstall

tingtli:
i don't have further comments. @immarvin  How about your opinion?

immarvin:
i am ok with the change

Issue #2396, opened 2017-01-11
[FVT]:configcumulus --switches edgecoresw1 --all could not find the license file


tingtli:
xCAT 2.13

This is not included in the cumulus documentation. I just run some test for the script. It is a nice to have function. So, i open this in 2.13.2. 

```
[root@fs4 /]# /opt/xcat/share/xcat/scripts/configcumulus --switches edgecoresw1 --all
/root/.ssh/known_hosts updated.
Original contents retained as /root/.ssh/known_hosts.old
ssh configured for edgecoresw1
install_license
file =
 is not exist
start to config_snmp
echo 'US/Eastern'>/etc/timezone;dpkg-reconfigure --frontend noninteractive tzdata
xdsh edgecoresw1 echo 'US/Eastern'>/etc/timezone;dpkg-reconfigure --frontend noninteractive tzdata
```
Take a look at the help information.
```
[root@fs4 ~]# /opt/xcat/share/xcat/scripts/configcumulus -h
Usage:
    configcumulus -h│--help
    configcumulus --switches switchnames --ssh
    configcumulus --switches switchnames --license filename
    configcumulus --switches switchnames --snmp [--user snmp_user] [--password snmp_password] [--group snmp_group]
    configcumulus --switches switchnames --ntp

    To set ssh, install license(license file: /root/license.txt), config snmp and ntp:
        configcumulus --switches switchnames --all

[root@fs4 ~]# ls /root/license.txt
/root/license.txt
```

immarvin:
only the ``configcumulus --switches switchnames --ssh`` is the documented option, the rest options should be supported by postsctipts,  is it ok to remove the other options? @cxhong

cxhong:
I like to keep those options around until cumulus switch process is well defined.

tingtli:
@cxhong @immarvin  I could not find /opt/xcat/share/xcat/scripts/configcumulus in today's build, why?

immarvin:
renamed to configonie see https://github.com/xcat2/xcat-core/commit/f843dcad3cc9484674caa4d75a877ce333f23226

cxhong:
@tingtli ,  yes,  it changed to /opt/xcat/share/xcat/scripts/configonie at PR #2395

tingtli:
I have review the defect and tried the verification on fs4. Since edgecoresw1 is down. The test failed, but the option is realized. Close this bug.

PR #2402, opened 2017-01-11
Add default license file for configonie command

cxhong:
For issue #2396 ,

the default license file will be /root/license.txt

### DIFF
--- a/xCAT-server/share/xcat/scripts/configonie
+++ b/xCAT-server/share/xcat/scripts/configonie
@@ -217,7 +217,7 @@ sub cumulus_exec {
 sub install_license {
     my @config_switches;
     print "install_license\n";
-    my $license_file;
+    my $license_file = "/root/license.txt";
     my $file_name = "/root/license.txt";
 
     if ($::LICENSE) {
@@ -226,7 +226,7 @@ sub install_license {
 
     print "file = $license_file\n";
     if (!(-e $license_file) ) {
-        print "$license_file is not exist\n";
+        print "license file $license_file is not exist\n";
         return;
     }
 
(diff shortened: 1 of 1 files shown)


xCAT-server/share/xcat/scripts/configonie
```diff
@@ -217,7 +217,7 @@ sub cumulus_exec {
 sub install_license {
     my @config_switches;
     print "install_license\n";
-    my $license_file;
+    my $license_file = "/root/license.txt";
```
whowutwut: What's the difference between this and file name below?

xCAT-server/share/xcat/scripts/configonie
```diff
@@ -226,7 +226,7 @@ sub install_license {
 
     print "file = $license_file\n";
     if (!(-e $license_file) ) {
-        print "$license_file is not exist\n";
+        print "license file $license_file is not exist\n";
```
whowutwut: Suggest change to "does not exist"  instead of is not

xCAT-server/share/xcat/scripts/configonie
cxhong: License_file is located in the MN, and can be changed with "-f" options.  filename is located on the switch.

xCAT-server/share/xcat/scripts/configonie
cxhong: updated

immarvin:
hi @cxhong , @whowutwut , according to the documentation http://10.3.5.21/xcat-doc/html/advanced/networks/edgecore_switches/onie_switches.html, the licence file will be installed via ``xdcp+xdsh``,  I think we'd better deprecate the ``--license`` option

cxhong:
I like to keep this option around until cumulus switch is well defined.

immarvin:
agree to merge

Issue #2391, opened 2017-01-10
[fvt]2.13.1 xcatprobe xcatmn need modify the dhcp's probe result's format


junxiawang:
env:rhels7.3 ppc64le
build:
Version 2.13.1 (git commit bce3ec37e0762a1f9dbe8828c10ef7dac80c51a4, built Fri Jan  6 06:16:18 EST 2017)
How to reproduce:
```
xcatprobe xcatmn -i eth0:
...
[mn]: Checking 'site' table is configured...                                                                      [ OK ]
[mn]: Checking provision network is configured...                                                                 [ OK ]
[mn]: Checking 'passwd' table is configured...                                                                    [ OK ]
[mn]: Checking important directories(installdir,tftpdir) are configured...                                        [ OK ]
[mn]: Checking SELinux is disabled...                                                                             [ OK ]
[mn]: Checking HTTP service is configured...                                                                      [ OK ]
[mn]: Checking TFTP service is configured...                                                                      [ OK ]
[mn]: Checking DNS service is configured...                                                                       [ OK ]
 Error: dhcp server is not running.  please start the dhcp server.                                                [mn]: Checking DHCP service is configured...                                                                      [FAIL]
------------------>The output here need modified.
```

PR #2403, opened 2017-01-12
fix issue 2391, modify probe result format

xuweibj:
#2391


hu-weihua:

PR #2405, opened 2017-01-12
[go-xcat] Fix a bug of the dirty hack for Debian/Ubuntu `devel' branch

neo954:


immarvin:
i am ok with the change, agree to merge 

thanks @neo954

PR #2411, opened 2017-01-13
get latest from xcat2/xcat-core

zVMopenstack:
Get the latest code from xCAT team


whowutwut:
I don't think this PR is correct, talking to @chuckbrazie  offline.  There's no changes attempting to be pushed to the project.  I think this is an attempt to sync up the forked copy ... so the PR should be in the forked copy, not in the upstream main xcat2 project.

PR #2419, opened 2017-01-18
Quote user and password in ipmitool

jjohnson42:
If a value had certain characters, the command would be incorrect.
For example, a password with a '#' in it would break.


zet809:

PR #2420, opened 2017-01-18
Fix rinv support for dimm

jjohnson42:


zet809:
Looks good to me, I agree to merge. Some verification shows that 'dimm' option only works for system X servers, will you pls update the manpage and usage of rinv for BMC/MPA specific servers? Thx a lot

Issue #2139, opened 2016-11-10
vmstoragemodel doesn't work for VMs on ppc64le hypervisor


zet809:
When using vmstoragemodel to specify the stroagemodel for a diskfile, it will be replaced by 'scsi' for a ppc64 or ppc64le hypervisor. 

```
 # lsdef testvm -i vmstoragemodel
 Object name: testvm
    vmstoragemodel=virtio
```

```
<disk type='file' device='disk'>
      <driver name='qemu' type='qcow2' cache='none'/>
      <source file='/var/lib/libvirt/images/testvm.sda.qcow2'/>
      <backingStore/>
      <target dev='sda' bus='scsi'/>             ==> Shall be dev='vda' and bus='virtio'
      <alias name='scsi0-0-0-0'/>
      <address type='drive' controller='0' bus='0' target='0' unit='0'/>
    </disk>
```
It is wrongly modified within our code when it detect the hypervisor is a ppc64/ppc64le model. Changing to 'ppc64$' only will work.
```
        if ($confdata->{$hyp}->{cpumodel} and $confdata->{$hyp}->{cpumodel} =~ /ppc64/i) {
            $confdata->{vm}->{$node}->[0]->{storagemodel} = "scsi";
        }
```
It is approved that 'virtio' also works for rhev4.x on ppc64le hypervisor, so need more test for vm related functions.

gurevichmark:
@zet809 I am a little confused by this issue. Did you already fix it with the code you identified above and just some testing needed, OR code needs to be changed as you identified above ?

I can not recreate this problem running on the latest xCAT build from Nov 14. When I create a VM with `vmstoragemodel=virtio` the disk information in the `kvm_nodedat`a does not have any `bus` setting.

```
[root@fs2vm100 gurevich]# lsdef mgXgeneratedvm1
Object name: mgXgeneratedvm1
    arch=ppc64le
    cons=kvm
    currchain=boot
    currstate=install rhels7.2-ppc64le-compute
    groups=vm,all
    initrd=xcat/osimage/rhels7.2-ppc64le-install-compute/initrd.img
    ip=10.3.31.215
    kcmdline=quiet inst.repo=http://!myipfn!:80/install/rhels7.2/ppc64le inst.ks=http://!myipfn!:80/install/autoinst/mgXgeneratedvm1 BOOTIF=42:d1:0a:03:1f:d7  ip=dhcp  inst.cmdline  console=tty0 console=hvc0,115200
[...]

PR #2421, opened 2017-01-18
Use correct prefix for disks when vmstoragemodel=virtio

gurevichmark:
This pull request fixes issue #2139.

If `vmstoragemodel` attribute of node definition is not specified, default to `scsi`.
If `vmstoragemodel=virtios` disk prefix should be `vd*`, CDROM prefix `sd*`


zet809:
@gurevichmark Hi, mark, the modification looks good to me, thx a lot!

Issue #2422, opened 2017-01-18
Missing command in genesis image


fyhertz:
Hi,

xcat version: 2.13.2 (git commit 4c72fee5435fa53f40f31b41df5809b4c55e4f6b, built Tue Jan 17 06:17:05 EST 2017)

Running on ubuntu 16.04 x86_64

When running bmcsetup on a compute node, running genesis image built with mknb x86_64:
`[xCAT Genesis running on node103 /]# bmcsetup
/bin/bmcsetup: line 107: which: command not found
No ipmitool find, please install it first
`

The command "which" is missing. After manually copying the "which" sh script in the genesis image, bmcsetup seems to work.

Regards

whowutwut:
@fyhertz Thanks for bringing this to our attention

zet809:
@fyhertz Thanks very much for your finding, it is a defect for our script, have fixed in pull request #2424

jonathanmills:
I discovered this the hard way last night, in 2.13.1.  Glad someone's already patched it.

whowutwut:
This has been fixed and verified.

PR #2424, opened 2017-01-19
Don't use "which" to check whether ipmitool is installed in genesis

zet809:
Fix issue #2422 
The command 'which' wasn't included in xcat-genesis-base, use other method to check whether ipmitool is installed.

### DIFF
--- a/xCAT-genesis-scripts/bin/bmcsetup
+++ b/xCAT-genesis-scripts/bin/bmcsetup
@@ -104,7 +104,7 @@ function snooze() {
     fi
 } 
 
-if which ipmitool 2>&1 | grep no; then 
+if ! ipmitool -V 2>/dev/null| grep "version"; then
     echo "No ipmitool find, please install it first"; 
     exit 1;
 fi


whowutwut:
Thanks @zet809  I review and will merge.

whowutwut:
I created a patch file for this fix...   [bmcsetup.2424.patch.txt](https://github.com/xcat2/xcat-core/files/764833/bmcsetup.2424.patch.txt)

```
# cat bmcsetup.2424.patch.txt
--- /opt/xcat/share/xcat/netboot/genesis/ppc64/fs/usr/bin/bmcsetup	2017-01-17 04:39:57.000000000 -0500
+++ /opt/xcat/share/xcat/netboot/genesis/ppc64/fs/usr/bin/bmcsetup	2017-02-09 14:43:29.631419050 -0500
@@ -104,7 +104,7 @@
     fi
 } 
 
-if which ipmitool 2>&1 | grep no; then 
+if ! ipmitool -V 2>/dev/null| grep "version"; then
     echo "No ipmitool find, please install it first"; 
     exit 1;
 fi
```

Create the file or download the file to the xCAT management node in `/tmp/` and run the following
```
cd /
patch -p0 < /tmp/bmcsetup.2424.patch.txt
```

Issue #2412, opened 2017-01-13
Regular Expression expansion is not support for nicips keywords


whowutwut:
Doing some testing today, it looks like the `nicips.<interface>` keyword in the group is not able to expand regular expressions.  Here's a simple example...

```
[root@fs4 research92]# mkdef -t group -o testnicips 
Warning: Cannot determine a member list for group 'testnicips'.
[root@fs4 research92]# chdef -t group -o testnicips ip='|\D+(\d+)\D+(\d+)|10.1.($1).($2)|'
1 object definitions have been created or modified.
[root@fs4 research92]# chdef -t group -o testnicips nicips.eth1='|\D+(\d+)\D+(\d+)|10.1.($1).($2)|'
1 object definitions have been created or modified.
[root@fs4 research92]# lsdef -t group -o testnicips
Object name: testnicips
    grouptype=static
    ip=|\D+(\d+)\D+(\d+)|10.1.($1).($2)|
    members=
    nicips.eth1=|\D+(\d+)\D+(\d+)|10.1.($1).($2)|
```

Creating a test node....
```
[root@fs4 research92]# mkdef frame10node10 groups=testnicips
1 object definitions have been created or modified.
[root@fs4 research92]# lsdef frame10node10
Object name: frame10node10
    groups=testnicips
    ip=10.1.10.10
    nicips.eth1=|\D+(\d+)\D+(\d+)|10.1.($1).($2)|
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
```

As you can see, the ip field expanded correctly, but the `nicips.eth1` field did not.  
I think all the IP related keywords should support this expansion....

Are there other keywords that should support this but does not?

zet809:
Hi, @whowutwut , @chenglch has experience about how to inherit attributes from groups, I'd like him to take a look at this issue.

whowutwut:
I am running with build: 
```
Version 2.13.1 (git commit f834844774a6bef8173db8eb17b23741dd36596f, built Wed Jan 18 06:16:23 EST 2017)
This is a Management Node
cfgloc=Pg:dbname=xcatdb;host=10.3.25.1|xcatadm
dbengine=Pg
dbname=xcatdb
dbhost=10.3.25.1
dbadmin=xcatadm
```

and the node definition has: 

```
    nicips.ib0=|\D+(\d+)\D+(\d+)|20.1.($1*1).($2*1)|
    nictypes.ib0=Infiniband
```

[...]

PR #2425, opened 2017-01-19
Make nics attribute inhertable

chenglch:
As the attribute value in the nics table can not match the regular
expresion format supported by the table library, this patch add the
logic in the def utils to handle the regular expresion.

close-issue: #2412

### DIFF
--- a/perl-xCAT/xCAT/Table.pm
+++ b/perl-xCAT/xCAT/Table.pm
@@ -2299,6 +2299,91 @@ $evalcpt->permit('require');
 
 #--------------------------------------------------------------------------
 
+=head3 transRegexAttrs
+
+    Description: Transform the regular expression attribute to the target value
+                 based on the node name.
+
+    Arguments:
+            Node
+            Attribute value (may have regular expression)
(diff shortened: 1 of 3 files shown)


chenglch:
@bybai  Not sure how `confignics` get the nics attribute, can you review this change?

perl-xCAT/xCAT/Table.pm
```diff
@@ -2299,6 +2299,91 @@ $evalcpt->permit('require');
 
 #--------------------------------------------------------------------------
 
+=head3 transRegexAttrs
+
+    Description: Transform the regular expresion attribute to the target value
```
bybai: expresion  ->  expression

chenglch: Done

bybai:
Hi @chenglch ,
I am OK to merge it.

1, function transRegexAttrs code is coming from part of getNodeAttribs, its function looks well.
2, In the past, the attribute did not support  regular expression, so the change will not affect old user to get the nics attribute.

PR #2426, opened 2017-01-20
modify docker test cases

junxiawang:
modify docker related cases for docker version updated and docker command output changed
@tingtli please help review


tingtli:
I agree to merge

PR #2317, opened 2016-12-15
fix copycds can't auto copycd centos iso

penguhyang:
fix #2303

PR #2429, opened 2017-01-20
Discinfo fixes

samveen:
- Add Centos 5.4, 7.2, 7.3 (x86_64) to discinfo
- Reorder Centos discinfo to be cleaner and easier to update.


immarvin:
agree to merge

immarvin:
hi @samveen , thanks for your fix. We have added some code logic to auto recognize the centos iso in https://github.com/xcat2/xcat-core/pull/2317/files, maybe you can try it so that no need bother to add each discinfo entry, it is appreciated if you can enhance it if any issue on the centos isos you are using. thanks

samveen:
@immarvin I had a look at pull request 2317. You are right, and will do as per your advice.

PR #2430, opened 2017-01-20
update psh_i test case for the CN information is not unified 

caomengmeng:
@tingtli please help to review.


tingtli:
i agree to merge.

PR #2431, opened 2017-01-20
Update rcons, so that it detects conserver

arif-ali:
I have a HA setup of xCAT, and the old method of detecting that conserver is running, doesn't work, and therefore, I keep getting that conserver is running.

The patch here, works for me. See if this fits in with how you guys would like to do this?


zet809:
@arif-ali Hi, arif, will you pls update the conflict and then we can have someone to review this Pull request. Thx a lot!

whowutwut:
I'm ok with the changes.

zet809:

Issue #2436, opened 2017-01-22
tftp continued after xcatd shutdown


arif-ali:
I have seen in numerous situations, where xcatd controls the tftp, to start, abut fails to stop it. I've had to on occasions kill the service, so that it releases the lock to the deirectory, it is using.

```
/usr/sbin/in.tftpd -v -l -s /tftpboot -m /etc/tftpmapfile4xcat.conf
```

Could we have a look at this, for the next release, so that we don't have this problem

immarvin:
hi @arif-ali , you mentioned that ``I've had to on occasions kill the service, so that it releases the lock to the deirectory``, what do you mean by "the lock to the deirectory"? can you explain more?

thanks

arif-ali:
I have setup HA with drbd setup, on various customer solutions. The drbd device is mounted under `/xcatdrbd`. When manually failing over from one machine to the other, on occasions, this service will still run, and therefore still hold onto it, as opened files. I am able to grab this information, by running `fuser -m /xcatdrbd`, and this usually points to a person logged in, and in one of the directories that is located under `/xcatdrbd` or a service is still running.

I hope that makes sense

immarvin:
hi @arif-ali ,thanks for your info, we will consider this request in xCAT 2.13.2 release

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2447

arif-ali:
Thanks @immarvin

whowutwut:
On a simple test, it seems to stop `tftp`
```
[root@fs3 dhcp]# lsxcatd -a 
Version 2.13.2 (git commit e05ae84ee0676779c83185cb52fbd7ddac737c48, built Tue Feb  7 06:15:50 EST 2017)
This is a Management Node

[root@fs3 dhcp]# ps -ef | grep tftp
root      31991      1  0 Feb07 ?        00:00:00 /usr/sbin/in.tftpd -v -l -s /tftpboot -m /etc/tftpmapfile4xcat.conf
root      72804  72760  0 15:27 pts/0    00:00:00 grep --color=auto tftp
[root@fs3 dhcp]# service xcatd stop 
Stopping xcatd (via systemctl):                            [  OK  ]

[root@fs3 dhcp]# ps -ef | grep tftp
root      72868  72760  0 15:27 pts/0    00:00:00 grep --color=auto tftp
```

I will close for now, please re-open if you see issues

arif-ali:
[...]

PR #2447, opened 2017-01-26
Stop tftp service when xcatd shutdown completes

chenglch:
close-issue: #2436

### DIFF
--- a/xCAT-server/lib/xcat/plugins/AAsn.pm
+++ b/xCAT-server/lib/xcat/plugins/AAsn.pm
@@ -125,7 +125,8 @@ sub init_plugin
                     if (xCAT::Utils->isServiceNode()) {    # service node
                         &setup_TFTP($nodename, $doreq);    # setup TFTP
                     } else {                               # management node
-                        &enable_TFTPhpa();
+                        &stop_TFTP();
+                        &enable_TFTP();
                     }
 
                 }
@@ -1204,10 +1205,13 @@ sub setup_TFTP
 
         }
     }
+    $rc = stop_TFTP();
+    if($rc !=0) {
+        xCAT::MsgUtils->message("S", "Failed to stop tftp service.");
+    }
 
     # enable the tftp-hpa
-    $rc = enable_TFTPhpa();
-
+    $rc = enable_TFTP();
     if ($rc == 0)
     {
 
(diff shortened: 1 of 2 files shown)


xCAT-server/lib/xcat/plugins/AAsn.pm
```diff
@@ -1262,29 +1266,95 @@ sub setup_HTTP
...
+#-----------------------------------------------------------------------------
+#
+#=head3 enable_TFTPhpa
+#
+#    Configure and enable the tftp-hpa in the xinetd.
+#
+#=cut
+#
+#-----------------------------------------------------------------------------
+sub enable_TFTPhpa
```
neo954: On Ubuntu, `tftpd-hpa` is another tftp server. It is "HPA's tftp server". Package `tftpd-hpa` conflict with `tftpd`, but provide similar function. The end user can choose to use `tftpd-hpa` to replace the ordinary `tftpd`. It is just another implementation of the tftp server.

I believe the function `enable_TFTPhpa` here is used to handle the `tftpd-hpa` in Ubuntu. Thus, change this function massively may change its original behavior. I think this change is improper.

xCAT-server/sbin/xcatd
```diff
@@ -1549,6 +1549,23 @@ if ($dbmaster) {
 
 # stop the monitoring process
 xCAT_monitoring::monitorctrl::stop($$);
+stop_tftp_service();
```
immarvin: will tftp process be stopped while stopping xcatd with SIGINT or SIGTERM? when xcatd is started with "xcatd -f", it is usually terminated with ``CTRL+C``

xCAT-server/lib/xcat/plugins/AAsn.pm
immarvin: hi @neo954 , the subroutine name might be misleading here, in the body of the subroutine, the ordinary tftp, tftpd-hpa and atftp are processed.

xCAT-server/sbin/xcatd
chenglch: yes, it will be stopped.

xCAT-server/lib/xcat/plugins/AAsn.pm
chenglch: This is a legacy problem, I will use `enable_TFTP` instead

immarvin:
agree to merge

PR #2163, opened 2016-11-16
Clean up logger messages for xcatroot in diskless boot

whowutwut:
Initial changes to clean up some of the log messages that are being generated during diskless provisioning to attempt to have some log messages to time the process.

PR #2448, opened 2017-01-26
Clean up logger messages for xcatroot in diskless boot (2)

whowutwut:
Replaces pull request #2163  , I  could not easily rebase the conflicts for the commit that cleaned up the tabs to 4 spaces.  so just decided to create a new pull request. 

please see discussions in that pull request

### DIFF
--- a/xCAT-server/share/xcat/netboot/rh/dracut_033/xcatroot
+++ b/xCAT-server/share/xcat/netboot/rh/dracut_033/xcatroot
@@ -43,6 +43,7 @@ if [ ! -z "$imgurl" ]; then
                 sleep 27
             fi
         done
+        [ "$xcatdebugmode" > "0" ] && logger -t $log_label -p local4.debug "Download complete."
     elif [ xnfs = x${imgurl%%:*} ]; then
         NFS=1
         SERVER=${imgurl#nfs:}
@@ -53,7 +54,6 @@ if [ ! -z "$imgurl" ]; then
         SERVER=${SERVER%:}
         ROOTDIR=/${ROOTDIR#*/} 
     fi
-    [ "$xcatdebugmode" > "0" ] && logger -t $log_label -p local4.debug "Download complete."
 fi
 #echo 0 > /proc/sys/vm/zone_reclaim_mode #Avoid kernel bug
 
--- a/xCAT-server/share/xcat/netboot/rh/dracut_033/xcatroot
+++ b/xCAT-server/share/xcat/netboot/rh/dracut_033/xcatroot
@@ -31,234 +31,234 @@ fi
 
 imgurl="$(getarg imgurl=)";
 if [ ! -z "$imgurl" ]; then
-	if [ xhttp = x${imgurl%%:*} ]; then
-		NFS=0
-		FILENAME=${imgurl##*/}
-		while [ ! -r "$FILENAME" ]; do
+    if [ xhttp = x${imgurl%%:*} ]; then
+        NFS=0
+        FILENAME=${imgurl##*/}
+        while [ ! -r "$FILENAME" ]; do
             [ "$xcatdebugmode" > "0" ] && logger -t xcat -p debug "downloading $imgurl...."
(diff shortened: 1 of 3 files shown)


whowutwut:
I am debating whether or not to wrap all the echos in logger commands.  `logger -s` will print the log message to the screen.  We only need the `xcatdebug > 0` test for messages that should be "trace debug" messages.  But general status maybe can be written using logger but to the screen.

whowutwut:
Although @samveen  suggested putting more log files into logger, I'd actually like to merge these changes first and then consider adding more later.  It seems like on the normal case, without `xcatdebug` turned on, we only see the following messages in `/var/log/messages` on the compute node. 

```
Feb  3 16:48:46 frame23cn22 dracut-initqueue: Getting http://192.168.3.25:80//install/netboot/rhels7.3-GA/ppc64le/compute/rootimg.cpio.gz...
Feb  3 16:48:48 frame23cn22 dracut-initqueue: 2017-02-03 16:48:48 URL:http://192.168.3.25//install/netboot/rhels7.3-GA/ppc64le/compute/rootimg.cpio.gz [242973035/242973035] -> "rootimg.cpio.gz" [1]
Feb  3 16:48:48 frame23cn22 dracut-initqueue: Setting up RAM-root tmpfs.
Feb  3 16:48:55 frame23cn22 dracut-initqueue: 1498833 blocks
Feb  3 16:48:55 frame23cn22 dracut-initqueue: Extracting the root filesystem...Done extracting the root filesystem...
```

So  it doesn't seem like these messages are very helpful ...

whowutwut:
@zet809 Are we okay to merge this for now?   One change I just added in the last commit is to split the echo of downloading the rootimg to 2 lines so that we can get the timing of the rootimg transfer in the normal logs without additional effort.  Currently the "Done" is printed on the same line so the timestamp for the message in `/var/log/messages` without debug turned on is useless.

xCAT-server/share/xcat/netboot/rh/dracut_033/xcatroot
```diff
@@ -18,247 +21,248 @@ if [ $? -ne 0 ]; then
...
+        done
+    elif [ xnfs = x${imgurl%%:*} ]; then
+        NFS=1
+        SERVER=${imgurl#nfs:}
+        SERVER=${SERVER#/}
+        SERVER=${SERVER#/}
+        ROOTDIR=$SERVER
+        SERVER=${SERVER%%/*}
+        SERVER=${SERVER%:}
+        ROOTDIR=/${ROOTDIR#*/}
```
zet809: It seems the variables are not used at all in this case, do you think we shall output an error message?

xCAT-server/share/xcat/netboot/rh/dracut_033/xcatroot
```diff
@@ -18,247 +21,248 @@ if [ $? -ne 0 ]; then
...
+        NFS=1
+        SERVER=${imgurl#nfs:}
+        SERVER=${SERVER#/}
+        SERVER=${SERVER#/}
+        ROOTDIR=$SERVER
+        SERVER=${SERVER%%/*}
+        SERVER=${SERVER%:}
+        ROOTDIR=/${ROOTDIR#*/} 
+    fi
+    [ "$xcatdebugmode" > "0" ] && logger -t $log_label -p local4.debug "Download complete."
```
zet809: I'd like to put this line before line 46 since only imgurl contains "http" need to download rootimg.

whowutwut: Ok, i will move this up

xCAT-server/share/xcat/netboot/rh/dracut_033/xcatroot
whowutwut: Right, seems like those variables are not used.   The NFS variable is not used either...  maybe this support is not enabled.  We could just print an error message saying unsupported, and then exit?

zet809:

PR #2456, opened 2017-02-01
Clean up useless commented code in the console files 

whowutwut:
Was looking specifically at the `ipmi` file and noticed there is lots of commented lines in these console files that have not been touched for many years,  Deleting those commented lines to keep things clean.  History is recorded in the source control.


neo954:
The patch looks good to me.

zet809:

Issue #2459, opened 2017-02-06
[DEV]"xcatd: install monitor" process(listening on ipv6 socket) will die on status update from node failed on reverse dns lookup 


immarvin:
steps to reproduce:
1.  on a rhels7.3 mn with at least 1 nic with ipv6 enabled
````
# ip addr|grep inet6
    inet6 ::1/128 scope host
    inet6 fe80::9abe:94ff:fe00:c478/64 scope link
    inet6 fe80::9abe:94ff:fe00:c47a/64 scope link
    inet6 fe80::9abe:94ff:fe00:c47b/64 scope link
````
2. install the package "perl-IO-Socket-INET6" and "perl-Socket6" , then restart xcatd
3. make sure the xcatd processes are listening on ipv6 stack:
````
# netstat -nltp|grep -w "3001\|3002"
tcp6       0      0 :::3001                 :::*                    LISTEN      90301/xcatd: SSL li
tcp6       0      0 :::3002                 :::*                    LISTEN      90305/xcatd: instal
````
4. then run "/xcatpost/updateflag.awk" from a node whose hostname can not be reverse lookup on ``nslookup <ip>``
````
/xcatpost/updateflag.awk fs4 3002 "installstatus booted"
````
5. then the process "xcatd: install monitor" will be found died on the mn
````
ps aux|grep "xcatd: install monitor" |grep -v grep
````

immarvin:
``inet_ntoa`` is an ipv4 function, will cause the caller process die with message ``Bad arg length for Socket::inet_ntoa, length is 16, should be 4 at /opt/xcat/sbin/xcatd line 397`` when a ipv6 network address is passed

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2460

immarvin:
the customer has verified the fix

PR #2460, opened 2017-02-06
fix issue  #2459

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2459:

use ``inet_ntop`` instead of ``inet_ntoa``

close the connection from the node which cannot be reverse lookup on the mn


immarvin:
hi @chenglch @neo954 ,would you please help to review this PR

thanks

neo954:
I suggest implement IPv4/IPv6 dual stack support. Even if variable`$inet6support` is `true`, it is still possible to communicate with an IPv4 partner.

neo954:
I will merge this one.

PR #2462, opened 2017-02-07
add /tmp into the exlist of rhels7 diskless osimage

immarvin:
the /tmp directory under diskless osimage might cause the compressed rootimg too large, add it into exlist


bybai:
I am OK to merge it.

Issue #2473, opened 2017-02-09
[DEV] "updatenode <node> -P" will complain "msgutil_r: command not found" if no "/xcatpost" on <node>


immarvin:
````
[root@fs3 postscripts]# updatenode frame1sw1 -P configinterface.vhu
frame1sw1: xcatdsklspost: downloaded postscripts successfully
frame1sw1: /tmp/fileADSxpq.dsh: line 66: msgutil_r: command not found
frame1sw1: Wed Feb  8 19:37:20 UTC 2017 Running postscript: configinterface.vhu
frame1sw1: postscript: configinterface.vhu exited with code 0
frame1sw1: Running of postscripts has completed.
frame1sw1: /tmp/fileADSxpq.dsh: line 66: msgutil_r: command not found
````

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2474

bybai:
I am ok to merge it.

PR #2474, opened 2017-02-09
fix issue [DEV] "updatenode <node> -P" will complain "msgutil_r: command not found" if no "/xcatpost" on <node> #2473

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2473


bybai:
I am ok to merge it.

PR #2477, opened 2017-02-10
[DON'T MERGE]To have petitboot and onie based node be able to get configuration from SN

zet809:
For petitboot and onie, they will get next server from servicenode and/or xcatmaster attribute in noderes table. But the problem is, original code doesn't read those 2 attributes from noderes table.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -1964,7 +1964,7 @@ sub process_request
         } else {
             $chainents = undef;
         }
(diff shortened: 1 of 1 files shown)


whowutwut:
@zet809  could you provide some description here too..   by putting those attributes into the hash table, how is it available during petitboot?

gurevichmark:
Verified that with this change, the request for  /tftpboot/petitboot/ config file comes to SN instead of MN.
I think this PR should be merged.

zet809:
Thx @gurevichmark , I will merge it by myself before the RC1 build. ^_^

whowutwut:
@zet809 Good!  I think we were waiting for the title to change to remove the "do not merge" before merging  :-)

zet809:
@whowutwut Modified, thx for the reminder!

PR #2478, opened 2017-02-10
Remove the reference to hwtype=bmc from the documentation

whowutwut:
Remove the references to hwtype=bmc from the documentation . This was only used internally to detect the temporary bmc node, and the display in the example output is causing confusion.  Replace the command to list out the discovered definition with a wildcard  ``lsdef /node-.*`` and this will help reduce the need to maintain the documentation when the output of the command changes. 

current output of ``bmcdiscover``

```
[root@fs1 common]# bmcdiscover --range 50.45.2.1 -z 
node-8335-gtb-100499a:
	objtype=node
	groups=all
	bmc=50.45.2.1
	cons=ipmi
	mgt=ipmi
	mtm=8335-GTB
	serial=100499A
```

### DIFF
--- a/docs/source/guides/admin-guides/manage_clusters/ppc64le/discovery/mtms/discovery_using_defined.rst
+++ b/docs/source/guides/admin-guides/manage_clusters/ppc64le/discovery/mtms/discovery_using_defined.rst
@@ -27,27 +27,17 @@ The BMC IP address is obtained by the open range dhcp server and the plan in thi
 
 #. Detect the BMCs and add the node definitions into xCAT.
 
-   Use the ``bmcdiscover`` command to discover the BMCs responding over an IP range and automatically write the output into the xCAT database.  You **must** use the ``-w`` option to automatically write the output into the xCAT database. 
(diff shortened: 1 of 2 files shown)


docs/source/guides/admin-guides/manage_clusters/ppc64le/discovery/mtms/discovery_using_defined.rst
```diff
@@ -27,27 +27,17 @@ The BMC IP address is obtained by the open range dhcp server and the plan in thi
...
-          groups=all
-          hwtype=bmc
-          mgt=ipmi
-          mtm=8247-22L
-          nodetype=mp
-          postbootscripts=otherpkgs
-          postscripts=syslog,remoteshell,syncfiles
-          serial=10112CA
-          nodetype=mp
+      lsdef /node-.*
```
zet809: Why there is a '/' before the nodename?

whowutwut: In order to use the wild card, there needs to be a slash in front... 

```
[root@fs4 ~]# nodels frame*
Error: Invalid nodes and/or groups in noderange: frame*
[root@fs4 ~]# nodels /frame.*
frame10node10
frame23cn16
frame23cn17
frame23cn18
frame23cn22
frame23cn24
```

zet809:
Looks good to me, thx a lot.

PR #2479, opened 2017-02-10
Enhance mtm discovery for non-IBM hardware

kcgthb:
Currently, the population of the `mtm` field in the `vpd` table only works for IBM/Lenovo machines. This adds a fallback for cases where the MTM can't be correctly detected in `dodiscovery` and `bmcdiscover`.

Generates a `mtm` field of the form `VENDOR::MODEL`, using:
 * `Product Manufacturer` and `Product Name` from the IPMI FRU pages
 * `sys_vendor` and `product_name` from DMI

### Example results
Via `dodiscovery`
```
# lsdef sh-101-01 -i mtm
Object name: sh-101-01
    mtm=DELL INC.:POWEREDGE C6320
```

Via `bmcdiscover`
```
# bmcdiscover -s nmap --range sh-101-01.ipmi  -z
node-dell-poweredge-c6320-62d4fd2:
        objtype=node
        groups=all
        bmc=sh-101-01.ipmi
        cons=ipmi
        mgt=ipmi
        mtm=DELL:PowerEdge C6320
        serial=62D4FD2
```

The separator (`:`) between vendor and model can obviously be changed if problematic, as well as spaces in vendor and model names.

### DIFF
--- a/xCAT-genesis-scripts/bin/dodiscovery
+++ b/xCAT-genesis-scripts/bin/dodiscovery
@@ -73,14 +73,11 @@ if [ -r /sys/devices/virtual/dmi/id/product_name ]; then  #x86
 	else 
 		MTM=`cat /sys/devices/virtual/dmi/id/product_name|awk -F'[' '{print $2}'|awk -F']' '{print $1}'`
 		if [ -z "$MTM" ]; then
-			SYS=`cat /sys/devices/virtual/dmi/id/sys_vendor`
-            PRD=`cat /sys/devices/virtual/dmi/id/product_name`
-			if [ ! -z "$SYS" ]; then
-				MTM=${SYS^^}":"${PRD^^}
-			else
-				MTM=${PRD^^}
-			fi
-		fi
+            FRU=`ipmitool fru print 0`
+            if [ $? -eq 0 ]; then
+                MTM=`echo "$FRU" | awk -F': ' '/Product Manufacturer/ {m=$2} /Product Name/ {n=$2} END {print m":"n}'`
+            fi
+		fi 
 		SERIAL=`cat /sys/devices/virtual/dmi/id/product_serial`
 	fi
         CPUCOUNT=`cat /proc/cpuinfo |grep "model name"|wc -l`
--- a/xCAT-genesis-scripts/bin/dodiscovery
+++ b/xCAT-genesis-scripts/bin/dodiscovery
@@ -72,6 +72,15 @@ if [ -r /sys/devices/virtual/dmi/id/product_name ]; then  #x86
 		MTM=VMware
 	else 
 		MTM=`cat /sys/devices/virtual/dmi/id/product_name|awk -F'[' '{print $2}'|awk -F']' '{print $1}'`
+		if [ -z "$MTM" ]; then
+			SYS=`cat /sys/devices/virtual/dmi/id/sys_vendor`
+            PRD=`cat /sys/devices/virtual/dmi/id/product_name`
+			if [ ! -z "$SYS" ]; then
+				MTM=${SYS^^}":"${PRD^^}
+			else
+				MTM=${PRD^^}
+			fi
+		fi
 		SERIAL=`cat /sys/devices/virtual/dmi/id/product_serial`
 	fi
         CPUCOUNT=`cat /proc/cpuinfo |grep "model name"|wc -l`
--- a/xCAT-genesis-scripts/bin/dodiscovery
+++ b/xCAT-genesis-scripts/bin/dodiscovery
@@ -73,10 +73,10 @@ if [ -r /sys/devices/virtual/dmi/id/product_name ]; then  #x86
 	else 
 		MTM=`cat /sys/devices/virtual/dmi/id/product_name|awk -F'[' '{print $2}'|awk -F']' '{print $1}'`
 		if [ -z "$MTM" ]; then
-            FRU=`ipmitool fru print 0`
-            if [ $? -eq 0 ]; then
-                MTM=`echo "$FRU" | awk -F': ' '/Product Manufacturer/ {m=$2} /Product Name/ {n=$2} END {print m":"n}'`
-            fi
+			FRU=`ipmitool fru print 0`
+			if [ $? -eq 0 ]; then
+				MTM=`echo "$FRU" | awk -F': ' '/Product Manufacturer/ {m=$2} /Product Name/ {n=$2} END {print m":"n}'`
+			fi
 		fi 
 		SERIAL=`cat /sys/devices/virtual/dmi/id/product_serial`
 	fi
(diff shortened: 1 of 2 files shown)


zet809:
@kcgthb Does it mean the dodiscovery will get "DELL INC.:POWEREDGE C6320" and bmcdiscover will get "DELL:PowerEdge C6320" for MTM based on your modification? If yes, it will cause MTMS based hardware discovery failed. If no, what is the output of MTM for the 2 scripts dodiscovery and bmcdiscover?

kcgthb:
That's correct, the resulting strings will be different. I didn't realize they needed to match for the MTMS discovery to work.

It could be a bit of a challenge, because `dodiscovery` gets the information from DMI, while `bmcdiscover` gets it from the BMC via IPMI. And there is no guarantee that they would match. As a matter of fact, they often don't. On the same node:
```
# ipmitool fru print 0
[...]
 Product Manufacturer  : DELL
 Product Name          : PowerEdge R630
 Product Version       : 01
[...]
# dmidecode  -s system-manufacturer; dmidecode  -s system-product-name
Dell Inc.
PowerEdge R630
```

We can capitalize all the strings, but in one case, the manufacturer is "Dell Inc.", and in the other one, it's "DELL", so it won't solve the problem.

Maybe it would be better to take the MTM information from the same place for both `dodiscovery` and `bmcdiscover`? Meaning getting the information from the BMC controller with `ipmitool` rather than getting it from DMI. `dodiscovery` could invoke `ipmitool` locally as it's running in the Genesis image. That way, we'll be certain that the information is consistent between `dodiscovery` and `bmcdiscover`.

What do you think?

kcgthb:
Just commited a proposal for using FRU information within `dodiscovery` instead of parsing DMI strings.

zet809:
@kcgthb Can you show me the information of all frus though IPMI? "ipmitool -I lanplus -H <bmc_Ip> -U <username> -P password fru" will output all the information about FRUs.

kcgthb:
Sure, here it is:
```
FRU Device Description : Builtin FRU Device (ID 0)
 Board Mfg Date        : Sat Sep  3 06:29:00 2016
 Board Mfg             : DELL
 Board Product         : PowerEdge C6320
 Board Serial          : CN747516930287
 Board Part Number     : 082F9MA01
 Product Manufacturer  : DELL
 Product Name          : PowerEdge C6320
 Product Version       : 01
 Product Serial        : 62D4FD2

FRU Device Description : OEM fru (ID 17)
```

zet809:
For IBM/Lenovo server is, there is 'Product Part Number' attribute in FRU 0 which is equal to 'dmidecode  -s system-product-name'.
And It seems for DELL servers, the 'Product Name' in FRU 0 equal to the 'dmidecode  -s system-product-name', they contain the same value: PowerEdge C6320. Why don't we just use it?

kcgthb:
The Product Name indeed matches between DMI and FRU in that specific case, but as those are two independent sources of information, there is no guarantee that it will always match.

Actually, I have cases where the DMI information is present, but the FRU is incomplete:
```
# dmidecode -s system-product-name
SYS-4027GR-TR
# ipmitool fru
FRU Device Description : Builtin FRU Device (ID 0)
 Board Mfg Date        : Sun Dec 31 16:00:00 1995
 Board Mfg             : Supermicro
 Board Serial          :
 Product Serial        :
```
As well as opposite cases where the FRU has the information, but the DMI is empty:
```
# cat /sys/devices/virtual/dmi/id/product_name
empty
# dmidecode -s system-product-name
empty
# ipmitool fru print
FRU Device Description : Builtin FRU Device (ID 0)
 Chassis Type          : Rack Mount Chassis
 Board Mfg Date        : Wed Jan 15 02:45:00 2014
 Board Mfg             : TYAN
 Board Product         : S7059GM2NR
 Board Serial          : C14D3DE47003
 Board Part Number     : 411T47600083
 Product Manufacturer  : TYAN
 Product Name          : B7059F77AV6R
 Product Part Number   : 5412T4760001
 Product Serial        : TCFSQ10C0034

FRU Device Description : AST2300
 Chassis Type          : Rack Mount Chassis
 Board Mfg Date        : Wed Jan 15 02:45:00 2014
 Board Mfg             : TYAN
 Board Product         : S7059GM2NR
 Board Serial          : C14D3DE47003
 Board Part Number     : 411T47600083
 Product Manufacturer  : TYAN
 Product Name          : B7059F77AV6R
 Product Part Number   : 5412T4760001
 Product Serial        : TCFSQ10C0034
```

Since the information from both `dodiscovery` and `bmcdiscover` have to match, and since `bmcdiscover` only has access to the FRU information, I believe that xCAT should only rely on the FRU information for the discovery. We can't be certain that DMI will match the FRU information, or that it will exist at all.

Also, I think it would be better for non-IBM/Lenovo servers to also have the manufacturer in the MTM field. It could be of tremendous value for heterogeneous clusters where a lot of different hardware types are managed under the same xCAT installation.

zet809:
@kcgthb I think your concern is reasonable, can you show me the output of MTM/serial for both dodiscovery and bmcdiscovery? If they are same, I think it is OK for me to merge this pull request. Thx!

kcgthb:
Sure! Here they are:

#### dodiscovery
```
[xCAT Genesis running on sh-101-02 /]# dodiscovery
xcat.genesis.dodiscovery: Beginning node discovery process...
xcat.genesis.dodiscovery: Waiting for nics to get addresses
eth0|10.10.101.2
xcat.genesis.dodiscovery: Network configuration complete, commencing transmit of discovery packets
xcat.genesis.dodiscovery: Restart network interfaces...
Done waiting

[xCAT Genesis running on sh-101-02 /]# cat /tmp/discopacket.gz | gzip -d| grep mtm
<mtm>DELL:PowerEdge C6320</mtm>
```

#### bmcdiscover
```
# bmcdiscover -s nmap --range sh-101-02.ipmi  -z
node-dell-poweredge-c6320-66d5nd2:
        objtype=node
        groups=all
        bmc=sh-101-02.ipmi
        cons=ipmi
        mgt=ipmi
        mtm=DELL:PowerEdge C6320
        serial=66F2DN5
```

xCAT-genesis-scripts/bin/dodiscovery
```diff
@@ -72,6 +72,12 @@ if [ -r /sys/devices/virtual/dmi/id/product_name ]; then  #x86
 		MTM=VMware
 	else 
 		MTM=`cat /sys/devices/virtual/dmi/id/product_name|awk -F'[' '{print $2}'|awk -F']' '{print $1}'`
+		if [ -z "$MTM" ]; then
+            FRU=`ipmitool fru print 0`
```
zet809: The modification looks good to me. Will you pls follow the indent? Once done, I will merge this modification.

kcgthb: Sure, here it is.

PR #2484, opened 2017-02-15
Update bootparams in destiny to support discovery without DB access

chenglch:


zet809:
Looks good to me, thx a lot!

Issue #2487, opened 2017-02-16
[fvt]2.13.2 xcat install monitor process hang and could not reort the node status


junxiawang:
env:rhels7.3 ppc64le
build:[root@c910f03c05k09 ~]# lsxcatd -v
Version 2.13.2 (git commit c707c7a003c196725f4b957fa100f9f4ea85286f, built Tue Feb 14 06:16:04 EST 2017)
[root@c910f03c05k09 ~]#  
How to reproduce:
```
1. 
provision one node
2.rinstall c910f03c05k02 osimage=rhels7.3-ppc64le-install-compute
3.Then the node status still in powering on status.

```

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2488

PR #2488, opened 2017-02-16
fix issue [fvt]2.13.2 xcat install monitor process hang and could not report the node status

immarvin:
fix https://github.com/xcat2/xcat-core/issues/2487


hu-weihua:
I have reviewed the code change and agree to merge, thanks @immarvin

PR #2503, opened 2017-02-20
Use bulk operation to update chain table for multiple nodes

chenglch:
partial-issue: #2024


zet809:
Looks good to me, I agree to merge. Thx very much!

Issue #2475, opened 2017-02-10
[Customer] Can't locate Getopt /Std.pm during genesis discovery


cxhong:
Lab Service reported this error during genesis discovery:
`````
<166>Feb 10 11:41:06 xcat.genesis.doxcat: Starting ntpd...
<166>Feb 10 11:41:06 xcat.genesis.doxcat: Waiting for ntpd to synchronize...
Can't locate Getopt/Std.pm in @INC (you may need to install the Getopt::Std module) (@INC contains: /usr/l
ocal/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib
64/perl5 /usr/share/perl5 .) at /usr/sbin/ntp-wait line 5.
BEGIN failed--compilation aborted at /usr/sbin/ntp-wait line 5.
<166>Feb 10 11:41:06 xcat.genesis.doxcat: ... ntpd did not synchronize.
<166>Feb 10 11:41:06 xcat.genesis.doxcat: Attempting to sync hardware clock...
`````
Here is xCAT version on their system:
```
[root@smn dev]# lsxcatd -a
Version 2.13.1 (git commit 62bd7a4072bb629ede16974272c993f4e707765f, built Wed Jan 11 08:45:41 EST 2017)
This is a Management Node
dbengine=SQLite
`````

immarvin:
hi @zet809 , would you please take a look at this?

whowutwut:
Assigned to @neo954  This issue occurs when running code changed by PR https://github.com/xcat2/xcat-core/pull/2333

zet809:
I need to rebuild xcat-genesis-base to include those libraries.

whowutwut:
Is this something that we have looked into potentially using containers to help simplify the re-build process of the dependencies?

zet809:
No, for genesis-base-ppc64, I was using my CN which had been installed Fedora23 before.

neo954:
Tested `xCAT-genesis-base-ppc64-2.13.2-snap201702150049.noarch` on a PowerVM LPAR. And get the following error message 

    [xCAT Genesis running on (none) /]# ntp-wait                                   Unexpected first line <System error>

neo954:
Anyway, the `ntp-wait` can be run in the genesis now. We still need to solve the network problem for PowerVM LPAR, which need linux kernel module `ibmveth`. It is another problem.

[...]

PR #2504, opened 2017-02-20
Fix issue 2475 to add perl library for ntp-wait

zet809:
Fix issue #2475 , include perl library within xcat-genesis-base package.


immarvin:
i am ok with the change

PR #2505, opened 2017-02-20
update rh7.2 and rh7.3 test cases for ppc64 and ppc64le bundle files

caomengmeng:
@tingtli 

update rh7.2 and rh7.3 test cases for ppc64 and ppc64le bundle files


tingtli:
i have reviewed and agree to merge.

Issue #2499, opened 2017-02-17
[openbmc] Support ssh based console in conserver 


whowutwut:
For openbmc, the SOL consoles is moving away from ``ipmitool sol active`` to ssh based on a port opened on the BMC.   This issue requests adding in very quick and basic support for ssh console so we can be ready to test it out once we get machines running openbmc.  

From my investigation, it seems like we can scaffold and just use the normal ssh port to do testing, and the concept is identical.   For example, create a "debugtest" user on a compute node, giving it password "abc123".    Hard code a "ssh" console: 

```
exec "echo abc123 | ssh debugtest\@10.3.25.1";
```

Set ``cons=ssh`` in the dummy xCAT node object and `rcons dummynode` and it will open a console to this machine. 

```
[root@c910f02c05p03 cons]# rcons test_openbmc_ssh
[Enter `^Ec?' for help]

[debugtest@fs4 ~]$ 
[debugtest@fs4 ~]$ 
[debugtest@fs4 ~]$ 
[debugtest@fs4 ~]$ uptime
 06:53:18 up 3 days,  6:11,  3 users,  load average: 0.01, 0.05, 0.05
[debugtest@fs4 ~]$ [disconnect]
[root@c910f02c05p03 cons]#
```

## Planned Externals: 

* new console ``cons=ssh``
 
* User will need to define in the passwd table the password for the ssh console, similar to the entry we have today for ``ipmi`` based consoles.  

* Ability to override the SSH_CONSOLE_PORT on the xCAT MN/SN, so in the ssh console, we will hard code the port based on openbmc, but to allow users to use this against a different port, and to help Unit testing, we should be able to ``export SSH_CONSOLE_PORT=22`` and not have to change the console file code.

whowutwut:
I target the 2.13.2 milestone, we don't need FVT to test this until we get some real machines with openbmc firmware.  This just allows us to be ready with what we can do before machines arrive.

cxhong:
Created a new plugin and cons for openbmc to support this feature
```
# lsdef openbmc1
Object name: openbmc1
    cons=openbmc
    groups=all,openbmc
    ip=x.x.x.x
    mgt=openbmc
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
````
when command `rcons openbmc1`  is running, it will call 
```
exec /opt/xcat/share/xcat/cons/openbmc openbmc1;
```
[...]

PR #2511, opened 2017-02-21
[Dev] Support rcons command for openbmc

cxhong:
for issue #2499 .

Create a new plugin and able to make conserver for openbmc node
1) Created openbmc node object
```
# lsdef openbmc1
Object name: openbmc1
    cons=openbmc
    groups=all,openbmc
    ip=9.3.x.x
    mgt=openbmc
    postbootscripts=otherpkgs
````
2) run makeconservercf command
````
makeconservercf openbmc1
````
3) rcons to openbmc1 node
````
# rcons openbmc1
[Enter `^Ec?' for help]
Ubuntu Zesty Zapus (development branch) w33l ttyS0

w33l login:
Ubuntu Zesty Zapus (development branch) w33l ttyS0

````

**NOTE**  Passwordless ssh needs to be configured.  Need to open another issue for that.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -294,7 +294,6 @@ sub parse_args {
 
 sub unsupported {
     my $callback = shift;
-    xCAT::SvrUtils::sendmsg("ENV: OPENBMC_DEVEL=$ENV{OPENBMC_DEVEL},$ENV{'OPENBMC_DEVEL'}\n",  $callback);
     if ($::OPENBMC_DEVEL ne "YES") {
         return ([ 1, "This function is currently not supported" ]);
     } else {
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -1,21 +1,11 @@
 #!/usr/bin/perl
-# IBM(c) 2017 EPL license http://www.eclipse.org/legal/epl-v10.html
-#-------------------------------------------------------
+## IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
 
-=head1
-  xCAT plugin package to handle openbmc 
-
-   Supported command:
-       getopenbmccons
-
-=cut
-
-#-------------------------------------------------------
 package xCAT_plugin::openbmc;
 
 BEGIN
 {
-    $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : -d '/opt/xcat' ? '/opt/xcat' : '/usr';
+    $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : '/opt/xcat';
 }
 use lib "$::XCATROOT/lib/perl";
 use strict;
@@ -31,19 +21,8 @@ use xCAT::SvrUtils;
 use File::Basename;
 use Data::Dumper;
 use JSON;
-use Getopt::Long;
-use File::Path;
-use Term::ANSIColor;
-use Time::Local;
-use strict;
-use Class::Struct;
-use xCAT::PasswordUtils;
 
 $::OPENBMC_DEVEL = $ENV{'OPENBMC_DEVEL'};
-
-my $VERBOSE = 0;
-my %allerrornodes = ();
-my $callback;
 
 #-------------------------------------------------------
 
@@ -154,6 +133,7 @@ my $async;
 my $cookie_jar;
 
 my $callback;
+my %allerrornodes = ();
 
 #-------------------------------------------------------
 
@@ -237,6 +217,7 @@ sub process_request {
     my $handle_id;
     my $content;
     $wait_node_num = keys %node_info;
+    my @donargs = ();
 
     foreach my $node (keys %node_info) {
         $bmcip = $node_info{$node}{bmc};
@@ -246,7 +227,17 @@ sub process_request {
         $handle_id_node{$handle_id} = $node;
         $node_info{$node}{cur_status} = $next_status{ $node_info{$node}{cur_status} };
         print "$node: DEBUG POST $login_url -d $content\n";
+        push @donargs, [ $node,$bmcip,$node_info{$node}{username}, $node_info{$node}{password}];
     }  
+
+    #process rcons
+    if ($request->{command}->[0] eq "getopenbmccons") {
+        foreach (@donargs) {
+            getopenbmccons($_, $callback);
+        }
+        return;
+    }
+
 
     while (1) { 
         last unless ($wait_node_num);
@@ -255,33 +246,6 @@ sub process_request {
         }
     } 
 
-
-    my $ipmitab     = xCAT::Table->new('ipmi');
-    my $ipmihash = $ipmitab->getNodesAttribs($noderange, [ 'bmc', 'username', 'password' ]);
-    my $authdata = xCAT::PasswordUtils::getIPMIAuth(noderange => $noderange, ipmihash => $ipmihash);
-
-    foreach (@$noderange) {
-        my $node     = $_;
-        my $bmcip;
-        my $nodeuser = $authdata->{$node}->{username};
-        my $nodepass = $authdata->{$node}->{password};
-        my $ent;
-        if (defined($ipmitab)) {
-            $ent = $ipmihash->{$node}->[0];
-            if (ref($ent) and defined $ent->{bmc}) { $bmcip = $ent->{bmc}; }
-        }
-        push @donargs, [ $node,$bmcip,$nodeuser, $nodepass];
-        my $output = "openbmc, get $username and $password from ipmi table for $bmcip";
-        xCAT::SvrUtils::sendmsg($output, $callback, $node, %allerrornodes);
-    }
-
-
-    if ($request->{command}->[0] eq "getopenbmccons") {
-        foreach (@donargs) {
-            getopenbmccons($_, $callback);
-        }
-        return;
-    }
     return;
 }
 
@@ -330,6 +294,7 @@ sub parse_args {
 
 sub unsupported {
     my $callback = shift;
+    xCAT::SvrUtils::sendmsg("ENV: OPENBMC_DEVEL=$ENV{OPENBMC_DEVEL},$ENV{'OPENBMC_DEVEL'}\n",  $callback);
     if ($::OPENBMC_DEVEL ne "YES") {
         return ([ 1, "This function is currently not supported" ]);
     } else {
@@ -673,7 +638,15 @@ sub rinv_response {
     return;
 }
 
+#-------------------------------------------------------
 
+=head3  getopenbmccons
+
+    Process getopenbmccons
+
+=cut
+
+#-------------------------------------------------------
 sub getopenbmccons {
     my $argr = shift;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by whowutwut** [comment]
Let's hold off for a few days before merging to see if we can find an alternative for sshpass..  also looking to see if we are able to send ssh keys to this bmc.

File: xCAT-server/lib/xcat/plugins/openbmc.pm [hunk 1]
```diff
@@ -0,0 +1,189 @@
...
+
+    #Assume the openbmc user/password is defined in the passwd
+    my $passtab = xCAT::Table->new('passwd');
+    if ($passtab) {
+       ($dba) = $passtab->getAttribs({key=>'openbmc'},qw(username password));
+        if ($dba) {
+            if ($dba->{username}) { $username = $dba->{username}; }
+            if ($dba->{password}) { $password = $dba->{password}; }
+      }
+    }
```
zet809: From my point of view, the openpower server using OpenBMC has no difference with servers using BMCs. Do you think we can reusing the ipmi table to store BMC/OpenBMC related information? Such as 'bmc ip', "bmc username", "bmc password". If so, the bmcusername/bmcpassword in ipmi table will be more specific than in 'passwd' table. You can reference xCAT::PasswordUtils::getIPMIAuth to get the BMC account.

zet809:
@cxhong Hi, casandra, from the user point of view, the server with OpenBMC shall not have difference with normal openpower server. I mean that we may don't need to define a BMC node. The info of BMC shall be just the attributes of a host node, for example, the bmc attribute of a node shall be the ip address of the BMC for it, bmcusername,bmcpassword shall have the similar functions. What is your idea?

samveen:
@whowutwut For copying keys to bmc, maybe do some magic with `ssh-copy-id`?

whowutwut:
@samveen  Thanks, will look into that.   @zet809 Right, we should not have to define a bmc node...   Let's at least implement the following...  

```
[root@fs1 ~]# mkdef opennode1 bmc=9.3.185.156 mgt=openbmc groups=all bmcusername=USERID bmcpassword=PASSW0RD
Error: Cannot set the attr='bmcpassword' attribute unless 'mgt=ipmi'.
Error: Cannot set the attr='bmcusername' attribute unless 'mgt=ipmi'.
Error: Cannot set the attr='bmc' attribute unless 'mgt=ipmi'.
```
So we need to enable the above to work...

Then the node object can override the password table entry for `openbmc` and we support a new management type called `openbmc`.    Yes, we are still under discussion about the actual name of the plugin, but we can change that easily enough.  

All this code being checked in is really unsupported and we are staging it in so we can develop functions in parallel.   If you think it's better, we could also create a new branch for this development work but since we can disable this plugin by default, it should be safe to commit in the master branch. 

After @cxhong  puts in this function, we can disable all the handled_commands unless ``export OPENBMC_DEVEL=TRUE``.

cxhong:
example of openbmc compute node for this case:
```
]# lsdef opennode1
Object name: opennode1
    bmc=9.3.185.161
    bmcpassword=0penbmc
    bmcusername=root
    cons=openbmc
    groups=all
    ip=192.168.45.161
    mgt=ipmi
    postbootscripts=otherpkgs
`````

**Comment by whowutwut** [comment]
I'm good with the changes.   Will plan to merge this in.   So for this initial support, the ssh keys need to be copied over to the BMC.  Will look into other options for providing password, but in our dev environment, this should be OK for now so we don't have to ship a 3rd party program: ``sshpass``

**Comment by cxhong** [comment]
@whowutwut , @zet809 ,  I added code to config openbmc passwordless.  right now,  it's only config when user run rcons.   It will send public key to openbmc

**Comment by whowutwut** [comment]
@cxhong  I'm not sure if we want to automatically do this configssh...   Once we do this any "root" user will be able to communicate with any openbmc..  

I also wonder if we want to do this under the covers, or at least need to print out I wonder we want to do this under the covers or we should actually have a configuration variable that turns that on or off.  For example... 
```
export XCAT_OPENBMC_CONFIGSSH=TRUE
```

then we will automatically send over the keys. 

Or just let it fail and print a message "If console does not appear, validate that you have passwordless ssh working for the bmc or export XCAT_OPENBMC_CONFIGSSH=TRUE, to allow xCAT to send over the public key to the BMC"

**Comment by samveen** [comment]
Maybe it would be better to restrict BMC access to just the `MN`/`SN` by prepending `from="$MN_IP"` to the `authorized_keys` entry for the `root` key on the BMC. (`from="$IP" ssh-rsa .....`)

**Comment by cxhong** [comment]
Thanks @samveen , somehow add from="$IP" didn't work for this system.
We will not enable passwordless for now.

File: xCAT-server/share/xcat/cons/openbmc [hunk 2]
```diff
@@ -0,0 +1,95 @@
...
+# IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
+use Fcntl qw(:DEFAULT :flock);
+use Time::HiRes qw(sleep);
+use File::Path;
+BEGIN {
+    $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : -d '/opt/xcat' ? '/opt/xcat' : '/usr';
+}
+my $sleepint = int(rand(10)); #Stagger start to avoid overwhelming conserver/xCATd
+my ($lockfd, $nodeip);
+my $username = 'UESRID';
```
whowutwut: This one is spelled wrong, but should we just default to the `root/0penBmc` defaults?

xCAT-server/share/xcat/cons/openbmc
```diff
@@ -0,0 +1,95 @@
...
+my $isintel = 0;
+my $sleepint;
+my $rc;
+ 
+my $sshport = 2200;
+if ($ENV{SSHCONSOLEPORT}) {
+    $sshport= $ENV{SSHCONSOLEPORT};
+}
+
+exec "ssh -p $sshport -l $username $nodeip";
```
whowutwut: If we add a print statement here... would it appear? 

`print "If the console cannot connect, please verify whether ssh keys has been configured on the bmc for $username user\n"; `

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -0,0 +1,292 @@
...
+
+#-------------------------------------------------------
+package xCAT_plugin::openbmc;
+
+BEGIN {
+    $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : -d '/opt/xcat' ? '/opt/xcat' : '/usr';
+}
+use lib "$::XCATROOT/lib/perl";
+
+use xCAT::Table;
```
whowutwut: Are all these imports needed up to line 39... Or just copied and pasted?  
Can we removed the unused ones?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -0,0 +1,292 @@
...
+        }
+        push @donargs, [ $node,$nodeip,$nodeuser, $nodepass];
+        my $output = "openbmc, get $username and $password from ipmi table for $nodeip";
+        xCAT::SvrUtils::sendmsg($output, $callback, $node, %allerrornodes);
+    }
+
+
+    if ($request->{command}->[0] eq "getopenbmccons") {
+        foreach (@donargs) {
+            #configssh($_, $callback);
```
whowutwut: On 2nd thought, could we just remove this code and function...   thinking about this.. it's easier to re-add later than to clean up :-)  but I think we want to go with a different method.  We can have more insight once we have some machines to play with.

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -0,0 +1,292 @@
+# IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
```
whowutwut: @daniceexi  @zet809  Is there a newer URL we should be using for new files or new boiler place for the license?   If not, at least we should date this 2017.

xCAT-server/share/xcat/cons/openbmc
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env perl
+# IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
```
whowutwut: Same comment about the license as above.. let's see what the response is...

cxhong: changed to 2017

xCAT-server/share/xcat/cons/openbmc
cxhong: doesn't appear

xCAT-server/share/xcat/cons/openbmc
cxhong: changed to default

xCAT-server/lib/xcat/plugins/openbmc.pm
cxhong: all the configssh routine removed

xCAT-server/share/xcat/cons/openbmc
zet809: What about putting the warning message Victor mentioned before exec? I think it shall print out to the stdout.

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -0,0 +1,179 @@
...
+        @exargs = @$extrargs;
+    }
+
+    my $ipmitab     = xCAT::Table->new('ipmi');
+    my $ipmihash = $ipmitab->getNodesAttribs($noderange, [ 'bmc', 'username', 'password' ]);
+    my $authdata = xCAT::PasswordUtils::getIPMIAuth(noderange => $noderange, ipmihash => $ipmihash);
+
+    foreach (@$noderange) {
+        my $node     = $_;
+        my $nodeip = $node;
```
zet809: What kind of node object will rcons work against? If the host node, the IP here shall be the host IP rather than bmc IP. If it is bmc node, it means we shall define a special BMC node, right?

whowutwut: Good point Er Tao... I am looking at this section more closely... 

 @cxhong  line 126 and line 129 is exactly the same.   
Line 131 we check if ``defined($ipmitab)``.. this check if really needed should be at line 121,, we already use the `$ipmitab` in line 121... so why are we checking it down here?   Rather, should we be checking the ipmihash... 

But I think the variables should be more clear... instead of nodeip, we should call it bmcip, because that's that IP address that we are going to pass back.. right?  We are first going against a node object. ..and then getting the `bmc` ip value ....  Have we tested both the `bmc=IP` and `bmc=<node>` if we define the BMC as a node object?

xCAT-server/share/xcat/cons/openbmc
whowutwut: @cxhong  can you try if you pass it a IP address that is invalid?  Because the line above I've seen print out to the console... 
`print "Console not ready, retrying in $sleepint seconds (Ctrl-e,c,o to skip delay) \n";``

xCAT-server/lib/xcat/plugins/openbmc.pm
cxhong: yeh, that nodeip should be bmcip, i will change the name.  In this PR, we didn't define openbmc object, still used ipmi as we discussed .

xCAT-server/share/xcat/cons/openbmc
cxhong: it didn't show up in the rcons, but it logged in the /var/log/console/nodename.  So, I will add that message in

zet809:
I think the code logic of rcons is good. Hi, @xuweibj , will you pls take a look at the code based on your implementation? Thx  a lot!

xuweibj:
I'm good with the changes.

whowutwut:
Thanks @cxhong !

PR #2520, opened 2017-02-22
Mychanges

billwaj:
Please pull in changes for 2.12 branch for:
011eaa95cf7b121475766d85436f794062521a7b: lsslp to correctly create the correct frame for the cec parent.

47d096bcde0077b01294a957a0c21dba59f5ce77: update rspconfig command to add sslmode option for Power hardware (P775 specifically) to disable or enable sslv3 HMC (dfm) connections to the fsp.

209907ad108960db95b12a00a60ebc43f5b15f38: update nodeset command to add a grub2 boot password and optional grub2 superuser id which requires the userid/password be entered if user tries to modify the xCAT OS Deployment panel from the console.


zet809:
@billwaj Hi, Bill, we will release 2.13.2 this Friday, the modification for grub2 may affect functions for OS deployment for VMs, so I'd like to move this pull request for 2.13.3 release, what is your idea?

billwaj:
Hi Er Tao, I had spoken to Victor and would like this to be put into the 2.12 branch if possible. All of our testing to date was with 2.12.3 and 2.12.4. I will follow up with Victor today. Thanks.

________________________________

whowutwut:
@zet809 Talking to Bill, I think the plan right now is he created this PR, they are testing the service pack on 2.12 release...  If they are ready, we can create a 2.12.4 release and they can use that to send as part of the SP.  

If we think the changes for grub2 should not be made, is there a change we should recommend?  

Moving forward, we would need to have him cherry pick this into master if they want to preserve the changes for forward releases... otherwise stay on 2.12.x release.

whowutwut:
Looking at the release history, 2.12.4 is a milestone that is released, so really we should be at 2.12.5  @billwaj  can you change the Version file and add it to your branch . Then push the change up again... 

```
git checkout mychanges 
vi Version 
git add Version
git commit -m "update the version to 2.12.5"
git push origin mychanges ```

This pull request will reflect that change .

**Comment by billwaj** [comment]
I updated Version to 2.12.5

File: xCAT-server/lib/xcat/plugins/grub2.pm [hunk 1]
```diff
@@ -253,8 +253,28 @@ sub setstate {
...
         if ($cref and $cref->{currstate} ne "offline") {
+            # process grub2 password and user arguments
+            if (defined $grub2_password) {
+                my $passwdtag = "password";
+                if ($grub2_password =~/grub.pbkdf2/) {
+                    $passwdtag = "password_pbkdf2";
+                }
+                if (defined $grub2_user) {
+                    print $pcfg "set superusers=\"$grub2_user\"\n";
+                    print $pcfg "$passwdtag $grub2_user $grub2_password\n";
```
zet809: Can "grub.pbkdf2" be the password? It seems to be a keyword. If yes, I think this pull request is OK to me.

billwaj: "grub.pbkdf2" can be a password. However, this test it is meant to check for an encrypted password generated from the command grub2-mkpasswd-pbkdf2. "grub.pbkdf2.sha512" is pre-pended to the encrypted password. i.e. 
"grub.pbkdf2.sha512.10000.6A0E2A0D036984CE53D85C2A1FD46F6E2092DDD32DEDB2D7A10D4AF184C59709179FFC907016EBBBEA12DB0CB794DB3107ACF0639A67F4A509673F0CD0FD1A70.0"EDEA959E232C6226382F0E391C8AF190264D2F5F7B9D7F9BD2977BDB171B417720B78EE07AE2548111441ED9BB393272B67C7561323EBFE39B1A0FCE5689082"

I think the better pattern match would be to check if the password starts with "^grub2.pbkdf2" and matches more than one character after that. I will make this change.

zet809:
@billwaj Hi, bill, looks good to me, I agree to merge. Thx a lot. BTW, do you want to merge this modification also to master branch?

billwaj:
Thank you. Yes you can merge these changes into master.

whowutwut:
@billwaj  Can you rename the pull request to something that better describes the changes contained instead of "Mychanges"? Thanks!

PR #2522, opened 2017-02-22
[perl-xCAT] Fix #2521 - 2.12 branch

### LINKED ISSUE #2521: `makedhcp` fails for nodes on remote networks
When I try to `makedhcp` for a node for a `!remote!` network, the lease configuration isn't added to the DHCPD lease file.

As per the man page for the `networks` table, the format of the field `mgtifname` is as follows:
```
[sgdc:samveen@inframgtlivexcatmn1 ~]$ man 5 networks |grep '^\s*mgtifname' |fmt -w 100
       mgtifname The interface name of the management/service node facing this network.
       !remote!<nicname> indicates a non-local network on a specific nic for relay DHCP.
```

However, the check for `mgtifname` in `NetworkUtils::nodeonmynet` does a string equivalence for `!remote!` instead of checking whether the field contains `!remote!`.
```
if ((defined $_->{mgtifname}) && ($_->{mgtifname} eq '!remote!'))
```
I'll be submitting the trivial fix as pull requests for branches 2.9 onwards till master. 

I also noticed that the code for loading network configuration in `dhcp::process_request` (in file `xCAT-server/.../plugins/dhcp.pm`) contains correct code for loading remote network information for use with DHCP, which is almost identically duplicated into `NetworkUtils::nodeonmynet`, except for the bug as described here.

Maybe the network loading code should be refactored into a single utility function, instead of being duplicated at multiple places.

samveen:
Fix for bug #2521 for 2.12 branch

### DIFF
--- a/perl-xCAT/xCAT/NetworkUtils.pm
+++ b/perl-xCAT/xCAT/NetworkUtils.pm
@@ -1471,7 +1471,7 @@ sub nodeonmynet
         my $nettab = xCAT::Table->new("networks");
         my @vnets = $nettab->getAllAttribs('net', 'mgtifname', 'mask');
         foreach (@vnets) {
-            if ((defined $_->{mgtifname}) && ($_->{mgtifname} eq '!remote!'))
+            if ((defined $_->{mgtifname}) && ($_->{mgtifname} =~ /!remote!/))
             {
                 if (xCAT::NetworkUtils->ishostinsubnet($nodeip, $_->{mask}, $_->{net}))
                 {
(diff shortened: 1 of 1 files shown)


zet809:
Had merged the modification to master and 2.13 branch. So close this pull request.

samveen:
@zet809 Merging into master does not backport these to older branches. That is why I created the merge requests to add the feature to 2.12 branch.

whowutwut:
Yes,  Agreed. I had a comment on #2523.  I think we should accept the changes on the older branches in the event that we do release a mod level , we don't need to cherry pick things that we want to put in...

zet809:
Hi, @whowutwut and @samveen , I had merged this pull request for 2.9.2.10.2.11 and 2.12 branch. For 2.13, it was merged from master, so it had already included this modification.

Issue #2528, opened 2017-02-22
[fvt]2.13.2  install xcat Error loading module /opt/xcat/lib/perl/xCAT_plugin/rhevm.pm 


junxiawang:
env: all os
build:[root@c910f04x12v02 ~]# lsxcatd -v
Version 2.13.2 (git commit bf47335dccb7dd365fb175e219c2c325b6bd46e0, built Tue Feb 21 07:16:03 EST 2017)
[root@c910f04x12
How to reproduce:
```
In cluster.log:
Feb 21 21:23:53 c910f03c17k27 xcat[13432]: Error loading module /opt/xcat/lib/perl/xCAT_plugin/rhevm.pm  ...skipping
Feb 21 21:23:53 c910f03c17k27 xcat[13432]: Error on restart xinetd
Feb 21 21:23:53 c910f03c17k27 xcat[13465]: INFO xcatd: command log process 13465 start
Feb 21 21:23:56 c910f03c17k27 xcat[13576]: xCAT: Allowing mknb ppc64 for root from localhost
Feb 21 21:24:41 c910f03c17k27 xcat[13592]: xCAT: Allowing mknb x86_64 for root from localhost
Feb 21 21:25:16 c910f03c17k27 xcat[13635]: xCAT: Allowing lsdef for root from localhost
Feb 21 21:25:16 c910f03c17k27 xcat[13645]: xCAT: Allowing lsxcatd -v for root from localhost
Feb 21 21:25:18 c910f03c17k27 xcat[13676]: xCAT: Allowing chdef -t node -o c910f03c17k28 arch=ppc64le netboot=grub2 vmcpus=4 vmhost=c910f03c17 serialport=0 serialspeed=115200 os=rhels7.3 mgt=kvm groups=all id=28 serialflow=hard vmstorage=phy:/dev/mapper/vdiskvg03-vdisk01n28 vmnics=brenP3p9s0f0 cons=kvm vmmemory=6144 vmnicnicmodel=virtio for root from localhost
```

zet809:
@hu-weihua Will you pls help to take a look at this issue? Thx!

PR #2530, opened 2017-02-22
Fix bug #2528 2.13.2 install xcat Error loading module /opt/xcat/lib/perl/xCAT_plugin/rhevm.pm

hu-weihua:


immarvin:
agree to merge

Issue #2501, opened 2017-02-17
"pping" command flag "-f" displays an error


gurevichmark:
xCAT man page for pping shows that `-f | --use_fping` option can be used.

When trying it, I get:

```
[root@fs4 ~]# pping -f frame23
Cannot open fping pipe: No such file or directory at /opt/xcat/bin/pping line 165.
[root@fs4 ~]# pping --use_fping frame23
Cannot open fping pipe: No such file or directory at /opt/xcat/bin/pping line 165.
[root@fs4 ~]#
```

gurevichmark:
Assigning to @zet809 so that he can assign this issue to the right person.

whowutwut:
@zet809 if you think someone can fix this in this release please assign it. Otherwise let's move it to 2.13.3.  @tingtli can you verify if we have any test cases that test this?

zet809:
It seems the fping package shall be installed to support fping functions. 
[root@xcat]# rpm -aq |grep fping
[root@xcat]# pping -f frame23
Cannot open fping pipe: No such file or directory at /opt/xcat/bin/pping line 165.
[root@xcat]# yum install fping
...
[root@xcat]# rpm -aq |grep fping
fping-2.4b2_to-2.ppc64le
[root@xcat]# pping -f frame23
frame23cn03: ping
frame23cn05: ping
frame23cn16: ping
frame23cn17: ping
...

PR #2531, opened 2017-02-22
Return error if no /usr/bin/fping is available

zet809:
Fix issue #2501 
Add checking that whether fping command is available, return error if not.

### DIFF
--- a/xCAT-client/bin/pping
+++ b/xCAT-client/bin/pping
@@ -162,6 +162,11 @@ sub fping_pping {
     }
     else
     {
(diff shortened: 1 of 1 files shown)


gurevichmark:
The code looks good to me. One suggestion is to change the message to something like:
"Please install missing package fping".

@zet809 if `fping` is required for this command, maybe it should just be added to "deps" package ?

zet809:
@gurevichmark I had modified the return information, pls review.
As for your question, I think there will be more effort involved if we add a new depended package. We need to maintain the build, releasing...

PR #2541, opened 2017-02-23
update cases for pdu support

tingtli:
@cxhong @junxiawang  Please help review. Thanks!

### DIFF
--- a/xCAT-test/autotest/linux.conf.template
+++ b/xCAT-test/autotest/linux.conf.template
@@ -115,6 +115,11 @@ bmcusername=USERID
 bmcpasswd=PASSW0RD
 #For bmc case u need to set username and passwd first
 #End
+#For pdu support
+PDU=f5pdu3
+PDUIP=50.0.0.8
+PORT=6
+#End
 [Custom]
 
 #commands run before test
--- a/xCAT-test/autotest/testcase/pdu/case0
+++ b/xCAT-test/autotest/testcase/pdu/case0
@@ -1,5 +1,5 @@
 start:lsdef_pdu_object
-description:list pdu object and it's attribute
+degscription:list pdu object and it's attribute
 cmd:lsdef -h
 check:output=~pdu
 cmd:mkdef $$PDU mgt=pdu nodetype=pdu groups=pdu
@@ -136,6 +136,8 @@ cmd:oldpdu=`lsdef $$CN |grep pdu|awk -F= '{print $2}'`;chdef $$CN pdu=$$PDU:$$PO
 check:rc==0
 cmd:lsdef $$CN -i pdu
 check:rc==0
+cmd:rpower $$CN stat |awk -F: '{print $2}' > /tmp/cnstat
+check:rc==0
 cmd:rpower $$CN pdustat
 check:rc==0
 check:output=~$$CN: $$PDU outlet $$PORT is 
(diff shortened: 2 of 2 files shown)


tingtli:
@cxhong  Please help review. Thanks!

xCAT-test/autotest/linux.conf.template
```diff
@@ -115,6 +115,11 @@ bmcusername=USERID
 bmcpasswd=PASSW0RD
 #For bmc case u need to set username and passwd first
 #End
+#For pdu support
+PDU=f5pdu3
+PDUIP=50.0.0.8
+PORT=6
+#End
```
cxhong: do u have a way to change this hard code of machine info?  the ip address maybe changed and most likely outlet connection to compute node maybe change too.

xCAT-test/autotest/testcase/pdu/case0
```diff
@@ -0,0 +1,158 @@
...
+check:output=~$$CN: $$PDU outlet $$PORT is off
+cmd:rpower $$CN pduon
+check:rc==0
+check:output=~$$CN: $$PDU outlet $$PORT is on
+cmd:oldpdu=`cat /tmp/pduvalue`;chdef $$CN pdu=$oldpdu
+cmd:rmdef $$PDU
+check:rc==0
+cmd:rm -rf /tmp/pduvalue
+end
+
```
whowutwut: Remove these extra spaces at the bottom please

xCAT-test/autotest/linux.conf.template
whowutwut: Do we have any test cases that tests the error conditions? 
* if the pdu config is not set up correctly, wrong outlet number, like outlet 100? 
* IF we have a bad config, rpower <cn> pduoff, do we check the rc of rpower?  to make sure it blocks...

xCAT-test/autotest/testcase/pdu/case0
```diff
@@ -0,0 +1,158 @@
...
+description:Verify rpower works for pdu object
+cmd:mkdef $$PDU groups=pdu ip=$$PDUIP mgt=pdu nodetype=pdu
+check:rc==0
+cmd:makehosts $$PDU
+check:rc==0
+cmd:grep $$PDUIP /etc/hosts
+check:rc==0
+cmd:snmpwalk -v1 -cpublic -mALL $$PDU .1.3.6.1.2.1.1
+check:rc==0
+cmd:rpower $$PDU off
```
whowutwut: Which PDU are you running this on? This would affect many compute nodes that are tied to it....

xCAT-test/autotest/testcase/pdu/case0
```diff
@@ -0,0 +1,158 @@
...
+check:rc==0
+check:output=~$$CN: $$PDU outlet $$PORT is off
+cmd:rpower $$CN pduon
+check:rc==0
+check:output=~$$CN: $$PDU outlet $$PORT is on
+cmd:oldpdu=`cat /tmp/pduvalue`;chdef $$CN pdu=$oldpdu
+cmd:rmdef $$PDU
+check:rc==0
+cmd:rm -rf /tmp/pduvalue
+end
```
whowutwut: Can we also turn the compute node back on at the end if the test.. essentially.. leave the compute node in the same state it was in at the start of the test.. 

 If the compute node is powered off, then the team might think this machine is broken or offline... ...

xCAT-test/autotest/testcase/pdu/case0
```diff
@@ -0,0 +1,158 @@
...
+check:output=~$$PDU
+check:output=~pdu
+cmd:tabdump pdu
+check:rc==0
+check:output=~$$PDU
+cmd:rmdef $$PDU
+check:rc==0
+end
+
+start:mkdef_pdu_object
```
whowutwut: I think this mkdef_pud_object test case is not required. we are doing mkdef in all the other tests, what does this provide that is different than the lsdef above or the noderm test below?

xCAT-test/autotest/linux.conf.template
tingtli: @cxhong  This is only an example for user's reference. When users use it, they need to modify the values.

xCAT-test/autotest/testcase/pdu/case0
tingtli: @whowutwut  The PDU  is defined in the configuration file. In this case, when I run test on fs4, it is f5pdu3. Users can define it in the configuration files.

xCAT-test/autotest/testcase/pdu/case0
tingtli: @whowutwut  We need to run mkdef for each test because we support test for single case.
For other scenarios, I will add 1 test case.

xCAT-test/autotest/linux.conf.template
tingtli: I will add a case for the error checking.

xCAT-test/autotest/testcase/pdu/case0
whowutwut: @tingtli Right, you can do rmdef single case but my point is to create all the objects once, quickly and then you can rmdef single case, rather than for each case, mkdef and them rmdef ...  just a suggestion that may help performance and testing a wider range of commands against noderanges... and various other things..  at the end you can do a wildcard...  IF you create the node defintiions unique...  

``rmdef fvt_autotest_pdu_.... ``
``rmdef /fvt_autotest_pdu.*`` will remove all to do full cleanup

tingtli:
@whowutwut  i've update the cases according to your comments. For the object operation, i will open a new bug to add some cases for the wildcase and range test.

tingtli:
@whowutwut @cxhong  If you don't have further comments for the pdu cases, would you please merge it. Thanks!

PR #2546, opened 2017-02-23
confignics: fix regexp for Ethernet checks

kcgthb:
Fix regular expressions in `confignics` that checks for Ethernet interfaces.

They missed `em1`-type interfaces, and prevented `confignics -r` to remove unused `emX` interfaces.


bybai:
Hi @kcgthb ,
This will be merge in 2.13.3 next monday. Thanks.

PR #2547, opened 2017-02-24
revert: use df instead of mount, and check for the disk that /boot is…

arif-ali:
… on, rather than /

I thought this was a problem last time, but I was wrong, so reverting the df concept here


zet809:
Hi, @arif-ali , the revert looks good, thx a lot!

Issue #2544, opened 2017-02-23
Duplicated output of postscripts in /var/log/xcat/xcat.log


kcgthb:
Hi!

It seems that postscripts execution generates duplicated output in `/var/log/xcat/xcat.log` (at least for stateful installation of RHEL 7 nodes).

```
# lsxcatd -v
Version 2.13.1 (git commit 62bd7a4072bb629ede16974272c993f4e707765f, built Wed Jan 11 08:45:41 EST 2017)
# lsdef sh-101-01 -i postscripts
Object name: sh-101-01
    postscripts=syslog,remoteshell,syncfiles,otherpkgs
```
After (successfull) deployment, when SSH'ing to the node to look at the log file, you can see that each postscript line is outputted twice:
```
# cat /var/log/xcat/xcat.log
[... postscripts cop from MN ...]
systemd 219
+PAM +AUDIT +SELINUX +IMA -APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ -LZ4 -SECCOMP +BLKID +ELFUTILS +KMOD +IDN
/etc/systemd/system/xcatpostinit1.service generated
xcatpostinit1.service enabled
Thu Feb 23 10:47:04 PST 2017 Running postscript: syslog
Thu Feb 23 10:47:04 PST 2017 Running postscript: syslog
postscript: syslog exited with code 0
Thu Feb 23 10:47:04 PST 2017 postscript syslog return with 0
Thu Feb 23 10:47:04 PST 2017 Running postscript: remoteshell
Thu Feb 23 10:47:04 PST 2017 Running postscript: remoteshell


Could not load host key: /etc/ssh/ssh_host_ed25519_key^M
Could not load host key: /etc/ssh/ssh_host_ed25519_key^M
postscript: remoteshell exited with code 0
Thu Feb 23 10:47:06 PST 2017 postscript remoteshell return with 0
Thu Feb 23 10:47:06 PST 2017 Running postscript: syncfiles
Thu Feb 23 10:47:06 PST 2017 Running postscript: syncfiles
postscript: syncfiles exited with code 0
Thu Feb 23 10:48:26 PST 2017 postscript syncfiles return with 0
Thu Feb 23 10:48:26 PST 2017 Running postscript: otherpkgs
Thu Feb 23 10:48:26 PST 2017 Running postscript: otherpkgs
[...]
```

[...]

PR #2556, opened 2017-02-24
Fix double logging of postscript output in /var/log/xcat/xcat.log

kcgthb:
Tentative fix for #2544


immarvin:
agree to merge, thanks @kcgthb

Issue #2555, opened 2017-02-24
Reusing the hardeths postscript for Cumulus Linux Switch OS added unsupported keyword


whowutwut:
Using the xCAT ``hardeths`` postscript, added in 2.13.2 #2485 ... works, but an unsupported keyword is added to the `/etc/network/interfaces` file. 

```
[root@fs3 ~]#  xdsh frame1sw1 cat /etc/network/interfaces
frame1sw1: auto eth0
frame1sw1: iface eth0 inet static
frame1sw1:     address 192.168.3.200
frame1sw1:     network 192.168.0.0  <--------------------
frame1sw1:     netmask 255.255.0.0
frame1sw1:     broadcast 192.168.255.255
frame1sw1:     gateway 192.168.27.1
frame1sw1:
frame1sw1:
frame1sw1: auto lo
frame1sw1: iface lo inet loopback
```

The error comes out when we run `ifreload -a` on the Cumulus Switch
```
frame1sw1: error: /etc/network/interfaces: line4: iface eth0: unsupported keyword (network)
```

After removing that line and running `ifreload -a`, no error message comes out...

```
root@frame1sw1:~# grep network /etc/network/interfaces
root@frame1sw1:~#
root@frame1sw1:~# ifreload -a
root@frame1sw1:~#
```

whowutwut:
I thought I had verified this a while back, but our Cumulus Switch is down right now, so will need to wait...

cxhong:
@whowutwut , yes, you verified and used this postscripts

whowutwut:
ok let's close.

PR #2565, opened 2017-02-27
Modify hardeths portscripts for Cumulus Linux Switch OS

cxhong:
For Issue #2555 ,  For cumulus switch, needs to remove "network" keywords from /etc/network/interfaces file and source /etc/network/interfaces.d/*.intf  files.


bybai:
I am OK to merge the fix.

PR #2568, opened 2017-02-28
Remove hard code of community and protocol value for snmp configuration

cxhong:
change hard code of snmp community string 
* use snmpc attribute define in the site table, if it's not define, use "public" as default for snmp community string

For configBNT scripts, get the value of switches.protocol  as login prompt, if it's not define, use "ssh" as default value.


immarvin:
agree to merge

Issue #2560, opened 2017-02-26
syncfiles postscript always exits with 0


kcgthb:
The [`syncfiles`](https://github.com/xcat2/xcat-core/blob/master/xCAT/postscripts/syncfiles) postscript seems to always exit with a return code of 0, not matter what. This could lead to a situation where the postscript fails, but still sends a message that it exited with 0. As a consequence, the user will be unaware of the failure.

```
++ /xcatpost/startsyncfiles.awk
+ returncode=1
+ '[' 1 -eq 0 ']'
+ '[' 0 -eq 0 ']'
+ quit=yes
+ let count=count-1
+ '[' yes = no ']'
+ '[' 1 -eq 0 ']'
+ logger -t xcat -p local4.err './syncfiles: Perform Syncing File action encountered error'
+ exit 0
++ ret_local=0
++ '[' 0 -ne 0 ']'
+++ date
++ msgutil_r 10.10.0.1 info 'Fri Feb 24 17:57:44 PST 2017 postscript syncfiles return with 0' /var/log/xcat/xcat.log
```
The corresponding code is:
```bash
if [ $returncode -eq 0 ]
then
  logger -t xcat -p local4.info "$0: Perform Syncing File action successfully"
else
  logger -t xcat -p local4.err "$0: Perform Syncing File action encountered error"
fi

exit 0
```
Is that the intended behavior? Shouldn't `syncfiles` return 1 in case of error?

neo954:
I would like to look into this issue.

tingtli:
This is verified by 3/1 auotest. Close it.

PR #2570, opened 2017-03-01
Fix the exit code of postscript `syncfiles`

neo954:
Fix github issue #2560.


immarvin:
agree to merge , thanks @neo954

PR #2571, opened 2017-03-01
confignics support MTU

### LINKED ISSUE #2535: MTU in networks table not carried through to static config  with confignics -s
https://github.com/xcat2/xcat-core/pull/1933 says it will "Add a MTU column to the networks table. If set, this will configure the MTU in DHCP so that nodes will have the correct MTU during provisioning. "

This works, however the MTU is lost if "confignics -s" is used to set the IP address statically.

It is possible to set this explicitly with something like:
nicextraparams.eth0=MTU=9000
 - but it would be more consistent if the MTU setting for the network were carried through to static config for the install NIC. This is particularly useful for Centos7 where network devices have a variety of different names depending on their position in the host.

bybai:
fix #2535 MTU in networks table not carried through to static config with confignics -s

support MTU in confignics.

### DIFF
--- a/xCAT/postscripts/configeth
+++ b/xCAT/postscripts/configeth
@@ -1,4 +1,4 @@
-#!/bin/bash -x 
+#!/bin/bash  
 # IBM(c) 2014 EPL license http://www.eclipse.org/legal/epl-v10.html
 # Internal script used by confignics only.
 # It configs the Ethernet adpaters on the node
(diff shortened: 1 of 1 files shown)


bybai:
Hi @immarvin 
MTU support includes:
1, If MTU is defined in networks table, confignics will atomatically add mtu value in the compute node network configuration file. MTU in networks table is the default mtu value for the specified network.
2, If MTU is also defined in the nicextraparams in nics table for noderange,  MTU value will use the value defined in the nicextraparams in nics table.  MTU defined in the nicextraparams is customized value.

xCAT/postscripts/configeth
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash 
+#!/bin/bash -x
```
whowutwut: I think we don't want to commit this -x into the code, right? This is for debug?

whowutwut:
I've pulled down these changes and have tried this on my MN node, and it seems to work, so I will merge...  @bybai  does `confignetworks` support MTU?  IF we need to configure the network on a VLAN?

PR #2569, opened 2017-02-28
[DO NOT MERGE] initial checkin openbmc.pm

gurevichmark:
Initial checkin of `mgt=openbmc` support for `rpower stat`

PR #2573, opened 2017-03-01
framework for openbmc support, support some rpower and rinv commands

xuweibj:
Supported command:
``rpower <node> status/on/off/reset/boot`` are done based on current urls.
``rinv <node> cpu/dimm/bios/all`` are done based on current urls.

All "print" for debug.

ENV configuration when test:

``export OPENBMC_DEVEL=YES`` in /etc/profile
``source /etc/profile`` load ENV

``restartxcatd``

Node Definition:
```
# lsdef redfishtest
Object name: redfishtest
    bmc=9.3.185.29
    bmcpassword=0penBmc
    bmcusername=root
    groups=test
    mgt=openbmc
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles,setupntp
```

rinv command:
```
# rinv redfishtest bios
redfishtest: BIOS is_fru : 1
redfishtest: BIOS version :
redfishtest: BIOS fru_type : SYSTEM
redfishtest: BIOS present : True
redfishtest: BIOS fault : False
```

rinv command with XCATBYPASS=1:
```
# XCATBYPASS=1 rinv redfishtest bios
$VAR1 = 'RINV_REQUEST';
$VAR2 = 'RINV_RESPONSE';
$VAR3 = 'LOGIN_REQUEST';
$VAR4 = 'LOGIN_RESPONSE';
$VAR5 = 'LOGIN_RESPONSE';
$VAR6 = 'RINV_REQUEST';

$VAR1 = 'redfishtest';
$VAR2 = {
          'password' => '0penBmc',
          'bmc' => '9.3.185.156',
          'cur_status' => 'LOGIN_REQUEST',
          'username' => 'root'
        };

redfishtest: POST https://9.3.185.156/login -d {"data": [ "root", "0penBmc" ] }
redfishtest: login_response 200 OK
redfishtest: GET https://9.3.185.156/org/openbmc/inventory/enumerate
redfishtest: rinv_response 200 OK
redfishtest: BIOS is_fru : 1
redfishtest: BIOS version :
redfishtest: BIOS fru_type : SYSTEM
redfishtest: BIOS present : True
redfishtest: BIOS fault : False
```

### DIFF
--- a/xCAT-server/lib/perl/xCAT/OPENBMC.pm
+++ b/xCAT-server/lib/perl/xCAT/OPENBMC.pm
@@ -1,0 +1,46 @@
+#!/usr/bin/perl
+## IBM(c) 2107 EPL license http://www.eclipse.org/legal/epl-v10.html
+
+package xCAT::OPENBMC;
+
+BEGIN
+{
+    $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : '/opt/xcat';
+}
+use lib "$::XCATROOT/lib/perl";
+use strict;
+use warnings "all";
+
+use HTTP::Async;
+use HTTP::Request;
+use HTTP::Headers;
+use HTTP::Cookies;
+use Data::Dumper;
+
+my $header = HTTP::Headers->new('Content-Type' => 'application/json');
+
+sub new {
+    my $async = shift;
+    $async = shift if (($async) && ($async =~ /OPENBMC/));
+    my $url = shift;
+    my $content = shift;
+    my $method = 'POST';
+
+    my $id = send_request( $async, $method, $url, $content );
+
+    return $id;
+}
+
+sub send_request {
+    my $async = shift;
+    $async = shift if (($async) && ($async =~ /OPENBMC/));
+    my $method = shift;
+    my $url = shift;
+    my $content = shift;
+
+    my $request = HTTP::Request->new( $method, $url, $header, $content );
+    my $id = $async->add_with_opts($request, {});
+    return $id;
+}
+
+1;
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -1,0 +1,599 @@
+#!/usr/bin/perl
+## IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
+
+package xCAT_plugin::openbmc;
+
+BEGIN
+{
+    $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : '/opt/xcat';
+}
+use lib "$::XCATROOT/lib/perl";
+use strict;
+use warnings "all";
+
+use HTTP::Async;
+use HTTP::Cookies;
+use xCAT::OPENBMC;
+use xCAT::Utils;
+use xCAT::Table;
+use xCAT::Usage;
+use xCAT::SvrUtils;
+use File::Basename;
+use Data::Dumper;
+use JSON;
+
+#-------------------------------------------------------
+
+=head3  handled_commands
+
+  Return list of commands handled by this plugin
+
+=cut
+
+#-------------------------------------------------------
+sub handled_commands {
+    return {
+        rpower => 'nodehm:mgt',
+        rinv    => 'nodehm:mgt',
+    };
+}
+
+#-------------------------------------------------------
+
+# The hash table to store method and url for request, 
+# process function for response 
+
+#-------------------------------------------------------
+my %status_info = (
+    LOGIN_REQUEST      => {
+        method         => "POST",
+        init_url       => "/login",
+    },
+    LOGIN_RESPONSE     => {
+        process        => \&login_response,
+    },
+
+    RPOWER_ON_REQUEST  => {
+        method         => "POST",
+        init_url       => "/power/on",
+    },
+    RPOWER_ON_RESPONSE => {
+        process        => \&rpower_response,
+    },
+    RPOWER_OFF_REQUEST  => {
+        method         => "POST",
+        init_url       => "/power/off",
+    },
+    RPOWER_OFF_RESPONSE => {
+        process        => \&rpower_response,
+    },
+    RPOWER_RESET_REQUEST  => {
+        method         => "POST",
+        init_url       => "/power/reset",
+    },
+    RPOWER_RESET_RESPONSE => {
+        process        => \&rpower_response,
+    },
+    RPOWER_STATUS_REQUEST  => {
+        method         => "GET",
+        init_url       => "/org/openbmc/settings/host0",
+    },
+    RPOWER_STATUS_RESPONSE => {
+        process        => \&rpower_response,
+    },
+
+    RINV_MTHBRD_REQUEST => {
+        method         => "GET",
+        init_url       => "/org/openbmc/inventory/system/chassis/motherboard",
+    },
+    RINV_MTHBRD_RESPONSE => {
+        process        => \&rinv_response,
+    },
+    RINV_CPU_REQUEST => {
+        method         => "GET",
+        init_url       => "/org/openbmc/inventory/system/chassis/motherboard/",
+    },
+    RINV_CPU_RESPONSE => {
+        process        => \&rinv_response,
+    },
+    RINV_DIMM_REQUEST => {
+        method         => "GET",
+        init_url       => "/org/openbmc/inventory/system/chassis/motherboard/",
+    },
+    RINV_DIMM_RESPONSE => {
+        process        => \&rinv_response,
+    },
+);
+
+
+#-----------------------------
+
+=head3 %node_info
+
+  $node_info = (
+      $node => {
+          bmc        => "x.x.x.x",
+          username   => "username",
+          password   => "password",
+          cur_status => "LOGIN_REQUEST",
+          cur_url    => "",
+          method     => "",
+          back_urls   => (),
+          $src => "",
+      },
+  );
+
+  'cur_url', 'method', 'back_urls' used for path has a trailing-slash
+
+=cut
+
+#-----------------------------
+my %node_info = ();
+
+my %next_status = ();
+
+my %handle_id_node = ();
+
+my $wait_node_num;
+
+my $async;
+
+my $cookie_jar;
+
+my $callback;
+
+#-------------------------------------------------------
+
+=head3  preprocess_request
+
+  preprocess the command
+
+=cut
+
+#-------------------------------------------------------
+sub preprocess_request {
+    my $request = shift;
+    if (defined $request->{_xcatpreprocessed}->[0] and $request->{_xcatpreprocessed}->[0] == 1) {
+        return [$request];
+    }
+
+    $callback  = shift;
+
+    my $command   = $request->{command}->[0];
+    my $noderange = $request->{node};
+    my $extrargs  = $request->{arg};
+    my @exargs    = ($request->{arg});
+    my @requests;
+
+    if (ref($extrargs)) {
+        @exargs = @$extrargs;
+    }
+    my $usage_string = xCAT::Usage->parseCommand($command, @exargs);
+    if ($usage_string) {
+        $callback->({ data => [$usage_string] });
+        $request = {};
+        return;
+    }
+
+    my $parse_result = parse_args($command, $extrargs);
+    if (ref($parse_result) eq 'ARRAY') {
+        $callback->({ error => $parse_result->[1], errorcode => $parse_result->[0] });
+        $request = {};
+        return;
+    }
+
+    my $sn = xCAT::ServiceNodeUtils->get_ServiceNode($noderange, "xcat", "MN");
+    foreach my $snkey (keys %$sn) {
+        my $reqcopy = {%$request};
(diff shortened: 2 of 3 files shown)


whowutwut:
@xuweibj   I will close @gurevichmark pull request https://github.com/xcat2/xcat-core/pull/2569 since you are further along, but I'd like to point out the exercise that we were doing initially,  

We do want to print out debug messages showing which endpoint we are connecting against... so it would be useful to have that set up.  

I'd also like to block the functions from running unless we do something like `export OPENBMC_DEVEL=TRUE`.  Then in our development and UT, we would export this to allow commands to run.  Otherwise it would simply print "This function is not supported at this time"  This allows us to ship this new plugin and have it turned off.  

Can you add those framework in additional commits in this PR?

xuweibj:
@chenglch @immarvin @hu-weihua Please help to give some comments, but please do not merge.
Thank you!

whowutwut:
@xuweibj  After a review has started, could we avoid using the ``--amend`` option for commit.  This will allow us to save some time by seeing the delta instead of looking at all the changes again.

whowutwut:
Can you show the output when doing the rpower command and "XCATBYPASS=TRUE" in the example above in the description?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -0,0 +1,599 @@
...
+    RPOWER_RESET_REQUEST  => {
+        method         => "POST",
+        init_url       => "/power/reset",
+    },
+    RPOWER_RESET_RESPONSE => {
+        process        => \&rpower_response,
+    },
+    RPOWER_STATUS_REQUEST  => {
+        method         => "GET",
+        init_url       => "/org/openbmc/settings/host0",
```
whowutwut: The `/org/openbmc' should be put into a variable so we can easily change this if needed in the future...

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -0,0 +1,599 @@
...
+            return ([ 1, "No option specified for rpower" ]);
+        }
+
+        if (scalar(@ARGV) > 1) {
+            return ([ 1, "Only one option is supportted at the same time" ]);
+        } 
+
+        my $subcommand = $ARGV[0];
+
+# now, only support status, delete when other command supported
```
whowutwut: This section should be indented here..

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -0,0 +1,599 @@
...
+if ($subcommand ne "status" and $subcommand ne "state" and $subcommand ne "stat") {
+    return ([ 1, "Only support status check currently" ])
+}
+#----------------------------------------------------------------
+
+        if ($subcommand eq "on") {
+            $next_status{LOGIN_RESPONSE} = "RPOWER_ON_REQUEST";
```
whowutwut: I think we should create some functions for this... 
the LOGIN code should be call from within the other functions instead of here on this if/else tests and we should keep track of whether we have already authenticated, and not try to login again... 

so if we had... 
```
login() {
   if logged in, return 
    log in
} 

on() {
   login()
   stuff....
} 
off() { 
    login()
    stuff....
} 
```
Then things like boot... we can do...
```
boot() { 
   login()
   off()
   on() 
} 
```

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -0,0 +1,599 @@
...
+            $next_status{RINV_MTHBRD_REQUEST} = "RINV_MTHBRD_RESPONSE";
+            $next_status{RINV_MTHBRD_RESPONSE} = "RINV_CPU_REQUEST";
+            $next_status{RINV_CPU_REQUEST} = "RINV_CPU_RESPONSE";
+            $next_status{RINV_CPU_RESPONSE} = "RINV_DIMM_REQUEST";
+            $next_status{RINV_DIMM_REQUEST} = "RINV_DIMM_RESPONSE";
+        } else {
+            return ([ 1, "Only 'cpu','dimm','all' are supportted at the same time" ]);
+        }
+    }
+
```
whowutwut: Here's where I would suggest that we figure out all the commands that should be supported here and define those functions as part of the skeleton framework. From there, we create a function that checks .... 
```
unsupported() { 
    if OPENBMC_DEVEL != TRUE
        print "This command is not currently supported" 
        exit 
}
```

Then all the functions will call the unsupported() function... Now we can safely ship this code "turned off" and devel and fvt would export that OPENBMC_DEVEL=TRUE when testing this new code.

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -0,0 +1,599 @@
...
+    print Dumper(%next_status) . "\n";
```
whowutwut: This is debug code? should be shipped off?  Can we test XCATBYPASS and print it if it's set?  Then we don't need to comment and uncomment if this debug is helpful.

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -0,0 +1,599 @@
...
+        xCAT::SvrUtils::sendmsg("off", $callback, $node);
+    }
+
+    if ($node_info{$node}{cur_status} eq "RPOWER_STATUS_RESPONSE") {
+        xCAT::SvrUtils::sendmsg($response_info->{'data'}->{system_state}, $callback, $node);
+    }
+
+    if ($next_status{ $node_info{$node}{cur_status} }) {
+        if ($node_info{$node}{cur_status} eq "RPOWER_STATUS_RESPONSE") {
+            if ($response_info->{'data'}->{system_state} =~ /HOST_POWERED_ON/) {
```
whowutwut: @zet809  asked this question on the interlock call. you can also ask them in slack, there should be ENUMs that we might want to define for the status. One suggestion is to parse the yaml file and take the ENUM / description there, may be neat if we did that.. then any firmware implementation we could get the enum/description they want to express from the yaml file... ...

xCAT-server/lib/perl/xCAT/OPENBMC.pm
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/perl
+## IBM(c) 2107 EPL license http://www.eclipse.org/legal/epl-v10.html
+
+package xCAT::OPENBMC;
```
immarvin: the file name and package name should better be in lower case

whowutwut:
@xuweibj  Thanks for adding the output above! 

I'm wondering what is the value of printing the following :
```
$VAR1 = 'RINV_REQUEST';
$VAR2 = 'RINV_RESPONSE';
$VAR3 = 'LOGIN_REQUEST';
$VAR4 = 'LOGIN_RESPONSE';
$VAR5 = 'LOGIN_RESPONSE';
$VAR6 = 'RINV_REQUEST';
```

Rather, I think we should print the endpoint that we are trying to go against, for example: 
```
Endpoint:  GET /org/openbmc/NetworkManager/Interface/enumerate
Response: 200 OK
```
I think we might be able to tell what we are doing based on the GET/POST/.. etc.. , the URL endpoint... and the response code...

xCAT-server/lib/xcat/plugins/openbmc.pm
whowutwut: The yaml is part of https://github.com/openbmc/phosphor-dbus-interfaces 
I haven't found a good example to look at.. but just pointing you to this FYI

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -0,0 +1,546 @@
...
+        $bmcip = $node_info{$node}{bmc};
+        $login_url = "https://$bmcip/login";
+        $content = '{"data": [ "' . $node_info{$node}{username} .'", "' . $node_info{$node}{password} . '" ] }';
+        $handle_id = xCAT::OPENBMC->new($async, $login_url, $content); 
+        $handle_id_node{$handle_id} = $node;
+        $node_info{$node}{cur_status} = $next_status{ $node_info{$node}{cur_status} };
+    }  
+
+    while (1) { 
+        last unless ($wait_node_num);
```
xuweibj: Yes, it worked.
If one node does not have any next status, $wait_node_num will minus 1. If all nodes have done, $wait_node_num will be '0'.

xuweibj:
@whowutwut 

```
$VAR1 = 'RINV_REQUEST';
$VAR2 = 'RINV_RESPONSE';
$VAR3 = 'LOGIN_REQUEST';
$VAR4 = 'LOGIN_RESPONSE';
$VAR5 = 'LOGIN_RESPONSE';
$VAR6 = 'RINV_REQUEST';
```
This is used like status machine.
For this scenario, status machine is ``login request->login response ->rinv request -> rinv response``.

whowutwut:
@xuweibj  The above comment, I can't follow how you got `login request->login response ->rinv request -> rinv response`.  

From the messages printed seems like it is... 
`rinv request -> rinv response (??) -> login request -> login response -> login response (??) -> rinv request (should be response?)`.    Seems like we have a few more steps than expected or maybe i'm missing some higher logic...

whowutwut:
:+1:  Thanks for this..
```
redfishtest: POST https://9.3.185.156/login -d {"data": [ "root", "0penBmc" ] }
redfishtest: login_response 200 OK
redfishtest: GET https://9.3.185.156/org/openbmc/inventory/enumerate
redfishtest: rinv_response 200 OK```

Can we add some header so that we can easily distinguish normal output and debug output? 

`<node name>: DEBUG POST http....`

Then we can do things like ... `XCATBYPASS=1; rinv redfishtest bios | grep DEBUG`

**Comment by xuweibj** [comment]
@whowutwut 

About this hash. For each node, initial status is ``LOGIN_REQUEST``. In this hash table ``$VAR3`` is it, so get its next status is ``$VAR4``. So the status machine is 3->4(5)->6(1)->2.

```
$VAR1 = 'RINV_REQUEST';
$VAR2 = 'RINV_RESPONSE';
$VAR3 = 'LOGIN_REQUEST';
$VAR4 = 'LOGIN_RESPONSE';
$VAR5 = 'LOGIN_RESPONSE';
$VAR6 = 'RINV_REQUEST';
```

**Comment by whowutwut** [comment]
@xuweibj  What is the scope for this pull request? It seems like you are adding more and more function in each day.. this is not the purpose here...   I'd like to have a defined scope for this PR and we can merge it in, Is it for framework?  If so, then let's scaffold in the framework which will allow mode developers to co develop functions. 

I'd like to implement the OPENBMC_DEVEL environment variable that will turn on and off this support.  I don't want to merge this code in without this switch.   Then we are able to scaffold the framework and actually have the all `handled_commands` defined... and just "turned off" or blocked.  After they are ready and tested by FVT, we remove the call to the function

To bypass this check during dev and fvt, we would export this variable and it will turn everything on, but this allows us to ship code turned off and protect the users from accidentally running it.

**Comment by xuweibj** [comment]
@whowutwut I have added ``OPENBMC_DEVEL`` environment variable and updated description. 
All commands are based on the current investigation. And for ``rinv``, did not filter information now, just show what got. 
Please help to review again. Thank you.

File: xCAT-server/lib/perl/xCAT/OPENBMC.pm (hunk 7 above)
Comment (whowutwut) [inline xCAT-server/lib/perl/xCAT/OPENBMC.pm]: @xuweibj  can you respond to this?

File: xCAT-server/lib/xcat/plugins/openbmc.pm [hunk 9]
```diff
@@ -0,0 +1,604 @@
...
+#-------------------------------------------------------
+sub preprocess_request {
+    my $request = shift;
+    if (defined $request->{_xcatpreprocessed}->[0] and $request->{_xcatpreprocessed}->[0] == 1) {
+        return [$request];
+    }
+
+    $callback  = shift;
+
+#-------------------------------------------------------
```
whowutwut: My idea here is that we would have function that we can call to enable and disable at a more fine control.  
```
sub unsupported { 
    if ($::OPENBMC_DEVEL ne "YES") {
        xCAT::SvrUtils::sendmsg("This function is currently not supported\n", $callback);
        return;
    } else {
        xCAT::SvrUtils::sendmsg("Warning: Currently running development code, use at your own risk\n",  $callback);
    }
}   
```

Then we would call `unsupported()` in all the functions that are planned to be supported in the parse_args..  this is part of the "scaffold framework code" that should have been done first...  @zet809  ...   And ONLY after we have FVT run through the commands testing ... do we open it up by removing the call to `unsupported()` and make a note in the release notes. .  This is how we can roll out support in phases and control it exactly.

xCAT-server/lib/perl/xCAT/OPENBMC.pm
xuweibj: @whowutwut , I have talked with yang song about this and decided that don't modify it.

whowutwut:
If ready to merge in, please change the title...

whowutwut:
Also I think @zet809  and @gurevichmark  may not be actively reviewing...  If not, please remove them from the reviewer list so we do not hold up the review/merge of the PR

zet809:
Looks good to me, thx a lot. One potential issue is the "next_status" won't work on Service Node. Pls modify.

whowutwut:
Linking PR to Parent Issue #2601

PR #2585, opened 2017-03-03
remove stop lable to have a full regression

tingtli:


bybai:
agree to merge. thanks

PR #2590, opened 2017-03-03
[go-xcat] Better Linux distro version detection

neo954:
Fix Linux distro version detection on RHEL 6.9 Beta.


immarvin:
agree to merge

PR #2595, opened 2017-03-03
Modify service node instructions

gurevichmark:
Update service node instructions so that postscripts from service group get included into the postscripts of the compute node.

### DIFF
--- a/docs/source/advanced/hierarchy/define_service_nodes.rst
+++ b/docs/source/advanced/hierarchy/define_service_nodes.rst
@@ -28,6 +28,11 @@ The following table illustrates the cluster being used in this example:
 |                      | r2n10                |
 +----------------------+----------------------+
 
+
+#. Modify ``site`` table attribute to include **service** group's postscripts in compute node definition: ::
+
+       chdef -t site hierarchicalattrs="postscripts"
+
 #. Select the compute nodes that will become service nodes 
      
         The first node in each rack, ``r1n01`` and ``r2n01``, is selected to become the xCAT service nodes and manage the compute nodes in that rack


docs/source/advanced/hierarchy/define_service_nodes.rst
```diff
@@ -28,6 +28,11 @@ The following table illustrates the cluster being used in this example:
 |                      | r2n10                |
 +----------------------+----------------------+
 
+
+#. Modify ``site`` table attribute to include **service** group's postscripts in compute node definition: ::
```
chenglch: `hierarchicalattrs` is a attribute in site table to indicate which attribute( postscript, postbootscript, etc, actually it can be any node attributes) should be inherited from multiple groups. In some scenario, we do not need this step. I think this description may make user confused.

gurevichmark: Ok, I understand.
I ran into this problem when I was setting up a service node: I had a compute node already defined, and it had its own postscript. Then I tried to make this node a service node. But when I assigned that compute node to a `service` group I noticed that the `service` group postscript called `servicenode` was not being executed.
After your help, setting `hierarchicalattrs` fixed the problem, but I wanted to add this information to our docs in case a user tries to create a `service` node from a compute node that already has its own postscript or is already assigned to a group before being assigned to a `service` group.

Perhaps a wording can be changed to explain it better, such as:
In section "Assigning Compute Nodes to their Service Nodes" I can say "Run `lsdef <node>` to verify postscripts, postbootscripts, etc. from `service` group are listed for the node's attributes. If they are not listed, make sure `hierarchicalattrs` attribute is set in the site table to inherit appropriate values"

Issue #2414, opened 2017-01-16
Power8 node replacement procedure.


markprez:
When replacing a power8le node or the BMC of a power8le node, we need a procedure or tools for plugging the node in and have it discovered/provisioned using the existing node definition.

whowutwut:
I agree that we should have some documentation here on the process of replacing a OpenPower compute node when we have xCAT configured with some hardware discovery method...   Just thinking out loud, I thin a high level process is the following: 

1. Machine is identified to be replaced 

2. Set the machine's BMC to DHCP (Should we power off compute side? #2226)) 

```
rspconfig <noderange> ip=dhcp 
```

3. Clear the mac address (If the node is being replaced, a new mac address should be detected) 
Perhaps this should be a function of `nodeset .. offline`.  @gurevichmark ? 

4. Set the node to offline 

```
nodeset <noderange> offline
```


Now, when the server is replaced and the new one is connected to the power supply, the BMC should boot and obtain a DHCP IP address. Then it's business as usual for discovery... 

```
bmcdiscover --range <range> -z -w
rsetboot /node-8335.* net 
rpower /node-8335.* boot 
```

markprez:
COMMENT in step 2
1. Machine is identified to be replaced

2. **_New node's BMCs should be set to use DHCP, this step would be needed if a BMC was already configured statically  you wanted to reset it to DHCP._**

Set the machine's BMC to DHCP (Should we power off compute side? #2226))
rspconfig <noderange> ip=dhcp 

3.   Clear the mac address (If the node is being replaced, a new mac address should be detected)
Perhaps this should be a function of nodeset .. offline. @gurevichmark ?

4.  Set the node to offline
nodeset <noderange> offline

Now, when the server is replaced and the new one is connected to the power supply, the BMC should boot and obtain a DHCP IP address. Then it's business as usual for discovery...
bmcdiscover --range <range> -z -w
rsetboot /node-8335.* net

whowutwut:
@markprez  I think in our situation, we're expecting the replacement machine to be coming from MFG so the server should be configured and default to be in DHCP mode.    Step 2 was to set the current machine being replaced (going out of the frame) to DHCP in order to more easily move it to another frame/rack in the datacenter and avoid the need to manually connect by Serial cable to undo the static IP configurations.   Let me update for clarity though...

whowutwut:
[...]

PR #2597, opened 2017-03-03
Add instructions for when compute nodes are services and replaced

whowutwut:
This is to address #2414 and is only a documentation change to help document a set process.


whowutwut:
@markprez if you are OK with these instructions, we can merge.

markprez:
The intructions look good, I think everything we discussed is covered.

whowutwut:
Thanks Mark, will merge this doc change.

Issue #2557, opened 2017-02-24
sendmsg is not defined in /opt/xcat/lib/perl/xCAT/Template.pm


kcgthb:
The `sendmsg` function is not declared (nor imported) in `/opt/xcat/lib/perl/xCAT/Template.pm`

As a result, all error messages supposedly sent back to the client are lost, and trying to use `sendmsg` results in the following being logged in `/var/log/xcat/cluster.log`:
```
Undefined subroutine &xCAT::Template::sendmsg called at /opt/xcat/lib/perl/xCAT/Template.pm line 1593.
```

zet809:
@chenglch Hi, cheng long, will you pls take a look at this issue? Thx a lot!

PR #2598, opened 2017-03-06
Add work aroud to replace sendmsg which is undefine method

chenglch:
This is a techenique debet that the callback is not transfered to
the Template library correctly, this fix just print error message
to syslog.

temp-fix: #2557


immarvin:
agree to merge

PR #2603, opened 2017-03-07
find mtu from networks table

### LINKED ISSUE #2535: MTU in networks table not carried through to static config  with confignics -s

bybai:
mtu value can be found from networks table if there is no network defined in nics table;
fix #2535

### DIFF
--- a/xCAT/postscripts/configeth
+++ b/xCAT/postscripts/configeth
@@ -420,18 +420,6 @@ elif [ "$1" = "-s" ];then
     str_inst_mask=''
     str_inst_gateway=''
     str_inst_mtu=''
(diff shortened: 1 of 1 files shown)


immarvin:
agree to merge

PR #2605, opened 2017-03-07
Add security bulleting for 02/16/17 openssl advisory

whowutwut:
Add a security bulletin for openssl vulnerability https://www.openssl.org/news/secadv/20170216.txt


whowutwut:
Doc change, will merge.

Issue #2508, opened 2017-02-20
rmdef takes a very long time to remove objects in xCAT DB when the DB is populated, nodeset offline is cause


whowutwut:
Creating some dummy entries into xCAT, created 1000 node definitions very quickly...

```
[root@fs1 research92]# time mkdef c460c[0001-1000] groups=regEX_CORAL
1000 object definitions have been created or modified.

real	0m2.506s
user	0m0.256s
sys	0m0.005s
```

Removing it was another story.   I tried to first use a wildcard noderange, and  I thought xCAT hung and killed it.... 
```
root        616 142804  0 11:48 ?        00:00:00  \_ sshd: root@pts/5
root        626    616  0 11:48 pts/5    00:00:00  |   \_ -bash
root       2913    626  0 12:00 pts/5    00:00:00  |       \_ /usr/bin/perl /opt/xcat/bin/rmdef /c460c.*

[root@fs1 research92]# time rmdef /c460c.*
Killed

real	53m27.144s
user	0m0.385s
sys	0m0.021s

```

I think after the investigation, it's probably just that it was going to take a very long time to remove the definitions.    DB is Postgres 

Here's some investigation data....  

```
[root@fs1 ~]# time rmdef c460c[0001-0010]
10 object definitions have been removed.

real	0m54.834s
user	0m0.256s
sys	0m0.015s

[root@fs1 ~]# time rmdef c460c[0011-0030]
20 object definitions have been removed.

real	1m48.741s
user	0m0.267s
sys	0m0.005s
```

Removing blocks of 100 (surprising after the first 100 was removed, the time actually increased...

```
[root@fs1 ~]# time rmdef c460c[0061-0160]
100 object definitions have been removed.

real	20m38.523s
user	0m0.313s
sys	0m0.014s
[root@fs1 ~]# time rmdef c460c[0161-0260]
100 object definitions have been removed.

real	34m36.380s
user	0m0.349s
sys	0m0.037s
[root@fs1 ~]# time rmdef c460c[0261-0360]
100 object definitions have been removed.

real	34m34.777s
user	0m0.374s
sys	0m0.005s
```

The good thing here is that ``noderm`` works a lot faster but it has an entry being printed out for each node removed. (Should consider making the output less verbose..) 

```
[root@fs1 ~]# time mkdef c460c[0001-1000] groups="regEX_CORAL"
1000 object definitions have been created or modified.

real	0m2.508s
user	0m0.247s
sys	0m0.015s
[...]

PR #2606, opened 2017-03-07
Make running nodeset offline optional for rmdef

gurevichmark:
* Use `-C|--cleanup` option to run `nodeset offline` when removing node objects with `rmdef` command.
* Update man page

### DIFF
--- a/xCAT-client/pods/man1/rmdef.1.pod
+++ b/xCAT-client/pods/man1/rmdef.1.pod
@@ -22,7 +22,17 @@ This command is used to remove xCAT object definitions that are stored in the xC
 
 =item B<-a|--all>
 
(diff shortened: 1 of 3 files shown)


xCAT-client/pods/man1/rmdef.1.pod
```diff
@@ -22,15 +22,11 @@ This command is used to remove xCAT object definitions that are stored in the xC
...
-In order to use xCAT commands again, you have two options.  You can restore your database from your backup by switching to bypass mode, and running the restorexCATdb command. 
-You switch to bypass mode by setting the XCATBYPASS environmant variable.  (ex. "export XCATBYPASS=yes") 
-A second option is to run xcatconfig -d.  This will restore the initial setup of the database as when xCAT was initially installed. 
-You can then restart xcatd and run xCAT commands.  
+Clear the whole xCAT database. A backup of the xCAT definitions should be saved before using this option. Once all the data is removed the xCAT daemon will no longer work. Most xCAT commands will fail. In order to use xCAT commands again, you have two options. You can restore your database from your backup by switching to bypass mode, and running the B<restorexCATdb> command. You can switch to bypass mode by setting the XCATBYPASS environmant variable.  (ex. "export XCATBYPASS=yes") A second option is to run B<xcatconfig -d>.  This will restore the initial setup of the database as when xCAT was initially installed. You can then restart xcatd and run xCAT commands.
```
whowutwut: Picking at this line....  optional rewording..  but would suggest... 

```
Clear the whole xCAT database.  A backup of the xCAT definitions should be saved before using this option as the xCAT daemons will no longer work once cleared.  

To restore: 1) export XCATBYPASS=1 and run the B<restorexCATdb> command,  or 2) Run B<xcatconfig -d> which initializes the database the same as when xCAT was installed. 
```

gurevichmark:
Updated man page as suggested.
Looks like this now:

image

gurevichmark:
Resolves issue #2508

PR #2609, opened 2017-03-08
add pegas support

tingtli:


whowutwut:
@tingtli How does this differ from running any other RHEL 7.x testing?  If it turns out it's close enough to other rhels 7.x releases, can't we just run the testing twice for the RHEL 7.x bundle and switch the OS under it?  That would help reduce the maintenance required for these additional files

tingtli:
@whowutwut  This is not very different from running other RHELS7.x testing. The reasons why we need to change is because we use the os in the node definition as the matching method for the bundle file. And when we first do OS support, we also need to do some changes for the bundle file to remove the unnecessary cases from the bundle file.

zet809:
hi, @hu-weihua , pls help to review this pull request. Thx!

whowutwut:
@tingtli  OK..  what about making it a symlink to the other file?

whowutwut:
Looking at the files some more under `xCAT-test/autotest/bundle`  Seems like there is no changes between the architectures..  Some of them have differences based on spaces in the file or sections in different places.  

```
[vhu@victorvm7 bundle]$ diff rhels7.2_ppc64le.bundle rhels7.3_ppc64le.bundle
31d30
< 
59a59
> 
254a255,258
> copycds_iso
> copycds_n
> copycds_a
> copycds_n_a
```
Why would we not want copycds testing between rhels 7.2 and rhels 7.3? Seems like these should be identical files.  My guess is that it takes too much effort to maintain.....  

Suggesting we change them to symlinks... and only break the links if we have to do some testing specific to an OS..   For example
```
[vhu@victorvm7 test]$ ls -ltr
total 20
-rw-rw-r--. 1 vhu vhu 4067 Feb 13 15:35 rhels7_x86_64.bundle
-rw-rw-r--. 1 vhu vhu 5996 Mar 14 10:42 rhels7_ppc64le.bundle
-rw-rw-r--. 1 vhu vhu 6101 Mar 14 10:42 rhels7_ppc64.bundle

[vhu@victorvm7 test]$ ln -s rhels7_ppc64.bundle rhels7.2_ppc64.bundle
[vhu@victorvm7 test]$ ln -s rhels7_ppc64.bundle rhels7.3_ppc64.bundle

[vhu@victorvm7 test]$ ln -s  rhels7_ppc64le.bundle rhels7.2_ppc64le.bundle
[vhu@victorvm7 test]$ ln -s  rhels7_ppc64le.bundle rhels7.3_ppc64le.bundle

[vhu@victorvm7 test]$ ln -s rhels7_x86_64.bundle rhels7.2_x86_64.bundle
[vhu@victorvm7 test]$ ln -s rhels7_x86_64.bundle rhels7.3_x86_64.bundle

# the following is for PEGAS
[vhu@victorvm7 test]$ ln -s rhels7_ppc64le.bundle rhelsPegas_ppc64le.bundle
```

Seems like this will drastically cut down time when we have to modify a bundle file.   for this last change, we had to modify 3 files... 

image


Could just be 1 moving forward...

tingtli:
@whowutwut  I think there is duplicated cases for copycds in rhels7.3_ppc64le.bundle. We should remove following lines in rhels7.3_ppc64le.bundle
254a255,258
> copycds_iso
> copycds_n
> copycds_a
> copycds_n_a.

See line 50- 55 in both  rhels7.3_ppc64le.bundle and  rhels7.2_ppc64le.bundle. The bundle files has copycds cases.
 50 copycds_iso
 51 copycds_n
 52 copycds_a
 53 copycds_n_a
 54 copycds_a_err
 55 copycds_n_err

So, i guess, there maybe some mistake when we create  rhels7.3_ppc64le.bundle file.

For the symlinks, we can use synlinks but the problem is , for a new supported os, we don't have migration cases in the bundle file since it is the first time to support. 
When we support the new os for several release, we can add migration cases to that bundle.

whowutwut:
@hu-weihua If you are ok with this you can feel free to merge.  my suggestion was only to try and attempt to mimimize the need to continue to modify the files like we have been in the past to switch OS.  In case we have to switch back for some reason a RC is bad, it causes more work if our testing is really puling from Git each day.

hu-weihua:
I think the current ``xcatjktest`` script can not support bundle name ``rhelsPegas_ppc64le.bundle ``. The reason is the same as https://github.ibm.com/xcat2/jenkins/pull/80.

we need time to modify ``xcatjktest``, but it can't been down very soon. If we can achieve it during next sprint, we will use new ``xcatjktest`` to cover pegas test, if not, we can use workaround, using ``rhels7.4-ppc64le`` to test pegas temporarily until new ``xcatjktest`` be ready.

So this work item won't be finished in xCAT2.13.3 sprint 2.  I will move it to 2.13.4 first.

hu-weihua:
For pegas support, we have different test requirement, so create new pull request #2944 to cover. 
Close this one.

PR #2610, opened 2017-03-08
Add support to autodetect beta iso files

whowutwut:
I assume that we will probably need to change and finalize this a little more when we get closer to GA, but this should be sufficient to help the development teams use xCAT with the beta iso images without modifying the os images and providing options to copycds.


zet809:
@whowutwut Hi, Victor, it seems to me that with this modification, copycds doesn't need to specify -n option at all. Thx for the modification.

Issue #2616, opened 2017-03-09
hostname is linux in sles12.2 statelite


bybai:
After provisioning statelite sles12.2, the system hostname is linux.

bybai:
Hi @immarvin,
Besides the code change, the following should be added into litefile table:
 "ALL","/etc/hostname","tmpfs",,

immarvin:
hi @bybai , should be mentioned in http://xcat-docs.readthedocs.io/en/latest/guides/admin-guides/manage_clusters/ppc64le/statelite/config_statelite.html#sample-data-for-sles-statelite-setup

bybai:
Hi @immarvin , the documentation is updated.

junxiawang:
@bybai 
Now the issue still exist
After sles12.2 statelite installtion:
On management node:
xdsh c910f03c11k14 hostname 

c910f03c11k14: linux
------> could not get hostname on sles12.2 statelite installation hierarchy by nfs

bybai:
The litefile_sles.csv in autotest case is not updated as my issue described. Fixed in #3263

bybai:
Verified in autotest environment.

PR #2621, opened 2017-03-09
fix hostname linux in statelite sles12.2

bybai:
In diskless we use `hostname -s` to change hostname in sles;
So add the same way to change hostname in statelite sles;

Fix issue #2616


immarvin:
agree to merge

Issue #2540, opened 2017-02-23
[fvt]2.13.3 xcatprobe osdeploy -n NODE:If NODE using illegal string such as "-h" , osdeploy did not  handle this scenario 


junxiawang:
env:all os
build:
[root@c910f03c05k100 ~]# lsxcatd -v
Version 2.13.2 (git commit bf47335dccb7dd365fb175e219c2c325b6bd46e0, built Tue Feb 21 07:16:03 EST 2017)
[root@c910f03c05k100 ~]# 
How to reproduce:
```
[root@c910f03c05k100 ~]#  xcatprobe osdeploy -n -h
The install NIC in current server is eth0                                                                         [INFO]
All nodes to be deployed are valid                                                                                [ OK ]
-------------------------------------------------------------
Start capturing every message during OS provision process....
-------------------------------------------------------------



```

junxiawang:
verified

PR #2626, opened 2017-03-10
fix issue 2540, check noderange validity

xuweibj:
Fixes Issue #2540 

Check whether ``noderange`` of input is begin with '-'.
If begin with '-', print error message.


hu-weihua:
It's good for me to merge, thanks @xuweibj .

Issue #2599, opened 2017-03-06
[customer]the entire statelite image is rw in sles12sp2 x86_64


bybai:
From customer Lenovo (critical) The entire image is RW – For some reason, the rootimg is mounted as rw and the compute nodes can actually write to anywhere is the rootimg and change it. needless to say, this is a huge issue for statelite J.   The elilo file in the tftproot dir looks ok and the line has ro as the mount option, I am not clear on when or how it was changed to rw. 

I can reproduce it in my env.

whowutwut:
How was this reported? User list?

immarvin:
hi @whowutwut , this issue is reported through mail.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2628#pullrequestreview-26231875 by @bybai

bybai:
how to test:
genimage <statelite_osimage>
liteimg <statelite_osimage>
rinstall <node> osimage=<statelite_osimage>

PR #2628, opened 2017-03-10
fix rw rootimg statelite for sles12.2

bybai:
Fix issue #2599 the entire statelite image is rw in sles12sp2 x86_64 

Use liteimg change rw to ro for `/` dir in rootimg/etc/fstab

### DIFF
--- a/xCAT-server/lib/xcat/plugins/statelite.pm
+++ b/xCAT-server/lib/xcat/plugins/statelite.pm
@@ -991,7 +991,7 @@ sub updateFstab {
     open($tfstab, ">", "$rootimg_dir/etc/fstab");
     foreach my $line (@fsdents) {
         if ( $line =~ /^$rootfs_name/ ) {
-            $line =~ s/rw/ro/;
+            $line =~ s/\brw\b/ro/;
         }
         print $tfstab $line;
     }
--- a/xCAT-server/lib/xcat/plugins/statelite.pm
+++ b/xCAT-server/lib/xcat/plugins/statelite.pm
@@ -181,6 +181,8 @@ sub process_request {
     system("mkdir -p $rootimg_dir/xcatpost");
     system("cp -r $installroot/postscripts/* $rootimg_dir/xcatpost/");
 
+    # update rw to ro for sles
+    updateFstab($rootimg_dir,$profile,$arch);
 
     #get the root password for the node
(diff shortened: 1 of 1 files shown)


immarvin:
agree to merge

xCAT-server/lib/xcat/plugins/statelite.pm
```diff
@@ -181,6 +181,8 @@ sub process_request {
     system("mkdir -p $rootimg_dir/xcatpost");
     system("cp -r $installroot/postscripts/* $rootimg_dir/xcatpost/");
 
+    # update rw to ro for sles
```
daniceexi: If this change only for sles, we should show it in the name of subroutine and also make this subroutine only be called for sles.

xCAT-server/lib/xcat/plugins/statelite.pm
```diff
@@ -968,5 +971,33 @@ sub liteItem {
     }
 }
 
+=head3
+    updateFstab
```
daniceexi: we may add description for the new subroutine.

xCAT-server/lib/xcat/plugins/statelite.pm
```diff
@@ -968,5 +971,33 @@ sub liteItem {
...
+=cut
+
```
daniceexi: I am a little surprise for the root cause of this issue. I was thinking that the root fs won't be remounted after the mount in the initrd.

From your change, looks like the mount option of rootfs is controlled by the /etc/fstab. That means the rootfs for nfs-based statelite is remounted after chroot base on the entry in the /etc/fstab, am I right?

BTW, is the statelite for sles12 using dracut or not?

bybai: Hi @daniceexi ,
The statelite for sles12 is using dracut.

In initrd, the sysroot is ro mounted, after cleanup and switchroot, it use systemd remount, it is based on /etc/fstab.

xCAT-server/lib/xcat/plugins/statelite.pm
bybai: Hi @daniceexi , In redhat, /etc/fstab is different from sles, there is no the same line, so the function will no run in redhat.

PR #2631, opened 2017-03-10
Handle purging block or raw VM disks when removing a VM

gurevichmark:
When removing a VM with a "purge" option, we try to remove the disks that were created when a VM was being created.

Normally it is a `qcaw2` file and we can easily remove it  since it was created as part of the VM creation (`vmstorage=dir:/xxx`) . However, when VM is created with `vmstorage=phy:/dev/sdx` or `vmstorage=phy:/dev/mapper/<lv-name` or `vmstorage=/xxx/yyy.img`, it is not easy to remove these physical entities because they were not created during VM creation.
Such physical devices will have `raw` or `block` attribute values. In cases like that, do not attempt to remove the device, and just log an info message into `cluster.log` file.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -1777,6 +1777,7 @@ sub rmvm {
             if ($disktype eq "cdrom") { next; }
 
             my @driver = $disk->parentNode()->findnodes("driver");
+            unless ($driver[0]) { next; }
             my $drivertype = $driver[0]->getAttribute("type");
             if (($drivertype eq "raw") || ($disktype eq "block")) { 
                 #For raw or block devices, do not remove, even if purge was specified. Log info message.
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -1775,7 +1775,17 @@ sub rmvm {
...
+            my @driver = $disk->parentNode()->findnodes("driver");
+            unless ($driver[0]) { next; }
+            my $drivertype = $driver[0]->getAttribute("type");
+            if (($drivertype eq "raw") || ($disktype eq "block")) { 
+                #For raw or block devices, do not remove, even if purge was specified. Log info message.
+                xCAT::MsgUtils->trace(0, "i", "Not purging raw or block storage device: $disk");
+                next; 
+            }
             my $file = $disk->getAttribute("file");
+            unless ($file) { next; }
```
zet809: I think a tracing message here is needed if not file attribute found.

gurevichmark:
@zet809 Thank you for the review. Warning log message added.

zet809:
hi, @gurevichmark , the modification looks good to me, I agree to merge. Thx a lot!

Issue #2625, opened 2017-03-10
[fvt]2.13.3 rhels6.9 need to modify postscript about  "/etc/os-release"


junxiawang:
env:rhels6.9
build:
[root@c910f02c01p13 postscripts]# lsxcatd -v
Version 2.13.3 (git commit b60eaae0e6d5b3fb040e50d3bafa5c7cffe5e668, built Thu Mar  9 06:15:36 EST 2017)
[root@c910f02c01p13 
How to reproduce:
```
1.
on cn's /var/log/xcat/xcat.log
910f02c02p27: Thu Mar  9 21:21:06 EST 2017 Running postbootscript: syslog
c910f02c02p27: Thu Mar  9 21:21:06 EST 2017 postbootscript syslog return with 0
c910f02c02p27: Thu Mar  9 21:21:06 EST 2017 Running postbootscript: remoteshell
c910f02c02p27: cat: /etc/os-release: No such file or directory
c910f02c02p27:
c910f02c02p27: Thu Mar  9 21:21:09 EST 2017 postbootscript remoteshell return with 0
c910f02c02p27: Thu Mar  9 21:21:09 EST 2017 Running postbootscript: syncfiles
c910f02c02p27: cat: /etc/os-release: No such file or directory
c910f02c02p27: Thu Mar  9 21:22:24 EST 2017 postbootscript syncfiles return with 1
c910f02c02p27: Thu Mar  9 21:22:24 EST 2017 Running postbootscript: setbootfromnet
c910f02c02p27: Thu Mar  9 21:22:28 EST 2017 postbootscript setbootfromnet return with
2.
[root@c910f02c01p13 postscripts]# grep "os-release" * -r
configib:    elif [ -f /etc/os-release ] && cat /etc/os-release |grep NAME|grep Ubuntu>/dev/null
configinterface:if ! cat /etc/os-release |grep -i  '^NAME=[ "]*Cumulus Linux[ "]*$' >/dev/null 2>&1 ; then
enablesnmp:if ! cat /etc/os-release |grep -i  '^NAME=[ "]*Cumulus Linux[ "]*$' >/dev/null 2>&1 ; then
hardeths:if (cat /etc/os-release |grep -i  '^NAME=[ "]*Cumulus Linux[ "]*$' >/dev/null 2>&1); then
otherpkgs:if cat /etc/os-release |grep -i -e "^NAME=[ \"']*Cumulus Linux[ \"']*$" >/dev/null 2>&1 ; then
[...]

PR #2632, opened 2017-03-10
checking if /etc/os-release/ file is existed

cxhong:
for issue #2625 , on older OS,  the /etc/os-release files may not existed.  need to check it if it is existed first.


immarvin:
agree to merge, thanks @cxhong

Issue #2630, opened 2017-03-10
copycds issues when using -p path option


whowutwut:
@immarvin  Pull request #2050 broke `copycds -p` ...  that was addressing issue #1852.  Decided to open another issue to track this in the correct release. 

## Issue 1 

Running the command: `copycds -p /install/RHEL-7.3-20161019.0-Server/ppc64le /iso/RHEL-7.3-20161019.0-Server-ppc64le-dvd1.iso`  destroys the path... 

```
    148                 print "VKHU_DEBUG path=$path\n";
    149                 $path=Cwd::realpath($path); <====== problem line
    150                 print "VKHU_DEBUG path=$path\n";
    151                 unless(substr($path,0,length("/install")) eq "/install"){
    152                     $callback->({ warning => "copycds: the specified path \"$path\" is not a subdirectory under /install. Make sure this path is configured for httpd/apache, otherwise, the provisioning w        ith this iso will fail!" });
    153                 }
```

```
VKHU_DEBUG path=/install/RHEL-7.3-20161019.0-Server/ppc64le
Use of uninitialized value $path in concatenation (.) or string at /opt/xcat/lib/perl/xCAT_plugin/copycds.pm line 150.
VKHU_DEBUG path=
```

If I run the command: 
```
copycds -p /install/RHEL-7.3-20161019.0-Server -a ppc64le /iso/RHEL-7.3-20161019.0-Server-ppc64le-dvd1.iso
```
it puts the ISO image in `/install/RHEL-7.3-20161019.0-Server/` instead of `/install/RHEL-7.3-20161019.0-Server/ppc64le` 

## Issue 2... 
```
# copycds -p /install/RHEL-7.3-20161019.0-Server/ppc64le /iso/RHEL-7.3-20161019.0-Server-ppc64le-dvd1.iso
Warning: copycds: the specified path "" is not a subdirectory under /install. Make sure this path is configured for httpd/apache, otherwise, the provisioning with this iso will fail!
Copying media to /install/rhels7.3/ppc64le
Media copy operation successful
```

I think this should NOT be a warning, we should prevent the copycds to happen if we are not using the path provided by the user.  The path is specified here and then we default to `rhels7.3/ppc64le`

immarvin:
hi @whowutwut , the issue 1 is the designed behavior, ``copycds -p $path`` will copy the iso contents into $path, ``-a $arch`` will not be appended to $path;

[...]

PR #2635, opened 2017-03-13
fix issue copycds issues when using -p path option #2630

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2630:
1. when $path specified with "-p" does not exist, copycds will complain 
````
Warning: copycds: the specified path "" is not a subdirectory under /install. Make sure this path is configured for httpd/apache, otherwise, the provisioning with this iso will fail!
Copying media to /install/rhels7.3/ppc64le

```` 
and copy /iso contents into default path without "-p"

2.  for ``copycds -p /installm`` , copycds should complain 
````
Warning: copycds: the specified path "" is not a subdirectory under /install. Make sure this path is configured for httpd/apache, otherwise, the provisioning with this iso will fail!
````

### DIFF
--- a/xCAT-server/lib/xcat/plugins/copycds.pm
+++ b/xCAT-server/lib/xcat/plugins/copycds.pm
@@ -145,8 +145,11 @@ sub process_request {
 
             if ($path)
             {
-                $path=Cwd::realpath($path);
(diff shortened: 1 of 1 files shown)


bybai:
Ｉ　ａｇｒｅｅｄ　ｔｏ merge.

PR #2642, opened 2017-03-14
Fix rmdef.pod error that broke the build

gurevichmark:
Left out `=back`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 docs/build
 # Ignore the temporary file for creating the Release version
 Release.save*
+# ignore the old generated man pages and html pages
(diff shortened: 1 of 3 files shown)


whowutwut:
@gurevichmark  We are missing the xcat man page, should we add that into the documentation? 
`man xcat` works on the xCAT nodes, but it's missing from http://xcat-docs.readthedocs.io/en/stable/guides/admin-guides/references/man1/xcat.1.html  

was it left out intentionally?

gurevichmark:
For future reference:
The `xcat` man page gets generated by `xCAT-client/xpod2man`, starting on line 234

PR #2645, opened 2017-03-15
modify bundle list according to disk diskless case modify

junxiawang:
@tingtli  I modify bundle list for diskfull and diskless test case integration。please help review


hu-weihua:
We didn't modify the case logic, just did some combine between cases. 
There is risk to result in installation verification failure due to some error during the combine.
I don't think it is a good time to merge this PR in 2.13.3 sprint2, due to developers are focusing on xcat installation bugs. If we replace cases now, maybe involve additional overhead. 
I will merge this PR in next sprint, if have case error, we will have time to fix and won't effect release.

hu-weihua:
I have reviewed the change and agree to merge it in 2.13.4 sprint 1

PR #2646, opened 2017-03-15
change command parse to support service for openbmc

xuweibj:


zet809:
Looks good to me, I agree to merge. Thx  a lot!

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -252,17 +254,56 @@ sub parse_args {
         return $check;
     }
 
-    $next_status{LOGIN_REQUEST} = "LOGIN_RESPONSE";
+    if (!defined($extrargs)) {
+        return ([ 1, "No option specified for rpower" ]);
+    }
 
+    if (scalar(@ARGV) > 1) {
+        return ([ 1, "Only one option is supportted at the same time" ]);
```
whowutwut: Spelling error here.

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -252,17 +254,56 @@ sub parse_args {
...
+    }
+
+    my $subcommand = $ARGV[0];
     if ($command eq "rpower") {
-        if (!defined($extrargs)) {
-            return ([ 1, "No option specified for rpower" ]);
+        if ($subcommand ne "on" and $subcommand ne "off" and $subcommand ne "reset" and $subcommand ne "status" and $subcommand ne "state" and $subcommand ne "stat" and $subcommand ne "boot") {
```
whowutwut: is there a better way to do this.. perhaps an array and run through supported commands in the array??

whowutwut:
@zet809  For the pull request, we should require some descriptions be added before allowing for merge.     For simple changes, maybe OK, but it's not a good habit to get into.... 

There's nothing in the code regarding "service" so what are we actually making changes for?

whowutwut:
Thanks @xuweibj  for adding the description! :+1:

PR #2648, opened 2017-03-15
 using systemd to control xcatd

robin2008:
1, Using xcatctl to control the xcatd daemon, and the unit file just
leverage this routine.
2, xcatd reload action will create a new Process, the behaviour is not
supported in systemd fork mode
3, Modify the related spec/debian files for above changes; and it should
cover the upgrade case from no systemd unit version


immarvin:
hi @robin2008 , suggest not to create PR from the master branch of your fork, this will get in trouble when you need to create another PR while this PR is still under review. Some guidelines can be found in http://xcat-docs.readthedocs.io/en/stable/developers/github/index.html

xCAT-server/sbin/xcatctl
```diff
@@ -0,0 +1,130 @@
...
+# This avoids the perl locale warnings
+if [ -z $LC_ALL ]; then
+  export LC_ALL=C
+fi
+
+if [ -r /etc/sysconfig/xcat ]; then
+  . /etc/sysconfig/xcat
+fi
+if [ -r /etc/profile.d/xcat.sh ]; then
+  . /etc/profile.d/xcat.sh
```
immarvin: hi @robin2008 , the ``/etc/profile.d/xcat.sh`` and ``/etc/sysconfig/xcat`` are the files to specify the environment
 variables for xcatd service, it seems that ``/etc/sysconfig/xcat`` is more popular for the systemd service unit, should we combine these 2 files into one?

robin2008: Currently,   /etc/sysconfig/xcat does not exist in xCAT.  So I don't want to introduce the new profile here but just keep the compatible with old SysV ( /etc/init.d/xcatd). I don't know the cause, but the non-existing /etc/sysconfig/xcat is sourced in it.)

And this pull-request is not an enhancement on xcatd control,  just could leverage systemd unit in a systemd enabled distro.

perl-xCAT/debian/postrm
```diff
@@ -21,9 +21,8 @@ set -e
 
 case "$1" in
     upgrade)
-	if [ -x /etc/init.d/xcatd ] && [ -f "/proc/cmdline" ]; then
-	    . /etc/profile.d/xcat.sh
-	    /etc/init.d/xcatd reload
+	if [ -x /usr/sbin/xcatctl ] && [ -f "/proc/cmdline" ]; then
+	    /usr/sbin/xcatctl reload
 	fi
```
robin2008: Anyone knows why this is not required in rpm package?

immarvin: ``debian/postrm`` is for building  deb packages on ubuntu, rpm build does not use this file

robin2008: I mean, in the corresponding spec it not required to restart or reload xcatd when perl-xCAT is upgrade.
(see below spec file, it does nothing,  I just remove the redundant codes)

whowutwut:
@immarvin  should we ask @robin2008 to re-create  this pull request before getting too far along in the review.    I think this makes it quite difficult to pull the upstream changes back into your fork if you are working directly on the master branch... but I could be wrong.

whowutwut:
It also creates all those unnecessary merge comments ...  I'm going to close for now.  We can re-open if you guys disagree.

PR #2652, opened 2017-03-16
Add man page related with openbmc

xuweibj:
#2663

### DIFF
--- a/perl-xCAT/xCAT/Usage.pm
+++ b/perl-xCAT/xCAT/Usage.pm
@@ -53,6 +53,8 @@ my %usage = (
      pdu specific:
        rpower noderange [off|on|stat]
        rpower noderange [pduoff|pduon|pdustat]
+     OpenBMC (using Redfish) specific:
+       rpower noderange [on|off|reset|boot|stat|state|status]
 ",
     "rbeacon" =>
       "Usage: rbeacon <noderange> [on|off|stat] [-V|--verbose]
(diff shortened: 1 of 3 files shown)


cxhong:
if pod file is modified, need to check in the rst file too.  I used Mark's tool to do that:
````
#cd xcat-core
#./create_man_pages.py --prefix=/u/gurevich/tools
````

File: perl-xCAT/xCAT/Usage.pm [hunk 1]
```diff
@@ -53,6 +53,8 @@ my %usage = (
      pdu specific:
        rpower noderange [off|on|stat]
        rpower noderange [pduoff|pduon|pdustat]
+     OpenBMC (using Redfish) specific:
```
whowutwut: I don't know if we should specify "redfish" because we are not using redfish 1.0 specification, correct?

whowutwut: On 2nd thought, should we just combine BMC/OpenBMC together, instead of creating a new section?

xCAT-client/pods/man1/rpower.1.pod
```diff
@@ -60,6 +60,10 @@ B<rpower> I<noderange> [B<start>|B<stop>|B<restart>|B<pause>|B<unpause>|B<state>
 
 B<rpower> I<noderange> [B<stat>|B<off>|B<on>]
 
+=head2 OpenBMC (using Redfish) specific:
```
whowutwut: Same comments as with above...

perl-xCAT/xCAT/Usage.pm
zet809: Hi, @whowutwut , for rpower, the option "wake/suspend/softoff" works for ipmi, but may not supported for Redfish. To make it clear that which option is supported for OpenBMC through Redfish interface, I think it is OK to have a independent section for OpenBMC. What do you think?

whowutwut: Ok, but I don't think we should specify "Redfish" because we are not providing supporting redfish at this time, we are supporting openbmc-restful....   If we were, we could do all the development against the redfish 1.0 specification and it should work against any BMC that is redfish 1.0 compliant.  So should  change to: 

`OpenBMC specific:`

zet809: Yes, I agree.

zet809:

Issue #2624, opened 2017-03-10
[fvt]2.13.3 syncfiles could not work on hierarchy provision node


junxiawang:
env:sles11.4 rhels7.3
build:
root@c910f04x12v02:~# lsxcatd -v
Version 2.13.3 (git commit b60eaae0e6d5b3fb040e50d3bafa5c7cffe5e668, built Thu Mar  9 06:17:13 EST 2017)
root@c910f04x12v02:~# 
How to reproduce:
```
1.provison diskless node
2.on cn:
/var/log/xcat/xcat.log

c910f04x12v04: Thu Mar 9 22:36:35 UTC 2017 Running postbootscript: remoteshell
c910f04x12v04:
c910f04x12v04: Thu Mar 9 22:36:37 UTC 2017 postbootscript remoteshell return with 0
c910f04x12v04: Thu Mar 9 22:36:37 UTC 2017 Running postbootscript: syncfiles
c910f04x12v04: Thu Mar 9 22:37:57 UTC 2017 postbootscript syncfiles return with 1
c910f04x12v04: Thu Mar 9 22:37:57 UTC 2017 Running postbootscript: setupntp
3.syncfiles could not work 
```

immarvin:
hi @junxiawang , for the file sync during installation in hierarchy environment, please refer to http://xcat-docs.readthedocs.io/en/stable/guides/admin-guides/manage_clusters/common/deployment/syncfile/syncfile_install.html#hierarchy-and-service-nodes

please make sure your test cases confirm to the DOC

junxiawang:
I do not think so .For this case does not be changed for long time and syncfile could not work just from 2.13.3 from our autotest log

junxiawang:
ok I will add updatenode cn -f to set up syncfile on servicenode

immarvin:
hi @junxiawang , since your fix in the test cases has been merged, would you please verify this issue?

junxiawang:
verified

PR #2655, opened 2017-03-17
modify diskeless hierarchy testcase for issue #2624

junxiawang:
@hu-weihua I modify hierarchy testcase for issue #2624


hu-weihua:
It's good for me, thanks @junxiawang .

PR #2656, opened 2017-03-17
modify ppc64le bundle for rhels7.3

junxiawang:
@hu-weihua modify rhels7.3ppc64le bundle file for not all case should be ran on this platform

### DIFF
--- a/xCAT-test/autotest/bundle/rhels7.3_ppc64le.bundle
+++ b/xCAT-test/autotest/bundle/rhels7.3_ppc64le.bundle
@@ -149,6 +149,13 @@ nodestat_err_node
 rinv_noderange_err
 rmdef_null
 rmdef_t_err
+rpower_off
+rpower_stat
+rpower_boot
+rpower_on
+rpower_reset
+rpower_noderange
+rpower_noderange_nodeps
 rpower_err_noderange
 rscan_noderange
 rscan_x
@@ -282,6 +289,7 @@ nodeset_cmdline
 nodeset_runimg
 makentp_v
 makentp_h
+nodeset_check_warninginfo
 runcmdinstaller_h
 runcmdinstaller_command
 get_xcat_postscripts_loginfo
--- a/xCAT-test/autotest/bundle/rhels7.3_ppc64le.bundle
+++ b/xCAT-test/autotest/bundle/rhels7.3_ppc64le.bundle
@@ -149,13 +149,6 @@ nodestat_err_node
 rinv_noderange_err
 rmdef_null
 rmdef_t_err
-rpower_off
-rpower_stat
-rpower_boot
-rpower_on
-rpower_reset
-rpower_noderange
-rpower_noderange_nodeps
 rpower_err_noderange
 rscan_noderange
 rscan_x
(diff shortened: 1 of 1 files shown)


xCAT-test/autotest/bundle/rhels7.3_ppc64le.bundle
```diff
@@ -149,13 +149,6 @@ nodestat_err_node
 rinv_noderange_err
 rmdef_null
 rmdef_t_err
-rpower_off
-rpower_stat
-rpower_boot
-rpower_on
-rpower_reset
-rpower_noderange
-rpower_noderange_nodeps
```
hu-weihua: I think we need to keep rpower command

whowutwut:
I'm not sure why we are removing test cases from the coverage?  Is it because it's tested in other bundles?  @junxiawang  Can you be more specific in your pull request as to the purpose of this removal?

hu-weihua:
@whowutwut,  due to some carelessness, we added some cases which is not suitable current scenario in bundle file. Take ``rvitals_all`` for example, it is designed for physical machine, but the node in our test environment are virtual machine, so ``rvitals_all`` will always failed.  @junxiawang is trying to clean up this kind cases.  

To be honest, this is not only problem of bundle file ``rhels7.3_ppc64le.bundle``,  other bundle files maybe have same problem.  There is not a easy way to find all of this kind cases out quickly. Maybe we need find out solutions for this problem. one for short term and one for long term.

My suggestion are 
     For short term, try to find such cases manually, reduce such cases as much as possible. 
     For long term, try to find out a good and easy way to maintain bundle files. This also relate issue you brought out in PR #2609

xCAT-test/autotest/bundle/rhels7.3_ppc64le.bundle
```diff
@@ -290,7 +282,6 @@ nodeset_cmdline
 nodeset_runimg
 makentp_v
 makentp_h
-nodeset_check_warninginfo
```
hu-weihua: I don't think we should delete ``nodeset_check_warninginfo``. It is a valid case.
This case is always failed because there is designed problem of case itself. I think we should modify our case, not deleted it from bundle file.

hu-weihua: I have opened a issue #2667 to track refining.

hu-weihua:

whowutwut:
@hu-weihua  Is it possible to add a keyword to a bundle that indicates VM or BareMetal.  Then if we set a global config about the nature of the test machines, it could skip or run the test cases based on that?

hu-weihua:
@whowutwut , in fact, in our current automation environment, all the nodes are VM( VM for x86 and pp64le, lpar for ppc64). So our automation environment can't cover any test against BareMetal or specific hardware, such like PDU and switch. I plan to write a document about this situation, category which part can be automated entirely now, which part need to be automated partly and which part need to be tested manually. In the future, we keep trying to increase the part which can be automated entirely, and reduce the part need to test manually.  But now, the reality is not perfect.

Issue #2586, opened 2017-03-03
[DEV] convert /etc/init.d/xcatd from sysVinit service unit to systemd service unit


immarvin:
currently, we depended on the backward compatibility of  systemd to manipulate the xcatd service, this caused some problem like https://github.com/xcat2/xcat-core/issues/2322

immarvin:
since this feature might influence the management of xcatd service, should better assigned to some expert on linux administration @neo954 @robin2008

whowutwut:
:+1:  I think this is something that has to be done.. thanks for opening this issue.

immarvin:
hi @whowutwut , seems the xcat postscript service in systemd has been finished in this PR https://github.com/xcat2/xcat-core/pull/2055

samveen:
@immarvin Victor is talking about converting the  xCAT-server init script `/etc/init.d/xcatd` to a proper systemd unit file, instead of depending on systemd's legacy support for init.d scripts.

samveen:
@immarvin #2055 just creates a unit file for xCAT `compute` nodes, not the xCAT `master`/`service` nodes.

immarvin:
hi @samveen , yes, the systemd unit of service on compute nodes to run PS and report node status has been finished. this issue is on the systemd service unit for xcatd itself

robin2008:
Main sub-tasks for this feature:
1, Build the service type unit file for xcatd.service to control xcatd daemon via systemd
2,Handle the systemd unit in install/uninstall/upgrade case for xCAT packages
3,Check if any codes or sample codes in xcat which is using legacy SysV init method via hard coding, and correct them if possible. (Note:  this is an interface change to external user and component)
4,Update document to cover the changes  

Assumption:
Both MN and SN should be covered
It should work on all OS which systemd is enabled by default.

robin2008:
In /etc/init.d/xcatd,  there are 4 action supported: status, start, stop, restart, reload.
To keep the compatible, we need to support them all in systemd service unit file.

If installed systemd unit file,  it is greatly recommended to use "systemctl" to control the service.
But "service xcatd <operation>" is still working as it will redirected to systemctl. 

However the output of systemd is totally different with SysV init, it requires that the usage of xcatd should be refined if any usage rely on the program output. 

#For example:

systemctl status xcatd
xcatd.service - xCAT management service
   Loaded: loaded (/usr/lib/systemd/system/xcatd.service; enabled)
[...]

PR #2660, opened 2017-03-17
(#2586) xcatd supports systemd

robin2008:
Resolves issue #2586
1, Using xcatctl to control the xcatd daemon, and the unit file just
leverage this routine.
2, xcatd reload action will create a new Process, the behaviour is not
supported in systemd fork mode
3, Modify the related spec/debian files for above changes; and it should
cover the upgrade case from no systemd unit version

### DIFF
--- a/perl-xCAT/debian/postrm
+++ b/perl-xCAT/debian/postrm
@@ -21,9 +21,8 @@ set -e
 
 case "$1" in
     upgrade)
-	if [ -x /etc/init.d/xcatd ] && [ -f "/proc/cmdline" ]; then
-	    . /etc/profile.d/xcat.sh
-	    /etc/init.d/xcatd reload
+	if [ -x /usr/sbin/xcatctl ] && [ -f "/proc/cmdline" ]; then
+	    /usr/sbin/xcatctl reload
 	fi
     ;;
     purge|remove|failed-upgrade|abort-install|abort-upgrade|disappear)
--- a/perl-xCAT/perl-xCAT.spec
+++ b/perl-xCAT/perl-xCAT.spec
@@ -136,14 +136,6 @@ fi
 %endif
 
 %post
-%ifos linux
-if [ "$1" -gt 1 ]; then #Ugrade only, restart daemon and migrate settings
-   if [ -x /etc/init.d/xcatd ] && [ -f "/proc/cmdline" ]; then
-      . /etc/profile.d/xcat.sh
-   fi
-fi
-%endif
-exit 0
 
 %changelog
 * Wed May 2 2007 - Norm Nott nott@us.ibm.com
--- a/xCAT-OpenStack/debian/postinst
+++ b/xCAT-OpenStack/debian/postinst
@@ -21,9 +21,7 @@ set -e
 case "$1" in
     configure)
         if [ -f "/proc/cmdline" ];then
-            if [ -f "/opt/xcat/sbin/xcatd" ];then
-                /etc/init.d/xcatd reload
-            fi
+            [ -f "/usr/sbin/xcatctl" ] && /usr/sbin/xcatctl reload
         fi
     ;;
 
--- a/xCAT-OpenStack/xCAT-OpenStack.spec
+++ b/xCAT-OpenStack/xCAT-OpenStack.spec
@@ -87,9 +87,7 @@ rm -rf $RPM_BUILD_ROOT
 %post
 %ifos linux
   if [ -f "/proc/cmdline" ]; then   # prevent running it during install into chroot image
-    if [ -f $RPM_INSTALL_PREFIX0/sbin/xcatd  ]; then
-      /etc/init.d/xcatd restart
-    fi
+    [ -f $RPM_INSTALL_PREFIX0/sbin/xcatctl ] && $RPM_INSTALL_PREFIX0/sbin/xcatctl restart
   fi
 %endif
 exit 0
--- a/xCAT-UI/xCAT-UI.spec
+++ b/xCAT-UI/xCAT-UI.spec
@@ -184,7 +184,7 @@ ln -sf ../bin/xcatclientnnr $RPM_BUILD_ROOT%{prefix}/bin/webportal
     if [ "$1" = 1 ] || [ "$1" = 2 ]             # Install or upgrade
     then
         # Restart xCAT
-        /etc/init.d/xcatd restart
+        /usr/sbin/xcatctl restart
 
         # Copy php.ini file into /opt/xcat/ui and turn off output_buffering
         if [ -e "/etc/redhat-release" ]; then
--- a/xCAT-rmc/debian/postinst
+++ b/xCAT-rmc/debian/postinst
@@ -58,9 +58,7 @@ case "$1" in
 
       if [[ `uname -s` = "Linux" ]] ; then
          if [ -f "/proc/cmdline" ]; then   # prevent running it during install into chroot image
-           if [ -f /opt/xcat/sbin/xcatd  ]; then
-              /etc/init.d/xcatd reload
-           fi
+            [ -f "/usr/sbin/xcatctl" ] && /usr/sbin/xcatctl reload
          fi
       else
       #restart the xcatd on if xCAT or xCATsn is installed already
--- a/xCAT-rmc/xCAT-rmc.spec
+++ b/xCAT-rmc/xCAT-rmc.spec
@@ -102,9 +102,7 @@ fi
 
 %ifos linux
   if [ -f "/proc/cmdline" ]; then   # prevent running it during install into chroot image
-    if [ -f $RPM_INSTALL_PREFIX0/sbin/xcatd  ]; then
-      /etc/init.d/xcatd restart
-    fi
+      [ -f $RPM_INSTALL_PREFIX0/sbin/xcatctl ] && $RPM_INSTALL_PREFIX0/sbin/xcatctl restart
   fi
 %else
   #restart the xcatd on if xCAT or xCATsn is installed already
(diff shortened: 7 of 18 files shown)


whowutwut:
@robin2008  Sorry for the churn.  This seems to be a big change to the way we are shipping and packaging xCAT.  While we are starting sprint2, I'd be more comfortable if we created this and ran it through the FVT testing... I created a branch `2.13.systemd` , If we could push the changes there and then I can activate the daily build of that branch.  Then FVT @tingtli  can at least run this build of xCAT through the automation before we merge it into master....

samveen:
This is not actually creating a `systemd` unit. It's just moving the existing init.d script to a new location and then simulating systemd support, which is in essence no different from the existing startup system. 

It defeats the purpose of moving to systemd.

robin2008:
@samveen Thanks for your comments, and the focusing is to make the xcatd daemon be controlled by both SysV and Systemd under current existing capabilities (start/stop/..). There is no plan to improve it to cover more new features provided by Systemd.  

And could I know what's your recommendation on which new features should be supported first when going with systemd?

samveen:
I would suggest that you leave the init script as is, add a systemd service unit, and then decide at deployment time which to use.

robin2008:
I just feel a bit confusing as I think I've done what you want.  'xcatctl' is the wrapper to control xcatd (exec xcatd inside,  most of the code logic is from /etc/init.d/xcatd to keep the compatible),  then the old /etc/init.d/xcatd and a new systemd service unit file will just invoke it to do start/stop/...   During the deployment,  the package will choose to install /etc/init.d/xcatd or new systemd service unit according to current OS.

samveen:
The systemd unit does not control the xCATd process directly. Instead it uses a forking model on xcatctl, which is not recommended by the `systemd` authors (but supported for legacy reasons). Please take a look at http://0pointer.de/public/systemd-man/daemon.html#New-Style%20Daemons for recommendations about how systemd services should ideally be (this is the website of one of `systemd`'s primary authors).

robin2008:
@samveen Thanks for the reference, it is very helpful. In fact, we are not building a new style daemon. Instead to workaround with less effort to make existing xcatd work well in a systemd environment. Systemd supports many method to keep compatible with the old style. For example, with fork and PID file, it can also get the similar functions as new style daemon. 

Of course, I agree with you that a new style daemon is better here. But I think the workable solution is also acceptable with low risk.

samveen:
@robin2008 As you pointed out, we are not building a new style daemon, so then is there a need to go through all these file renames, just to add a `label` without making any structural changes to xCATd? 

What you are doing is exactly what `systemd` does automatically for `init.d` scripts ( the `systemd` legacy support creates a forking type service unit around `init.d` scripts), so why waste time and effort?

On the flip side, it is trivial to change xCAT do to fit the `simple` type service unit. The only place that the xCAT would need any changes would be trivial changes in `xcatd::daemon()`, and `print` statements elsewhere in `xcatd`. All that the parent does after `xfork()` is 
- wait for the child to report `SUCCESS` and then `exit()`.
- or if not `SUCCESS`, issue a warning and `exit(-1)`.

Further, the xCATd process does nothing to become a session owner, and the closure of the standard IO FDs is trivial to disable. Finally, a simple string replacement, changing the FD `$startupparent` to `STDERR` in `print` statements will just push the errors to STDERR, as required.

robin2008:
We can easily implemented a 'simple' type of systemd service unit via:
`ExecStart=/bin/bash -c 'source /etc/profile.d/xcat.sh; exec /usr/sbin/xcatd -p /var/run/xcatd.pid -f'`

But using 'simple' type of system unit file, it will make no fast-restart of xcatd,  Note:  "restartxcatd" is introduced in systemd environment for such requirement.
```
man restartxcatd:

         It will perform the xcatd fast restart. The xcatd fast restart is a specific restart which has two advantages compares to the stop and then start.
             1. The interval of xcatd out of service is very short.
             2. The in processing request which initiated by old xcatd will not be stopped by force. The old xcatd will hand over the sockets to new xcatd, but old xcat will
         still be waiting for the in processing request to finish before the exit.

         It does the same thing as 'service xcatd restart' on NON-systemd enabled Operating System like rh6.x and sles11.x. But for the systemd enabled Operating System like
         rh7 and sles12, the 'service xcatd restart' just do the stop and start instead of xcatd fast restart.

         It's recommended to use restartxcatd command to restart xcatd on systemd enable system like rh7 and sles12 instead of 'service xcatd restart' or 'systemctl restart
         xcatd'.
```

If we use the simple type systemd unit file, for example,  the auto-generated unit file from SysV init script,  after run  'restartxcatd',  systemctl status xcatd will report error.  (#2742)

However if we still using forking mode and PIDfile,  it works as SysV.
```
[root@xcatmn ~]# systemctl status xcatd
● xcatd.service - xCAT management service
   Loaded: loaded (/opt/xcat/etc/systemd/xcatd.service; enabled; vendor preset: disabled)
   Active: active (running) since Sat 2017-04-01 04:26:07 UTC; 22min ago
  Process: 29873 ExecStop=/usr/sbin/xcatctl stop (code=exited, status=0/SUCCESS)
  Process: 29920 ExecStart=/usr/sbin/xcatctl start (code=exited, status=0/SUCCESS)
 Main PID: 30037 (xcatd: SSL list)
   CGroup: /system.slice/xcatd.service
           ‣ 30037 xcatd: SSL listener

Apr 01 04:26:05 xcatmn systemd[1]: Starting xCAT management service...
Apr 01 04:26:07 xcatmn xcat[29962]: INFO xcatd: command log process 29962 start
Apr 01 04:26:07 xcatmn systemd[1]: Started xCAT management service.
Apr 01 04:26:27 xcatmn xcat[29961]: xcatd install monitor 29961 quiescing
Apr 01 04:26:27 xcatmn xcat[29962]: INFO xcatd: 'Command log writer' process 29962 is terminated by USR2 signal
Apr 01 04:26:27 xcatmn xcat[29960]: xcatd udp service 29960 quiescing
Apr 01 04:26:28 xcatmn xcat[29958]: xcatd main service 29958 quiescing
[root@xcatmn ~]# restartxcatd
restartxcatd invoked by root.

Restarting xCATd                                           [  OK  ]
[root@xcatmn ~]# systemctl status xcatd
● xcatd.service - xCAT management service
   Loaded: loaded (/opt/xcat/etc/systemd/xcatd.service; enabled; vendor preset: disabled)
   Active: active (running) since Sat 2017-04-01 04:26:07 UTC; 22min ago
  Process: 29873 ExecStop=/usr/sbin/xcatctl stop (code=exited, status=0/SUCCESS)
  Process: 29920 ExecStart=/usr/sbin/xcatctl start (code=exited, status=0/SUCCESS)
 Main PID: 30140 (xcatd: SSL list)
   CGroup: /system.slice/xcatd.service
           ‣ 30140 xcatd: SSL listener

Apr 01 04:26:05 xcatmn systemd[1]: Starting xCAT management service...
Apr 01 04:26:07 xcatmn xcat[29962]: INFO xcatd: command log process 29962 start
Apr 01 04:26:07 xcatmn systemd[1]: Started xCAT management service.
Apr 01 04:26:27 xcatmn xcat[29961]: xcatd install monitor 29961 quiescing
Apr 01 04:26:27 xcatmn xcat[29962]: INFO xcatd: 'Command log writer' process 29962 is terminated by USR2 signal
Apr 01 04:26:27 xcatmn xcat[29960]: xcatd udp service 29960 quiescing
Apr 01 04:26:28 xcatmn xcat[29958]: xcatd main service 29958 quiescing
Apr 01 04:48:25 xcatmn systemd[1]: xcatd.service: Supervising process 30140 which is not our child. We'll most likely not notice when it exits.
[root@xcatmn ~]# ps -ef|grep 30140
root     30140     1  0 04:48 ?        00:00:00 xcatd: SSL listener
root     30141 30140  0 04:48 ?        00:00:00 xcatd: DB Access
root     30142 30140  0 04:48 ?        00:00:00 xcatd: UDP listener
root     30143 30140  0 04:48 ?        00:00:00 xcatd: install monitor
root     30144 30140  0 04:48 ?        00:00:00 xcatd: Command log writer
root     30155 25349  0 04:48 pts/3    00:00:00 grep --color=auto 30140
[root@xcatmn ~]# kill 30140
[root@xcatmn ~]# systemctl status xcatd
● xcatd.service - xCAT management service
   Loaded: loaded (/opt/xcat/etc/systemd/xcatd.service; enabled; vendor preset: disabled)
   Active: inactive (dead) since Sat 2017-04-01 04:49:12 UTC; 5s ago
  Process: 30166 ExecStop=/usr/sbin/xcatctl stop (code=exited, status=0/SUCCESS)
  Process: 29920 ExecStart=/usr/sbin/xcatctl start (code=exited, status=0/SUCCESS)
 Main PID: 30140 (code=exited, status=0/SUCCESS)

Apr 01 04:26:05 xcatmn systemd[1]: Starting xCAT management service...
Apr 01 04:26:07 xcatmn xcat[29962]: INFO xcatd: command log process 29962 start
Apr 01 04:26:07 xcatmn systemd[1]: Started xCAT management service.
Apr 01 04:26:27 xcatmn xcat[29961]: xcatd install monitor 29961 quiescing
Apr 01 04:26:27 xcatmn xcat[29962]: INFO xcatd: 'Command log writer' process 29962 is terminated by USR2 signal
Apr 01 04:26:27 xcatmn xcat[29960]: xcatd udp service 29960 quiescing
Apr 01 04:26:28 xcatmn xcat[29958]: xcatd main service 29958 quiescing
Apr 01 04:48:25 xcatmn systemd[1]: xcatd.service: Supervising process 30140 which is not our child. We'll most likely not notice when it exits.
Apr 01 04:49:12 xcatmn xcatctl[30166]: xcatd is not running, not stopping
```

robin2008:
Also for the 'reload', currently it will restart the xcatd with special environment, it will fail too when using 'simple' type  unit.  

Without a daemon itself enhancement to cover the reload/fast-restart, a conservative method is to using forking mode service unit. It could work with the systemd on new version of OS, and no such known issue we already hit.

robin2008:
Take https://github.com/xcat2/xcat-core/pull/5394 instead.

Issue #2668, opened 2017-03-19
[DEV] Install xcat on ubuntu14.4.4+ppc64le by go-xcat show  "Restarting xcatd ERROR:". That is why Diskless_installation_flat_p8_le always fail and stay on ''powering_on``


hu-weihua:
In ubuntu14.4.4+ppc64le, install xcat by go-xcat show below error message:

    .....
    Created xCAT certificate.
    Created xCAT certificate.
    Restarting xcatd ERROR: Can not execute command /usr/sbin/in.tftpd  -v -l -s /tftpboot -m /etc/tftpmapfile4xcat.conf.
     .....

Also find the same error message in ``cluster.log``

     Mar 18 16:46:12 c910f03c17k22 xcat:  Install: syslog setup
     Mar 18 16:46:18 c910f03c17k22 xcat[21109]:  ERROR: Can not execute command /usr/sbin/in.tftpd  -v -l -s /tftpboot -m /etc/tftpmapfile4xcat.conf.
     Mar 18 16:46:18 c910f03c17k22 xcat[21159]:  INFO xcatd: command log process 21159 start
     Mar 18 16:46:21 c910f03c17k22 xcat[21376]:  xCAT: Allowing mknb ppc64 for root from localhost
      .....

So it seems we have problem of tftp service, so when doing provision bases on this kind  of xcat MN,  after reboot CN,  CN will failed at downloading through tftp and stay on "powering-on" state.

    RUN:rinstall c910f03c17k24 osimage=ubuntu14.04.4-ppc64el-netboot-compute

    [rinstall c910f03c17k24 osimage=ubuntu14.04.4-ppc64el-netboot-compute] Running Time:4 sec
    RETURN rc = 0
    OUTPUT:
    Provision node(s): c910f03c17k24
    CHECK:rc == 0   [Pass]
    CHECK:output =~ Provision node\(s\)\: c910f03c17k24     [Pass]

    RUN:sleep 200

    [sleep 200] Running Time:200 sec
    RETURN rc = 0
    OUTPUT:

    RUN:a=0;while ! `lsdef -l c910f03c17k24|grep status|grep booted >/dev/null`; do sleep 10;((a++));if [ $a -gt 100 ];then break;fi done

    [a=0;while ! `lsdef -l c910f03c17k24|grep status|grep booted >/dev/null`; do sleep 10;((a++));if [ $a -gt 100 ];then break;fi done] Running Time:1060 sec
    RETURN rc = 0
[...]

PR #2672, opened 2017-03-20
fix issue [DEV] Install xcat on ubuntu14.4.4+ppc64le by go-xcat show "Restarting xcatd ERROR:". That is why Diskless_installation_flat_p8_le always fail and stay on powering_on

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2668#issuecomment-287673025


neo954:
The patch looks good. I will merge it.

Issue #2663, opened 2017-03-17
[openbmc] Create framework skeleton for all interactions to the bmc


whowutwut:
Creating a initial list of commands that potentially interact with the BMC.  

https://github.com/xcat2/xcat-core/blob/master/xCAT-server/lib/xcat/plugins/ipmi.pm defines the following `handled_commands` 

```
sub handled_commands {
    return {
        rpower  => 'nodehm:power,mgt',
        renergy => 'nodehm:mgt',         
        getipmicons => 'ipmi',           
        rspconfig   => 'nodehm:mgt',
        rspreset    => 'nodehm:mgt', 
        rvitals     => 'nodehm:mgt',   
        rinv        => 'nodehm:mgt',    
        rflash      => 'nodehm:mgt',   
        rsetboot    => 'nodehm:mgt',
        rbeacon     => 'nodehm:mgt',
        reventlog   => 'nodehm:mgt',
        ripmi       => 'ipmi',
        getrvidparms => 'nodehm:mgt',    
        rscan => 'nodehm:mgt', 
      }
}

```

* rspconfig
* rvitals
* rinv
* rflash
* rpower
* bmcdiscover
* rspreset
* ipmi
* wvid
* winstall
* passwd
* mknb
* rcons 
* rinstall
* nodeset
* pasu
* rbeacon
* reventlog

zet809:
This task had been done with #2688 , so close this task.

PR #2688, opened 2017-03-20
Scaffold in framework for openbmc functionality

whowutwut:
This pull request partially addresses #2663 and attempts to start adding in some scaffold in the framework for the various commands that should be supported with the new openbmc management protocol.  The functions will be disabled and controlled mostly by the OPENBMC_DEVEL environment variable, even when set to YES, will still disable certain calls if it's not implemented yet. 

* b045fe79f0813cfd9b208ba91eda6903d3eb6096 - by moving unsupported calls, we enable rcons without OPENBMC_DEVEL=YES.  We want to allow users to call this function. 

* b831595b70f02b52946fefeecf12327a0d2a04e7 Identifies some function we should support based on ipmi.pm and returns unsupported
   ```
   [root@fs2vm105 opt]# rvitals openbmc1
   Error: Command is not supported.
   [root@fs2vm105 opt]# rbeacon openbmc1 on 
   Error: Command is not supported.
   ```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -1,5 +1,5 @@
 #!/usr/bin/perl
-## IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
+## IBM(c) 2017 EPL license http://www.eclipse.org/legal/epl-v10.html
 
 package xCAT_plugin::openbmc;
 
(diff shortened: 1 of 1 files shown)


whowutwut:
@zet809 @xuweibj This pull request can be merged for now.  I'm just starting to look at the openbmc code so it might take longer to think about how to scaffold more of the framework in and create some functions... but for now, this is ready to be merged.  I will create another pull request if I can see some more ways to set up the framework.  Or if you have comments, I can make some more code changes

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -38,10 +50,22 @@ sub handled_commands {
...
+        rsetboot       => 'nodehm:mgt',
+        rspconfig      => 'nodehm:mgt',
+        rvitals        => 'nodehm:mgt',
+        rflash         => 'nodehm:mgt',
+        reventlog      => 'nodehm:mgt',
+        rspreset       => 'nodehm:mgt',
+        rbeacon        => 'nodehm:mgt',
+        renergy        => 'nodehm:mgt',
+        rscan          => 'nodehm:mgt',
+        ripmi          => 'ipmi',
```
zet809: Seems rscan and ripmi is not needed.

whowutwut:
@zet809 Thanks, I removed `rscan`, `ripmi` , and `getrvidparms` in the latest commit

Issue #1001, opened 2016-04-28
[DEV] xdcp from non-root user copies files as root by default


pdlun92:
In the following example, `vhu` is xCAT enabled non-root user.

```
[vhu@fs2vm100 ~]$ xdcp fs2vm101 /tmp/victor_file /tmp/testcopy/
[vhu@fs2vm100 ~]$ xdsh fs2vm101 ls -l /tmp/testcopy
fs2vm101: total 0
fs2vm101: -rw-r--r-- 1 root root 0 Apr 28 11:46 victor_file
```

Work around is to use `-l` option to specify user

```
[vhu@fs2vm100 ~]$ xdcp fs2vm101 -l vhu /tmp/victor_file /tmp/testcopy/
[vhu@fs2vm100 ~]$ xdsh fs2vm101 ls -l /tmp/testcopy
fs2vm101: total 0
fs2vm101: -rw-r--r-- 1 vhu users 0 Apr 28 11:48 victor_file
```

@daniceexi  should xdcp issued from non-root user preserve user by default?

daniceexi:
@pdlun92 could you take a look whether the **xdsh -p|--preserve** option works for you?

pdlun92:
@daniceexi seems to be the same
```
[vhu@fs2vm100 ~]$ xdcp fs2vm101 -p /tmp/victor_file /tmp/testcopy
[vhu@fs2vm100 ~]$ xdsh fs2vm101 ls -l /tmp/testcopy
fs2vm101: total 64
fs2vm101: -rw-rw-r-- 1 root root 4 May 10 15:31 victor_file
```

whowutwut:
@neo954   I think we had talked about this on the interlock call.   The issue here is that a non root user has been given permission to use xdcp.  (distributed copy).  As a non root user, using `xdcp` to copy a file over to another node results in that file being created as the root user.  This is the bug. 

If we look at normal scp, the behavior preserves the ownership of the file.  

Taking two VMs with the `vhu` user ...

1) Create a file: 
```
[vhu@c910f04x42vm101 ~]$ touch vm101_testfile
```
2) Send the file over to vm106 as the user... 
```
[vhu@c910f04x42vm101 ~]$ scp vm101_testfile vhu@c910f04x42vm106:/tmp
Warning: Permanently added 'c910f04x42vm106,10.4.42.106' (RSA) to the list of known hosts.
vhu@c910f04x42vm106's password:
[...]

PR #2689, opened 2017-03-20
Preserve the user ownership of the file for xdcp command

gurevichmark:
Resolves Issue #1001  Use the same logic for `xdcp` as `xdsh -K` uses to save the userid of the user running the command.
The userid is saved in `DSH_FROM_USERID` environment variable by `xCAT-client/bin/xdsh`. Later it is extracted by  `perl-xCAT/xCAT/DSHCLI.pm` from that environment variable. 
It is not possible to determine the userid using `getpwuid()` inside `perl-xCAT/xCAT/DSHCLI.pm` directly, because it runs under `root`.

Also fix links, formatting and grammar in related doc pages.

### DIFF
--- a/xCAT-client/bin/xdsh
+++ b/xCAT-client/bin/xdsh
@@ -666,6 +666,24 @@ sub parse_args_xdcp
         $::NODE_RCP = 1;
     }
 
+    # find out who is the current user running xdcp
+    my $current_userid = getpwuid($>);
+
+    $ENV{DSH_FROM_USERID} = $current_userid;
+
+    my $to_userid;
+    if ($options{'user'})    # if -l option
+    {
+        $to_userid = $options{'user'};
+    }
+    else
+    {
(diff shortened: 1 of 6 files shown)


xCAT-client/bin/xdsh
```diff
@@ -666,6 +666,24 @@ sub parse_args_xdcp
...
+    $ENV{DSH_FROM_USERID} = $current_userid;
+
+    my $to_userid;
+    if ($options{'user'})    # if -l option
+    {
+        $to_userid = $options{'user'};
+    }
+    else
+    {
+        $to_userid = $current_userid;
```
whowutwut: Should we move the above code, linke 669-672 to this else clause? Only check the user if the -l option is not specified?  Looks like the DSHCLI.pm code only looks for this environment if user is not specified.

gurevichmark:
Agreed. Made suggested changes.

Issue #2687, opened 2017-03-20
openbmc unable to get default username/password from the password table


whowutwut:
The current development build: 
```
Version 2.13.3 (git commit 7bd281016189ca71a1e7c8e93774c077477330b0, built Mon Mar 20 06:15:41 EDT 2017)
```

I've defined the openbmc attribute in the `passwd` table but the code is looking only in the `openbmc` table. 

```
[root@fs2vm105 ~]# tabdump passwd | grep openbmc
"openbmc","root","XXXXXXX",,,,
```

When running rinv.... 

```
[root@fs2vm105 ~]# rinv openbmc1 cpu
Warning: Currently running development code, use at your own risk

openbmc1: Unable to get attribute username
```

Once I add the username, then it complains about the password: 

```
[root@fs2vm105 ~]# chdef openbmc1 bmcusername=root
1 object definitions have been created or modified.

[root@fs2vm105 ~]# rinv openbmc1 cpu
Warning: Currently running development code, use at your own risk

openbmc1: Unable to get attribute password
```

All the commands should default to using the password table entry and then override with the node object define.  Assign to @cxhong  and @xuweibj  . Whomever is looking at it can remove the other person.

hu-weihua:
@whowutwut , could you help to verify this fix? thanks.

whowutwut:
Sure, will try and verify today..  But this should be okay...  as the function is disabled by default to the end user at this time.

whowutwut:
Verified with the following build:
```
Version 2.13.3 (git commit 1d01f56aacf3cb33ac3e70a5845dddfad0ac5603, built Mon Mar 27 14:12:09 EDT 2017)
```

Password is not set on the object: 
```
[root@fs2vm105 ~]# lsdef openbmc2 -i bmcusername,bmcpassword
Object name: openbmc2
    bmcpassword=
    bmcusername=
```
it is set on the password table
```
[root@fs2vm105 ~]#  tabdump passwd | grep openbmc
"openbmc","root","XXXXXXX",,,,
```

running rinv works now.. 
```
[root@fs2vm105 ~]# rinv openbmc2 cpu
Warning: Currently running development code, use at your own risk.  Unset OPENBMC_DEVEL and `restartxcatd` to disable.
openbmc2: CPU0 CORE0 is_fru : 0
openbmc2: CPU0 CORE0 version :
openbmc2: CPU0 CORE0 fru_type : CORE
openbmc2: CPU0 CORE0 present : False
[...]

PR #2695, opened 2017-03-21
Get default username and password from passwd table for openbmc

xuweibj:
#2687 
If can get username and password from openbmc table, use it.
If can not get username or password from openbmc table, check passwd table.
If passwd table has, use the value in passwd table.

Modified error msg from ``xxxxxx is not supported for rpower`` to ``Unsupported command: rpower xxxxxxx``


whowutwut:
@cxhong  can you review this, was the getopenbmccons also using some code that got the password from the `passwd` table? I don't see that here, was that removed?  Should the code be shared?

cxhong:
getopenbmccons will get the values from $node_info{$node}.  the codes looks good

xuweibj:
Hi @whowutwut , Could this pull request be merged?

Issue #2492, opened 2017-02-16
When "site" attribute "sharedtftp" is false, the "nodeset" command fails


gurevichmark:
It seem that when `site` attribute `sharedtftp` is true, the `nodeset` command works as expected.
But when `sharedtftp` is false an error is displayed:

```
[root@fs4 petitboot]# lsdef -t site clustersite -i sharedtftp -c
clustersite: sharedtftp=1
[root@fs4 petitboot]# nodeset frame23cn17 osimage
frame23cn17: netboot rhels7.3-ppc64le-compute
[root@fs4 petitboot]# chdef -t site clustersite sharedtftp=0
1 object definitions have been created or modified.
[root@fs4 petitboot]# lsdef -t site clustersite -i sharedtftp -c
clustersite: sharedtftp=0
[root@fs4 petitboot]# nodeset frame23cn17 osimage
Error: petitboot plugin bug, pid 24903, process description: 'xcatd SSL: nodeset to frame23cn17 for root@localhost: petitboot instance' with error 'Can't locate object method "get_broadcast_scope" via package "xCAT::Scope" (perhaps you forgot to load "xCAT::Scope"?) at /opt/xcat/lib/perl/xCAT_plugin/petitboot.pm line 357.
' while trying to fulfill request for the following nodes: frame23cn17
[root@fs4 petitboot]#
```

gurevichmark:
Assigning to @zet809 so that he can assign this issue to the right person.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2700

immarvin:
hi @gurevichmark , this fix has been merged, please verify it if convenient, thanks

gurevichmark:
Works now!

PR #2700, opened 2017-03-21
fix issue When "site" attribute "sharedtftp" is false, the "nodeset" command fails #2492

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2492


bybai:
Missing xCAT::Scope, I agree to merge.

PR #2712, opened 2017-03-21
dodiscovery: add a fallback for getting MTM from DMI

kcgthb:
In some cases, FRU pages are not available, or don't contain relevant
readable or relevant information. In those cases, we could still fall
back to DMI to get the MTM information, and log a warning that
MTMS-based discovery may not work as expected.


zet809:

Issue #2634, opened 2017-03-10
Quotes in tables' comment field break mypostscript


kcgthb:
Using simple quotes (`'`) in the comment field of a table (at least the `networks`) table will break postscripts, as the `mypostscript` template will dump table entries without escaping those simple quotes. As a result, we get mismatched quotes in `/xcatpost/mypostscript`.

For instance:
```
# lsdef -t network test_net
Object name: test_net
    mask=255.255.255.0
    net=1.2.3.4
    usercomment=don't use
```
results in the following line in `/xcatpost/mypostscript`, being generated by `tabdump(networks)` in `mypostscript.tmpl`:
```
NETWORKS_LINE1='netname=test_net||net=1.2.3.4||mask=255.255.255.0||mgtifname=||gateway=||dhcpserver=||tftpserver=||nameservers=||ntpservers=||logservers=||dynamicrange=||staticrange=||staticrangeincrement=||nodehostname=||ddnsdomain=||vlanid=||domain=||mtu=||disable=||comments=don't use'
export NETWORKS_LINE1
```
(notice the un-escaped quote in `don't`)

The following error is logged in `xcat/commands.log` on the MN when running a postscript:
```
<nodename>: //xcatpost/mypostscript: line 271: unexpected EOF while looking for matching `''
<nodename>: //xcatpost/mypostscript: line 277: syntax error: unexpected end of file
```

Single quotes should probably be escaped by the `tabdump()` function

zet809:
HI, @bybai , will you pls take a look at this issue? Thx!

bybai:
Hi @kcgthb , this will be fixed within this sprint within 3 weeks.

bybai:
Fixed in pull request #2716

hu-weihua:
@junxiawang , Could you help to verify this issue? thanks

bybai:
testcase #2769

bybai:
It is verified through autotest.

PR #2716, opened 2017-03-22
fix issue 2634 Quotes in tables' comment field break mypostscript

bybai:
fix issue #2634 Quotes in tables' comment field break mypostscript

### DIFF
--- a/xCAT-server/lib/perl/xCAT/Postage.pm
+++ b/xCAT-server/lib/perl/xCAT/Postage.pm
@@ -1585,6 +1585,7 @@ sub dump_all_attribs_in_tabs
                         # Updated on 2017-03-22 for issue 2634 Quotes in tables' comment field break mypostscript
                         # The original line is : $values .= "comments=$t";
                         # In order to fix issue 2634, change this line to : $values .= "comments=";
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/perl/xCAT/Postage.pm
```diff
@@ -1582,7 +1582,10 @@ sub dump_all_attribs_in_tabs
                 } else {
                     $values .= "$attrib=$val||";
                     if ($attrib =~ /^disable$/) {
-                        $values .= "comments=$t";
+                        # Updated on 2017-03-22 for issue 2634 Quotes in tables' comment field break mypostscript
+                        # The original line is : $values .= "comments=$t";
+                        # In order to fix issue 2634, change this line to : $values .= "comments=";
```
immarvin: should better be ``To keep the free-style of comments and avoid the issues caused by special characters in the comments, the value of comments will not be appending to network related environment variables in mypostscript``

bybai:
Hi @immarvin , I commit new changes. Thanks.

immarvin:
agree to merge,

kcgthb:
Wouldn't it be better to just escape the single quotes in the comment field rather than dropping the comments altogether?

What happens if another field (other than "comments") contains a single quote?

Issue #2717, opened 2017-03-22
[fvt]2.13.3 makehosts -d will delete erroneously entries  /etc/hosts if one node "s01" another is "s01r*"1.


junxiawang:
env:all os
build:[root@c910f02c08p05 ~]# lsxcatd -v
Version 2.13.3 (git commit 3a54d642de9f34acb279f5f540515ced5f6f90ce, built Tue Mar 21 06:15:44 EDT 2017)
[root@c910f02c08p05 ~]# 
How to reproduce:
```
1.
nodeadd s01 groups=service
nodeadd s01r1b01 groups=compute

tabdump hosts
#node,ip,hostnames,otherinterfaces,comments,disable
"compute","|s(\d+)r(\d)b(\d+)$|10.2.0.1|",,,,
"service","|s(\d+)$|10.2.0.254|",,,,

lsdef s01
Object name: s01
groups=service
ip=10.2.0.254
postbootscripts=otherpkgs
postscripts=syslog,remoteshell,syncfiles
lsdef s01r1b01
Object name: s01r1b01
groups=compute
ip=10.2.0.1
postbootscripts=otherpkgs
postscripts=syslog,remoteshell,syncfiles
2.makehosts
cat /etc/hosts
10.2.0.254 s01 s01.pok.stglabs.ibm.com
10.2.0.1 s01r1b01 s01r1b01.pok.stglabs.ibm.com
3.
makehosts -d s01
could not find s01r1b01 on /etc/hosts
```

bybai:
fixed in 2718

hu-weihua:
@junxiawang , Could you help to verify this fix. thanks

junxiawang:
verified

PR #2718, opened 2017-03-22
fix issue 2717 makehosts -d will delete erroneously entries /etc/host…

bybai:
fix issue #2717 makehosts -d will delete erroneously entries /etc/host…


immarvin:
agree to merge

thanks @bybai

Issue #2348, opened 2016-12-22
[DEV] bmcdiscover did not report password wrong when no password input for Tuleta


xuweibj:
```
# bmcdiscover --range 50.3.19.2 -z -w
node-32031302:
	objtype=node
	groups=all
	bmc=50.3.19.2
	cons=ipmi
	mgt=ipmi
```

More infomation:

```
# XCATBYPASS=1 bmcdiscover --range 50.3.19.2 -z -w
>> Sending IPMI command payload
>>    netfn   : 0x06
>>    command : 0x38
>>    data    : 0x8e 0x04

>> SENDING AN OPEN SESSION REQUEST

>> Console generated random number (16 bytes)
 65 30 7b af 9a ca 09 f8 81 a0 87 a3 66 05 42 f2
>> SENDING A RAKP 1 MESSAGE

RAKP 2 message indicates an error : illegal parameter
Error: Unable to establish IPMI v2 / RMCP+ session
<<OPEN SESSION RESPONSE
<<  Message tag                        : 0x00
<<  RMCP+ status                       : no errors
<<  Maximum privilege level            : admin
<<  Console Session ID                 : 0xa0a2a3a4
<<  BMC Session ID                     : 0x081b19ed
<<  Negotiated authenticatin algorithm : hmac_sha1
<<  Negotiated integrity algorithm     : hmac_sha1_96
<<  Negotiated encryption algorithm    : aes_cbc_128

<<RAKP 2 MESSAGE
<<  Message tag                   : 0x00
<<  RMCP+ status                  : illegal parameter
<<  Console Session ID            : 0x00000000
<<  BMC random number             : 0x00000000000000000000000000000000
<<  BMC GUID                      : 0x00000000000000000000000000000000
<<  Key exchange auth code [sha1] : 0xffff020700000000000000000000000000000000
```

xuweibj:
Verified

PR #2720, opened 2017-03-22
fix issue 2348, check error msg when run bmcdiscover

xuweibj:
#2348 
Before modified: 
When run ``bmcdiscover`` to discover tuleta, the result will like below:
```
# bmcdiscover --range 50.3.19.2 -z -w
node-32031302:
	objtype=node
	groups=all
	bmc=50.3.19.2
	cons=ipmi
	mgt=imp
```

There is no ``mtm`` and ``serial``.

When print more msg, will see ``RAKP 2 message indicates an error : illegal parameter ``.
So add error check.

Check if there is ``RAKP 2 message indicates an error : `` in return msg when run bmcdiscover.
If there is, print error msg and return.

After modified:

```
Error: Unable to establish IPMI v2 / RMCP+ session: illegal parameter for 50.3.19.2
```


zet809:
Looks good to me, thx!

PR #2721, opened 2017-03-22
Modify syntex error of case pre_deploy_sn

hu-weihua:
Just modify syntex error of case ``pre_deploy_sn``.


neo954:
Thanks @hu-weihua.

Issue #2617, opened 2017-03-09
Wrong permissions for ssh host private keys on CentOS 7


thiell:
The **remoteshell** postscript sets wrong permissions and wrong group ownership on `/etc/ssh/ssh_host_*key` files (0600), that is not compatible with hostbased authentication (for users) on CentOS 7.

At least on RHEL 7/CentOS 7, the openssh package creates ssh host key files with the group `ssh_keys` and permissions 0640 for user host-based ssh authentification to work (the setuid helper program `/usr/libexec/openssh/ssh-keysign` is used to read the keys and requires these permissions).

Expected permissions:
```
-rw-r-----. 1 root ssh_keys    227 May 19  2016 ssh_host_ecdsa_key
-rw-r--r--. 1 root root        162 May 19  2016 ssh_host_ecdsa_key.pub
-rw-r-----. 1 root ssh_keys    387 May 19  2016 ssh_host_ed25519_key
-rw-r--r--. 1 root root         82 May 19  2016 ssh_host_ed25519_key.pub
-rw-r-----. 1 root ssh_keys   1675 May 19  2016 ssh_host_rsa_key
-rw-r--r--. 1 root root        382 May 19  2016 ssh_host_rsa_key.pub
```

xCAT permissions:
```
-rw-------. 1 root root        668 Mar  8 17:24 ssh_host_dsa_key
-rw-r--r--. 1 root root        589 Mar  8 17:24 ssh_host_dsa_key.pub
-rw-------. 1 root root        227 Mar  8 17:24 ssh_host_ecdsa_key
-rw-r--r--. 1 root root        161 Mar  8 17:24 ssh_host_ecdsa_key.pub
-rw-r-----  1 root ssh_keys    387 Mar  8 17:26 ssh_host_ed25519_key
-rw-r--r--  1 root root         82 Mar  8 17:26 ssh_host_ed25519_key.pub
-rw-------. 1 root root       1675 Mar  8 17:24 ssh_host_rsa_key
-rw-r--r--. 1 root root        381 Mar  8 17:24 ssh_host_rsa_key.pub
```

Related link: https://bugzilla.redhat.com/show_bug.cgi?id=819896

zet809:
@immarvin Hi, Yang song, will you pls take a look at this issue? Thx!

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2724

immarvin:
[...]

PR #2724, opened 2017-03-22
fix issue Wrong permissions for ssh host private keys on CentOS 7 #2617

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2617


neo954:
Please use perl function `chown` and `chmod` to do the work.

neo954:
I will merge this one.

Issue #2693, opened 2017-03-21
rflash fails to update some nodes due to timeout.


ralphbellofatto:
we have two nodes out of our 150+ node cluster that that rflash failed to update:

```
c460c138: rflash started, please wait.......
c460c138: Error: Timeout to check the bmc status
c460c138: Error: failed to update firmware

c460c142: rflash started, please wait.......
c460c142: Error: Timeout to check the bmc status
c460c142: Error: failed to update firmware
```

This appears to be due to the timeout value (which appears to be 24 5 second intervals (or 2 minutes) is too small for some nodes.

we ran a sweep of the time it takes to get the ready signal after doing an mc reset with the following code:

[bmc-reset-ready.zip](https://github.com/xcat2/xcat-core/files/856982/bmc-reset-ready.zip)

Inside this zip file is the sweep.sh program that we used to sweep the reset times, the log of its output AND a csv file of that output and an excel spreadsheet file and graph.

![image](https://cloud.githubusercontent.com/assets/4139514/24130127/337abbba-0dbd-11e7-9d1a-ce5611b78b68.png)

this data shows that there a few nodes clustered that take about 90 seconds, most take around 112-120, AND there are a couple outsiders at > 200 seconds.

This is showing that a 120 second wait time is marginal, but there is probably something off on the outliers that are preventing it from coming back ready in time.

The wait time should probably be increased to 4 or 5 minutes (48 or 60 iterations of the loop) to cover the variation we are currently seeing in the nodes that we have to date.

The code requiring change can be found here:

/opt/xcat/lib/perl/xCAT_plugin/ipmi.pm

specifically: 
```
    # check reset status
    unless (check_bmc_status_with_ipmitool($pre_cmd, 5, 24)) {
        xCAT::SvrUtils::sendmsg([ 1, "Timeout to check the bmc status" ],
            $callback, $sessdata->{node}, %allerrornodes);
        exit -1;
    }
```

Change 24 to 60 would cover the variance we are seeing with a margin for it to be about 20% worse in future units.

ralphbellofatto:
note: this is for the garision compute nodes.

whowutwut:
Hi @chenglch, @ralphbellofatto also indicated that after a firmware flash was unsuccessful, he tried to run another rflash and hit the problem with messages similar to:`` Error: Unknown FRU header version 0x00 ``  This doesn't help the user.  

[...]

Issue #2682, opened 2017-03-20
rflash reports Error: timeout and then success.


ralphbellofatto:
we have two nodes what are currently not responding to the BMC.

When i attempt to use rflash on them, we see:
```
c460c010: Error: timeout
c460c003: Error: timeout
c460c003: success to update firmware
c460c010: success to update firmware
```

This is picitulary problematic when we are attempting to use some automation on large numbers of nodes to determine that things went well...

for example:
```

c460c017: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c017.log"
c460c006: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c006.log"
c460c010: Error: timeout
c460c003: Error: timeout
c460c003: success to update firmware
c460c010: success to update firmware
c460c001: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c001.log"
c460c004: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c004.log"
c460c008: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c008.log"
c460c002: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c002.log"
c460c012: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c012.log"
c460c015: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c015.log"
c460c013: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c013.log"
c460c005: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c005.log"
c460c007: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c007.log"
c460c018: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c018.log"
c460c014: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c014.log"
c460c009: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c009.log"
c460c011: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c011.log"
c460c016: rflashing ... See the detail progress : "tail -f /var/log/xcat/rflash/c460c016.log"
c460c001: success to update firmware
```

whowutwut:
[...]

PR #2725, opened 2017-03-22
Fix timeout issue if bmc is unreachable for rflash

chenglch:
As the original ipmi code do not exit with error code, it is
not possible to judge the success or not from return code of child
procss. This patch will send message from child processes directly.
This patch also change the wait time to check whether the ipmi
session is usable and add power on at the end of rflash process.

close-issue: #2693
close-issue: #2682


zet809:
Looks good to me, thx very much for the modification.

Issue #2665, opened 2017-03-17
postscript fails and status=booted.


ralphbellofatto:
When we have a postscript that fails,. the overall boot status still shows booted rather than failed.  This makes us blind to failures in the install.

Here is what we did to demonstrate the problem.
Clone an osimage
and then setup
```
cd /install/postscripts/custom
mkdir rh73_post_test
cd rh73_post_test
cat >compute.postscript <<EOF
#/bin/bash
set -x; export PS4='+(`basename ${BASH_SOURCE}`:${LINENO}):'; 
echo "failing postscript"
exit 1
EOF

cat >compute.postbootscript <<EOF
#/bin/bash
set -x; export PS4='+(`basename ${BASH_SOURCE}`:${LINENO}):'; 
exit 0
EOF
chmod +x compute*
```

```
chdef -t osimage rh73_post_test postbootscripts=custom/rh73_post_test/compute.postbootscript
chdef -t osimage rh73_post_test postscripts=custom/rh73_post_test/compute.postscript
```

Final output:
```
[root@c460mgt01 rh73_post_test]# lsdef -t osimage rh73_post_test
Object name: rh73_post_test
    imagetype=linux
    osarch=ppc64le
    osdistroname=rhels7.3-ppc64le
    osname=Linux
    osvers=rhels7.3
    otherpkgdir=/install/post/otherpkgs/rhels7.3/ppc64le
    pkgdir=/install/RHEL-7.3-20161019.0-Server/ppc64le
    pkglist=/opt/xcat/share/xcat/install/rh/compute.rhels7.pkglist
    postbootscripts=custom/rh73_post_test/compute.postbootscript
    postscripts=custom/rh73_post_test/compute.postscript
    profile=compute
    provmethod=install
    template=/opt/xcat/share/xcat/install/rh/compute.rhels7.tmpl
```

```
nodeset c460c024 osimage=rh73_post_test
xdsh c460c024 reboot

while true; do 
  date; 
  lsdef c460c024 -i status -c; sleep 60; 
done
```

Wait until the node boots.

```
Fri Mar 17 10:46:04 EDT 2017
c460c024: status=powering-off
Fri Mar 17 10:49:05 EDT 2017
c460c024: status=installing
Fri Mar 17 10:56:08 EDT 2017
c460c024: status=booting
Fri Mar 17 11:06:12 EDT 2017
c460c024: status=booted
```
```
[...]

PR #2728, opened 2017-03-23
fix issue postscript fails and status=booted. #2665

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2665:

1. if some PS fail during diskfull installation, save the ``POSTSCRIPTS_RC=1`` into ``/opt/xcat/xcatinfo`` in the installed root filesystem; 
2. during running of PBS on 1st reboot after installation,  determine the final node provision status based on the result of PS and PBS, only if all the PBS and PS are run successfully, the ``nodelist.status`` can be "booted", otherwise, the  ``nodelist.status`` will be "failed"
3. this feature should not affect the behavior of "updatenode" and reboot on "site.runbootscripts"

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/post.xcat
+++ b/xCAT-server/share/xcat/install/scripts/post.xcat
@@ -180,6 +180,7 @@ echo "
 
 # global value to store the running status of the postbootscripts,the value is non-zero if one postbootscript failed
 return_value=0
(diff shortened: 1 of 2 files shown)


cxhong:
rerun the test in the test system.  worked fine.  will merge this PR.

Issue #2674, opened 2017-03-20
[Test case] add test case to makehosts erroneously deletes entries from /etc/hosts


bybai:
add test case for issue #2657 makehosts erroneously deletes entries from /etc/hosts

The test case should check: the **makehosts** command does not remove nodes from /etc/hosts which are a super-string of the pattern match e.g. s01r1b01 is a super string of s01.  

**nodeadd s01 groups=service
nodeadd s01r1b01 groups=compute**

**tabdump hosts**
#node,ip,hostnames,otherinterfaces,comments,disable
"compute","|s(\d+)r(\d)b(\d+)$|10.2.0.1|",,,,
"service","|s(\d+)$|10.2.0.254|",,,,

**lsdef  s01**
Object name: s01
    groups=service
    ip=10.2.0.254
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
**lsdef  s01r1b01**
Object name: s01r1b01
    groups=compute
    ip=10.2.0.1
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles

\# Creates entry in /etc/hosts for s01r1b01 as expected.
**makehosts compute
grep s01 /etc/hosts**
10.2.0.1 s01r1b01 s01r1b01.lab1.net 

\# Create s01 entry as expected, does not delete s01r1b01
**makehosts service
grep s01 /etc/hosts**
10.2.0.254 s01 s01.lab1.net 
10.2.0.1 s01r1b01 s01r1b01.lab1.net

\# However works OK, with single invocation of the command.
**makehosts compute,service
grep s01 /etc/hosts**
10.2.0.254 s01 s01.lab1.net 
10.2.0.1 s01r1b01 s01r1b01.lab1.net

bybai:
Fixed in #2729

PR #2729, opened 2017-03-23
modify makehosts testcase according to issue:add test case to makehos…

junxiawang:
add test case to makehosts erroneously deletes entries from /etc/hosts #2674 
@bybai @hu-weihua  please help review


bybai:
I agree to merge it.

hu-weihua:
It is good, thanks @junxiawang .

Issue #2727, opened 2017-03-23
[FVT]:Starting xcatd Error loading module /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm


tingtli:
xCAT 2.13.3 sles12.2 20170323 build

```
c910f03c05k23:/ # service xcatd status
● xcatd.service - LSB: xcatd
   Loaded: loaded (/etc/init.d/xcatd; bad; vendor preset: disabled)
   Active: active (running) since Thu 2017-03-23 02:49:24 EDT; 6min ago
     Docs: man:systemd-sysv-generator(8)
  Process: 11930 ExecStop=/etc/init.d/xcatd stop (code=exited, status=0/SUCCESS)
  Process: 11964 ExecStart=/etc/init.d/xcatd start (code=exited, status=0/SUCCESS)
    Tasks: 7 (limit: 512)
   CGroup: /system.slice/xcatd.service
           ├─12003 /usr/sbin/in.tftpd -v -l -s /tftpboot -m /etc/tftpmapfile4xcat.conf
           ├─12004 xcatd: SSL listener
           ├─12005 xcatd: DB Access
           ├─12006 xcatd: UDP listener
           ├─12007 xcatd: Discovery worker
           ├─12008 xcatd: install monitor
           └─12009 xcatd: Command log writer

Mar 23 02:49:22 c910f03c05k23 systemd[1]: Starting LSB: xcatd...
Mar 23 02:49:24 c910f03c05k23 xcatd[11964]: Starting xcatd Error loading module /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm  ...skipping
Mar 23 02:49:24 c910f03c05k23 xcat[11981]: Error loading module /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm  ...skipping
Mar 23 02:49:24 c910f03c05k23 xcat[12009]: INFO xcatd: command log process 12009 start
Mar 23 02:49:24 c910f03c05k23 xcatd[11964]: ..done
Mar 23 02:49:24 c910f03c05k23 systemd[1]: Started LSB: xcatd.
Mar 23 02:49:24 c910f03c05k23 xcat[12021]: xCAT: Allowing mknb ppc64 for root from localhost
Mar 23 02:51:35 c910f03c05k23 xcat[12037]: xCAT: Allowing mknb x86_64 for root from localhost
[...]

PR #2730, opened 2017-03-23
fix issue 2727, add Async path to @INC when want to use it

xuweibj:
#2727 

In docker.pm & openbmc.pm, if there is no Async.pm's path: ``/usr/local/share/perl5/``, push it to ``@INC``.


immarvin:
i am ok with the change,

samveen:
@immarvin Why is `HTTP::Async` installed in `/usr/local/` heirarachy? If it is a dependency ,then it should be correctly packaged as with other xCAT dependencies!!

PR #2731, opened 2017-03-23
fix 2666 Test cases related rpower need to modify

### LINKED ISSUE #2666: [DEV] Test cases related rpower need to modify
Test cases about ``rpower``, such like ``rpower_off, rpower_stat, rpower_boot, rpower_on, rpower_reset, rpower_noderange, rpower_noderange_nodeps``, are checking some key work of the output of ``rpower`` command. But in different platform, the same intention are expressed by different word. For example, to express the node is working,

       on ppc64le
       # rpower c910f03c17k24 stat
       c910f03c17k24: on

       on ppc64
       # rpower c910f02c04p04 stat
       c910f02c04p04: Running

But in our test case  ``rpower_stat``, we hard code checking  key word ``Running``.
 
      start:rpower_stat
      cmd:rpower $$CN stat
      check:rc==0
      check:output=~Running

So these cases will failed in some platform.

Form xCAT point of view,  it is unnecessary to unify the output. So we have to modify our such test cases to make it more flexible.

bybai:
fix #2666 
Finished UT in both x86_64 vm and pp64le system.

### DIFF
--- a/xCAT-test/autotest/testcase/rpower/cases0
+++ b/xCAT-test/autotest/testcase/rpower/cases0
@@ -1,75 +1,79 @@
 start:rpower_off
-cmd:rpower $$CN stat
-check:ouptut=~Running
+cmd:rpower $$CN on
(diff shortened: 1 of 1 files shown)


hu-weihua:
Thank you very much @bybai

PR #2733, opened 2017-03-23
modify bundle files for makehosts testcase modified

junxiawang:
@hu-weihua modify bundle files for makehosts testcase modified according to issue:#2674


hu-weihua:
We have 13 valid bundle files, but @junxiawang you just modified 12 of them . Could you find out which one was missed ? thanks

hu-weihua:
I have reviewed the change and agree to merge. thanks @junxiawang .

PR #2740, opened 2017-03-23
Add DSH_VERIFY environment variable to xdsh command

gurevichmark:
Add DSH_VERIFY environment variable to allow same control as `-verify` command line flag of `xdsh` command.
Add DSH_VERIFY environment variable description to `man xdsh`


whowutwut:
approved and will merge

PR #2750, opened 2017-03-27
enhance rpower testcase

bybai:
enhance test case for rpower #2666


hu-weihua:
It is good, thanks @bybai .

Issue #2684, opened 2017-03-20
[Test Case] addkcmdline support for nodeset


chenglch:
Check whether the addkcmline value from bootparams table is applied to configuration file for various platforms like xnba, grub etc.

#2614
#2620

neo954:
@chenglch, please review the pull request #2754. Thanks.

PR #2754, opened 2017-03-27
Add test case for the addkcomdline attribute of node definition

neo954:
For issue #2684.

### DIFF
--- a/xCAT-test/autotest/testcase/nodeset/cases1
+++ b/xCAT-test/autotest/testcase/nodeset/cases1
@@ -1,0 +1,9 @@
+start:nodeset_addkcmdline
+cmd:chdef $$CN addkcmdline=debug
+check:rc==0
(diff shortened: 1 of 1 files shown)


xCAT-test/autotest/testcase/nodeset/cases1
```diff
@@ -0,0 +1,9 @@
+start:nodeset_addkcmdline
+cmd:chdef $$CN addkcmdline=debug
+check:rc==0
+cmd:nodeset $$CN osimage=__GETNODEATTR($$CN,os)__-__GETNODEATTR($$CN,arch)__-install-compute
+check:rc==0
+cmd:find -L /tftpboot -name '*$$CN*' | grep debug
```
chenglch: Does `cat` or some command else should be used to print the content of the file before using `grep`?

neo954: That is my fault. Will fix this.

chenglch:
Looks good to me.

Issue #2582, opened 2017-03-02
Nodes bmc definition duplicates its hostname


CJCShadowsan:
Hi,

On an xCAT 2.13.2 system:

```
[root@mgmt01 nbfs]# lsdef mgmt02
Object name: mgmt02
    arch=x86_64
    bmc=**mgmt02mgmt02-imm**
```

The ipmi table is defined as such:

```
[root@mgmt01 nbfs]# tabdump ipmi
#node,bmc,bmcport,taggedvlan,bmcid,username,password,comments,disable
"ipmi","|\z|-imm|",,,,,,,
```

For some reason the hostname is looping around twice - meaning no BMCs can be defined correctly with this regex.

Note that this configuration works in 2.13.1.

zet809:
Can you show me the whole node definition? lsdef -t node mgmt02
And how the group "ipmi" is defined? lsdef -t group ipmi

CJCShadowsan:
Sure:

```
[root@mgmt01 nbfs]# lsdef -t group ipmi
Object name: ipmi
    bmc=|\z|-imm|
    chain=runimage=http://10.80.0.3/install/nbfs/raid_asu.tgz,runcmd=bmcsetup,standby
    cons=ipmi
[...]

Issue #2685, opened 2017-03-20
[Test Case] Nodes bmc definition duplicates its hostname


chenglch:
check the all kinds of regular expression that xcat support to define the node attributes based on node name.
#2582
#2587

neo954:
Test case added in pull request #2756. @chenglch, please review the pull request.

PR #2756, opened 2017-03-27
Add test case to verify github issue #2582

neo954:
For issue #2582


neo954:
For issue #2685

whowutwut:
@neo954  @hu-weihua Can we standardize pull request must contain some sort of validation or description.... Creating pull request without any text describing the changes is not a good habit. 
 
I also want to make sure the test case does infact work... Ideally we would execute this single testcase on a system that has xCAT 2.13.0 GA, and it should FAIL, then we run it on a system that has 2.13.3 installed and it should PASS.  These validation criteria should be part of the description for the pull request....    Just copy and paste it in...

PR #2761, opened 2017-03-28
fix issue [DEV] "remoteshell" postscript complains "chown: invalid group: root:ssh_keys" and "Permissions 0640 for /etc/ssh/ssh_host_dsa_key are too open" on sles11.4

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2757:

set the umask to be "0077", so that all the newly created is created with 600


neo954:

PR #2764, opened 2017-03-28
Fix if macdata does not exist

jjohnson42:
If macdata undefined, it took out all of inventory.


zet809:

Issue #2745, opened 2017-03-24
rinstall is not able to set or detect the arch , does not happen with nodeset


whowutwut:
When doing discovery of a ppc64le node, the genesis kernel is ppc64 , so the arch is set to `ppc64`. 

Doing `rinstall` after discovery flags the following error message: 
```
# rinstall sn02 osimage=rhels7.3-ppc64le-install-service
Error: sn02: The value of 'arch' attribute is not same as the 'osarch' attribute for osimage.
Error: No available nodes for provision.
```

However, the older step by step method is OK and  the arch is changed to `ppc64le` 

```
# nodeset sn02 osimage=rhels7.3-ppc64le-install-service
sn02: install rhels7.3-ppc64le-service
# rsetboot sn02 net
sn02: Network
# rpower sn02 boot
sn02: reset
```

immarvin:
hi @whowutwut , what is the node definition after discovery? this message is throw out by the pre-check of ``rinstall``  to make sure the arch of node match the arch of osimage.

whowutwut:
@immarvin  The arch of the node is `ppc64` after discovery

Correct, I suspected that `rinstall` does some pre checks.  But this means we cannot use `rinstall` after we do discovery of PPC64LE machines.   Here's an example.... 

If arch is set right, rinstall works.... 
```
[root@fs1 ~]# lsdef frame45cn05 -i arch
Object name: frame45cn05
    arch=ppc64le
[root@fs1 ~]# rinstall frame45cn05 osimage=rhels7.4-pegas-ppc64le-install-compute
Provision node(s): frame45cn05
```

Force the arch to `ppc64`....  and we hit the problem 
```
[root@fs1 ~]# chdef frame45cn05 arch=ppc64
1 object definitions have been created or modified.
[root@fs1 ~]# rinstall frame45cn05 osimage=rhels7.4-pegas-ppc64le-install-compute
Error: frame45cn05: The value of 'arch' attribute is not same as the 'osarch' attribute for osimage.
Error: No available nodes for provision.
```

But nodeset will change it regardless..... 
```
[root@fs1 ~]# nodeset frame45cn05 osimage=rhels7.4-pegas-ppc64le-install-compute
frame45cn05: install rhels7.4-pegas-ppc64le-compute
[root@fs1 ~]# lsdef frame45cn05 -i arch
Object name: frame45cn05
    arch=ppc64le
```

Even copying a x86_64 osimage, it changes it to x86_64 to match...  
```
[...]

PR #2766, opened 2017-03-29
fix issue rinstall is not able to set or detect the arch , does not happen with nodeset #2745

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2745:

loosen the ``rinstall`` code logic on arch validation for "ppc64" and ppc64le""


zet809:

bybai:
I agreed to merge this. Thanks

PR #2768, opened 2017-03-29
modify makehosts testcase according to issue:add test case to makehos…

junxiawang:
…ts erroneously deletes entries from /etc/hosts #2674
@bybai  please help review


bybai:
I agreed to merge it.

Issue #1174, opened 2016-05-20
Enhancements for getadapter command to support VM (ppc64le)


gurevichmark:
Following enhancements are proposed:
1. Improve error message when `getadapter` is issued against a VM, currently the message is `Error: Failed to run command: rnetboot <vmname>`. Perhaps it should say `Error: getadapter is not supported for KVM VMs`. Or maybe `getadapter` should be enhanced to work for VMs.
2. Since `getadapter` boots genesis, it could take several minutes for the command to complete. Currently there is no feedback to the user as to what is going on. Perhaps when the command starts, if genesis boot is needed, it could display a message saying something like `Booting node into genesis, this could take several minutes...`
3. Improve the man page for `getadapter` command to make it very clear that the node might be booted into genesis. Also explain what `local` is. Today the man page just says "If the target node is scaned for the first time, getadapter will trigger genesis to collect information then save the information at local.", but it is not explained what `local` is.

whowutwut:
@gurevichmark  can you help verify and close if it looks OK

gurevichmark:
Nope. Still not working for a VM.

It seems that PR #1303 just fixed the error message displayed, but I still can not get `getadapter` command to return anything:

1. For VM, the `getadapter` reboots the VM, but does **not** go into genesis and the command just hangs.
2. For a real server, the `getadapter` reboots the server, but it just enters this state:
```
<166>Feb 14 16:47:34 xcat.genesis.doxcat: Getting initial certificate --> 192.168.3.29:3001
<166>Feb 14 16:47:35 xcat.genesis.doxcat: The destiny=runcmd, destiny parameters=getadapter
<166>Feb 14 16:47:35 xcat.genesis.doxcat: Running nextdestiny 192.168.3.29:3001...

Retrying in 39 seconds
```
And keeps retrying forever, the `getadapter` command never returns.

gurevichmark:
Assigning to @zet809 so he can choose the right person to work on this issue.

zet809:
@chenglch Hi, Cheng long, pls talk a look at this issue, thx!

robin2008:
Did we verify this issue with 2.13.5?

zet809:
Will be covered by issue #3382, so close this one.

PR #2770, opened 2017-03-29
Use rinstall to replace nodeset to boot into genesis for getadapter

chenglch:
close-issue: #1174


hu-weihua:
I have reviewed the change and agree to merge, thanks @chenglch .

Issue #2769, opened 2017-03-29
Add test case for Quotes in tables' comment field break mypostscript 


bybai:
Add test case for Quotes in tables' comment field break mypostscript #2634

bybai:
add in pull request #2773

bybai:
Verified in autotest.

PR #2773, opened 2017-03-29
fix 2769 Add test case for Quotes in tables' comment field break mypo…

bybai:
add case #2769


hu-weihua:
I have reviewed the change and agree to merge, thanks @bybai .

PR #2784, opened 2017-03-30
fix a typo

immarvin:


hu-weihua:

PR #2790, opened 2017-03-31
modify mtu value and clear mtu

bybai:
modify confignics_s testcase mtu 1496,  and backup original networks table.


whowutwut:
@bybai  What is the purpose of this testcase? We are just making sure that the value of the MTU is set correctly?

bybai:
Hi @whowutwut , this case is to 
1, change MTU 1501 to 1496. 
2, clear up MTU in networks table after the test case is finished.

PR #2792, opened 2017-03-31
[go-xcat] Ignore any error of `yum clean metadata`

neo954:


hu-weihua:
@neo954 , could you offer some description for this pull request? thanks

neo954:
This patch will let the script ignore any error from `yum clean metadata`. Since this error is not important. If the script still reach error when do `yum update` later. It will do the error handling there.

hu-weihua:
I have reviewed the change and agree to merge, thanks @neo954.

neo954:
Thanks @hu-weihua

Issue #2558, opened 2017-02-24
Support for encrypted passwords in the passwd table is broken


kcgthb:
According to [the documentation]( https://xcat-docs.readthedocs.io/en/stable/guides/admin-guides/manage_clusters/ppc64le/configure/password.html#configure-passwords), one can store encrypted password in the `passwd` table. The provided example (using `openssl passwd -1` generates password hashes using `md5`.

Now, with #2168, the default crypt method for passwords in the `passwd` table seem to be `sha256`, which is a first problem. The documentation should probably be updated to mention that, and give an example to generate a `sha256` encrypted password hash.

Now, looking at #2168 in more details, especially https://github.com/xcat2/xcat-core/pull/2168/files?diff=unified#diff-97ef2682510dad0de95d0570c8f81106, and unless I missed something, I think there's been a major regression in functionality. 

Before, the value stored in the `passwd` table was checked to see if it started with `$1`, `$5$` or `$6$` (the identifiers for `md5`, `sha256` and `sha512` respectively). If it started with one of those, it was recognized as being already hashed, and was returned as-is by the `crydb()` function. If it did not, it was then hashed with `crypt()` function.

Now, it looks to me that the value from the `passwd` table is crypted no matter what, even if it already is. So as a result, if I store a hashed value in `passwd`, it gets hashed again, and the resulting hash in `/etc/passwd` doesn't correspond to anything anymore, meaning that I can't use the password I chose to authenticate as root.

I believe this is also the case for the stateless and statelite cases (`xCAT-server/lib/xcat/plugins/packimage.pm` and `xCAT-server/lib/xcat/plugins/statelite.pm`).

Could you please restore the previous behavior? We need to be able to store encrypted root passwords in the `passwd` table.

kcgthb:
Also the helpfile for the `passwd` table says:
```
       cryptmethod
                 Indicates the method that was used to encrypt the password attribute.
                 On AIX systems, if a value is provided for this attribute it indicates
                 that the password attribute is encrypted.  If the cryptmethod value is
                 not set it indicates the password is a simple string value. On Linux
                 systems, the cryptmethod can be set to md5, sha256 or sha512. If not
                 set, sha256 will be used as default.
```
Maybe it should say instead that the "cryptmethod" field indicates the method that xCAT will use to generate the encrypted password, and that the password should be stored in plain-text.
[...]

PR #2855, opened 2017-04-12
add test cases for #2558: Support for encrypted passwords in the passwd table is broken

tingtli:
@chenglch  This is to add test cases for  bug #2558. Please help review. Thanks!
We have 8 test cases. These cases is to test when we use encryped method in password table, we have correct encrypted password to do provision. 
encrypted_passwd_md5_diskfull
ecrypted_passwd_md5_diskless
encrypted_passwd_sha256_diskfull
encrypted_passwd_sha256_diskless
encrypted_passwd_sha512_diskfull
encrypted_passwd_sha512_diskless
encrypted_passwd_openssl_diskfull
encrypted_passwd_openssl_diskless

We will put encrypted_passwd_sha512_diskfull, encrypted_passwd_sha512_diskless and encrypted_passwd_openssl_diskfull in our daily run bucket.

### DIFF
--- a/xCAT-test/autotest/testcase/passwd/case0
+++ b/xCAT-test/autotest/testcase/passwd/case0
@@ -1,0 +1,304 @@
+start:encrypted_passwd_md5_diskfull
+description: this case is to test when  cryptmethod=md5 in passwd table, the password is encryped with the md5 method. When compute node is diskfull provisioned, users can login node with correct password. This case should be run when node is correctly defined.
+cmd:tabdump passwd |grep -i system| awk -F, '{print $4}' > /tmp/tmpcryptmethod
+check:rc==0
+cmd:chtab key=system passwd.cryptmethod=md5
+check:rc==0
+cmd:tabdump passwd |grep -i md5
+check:rc==0
+cmd:nodeset $$CN osimage=__GETNODEATTR($$CN,os)__-__GETNODEATTR($$CN,arch)__-install-compute
+check:rc==0
+cmd:grep '\$1\$' /install/autoinst/$$CN
+check:rc==0 
+cmd:oldcryptmethod=`cat /tmp/tmpcryptmethod |sed 's/\"//g'`;chtab key=system passwd.cryptmethod=$oldcryptmethod
+check:rc==0
+cmd:nodeset $$CN osimage=__GETNODEATTR($$CN,os)__-__GETNODEATTR($$CN,arch)__-install-compute
+check:rc==0
+cmd:rm -rf /tmp/tmpcryptmethod
+end
+
+start:encrypted_passwd_md5_diskless
+description: this case is to test when  cryptmethod=md5 in passwd table, the password is encryped with the md5 method. When compute node is diskless provisioned, users can login node with correct password. This case should be run when node is correctly defined.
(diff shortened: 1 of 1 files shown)


xCAT-test/autotest/testcase/passwd/case0
```diff
@@ -0,0 +1,304 @@
+start:encrypted_passwd_md5_diskfull
+description: this case is to test when  cryptmethod=md5 in passwd table, the password is encryped with the md5 method. When compute node is diskfull provisioned, users can login node with correct password. This case should be run when node is correctly defined.
+cmd:tabdump passwd |grep -i system| awk -F, '{print $4}' > /tmp/tmpcryptmethod
```
chenglch: I have a little concern about the `awk {print $4}` although the column is right and we may not change the `passwd`  table schema. But if we change it, this case may fail as $4 column may not be the `cryptmethod`.
Does `gettab` work here?

xCAT-test/autotest/testcase/passwd/case0
```diff
@@ -0,0 +1,304 @@
...
+check:rc==0
+cmd:chtab key=system passwd.cryptmethod=md5
+check:rc==0
+cmd:tabdump passwd |grep -i md5
```
chenglch: We may not need to check the the output of `passwd`.  From my point view, this is redundant case, it has already been checked in the `tabdump` cases of other `bundle`.

xCAT-test/autotest/testcase/passwd/case0
```diff
@@ -0,0 +1,304 @@
...
+check:rc==0
+#config CN to do diskless provision
+cmd:chdef -t node -o $$CN servicenode= monserver=$$MN nfsserver=$$MN tftpserver=$$MN  xcatmaster=$$MN
+check:rc==0
+cmd:makedns -n
+check:rc==0
+cmd:makeconservercf $$CN
```
chenglch: I do not have enough experience about diskless. But if our purpose to to check the cryptmethod, do we need to enable console here?  If we think crypt password is a part of node provision, we should add the password check at the test case for provision. Add the case like this will consume too much time.

daniceexi:
@tingtli @hu-weihua I think we need to add more comments for reviewer and later user to understand the test cases. This is a comment for all the test cases.

xCAT-test/autotest/testcase/passwd/case0
```diff
@@ -0,0 +1,298 @@
...
+check:rc==0
+cmd:chtab key=system passwd.username=root passwd.password=`openssl passwd -1 abc123`
+check:rc==0
+cmd:gettab key=system passwd.password > /tmp/tmpcryptedpasswd 
+check:rc==0
+cmd:nodeset $$CN osimage=__GETNODEATTR($$CN,os)__-__GETNODEATTR($$CN,arch)__-install-compute
+check:rc==0
+cmd:grep '\$1\$' /install/autoinst/$$CN |awk -F " " '{print $3}' > /tmp/instcryptedpasswd
+check:rc==0
+cmd:diff /tmp/tmpcryptedpasswd /tmp/instcryptedpasswd
```
chenglch: I have little concern about this way to compare the string, but it works.

PR #2860, opened 2017-04-13
Add timestamp for every cmd running in xcatteat log to convenient debug in the future

hu-weihua:
This pull request modify xcattest to add timestamp for every cmd running in xcatteat log to convenient debug in the future

### DIFF
--- a/xCAT-test/xcattest
+++ b/xCAT-test/xcattest
@@ -907,7 +907,8 @@ sub runcase
 
             #by
             my $runstart = timelocal(localtime());
-            log_this("\nRUN:$cmd");
+            my $runstartstr = scalar(localtime());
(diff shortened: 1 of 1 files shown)


xCAT-test/xcattest
```diff
@@ -907,7 +907,8 @@ sub runcase
 
             #by
             my $runstart = timelocal(localtime());
-            log_this("\nRUN:$cmd");
+            my $runstartstr = scalar(localtime());
```
neo954: It seems it will get a different time from line 909. It may be a problem.

hu-weihua: This is just a reference time, has error in millisecond.

neo954:
Will merge this one.

PR #2869, opened 2017-04-14
xCAT-server: fix S7200AP ipmi detection

khm:
The Intel Xeon PHI S7200AP board uses a Manufacturer ID of 343, so xCAT attempts to use intelplus for the IPMI console.  However, this board does not use intelplus, so we should exclude it from the check (by also testing against the Product ID).


zet809:
Looks good to me, Thx!

Issue #2835, opened 2017-04-07
[Test case] refine sles migration test cases


zet809:

neo954:
This test case calls `go-xcat`. The problem described in the issue is fixed in commit d13f6f0f5be2e3d950a69cf6f27d14f48c110c3a. See pull request #2876.

neo954:
Test cases updated with pull request #2878

Issue #2409, opened 2017-01-13
Migration on sles12.* failes. xCAT-genesis-scripts-ppc64 gives vendor change issue


tingtli:
This issue happens on sles12.1 and sles12.2
Here is the steps
1. Install xCAT 2.13 1/5 build
```
c910f03c17k22:/etc/zypp/repos.d # rpm -qa |grep -i xcat
conserver-xcat-8.1.16-10.ppc64le
xCAT-probe-2.13.1-snap201701050616.noarch
grub2-xcat-2.02-0.16.el7.snap201506090204.noarch
xCAT-genesis-scripts-x86_64-2.13.1-snap201701050616.noarch
xCAT-buildkit-2.13.1-snap201701050616.noarch
xCAT-server-2.13.1-snap201701050615.noarch
xCAT-genesis-base-x86_64-2.13.1-snap201701110114.noarch
xCAT-2.13.1-snap201701050616.ppc64le
elilo-xcat-3.14-4.noarch
xCAT-genesis-scripts-ppc64-2.13.1-snap201701050616.noarch
syslinux-xcat-3.86-2.noarch
perl-xCAT-2.13.1-snap201701050615.noarch
ipmitool-xcat-1.8.17-1.ppc64le
xCAT-genesis-base-ppc64-2.13.1-snap201701110052.noarch
xCAT-client-2.13.1-snap201701050615.noarch
yaboot-xcat-1.3.17-rc1.noarch
xCAT-test-2.13.1-snap201701050616.noarch
```
2. download xcat 2.13.1 1/11 build and create the new repo
```
c910f03c17k22:/etc/zypp/repos.d # zypper search xCAT-genesis-scripts-ppc64
Loading repository data...
Reading installed packages...

S | Name                       | Summary                                   | Type
--+----------------------------+-------------------------------------------+--------
i | xCAT-genesis-scripts-ppc64 | xCAT Genesis netboot image - Core content | package

c910f03c17k22:/etc/zypp/repos.d # ls /xcat-core/xCAT-genesis-scripts-ppc64-2.13.1-snap201701110845.noarch.rpm
/xcat-core/xCAT-genesis-scripts-ppc64-2.13.1-snap201701110845.noarch.rpm

```
3. Update xCAT-genesis-scripts-ppc64 and get the result.
```
c910f03c17k22:/etc/zypp/repos.d # zypper update xCAT-genesis-scripts-ppc64
Loading repository data...
Reading installed packages...
[...]

PR #2876, opened 2017-04-17
[go-xcat] Fix repository query and force problem resolution for zypper

neo954:
This pull request fix multiple problems on SLES.

### DIFF
--- a/xCAT-server/share/xcat/tools/go-xcat
+++ b/xCAT-server/share/xcat/tools/go-xcat
@@ -2,7 +2,7 @@
 #
 # go-xcat - Install xCAT automatically.
 #
-# Version 1.0.17
+# Version 1.0.18
 #
 # Copyright (C) 2016 International Business Machines
 # Eclipse Public License, Version 1.0 (EPL-1.0)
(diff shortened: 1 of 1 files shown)


neo954:
This pull request solve issue #2835.

neo954:
@robin2008 and @immarvin Please help review this pull request. The second patch of this pull request solve the migration problem on SLES. Which is due to xCAT changed the vendor name from `IBM Corp` to `IBM Corp.` in the metadata of some of the xCAT rpm packages.

neo954:
See issue #2409.

robin2008:
You fix may work on go-xcat, but I don't think this PR can resolve #2835 , as in xCAT-test package, for the migration in SLES,  it does not leverage go-xcat (BTW:  I don't see any zypper update in go-xcat,  is the tool support upgrade?) .

immarvin:
seems there has been some discussion and decision in https://github.com/xcat2/xcat-core/issues/2409

neo954:
@robin2008 This patch will not resolve #2835, since in #2835, there is no `go-xcat` involved. This patch only partly covered the problem. If the end user choose use `go-xcat` to do the xCAT upgrade, he/she will not hit the migration problem any more.

BTW, `go-xcat` does not call `zypper update`, instead, `go-xcat` calls `zypper install`. That is due to `zypper update` will update all the rpm packages installed on the system. But `go-xcat` will only update the packages xCAT shipped. Thus, `go-xcat` calls `zypper install` with a specific package list, in this way, only the packages shipped by xCAT will be updated.

neo954:
@immarvin, issue #2409 will leave there as is. This patch only partly covered that issue if `go-xcat` is involved during xCAT upgrade procedure by adding the `--foce-resolution` command line argument to `zypper`.

robin2008:
@neo954 
To current sprint, enhance go-xcat is another new task which is not planned. Our focusing is to clean up the failure in daily automation testing. Please enhance the SLES migration case to cover the specific version upgrade (you can added the version parsing there and do the corresponding upgrade action based on x -> y)

robin2008:
Fine with the code change,  but I'm not very sure about the automatic resolution choice made by zypper when "--force-resolution" is applied. Assume that you have done the UT for go-xcat via xcattest.

robin2008:
Code change is merged into go-xcat, let's verify it ASAP.

PR #2878, opened 2017-04-18
[testcase] Do xCAT upgrade by add `--force-resolution` to `zypper`

neo954:
For issue #2835.


robin2008:
I'd like to introduce "--force-resolution" for specific version upgrade as a workaround, because the case should be consistent with steps in xcat guide in most of the case.

BTW:  why not use '-y' option here?  Obviously we need a non-interactive mode.

neo954:
@robin2008 For `zypper`, `-n` is stand for non-interactive mode.

       -n, --non-interactive
              Switches  to  non-interactive  mode. In this mode zypper doesn't ask user to type answers to
              various prompts, but uses default answers automatically. The behaviour  of  this  option  is
              somewhat  different  than  that  of  options like '--yes', since zypper can answer different
              answers to different questions. The answers also depend on  other  options  like  '--no-gpg-
              checks'.

immarvin:
hi @neo954 , is it possible to apply "--force-resolution" only for the migration from the problematic xCAT release? I am afraid that apply "--force-resolution" without condition might hide some real problem during upgrade

neo954:
Hello all, the test cases are updated based on your comments.

Issue #2822, opened 2017-04-07
[Test case]Add test case for 2809: enhance bmcdiscover for openbmc


zet809:
Test case for #2809

zet809:
hi, @junxiawang , will you pls reference this issue within the Pull Request than refine or add test cases for bmcdiscover? It make us confusing that the task is closed but how is the task done. Thx a lot.!

PR #2879, opened 2017-04-18
add bmcdiscovery testcases #2822

junxiawang:
Modify bmcdiscover testcase
If case passed so we could verify bmcdiscover command could work normally

### DIFF
--- a/xCAT-test/autotest/testcase/bmcdiscover/cases0
+++ b/xCAT-test/autotest/testcase/bmcdiscover/cases0
@@ -1,3 +1,33 @@
+start:bmcdiscover_q
+os:linux
+decription:query usage for bmcdiscover
+cmd:bmcdiscover ?
+check:rc==0
+check:output=~bmcdiscover - Discover BMC
+check:output=~Usage
+check:output=~bmcdiscover \[-?|-h|--help\]
+check:output=~bmcdiscover \[-v|--version\]
+end
+
+start:bmcdiscover_help
+os:linux
+decription:show bmcdiscover help information 
+cmd:bmcdiscover help
+check:rc==0
+check:output=~bmcdiscover \- Discover BMC \(Baseboard Management Controller\) using the specified scan method
+check:output=~Usage:
+check:output=~bmcdiscover \[-?|-h|--help\]
+check:output=~bmcdiscover \[-v|--version\]
+end
+
+start:bmcdiscover_version
+os:linux
+description:show bmcdiscover version
+cmd:bmcdiscover --version
+check:rc==0
+check:output=~bmcdiscover - xCAT Version
+end
+
 start:bmcdiscover_h
 os:Linux
 decription:help
@@ -42,6 +72,12 @@ check:rc==0
 check:output=~bmc=$$bmcrange
 end
 
+start:bmcdiscover_range_u_p_i_ipsource
+description:check the bmcrange is which type address
+cmd:bmcdiscover -i  $$bmcrange -u $$bmcusername -p $$bmcpasswd --ipsource
+check:rc==0
+check:output=~IP Address Source       : Static Address
+end
 start:bmcdiscover_range_z
 cmd:bmcdiscover --range  $$bmcrange -u $$bmcusername -p $$bmcpasswd -z
 check:rc==0


xCAT-test/autotest/testcase/bmcdiscover/cases0
```diff
@@ -42,6 +72,12 @@ check:rc==0
 check:output=~bmc=$$bmcrange
 end
 
+start:bmcdiscover_range_u_p_i_ipsource
+description:check the bmcrange is which type address
+cmd:bmcdiscover -i  $$bmcrange -u $$bmcusername -p $$bmcpasswd --ipsource
+check:rc==0
+check:output=~IP Address Source       : Static Address
```
bybai: I think ip source is not just Static Address, it may from dhcp.

bybai:
Although the ip source may static address or from dhcp, we do not have dhcp configured system now, we do not know the dhcp type of output, so we only check `Static Address` from output currently. Agree to merge now.

bybai:
Although there are static address or address from dhcp, we do not have system address from dhcp, we do not know the output of dhcp address for this command, so we only check static address currently.

xCAT-test/autotest/testcase/bmcdiscover/cases0
```diff
@@ -42,6 +72,12 @@ check:rc==0
 check:output=~bmc=$$bmcrange
 end
 
+start:bmcdiscover_range_u_p_i_ipsource
+description:check the bmcrange is which type address
+cmd:bmcdiscover -i  $$bmcrange -u $$bmcusername -p $$bmcpasswd --ipsource
```
whowutwut: Where are the variables defined?

xCAT-test/autotest/testcase/bmcdiscover/cases0
```diff
@@ -1,3 +1,33 @@
...
+check:output=~bmcdiscover \- Discover BMC \(Baseboard Management Controller\) using the specified scan method
+check:output=~Usage:
+check:output=~bmcdiscover \[-?|-h|--help\]
+check:output=~bmcdiscover \[-v|--version\]
+end
+
+start:bmcdiscover_version
+os:linux
+description:show bmcdiscover version
+cmd:bmcdiscover --version
```
whowutwut: You check short help (?) and long help (--help) why only check long version? What about short version? 

Regardless how long do these test cases take to run and how does the function break? Seems like this should be added to long running full regression bucket but skipped in a short bucket because the likelihood of it breaking may be quite low

xCAT-test/autotest/testcase/bmcdiscover/cases0
whowutwut: Can we check for either ?

whowutwut:
@junxiawang please set as much information in thr right side bar for your pull request. Thanks

whowutwut:
@bybai please do not merge unless the info is complete on the right side. Thanks!

xCAT-test/autotest/testcase/bmcdiscover/cases0
hu-weihua: @whowutwut, these variables are defined in the  ``[system]`` section in ``xcattest`` config file. One test case, just like a script coded by developer, in order to be used in different environment, some parameter need to be passed from command line.  So for xcat test case, due to all cases are parsed and triggered by ``xcattest`` script, so all the parameters needed by cases should be assigned value in ``xcattest`` 's config file ahead.  

That is why you asked me if we can offer a default ``xcattest`` config file to let developer to use for their UT, I said it is hard, it do depending on which cases invoked by developer. Developer invoked different cases, the corresponding config file is different due to these cases have different parameter. 

I am trying to find a way to make it is easy for developer to know how many and which parameters need to assign value before they invoke a test case or a bundle of test case.

xCAT-test/autotest/testcase/bmcdiscover/cases0
junxiawang: short version has been checked

xCAT-test/autotest/testcase/bmcdiscover/cases0
junxiawang: short version has been checked

xCAT-test/autotest/testcase/bmcdiscover/cases0
junxiawang: no  bmc ip is dhcp now and will modify the case again when we have dhcp bmc ip.

Issue #2808, opened 2017-04-07
[openbmc] Support rsetboot for openbmc server


zet809:

xuweibj:
The task was finished in pull request #2848 .

whowutwut:
@zet809 @xuweibj @hu-weihua We can start testing and validating this....

The firmware applied on frame6u05 supports this... 
```
[root@stratton01 frame6u05]# rsetboot p9euh01 net
p9euh01: Error: 403 Forbidden - This function is not yet available in OpenBMC firmware.
}[root@stratton01 frame6u05]# rsetboot frame6u05 net
frame6u05: Network
```

PR #2880, opened 2017-04-18
add rsetboot testcase for issue 2808

junxiawang:
for issue #2808 add rsetboot test cases


xuweibj:
Look good to me. But I think we need more cases, I will open another issue to trace.
@junxiawang Thank you.

PR #2881, opened 2017-04-18
add tips for interface modification

zet809:
Add tips for modifying external or internal interfaces.

It looks like this:

```
For interface modification

For the modification of interface either external or internal, the developer must pay much attention to it.

    1. The owner shall get an agreement from all developers before modifying.
    2. The modification must pass a set of testcases related to the modified interface.
```

### DIFF
--- a/docs/source/developers/guides/code/tips.rst
+++ b/docs/source/developers/guides/code/tips.rst
@@ -1,2 +1,11 @@
 Tips
 ====
+
+For interface modification
+--------------------------
+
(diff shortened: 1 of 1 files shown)


whowutwut:
@zet809 What is this in reference to? An external behavior interface modification?  Can you give an example of a recent event that happened to spawn this doc change?

whowutwut:
Closing, please re-open if we want to add it.

Issue #2795, opened 2017-03-31
performing yum update xCAT* overwrote our log rotate settings.


ralphbellofatto:
We have the following log rotate settings for xcat (which we modified after install).

```
/var/log/xcat/*.log {
    missingok
    sharedscripts
    compress
    postrotate
        test -f /var/run/rsyslogd.pid && kill -HUP `cat /var/run/rsyslogd.pid 2> /dev/null` 2> /dev/null || true
        test -f /var/run/syslogd.pid && kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true
        test -f /var/run/xcat/cmdlogservice.pid && kill -HUP `cat /var/run/xcat/cmdlogservice.pid 2> /dev/null` 2> /dev/null || true
    endscript
}
```

After doing a yum update xCAT* we were left with:
```
/var/log/xcat/*.log {
    missingok
    sharedscripts
    postrotate
        test -f /var/run/rsyslogd.pid && kill -HUP `cat /var/run/rsyslogd.pid 2> /dev/null` 2> /dev/null || true
        test -f /var/run/syslogd.pid && kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true
        test -f /var/run/xcat/cmdlogservice.pid && kill -HUP `cat /var/run/xcat/cmdlogservice.pid 2> /dev/null` 2> /dev/null || true
    endscript
}
```

```
FILE COMPARISON
Produced: 3/31/2017 3:28:53 PM

Mode:  Just Differences

Left file: c:\wtools\r92gitlab\research-92-cluster\r92config\system\etc\logrotate.d\xcat
Right file: c:\wtools\r92gitlab\research-92-cluster\r92config\r92rsync\system\etc\logrotate.d\xcat
    compress +-
----------------
```

Please don't throw away our settings after performing an update...

whowutwut:
@robin2008  can you take a look at this?

robin2008:
xCAt don't want customer modify the OOB configuration for this files as we don't know if those changes w/o the support of xCAT will impact xCAT functionalities. (In this case, it seems no impact as only compress the historic logs.)

In that, xCAT does not consider this file as a config file (%config) in spec, but a normal package file(%file). And no additional steps in spec scripts to do the backup and restore.  Then, this file will be always be replaced with the file in package of the new version.
[...]

PR #2884, opened 2017-04-19
fix issue performing yum update xCAT* overwrote our log rotate settings. #2795

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2795:

take /etc/logrotate.d/xcat as a config file. it follows the rules of processing config files during rpm upgrade in http://rpm5.org/docs/max-rpm.html#s2-rpm-upgrade-config-file-magic:

with this fix, the scenario in issue https://github.com/xcat2/xcat-core/issues/2795 will follow the 
````
**Original file = X, Current file = Y, New file = X**

Here we have a file that was changed at some point. However, the new file is identical to the existing file prior to the local modifications.

In this case, RPM takes the viewpoint that since the original file and the new file are identical, the modifications made to the original version must still be valid for the new version. It leaves the existing, modified file in place. 

````

if the /etc/logrotate.d/xcat shipped by xCAT changed in the future, it follows the  rule:

````
 **Original file = X, Current file = Y, New file = Z**

Here the original file was modified at some point. The new file is different from both the original and the modified versions of the original file.

RPM is not able to analyze the contents of the files, and determine what is going on. In this instance, it takes the best possible approach. The new file is known to work properly with the rest of the software in the new package — at least the people building the new package should have insured that it does. The modified original file is an unknown: it might work with the new package, it might not. So RPM installs the new file.

BUT… The existing file was definitely modified. Someone made an effort to change the file, for some reason. Perhaps the information contained in the file is still of use. Therefore, RPM saves the modified file, naming it <file>.rpmsave, and prints a warning, so the user knows what happened:


warning: /etc/skel/.bashrc saved as /etc/skel/.bashrc.rpmsave

            
````

### DIFF
--- a/xCAT/xCAT.spec
+++ b/xCAT/xCAT.spec
@@ -251,7 +251,7 @@ exit 0
 /install/postscripts
 /install/prescripts
 %ifos linux
-/etc/logrotate.d/xcat
+%config /etc/logrotate.d/xcat
 /etc/rsyslog.d/xcat-cluster.conf
 /etc/rsyslog.d/xcat-compute.conf
 /etc/rsyslog.d/xcat-debug.conf


robin2008:
Looks fine to me for RPM based OS,  and we can treat DEB file in other issue.

PR #2887, opened 2017-04-19
modify rpower testcase :add rpower node status

junxiawang:
add rpower node status test case for openbmc needed 
@xuweibj please help review


xuweibj:
Looks good to me.
@junxiawang  Thank you.

PR #2888, opened 2017-04-19
standardize rpower testcase name

junxiawang:
standardize rpower test cases
@hu-weihua  please help review


hu-weihua:
I have reviewed the change and agree to merge, thanks @junxiawang

PR #2890, opened 2017-04-19
Display error information when rflash fails.

gurevichmark:
This pull requests fixes 2 small `rflash` problems:

* If `--retry` flag with value of `0` is specified for `rflash` command: `rflash --retry=0`, it is treated as a boolean `false` in the `if` statement. It should always be treated as a numeric value, therefore `defined` should be used in the `if` statement.

* Since all of the output (standard error and out) from the `hpm update` command gets logged into a log file, nothing gets returned from the command execution, even if there was an error. Code added to `grep` that log file for "Error" string and display whatever is returned. For example:

```
c460c001: 14: 20:56 rflashing ... See the detail progress :"tail -f /var/log/xcat/rflash/c460c001.log"
Error: c460c001: Running ipmitool command /opt/xcat/bin/ipmitool-xcat -H 10.2.1.1 -I lanplus -U ADMIN -P admin -z 10000 hpm upgrade /root/hpm/garrison/8335GTB_820.1642.20170327n_update.hpm force  >/var/log/xcat/rflash/c460c001.log 2>&1 failed:
 |*  2|BIOS         |   1.12 46000000 | ---.-- -------- |   1.12 70000000 |Error uploading firmware block.
 Error in Upload FIRMWARE command [rc=-1]
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -1863,7 +1863,7 @@ sub do_firmware_update {
         }
         if ($opt =~ /retry=/) {
             my ($attribute, $retry_value) = split(/=/, $opt);
-            if ($retry_value) {
+            if (defined $retry_value) {
                 # retry option was passed in, reset the default
                 $retry = $retry_value;
             }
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1941,14 +1941,19 @@ RETRY_UPGRADE:
     # report an error, exit to the caller and leave node in powered off state
     # otherwise report an error, power on the node  and try upgrade again
     if ($::RUNCMD_RC != 0) {
+        # Since "hpm update" command in step 4 above redirects standard out and error to a log file,
+        # nothing gets returned from execution of the command. Here if RC is not zero, we
+        # extract all lines containing "Error" from that log file and display them in the sendmsg below
+        my $get_error_cmd = "/usr/bin/cat ".$rflash_log_file." | grep Error";
```
whowutwut: Any advantage to doing `grep Error $rflash_log_file` instead of cat, then grep?

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1941,14 +1941,19 @@ RETRY_UPGRADE:
...
+        $output = xCAT::Utils->runcmd($get_error_cmd, 0);
         if ($retry == 0) {
             # No more retries left, report and error and exit
             $exit_with_error_func->($sessdata->{node}, $callback,
-                "Running ipmitool command $cmd failed: $output");
+                "Running ipmitool command $cmd failed:\n $output");
```
whowutwut: Suggest changing to something like... 
```
Running ipmitool command $cmd failed with rc=$::RUNCMD_RC and the following Error messages:
$output
See the $rflash_log_file for details
```

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1941,14 +1941,19 @@ RETRY_UPGRADE:
...
         }
         else {
             # Error upgrading, set a flag to attempt a retry
-            xCAT::SvrUtils::sendmsg("Running ipmitool command $cmd failed: $output", $callback, $sessdata->{node}, %allerrornodes);
+            xCAT::SvrUtils::sendmsg("Running ipmitool command $cmd failed:\n $output", $callback, $sessdata->{node}, %allerrornodes);
```
whowutwut: Same comment as above, but should we put the iteration in there? 

```
Running attempt $retry of ipmitool command $cmd failed with.... 
```

gurevichmark:
Added a pull request with following changes:
1. Suggestions from @whowutwut review:
   a. Just use `grep` instead of `cat` and `grep`
   b. Modified messages as suggested.

2. Additional small tweaks:
   a. Reduced number of messages displayed when doing `rflash` without `-v` flag by displaying some messages only with `-V`.
   b. Improved verification of `buffer-size` and `retry` input values.
   c. Increased the number of iterations to sustain `00` BMC ready code after `mc reset cold`.
   d. Decreased the number of iterations to sustain `00` BMC ready code after `hpm update`

PR #2891, opened 2017-04-20
Add test case for issue 2225 : [CUSTOMER] Stale rootimg.gz file left in the rootimg dir caused node boot failed.

tingtli:
Add 1 case in this pull request.
This case is added for issue#2225

[Case 1]
Case Name: rmimage_diskless
Description: This case isto test rmimage could work correctly to remove all image files.
Attribute: CN is the compute node and is defined in xcatconfig file

@cxhong  Please review. Thanks!


daniceexi:
@tingtli @hu-weihua Do we need to follow the new post format for our new case adding?

tingtli:
@hu-weihua  is te post format formally released?
If yes, i will modify it according to that.

xCAT-test/autotest/testcase/rmimage/case0
```diff
@@ -0,0 +1,26 @@
+start:rmimage_diskless_bug2225
```
hu-weihua: I think one issue brought up some xcat problems in one scenario,  we add test case to cover this scenario. So the case ``name`` and ``description`` should be expressed against the scenario, not the issue number. Could you change the ``name`` and ``description`` to express problem? thanks

hu-weihua:
I have reviewed and agree to merge, thanks @tingtli

Issue #2450, opened 2017-01-28
The "switches" table doesn't support groups


kcgthb:
### Issue
The `switches` table doesn't support specifying groups as the first field. It could be very useful for installation with many similar switches.

### Observed behavior
If I define the `switchtype` for a group name in the `switches` table:
```
# tabdump switches
#switch,snmpversion,username,password,privacy,auth,linkports,sshusername,sshpassword,protocol,switchtype,comments,disable
"switch_group",,,,,,,,,,"Dell",,
```
the nodes belonging to this group don't inherit the switchtype:
```
# lsdef test_switch -i groups,switchtype
Object name: test_switch
    groups=switch_group
    switchtype=
```

### Expected behavior
```
# lsdef test_switch -i groups,switchtype
Object name: test_switch
    groups=switch_group
    switchtype=Dell
```

immarvin:
hi @bybai ,would you please take a look at this?

thanks

zet809:
Hi, @bybai , will you pls verify this issue? Thx!

bybai:
Hi @zet809 , the issue can be reproduced using new code.
Hi @cxhong , could you look at this issue, there are switch table and switches table, I think you know it better.

cxhong:
yes,  I noticed this issue while I was working on PDU object.  As I mentioned before, if we change to have switch object,   it will lost all the information in the switches table while user upgrade the xcatd.  User needs to save infos in the switches table, after upgrade xcatd , then restore the switches table.   We need to very carefully to document this process to somewhere.

zet809:
Hi, @cxhong , is there something wrong when we deal with the switches table? Why the information lost when updating xCAT?

cxhong:
@zet809 ,  it looks like group only supports node object.  we can create switch object, but won't inherit attributes from group.     ```getNodeAttribs```  routine are getting attributes from nodelist which build from "node", "groups".    I don't think there is easy way to create a switchlist from "switch" and "group" based on xcat framework.

kcgthb:
@cxhong I thought switches were a regular node object, with a specific "nodetype" attribute:
```
#  lsdef -t node sh-ethsw-101-01 -i nodetype
Object name: sh-ethsw-101-01
    nodetype=switch
```

Actually, when I list all the nodes in my system with `lsdef -t node`, I get the switches in the list.

cxhong:
yes, switch is in the nodelist, but switches table use "switch" as key.
````
switches => {
[...]

PR #2893, opened 2017-04-20
Ignore groups inheritance for ppcdirect table as the special desgin

chenglch:
This patch aims to fix the error from automation testcase
`chdef_multiple_keys` due to the changes to support groups inheritance
for switches table.
As the strange design about access_tabentry, ppcdirect works like site
table and must not support groups.

Close-issue: #2450


immarvin:
agree to merge

Issue #2828, opened 2017-04-07
[Test case]Add test case for 2521: `makedhcp` fails for nodes on remote networks


zet809:
Test case for #2521

hu-weihua:
This task is being handled by pull request #2894

PR #2894, opened 2017-04-20
Add test case for issue 2521 : `makedhcp` fails for nodes on remote networks

tingtli:
Add 1 case in this pull request.
This case is added for issue#2521

[Case 1]
Case Name: makedhcp_remote_network
Description: This case is to verify bug 2521 to test when there is mgtifname='!remote!<nicname>', makedhcp could work correctly and create entrys in dhcp lease file.
Attribute:  N/A

### DIFF
--- a/xCAT-test/autotest/testcase/makedhcp/cases0
+++ b/xCAT-test/autotest/testcase/makedhcp/cases0
@@ -176,8 +176,8 @@ check:output!~$$CN
 cmd:makedhcp -a
 end
 
-start:makedhcp_remote_network_bug2521
-descriptiion:This case is to verify bug 2521 to test when there is mgtifname='!remote!<nicname>', makedhcp could work correctly and create entrys in dhcp lease file.
+start:makedhcp_remote_network
+descriptiion:This case is to test when there is mgtifname='!remote!<nicname>', makedhcp could work correctly and create entrys in dhcp lease file.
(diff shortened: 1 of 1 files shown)


daniceexi:
@tingtli @hu-weihua I think we should not add bug number in the case name and description as case is used to cover a scenario. The bug describes the scenario failed, so we should map a case to bug, but to the scenario itself.

daniceexi:
BTW, set the **[Case index]** to a number.

hu-weihua:
This pull request is handling task #2828

xCAT-test/autotest/testcase/makedhcp/cases0
```diff
@@ -176,3 +176,31 @@ check:output!~$$CN
...
+cmd:lsdef -t network
+check:rc==0
+check:output=~testnetwork
+check:rc==0
+cmd:mkdef -t node -o testnode ip=100.100.100.2 groups=all mac=42:3d:0a:05:27:0b 
+check:rc==0
+cmd: cp -f /etc/hosts /etc/hosts.bak
+cmd:echo "100.100.100.2 testnode" >> /etc/hosts
+check:rc==0
+cmd:makedhcp testnode
```
bybai: Since user often use testnode in their env, there may be old testnode entry left in dhcp lease file. I think  we can add more line  `makedhcp -d testnode` before `makedhcp testnode` to make sure there is no old testnode entry  info in dhcp file.

bybai:
Since `makedhcp -d testnode` is to clean old entry, it is no need to check rc; I am OK to merge it.

Issue #2637, opened 2017-03-13
when file is copied with xdcp using different user, owner is always "root"


sanomiya:
User _diagadmin_ has all xcat privileges. When a file is copied using this user, the ownership of the file on the target node is root instead of _diagadmin_.

```
c460mgt01:/home/diagadmin > id
uid=642490(diagadmin) gid=1043351(diagadmin) groups=1043351(diagadmin),1043350(csmadmin)
c460mgt01:/home/diagadmin > xdcp c460c034 1 /tmp/1
c460mgt01:/home/diagadmin > echo $?
0

```
```
c460c034:/tmp > ls -l 1
-rw-r--r-- 1 root root 2525 Mar 13 15:05 1
```

gurevichmark:
I think this is the same as issue #1001

gurevichmark:
This issue is fixed by PR #2689 
Please verify and close this issue if satisfactory.

sanomiya:
I have no way of testing this fix in my development environment . Closing it.
Will reopen, if problem persists when this version is installed in r92.

PR #2895, opened 2017-04-20
Add test case for verify the github issue #2637: Preserve the user ownership of the file for xdcp command

neo954:
Add one cases in this pull request.
This case is added for issue #2637

[Case index]
Case Name: xdcp_nonroot_user
description: The test case add a ordinary system user on the manage node and compute node, and then use this non-root system user to perform `xdcp`, and check the owner of the copied file after that.

### DIFF
--- a/xCAT-test/autotest/testcase/xdcp/cases1
+++ b/xCAT-test/autotest/testcase/xdcp/cases1
@@ -1,4 +1,4 @@
-start:xdcp_github_issue2637
+start:xdcp_nonroot_user
 cmd:useradd -m xyzzy
 check:rc==0
 cmd:( cd ~ && tar cf - .xcat .ssh ) | ( cd ~xyzzy && tar xf - )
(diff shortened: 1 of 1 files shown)


hu-weihua:
@neo954, It seems you linked a wrong issue number in the post box.  2537 should be change to 2637

xCAT-test/autotest/testcase/xdcp/cases1
```diff
@@ -0,0 +1,22 @@
+start:xdcp_github_issue2637
```
hu-weihua: please add description for this case

hu-weihua:
Please follow the template to pull a request against test case. https://github.com/xcat2/xcat-process/blob/master/basic-process/The%20Template%20of%20Pull%20Request%20of%20Test%20Case.rst

xCAT-test/autotest/testcase/xdcp/cases1
hu-weihua: I think one issue brought up some xcat problems in one scenario, we add test case to cover this scenario. So the case name and description should be expressed against the scenario, not the issue number. Could you change the name to express problem itself? thanks

hu-weihua:
I have reviewed the change and agree to merge, thanks @neo954

PR #2897, opened 2017-04-21
Using bulk operation to invoke makedhcp to imporve nodeset perforamnce (#2024)

robin2008:
This patch aims to fix the 'nodeset' command performance issue when using 'petitboot' netboot attribute for ppc64le nodes. It leverage bulk operation to invoke 'makedhcp', instead of run 'makedhcp' in a loop for each node one by one. (The same design as 'nodeset' in xnba plugin for x86_64 nodes)
In a cluster with thousandths nodes, it could reduce the time of running 'makedhcp' greatly.


robin2008:
To fix one performance issue in #2024

immarvin:
I am ok with the change

PR #2902, opened 2017-04-21
For issue xCAT Object Name Format error when trying to name osimages …

junxiawang:
…starting with rhels7.3 #2157
when rhels7.3 osimage could create the case will pass


immarvin:
i am ok with the change

hu-weihua:
@junxiawang , could you follow the rule https://github.com/xcat2/xcat-process/blob/master/basic-process/The%20Template%20of%20Pull%20Request%20of%20Test%20Case.rst to pull a request? thanks

PR #2849, opened 2017-04-11
modify installation testcase for issue:2834

junxiawang:
@hu-weihua ,I modify diskfull and diskless install cases for syncfile for issue #2834 .
If diskfull and diskless case passed then syncfile function is supported.

PR #2922, opened 2017-04-24
modify synclist file folder

junxiawang:
an enhancement for PR https://github.com/xcat2/xcat-core/pull/2849
The ``/tmp`` directory might be cleaned up across system reboots on ubuntu

I modify synclist folder for on ubuntu could not create synfile under /tmp
@immarvin please help review


immarvin:
agree to merge

PR #2943, opened 2017-04-26
docker version updated and the output for docker command changed so m…

junxiawang:
…odify the testcase

Modify dockercommand testcase for the docker  changed and docker command's output will changed .So we could not using output to check the test result and only using return value .

Before modify the case:
1.The output should be:
........ "Stop the container before attempting removal or use -f"
When using rmdocker $$CONTAINER if container is running
2.But now the docker version changed:
host12c04: Error: {"message": "You cannot remove a running container c35494a228182b1b9167ecb547cd7917087f0dc83123106d19353d63c15a430f. Stop the container before attempting removal or force remove"}


zet809:

PR #2950, opened 2017-04-27
modify makehosts_n_r testcase

junxiawang:
Modify makehost_n_r testcase 
For sometime autotest could not get the node information in "/etc/hosts",so I add sleep after makehosts and cat /etc/hosts to probe the issue.


bybai:
add sleep and cat to debug the makehost_n_r testcase failure. I am OK to merge it .

Issue #2935, opened 2017-04-25
[PGS support]PGS diskful deployment support without manual step


zet809:
1， Trace and work around udevadmin issue ：https://bugzilla.linux.ibm.com/show_bug.cgi?id=153609
2， Trace issue #2905

bybai:
Discussed in the meeting, closed this task, and open a new task https://github.com/xcat2/xcat-core/issues/3011 to trace LTC issue.

bybai:
I verified， do not need "rd.dm=0 rd.md=0 nodmraid" in kcmdline, default petitboot file is enough for diskful installation. For example:
```
]# vi p9euh01
#install rhels7.4-pegas-by-ppc64le-compute
default xCAT
label xCAT
        kernel http://10.6.29.1:80/tftpboot/xcat/osimage/rhels7.4-pegas-by-ppc64le-install-compute/vmlinuz
        initrd http://10.6.29.1:80/tftpboot/xcat/osimage/rhels7.4-pegas-by-ppc64le-install-compute/initrd.img
        append "quiet inst.repo=http://10.6.29.1:80/install/rhels7.4-pegas-by/ppc64le inst.ks=http://10.6.29.1:80/install/autoinst/p9euh01 BOOTIF=70:e2:84:14:24:84  ip=dhcp  inst.loglevel=debug inst.syslog=10.6.29.1 inst.sshd"
```

bybai:
Fixed in #3038

PR #2953, opened 2017-04-27
work around installer hang for udevadm command in p9 pegas.

bybai:
Work around the issue: installation hang in anaconda with "TypeError: argument of type 'NoneType' is not iterable" after running of "udevadm info --attribute-walk --name=/dev/sda" in %pre section of kickstart.
LTC issue link: https://bugzilla.linux.ibm.com/show_bug.cgi?id=153609

Traced in github task: #2935

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/getinstdisk
+++ b/xCAT-server/share/xcat/install/scripts/getinstdisk
@@ -141,8 +141,22 @@ if [ -z "$install_disk" ]; then
         disk_wwn=$(echo $output_for_wwn | $utolcmd)
         output_for_path=$(IFS= ;echo $disk_info | grep DEVPATH | cut -d "=" -f2)
         disk_path=$(echo $output_for_path | $utolcmd)
-        disk_driver=$(udevadm info --attribute-walk --name=$disk | grep DRIVERS| grep -v '""'| grep -v '"sd"'|
+
+        # Work around the issue Pegas running on Power9 installation hang in anaconda 
+        # with "TypeError: argument of type 'NoneType' is not iterable", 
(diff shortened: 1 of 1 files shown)


immarvin:
i am ok to merge this.

Since this is a bug in OS installer, the root fix should be delivered by the OS distribution. 
The HD driver type is the last criterial for the disk selection, the disk selection logic based on WWN number should work for 95% cases.  The workaround in this PR is adequate for xCAT in this phase.

thanks @bybai

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -141,8 +141,22 @@ if [ -z "$install_disk" ]; then
...
+        # with "TypeError: argument of type 'NoneType' is not iterable", 
+        # after running of "udevadm info --attribute-walk --name=/dev/sda" in %pre section of kickstart 
+        exec 42>&1
+        anaconda_version=$(anaconda --version 2>&1 >&42)
+        exec 42>&-
+        cpu_model=$(awk '/model/ { print $NF }' /proc/cpuinfo)
+        p9_bool=$(grep POWER9 /proc/cpuinfo|head -n 1|awk '{print $3}')
+        if [ "$anaconda_version" == "anaconda 21.48.22.93-1" -a \
+             "$cpu_model" == "0000000000000000" -a "$p9_bool" == "POWER9" ]; then
+            disk_driver=""
```
daniceexi: What does that mean that set **disk_driver=""**

bybai: disk_driver is for find first available block device. If can get disk_driver for disks, sort disks by WWN/PATH and driver type, select the first one. If disk_driver="", select the default one "/dev/sda".

PR #2954, opened 2017-04-27
modify diskfull and diskless hirerarchy testcase for synclist could n…

junxiawang:
…ot work on these cases
Add debug information on diskfull and diskless hirerarchy testcases for synlist could not work on these cases.


zet809:

Issue #2942, opened 2017-04-25
Handle case where user defined snmp auth


markprez:
Discussed a problem with Casandra where we tried to bring in a switch that already had snmp auth defined. This issue to handle defined snmp auth,

immarvin:
hi @markprez @cxhong , what did you plan to do on this ticket?

cxhong:
currently,  xCAT used postscripts "enablesnmp" to config snmp for cumulus switch with default username and password.   If user defined snmp information in the switches table,  enablesnmp postscripts will not pick up and cause switch_macmap function failed and any compute nodes connect to the cumulus switch will not able to discovery.   To fix this,  1) we need to get snmp information from switches table and pass to ```enablesnmp``` postscripts, or 2)  use ```configonie --snmp``` command to send the snmp information to the switch.

whowutwut:
@cxhong If you think this is resolved, we can close it on behalf of Mark.

cxhong:
sure, I think we can close this one

PR #2956, opened 2017-04-27
Supports snmpv3 configuration based on the attributes defined in the switches table

cxhong:
This is for issue #2942,   If user defined snmpv3 user in the switches table for ONIE switch,  we need to add those attribute to the snmpd configuration file on the switch.  otherwise the switcch_macmap command will failed and any nodes connect to that switch will not able to discovery.   
In order to do this, we modified ```configonie --switches onie_switch --snmp```,  it will get snmp attributes from switches table and process following steps on the onie switch:
1) stop snmpd service
2) remove snmp user in the /var/lib/snmp/snmpd.conf file if already defined
3) create or modify snmp user in the /etc/snmp/snmpd.conf
4) start snmpd service

### DIFF
--- a/docs/source/advanced/networks/onie_switches/os_cumulus/install.rst
+++ b/docs/source/advanced/networks/onie_switches/os_cumulus/install.rst
@@ -109,4 +109,39 @@ In order to utilize ``xcatprobe switch_macmap``, snmp needs to be enabled.  To e
 
     updatenode frame01sw1 -P enablesnmp
 
+To configuring SNMPv3 after enable snmp,  set user, authentication and/or encryption for the switches: ::
+
+    chdef frame01sw1 snmpauth=sha snmpassword=xcatpassw0rd snmprivacy=DES snmusername=xcatadmin
+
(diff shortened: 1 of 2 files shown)


docs/source/advanced/networks/onie_switches/os_cumulus/install.rst
```diff
@@ -109,4 +109,39 @@ In order to utilize ``xcatprobe switch_macmap``, snmp needs to be enabled.  To e
 
     updatenode frame01sw1 -P enablesnmp
 
+To configuring SNMPv3 after enable snmp,  set user, authentication and/or encryption for the switches: ::
+
+    chdef frame01sw1 snmpauth=sha snmpassword=xcatpassw0rd snmprivacy=DES snmusername=xcatadmin
```
immarvin: should this be ``snmpusername`` and ``snmppassword``?

cxhong: thanks, typo

immarvin:
i am o with the change

PR #2959, opened 2017-04-28
delete synclist debug information and modify the testcase for synclist could not work on hierarchy install

junxiawang:
Modify hierarchy testcases for synlist could not work 
Add nodeset CN osimage before "updatenode $$CN -f"  
Delete synlist debug information for we could not use it.


immarvin:
agree to merge

immarvin:
agree to merge

Issue #2805, opened 2017-04-06
Attempt to clone a vm with "clonevm" command produces confusing error


gurevichmark:
Running a `clonevm` command following an example in man page, produces a confusing message:

```
[root@fs2vm102 ~]# clonevm lsvm_test1 -t appserver
lsvm_test1: Error: //<server>/<path>/<mastername> support for cloning at this moment
[root@fs2vm102 ~]#
```

Also, reading the code, it seems that `clonevm` is only supported for VM with NFS mounted disks. Should documentation pages be modified to make that clear to the user ?

whowutwut:
Not in backlog, removing sprint label

gurevichmark:
Fixed by PR #2965. The new output is now:

```
[root@c910f02c05p03]# clonevm fs2vm105 -t appserver
fs2vm105: Error: VM cloning is only supported for nfs server vmstorage attribute.

[root@c910f02c05p03]# echo $?
1
```

PR #2965, opened 2017-05-01
Remove : for the output error message

gurevichmark:
This pull request fixes issue #2805 

`xCAT::SvrUtils::sendmsg()` function has a problem dealing with a `:` in the message text for error message types.
This pull request removes used `xCAT::MsgUtils->message()` instead of `xCAT::MsgUtils->message()`

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -2712,7 +2712,9 @@ sub promote_vm_to_master {
         $target = $sourcedir . "/" . $target;
     }
     unless ($target =~ /^nfs:\/\//) {
-        xCAT::SvrUtils::sendmsg([ 1, "KVM plugin only has nfs://<server>/<path>/<mastername> support for cloning at this moment" ], $callback, $node);
+        my $rsp;
+        push @{ $rsp->{data} }, "VM cloning is only supported for nfs server vmstorage attribute. Current setting is $target";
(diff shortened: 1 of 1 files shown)


gurevichmark:
Error output is now:
```
 clonevm fs2vm105 -t appserver
Error: VM cloning is only supported for nfs server vmstorage attribute. Current setting is dir:///var/lib/libvirt/images/appserver
[root@c910f02c05p03 xcat]# 
```

xCAT-server/lib/xcat/plugins/kvm.pm
```diff
@@ -2712,7 +2712,9 @@ sub promote_vm_to_master {
         $target = $sourcedir . "/" . $target;
     }
     unless ($target =~ /^nfs:\/\//) {
-        xCAT::SvrUtils::sendmsg([ 1, "KVM plugin only has nfs://<server>/<path>/<mastername> support for cloning at this moment" ], $callback, $node);
+        my $rsp;
+        push @{ $rsp->{data} }, "VM cloning is only supported for nfs server vmstorage attribute. Current setting is $target";
```
whowutwut: Picky comment, does it make more sense to the user if we write  `..supported for vmstorage=nfs attribute.  Current setting is $target`

whowutwut: Oh, just realized that it has to be `nfs://....` as the attribute.

Issue #2913, opened 2017-04-24
[openbmc]rspconfig support option: ip/netmask/gateway/vlan


zet809:
Acceptance: support option ip/netmask/gateway/vlan for rspconfig against openbmc

whowutwut:
In the process of investigating this new support Issue, it would be good if we can put in some description pointing to openbmc URLs where we find out the endpoint that we need to hit to make the changes required here..

xuweibj:
The url that provided by openbmc team is ``/xyz/openbmc_project/network///0/`` now.
We start to work on ``rspconfig`` command according this. The interface is not enough for us to send real request.

xuweibj:
Interface:

```
 # rspconfig node1 [ip|netmask|gateway]=x.x.x.x
BMC [IP|Netmask|Gateway]: x.x.x.x
```

```
# rspconfig node1 ip=x.x.x.x netmask=x.x.x.x gateway=x.x.x.x
BMC IP: x.x.x.x
BMC Netmask: x.x.x.x
BMC Gateway: x.x.x.x
```

whowutwut:
I should have added this also, we need to be able to tag the VLAN using rspconfig. Is this possible?

whowutwut:
@xuweibj  Regarding `/xyz/openbmc_project/network///0/` in the [README.md](https://github.com/openbmc/phosphor-dbus-interfaces/blob/master/xyz/openbmc_project/Network/README.md) , you have to view it raw because the markdown is not rending correctly.  Then you can see the interfaces planned and we can code against that API.

xuweibj:
Hi @whowutwut , now the code is just framework, I will modify it when I got the final interface.

xuweibj:
The task was finished in pull request #2966 .(framework)

PR #2966, opened 2017-05-02
framework of rspconfig command for openbmc

xuweibj:
#2913 

1. Framework of ``rspconfig`` command for openbmc
    Supported options: ip/netmask/gateway/vlan
```
    # rspconfig node1 ip=x.x.x.x
    BMC IP: x.x.x.x
```

    Support interface:
```
    # rspconfig node1 ip=x.x.x.x netmask=x.x.x.x gateway=x.x.x.x vlan=x.x.x.x
    BMC IP: x.x.x.x
    BMC Netmask: x.x.x.x
    BMC Gateway: x.x.x.x
    BMC VLAN ID enabled: x.x.x.x
```

2. Modify status machine for ``rsetboot`` command (set)
    The flow is ``login request->login response->set request->set response->get request->get response``.

### DIFF
--- a/perl-xCAT/xCAT/Usage.pm
+++ b/perl-xCAT/xCAT/Usage.pm
@@ -132,11 +132,11 @@ my %usage = (
        rspconfig <noderange> [snmpdest|alert|community] [-V|--verbose]
        rspconfig <noderange> [snmpdest=<dest ip address>|alert=<on|off|en|dis|enable|disable>|community=<string>]
    BMC specific:
-       rspconfig <noderange> [vlan|ip|netmask|gateway|backupgateway|garp]
+       rspconfig <noderange> [ip|netmask|gateway|backupgateway|garp|vlan]
        rspconfig <noderange> [garp=<number of 1/2 second>]
        rspconfig <noderange> [userid=<userid> username=<username> password=<password>]
    OpenBMC specific:
-       rspconfig <noderange> [vlan|ip|netmask|gateway]
+       rspconfig <noderange> [ip|netmask|gateway|vlan]
    iDataplex specific:
        rspconfig <noderange> [thermprofile]
        rspconfig <noderange> [thermprofile=<two digit number from chassis>]
--- a/perl-xCAT/xCAT/Usage.pm
+++ b/perl-xCAT/xCAT/Usage.pm
@@ -132,9 +132,11 @@ my %usage = (
        rspconfig <noderange> [snmpdest|alert|community] [-V|--verbose]
        rspconfig <noderange> [snmpdest=<dest ip address>|alert=<on|off|en|dis|enable|disable>|community=<string>]
    BMC specific:
-       rspconfig <noderange> [ip|netmask|gateway|backupgateway|garp]
+       rspconfig <noderange> [vlan|ip|netmask|gateway|backupgateway|garp]
        rspconfig <noderange> [garp=<number of 1/2 second>]
        rspconfig <noderange> [userid=<userid> username=<username> password=<password>]
+   OpenBMC specific:
+       rspconfig <noderange> [vlan|ip|netmask|gateway]
    iDataplex specific:
        rspconfig <noderange> [thermprofile]
        rspconfig <noderange> [thermprofile=<two digit number from chassis>]
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -356,9 +356,6 @@ sub parse_args {
     my $noderange = shift;
     my $check = undef;
  
-    # disable function until fully tested
-    $check = unsupported($callback); if (ref($check) eq "ARRAY") { return $check; }
-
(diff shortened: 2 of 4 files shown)


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1009,4 +1025,57 @@ sub reventlog_response {
...
+            push @output, "BMC Gateway: $data";
+        }
+
+        xCAT::SvrUtils::sendmsg("$_", $callback, $node) foreach (@output);
+    }
+
+    if ($node_info{$node}{cur_status} eq "RSPCONFIG_SET_RESPONSE" and $response_info->{'message'} eq $::RESPONSE_OK) {
+        if ($status_info{RSPCONFIG_SET_RESPONSE}{ip}) {
+            $node_info{$node}{bmc} = $status_info{RSPCONFIG_SET_RESPONSE}{ip};
+            print "$node: DEBUG BMC IP is $node_info{$node}{bmc}\n";
```
whowutwut: Is only the IP setting supported at this time? Where is the debug message for Netmask and Gateway?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -412,7 +405,37 @@ sub parse_args {
...
+            return ([ 1, "No option specified for $command" ]);
+        }
+        #
+        # disable function until fully tested
+        #
+        $check = unsupported($callback); if (ref($check) eq "ARRAY") { return $check; }
 
+        my $setorget;
+        foreach $subcommand (@ARGV) {
+            if ($subcommand =~ /^(\w+)=(.*)/) {
```
whowutwut: Is there a reason why we need to do all the parsing here instead of in the rspconfig function down at line 544?  If we compare with the ipmi.pm code.... 

```
elsif ($command eq "rspconfig") {
        shift @{ $sessdata->{extraargs} };

        # set username or/and password for specified userid #
        if ($sessdata->{subcommand} =~ /userid|username|password/) {
            setpassword($sessdata);
        } else {
            if ($sessdata->{subcommand} =~ /=/) {
                setnetinfo($sessdata);
            } else {
                getnetinfo($sessdata);
            }
        }
    }
```

By doing this, we are able to reduce the complex parsing and move this code down to the set/get functions.

xuweibj: For ipmi, only can set one attribute value once. But for openbmc, we hope that we can support to set more attributes for one or more nodes once. But we do not support set and get at the same time. So we need check input here.

xCAT-server/lib/xcat/plugins/openbmc.pm
xuweibj: This part is used to change bmc's value in node_hash value. If ``set`` failed, in ``deal_with_response`` function will print error msg. And if success, will enter ``rspconfig_response``. And if did not set ``ip``, will get next status(get) and then send get request directly. But if is ``ip``, we must change bmc's value so that can send correct request to get. Maybe we will move error process from deal_with_response if we got correct response msg in the future.

xuweibj:
@whowutwut Thank you for your comments. Please help to review again.

perl-xCAT/xCAT/Usage.pm
```diff
@@ -132,9 +132,11 @@ my %usage = (
        rspconfig <noderange> [snmpdest|alert|community] [-V|--verbose]
        rspconfig <noderange> [snmpdest=<dest ip address>|alert=<on|off|en|dis|enable|disable>|community=<string>]
    BMC specific:
-       rspconfig <noderange> [ip|netmask|gateway|backupgateway|garp]
+       rspconfig <noderange> [vlan|ip|netmask|gateway|backupgateway|garp]
```
whowutwut: I think the vlan option should go at the end.

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -355,72 +355,47 @@ sub parse_args {
     my $extrargs = shift;
     my $noderange = shift;
     my $check = undef;
+
```
whowutwut: Hi @xuweibj Maybe some misunderstanding in my comments.  The `unsupported` function needs to be run at a command level, because pretty soon, we want to be able to turn on the function without the user `export XCAT_OPENBMC_DEVEL`.  

So these calls should be put back...

whowutwut: What I meant is the following code is very similar:
```
-        if (!defined($extrargs)) {		
 -            return ([ 1, "No option specified for rpower" ]);		
 -        }
```

and 
```
-        if (!defined($extrargs)) {		
 -            return ([ 1, "No option specified for $command" ]);		
 -        }
```

and it appears in multiple places, and it' inconsistent.  We are using $command in most of them and `rpower` in one of them.  So this way the output message is printed in a function and it will be the same for each call to this function.  Maybe we just pass in the command name.

xuweibj:
@whowutwut I have modified. Please help to review again. Thank you.

whowutwut:
I have reviewed and will merge.

PR #2968, opened 2017-05-02
Turn off xcatpost at beginning rather than end

jjohnson42:
If a postscript reboots, make sure that our logic for
self-termination is at the least executed.


immarvin:
I am ok with the change

PR #2970, opened 2017-05-02
Piflash man

aduffy19:


daniceexi:
@aduffy19 even the merge commit does not hurt anything, but we still should avoid to include meaningless merge commit in the pull request.

PR #2979, opened 2017-05-04
Add docs about rinv for openbmc

xuweibj:
Add docs to support rinv command for openbmc.

1. usage
2. manpage

### DIFF
--- a/xCAT-client/pods/man1/rinv.1.pod
+++ b/xCAT-client/pods/man1/rinv.1.pod
@@ -10,9 +10,13 @@ B<rinv> [B<-h>|B<--help>|B<-v>|B<--version>]
 
 B<rinv> I<noderange> {B<pci>|B<model>|B<serial>|B<asset>|B<vpd>|B<mprom>|B<deviceid>|B<guid>|B<firm>|B<diag>|B<dimm>|B<bios>|B<mparom>|B<mac>|B<all>}
 
-=head2 OpenPOWER server specific:
+=head2 OpenPOWER (using ipmi) server specific:
   
 B<rinv> I<noderange> {B<model>|B<serial>|B<deviceid>|B<uuid>|B<guid>|B<vpd>|B<mprom>|B<firm>|B<all>} 
+
+=head2 OpenPOWER (using openbmc) server specific:
+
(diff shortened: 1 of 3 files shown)


xCAT-client/pods/man1/rinv.1.pod
```diff
@@ -10,10 +10,14 @@ B<rinv> [B<-h>|B<--help>|B<-v>|B<--version>]
...
 
-=head2 OpenPOWER server specific:
+=head2 OpenPOWER (using ipmi) server specific:
   
 B<rinv> I<noderange> {B<model>|B<serial>|B<deviceid>|B<uuid>|B<guid>|B<vpd>|B<mprom>|B<firm>|B<all>} 
 
+=head2 OpenPOWER (using openbmc) server specific:
+
+B<rinv> I<noderange> {B<model>|B<serial>|B<deviceid>|B<uuid>|B<guid>|B<vpd>|B<mprom>|B<firm>|B<cpu>|B<dimm>|B<all>}
+
```
gurevichmark: There seems to be a mismatch between the synopsis in man page and "usage" output. In man page we use `{}` to indicate "one of",  but in "usage" output we use `[]` to indicagte "optional". 
I think it should be `[]` because everything is optional and user can just run `rinv <noderange>`.
The question is should this PR just change the OpenPOWER synopsis lines or all of them. (zVM appears to be using correct `[]`)

xuweibj: I just follow the format. If modify this,  we need to modify all.

gurevichmark:
My recommendation would be to modify **all** synopsis lines in `.pod` file to match the format in "Usage" output. "Usage" output correctly uses `[]` to indicate optional choices.

xuweibj:
Hi @gurevichmark , you are right. But I do not think modify it in this pull request is a good idea. We decide to create a new task to modify files for all commands.

xuweibj:
Hi @gurevichmark , do have any other comments about this pull request. If not, I think we could merge it first and then open an new issue to modify all synopsis lines.

gurevichmark:
@xuweibj Will merge this pull request. Please open a separate issue to fix man page synopsis and usage output.

PR #2980, opened 2017-05-04
add consport and taggedvlan attributes in openbmc table

xuweibj:
add consport and taggedvlan attributes in openbmc table


zet809:

whowutwut:
@xuweibj What's the purpose for the console port?  If it changes from 2200 to something else??

zet809:
Hi, @whowutwut , your are right. For example, for the qemu simulated openbmc environment, there must be most than 1 qemu VM(represent one openbmc) started on one host, then console client need to access different port of the host to forward data to the qemu VM.

gurevichmark:
Not sure if another PR was used, but changes to Schema.pm file cause `.pod` man pages to get generated and `.rst` files to be built. The `.rst` files need to be checked in.

PR #2981, opened 2017-05-04
modify makehosts and statelite case

junxiawang:
1.statelite ramdisk based could write on /
2.makehosts alway have timing issue 
So I modify the testcases


bybai:
I am OK to merge.

PR #2982, opened 2017-05-04
Fix issue #2794, Nodeset fails to generate /tftpboot config files if …

### LINKED ISSUE #2794: Nodeset fails to generate /tftpboot config files if ip attribute is not set
If `ip` attribute for the VM is not set, `nodeset` returns with `RC=0`, but config files for that VM are not generated in `/tftboot/grub2` and the log file has these entries:

```
xCAT: Allowing nodeset to lsvm_test1 osimage=rhels7.3-ppc64le-install-compute for root from localhost
lsvm_test1: grub2 netboot: stop configuration because of none sharedtftp and not on same network with its xcatmaster.
xCAT: grub2 netboot: no valid nodes. Stop the operation on this server.
```
This happens for VMs, but might also happen for real servers

@daniceexi suggested:
*If nodeset does not generate an usable configuration, it should display an error and return non-zero.*

robin2008:
Resolve #2794 

When sharedtftp=0 in site table, xcatd will broadcast the 'nodeset' request with the whole Node Range to all SNs. We cannot filter the node range first by explicit 'servicenode' definition in 'noderes' table prior to the broadcasting,  as all SNs or MN in the same subnet should be notified to make the changes on the corresponding boot configuration.

In previous implementation,  if the node does not have IP or the IP is not in the same network, it will only log but not complain to user. And the RC is still zero. It will cause confusing.

Now the patch will add some error reporting in such case, and to avoid too many redundant message.  **It is decided that only if 'servicenode' is explicitly defined for a node, but this service node cannot handle nodeset request for the node,  it will report error.**

### DIFF
--- a/xCAT-server/lib/xcat/plugins/grub2.pm
+++ b/xCAT-server/lib/xcat/plugins/grub2.pm
@@ -513,11 +513,35 @@ sub process_request {
     #if not shared tftpdir, then filter, otherwise, set up everything
     if ($request->{'_disparatetftp'}->[0]) { #reading hint from preprocess_command
         @nodes = ();
+        my $cur_xmaster;
+        my %managed = {};
+
+        # Get current server managed node list
+        my $sn_hash = xCAT::ServiceNodeUtils->getSNformattedhash(\@rnodes, "xcat", "MN");
+        foreach my $nn (keys %$sn_hash) {
+            unless ( xCAT::NetworkUtils->thishostisnot($nn) ) {
+                $cur_xmaster = $nn;
+                foreach (@{ $sn_hash->{$cur_xmaster} }) { $managed{$_} = 1; }
+                last;
+            }
+        }
+
         foreach (@rnodes) {
             if (xCAT::NetworkUtils->nodeonmynet($_)) {
                 push @nodes, $_;
             } else {
-                xCAT::MsgUtils->message("S", "$_: grub2 netboot: stop configuration because of none sharedtftp and not on same network with its xcatmaster.");
+                my $msg = "grub2 configuration file was not created for node [$_] because sharedtftp attribute is not set and the node is not on same network as this xcatmaster";
+                if ( $cur_xmaster ) {
+                    $msg .= ": $cur_xmaster";
+                }
+                if ( exists( $managed{$_} ) ) {
+                    # report error when it is under my control but I cannot handle it.
+                    my $rsp;
+                    $rsp->{data}->[0] = $msg;
+                    xCAT::MsgUtils->message("E", $rsp, $callback);
+                } else {
+                    xCAT::MsgUtils->message("S", $msg);
+                }
             }
         }
     } else {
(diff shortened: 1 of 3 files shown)


robin2008:
UT:
1, set sharedtftp=0 in site
2,create a node without IP or IP is not in the same subnet of MN
3, run nodeset to see the result:

```
[root@xcatmn xcat]# nodeset fake1 osimage
Error: Ignore xnba configuration for fake1 because of none sharedtftp and not on same network with this xcatmaster: xcatmn
[root@xcatmn xcat]# echo $?
1
[root@xcatmn xcat]#
```

4, define another node with good IP definition
```
[root@xcatmn xcat]# nodeset fake1,fake2 osimage
Error: Ignore xnba configuration for fake1 because of none sharedtftp and not on same network with this xcatmaster: xcatmn
fake2: install centos7.3-x86_64-compute
[root@xcatmn xcat]# echo $?
1
```

5, Test with active SNs - TBC

gurevichmark:
The code looks good to me. One small suggestion is to make the error message a little more clear.
For example, for `grub2` case, I would suggest:
"grub2 configuration file was not created for node $_ because sharedtftp attribute is not set and the node is not on same network as this xcatmaster"

xCAT-server/lib/xcat/plugins/petitboot.pm
```diff
@@ -404,11 +404,35 @@ sub process_request {
...
+                last;
+            }
+        }
+
         foreach (@rnodes) {
             if (xCAT::NetworkUtils->nodeonmynet($_)) {
                 push @nodes, $_;
             } else {
-                xCAT::MsgUtils->message("S", "$_: petitboot netboot: stop configuration because of none sharedtftp and not on same network with its xcatmaster.");
+                my $msg = "Ignore petitboot configuration for $_ because of none sharedtftp and not on same network with this xcatmaster";
```
daniceexi: Could we emphasis the node name in the message with something like **[nodename]**?

robin2008:
Update according to the comments, and after applying new changes:
```
@xcatmn xcat]# nodeset fake1,fake2,fake3 osimage
Error: xnba configuration file was not created for node [fake1] because sharedtftp attribute is not set and the node is not on same network as this xcatmaster: xcatmn
Error: xnba configuration file was not created for node [fake3] because sharedtftp attribute is not set and the node is not on same network as this xcatmaster: xcatmn
fake2: install centos7.3-x86_64-compute
```

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -513,11 +513,35 @@ sub process_request {
     #if not shared tftpdir, then filter, otherwise, set up everything
     if ($request->{'_disparatetftp'}->[0]) { #reading hint from preprocess_command
         @nodes = ();
+        my $cur_xmaster;
+        my %managed = {};
+
+        # Get current server managed node list
+        my $sn_hash = xCAT::ServiceNodeUtils->getSNformattedhash(\@rnodes, "xcat", "MN");
+        foreach my $nn (keys %$sn_hash) {
```
immarvin:
````
#-----------------------------------------------------------------------------

=head3 determinehostname  and ip address(s)

  Used on the service node to figure out what hostname and ip address(s)
  are valid names and addresses
  Input: None
  Output: ipaddress(s),nodename
=cut

#-----------------------------------------------------------------------------
````


``xCAT::NetworkUtils->determinehostname()`` is a subroutine to obtain the node name of the current node, maybe it can be used here to determine the node name of the SN to avoid the ``foreach`` loop to improve the performance.

immarvin:
i am ok with the change

Issue #2967, opened 2017-05-02
 xCAT::SvrUtils::sendmsg() does not properly handle ":" in the error message text


gurevichmark:
There seems to be a problem how xCAT::SvrUtils::sendmsg() parses the error message text containing `:`

For example:
1. If message is error type and there is NO `:` in the text, it gets displayed properly:
    Coded message: `xCAT::SvrUtils::sendmsg([ 1, "Mark1-Mark2"], $callback, $node);`
    Output: `fs2vm105: Error: Mark1-Mark2`

2. If message is error type and there is `:` in the text, it is NOT displayed properly:
    Coded message: `xCAT::SvrUtils::sendmsg([ 1, "Mark1:Mark2"], $callback, $node);`
    Output: `fs2vm105: Error: Mark2`

3. If message is NOT error type and there is  `:` in the text, it gets displayed properly (but an extra space is inserted after the `:` :
    Coded message: `xCAT::SvrUtils::sendmsg([ 0, "Mark1:Mark2"], $callback, $node);`
    Output: `fs2vm105: Mark1: Mark2`

immarvin:
hi @gurevichmark ,

The response format in xCAT defined in ``perl-xCAT/xCAT/Client.pm`` is like this:

````
1071 # Format of the response hash:
1072 #  {data => [ 'data str1', 'data str2', '...' ] }
1073 #
1074 #    Results are printed as:
1075 #       data str1
1076 #       data str2
1077 #
1078 # or:
1079 #  {data => [ {desc => [ 'desc1' ],
1080 #              contents => [ 'contents1' ] },
1081 #             {desc => [ 'desc2 ],
1082 #              contents => [ 'contents2' ] }
1083 #                :
1084 #            ] }
1085 #    NOTE:  In this format, only the data array can have more than one
1086 #           element. All other arrays are assumed to be a single element.
1087 #    Results are printed as:
1088 #       desc1: contents1
1089 #       desc2: contents2
1090 #
1091 # or:
1092 #  {node => [ {name => ['node1'],
1093 #              data => [ {desc => [ 'node1 desc' ],
1094 #                         contents => [ 'node1 contents' ] } ] },
1095 #             {name => ['node2'],
1096 #              data => [ {desc => [ 'node2 desc' ],
1097 #                         contents => [ 'node2 contents' ] } ] },
1098 #                :
1099 #             ] }
1100 #    NOTE:  Only the node array can have more than one element.
[...]

PR #2983, opened 2017-05-04
Parsing : fix for sendmsg() error message

gurevichmark:
Fixes issue #2967 
Preserve original text string to use when displaying error message from sendmsg().


immarvin:
i am ok with the change

Issue #2809, opened 2017-04-07
[openbmc]enhance bmcdiscover for openbmc


zet809:
Acceptance: able to discovery an openbmc with ip, mtms.

PR #2986, opened 2017-05-05
Enhance bmcdiscover to support openbmc

xuweibj:
#2809 

1. Check node's type after forked for each node.
    Check whether 2200 port is open. If open, consider the node as openbmc type, enter ``bmcdiscovery_openbmc``. If not, enter ``bmcdiscovery_ipmi``.

2. In ``bmcdiscovery_openbmc``, get mtm and serial by using ``curl`` command.

3. Add ``hwtype`` attribute in Schema to support set hwtype for openbmc.

Output like this: (currently, can not get mtm value, so write **unknown** to test)
```
# bmcdiscover -s nmap --range 9.3.185.161 -u root -p 0penBmc -w -z
node-unknown-y430uf6cl002:
	objtype=node
	groups=all
	bmc=9.3.185.161
	cons=openbmc
	mgt=openbmc
	mtm=unknown
	serial=Y430UF6CL002
	bmcusername=root
	bmcpassword=0penBmc
```

```
# lsdef node-unknown-y430uf6cl002
Object name: node-unknown-y430uf6cl002
    bmc=9.3.185.161
    bmcpassword=0penBmc
    bmcusername=root
    cons=openbmc
    groups=all
    hwtype=bmc
    mgt=openbmc
    mtm=unknown
    nodetype=mp
    postbootscripts=otherpkgs
    serial=Y430UF6CL002
```


xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -976,4 +990,73 @@ sub bmcdiscovery_ipmi {
...
+            $ip .= ",,";
+        }
+        $ip .= ",mp,bmc";
+        if ($mtm and $serial) {
+            $node = "node-$mtm-$serial";
+            $node =~ s/(.*)/\L$1/g;
+            $node =~ s/[\s:\._]/-/g;
+        }
+
+        `rm -f $cjar_file`;
```
chenglch: Why do you call the shell command here?

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -976,4 +990,73 @@ sub bmcdiscovery_ipmi {
...
+    my $opw             = shift;
+    my $request_command = shift;
+    my $node            = sprintf("node-%08x", unpack("N*", inet_aton($ip)));
+
+    my $cjar_file = "/tmp/cjar_$ip";
+    my $data = '{"data": [ "' . $bmc_user .'", "' . $bmc_pass . '" ] }';
+
+    my $output = `curl -c $cjar_file -k -X POST -H \"Content-Type: application/json\" -d '$data' https://$ip/login`;    
+    
+    my $login_rsp = decode_json $output;
```
chenglch: Is the output always json?

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -976,4 +990,73 @@ sub bmcdiscovery_ipmi {
...
+    my $ip              = shift;
+    my $opz             = shift;
+    my $opw             = shift;
+    my $request_command = shift;
+    my $node            = sprintf("node-%08x", unpack("N*", inet_aton($ip)));
+
+    my $cjar_file = "/tmp/cjar_$ip";
+    my $data = '{"data": [ "' . $bmc_user .'", "' . $bmc_pass . '" ] }';
+
+    my $output = `curl -c $cjar_file -k -X POST -H \"Content-Type: application/json\" -d '$data' https://$ip/login`;
```
chenglch: Too much hard code

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -976,4 +990,73 @@ sub bmcdiscovery_ipmi {
...
+
+    my $output = `curl -c $cjar_file -k -X POST -H \"Content-Type: application/json\" -d '$data' https://$ip/login`;    
+    
+    my $login_rsp = decode_json $output;
+    if ($login_rsp->{status} eq "ok") {
+        my $req_output = `curl -b $cjar_file -k https://$ip/xyz/openbmc_project/inventory/system/chassis/motherboard`;
+        my $response = decode_json $req_output;
+        my $mtm = $response->{data}->{Model};
+        my $serial = $response->{data}->{SerialNumber}; 
+        $mtm =~ s/^\s+|\s+$//g;
```
chenglch: Can you add some comments or the output example here?

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -665,11 +678,12 @@ sub format_stanza {
 sub write_to_xcatdb {
     my $node = shift;
     my $data = shift;
+    my $mgt_type = shift;
     my ($bmcip, $bmcmtm, $bmcserial, $bmcuser, $bmcpass, $nodetype, $hwtype) = split(/,/, $data);
     my $request_command = shift;
     my $ret;
 
-    $ret = xCAT::Utils->runxcmd({ command => ['chdef'], arg => [ '-t', 'node', '-o', $node, "bmc=$bmcip", "cons=ipmi", "mgt=ipmi", "mtm=$bmcmtm", "serial=$bmcserial", "bmcusername=$bmcuser", "bmcpassword=$bmcpass", "nodetype=$nodetype", "hwtype=$hwtype", "groups=all" ] }, $request_command, 0, 1);
+    $ret = xCAT::Utils->runxcmd({ command => ['chdef'], arg => [ '-t', 'node', '-o', $node, "bmc=$bmcip", "cons=$mgt_type", "mgt=$mgt_type", "mtm=$bmcmtm", "serial=$bmcserial", "bmcusername=$bmcuser", "bmcpassword=$bmcpass", "nodetype=$nodetype", "hwtype=$hwtype", "groups=all" ] }, $request_command, 0, 1);
```
chenglch: Actually, I don't like this way which call xcat command within the xcat request.  I understand this is a history reason, but can you format this line? It's too long and hard to read or modify.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -575,7 +576,18 @@ sub scan_process {
 
                 # Set child process default, if not the function runcmd may return error
                 $SIG{CHLD} = 'DEFAULT';
-                bmcdiscovery_ipmi(${$live_ip}[$i], $opz, $opw, $request_command);
+
+                my $nmap_cmd = "nmap ${$live_ip}[$i] -p 2200 -Pn";
+                my $nmap_output = xCAT::Utils->runcmd($nmap_cmd, -1);
+                if ($nmap_output =~ /2200\/tcp (\w+)/) {
```
chenglch: As you have added a `if` condition, can you tell me why this child process always return `0`. Can you add some log for this? It is not trackable code, if command fails, nobody know where the error is.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
xuweibj: yes

chenglch: Please see the man page of curl.

whowutwut:
Is our p9 EUH machines also reporting `Unknown` back for the mtm?  I know the SN is currently all zeros...

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
xuweibj: I have tested, it's OK. And could get correct value.

xuweibj:
Hi @whowutwut , the machine I used has no mtm value, so I set it as ``unknown`` in my code to test.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
xuweibj: I have add ``else`` branch to print error msg.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
chenglch: Why do you confirm a network service is available all the time?

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -665,11 +681,17 @@ sub format_stanza {
...
     my $node = shift;
     my $data = shift;
+    my $mgt_type = shift;
     my ($bmcip, $bmcmtm, $bmcserial, $bmcuser, $bmcpass, $nodetype, $hwtype) = split(/,/, $data);
     my $request_command = shift;
     my $ret;
 
-    $ret = xCAT::Utils->runxcmd({ command => ['chdef'], arg => [ '-t', 'node', '-o', $node, "bmc=$bmcip", "cons=ipmi", "mgt=ipmi", "mtm=$bmcmtm", "serial=$bmcserial", "bmcusername=$bmcuser", "bmcpassword=$bmcpass", "nodetype=$nodetype", "hwtype=$hwtype", "groups=all" ] }, $request_command, 0, 1);
+    $ret = xCAT::Utils->runxcmd({ command => ['chdef'], 
+                                  arg => [ '-t', 'node', '-o', $node, "bmc=$bmcip", "cons=$mgt_type",
```
chenglch: indent

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -976,4 +998,77 @@ sub bmcdiscovery_ipmi {
...
+
+    my $cjar_file = "/tmp/cjar_$ip";
+    my $data = '{"data": [ "' . $bmc_user .'", "' . $bmc_pass . '" ] }';
+
+    my $output = `curl -c $cjar_file -k -X POST -H \"Content-Type: application/json\" -d '$data' https://$ip/login`;
+    
+    if ($output =~ /\"status\": \"ok\"/) {
+        my $req_output = `curl -b $cjar_file -k https://$ip/xyz/openbmc_project/inventory/system/chassis/motherboard`;
+        my $response = decode_json $req_output;
+        my $mtm = $response->{data}->{Model};
```
chenglch: As the http status code is checked with regex expression,  I have a little concern about this hash structure. If the hash object do not has 'data' or 'Model' field, this calls will fail unexpectedly.  Have you tested the decode_json subroutine. If `$req_output` is null or not json format string, what the value of $response will be? This result can help us decide if we need to add some check logic here.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -976,4 +998,77 @@ sub bmcdiscovery_ipmi {
...
+        my $serial = $response->{data}->{SerialNumber}; 
+
+        # delete space before and after
+        $mtm =~ s/^\s+|\s+$//g; 
+        $serial =~ s/^\s+|\s+$//g;
+
+        # format ip string for format_stanza function
+        $ip .= ",$mtm";
```
chenglch: Maybe this is not your problem, but `ip` is a strange variable, it concat all kinds of other kinds of string like: `mtm`, `serail` etc. A string with comma is constructed, and then call the `format_stanza` subroutine which uses `split(,)` to split this string into multiple parts again. So strange.  Why not to use array or hash directly?

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
xuweibj: I have tested it. If ``$req_output`` is null, ``$response`` is null too.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
xuweibj: I copied it from ``bmcdiscovery_ipmi``, so I did not modify it.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
chenglch: If it is some string , not null , just like "status" : "ok"?

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
chenglch: I understand, but you should think about it,  is `$ip` a meaningful variable?

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
daniceexi: did you use tab instead of 4 spaces to indent?

xuweibj: No, I used space. But I made ``arg`` align with ``command``.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -962,18 +986,99 @@ sub bmcdiscovery_ipmi {
...
+            $node =~ s/(.*)/\L$1/g;
+            $node =~ s/[\s:\._]/-/g;
+        }
+
+        unlink $cjar_file;
+    } else {
+        if ($output =~ /\"description\": \"Invalid username or password\"/) {
+            xCAT::MsgUtils->message("E", { data => ["BMC username/password is incorrect for $ip"] }, $::CALLBACK); 
+        }
+        return 1;
```
chenglch: return code of this subroutine is not unified. See line 1038 of this file just `return;`. If you keep `return 1` in this subroutine which indicate the error result, you should leverage this return code when calling `bmcdiscovery_openbmc` subroutine.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -936,44 +959,125 @@ sub bmcdiscovery_ipmi {
...
+    my $opz             = shift;
+    my $opw             = shift;
+    my $request_command = shift;
+    my $node            = sprintf("node-%08x", unpack("N*", inet_aton($ip)));
+
+    my $node_data = $ip;
+    my $cjar_file = "/tmp/cjar_$ip";
+    my $data = '{"data": [ "' . $bmc_user .'", "' . $bmc_pass . '" ] }';
+
+    my $output = `curl -c $cjar_file -k -X POST -H \"Content-Type: application/json\" -d '$data' https://$ip/login`;
```
chenglch: Sending rest request with curl command is only a workaround and the current method to check the http return code is too. xcat should provide a library for the rest api request.

PR #2989, opened 2017-05-05
add info on retry times to check whether the netwoek connection is ok

immarvin:
To help debug on failed cases that the xcatinstallpost script exit when the network connection between mn and cn are not setup


hu-weihua:

Issue #2810, opened 2017-04-07
[openbmc]enhance bmcsetup for openbmc


zet809:
Modify bmcsetup script to disable user configuration in-band through IPMI for WSP server.
Acceptance: bmcsetup able to run within host OS or genesis if possible.

zet809:
Hi, @whowutwut , the bmcsetup related to multiple ipmi in-band command supporting for WSP, we have the page https://github.ibm.com/vhu/c910env/blob/master/notes/openbmc/P1_items_openbmc_planning.md to record the test result for the ipmitool command used in bmcsetup. Will you pls check with Firmware team that for the failed command, is there any plan to support it?If yes, when? As for the OK command, will it be changed or deprecated some day? We really need their confirm, thx a lot.

zet809:
It seems the ipmitool in-band interface doesn't work at all:
```
[xCAT Genesis running on p9euh02 /]# ipmitool lan set 1 ipaddr 10.6.9.200
Setting LAN IP Address to 10.6.9.200
[xCAT Genesis running on p9euh02 /]# ipmitool lan print 1
Set in Progress         : Set Complete
Auth Type Support       : MD5
Auth Type Enable        : Callback : MD5
                        : User     : MD5
                        : Operator : MD5
                        : Admin    : MD5
                        : OEM      : MD5
IP Address              : 10.6.9.254
Subnet Mask             : 255.0.0.0
MAC Address             : 70:e2:84:14:24:74
Default Gateway IP      : 10.0.0.101
Cipher Suite Priv Max   : Not Available
Bad Password Threshold  : Not Available
```

daniceexi:
This is a high risk issue for xCAT!!!

zet809:
Hi, @whowutwut, I close this task and use #3051 to tracing the interface implementing by OpenBMC team.

PR #2990, opened 2017-05-05
enhance bmcsetup logic for openbmc

zet809:
Include 3 main modification:
1. add `isopenbmc` attribute for the `getbmcconfig` command send by `getipmi` script in xcat-genesis-scripts
2. modify `bmcsetup` script to disable unsupported ipmitool in-band command for openbmc
3. in bmcconfig.pm, read bmc information from `openbmc` table if the request come from an openbmc.
4. read default password with key `openbmc` from `passwd` table if the request come from an openbmc.

Before modifying:
```
[p9euh02 /]# ipmitool lan print 1
Set in Progress         : Set Complete
...
IP Address              : 10.6.9.200
Subnet Mask             : 255.0.0.0
MAC Address             : 70:e2:84:14:24:74
Default Gateway IP      : 10.0.0.101
...
```
Configuring BMC:
```
[p9euh02 /]# /bin/bmcsetup
ipmitool version 1.8.18
Invalid channel: 16
/bin/bmcsetup: line 151:  3709 Terminated              allowcred.awk
<167>May  8 05:24:45 xcat.genesis.bmcsetup: BMC Information obtained from xCAT
<167>May  8 05:24:45 xcat.genesis.bmcsetup: NUMBMCS=1 ==> BMC IP=10.6.9.254/255.0.0.0, GW=10.0.0.101, VLAN=off
<166>May  8 05:24:45 xcat.genesis.bmcsetup: IPMIVER=2.0, IPMIMFG=42817, XPROD=16975
<166>May  8 05:24:45 xcat.genesis.bmcsetup: Auto detecting LAN channel...
<166>May  8 05:24:45 xcat.genesis.bmcsetup: Detected LAN channel 1
Setting LAN IP Address to 10.6.9.254
Setting LAN Subnet Mask to 255.0.0.0
Setting LAN Default Gateway IP to 10.0.0.101
<167>May  8 05:24:47 xcat.genesis.bmcsetup: snooze for 1 second...
```
After modifying:
```
[p9euh02 /]# ipmitool lan print 1
Set in Progress         : Set Complete
...
IP Address              : 10.6.9.254
Subnet Mask             : 255.0.0.0
MAC Address             : 70:e2:84:14:24:74
Default Gateway IP      : 10.0.0.101
...
```
For the task #2810

### DIFF
--- a/xCAT-genesis-scripts/bin/bmcsetup
+++ b/xCAT-genesis-scripts/bin/bmcsetup
@@ -266,7 +266,7 @@ elif [ "$IPMIMFG" == "674" ]; then # DELL
         ipmitool delloem lan set shared with lom$BMCPORT &>/dev/null
         ipmitool delloem lan set shared with failover all loms &>dev/null
     fi
-elif [ "$IPMIMFG" == "42817" -a "$XPROD" == "16975" ]; then
+elif [ "$IPMIMFG" = "42817" -a "$XPROD" = "16975" ]; then
     ISOPENBMC=1
 fi
 
@@ -385,18 +385,40 @@ for b in $BMCVLAN; do
     let idev=idev+1
 done
 
-# After network commands are issued, pause to allow the BMC to apply (OpenPower)
-snooze 
+
 
 # update the node status to 'bmcready' for openbmc, no more configuration is needed.
 if [ ! -z "$ISOPENBMC" ]; then
+    # To enable network configuration for openbmc
+    let idev=0
+    TRIES=0
+    bmc_config_rc=0
+    # Set Channel Access to apply network setting 
+    #while ! ipmitool -d $idev lan set $LANCHAN access on; do
+    while ! ipmitool -d $idev raw 0x06 0x40 $LANCHAN 0x42 0x44 > /dev/null; do
+        snooze
+        let TRIES=TRIES+1
+        if [ $TRIES -gt $TIMEOUT ]; then 
+            bmc_config_rc=1
+            break;
+        fi
+    done
+
     # update the node status to 'bmcready'
     if [ ! -z "$XCATMASTER" ]; then
-        updateflag.awk $XCATMASTER 3002 "installstatus bmcready"
+        if [ "$bmc_config_rc" = "0" ]; then
+            # Wait for some time for the new network setting is ready 
+            snooze
+            updateflag.awk $XCATMASTER 3002 "installstatus bmcready"
+        else 
+            updateflag.awk $XCATMASTER 3002 "installstatus failed"
+        fi
     fi 
     rm -f /tmp/ipmicfg.xml
-    exit 0
+    exit $bmc_config_rc
 fi
+# After network commands are issued, pause to allow the BMC to apply (OpenPower)
+snooze 
 
 let idev=NUMBMCS-1
 for user in $BMCUS; do
--- a/xCAT-genesis-scripts/bin/bmcsetup
+++ b/xCAT-genesis-scripts/bin/bmcsetup
@@ -112,6 +112,13 @@ fi
 # Add ipmi_devintf module to allow the ipmitool operation in-band
 modprobe ipmi_devintf
 
+for parm in `cat /proc/cmdline`; do
+    key=`echo $parm|awk -F= '{print $1}'`
+    if [ "$key" = "xcatd" ]; then
+        XCATMASTER=`echo $parm|awk -F= '{print $2}'|awk -F: '{print $1}'`
+    fi
+done
+
 allowcred.awk &
 CREDPID=$!
 sleep 5
@@ -161,6 +168,7 @@ logger -s -t $log_label -p local4.info "IPMIVER=$IPMIVER, IPMIMFG=$IPMIMFG, XPRO
 #
 # IPMIMFG=2 = IBM
 # IPMIMFG=0 = OpenPower 
+# IPMIMFG=42817 and XPROD=16975 = OpenBMC
 #
 if [ "$IPMIMFG" == 2 ]; then #IBM
     if [ "$XPROD" == "220" ]; then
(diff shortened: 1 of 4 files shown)


xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -161,6 +168,7 @@ logger -s -t $log_label -p local4.info "IPMIVER=$IPMIVER, IPMIMFG=$IPMIMFG, XPRO
 #
 # IPMIMFG=2 = IBM
 # IPMIMFG=0 = OpenPower 
+# IPMIMFG=42817 and XPROD=16975 = OpenBMC
 #
 if [ "$IPMIMFG" == 2 ]; then #IBM
```
gurevichmark: Not sure if this is very important, but in some cases in this file we use strings for comparisons and in some cases integers. Should we make this consistent:
if [ "$xxxx" == "yyy" ]
if [ "$xxxx" == yyy ]

zet809: Hi, @gurevichmark , thx very much for the comments. You are right we shall keep the consistent. I do that for the OpenBMC part.
But for others, UT is needed if there is any modifications. But we don't need to put too much effort for the low priority items, right? What is your idea?

gurevichmark: Yes, I agree. Maybe just change the code that gets executed for this PR.

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -258,13 +266,20 @@ elif [ "$IPMIMFG" == "674" ]; then # DELL
         ipmitool delloem lan set shared with lom$BMCPORT &>/dev/null
         ipmitool delloem lan set shared with failover all loms &>dev/null
     fi
+elif [ "$IPMIMFG" = "42817" -a "$XPROD" = "16975" ]; then
+    ISOPENBMC=1
 fi
 
+if [ -z "$ISOPENBMC" ]; then
```
whowutwut: Suggest that we test for OpenBMC and override the default, and be consistent with this.  So default to non-openbmc and then override for OpenBMC always... 

Example: 
```

LAN_MED_TYPE="802.3" 
if [ ! -z "${ISOPENBMC+x}" ]; then 
    # OpenBMC overrides
    LAN_MED_TYPE="Other LAN"
    # other here...  
fi 
```

However, it looks like nothing else is defined at this time to put into this statement, but perhaps it will set the stage for future additions.

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -363,6 +385,38 @@ for b in $BMCVLAN; do
...
+
+
+# update the node status to 'bmcready' for openbmc, no more configuration is needed.
+if [ ! -z "$ISOPENBMC" ]; then
+    # To enable network configuration for openbmc
+    let idev=0
+    TRIES=0
+    bmc_config_rc=0
+    # Set Channel Access to apply network setting 
+    #while ! ipmitool -d $idev lan set $LANCHAN access on; do
```
whowutwut: This line should be deleted.

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -363,6 +385,38 @@ for b in $BMCVLAN; do
     let idev=idev+1
 done
 
+
+
+# update the node status to 'bmcready' for openbmc, no more configuration is needed.
+if [ ! -z "$ISOPENBMC" ]; then
+    # To enable network configuration for openbmc
+    let idev=0
```
whowutwut: Why do we need to set this `idev=0` if it seems to never increase or change...?

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -363,6 +385,38 @@ for b in $BMCVLAN; do
...
+            bmc_config_rc=1
+            break;
+        fi
+    done
+
+    # update the node status to 'bmcready'
+    if [ ! -z "$XCATMASTER" ]; then
+        if [ "$bmc_config_rc" = "0" ]; then
+            # Wait for some time for the new network setting is ready 
+            snooze
```
whowutwut: Did we check with firmware team about this snooze, it will default to 1 second, is it even needed?

xCAT-genesis-scripts/bin/getipmi
```diff
@@ -29,10 +29,17 @@ for LANCHAN in 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16; do
 done
 BMCMAC=`ipmitool lan print $LANCHAN|grep ^MAC|awk '{print $4}'` #bmcconfig may opt to use DHCP, if so we need to feed up the mac address
 #TODO: need a way to get the DUID the service processor may use, perhaps reserve that for 'ibmsetup' since spec doesn't touch ipv6?
-
+IPMIMFG=`ipmitool mc info |grep "^Manufacturer ID"|awk  '{print $4}'`
```
whowutwut: Should we consider having a common file to move some of the common functions into another place?  
```
[vhu@victorvm7 test]$ cat ipmi_common 
#!/bin/sh

IPMIMFG="42817"
XPROD="16975"

if [ "$IPMIMFG" == "42817" -a "$XPROD" == "16975" ]; then
    ISOPENBMC=1
else
    ISOPENBMC=0
fi
```

If we source it in another file, we will get the variables set....
```
[vhu@victorvm7 test]$ cat getipmi 
#!/bin/sh 

echo "BEFORE: ISOPENBMC=$ISOPENBMC"

. ./ipmi_common

echo "AFTER:ISOPENBMC=$ISOPENBMC"
```

Result: 
```
[vhu@victorvm7 test]$ ./getipmi 
BEFORE: ISOPENBMC=
AFTER:ISOPENBMC=1
```

One downside is that we can't just use relative path `. ./ipmi_common` and we have to make sure we don't get "File not found..."

xCAT-server/lib/xcat/plugins/bmcconfig.pm
```diff
@@ -96,20 +96,24 @@ sub process_request {
     my $request  = shift;
     my $callback = shift;
     my $node     = $request->{'_xcat_clienthost'}->[0];
+    my $open_table = "ipmi";
```
whowutwut: This variable name doesn't really make sense.. "open table"...   Suggest  `bmc_mgmt_type` ?

zet809:
Hi, @vhu, I had modified based on some of your comments.
Not sure what happened, I can not reply your comments directly, so reply here.
1. For the question about snooze, I hadn't asked Firmware, but my verification shows that the network can not work immediately after applied. Wait for 1 seconds works for our p9euh01 and p9euh02 server.
2. Use a common file, I agree with you that it is a good idea. But this code section doesn't work for both getipmi and bmcsetup. So I didn't modify it.
3. For the line 'set access on', the command will be supported by Firmware later, I can comment it out when it is supported. So, I didn't delete that line.
Thx!

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -363,6 +385,37 @@ for b in $BMCVLAN; do
...
+
+
+# update the node status to 'bmcready' for openbmc, no more configuration is needed.
+if [ ! -z "$ISOPENBMC" ]; then
+    # To enable network configuration for openbmc
+    TRIES=0
+    bmc_config_rc=0
+    # Set Channel Access to apply network setting 
+    #while ! ipmitool -d 0 lan set $LANCHAN access on; do
+    while ! ipmitool -d 0 raw 0x06 0x40 $LANCHAN 0x42 0x44 > /dev/null; do
```
whowutwut: @zet809  I just followed the github issue you opened, I think we should add some comments and point to the openbmc issue so that we don't forgot why we are doing this `raw 0x06 0x40` workaround...

PR #2992, opened 2017-05-05
Remove go-xcat test cases from all platforms except rhels7.3_ppc64le and ubuntu16.04.1_ppc64le

neo954:
This is due to these test cases consume too much Internet traffic.


hu-weihua:
agree to merge, thanks @neo954

PR #2993, opened 2017-05-05
Add rsync as an xCAT dependency

jjohnson42:


zet809:
Hi, @jjohnson42 , I had noticed that we had used rsync command in our code, maybe the rpm dependency calculation logic deal with the installation of rsync. 
Is there any issue have encountered if not write rsync in spec file?
If yes, what about ubuntu? Thx!

zet809:
Looks good to me, and I will create another PR to include rsync for ubuntu.

Issue #2350, opened 2016-12-22
Wcons IPMI not working due to xcat code


jjinjin123:
This problem occurs with xCAT 2.12.2.
The script (/opt/xcat/share/xcat/cons/ipmi) will query the BMC on a node and if it detects that the manufacturer id is 343 (intel), then it will use the -o intelplus option with ipmitool. But this function breaks conserver connections to one of our system.
Fixed by modifying the code thus:

if ($isintel) {
##Disabled as breaks Terri
#$inteloption = " -o intelplus";
}

zet809:
@jjinjin123 Which kind of server your system is? No an inter produced BMC? 
Will you pls show me the output of 'ipmitool ... mc info'?

jjinjin123:
Sorry for delay.
SGI UV10

[root@terri ~]# ipmitool mc info
Device ID                 : 33
Device Revision           : 1
Firmware Revision         : 0.20
IPMI Version              : 2.0
Manufacturer ID           : 343
Manufacturer Name         : Intel Corporation
Product ID                : 64 (0x0040)
Product Name              : Unknown (0x40)
Device Available          : yes
Provides Device SDRs      : no
Additional Device Support :
    Sensor Device
    SDR Repository Device
    SEL Device
    FRU Inventory Device
    IPMB Event Receiver
    IPMB Event Generator
    Chassis Device
Aux Firmware Rev Info     : 
    0x00
    0x23
    0x00
    0x20

whowutwut:
@jjohnson42  Any idea about this one?

zet809:
Hi, fixed with #2998, so close this issue.

PR #2998, opened 2017-05-08
Fix issue 2350: Wcons IPMI not working due to xcat code

zet809:
Fix issue #2350 

The modification is, if "Product ID" is "64", won't set the 'isintel' flag.


zet809:
Hi, @gurevichmark , will you pls help to merge this pull request? Thx!

PR #2999, opened 2017-05-08
Fix 2794 - Nodeset fails to generate /tftpboot config files if ip attribute is not set

robin2008:
This is a amend fix for #2794 - correct the wrong plugin name in trace log


immarvin:
agree to merge

Issue #3010, opened 2017-05-09
[DEV] please avoid using "exit" inside subroutines in xCAT plugin 


immarvin:
any process invoke the plugin thru ``plugin_command`` might die unconscious
the xCAT commands invoked by ``runxcmd``  might die unconscious and have no chance to capture the output message sent by ``sendmsg/callback/MsgUtils``

btw, https://github.com/xcat2/xcat-core/issues/2905 might be caused by this:

````
#in petitboot.pm:

476     # Fix the bug 4611: PowerNV stateful CN provision will hang at reboot stage#
477     if ($args[0] eq 'next') {
478         $sub_req->({ command => ['rsetboot'],
479                 node => \@nodes,
480                 arg  => ['default'] });
481         xCAT::MsgUtils->message("S", "xCAT: petitboot netboot: clear node(s): @nodes boot device setting.");
482     }
````

````
#in xcatd:

 459         eval {
 460             alarm(2);
 461             print $conn "ready\n";
 462             while (my $text = <$conn>) {
 463                 alarm(0);
 464                 print $conn "done\n";
 465                 $text =~ s/\r//g;
 466                 if ($text =~ /next/) {
 467                     my %request = (
 468                         command => ['nodeset'],
 469                         node    => [$node],
 470                         arg     => ['next'],
 471                     );
 472
 473                     # node should be blocked, race condition may occur otherwise
 474                     #my $pid=xCAT::Utils->xfork();
 475                     #unless ($pid) { # fork off the nodeset and potential slowness
 476                     plugin_command(\%request, undef, \&build_response);
 477
 478                     #exit(0);
 479                     #}
 480                     close($conn);
 481                 } elsif ($text =~ /installstatus/) {
````

Issue #3002, opened 2017-05-08
[fvt]2.13.5:openbmc :return value for rpower node (invalid node)  wrong


junxiawang:
env:openbmc
build:lsxcatd -v
Version 2.13.4 (git commit 6331ada77447f9eaaf4e6b71224f38d4d6643eff, built Sat May  6 06:15:59 EDT 2017)
How to reproduce:
```
1.
[root@c910f03c11k03 ~]# rpower testnode stat
This openbmc related function is unsupported and disabled. To bypass, run the following: 
	export XCAT_OPENBMC_DEVEL=YES
[root@c910f03c11k03 ~]# XCAT_OPENBMC_DEVEL=YES rpower testnode stat
Warning: Currently running development code, use at your own risk.  Unset XCAT_OPENBMC_DEVEL to disable.
testnode: Unable to get information from openbmc table
2.
[root@c910f03c11k03 ~]# echo $?
0
3.
[root@c910f03c11k03 ~]# lsdef testnode
Object name: testnode
    groups=all
    mgt=openbmc
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
[root@c910f03c11k03 ~]# 

```

xuweibj:
The issue was fixed in pull request #3016

junxiawang:
I have verified

PR #3016, opened 2017-05-10
modify all exit to return & change return value when configuration wrong

xuweibj:
#3010 #3002 
1. Rm ``exit`` in function, judge return of function to exit.
2. If can not get bmc/bmcusername/bmcpassword, return 1.

```
# rpower redfishtest status
Warning: Currently running development code, use at your own risk.  Unset XCAT_OPENBMC_DEVEL to disable.
redfishtest: Unable to get attribute password
# echo $?
1
```


immarvin:
i am ok with the change

Issue #2726, opened 2017-03-23
[customer]remoteshell failed in line 488 in statelite sles


bybai:
From Gilad Berman, lenovo:
In the remoshell script you are enabling ssh service (line 488) – 
```
else

    #service sshd restart

    # sshd is not enabled on SLES 12 by default

    # does not hurt anything to re-enable if it is enabled already

    enableservice sshd

    restartservice sshd

fi
```
For statelite, the /etc/systemd/system..  is a read only file system, of course. This makes the script fail with error message. 
One obvious solution is to place it in tmpfs (which we did), but ideally, we don’t want to do that, there is no reason for it to be in tmpfs. 
I guess a simple “if” will solve it 😊

daniceexi:
@bybai what's your idea for this change? disable the action **enable the sshd** should be accepted for diskless/stetelite.

bybai:
Hi @daniceexi ， There is comments in the remoteshell here:
```
else
    #service sshd restart
    # sshd is not enabled on SLES 12 by default
    # does not hurt anything to re-enable if it is enabled already
    enableservice sshd
    restartservice sshd
fi 
```
I think there is "sshd is not enabled on SLES 12 by default",  so I will have a test in sles12sp2 to look into it in 2.13.4 sprint1, then we can decide if disable the action `enable the sshd` in all the OS.

daniceexi:
@bybai I think the problem is the **remoteshell** script is common for both diskless and diskful. I do think 'enable sshd' is necessary for diskful. So we can just disable the 'sshd enablement' for diskless and statelite.

bybai:
@daniceexi , that's fine, thank you.

bybai:
fixed in #3017

bybai:
fixed in #3027

bybai:
Verified in auto test env.

PR #3017, opened 2017-05-10
fix issue2726 remoteshell failed in line 488 in statelite

bybai:
fix issue #2726 

remoteshell script is common for both diskless and diskful.  'enable sshd' is necessary for diskful. Just disable the 'sshd enablement' for diskless and statelite.

### DIFF
--- a/xCAT/postscripts/remoteshell
+++ b/xCAT/postscripts/remoteshell
@@ -486,7 +486,7 @@ else
     # sshd is not enabled on SLES 12 by default
     # does not hurt anything to re-enable if it is enabled already
     # and disable enable service for diskless and statelite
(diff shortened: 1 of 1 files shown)


xCAT/postscripts/remoteshell
```diff
@@ -485,10 +485,12 @@ else
     #service sshd restart
     # sshd is not enabled on SLES 12 by default
     # does not hurt anything to re-enable if it is enabled already
-    enableservice sshd
+    # and disable enable service for diskless and statelite
+    if [[ $NODESETSTATE == install ]]; then
```
immarvin: should better be ``[ "$NODESETSTATE" = install ]``

immarvin:
looks good for me

PR #3018, opened 2017-05-10
moidfy docker host setup testcase to increase wait time during install

junxiawang:
I modify docker host setup testcase.
Before modify the test case using 10m during installation but seems the waiting time is limited,so I increase the waiting for installation time to 60m.


hu-weihua:
I have reviewed and agree to merge. thanks.

Issue #3020, opened 2017-05-10
[DEV]pgs install will cause process "getpostscript for p9euh01" hang on MN


immarvin:
pgs install 
````
root      89050  0.0  0.0 167744 55872 pts/8    S+   May09   0:00  |       \_ xcatd: SSL listener
root      91748  0.0  0.0 168320 53376 pts/8    S+   03:43   0:01  |           \_ xcatd SSL: getpostscript for p9euh01
root      91749  0.0  0.0 169792 52288 pts/8    T    03:43   0:00  |           |   \_ xcatd SSL: getpostscript for p9euh01: getpos
root      91751  0.0  0.0 111808  3584 pts/8    T    03:43   0:00  |           |       \_ sh -c grep net.ifnames=0  >/dev/null 2>&
root      91752  0.0  0.0 110784  2944 pts/8    T    03:43   0:00  |           |           \_ grep net.ifnames=0
root      92671  0.0  0.0 168320 53376 pts/8    S+   05:21   0:00  |           \_ xcatd SSL: getpostscript for p9euh01
root      92672  0.0  0.0 169792 52288 pts/8    T    05:21   0:00  |           |   \_ xcatd SSL: getpostscript for p9euh01: getpos
root      92674  0.0  0.0 111808  3520 pts/8    T    05:21   0:00  |           |       \_ sh -c grep net.ifnames=0  >/dev/null 2>&
root      92675  0.0  0.0 110784  2880 pts/8    T    05:21   0:00  |           |           \_ grep net.ifnames=0
root      92948  0.0  0.0 168320 53376 pts/8    S+   05:54   0:00  |           \_ xcatd SSL: getpostscript for p9euh01
root      92949  0.0  0.0 169792 52288 pts/8    T    05:54   0:00  |               \_ xcatd SSL: getpostscript for p9euh01: getpos
root      92951  0.0  0.0 111808  3520 pts/8    T    05:54   0:00  |                   \_ sh -c grep net.ifnames=0  >/dev/null 2>&
root      92952  0.0  0.0 110784  2880 pts/8    T    05:54   0:00  |                       \_ grep net.ifnames=0
root      89146  0.0  0.0  28672 19648 ?        Ss   May09   0:00  \_ sshd: root@pts/9
````

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3021

PR #3021, opened 2017-05-10
check existence of netboot config file in postage.pm

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3020:

1. add the netboot config file path for netboot="petitboot"
2. check whether netboot config fix exists in postage.pm


bybai:
I am OK to the enhancement.

Issue #2816, opened 2017-04-07
[openbmc] rvitals support


zet809:
The vitals information for OpenBMC-Rest is provided by various Sensor objects. 
**API:**  [xyz/openbmc_project/Sensors](https://github.com/openbmc/phosphor-dbus-interfaces/tree/master/xyz/openbmc_project/Sensor)

The intent is to keep as much of the externals the same as existing machines...

gurevichmark:
Taking openbmc `rvitals` command to work on.

whowutwut:
With the above pull requests, I think we can close out this initial support task and open issues as we find problems or changes in the openbmc framework.

PR #3024, opened 2017-05-10
Initial pass for OpenBMC rvitals

gurevichmark:
Initial check in of OpenBMC `rvitals` command.  
Implements partially Issue #2816 

1. Output is not pretty yet, will be refined later:
```
p9euh01: /xyz/openbmc_project/sensors/fan_tach/fan1 7828
p9euh01: /xyz/openbmc_project/sensors/temperature/pcie 26750
p9euh01: /xyz/openbmc_project/sensors/fan_tach/fan2 7796
p9euh01: /xyz/openbmc_project/sensors/fan_tach/fan0 7828
p9euh01: /xyz/openbmc_project/sensors/fan_tach/fan3 7828
```
2. Fixed usage and man page for OpenPower `rvitals`, to show that all parameters are optional `[]` instead of `{}`. (no parameters defaults to `all`). Changed only OpenPower synopsis, other hardware probably the same, but can not verify, left as it was.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -168,6 +168,13 @@ my %status_info = (
     RSPCONFIG_SET_RESPONSE => {
         process        => \&rspconfig_response,
     },
+    RVITALS_REQUEST => {
+        method         => "GET",
+        init_url       => "$openbmc_project_url/sensors/enumerate",
+    },
+    RVITALS_RESPONSE => {
+        process        => \&rvitals_response,
+    },
 );
 
 $::RESPONSE_OK                  = "200 OK";
@@ -361,7 +368,7 @@ sub parse_args {
         return ([ 1, "No option specified for $command" ]);
     }
 
-    if (scalar(@ARGV) > 1 and ($command =~ /rpower|rinv|rsetboot/)) {
+    if (scalar(@ARGV) > 1 and ($command =~ /rpower|rinv|rsetboot|rvitals/)) {
         return ([ 1, "Only one option is supported at the same time" ]);
     }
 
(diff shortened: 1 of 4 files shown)


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1090,4 +1115,48 @@ sub rspconfig_response {
...
+    my $grep_string = $status_info{RVITALS_RESPONSE}{argv};
+    my $src;
+    my $content_info;
+    my $sensor_value;
+    
+    xCAT::SvrUtils::sendmsg("Processing command rvitals $grep_string", $callback, $node);
+
+    foreach my $key_url (keys %{$response_info->{data}}) {
+        my %content = %{ ${ $response_info->{data} }{$key_url} };
+        #print Dumper(%{$response_info->{data}}) . "\n";
```
whowutwut: I think we should either remove these debug lines, or build it into the debug when we set XCATBYPASS=1.    Wish there was better debug levels so that we could trace...   similar to ipmitool  -v, -vv, -vvv, -vvv after XCATBYPASS is set.

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1090,4 +1115,48 @@ sub rspconfig_response {
...
+        #print Dumper(%content) . "\n";
+        $sensor_value = $key_url . " " . $content{Value};
```
whowutwut: For this, I think we can for now just strip the data and print the last value.... 
```
node:  fan0
node: fan1
node: fan2
node: fan3
 etc
```

gurevichmark:
Implemented suggested changes.
The output now is:
```
p9euh01: fan_tach/fan1 7828
p9euh01: temperature/pcie 27000
p9euh01: fan_tach/fan2 7796
p9euh01: fan_tach/fan0 7575
p9euh01: fan_tach/fan3 7796
```

gurevichmark:
Yes, there is still more work to do for this command, like checking the parameters, filtering out the output for a specific type of sensors, sorting the output.
I wanted to do the initial pass for this command and get it merged, so that others can keep changing this file, reducing the likelihood of merge conflicts.

PR #3025, opened 2017-05-11
modify return value of bmcdiscover check and ipsource options

xuweibj:
Changed all message level for check and ipsource options to warning, so that the return code is 0 not 1.


zet809:

PR #3027, opened 2017-05-11
disable the 'sshd enablement' for diskless and statelite

bybai:
fix issue #2726

remoteshell script is common for both diskless and diskful. 'enable sshd' is necessary for diskful. Just disable the 'sshd enablement' for diskless and statelite.


immarvin:
i am ok with the change

Issue #3026, opened 2017-05-11
[openbmc]xcatd reported "Died at /opt/xcat/sbin/xcatd" when openbmc nodes updated its status from "installing" to "booting"


immarvin:
xcatd started in foreground on MN, when openbmc nodes updated its status from "installing" to "booting", the following message is observed: 
````
[root@stratton01 xCAT_plugin]# xcatd -f
xCAT: Allowing nodeset to p9euh01 osimage for root from localhost
monitorctrl::setNodeStatusAttributes called
readline() on closed filehandle GEN3 at /opt/xcat/sbin/xcatd line 448.
...
monitorctrl::setNodeStatusAttributes called
...
Not a reference at /opt/xcat/sbin/xcatd line 2450, <GEN33> line 1.
...
Died at /opt/xcat/sbin/xcatd line 1974, <GEN33> line 1.

xcatd: possible BUG encountered by xCAT install monitor service: Died at /opt/xcat/sbin/xcatd line 1974, <GEN33> line 1.

````

daniceexi:
@immarvin what's the command was called in the command to update the status? I mean we need to figure out an easy way to recreate the problem so that @chenglch or someone can take a look easily.

immarvin:
the steps to simulate the provision node status update, recreate the issue : 

on MN run:

start xcatd in foreground:
````
xcatd -f
````
then run
````
# nodeset p9euh01 osimage
````

ssh to the CN, run:
````
 /xcatpost/updateflag.awk stratton01 "installstatus installing"
````
and 
````
 /xcatpost/updateflag.awk stratton01
````


some hints to fix:
````
in petitboot.pm:
500     # Fix the bug 4611: PowerNV stateful CN provision will hang at reboot stage#
501     if ($args[0] eq 'next') {
502         $sub_req->({ command => ['rsetboot'],
503                 node => \@nodes,
504     #            environment => {XCAT_OPENBMC_DEVEL=>"YES"},
505                 arg  => ['default'] });
506         xCAT::MsgUtils->message("S", "xCAT: petitboot netboot: clear node(s): @nodes boot device setting.");
507     }
````

````
in xcatd:

2438 sub build_response {
2439
2440     # Handle responses from do_request calls made directly from a plugin
2441     #   Merge this response into the full response hash.  We'll collect all
2442     #   the responses and ship it back on the return to the plugin.
2443     #   Note:  Need to create a new "deep clone" copy of each response structure
[...]

PR #3029, opened 2017-05-11
modify output format when preprocess print error message

xuweibj:
fix issue #3026 
modify output format when preprocess print error message


zet809:

Issue #2803, opened 2017-04-06
rinv ocasionally produces BMC error without telling which node was in error.


ralphbellofatto:
we have a test where we perform rinv to fetch the firmware levels and check it.

For example, it does the following:
```
rinv c460c[004-009,011-013,015-018,037-040,042-051,053-058,060-061,064-070,073-080,082-085,087-089,133,135-136,139-141,143-147,149-153,155],c460login[01-02] firmware 2>/tmp/40MalkCHMF/stderr
expected firmware levels: bmc_firmware=2.13,sys_firmware=IBM-garrison-ibm-OP8_v1.12_2.70
(ERROR) Error: Error parsing FRU data from BMC

cat /tmp/40MalkCHMF/stderr
```

occasionally this test displays:
```
 Error: Error parsing FRU data from BMC
```

Where the error message parsing fru data from bmc fails to display which node the error came from.

zet809:
Thx, @ralphbellofatto ,we need to fix it in 2.13.4 release.

zet809:
Hi, @ralphbellofatto , this issue was fixed with Pull Request #3030 , pls reference it for more information. Thx!

whowutwut:
@zet809 @ralphbellofatto  Can you help close this out of it's resolved?

whowutwut:
Closing for now, please re-open if the issue persists.

PR #3030, opened 2017-05-11
Fix issue 2803: rinv ocasionally produces BMC error without telling w…

zet809:
…hich node was in error

Fix issue #2803 

The modification is:
When sendmsg, will get nodename from global variable when was used to store sessdata for a BMC when parsing FRU information.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -4525,8 +4525,9 @@ sub extractfield { #idx is location of the type/length byte, returns something a
     my $language = shift;
     my $data;
     if ($idx >= scalar @$area) {
-        xCAT::SvrUtils::sendmsg([ 1, "Error parsing FRU data from BMC" ], $callback);
-        return -1, undef, undef;
+        # The global_sessdata store the sessdata for a node when parsefru, and it is cleaned after parsefru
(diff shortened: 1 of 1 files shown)


whowutwut:
@zet809 we really should have some externals printed as part of the description of the pull request so that we can see the changes in behavior. .it would help with closing out issues.

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -4525,8 +4525,9 @@ sub extractfield { #idx is location of the type/length byte, returns something a
     my $language = shift;
     my $data;
     if ($idx >= scalar @$area) {
-        xCAT::SvrUtils::sendmsg([ 1, "Error parsing FRU data from BMC" ], $callback);
-        return -1, undef, undef;
+        # The global_sessdata store the sessdata for a node when parsefru, and it is cleaned after parsefru
+        xCAT::SvrUtils::sendmsg([ 1, "Error encountered when parsing FRU data from BMC" ], $callback, $global_sessdata->{node}, %allerrornodes);
+        return 0, undef, undef;
```
whowutwut: @zet809 Why do you return 0 here? instead of -1?

zet809: If most of the ipmi.pm code that to call extractfield, it use "unless" to check the first return value. If it is not 0, it will continue. 
For your concern, it is in line 4517 only, but it will exit the loop with the "while ($currsize > 0)" condition.

Issue #3036, opened 2017-05-11
[openbmc] bmcdiscover does not take password defined in password table 


whowutwut:
bmcdiscover function was added into xCAT tracked in this issue.   https://github.com/xcat2/xcat-core/issues/2809 

Running the command against a openbmc node using the following build: 
```
[root@stratton01 consoles]# lsxcatd -a
Version 2.13.4 (git commit d2fd44fbdf64834803d4b77a8a1ed0df10e5729d, built Thu May 11 06:15:46 EDT 2017)
This is a Management Node
dbengine=SQLite
```

The password is not taken from the password table, like it is on OpenPower (IPMI managed) 
```
[root@stratton01 consoles]# bmcdiscover --range 50.5.39.1
50.5.39.1,8335-GTA,2100EEA,,,mp,bmc
```

When not providing a password, it fails: 

```
[root@stratton01 consoles]# bmcdiscover --range 10.6.7.254
Warning: Invalid username or password for 10.6.7.254
```
But manually passing in the user/pass, it works...
```
[root@stratton01 consoles]# bmcdiscover --range 10.6.7.254 -u root -p 0penBmc
10.6.7.254,,Y130UF72700T,root,0penBmc,mp,bmc
```

Since we are going against the bmc using restful, we should be able to get the openbmc password from the passwd table.

xuweibj:
Hi @whowutwut 
For bmcdiscover, if there is no username or password input,  will get username and password for ipmi. So in this case, it got wrong username and password. I will try to fix this.

xuweibj:
The issue was fixed in pull request #3037

whowutwut:
Looks good, verified on today's build
```
[root@stratton01 ~]# lsxcatd -v
Version 2.13.5 (git commit 8ea4e5ae19806c890dd9149afa82fd7fd1b9a7fc, built Tue May 23 06:15:36 EDT 2017)
```

Testing against OpenPower (IPMI) 
```
[root@stratton01 ~]#  bmcdiscover --range 50.5.39.1
50.5.39.1,8335-GTA,2100EEA,,,mp,bmc
```
Testing against OpenPower OpenBMC
```
[root@stratton01 ~]# bmcdiscover --range 10.6.7.254
10.6.7.254,,Y130UF72700T,,,mp,bmc
```

Passwords are defined in `passwd` table

PR #3037, opened 2017-05-12
Get username and password from passwd table for openbmc discover

xuweibj:
#3036 

1. Define default username and password value for node with openbmc mgt.
2. Get username and password defined in passwd table.
3. If there is no openbmc in passwd table, use default value. Or use value in passwd(even empty).
4. Use ``openbmcusername`` and ``openbmcpassword`` in bmcdiscovery_openbmc function.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -41,6 +41,8 @@ if ($tempstring =~ /debian/ || $tempstring =~ /ubuntu/) {
 my $parent_fd;
 my $bmc_user;
 my $bmc_pass;
+my $openbmc_user;
+my $openbmc_pass;
 
 #-------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


zet809:
Looks good to me

whowutwut:
delegate review to @cxhong  , can you review this code ?

cxhong:
will this option also work for openbmc?
```
push @{ $rsp->{data} }, "\tDisplay the BMC IP configuration:\n";
    push @{ $rsp->{data} }, "\t\tbmcdiscover [-u bmc_user] [-p bmc_passwd] -i bmc_ip --ipsource";
````
if it is, the bmc_user and bmc_passwd will be incorrect if  get from passwd table.

**Comment by xuweibj** [comment]
@cxhong ， currently, bmcdiscover for openbmc did not support ``ipsource`` and ``check`` options.

**Comment by xuweibj** [comment]
Hi @cxhong , do you have any other comments for this pull request? Or could you help to merge it?

**Comment by whowutwut** [comment]
@xuweibj  I think we should be able to put in the code for ipsource option, even though it's not implemented yet, because the interface has been defined.  I think it might be `DHCPEnabled` here.. https://github.com/openbmc/phosphor-dbus-interfaces/blob/master/xyz/openbmc_project/Network/EthernetInterface.interface.yaml  

If the API is not clear, we need to bring it up to openbmc team.  

for `check` we can also try to login with the credentials.  It doesn't tell us what is wrong, but it does not authenticate if incorrect credentials. 
```
[root@stratton01 ~]# curl -c cjar -b cjar -k -H "Content-Type: application/json" -X POST https://10.6.9.254/login -d "{\"data\": [ \"abc\", \"invalid\" ] }"
{
  "data": {
    "description": "Invalid username or password"
  },
  "message": "401 Unauthorized",
  "status": "error"
```

xuweibj:
@whowutwut , I think the problem is not because of API. 
For options ``check`` and ``ipsource``, currently we use ipmitool-xcat command. If we want to support openbmc, we must check whether the node is openbmc. If openbmc we use restful, if not use ipmitool. So that we must add judgement in function for these options or restructure code. So we decide support them later.

PR #3038, opened 2017-05-12
update udevadmin work around

### LINKED ISSUE #2935: [PGS support]PGS diskful deployment support without manual step

bybai:
Fixed #2935 

refine the PR https://github.com/xcat2/xcat-core/pull/2953

Changed work around， if it is redhat and power9, will use this fix.

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/getinstdisk
+++ b/xCAT-server/share/xcat/install/scripts/getinstdisk
@@ -145,13 +145,10 @@ if [ -z "$install_disk" ]; then
         # Work around the issue Pegas running on Power9 installation hang in anaconda 
         # with "TypeError: argument of type 'NoneType' is not iterable", 
         # after running of "udevadm info --attribute-walk --name=/dev/sda" in %pre section of kickstart 
-        exec 42>&1
-        anaconda_version=$(anaconda --version 2>&1 >&42)
-        exec 42>&-
         cpu_model=$(awk '/model/ { print $NF }' /proc/cpuinfo)
         p9_bool=$(grep POWER9 /proc/cpuinfo|head -n 1|awk '{print $3}')
-        if [ "$anaconda_version" == "anaconda 21.48.22.93-1" -a \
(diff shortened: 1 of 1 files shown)


immarvin:
i am ok with the change

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -145,13 +145,10 @@ if [ -z "$install_disk" ]; then
...
         # after running of "udevadm info --attribute-walk --name=/dev/sda" in %pre section of kickstart 
-        exec 42>&1
-        anaconda_version=$(anaconda --version 2>&1 >&42)
-        exec 42>&-
         cpu_model=$(awk '/model/ { print $NF }' /proc/cpuinfo)
         p9_bool=$(grep POWER9 /proc/cpuinfo|head -n 1|awk '{print $3}')
-        if [ "$anaconda_version" == "anaconda 21.48.22.93-1" -a \
-             "$cpu_model" == "0000000000000000" -a "$p9_bool" == "POWER9" ]; then
+        anaconda --version >/dev/null  2>&1
+        if [ $? -eq 0 ] && [ "$cpu_model" = "0000000000000000" ] && [ "$p9_bool" = "POWER9" ]; then
```
whowutwut: @bybai Do we know what the intended CPU model is going to be and do we care? If not, maybe it's sufficient to just check for POWER9?

bybai: Hi @whowutwut , It is a work around, and now we do not know the  intended CPU model, current value is coming from p9euh01 system,  I think if there is new POWER9 system with different cpu_model coming, we can test if udevadmin issue exist or not. If the same issue is existed, we can just use POWER9, if the issue is not existed, that is to say, only  "$cpu_model" = "0000000000000000" system has the udevadmin problem. At that time, we can decide if cpu_model is needed or not. What is your idea?

PR #3040, opened 2017-05-12
Change bundle name of pegas+p9 from rhels7.104-ppc64el.bundle to rhels7.4-pegas_ppc64le.bundle

hu-weihua:
At first, I though we could use ``rhels7.104`` to indicate ``pegas``, in case some day later we need to support real rhels7.4. But after test, I found it did not work.  Because after ``copycds``, the default image name is `` rhels7.4-pegas-ppc64le-compute``,  failed to use ``rhels7.104`` to find out related image name `` rhels7.4-pegas-ppc64le-compute``. So, after some test, using item `` rhels7.4-pegas`` can work


bybai:
I am OK to merge it.

Issue #2831, opened 2017-04-07
[Test case]Add test case for 1276: nicips.enP5p7s0f1 can not removed from node definition


zet809:
Test case for #1276

junxiawang:
Test case:
chdef_nic

whowutwut:
@junxiawang  I'm not understanding the above comment.. .  is that the planned name for the testcase to be created?

Issue #1276, opened 2016-06-07
nicips.enP5p7s0f1 can not removed from node definition. 


wwzhubj:
Issue: After I set the nicips.enP5p7s0f1 to empty, nicips.enP5p7s0f1 is still there. 
chdef c712f7n07-pvt nicips.enP5p7s0f1=

[root@c712ems4 ~]# lsdef c712f7n07-pvt -i nicips
Object name: c712f7n07-pvt
    nicips.enP5p7s0f1=9.114.75.37

My env: 

[root@c712ems4 ~]# uname -a
Linux c712ems4 3.10.0-327.13.1.el7.openpower3.ppc64le #1 SMP Thu Apr 28 13:22:48 EDT 2016 ppc64le ppc64le ppc64le GNU/Linux
[root@c712ems4 ~]#

[root@c712ems4 ~]# lsxcatd -v
Version 2.12 (git commit 5df15f39216da9c64759d61b19fbabad1674727d, built Wed May 18 07:30:41 EDT 2016)

wwzhubj:
I also  found nicips.ib0, nicnetworks.ib0, nictypes.ib0 can not be removed when I set them  to empty using chdef. 
But for ib1 , it can be removed.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/1311

immarvin:
ping @tingtli , a test case should be added on this bug

wwzhubj:
I have verified it in below version
[root@c712ems4 build_input]# lsxcatd -v
Version 2.12.1 (git commit 0919bf649d99fc957236dbb42cfa3afc087ddad2, built Thu Jul  7 06:18:01 EDT 2016)

PR #3041, opened 2017-05-12
Add test case for 1276: nicips.enP5p7s0f1 can not removed from node definition.for issue:1276

junxiawang:
for issue:1276
[Test case]Add test case for 1276: nicips.enP5p7s0f1 can not removed from node definition #2831
nicips.enP5p7s0f1 can not removed from node definition. #1276
After add the testcase nicips could be set tu null


immarvin:
i am ok with the change

Issue #2662, opened 2017-03-17
Support Cumulus ZTP for autoconfiguration


mattaezell:
After a fresh Cumulus install, it will attempt to run Zero Touch Provisioning ([ZTP](https://docs.cumulusnetworks.com/display/DOCS/Zero+Touch+Provisioning+-+ZTP)).  If given the correct DHCP option, it will download a script and execute it.  We can use this to copy over xCAT keys and possibly do some of the `configonie` steps there as well.  Could also be used as a Genesis-like environment and support running things from the chain table.

immarvin:
hi @mattaezell , thanks for your proposal.
This is an interesting reminder for us, since we can not control finely for the cumulus installer as ordinary linux installers such as anaconda, autoyast.. etc,  we did not have a chance to insert some hooks to perform some user specified postbootscripts on fresh installation, the ZTP might be a hook to perform some interesting configuration and interaction with the management node. 

I will look into this with @cxhong

mattaezell:
Thanks @immarvin.  You could also hook into the ONIE installer by creating a Genesis script that runs instead of the Cumulus installer - I believe it has a busybox shell that you could use to send a discovery packet.  However I believe the switches we are getting for CORAL will have Cumulus preinstalled so we won't be able to take advantage of ONIE-level features (unless we manually reset them to ONIE mode).

cxhong:
For this issue, there are two parts.  
1)  use ZTP autoconfigure the cumulus switch
2) genesis like switch discovery.

as talked to @immarvin , genesis discovery may require  a lot of hook on cumulus switch.  we need to have more time to investigate the ONIE features.  So for the first phase,  we will work on to use ZTP auto configure the cumulus switch.

mattaezell:
Hello @cxhong.  We recently replaced our Cumulus switches, but we had to manually reconfigure them.  I looked into what it would take to get them to autodiscover, but I didn't want to duplicate or conflict with any work you have already done.  What all are you investigating, and how far along are you?  Thanks!

cxhong:
I was working to use ZTP autoconfigure the cumulus switch.  didn't have chance to look into autodiscover yet.  what method did u try to use autodiscover?

mattaezell:
[...]

PR #3059, opened 2017-05-15
Enable switches to run findme for autodiscovery (#2662)

mattaezell:
This lays some groundwork for the autodiscovery part of #2662

### DIFF
--- a/xCAT-server/lib/xcat/plugins/switch.pm
+++ b/xCAT-server/lib/xcat/plugins/switch.pm
@@ -293,15 +293,19 @@ sub process_request {
             #discovery working.  Food for thought.
             return;
         }
+        my $discoverswitch = 0;
+        if (defined $req->{nodetype} and $req->{nodetype}->[0] eq 'switch') {
+            $discoverswitch = 1;
+        }
         my $firstpass = 1;
         if ($mac) {
-            $node = $macmap->find_mac($mac, $req->{cacheonly}->[0]);
+            $node = $macmap->find_mac($mac, $req->{cacheonly}->[0], $discoverswitch);
             $firstpass = 0;
         }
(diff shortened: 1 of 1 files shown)


cxhong:
Thanks @mattaezell 
How did u able to send request?  did u create xml file?

mattaezell:
@cxhong I copied some code from dodiscovery in genesis, but simplified it because we don't need to handle BMCs, etc.

Draft version at:
https://gist.github.com/mattaezell/f385baa5bdccf6f63145489f3c172da2

Feel free to adapt that into your #3085 PR

zet809:
Hi, @cxhong , seems to me that when the switch send out the 'findme' request, it need to have 'nodetype=switch' attribute set, then our find_mac subroutine will search only switch node for this request. 
But, you are right that how to send a request is not included, you are reference xCAT-genesis-scripts/bin/dodiscovery for how to generate the findme request and send it to xCAT server.

Thx @mattaezell , do you think 'productname=cumulus' more make sense?

mattaezell:
Hello @zet809.  As you noticed, by default, *find_mac* skips switches.  If *discoverswitch* is true, then it does not skip the switches. *nodetype=switch* is generic and could be used for any type of switch, Cumulus, Mellanox (maybe through Puppet integration), or some future product.  *productname=cumulus* might be an additional element to send in the discovery packet.

cxhong:
@mattaezell , thanks so much for the scripts.  I think we can use for autodiscover cumulus process.

@zet809 , currently, the switchdiscover command used find_mac has special value for switch
````
xCAT-server/lib/xcat/plugins/switchdiscover.pm:        my $node = $macmap->find_mac($mac,0,1);
````
so, I think ```nodeype=switch```  should be good enough.

Issue #2937, opened 2017-04-25
if rpower state is accurate or not for openbmc


bybai:
When I execute rpower CN state, it show "on" state, but I login bmc, the power state is on but host state is Quiesced. rpower ignore Quiesced state, where do we record this host state?
The openbmc issue link: https://github.com/openbmc/openbmc/issues/1446
```
[root@stratton01 ~]# rpower p9euh01 state
Warning: Currently running development code, use at your own risk.  Unset OPENBMC_DEVEL and `restartxcatd` to disable.
p9euh01: on
```
login the bmc:
```
root@witherspoon:~# obmcutil state
CurrentBMCState:    xyz.openbmc_project.State.BMC.BMCState.Ready
CurrentPowerState:  xyz.openbmc_project.State.Chassis.PowerState.On
CurrentHostState:   xyz.openbmc_project.State.Host.HostState.Quiesced
root@witherspoon:~# cat /etc/*release
ID="openbmc-phosphor"
NAME="Phosphor OpenBMC (Phosphor OpenBMC Project Reference Distro)"
VERSION="ibm-v1.99.4"
VERSION_ID="ibm-v1.99.4-0-r0"
PRETTY_NAME="Phosphor OpenBMC (Phosphor OpenBMC Project Reference Distro) ibm-v1.99.4"
BUILD_ID="ibm-v1.99.4-0-r0"
```

xuweibj:
The issue was fixed in pull request #3075 .

bybai:
I verified this.

PR #3075, opened 2017-05-16
Modify output for rpower state when openbmc return Quiesced 

xuweibj:
#2937 

If receive ``CurrentHostState:   xyz.openbmc_project.State.Host.HostState.Quiesced`` info when use ``power state`` command to get openbmc's status, output will be like this:

```
# rpower simulation_test stat
simulation_test: quiesced
```


bybai:
I am ok to merge it.

PR #3079, opened 2017-05-17
In automation environment, disable SN when doing Migration and add 'stop' label for all provision cases for Dev debug

junxiawang:
1.Migration test just need to run under flat environment or will two dns server:SN and MN for CN
2.Add stop label for DEV to debug


hu-weihua:

daniceexi:
I dont' understand the words **or will two dns server:SN and MN for CN** in the description.

Issue #2909, opened 2017-04-23
[FVT] add a list of people who check in code during last 24 hours in the "xCAT Jenkins Test Report" 


hu-weihua:
Acceptance :
1 Depending on long commit id in ``build.info`` to calculate the list of people who check in code during last 24 hours. 

2 highlight this list in ``xCAT Jenkins Test Report``

[DEADLINE] : May 2

neo954:
The code is ready for review.

https://github.ibm.com/xcat2/jenkins/pull/109

PR #3084, opened 2017-05-17
Create a mailmap file to coalesce together commits by the same person

neo954:
This will make analyze the output of `git log` easier. This will help task #2909.

Please see the *MAPPING AUTHORS* section of `man git-shortlog` for details of git mailmap feature.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,69 @@
+BAI Yuan	<bybai@cn.ibm.com>
+Bruce Potter	<bp@us.ibm.com>
+CAO Li		<caoli@8638fb3e-16cb-4fca-ae20-7b5d299a9bcd>
+CAO Meng Meng	<bjcaomm@cn.ibm.com>
+CAO Meng Meng	<caomengmeng@users.noreply.github.com>
+CAO Meng Meng	<caomengmeng_1011@sina.com>
+Casandra Qiu	<cxhong@us.ibm.com>	<abcdqiu@gmail.com>
+Casandra Qiu	<cxhong@us.ibm.com>
+CHENG Long	<chenglch@cn.ibm.com>
+CHU De Gao	<degaochu@cn.ibm.com>	<chudegao@8638fb3e-16cb-4fca-ae20-7b5d299a9bcd>
+CHU De Gao	<degaochu@cn.ibm.com>
+CHU De Gao	<degaochu@cn.ibm.com>	<degaochu@ibm.com.cn>
+GAO Feng	<gfgaogf@cn.ibm.com>
+GAO Ling	<linggao@8638fb3e-16cb-4fca-ae20-7b5d299a9bcd>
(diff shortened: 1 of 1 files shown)


whowutwut:
@neo954 Do we need non active contributors to this list? (bruce, ling, lissa, meng meng guang cheng etc...)  Missing Jarrod and Lenovo folks who have been on more recent pull requests...

whowutwut:
Other question is why we need this in xcat-core repo, could we put this into the jenkins repo in GHE?

neo954:
@vhu, Yes, we do need this `.mailmap` file under the top directory of the `xcat-core` repository. Since we need to run `git log --use-mailmap` or `git shortlog` against `xcat-core` repository.

Yes, we can put this list in the `jenkins` repo. But when we do `git shortlog`, we still need to copy/move, or create a `.mailmap` file under `xcat-core` repository.

Put the `.mailmap` file here is more straightforward. It is just like we put `.gitignore` file in the repository.

neo954:
@vhu, Jarrod is already in the list. If you think we need add/remove entries to/from the list. We can revise the list.

Here is how I created this list in the first place. Run `git log master --pretty="%an <%ae>" | sort -u` against `xcat-core` git repository. All the other work is sort it out manually.

whowutwut:
Thanks @neo954  Instead of having two copies of this list, can we put this list in GHE under the jenkins project and wget it to the xcat-core repo before we need to consume it? 

 I'd like to keep a single copy somewhere, would prefer an internal repo first if users do not use it.

neo954:
@vhu, I do not see your point of user do not use it. If you mean common xCAT user do not use it. That is fine. But common xCAT user get their xCAT shipped with rpm packages or deb packages. They will not see the mail-map file.

Anyone access `xcat-core` git repository will get benefit from the mail-map file. Plus this mail-map file only make the author names and email addresses in the git log looks consistent. Otherwise no one knows `ertaozh` and `zet809` are the same person. And no one knows `Victor Hu` with email address `vhu@victorvm7.pok.ibm.com` is also use email `whowutwut@gmail.com` to do his work.

Reality is messy. This simple file helps us make life easier.

neo954:
@vhu, I do not see your point of put this mail-map file in GHE as well. Since this solution just make things more complex. We could have a more simple and straightforward solution. Which is put this mail-map file here, in the `xcat-core` repository.

neo954:
We will not have same mail-map list between GHE jenkins repo and here. Since GHE user profile is synchronized from the bluepage in the first place. So most of the code authors in have different style of his/her name spelling between GHE and here. For example `Victor Hu` vs `VICTOR K. HU`.

So there are two lists in two repositories. No relationship between them. We should maintain two lists separately.

neo954:
@zet809, Please help on review and merge this pull request.

zet809:

Issue #3060, opened 2017-05-16
Investigation of the ZTP based discovery and auto configuration of Cumulus switch


immarvin:
Background:

1. what is ZTP?
Zero touch provisioning  (ZTP) enables network devices to be quickly deployed in large-scale environments. On first boot, Cumulus Linux will invoke ZTP, which executes the provisioning automation used to deploy the device for its intended role in the network.

The provisioning framework allows for a one-time, user-provided script to be executed. You can develop this script using a variety of automation tools and scripting languages, providing ample flexibility for you to design the provisioning scheme to meet your needs. You can also use it to add the switch to a configuration management (CM) platform such as Puppet, Chef, CFEngine or possibly a custom, proprietary tool.

While developing and testing the provisioning logic, you can use the ztp command in Cumulus Linux to manually invoke your provisioning script on a device.

ZTP in Cumulus Linux can occur automatically in one of the following ways, in this order:

    Via a local file
    Using a USB drive inserted into the switch (ZTP-USB)
    Via DHCP

2. The basic idea of ZTP based switch discovery and auto configuration
Currently, xCAT provides the switch discovery and configuration mechanism initiated from the management node. With the help of ZTP, this discovery and configuration can be initiated by the switch itself on the switch power-on/port plugin, the process is like this: the cumulus switch request a ZTP script from MN via DHCP  request, then invoke the script on the switch. This script can be something like "findme" in xcat node discovery, The switch will report some switch information about itself to MN and create/update the switch definition on the MN, then initiate a configuration script or request from MN.  

Acceptance Criteria:
1.  A feasible work manual flow of ZTP based discovery and auto configuration of Cumulus switch
2.  mini-design of the ZTP based discovery and auto configuration

Difficulties:
1. how to trigger the ZTP process on a cumulus switch with OS provisioned and licence file installed?

see https://github.com/xcat2/xcat-core/issues/2662

cxhong:
the ztp is disable if license is already installed.
````
cumulus@cumulus:~$ sudo systemctl -l status ztp.service
● ztp.service - Cumulus Linux ZTP
   Loaded: loaded (/lib/systemd/system/ztp.service; enabled)
   Active: inactive (dead) since Tue 2017-05-16 12:34:12 UTC; 13min ago
     Docs: man:ztp(8)
 Main PID: 1328 (code=exited, status=2)

May 16 12:34:07 cumulus systemd[1]: Started Cumulus Linux ZTP.
[...]

PR #3085, opened 2017-05-17
Support ZTP process for Cumulus Switch

cxhong:
That's the init steps to support ZTP process on Cumulus switch.  for issue #3060 and #2662 
on the lab, cumulus switch already has OS and licensed installed.  ZTP service will disabled.  To enable the ZTP service and reset ZTP process for next reboot,  need to do following:
```
#ssh -l cumulus x.x.x.x
#$ sudo ztp -R
[sudo] password for cumulus:
Created symlink from /etc/systemd/system/multi-user.target.wants/ztp.service to /lib/systemd/system/ztp.service.
#sudo reboot
````
xCAT will receive DHCP request from cumulus switch with cumulus-provision-url option, then xCAT will send back DHCP Ack with the ZTP scripts.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -2271,6 +2271,7 @@ sub addnet6
             push @netent, "    option domain-search  $domainstring;\n";
         }
     }
+    
 
     my $nameservers = $netcfgs{$net}->{nameservers};
     if ($nameservers and $nameservers =~ /:/) {
@@ -2574,6 +2575,7 @@ sub addnet
             push @netent, "    option interface-mtu $mtu;\n";
         }
 
+
         #  add domain-search if not sles10 or rh5
         my $osv = xCAT::Utils->osver();
         unless (($osv =~ /^sle[sc]10/) || ($osv =~ /^rh.*5$/)) {
@@ -2591,6 +2593,8 @@ sub addnet
                 push @netent, "    option domain-search  $domainstring;\n";
             }
         }
+        #for cumulus ZTP process
+        push @netent, "    option cumulus-provision-url \"http://$tftp/install/postscripts/onieztp\";\n";
 
         my $ddnserver = $nameservers;
         $ddnserver =~ s/,.*//;
@@ -2928,6 +2932,7 @@ sub newconfig
     }
     push @dhcpconf, "option gpxe.no-pxedhcp 1;\n";
     push @dhcpconf, "option www-server code 114 = string;\n";
+    push @dhcpconf, "option cumulus-provision-url code 239 = text;\n";
     push @dhcpconf, "\n";
     push @dhcpconf, "omapi-port 7911;\n";            #Enable omapi...
     push @dhcpconf, "key xcat_key {\n";
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -2594,7 +2594,7 @@ sub addnet
             }
         }
         #for cumulus ZTP process
-        push @netent, "    option cumulus-provision-url \"http://$tftp/install/postscripts/onieztp\";\n";
+        push @netent, "    option cumulus-provision-url \"http://$tftp/install/postscripts/cumulusztp\";\n";
 
         my $ddnserver = $nameservers;
         $ddnserver =~ s/,.*//;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: xCAT/postscripts/onieztp [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+ 
+trap error ERR
+ 
+#Add Debian Repositories
+echo "deb http://http.us.debian.org/debian jessie main" >> /etc/apt/sources.list
+echo "deb http://security.debian.org/ jessie/updates main" >> /etc/apt/sources.list
+ 
+#push root ssh keys, config passwordless
+echo "CumulusLinux!" | sudo -S mkdir -p /root/.ssh
+echo "CumulusLinux!" | sudo -S /usr/bin/wget -O /root/.ssh/authorized_keys http://172.21.253.27/install/postscripts/_ssh/authorized_keys
```
mattaezell: This should be the management/service node that provided the DHCP packet, same for below

xCAT-server/lib/xcat/plugins/dhcp.pm
```diff
@@ -2591,6 +2593,8 @@ sub addnet
                 push @netent, "    option domain-search  $domainstring;\n";
             }
         }
+        #for cumulus ZTP process
+        push @netent, "    option cumulus-provision-url \"http://$tftp/install/postscripts/onieztp\";\n";
```
mattaezell: Should the script be called cumulusztp?  Regular onie doesn't support ZTP, only if Cumulus is installed

xCAT/postscripts/onieztp
cxhong: oh, right,  Thanks.  I will make changes

xCAT-server/lib/xcat/plugins/dhcp.pm
cxhong: yeh, good

immarvin:
when to run ``sudo ztp -R``?  if we have to run this command manually, the auto discovery and auto configuration work flow will be broken

immarvin:
I have an idea on auto discovery and configuration of cumulus switch,  the whole workflow is trigged by the 1st dhcp interaction on switch power on.  with the dhcpd-eval: http://jpmens.net/2011/07/06/execute-a-script-when-isc-dhcp-hands-out-a-new-lease/(proposed by @mattaezell ), a script will be invoked when a lease is granted to the switch(we can identify the dhcp request is from vendor info in mac address and vendor-class-identifier in dhcp request), this script will check the existence of the predefined switch definition according to the mac, creating the switch definition if not , and change the switch status to "discovered", then enable the ztp on switch with "ssh -lxxx <switch> ztp -R"(the default user and password)  to initiate the ztp process and invoke the ztp script in this PR to finish the switch configuration. After switch configuration finished, update the switch status to "configured".

any comments? @cxhong @mattaezell

mattaezell:
@immarvin If we must login to the switch to make it run ZTP, might as well just configure it.  If the switch came in ONIE mode and ran the Cumulus installer, first boot would automatically initiate ZTP.  LBS requested that the CORAL switches come pre-installed with Cumulus and a license, but I have an open question to LBS to see if the switches can come unconfigured.

Either way, I think we will need to support *dhcpd-eval* for PDUs and Mellanox switches.

cxhong:
This PR will cover how xCAT handles ZTP request from cumulus switch.
how to trigger the ZTP process(manually, or send ssh command in the autodiscover, or  non pre-installed switch) should be discussed in the issue #3060

immarvin:
hi @mattaezell ,even LBS agrees to ship the switch un-configured, I think we still need to support ZTP re-enable, suppose a scenario that a switch, which was already discovered and configured sometime before, is powered off or plugged-out, then power on or plugin in, it should still be discovered(and reconfigured?).

I noticed that dhcpd-eval should only perform some asynchronous
 tasks which does not take too much time and have no possibility to hang, otherwise, the performance of dhcpd might decrease in a large cluster, in the worst case, the dhcpd will stop response dhcp requests if the execution of dhcpd-eval script hang. So I think dhcpd-eval can submit a job or send a message to tell xcatd to so something like "ssh xxx ztp -R" or create/update switch definition,but does not perform the task by itself or wait for the task result.

what do you think?

immarvin:
hi @mattaezell , just as @cxhong suggest, how about keep on discussion in https://github.com/xcat2/xcat-core/issues/3060? thanks

immarvin:
agree to merge this first to keep on working based on this pr

PR #3105, opened 2017-05-19
Sync up .rst files

gurevichmark:
Checking in out of sync `.rst` files. Most likely `Schema.pm` file was changed, but `.rst` files were not checked in.


whowutwut:
Thanks @gurevichmark

PR #3107, opened 2017-05-22
Fix bug: xcattest failed to handle multiple cases with the same case name

hu-weihua:
Depending on test, new version ``xcattest`` has a bug, it can't handle one scenario which multiple cases have the same case name. Fix this problem. now just load the last valid case.

Multiple cases with the same case name is a history problem. Now we have a rule which is to use one case to cover multiple platform, not use multiple cases with the same case name to do it.

### DIFF
--- a/xCAT-test/xcattest
+++ b/xCAT-test/xcattest
@@ -819,6 +819,7 @@ sub load_case {
     my $fd          = undef;
 
     my %invalidcases;
+    my %case_name_index_map_bak;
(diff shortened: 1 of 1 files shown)


bybai:
Our goal is to use one case to cover multiple platform, not use multiple cases with the same case name to do it. I am ok to merge to avoid case load failure in multiple case with the same name scenario. Let autotest env to test it first.

PR #3110, opened 2017-05-22
Check that there are failed nodes in the array before printing message 

whowutwut:
When running some tests using `rinstall`, get back error messages with blank nodes. 

```
#  rinstall sn01 'osimage=rhels7.3-ppc64le-MLNX-install-service'
Provision node(s): sn01
sn01: install rhels7.3-ppc64le-service
Unable to dispatch hierarchical sub-command to sn03:3001.  Error: Connection failure: IO::Socket::INET: connect: No route to host at /opt/xcat/lib/perl/xCAT/Client.pm line 248.
Error: failed to run 'nodeset' against the following nodes:
#
```

This pull request just checks that there are failed nodes in the array before printing.  It does not test the validity of the checks..

### DIFF
--- a/xCAT-server/lib/xcat/plugins/rinstall.pm
+++ b/xCAT-server/lib/xcat/plugins/rinstall.pm
@@ -364,9 +364,11 @@ sub rinstall {
             }
         }
         my $rsp = {};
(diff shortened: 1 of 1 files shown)


immarvin:
i am ok with the change

Issue #3100, opened 2017-05-19
[DEV] Unable print useful msg for 'curl is not installed'  when run bmcdiscover on Ubuntu


xuweibj:
OS: Unbuntu 16.04.1
Arch: PPC64LE

When run ``bmcdiscover -s nmap --range x.x.x.x -u root -p 0penBmc``, return nothing.

More detail:
```
# XCATBYPASS=1 bmcdiscover -s nmap --range x.x.x.x -u root -p 0penBmc
sh: 1: curl: not found
```

xuweibj:
The issue was fixed in pull request #3122

xuweibj:
Verified on Ubuntu 16.04
```
# bmcdiscover -s nmap --range 10.200.0.1 -u root -p 0penBmc
10.200.0.1,SIMULATER,Y130UF72701M,root,0penBmc,mp,bmc
```

PR #3116, opened 2017-05-23
Use LWP to send https request in bmcdiscover instead of curl command

xuweibj:
#3100 
Use ``LWP`` to send login and 'GET' request to avoid use ``curl`` command in bmcdiscover.pm

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -29,6 +29,9 @@ use Data::Dumper;
 use File::Basename;
 use File::Path;
 use Cwd;
+use LWP;
+use HTTP::Cookies;
+use HTTP::Response;
 use JSON;
 
 my $nmap_path;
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -1040,14 +1043,20 @@ sub bmcdiscovery_openbmc{
...
+    my $response = $brower->request($request);
     
-    if ($output =~ /\"status\": \"ok\"/) {
-        my $req_output = `curl -b $cjar_file -k https://$ip/xyz/openbmc_project/inventory/system/chassis/motherboard`;
-        my $response = decode_json $req_output;
+    if ($response->is_success) {
+        my $req_url = "https://$ip/xyz/openbmc_project/inventory/system/chassis/motherboard";
+        my $get_req = HTTP::Request->new('GET', $req_url, $header);
+        my $req_output = $brower->request($get_req);
+        my $response = decode_json $req_output->content;
```
chenglch: Why don't you check the request result like line 1055?

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -1080,10 +1089,8 @@ sub bmcdiscovery_openbmc{
             $node =~ s/(.*)/\L$1/g;
             $node =~ s/[\s:\._]/-/g;
         }
-
-        unlink $cjar_file;
     } else {
-        if ($output =~ /\"description\": \"Invalid username or password\"/) {
+        if ($response->content =~ /\"description\": \"Invalid username or password\"/) {
             xCAT::MsgUtils->message("W", { data => ["Invalid username or password for $ip"] }, $::CALLBACK); 
         }
```
chenglch: As you use the library now, if there is unknow failure( maybe 404, 503), you should keep some log or send some message back to the user, return immediately may not be a good choice.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -1040,14 +1043,20 @@ sub bmcdiscovery_openbmc{
...
+    my $request = HTTP::Request->new( 'POST', $url, $header, $data );
+    my $response = $brower->request($request);
     
-    if ($output =~ /\"status\": \"ok\"/) {
-        my $req_output = `curl -b $cjar_file -k https://$ip/xyz/openbmc_project/inventory/system/chassis/motherboard`;
-        my $response = decode_json $req_output;
+    if ($response->is_success) {
+        my $req_url = "https://$ip/xyz/openbmc_project/inventory/system/chassis/motherboard";
+        my $get_req = HTTP::Request->new('GET', $req_url, $header);
+        my $req_output = $brower->request($get_req);
```
chenglch: Suggest considering a proper name for these variables, now there are $request, $get_req, $req_output and $response.

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
xuweibj: If it returns other error code, we did not consider it as a bmc, so will not print message.

PR #3117, opened 2017-05-23
modify migration test for environment's servicenode need to clear

junxiawang:
Modify migration testcases for testcase sometimes fail due to servicenode's dns service influence .
1. power off service node
2.
Before modify we only "power off" servicenode but "makdhcp " always wrong for servicenode off .So I delete the servicenode's definition "groups=service"


hu-weihua:

PR #3118, opened 2017-05-23
Fix #3087: nodeset improvement - using multi-process to parallelize

### LINKED ISSUE #3087: [xCAT performance] nodeset improvement - try to using multi-process to parallelize
Using multi-process to parallelize the nodeset to reduce the time consuming,  and the method could be used by other plugins.

robin2008:
Fix #3087: nodeset improvement - using multi-process to parallelize 
- Before deliver the request to xcatd, in preprocess method, we can split the request to multiple sub-requests which contains different noderange
 - In that, xcatd could handle those sub-requests in parallel, just like issuing multiple commands in the same time.

Leave below tasks to next sprint:
 - Make the parallel requests parameters configurable
 - Using callback method to filter the noderange first to make the function could be leveraged by different purpose plugins (the plugin could have their own filter callback functions)

### DIFF
--- a/perl-xCAT/xCAT/Scope.pm
+++ b/perl-xCAT/xCAT/Scope.pm
@@ -1,7 +1,165 @@
 package xCAT::Scope;
+use POSIX;
 use xCAT::Utils;
 use xCAT::Table;
 use xCAT::ServiceNodeUtils qw(getSNList);
+
+
+#-----------------------------------------------------------------------------
+
+=head3 split_node_array
+
+    Split a node array into multiple subsets in case to handle them in parallel.
+
+    Arguments:
+       Reference of source array
+       Maximum subset number
+       Default element capacity in each subset
+    Returns: An array of all subsets
+    Error:
+        none
+    Example:
+        my $subsets = split_node_array(\@nodes, 5, 250);
+=cut
+
+#-----------------------------------------------------------------------------
+sub split_node_array {
+    my $source = shift;
+    if ($source =~ /xCAT::Scope/) {
+        $source = shift;
+    }
+    my $max_sub = shift;
+    my $capacity = shift;
+
+    if ($max_sub < 2) {return [$source];}
+ 
+    my @dest = {};
+    my $total = $#{$source} + 1;
+    my $n_sub = POSIX::ceil($total / $capacity);
+    if ( $n_sub <= 1 ) {
+        $dest[0] = $source;
+
+    } elsif ( $n_sub > $max_sub ) {
+
+        $capacity = int ($total / $max_sub);
+        if ( $total % $max_sub > 0 ) {
+            $capacity += 1;
+        }
+        my $start = $end = 0;
+        for (1..$max_sub) {
+            $end = $start + $capacity - 1;
+            if ( $end > $total -1 ) {
+                $end = $total -1
+            }
+
+            my @temp = @$source[$start..$end];
+            $dest[$_-1]=\@temp;
+            $start = $end + 1;
+        }
+
+    } else {
+        # Only require $n_sub subset.
+        my $start = $end = 0;
+        for (1..$n_sub) {
+            $end = $start + $capacity - 1;
+            if ( $end > $total -1 ) {
+                $end = $total -1
+            }
+            #print "subset #$_: $start to $end";
+            my @temp = @$source[$start..$end];
+            $dest[$_-1]=\@temp;
+            $start = $end + 1;
+        }
+    }
+
+    return \@dest;
+}
+
+#-----------------------------------------------------------------------------
+
+=head3 get_parallel_scope
+
+    Convert a request object to an array of multiple requests according to the 
+    splitted node range.
+
+    Arguments:
+       Reference of request
+       Maximum subset number: Optional, default is 5
+       Default element capacity in each subset: Optional, default is 250
+    Returns: An array of requests
+    Error:
+        none
+    Example:
+        my $reqs = xCAT::Scope->get_parallel_scope($request);
+=cut
+
+#-----------------------------------------------------------------------------
+sub get_parallel_scope {
+    my $req = shift;
+    if ($req =~ /xCAT::Scope/) {
+        $req = shift;
+    }
+    my ($max_sub, $capacity) = @_;
+    #TODO, make the value configurable
+    unless ($max_sub) { $max_sub = 5; }
+    unless ($capacity) { $capacity = 250; }
+
+    my $subsets = split_node_array(\@{ $req->{node} }, $max_sub, $capacity);
+    # Just return the origin one if node range is not big enough.
+    if ($#{$subsets} < 1) { return [$req]; }
+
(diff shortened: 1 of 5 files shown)


robin2008:
Some test result on an physical power8 server:

1, using max 5 processes for parallel , and each have at least 250 nodes
```
5000; nodeset; 1; 2; SUCESS; "nodeset fake[1-1] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 50; 6; SUCESS; "nodeset fake[1-50] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 100; 11; SUCESS; "nodeset fake[1-100] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 250; 26; SUCESS; "nodeset fake[1-250] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 500; 43; SUCESS; "nodeset fake[1-500] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 1000; 81; SUCESS; "nodeset fake[1-1000] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 2500; 217; SUCESS; "nodeset fake[1-2500] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 5000; 545; SUCESS; "nodeset fake[1-5000] osimage=rhels7.3-GA-ppc64le-install-compute"
```

2, 10 processes for parallel, and each have at least 250
```
5000; nodeset; 1; 1; SUCESS; "nodeset fake[1-1] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 50; 7; SUCESS; "nodeset fake[1-50] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 100; 13; SUCESS; "nodeset fake[1-100] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 250; 30; SUCESS; "nodeset fake[1-250] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 500; 47; SUCESS; "nodeset fake[1-500] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 1000; 77; SUCESS; "nodeset fake[1-1000] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 2500; 198; SUCESS; "nodeset fake[1-2500] osimage=rhels7.3-GA-ppc64le-install-compute"
5000; nodeset; 5000; 460; SUCESS; "nodeset fake[1-5000] osimage=rhels7.3-GA-ppc64le-install-compute"
```

perl-xCAT/xCAT/Scope.pm
```diff
@@ -1,8 +1,166 @@
...
+#-----------------------------------------------------------------------------
+
+=head3 get_broadcast_scope_with_parallel
+
+    Convert a request object to an array of multiple requests according to the 
+    splitted node range.
+
+    Arguments:
+       Reference of request
+       Callback: Optional, default is 5
```
immarvin: ``Callback: Optional, default is 5``? typo?

perl-xCAT/xCAT/Scope.pm
```diff
@@ -1,8 +1,166 @@
...
+
+#-----------------------------------------------------------------------------
+sub get_parallel_scope {
+    my $req = shift;
+    if ($req =~ /xCAT::Scope/) {
+        $req = shift;
+    }
+    my ($max_sub, $capacity) = @_;
+    #TODO, make the value configurable
+    unless ($max_sub) { $max_sub = 5; }
```
immarvin: is it better to make "5" and "250" configurable in site table? or environment variables for ``nodeset``  ?

perl-xCAT/xCAT/Scope.pm
```diff
@@ -1,8 +1,166 @@
...
+        $source = shift;
+    }
+    my $max_sub = shift;
+    my $capacity = shift;
+
+    if ($max_sub < 2) {return [$source];}
+ 
+    my @dest = {};
+    my $total = $#{$source} + 1;
+    my $n_sub = POSIX::ceil($total / $capacity);
```
immarvin: is it possible to avoid using ``ceil``? it is used once but introduces a new dependency module

perl-xCAT/xCAT/Scope.pm
```diff
@@ -1,8 +1,166 @@
...
+        }
+        my $start = $end = 0;
+        for (1..$max_sub) {
+            $end = $start + $capacity - 1;
+            if ( $end > $total -1 ) {
+                $end = $total -1
+            }
+
+            my @temp = @$source[$start..$end];
+            $dest[$_-1]=\@temp;
```
immarvin: ``@dest `` is an array of references to an array, is this compatible with the structure of hash  $req in the calling subroutine?

perl-xCAT/xCAT/Scope.pm
robin2008: I'll consider it.

perl-xCAT/xCAT/Scope.pm
robin2008: In $req,  the 'node' element is an array. 
And yes, I think in the calling method it requires to make the conversion.

This method is general purpose,  it covert an array to a new array whose element is a reference of a subset of the source array.

perl-xCAT/xCAT/Scope.pm
robin2008: Yes,  it is in TODO.  Plan to add it into site table. Maybe in environment is better,   As this method is also designed for other plugins which want to parallelize its handling, not only for nodeset.

perl-xCAT/xCAT/Scope.pm
robin2008: It's a typo,  forget to make the changes accordingly.

perl-xCAT/xCAT/Scope.pm
```diff
@@ -33,25 +33,26 @@ sub split_node_array {
...
     my $n_sub = int ($total / $capacity);
     unless ($n_sub * $capacity == $total) { $n_sub++;} #POSIX::ceil
 
     if ( $n_sub <= 1 ) {
+        # Only 1 subset is enough
         $dest[0] = $source;
 
     } elsif ( $n_sub > $max_sub ) {
-
+        # Caculate how many subsets is required.
```
daniceexi: Could you add more comments here?  I cannot understand the logic here.

daniceexi: Could you add more comments here?  I cannot understand the logic here.

robin2008: You are right,  I put the comment in wrong else clause. This "elsif" is for noderange exceed the max, then we have to recalculate the capacity for each subset as we only allow the max_sub subsets.

immarvin:
i am ok with the change

Issue #3046, opened 2017-05-12
getadapter does not work in service node configuration (bare metal ppc64le) 


whowutwut:
*Note* If it will help to quicky get this working 100% of the time, we should reduce the support scope limited to certain types of ARCH/OS/Model... we can get down to the model level if that is even a concern to get this working quickly.  (i.e Supported only for ppc64le rhels 7.3 only on 8335-GTB models) 

Trying to run the `getadapter` code, it does not work.   Not sure why.  Or where the blocking is coming from, but opening this issue to track it.  Ran this against 8335-GTB node, the following is seen....

```
[root@fs3 ~]# date; getadapter sn01; date
Thu May 11 17:40:13 EDT 2017
Thu May 11 20:01:52 EDT 2017
[root@fs3 ~]#
```

Essentially, it takes 2+ hours to return control back to the command, and no output messages are seen.  Looking in in the command logs...

```
====================================================
[Date]       2017-05-11 17:40:13
[ClientType] cli
[Request]    getadapter sn01
[Response]
[ElapsedTime] 8499 s
====================================================
```

So really there's no good way to debug this... can we implement a -V option or allow it to print some more messages so that we can track some sort of progress.   

@chenglch  has done some investigation , so will allow him to fill in.

gurevichmark:
Referencing issue #1174 that I opened originally for this problem.
1. I have not been able to run `getadapter` against a VM.
2. `getmacs` command should also be tried as it might have the same problem.

whowutwut:
@gurevichmark Thanks, I updated the titles for both issues to attempt to reduce the scope the work to support it...

chenglch:
@whowutwut , no need to add the verbose message in getadapter. The root cause here is `rinstall`. `rinstall` will hangs for a long time(maybe forever) in the `runxcmd` calls. (only in `runxcmd` calls) 
https://github.com/xcat2/xcat-core/blob/master/xCAT-server/lib/xcat/plugins/getadapter.pm#L307

```
 my $outref = xCAT::Utils->runxcmd(
            {
                command => ['rinstall'],    # --> nodeset 
                node    => ["$node"],
                arg     => ['runcmd=getadapter'],
            },
            $subreq, 0, 1);
```
If I change the command with `nodeset`, it will return immediately. 
```
[root@fs3 xcat]# getadapter sn02
sn02: runcmd=getadapter
[...]

PR #3121, opened 2017-05-24
Avoid of circular reference in a loop of rinstall

chenglch:
This is a temporary fix for getadapter which will hang forever if
error happens in `rinstall stage`. The root cause is the circular
reference in `runxcmd`. As global variable `$::xcmd_outref` and
`%::xcmd_outref_hash` are used to save the output message, if the
plugin command are called with runxcmd and the plugin called also
calls another plugin with `runxcmd`, the circular reference will happen.
In this situation, if the output reponse is used in a iterator,
like:
```
   foreach my $line (@$res) {
      xCAT::MsgUtils->message("I", $rsp, $callback);
```
`$res` points to the global variable, and `MsgUtils->message` in the
runxcmd calls also append the value in the global variable, as the
iter object is changed, the loop will not be terminated.

This patch can not solve the circular reference problem, it is just
a workaround to avoid infinite loop.

Some known issues: the error message will print twice as the
`MsgUtils->message` called in runxcmd. But we can not simply remove
it as the plugin may not be called with `runxcmd`.

Error message example:

```
Error: frame10node10: mgt configuration can not be found.
frame10node10: Unable to identify plugin for this command, check relevant tables: noderes.netboot
frame10node10: Unable to identify plugin for this command, check relevant tables: noderes.netboot
Failed to run 'nodeset' against the following nodes: frame10node10
Error: failed to run command: rinstall frame10node10 runcmd=getadapter
```
partial-fix: #3046

### DIFF
--- a/xCAT-server/lib/xcat/plugins/rinstall.pm
+++ b/xCAT-server/lib/xcat/plugins/rinstall.pm
@@ -340,7 +340,9 @@ sub rinstall {
         # We got an error with the nodeset
         my @successnodes;
         my @failurenodes;
-        foreach my $line (@$res) {
+        # copy into a temporary variable to avoid of circular reference
+        my @lines = @$res;
+        foreach my $line (@lines) {
             $rsp->{data}->[0] = $line;
             if (($line =~ /: install/) or ($line =~ /: netboot/)) {
                 my $successnode;
@@ -403,7 +405,6 @@ sub rinstall {
                 },
                 $subreq, -1, 1);
 
-
             $rc = $::RUNCMD_RC;
             my $rsp = {};
             if ($VERBOSE) {
@@ -416,7 +417,9 @@ sub rinstall {
 
                 # We got an error with the rnetboot
                 my @failurenodes;
-                foreach my $line (@$res) {
+                # copy into a temporary variable to avoid of circular reference
+                my @lines = @$res;
+                foreach my $line (@lines) {
                     $rsp->{data}->[0] = $line;
                     if ($line =~ /: Success/) {
                         my $successnode;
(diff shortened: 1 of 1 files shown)


immarvin:
so the nested runxcmds are in the same process?

chenglch:
yes, they are in the same process

immarvin:
hi @chenglch , I have some question on the infinite loop you mention,
in perl-xCAT/xCAT/Utils.pm:
````
1178 sub runxcmd
1179
1180 {
...
1211     $::xcmd_outref      = [];
1212     %::xcmd_outref_hash = ();
````

seems the each time the global variables will be reinitialized in each runxcmd() call, since the nested runxcmd() calls are in same process,  when run to 
````
 foreach my $line (@$res) {
      xCAT::MsgUtils->message("I", $rsp, $callback);
````
, there should not be a nested runxcmd() is pushing sth to  %::xcmd_outref_hash or @$::xcmd_outref , so how does the infinite loop occur?

chenglch:
hi @immarvin , if we call the `runxcmd` in a general plugin, `xCAT::MsgUtils->message` in the `runxcmd` will send the message to `$::xcmd_outref` and the `xCAT::MsgUtils->message` out of `runxcmd` will send the message to client. It works well.

Now in the nest scenario,  I divided it into three level: 
`wrap`: out of `runxcmd` ---- getadapter
`outer`: the first level of `runxcmd`  -- rinstall
`inner`: the second level of `runxcmd` -- nodeset

At first, `wrap` call the `runxcmd` , it initialize the `$::xcmd_outref`,  currently the `runxcmd` do not return, and then call the `runxcmd` again, `$::xcmd_outref` is initialized again, Now only **one global variable `$::xcmd_outref`**( initialize at the second time), the global variable initialized at first time is lost. 
The infinite loop is in the `outer` level,  the output message from `inner` level is filled in the `$::xcmd_outref` . Now in the `for loop`, we get the line from `$::xcmd_outref`, and use  `xCAT::MsgUtils->message`  send the line back to the `$::xcmd_outref`**(only one global variable, explained above)** again.

immarvin:
agree to merge

PR #3127, opened 2017-05-24
Fix automation bug: xcattest failed to load case depending on arch label

hu-weihua:
Depending on the result of daily regression these day,  found xcattest failed to load case depending on arch label. Fix this bug.


bybai:
Our test cases contain arch or hcp to distinguish test cases with same test case name, we should pick up these old code into xcattest. I am OK to merge it to let autotest to verify it.

Issue #3130, opened 2017-05-24
Wrong mac address showing for switchdiscover


cganglin:
When doing the following:
````
switchdiscover --range 10.80.1.39 -s snmp
Discovering switches using snmpwalk for 10.80.1.39 .....
ip              name                    vendor                                                  mac       
------------    ------------            ------------                                            ------------
10.80.1.39      ms1                     Cumulus Linux 3.0.1 (Linux Kernel 4.1.19-1+cl3u3)       70:e2:84:14:08:37
Switch discovered: ms1
````

The mac address returned is NOT the mac address I see when I log on to the physical switch. 
````
br0       Link encap:Ethernet  HWaddr cc:37:ab:b2:36:c1
          inet6 addr: fe80::ce37:abff:feb2:36c1/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:26491742 errors:0 dropped:149 overruns:0 frame:0
          TX packets:95215091 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:26143096274 (24.3 GiB)  TX bytes:35036594330 (32.6 GiB)

eth0      Link encap:Ethernet  HWaddr cc:37:ab:b2:36:c0
          inet addr:10.80.1.39  Bcast:0.0.0.0  Mask:255.192.0.0
          inet6 addr: fe80::ce37:abff:feb2:36c0/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:3584980 errors:0 dropped:2 overruns:0 frame:0
          TX packets:40712 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:314377568 (299.8 MiB)  TX bytes:6733764 (6.4 MiB)
          Interrupt:12 Base address:0x6000
````
Please check into this issue.

cxhong:
Clive verified on his test system.  I will close this issue

PR #3131, opened 2017-05-24
Wrong mac address showing for switchdiscover command

cxhong:
for issue #3130 , the switchdiscover command called get_mac_by_arp() routine to get mac address for specified ip.   In that routine, used !~ to compare two ip address, which caused issue. 
Verified with Clive,   changed !~ to ne, the command return correct mac address.

this PR also added two global identity to define cumulus switch for switchdiscover command.


immarvin:
agree to merge

PR #3133, opened 2017-05-24
Enhance MTM detection from FRU pages

kcgthb:
For systems that don't provide "Product Name", try to use "Product Part Number" instead. 

Supermicro does this, for instance:

    # ipmitool fru print 0
     Chassis Type          : Other
     Chassis Part Number   : CSE-118GQETS-R2K03P
     Chassis Serial        : [...]
     Board Mfg Date        : Sun Dec 31 16:00:00 1995
     Board Mfg             : Supermicro
     Board Serial          : [...]
     Board Part Number     : X10DGQ
     Product Manufacturer  : Supermicro
     Product Part Number   : SYS-1028GQ-TR
     Product Serial        : [...]

This will result in the following MTM information:
```
# lsdef sh-112-12 -i mtm
Object name: sh-112-12
    mtm=Supermicro:SYS-1028GQ-TR
```

### DIFF
--- a/xCAT-genesis-scripts/bin/dodiscovery
+++ b/xCAT-genesis-scripts/bin/dodiscovery
@@ -75,13 +75,17 @@ if [ -r /sys/devices/virtual/dmi/id/product_name ]; then  #x86
 		if [ -z "$MTM" ]; then
 			FRU=`ipmitool fru print 0`
 			if [ $? -eq 0 ]; then
-				MTM=`echo "$FRU" | awk -F': ' '/Product Manufacturer/ {m=$2} /Product Name/ {n=$2} END {print m":"n}'`
+				m=`echo "$FRU" | awk -F': ' '/Product Manufacturer/ {print $2}'`
+				n=`echo "$FRU" | awk -F': ' '/Product Name/ {print $2}'`
+				if [ -z "$n" ]; then
+					n=`echo "$FRU" | awk -F': ' '/Product Part Number/ {print $2}'`
+				fi
 			else
 				logger -s -t $log_label -p local4.warning "Couldn't find MTM information in FRU, falling back to DMI (MTMS-based discovery may fail)"
 				m=`cat /sys/devices/virtual/dmi/id/sys_vendor`
 				n=`cat /sys/devices/virtual/dmi/id/product_name`
-				MTM="$m:$n"
 			fi
+			MTM="$m:$n"
 		fi 
 		SERIAL=`cat /sys/devices/virtual/dmi/id/product_serial`
 	fi
--- a/xCAT-genesis-scripts/bin/dodiscovery
+++ b/xCAT-genesis-scripts/bin/dodiscovery
@@ -77,7 +77,7 @@ if [ -r /sys/devices/virtual/dmi/id/product_name ]; then  #x86
 			if [ $? -eq 0 ]; then
 				m=`echo "$FRU" | awk -F': ' '/Product Manufacturer/ {print $2}'`
 				n=`echo "$FRU" | awk -F': ' '/Product Name/ {print $2}'`
-				if [ -z "$n" ]; then
+				if [ -z "$n" -o "$n" == "NONE" ]; then
 					n=`echo "$FRU" | awk -F': ' '/Product Part Number/ {print $2}'`
 				fi
 			else


zet809:
hi, @kcgthb , the code seems good to me, but in order to have hardware discovery worked, you shall also modify the bmcdiscover.pm to take the similar logic to get mtm. Otherwise, the host can not match the temp BMC node.

kcgthb:
Hi @zet809 

Oh right, I forgot to update the `bmcdiscover.pm` part, sorry about that.

I just started to look at it and realized there's some discrepancy between the two that I didn't notice before: `bmcdiscover.pm` uses `Chassis Part Number` and `Chassis Serial` as a first choice if they're present in the FRU page, but `dodiscovery` completely ignores that and focuses on `Product *` strings.

So for any server that presents `Chassis` information in the FRU page, `bmcdiscover` will pick that up, and `dodiscovery` will report `Product` entries, so there couldn't be any match.

There is also so MTM parsing code in [`SPD.pm`](https://github.com/xcat2/xcat-core/blob/master/perl-xCAT/xCAT/SPD.pm) that is use, among other places I assume, by `rinv`. So it looks like there's at least 3 places in the code where xCAT tries to get server model and serial number information. And it's done quite differently each time:
* `dodiscovery`
* `bmcdiscover.pm`
* `SPD.pm`

Do you think there would be some way to centralize or re-use some code there? Just so it stays coherent and easier to maintain.

zet809:
Hi, @kcgthb , currently for the logic in bmcdiscover.pm, it need to support Both x86_64 and OpenPOWER servers and they have different fru output. 
I agree with you that we shall keep the logic to retrieve MTM/Serial consistent for bmcdiscover.pm and dodiscovery, will consistent it in next release.

kcgthb:
Hi @zet809 

I see. Is there any list of typical FRU outputs for all the servers types xCAT should support? If not, maybe it would be interesting to start building such a list, and collect example FRU pages from those machine types. That way, the MTM code could be tested against them to make sure all server types are correctly detected and the information correctly parsed.

It could be even better to collect both FRU pages and information from `/sys/devices/virtual/dmi/id` as this is also used as a fallback in `dodiscovery`.

What do you think?

I can provide FRU pages and DMI information for a set of Dell and Supermicro servers, probably other types too.

zet809:
Hi, @kcgthb , we don't have the list so it is a good idea to have that.

kcgthb:
Well, if you ever start one, please let me know, and I'll contribute information about the few different system types I have access to.

whowutwut:
@zet809 Can you do a review and select request changes with the requirement for `bmcdiscover.pm` , this way we know this should not be merged yet.  Also 2.13.5 is coming to an end this week so I will like to move this out this milestone.

zet809:
As @kcgthb mentioned above, need to have the list of FRU outputs for the server that xCAT supported and then modified based on that list.

xCAT-genesis-scripts/bin/dodiscovery
```diff
@@ -75,13 +75,17 @@ if [ -r /sys/devices/virtual/dmi/id/product_name ]; then  #x86
 		if [ -z "$MTM" ]; then
 			FRU=`ipmitool fru print 0`
 			if [ $? -eq 0 ]; then
-				MTM=`echo "$FRU" | awk -F': ' '/Product Manufacturer/ {m=$2} /Product Name/ {n=$2} END {print m":"n}'`
+				m=`echo "$FRU" | awk -F': ' '/Product Manufacturer/ {print $2}'`
+				n=`echo "$FRU" | awk -F': ' '/Product Name/ {print $2}'`
+				if [ -z "$n" -o "$n" == "NONE" ]; then
+					n=`echo "$FRU" | awk -F': ' '/Product Part Number/ {print $2}'`
```
samveen: You can easily collapse this back into one `awk` request:

```
MTM=`echo "$FRU" | awk -F': ' '/Product Manufacturer/ {m=$2} /Product Name|Product Part Number/ {if (n=="") {n=$2}} END {print m":"n}'`
```
This works because that `Product Name` will _always_ be printed before 'Product Part Number` in `ipmitool fru...` output.

kcgthb: True! Although I think that keeping "manufacturer" and "name" as two separate variables helps keep the code more readable, and is consistent with the DMI fall-back, which requires reading 2 files anyway.

I don't think the extra awk invocation adds a lot of overhehad to the whole `dodiscovery` script, do you?

samveen: Not much overhead, but a slightly cleaner git diff :D

whowutwut:
Let's set the milestone when we merge it into the xCAT code.

samveen:
We hit the bug this MR solves today. Any timeline for merging this?

whowutwut:
@zet809  you had requested changes, perhaps it was not too clear.  It seems like you mean that we should apple the same changes to `bmcdiscover` code.. .  Maybe we can merge this first, then someone can create a PR to make the similar changes for bmcdiscover?

zet809:
Oh, yes, so, I will merge this PR and will you pls modify the bmcdiscover.pm to make hardware discovery work with this modification? Thx!

PR #3134, opened 2017-05-25
Add support for the rvitals option for openbmc based on the defined API

whowutwut:
_Based on the OpenBMC Sensor API, complete the development code
for the supported options based on the possible values that are
currently allowed to be returned.  As more data is added, the code
should be able to handle the printing of the data automatically_

Adds support for `rvitals`, as part of issue #2816 

Looking at the [Value.yaml](https://github.com/openbmc/phosphor-dbus-interfaces/blob/fca9cc6df49f623126858ad7ec0a13ffae17c3f6/xyz/openbmc_project/Sensor/Value.interface.yaml) file as part of the openbmc/sensor API, complete the support for the options in `rvitals`

### External Interfaces supported based on the possible valid values: 

```
rvitals noderange [temp|voltage|wattage|fanspeed|power|length|all]
```

### Example output:

No options specified
```
[root@fs2vm110 ~]# rvitals p9euh02 
p9euh02: fan_tach/fan1: 7470 RPMS
p9euh02: temperature/pcie: 29.75 C
p9euh02: fan_tach/fan2: 7560 RPMS
p9euh02: fan_tach/fan0: 7440 RPMS
p9euh02: fan_tach/fan3: 7530 RPMS
```

Getting `temp` only
```
[root@fs2vm110 ~]# rvitals p9euh02 temp
p9euh02: temperature/pcie: 29.75 C
```

Getting `fanspeed` only:
```
[root@fs2vm110 ~]# rvitals p9euh02 fanspeed
p9euh02: fan_tach/fan1: 7440 RPMS
p9euh02: fan_tach/fan2: 7530 RPMS
p9euh02: fan_tach/fan0: 7500 RPMS
p9euh02: fan_tach/fan3: 7500 RPMS
```

PR #3135, opened 2017-05-25
Expose the rvitals command as part of openbmc support [DO NOT MERGE]

whowutwut:
Adds support for rvitals, as part of issue #2816

After PR #3134 is merged, I believe that completes the development effort for rvitals and I think we could merge this and open up the function.


whowutwut:
Without zenhub, it looks like we can't see this dependency.. perhaps the do not merge is more clear in the title.  Waiting for dependent pull requests to be merged before this.

Issue #3137, opened 2017-05-25
[ xCAT 2.13.2 Centos 7.3 ] Error: Unable to identify plugin for this command, check relevant tables: nodehm.power,mgt;nodehm.mgt;pduoutlet.pdu=.* 


dspatil7768:
Hello, I'm trying for Switch based discovery. I'm using `Dlink` switch:`DGS-1510-52'`. But while `rpower` I'm facing the issue:

- Following this document:
[http://xcat-docs.readthedocs.io/en/stable/guides/admin-guides/manage_clusters/ppc64le/discovery/switch_discovery.html#switch-based-discovery](url)
```
[root@xcatmgt ~]# rpower node-..................-..................
node-..................-..................: Error: Unable to identify plugin for this command, check relevant tables: nodehm.power,mgt;nodehm.mgt;pduoutlet.pdu=.*
```
- And while `bmcdiscovery` it not showing serial no.

`[root@xcatmgt ~]# bmcdiscover --range 172.40.1.5 -z -w`
`node-..................-..................:`
 `       objtype=node`
`        groups=all`
 `       bmc=172.40.1.5`
`        cons=ipmi`
 `       mgt=ipmi`
`        mtm=..................`
 `       serial=..................`

`[root@xcatmgt ~]# lsdef /node-.*`
`Object name: node-..................-..................`
 `   bmc=172.40.1.5`
`    cons=ipmi`
 `   groups=all`
`    hwtype=bmc`
 `   mgt=ipmi`
`    mtm=..................`
`    nodetype=mp`
 `   postbootscripts=otherpkgs`
`    postscripts=syslog,remoteshell,syncfiles`
`    power=ipmi`
 `   serial=..................`

- Below is some information about my configuration:

`[root@xcatmgt ~]# lsdef cn1`
`Object name: cn1`
 `   bmc=172.40.1.5`
  `  chain=runcmd=bmcsetup,osimage=centos7.3-x86_64-install-compute`
   ` cons=ipmi`
    `groups=all`
    `installnic=A4:BF:01:26:AD:DC`
   ` ip=172.10.1.5`
    `mgt=ipmi`
    `netboot=pxe`
    `postbootscripts=otherpkgs`
    `postscripts=syslog,remoteshell,syncfiles`
    `power=ipmi`
    `primarynic=A4:BF:01:26:AD:DC`
    `switch=dlink`
    `switchport=3`

`[root@xcatmgt ~]# lsdef dlink`
`Object name: dlink`
[...]

PR #3139, opened 2017-05-25
Fix issue 3137: bmcdiscover can not get mtms if fru only had board information

zet809:
For issue #3137


whowutwut:
@zet809  Can you put a summary here in addition to pointing to the issue.   For example, where are we hitting this problem and some sample externals on how the fix  results in attributes filled in.

zet809:
Close this PR since the issue #3137 was closed by creator.

Issue #2996, opened 2017-05-07
RHEL update caused boot failure due to selinux


mattaezell:
I updated our RHEL 7.3 repositories to pick up recent security updates.  The node failed to boot with the following message:

```
[Fri May  5 16:25:38 2017][   31.941132] systemd-journald[906]: Received SIGTERM from PID 1 (systemd).
[Fri May  5 16:25:38 2017][   31.955732] type=1404 audit(1494015917.470:2): enforcing=1 old_enforcing=0 auid=4294967295 ses=4294967295
[Fri May  5 16:25:38 2017][   32.001858] systemd[1]: Failed to load SELinux policy.
[Fri May  5 16:25:38 2017][!!!!!!] Failed to load SELinux policy, freezing.
```

I tracked down what changed, and apparently sssd-common now requires selinux-policy

```
* Fri Feb 03 2017 Jakub Hrozek <jhrozek@redhat.com> - 1.14.0-43.13
- Resolves: rhbz#1418943 - If a long-running task (e.g. enumeration) blocks the
                           sssd_be process, sssd_be can deadlock
- Also Require a new-enough version of selinux-policy so that setpgid() by sssd
  is allowed
```

Is xCAT expected to work with selinux?  I worked around this by adding *selinux=0* to addkcmdline.    I found a postinstall script that updates /etc/selinux/config, is that the preferred method?

immarvin:
seems similar situation with issue https://github.com/xcat2/xcat-core/issues/1666

xCAT expect selinux to be disable, xCAT will disable selinux on xCAT MN, we also disabled selinux on diskless node to fix the issue above.

do you mean ``xCAT-server/share/xcat/netboot/rh/compute.rhels7.xxxx.postinstall``? it's for diskless.

seems we might need to disable selinux on diskful nodes since selinux might be installed/updated by default

mattaezell:
We are running diskless, but we have a custom postinstall script that does other stuff (enable non-default services, fixup some config files, etc).  I guess we aren't running the default stuff which you expect will handle this, so I guess it's our responsibility to integrate the existing postscript into our running config (and watch it for updates between versions, since it evolves over time).  The other option is that xCAT unconditionally passes *selinux=0* on the kcmdline so sed in a postscript is not needed

immarvin:
maybe passing `` selinux=0`` kernel option is better than word substitution in compute.rhels7.ppc64.postinstall file. we will consider this in the following release. thanks @mattaezell

[...]

PR #3140, opened 2017-05-25
fix issue RHEL update caused boot failure due to selinux #2996

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/2996:

pass "selinux=0" to the kernel cmdline for redhat7 diskless provision, so that selinux can be disabled on system bootup.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/anaconda.pm
+++ b/xCAT-server/lib/xcat/plugins/anaconda.pm
@@ -859,7 +859,7 @@ sub mknetboot
         }
 
         # turn off the selinux
-        if ($osver =~ m/fedora12/ || $osver =~ m/fedora13/) {
+        if ($osver =~ m/(fedora12|fedora13|rhels7)/) {
             $kcmdline .= " selinux=0 ";
         }
 


daniceexi:
@immarvin Did not this issue happen on earlier version of rh7?

whowutwut:
@daniceexi Yea it's strange we didn't see this happen all the time... .

Why do we need to search certain versions? What happens if we always send selinux=0. . Are we unable to operate with SeLINUX on?

immarvin:
hi @daniceexi @whowutwut , from rhels7.3, selinux related packages is installed as dependency on cn and selinux is enabled by default

the selinux is disabled in the ``xCAT-server/share/xcat/netboot/rh/compute.rhels7.ppc64le.postinstall``

````
#xCAT-server/share/xcat/netboot/rh/compute.rhels7.ppc64le.postinstall
 
35 #-- Disable SELinux in the rootimg
 36 #-- Redhat 7.3 will install selinux-policy and selinux is enabled by default
 37 #-- Need to disable selinux, otherwise, the booting  will hang on "Loading selinux policy"
 38 if [ -f "$installroot/etc/selinux/config" ]
 39 then
 40     sed -i 's/SELINUX=enforcing\|permissive/SELINUX=disabled/' $installroot/etc/selinux/config
 41 fi
````

 in https://github.com/xcat2/xcat-core/issues/2996, @mattaezell uses a customized *.postinstall file, which does not contain the stuff above. This PR covers this scenario.

immarvin:
hi @whowutwut , the rhels7.3 diskless provision will hang if selinux is on. see https://github.com/xcat2/xcat-core/issues/1666

bybai:
Both rhel7 "selinux=0" and compute.rhels7.ppc64le.postinstall disable selinux, since old user may use compute.rhels7.ppc64le.postinstall  for other OS like centos, keep disable selinux in compute.rhels7.ppc64le.postinstall. I am OK to merge.

whowutwut:
@immarvin  Yes, I see that in the issue you mention.  But what about having xCAT support SELINUX? Can we successfully provision if SELINUX is turned on?  I think in those issues it doesn't explain why we need to disable it....

PR #3143, opened 2017-05-26
add future to autodelete unneded nodes when using makeconfluentcfg -l

banuchka:
I'm using multiple instances with confluent as console servers in my environment. I need "makeconfluentcfg -l" execution on all of them via cron, because someone may add/delete/update some node or its attributes.
The problem is when I need to use for example console server "xcat-sn3" for "nodeX" but "nodeX" already present on console server "xcat-sn1". There is a "race condition" because all of 2 servers tried to get the console(ipmi SOL).
Here I used to compare "present list of servers" VS "consoles for that console server" when do "makeconfluentcfg -l" and delete not needed consoles.


banuchka:
Sorry move PR to #3144 because previously i haven't use fresh master

Issue #2664, opened 2017-03-17
[openbmc] rinv support 


whowutwut:
Issue opened to support rinv using openbmc mgmt.

* Inventory is provided by the [openbmc_project/Inventory](https://github.com/openbmc/phosphor-dbus-interfaces/tree/master/xyz/openbmc_project/Inventory) endpoint.
* Firmware related information is provided by the [openbmc_project/Software](https://github.com/openbmc/phosphor-dbus-interfaces/tree/master/xyz/openbmc_project/Software) endpoint 

### OpenPower IPMI
```
   OpenPOWER (using ipmi) server specific:
       rinv noderange [model|serial|deviceid|uuid|guid|vpd|mprom|firm|all]
```

### OpenPower OpenBMC
```
   OpenPOWER (using openbmc) server specific:
       rinv noderange [model|serial|deviceid|firm|cpu|dimm|all]
```

zet809:
Pull Request #2573 include "rinv <node> all".

whowutwut:
@zet809  I will add the issue to the pull request, this way there will be a link appearing on this issue page, instead of the other way around

xuweibj:
For pr #2848 , supported options: cpu, dimm, model, serial, bios, all

Plan to do: memory, firm

xuweibj:
The task was finished in pull request #2848 .

whowutwut:
@gurevichmark I think the development for this function is complete.  If things change from the openbmc code, we should just open issues to track this, or should we keep this epic open for another release?

gurevichmark:
I agree, from what is known now about openbmc, the implementation of `sinv` is finished.

PR #3145, opened 2017-05-26
Enhancements to rinv code to complete development

whowutwut:
Help resolve issue #2664 

Inventory Information is provided by this API: [Inventory](https://github.com/openbmc/phosphor-dbus-interfaces/tree/master/xyz/openbmc_project/Inventory)

In this pull request, I'm looking to fix up the rinv code and attempt to complete all the development required based on the API.  

Some things that are fixed: 
* sorting based on alpha, then numeric 
* only contain the commands that plan to have data , for example, we can probably remove guid and uuid
* Still need to re-evaluate the way we are selecting options in the rinv

### DIFF
--- a/docs/source/guides/admin-guides/references/man1/rinv.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/rinv.1.rst
@@ -28,18 +28,18 @@ BMC/MPA specific:
 \ **rinv**\  \ *noderange*\  {\ **pci | model | serial | asset | vpd | mprom | deviceid | guid | firm | diag | dimm | bios | mparom | mac | all**\ }
 
 
-OpenPOWER (using ipmi) server specific:
-=======================================
+OpenPOWER (IPMI) server specific:
+=================================
 
 
 \ **rinv**\  \ *noderange*\  [\ **model | serial | deviceid | uuid | guid | vpd | mprom | firm | all**\ ]
 
 
-OpenPOWER (using openbmc) server specific:
-==========================================
+OpenPOWER (OpenBMC) server specific:
+====================================
 
 
-\ **rinv**\  \ *noderange*\  [\ **model | serial | deviceid | uuid | guid | vpd | mprom | firm | cpu | dimm | all**\ ]
+\ **rinv**\  \ *noderange*\  [\ **model | serial | mprom | firm | cpu | dimm | all**\ ]
 
 
 PPC (with HMC) specific:
@@ -195,25 +195,25 @@ Calling \ **rinv**\  for VMware will display the UUID/GUID, nuumber of CPUs, amo
 
 \ **mprom**\ 
  
- Retrieves mprom firmware level
+ Retrieves mprom firmware level.
  
 
 
 \ **deviceid**\ 
  
- Retrieves device identification. Usually device, manufacturing and product ids.
+ Retrieves device identification. Usually device, manufacturing and product IDs.
  
 
 
 \ **uuid**\ 
  
- Retrieves the universally unique identifier
+ Retrieves the universally unique identifier.
  
 
 
 \ **guid**\ 
  
- Retrieves the global unique identifier
+ Retrieves the global unique identifier .
  
 
 
--- a/docs/source/guides/admin-guides/references/man1/rinv.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/rinv.1.rst
@@ -39,7 +39,7 @@ OpenPOWER (OpenBMC) server specific:
 ====================================
 
 
-\ **rinv**\  \ *noderange*\  [\ **model | serial | mprom | firm | cpu | dimm | all**\ ]
+\ **rinv**\  \ *noderange*\  [\ **model | serial | firm | cpu | dimm | all**\ ]
 
 
 PPC (with HMC) specific:
--- a/xCAT-client/pods/man1/rinv.1.pod
+++ b/xCAT-client/pods/man1/rinv.1.pod
@@ -16,7 +16,7 @@ B<rinv> I<noderange> [B<model>|B<serial>|B<deviceid>|B<uuid>|B<guid>|B<vpd>|B<mp
 
 =head2 OpenPOWER (OpenBMC) server specific:
 
-B<rinv> I<noderange> [B<model>|B<serial>|B<mprom>|B<firm>|B<cpu>|B<dimm>|B<all>]
+B<rinv> I<noderange> [B<model>|B<serial>|B<firm>|B<cpu>|B<dimm>|B<all>]
 
 =head2 PPC (with HMC) specific:
 
--- a/xCAT-client/pods/man1/rinv.1.pod
+++ b/xCAT-client/pods/man1/rinv.1.pod
@@ -10,13 +10,13 @@ B<rinv> [B<-h>|B<--help>|B<-v>|B<--version>]
 
 B<rinv> I<noderange> {B<pci>|B<model>|B<serial>|B<asset>|B<vpd>|B<mprom>|B<deviceid>|B<guid>|B<firm>|B<diag>|B<dimm>|B<bios>|B<mparom>|B<mac>|B<all>}
 
-=head2 OpenPOWER (using ipmi) server specific:
+=head2 OpenPOWER (IPMI) server specific:
   
 B<rinv> I<noderange> [B<model>|B<serial>|B<deviceid>|B<uuid>|B<guid>|B<vpd>|B<mprom>|B<firm>|B<all>] 
 
-=head2 OpenPOWER (using openbmc) server specific:
+=head2 OpenPOWER (OpenBMC) server specific:
 
-B<rinv> I<noderange> [B<model>|B<serial>|B<deviceid>|B<uuid>|B<guid>|B<vpd>|B<mprom>|B<firm>|B<cpu>|B<dimm>|B<all>]
+B<rinv> I<noderange> [B<model>|B<serial>|B<mprom>|B<firm>|B<cpu>|B<dimm>|B<all>]
 
 =head2 PPC (with HMC) specific:
 
@@ -125,19 +125,19 @@ Diagnostics information of firmware.
 
 =item B<mprom>
 
-Retrieves mprom firmware level
+Retrieves mprom firmware level.
 
 =item B<deviceid>
 
-Retrieves device identification. Usually device, manufacturing and product ids. 
+Retrieves device identification. Usually device, manufacturing and product IDs. 
 
 =item B<uuid>
 
-Retrieves the universally unique identifier
+Retrieves the universally unique identifier.
 
 =item B<guid>
 
-Retrieves the global unique identifier 
+Retrieves the global unique identifier .
 
 =item B<all>
 
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -935,9 +935,17 @@ sub rinv_response {
         my %content = %{ ${ $response_info->{data} }{$key_url} };
 
         if ($grep_string eq "firm") {
+            # This handles the data from the /xyz/openbmc_project/Software endpoint.
+            #
+            # Handle printing out all posssible Software values in a generic format: 
+            #    node: <Purpose> Software: <version> (<Activation>)
+            #
             if (defined($content{Version}) and $content{Version}) {
-                my $firm_ver = "System Firmware Product Version: " . "$content{Version}";
-                xCAT::SvrUtils::sendmsg("$firm_ver", $callback, $node);
+                # TODO: In future, if we want to support ExtendedVersion, we should enable Verbose output.
+                my $purpose_value = (split(/\./, $content{Purpose}))[-1];
+                my $activation_value = (split(/\./, $content{Activation}))[-1];
+                my $software_str = "$purpose_value Software: $content{Version} (Activation=$activation_value)";
+                xCAT::SvrUtils::sendmsg("$software_str", $callback, $node);
                 next;
             }
         }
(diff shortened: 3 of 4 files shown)


whowutwut:
@xuweibj I added code to check the value of Present.  And if it's not present, I do not print it.  
But thinking about this more, do we think we want to have missing hardware indicated.   For example, if there's 16 cores avail, but the machine only has 2, does it make sense to print all 16 and force the user to determine which ones are missing?

whowutwut:
@xuweibj , After some discussion with @gurevichmark in our meeting, I think for now, will continue to print out the items even if Present=0.  I'm not sure how the user will consume this information and perhaps if a core is missing, it would flag an error.

xuweibj:
@whowutwut I agree with you. We could check ``Present=0`` in the future.

whowutwut:
With the last commit, I've removed the following options: 

rinv <noderange> [model|serial|~~deviceid|uuid|guid|vpd|mprom~~|firm|cpu|dimm|all]

I searched the API and I don't see anything that describes those fields. 

In addition, for the keywords that we do support, things like `model` and `serial`, we will not scan all the attributes and print out any model number or serial number returned with the inventory. 

For example: 
```
[root@fs2vm110 ~]# rinv p9euh02 serial
p9euh02: CPU0 SerialNumber : YA3933758688
p9euh02: CPU1 SerialNumber : YA3933756868
p9euh02: DIMM0 SerialNumber : 0x34a38483
p9euh02: DIMM1 SerialNumber : 0x34a3736d
p9euh02: DIMM2 SerialNumber : 0x34a37536
p9euh02: DIMM3 SerialNumber : 0x34a375a6
p9euh02: DIMM4 SerialNumber : 0x34a3847f
p9euh02: DIMM5 SerialNumber : 0x342a9a69
p9euh02: DIMM6 SerialNumber : 0x34a36a59
p9euh02: DIMM7 SerialNumber : 0x34a36afc
p9euh02: DIMM8 SerialNumber : 0x347a72d8
p9euh02: DIMM9 SerialNumber : 0x342a9a26
p9euh02: DIMM10 SerialNumber : 0x342a9a53
p9euh02: DIMM11 SerialNumber : 0x342a9a67
p9euh02: DIMM12 SerialNumber : 0x347a75c8
p9euh02: DIMM13 SerialNumber : 0x347a79ab
p9euh02: DIMM14 SerialNumber : 0x347a75ca
p9euh02: DIMM15 SerialNumber : 0x347a72d9
p9euh02: MOTHERBOARD SerialNumber : Y130UF72701M
p9euh02: SYSTEM SerialNumber : 0
```

Again, we are not looking specifically for data, but instead returning all the applicable values that the API passes to us...

whowutwut:
I just rebased my sandbox and added in the msg when there are 0 objects to return.

whowutwut:
@gurevichmark @xuweibj  This pull request is complete and Ready for review/merge

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -935,70 +935,52 @@ sub rinv_response {
...
-        if ($grep_string eq "mac" and $key_url =~ /\/ethernet/) {
-            my $macaddress = "MAC: " . $content{MACAddress};
-            xCAT::SvrUtils::sendmsg("$macaddress", $callback, $node);
-            next;
-        } 
+        if (! defined $content{Present}) {
+            # This should never happen, but if we find this, contact firmware team to fix...
+            xCAT::SvrUtils::sendmsg("ERROR: Invalid data for $key_url, contact firmware team!", $callback, $node);
+            next; 
+        }
```
xuweibj: I did not see the code to print ``model`` and ``serial`` information, did I miss something?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -935,70 +935,52 @@ sub rinv_response {
...
+            $src = basename $key_url;
+        }
 
-            foreach my $key (keys %content) {
-                $content_info = uc ($src) . " " . $key . " : " . $content{$key};
-                push (@sorted_output, $node . ": ". $content_info); #Save output in array
+        foreach my $key (keys %content) {
+            # If not all options is specified, check whether the key string contains
+            # the keyword option.  If so, add it to the return data
+            if ($grep_string ne "all" and lc($key) !~ m/$grep_string/i ) {
```
xuweibj: Is that right to check whether ``$key`` and ``$grep_string`` is matching ? I think it should be ``$key_url`` not ``$key``.

xCAT-server/lib/xcat/plugins/openbmc.pm
whowutwut: The check is down below at line 968, will reply to your next comment regarding this.

xCAT-server/lib/xcat/plugins/openbmc.pm
whowutwut: The `key_url` is the url string...  ...`chassis/system/motherboard... `   At this point, we are looping through the elements that are under that URL.  

For example: 
```
    "/xyz/openbmc_project/inventory/system/chassis/motherboard": {
      "BuildDate": "",
      "Manufacturer": "0000000000000000",
      "Model": "",
      "PartNumber": "00VK525         ",
      "Present": 0,
      "PrettyName": "SYSTEM PLANAR   ",
      "SerialNumber": "Y130UF72701M    "
    },
```

So in this for loop, the `$key` is the BuildDate, Model, SerialNumber and I'm seeing if the search string is contained in that key.  If it's NOT, then skip, otherwise  add to the `@sorted_output` structure to print out later.

xuweibj: I don't know why we did not print other information such as ``Manufacturer``. And for ``rinv model`` command, we print ``"BOARD Part Number: " . "$content{PartNumber}"`` before, why we change it now?

whowutwut: @xuweibj  I am talking with the firmware team to understand the Inventory interface and push for some more documentation.   

You might be right that we have to use "Part Number".  The `model` for 8335-GTA/GTB is the MTM number, but the BOARD Part number is the motherborad Part Number.  We might have to extract it from system... 
```
    "/xyz/openbmc_project/inventory/system": {
      "BuildDate": "",
      "Cached": 0,
      "FieldReplaceable": 0,
      "Manufacturer": "",
      "Model": "2",
      "PartNumber": "0000000000000000",
      "Present": 1,
      "PrettyName": "",
      "SerialNumber": "0000000000000000"
    },```

So perhaps we should hold off on merging this pull request while I try and get more information about the return values from this API.

**Comment by whowutwut** [comment]
Changed to [DO NOT MERGE] as I try to get more information about the API

File: xCAT-server/lib/xcat/plugins/openbmc.pm (hunk 2 above)
Comment (xuweibj) [inline xCAT-server/lib/xcat/plugins/openbmc.pm]: OK. But another question is that when run command such as ``rinv dimm``, we need to print all dimm information. The word ``dimm`` is in URL, if we use ``$key_url`` to grep, it can be matched. But if we use ``$key``, we can not print all information of dimm.

Comment (whowutwut) [inline xCAT-server/lib/xcat/plugins/openbmc.pm]: :+1: Yes thank you! That is right, I will fix it.

**Comment by whowutwut** [comment]
@xuweibj  @gurevichmark  I think this pull request is complete and ready for merge...


After talking to the firmware team will currently just print all variables out to the user for now.  With your comment about dimm, the current code will check the $key_url and here's an example of what we will see: 

```
[root@fs2vm110 ~]#  rinv p9euh02 dimm | more 
p9euh02: DIMM0 BuildDate : 
p9euh02: DIMM0 Cached : 0
p9euh02: DIMM0 FieldReplaceable : 1
p9euh02: DIMM0 Functional : 1
p9euh02: DIMM0 Manufacturer : 0xce80
p9euh02: DIMM0 Model : M393A4K40BB2-CTD
p9euh02: DIMM0 PartNumber : 
p9euh02: DIMM0 Present : 1
p9euh02: DIMM0 PrettyName : 0x0c
p9euh02: DIMM0 SerialNumber : 0x34a38483
p9euh02: DIMM0 Version : 0x0
p9euh02: DIMM1 BuildDate : 
...
```

File: xCAT-server/lib/xcat/plugins/openbmc.pm [hunk 3]
```diff
@@ -935,70 +935,64 @@ sub rinv_response {
...
             } else {
                 $src = basename $key_url;
             }
 
             foreach my $key (keys %content) {
+                # If not all options is specified, check whether the key string contains
+                # the keyword option.  If so, add it to the return data
+                if ($grep_string ne "all" and ((lc($key) !~ m/$grep_string/i) and ($key_url !~ m/$grep_string/i)) ) {
+                    next;
+                }
```
xuweibj: For ``model`` option, we only need to print motherboard's PartNumber. And for ``serial``, we only need to print motherboard's SerialNumber. If changed like this, we will print other information. And the output format is different from ipmi mgt.

whowutwut: Yes, but this is a new management interface.  

While we want to preserve the behavior in the past and we do for certain things in xCAT, for example `bmcdiscover` is going to take Serial/Model from one attribute.    `rinv` is a command that displays inventory information..   

Why would we limit the information to just a single piece? That means we have to write extra code and cannot generically handle data that is passed to us. 

I think we should return as much applicable information as we can and from the API, many of the inventory data would have Model and Serial number associated with it.   Otherwise users in the future maybe will ask....  "Please add support to display model information for DIMM" and then we have to make extra code changes to support this.    If they do not want to see it, they can always grep to filter more...

zet809: Hi, @whowutwut , we didn't clearly define what shall the 'serial' output. But from my point of view, 'rinv serial' need to output the Serial Number for the server and 'rinv model' output the Model Number(Shall be system PartNumber). For the serial and model of DIMMs, it shall be dealt by the 'rinv dimm' or 'rinv all'. And, for such an information "DIMM0 SerialNumber : 0x34a38483", the user still need to know what "DIMM0" is, and what the Model number for it, right?

docs/source/guides/admin-guides/references/man1/rinv.1.rst
```diff
@@ -28,18 +28,18 @@ BMC/MPA specific:
...
 
 
-OpenPOWER (using openbmc) server specific:
-==========================================
+OpenPOWER (OpenBMC) server specific:
+====================================
 
 
-\ **rinv**\  \ *noderange*\  [\ **model | serial | deviceid | uuid | guid | vpd | mprom | firm | cpu | dimm | all**\ ]
+\ **rinv**\  \ *noderange*\  [\ **model | serial | mprom | firm | cpu | dimm | all**\ ]
```
zet809: The mprom shall be removed here.

xCAT-client/pods/man1/rinv.1.pod
```diff
@@ -10,13 +10,13 @@ B<rinv> [B<-h>|B<--help>|B<-v>|B<--version>]
...
-=head2 OpenPOWER (using ipmi) server specific:
+=head2 OpenPOWER (IPMI) server specific:
   
 B<rinv> I<noderange> [B<model>|B<serial>|B<deviceid>|B<uuid>|B<guid>|B<vpd>|B<mprom>|B<firm>|B<all>] 
 
-=head2 OpenPOWER (using openbmc) server specific:
+=head2 OpenPOWER (OpenBMC) server specific:
 
-B<rinv> I<noderange> [B<model>|B<serial>|B<deviceid>|B<uuid>|B<guid>|B<vpd>|B<mprom>|B<firm>|B<cpu>|B<dimm>|B<all>]
+B<rinv> I<noderange> [B<model>|B<serial>|B<mprom>|B<firm>|B<cpu>|B<dimm>|B<all>]
```
zet809: The mprom shall be removed here.

docs/source/guides/admin-guides/references/man1/rinv.1.rst
zet809: And as you had mentioned in #3184, the uuid may be provided, so, we can add it in the manpage and usage, and just output warning message for the UUID option until it is available. What is you idea?

whowutwut: I'd like to keep this out until we have some idea about the API interface .. right now there's nothing so I cannot complete the code without the need to come back and revisit.  If the API is defined, then I would be OK with your suggestion.   Will continue to track the UUID and make a reference to the issue if I can find out more.

xCAT-client/pods/man1/rinv.1.pod
whowutwut: thanks! nice catch

docs/source/guides/admin-guides/references/man1/rinv.1.rst
zet809: OK, thx

xCAT-server/lib/xcat/plugins/openbmc.pm
zet809: Hi, @whowutwut , any ideas for my suggest?

whowutwut:
@zet809 if you are ok please approve the review

whowutwut:
@zet809 is out of the office for 1/2 of tomorrow, so since we have addressed his questions,  I will remove him as required reviewer.  Should try to get this merged in before tomorrows build.

xCAT-server/lib/xcat/plugins/openbmc.pm
whowutwut: Ok I can change the serial and model to just return the system info... will try to do this today or tomorrow. And can merge after

whowutwut:
@zet809  As you had suggested in the comments, `serial` and `model` will be special case and return only the system level information.  I added the change in the last commit, and the external output is as follows:  

(FYI @junxiawang )

```
[root@fs2vm110 ~]# rinv p9euh01 serial
p9euh01: SYSTEM SerialNumber : 0

[root@fs2vm110 ~]# rinv p9euh01 model
p9euh01: SYSTEM Model : 2
```

zet809:
Looks good to me, I agree to merge. Thx @whowutwut

Issue #2971, opened 2017-05-03
[fvt]2.13.4 nodeset osimage have no right output if command wrong


junxiawang:
env:all os
build:c910f02c01p14:/opt/xcat/share/xcat/tools/autotest/testcase/installation # lsxcatd  -v
Version 2.13.4 (git commit 7a5456217d96284a524433d070311c5bbea2958e, built Fri Apr 28 06:16:00 EDT 2017)
c910f02c01p14:/opt/xcat/share/xcat/tools/autotest/testcase/installation # 
How to reproduce:
```
1. Before the node change osimage:
c910f02c01p14:/opt/xcat/share/xcat/tools/autotest/testcase/installation # lsdef c910f02c02p28
Object name: c910f02c02p28
    arch=ppc64
    cons=hmc
    currchain=boot
    currstate=install sles11.4-ppc64-compute
    groups=lpar,all
    hcp=c910hmc01
    hidden=0
    hwtype=lpar
    id=28
    mac=da:08:4f:d8:b2:03
    mgt=hmc
    monserver=c910f02c01p08
    netboot=grub2
    nfsserver=c910f02c01p08
    nodetype=ppc,osi
    os=sles11.4
    parent=c910f02fsp02
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    pprofile=c910f02c02p28
    profile=compute
    provmethod=sles11.4-ppc64-netboot-compute
    servicenode=c910f02c01p08
2.c910f02c01p14:/opt/xcat/share/xcat/tools/autotest/testcase/installation 
 nodeset c910f02c02p28 osimage= sles11.4-ppc64-install-compute
Error: Did you run "genimage" before running "packimage"? kernel cannot be found
------------------------------> could not change osimage from diskless to diskfull if osimage have space with "=" and osimage name



```

immarvin:
The syntax ``nodeset xxxx osimage=[:SPACE:]yyyyy`` is illegal, will add some logic to check the command syntax of this kind in ``nodeset``.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3147

PR #3147, opened 2017-05-27
fix issue [fvt]2.13.4 nodeset osimage have no right output if command wrong #2971

immarvin:
fix issue: https://github.com/xcat2/xcat-core/issues/2971

complain and return if the argument is invalid, i.e, ``nodeset xxx osimage= [xxxx]``  or ``osimage= [xxxx]``in the chain

````
[root@c910f03c05k21 ~]# nodeset c910f03c05k27 osimage= rhels7.3-ppc64le-netboot-compute
Error: Invalid argument: 'osimage=rhels7.3-ppc64le-install-compute xxxx'
Usage: nodeset <noderange> [shell|boot|runcmd=bmcsetup|iscsiboot|osimage[=<imagename>]|offline]
[root@c910f03c05k21 perl]# echo $?
1
````

### DIFF
--- a/perl-xCAT/xCAT/Usage.pm
+++ b/perl-xCAT/xCAT/Usage.pm
@@ -620,3 +620,41 @@ sub parseCommand {
     return "";
 }
 
+#------------------------------------------------------------------------------
+
+=head3   validateArgs
+    This function validates the arguments of the specified command
+    Arguments:
+        command
+        arguments(array @)
+    Returns:
+        $ref:         a reference to array of [$retcode(integer),$info(string)]
+        $ref->[0]=0 : validation passed
+        $ret->[0]!=0: validation failed, the error info is returned in $ref->[1]
+
+=cut
+
+#-------------------------------------------------------------------------------
+
+sub validateArgs {
+    my $command=shift; 
+    if ($command =~ /xCAT::Usage/) { $command = shift; }
+  
+    my $count=0; 
+    my @extrargs=@_;
+    if($command =~ m/^(nodeset|rinstall|winstall)$/ ){
+        #suppose that argument like "-p foo" have been processed and 
+        #filtered by GetOpt subroutine 
+        #fortunately the commands in this branch does not have such options
+        foreach(@extrargs){
+            if($_ !~ m/^-[-]?\S+/){
+                $count+=1;
+            }
+        }
+        if ($count!=1) {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by robin2008** [comment]
@immarvin 
why not handle the CLI parameters checking in the first plugins, like xnba, petitboot, etc.
It is better to complain the option error, especially for the option format error,  as early as possible then we will not go deeper in business.

**Comment by immarvin** [comment]
hi @robin2008 ,code updated

File: perl-xCAT/xCAT/Usage.pm [hunk 1]
```diff
@@ -620,3 +620,38 @@ sub parseCommand {
...
+
+sub validateArgs {
+    my $command=shift; 
+    if ($command =~ /xCAT::Usage/) { $command = shift; }
+   
+    my @extrargs=@_;
+
+    if($command =~ m/^(nodeset|rinstall|winstall)$/ ){
+        foreach(@extrargs){
+            if($_ !~ m/^-[-]?\S+/){
```
robin2008: So we know our commands does not allow "-p foo" and "--port 12345",  better to describe the limitation in func header.

perl-xCAT/xCAT/Usage.pm
```diff
@@ -620,3 +620,38 @@ sub parseCommand {
...
+                $count+=1;
+            }
+        }
+        if ($count!=1) {
+           return [1,"Invalid argument: '".join(" ",@extrargs)."'"];
```
robin2008: As this error is very straightforward, we can make the error message more descriptive,  just like: 
"Invalid arguments: check if any space is using by wrong in 'xxx'"

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -426,10 +426,13 @@ sub preprocess_request {
         return;
     }
 
-    if (@ARGV == 0) {
+   my $ret=xCAT::Usage->validateArgs($command,@ARGV);
```
robin2008: It may not handle @ARGV == 0 case

immarvin: if ``scala @argv==0`` ,  validateArgs will run into
````
 if ($count!=1) {
 +           return [1,"Invalid argument: '".join(" ",@extrargs)."'"];
````

PR #3153, opened 2017-05-31
Modify 2 bugs for xcattest

hu-weihua:
Found 2 bugs for new version of ``xcattest``

* Failed to run one cases twice due to first run has changed the cases definition.
* Not support ``cmdcheck`` label correctly.


xCAT-test/xcattest
```diff
@@ -1175,11 +1175,11 @@ sub run_case {
 
                 #to run singal line command
```
bybai: Could you fix this typo error "singal" to "single"  by the way?

PR #3160, opened 2017-05-31
Spelling fixes for admin guides

gurevichmark:


whowutwut:
Are you using a tool to find some of these spelling errors?

gurevichmark:
The process is rather manual. I concat all `.rst` files in a subdirectory like `docs/source/guides/admin-guides` into a single file and then view that file in MS Word with spell checking turned on. I then visually scan that file for misspellings.

Issue #2856, opened 2017-04-12
[DEV] nodes provisioned with redhat osimages created by xCAT 2.12.3 or earlier missed yum repo to MN in xCAT 2.12.4 or later


immarvin:
this issue is introduced in PR https://github.com/xcat2/xcat-core/pull/2050/files

neo954:
This problem was found on a testing environment, below are the detailed steps of how to recreate this problem.

* Install xCAT 2.12.4 on a clean newly installed SLES 11.4 x86-64 machine.
* Run `copycds` against RHEL 6.9 iso.
* Update the installed xCAT 2.12.4 to 2.13.2
* Deploy a new compute node with the existing osimage definition `rhels6.9-x86_64-install-compute`
* After a successfully deployment of the compute node, ssh to the compute node, and run `yum install -y bzip2`.

The `yum install` operation on the compute node will fail, due to there is no existing yum repository for the operating system. The correct behavior of xCAT is that xCAT should put a workable yum repository configuration on the newly installed compute node.

neo954:
I consider this problem a **block** issue of the nearing released xCAT 2.13.3. Due to even the default `osimage` created by xCAT, such as `rhels6.9-x86_64-install-compute`, is affected.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/2857

whowutwut:
I provisioned a machine and the local-repository file name seemed to have changed...

```
[root@fs3 yum.repos.d]# cat local-repository-0.repo
[local-rhels7.3-ppc64le-ppc64le]
name=xCAT configured yum repository for /install/rhels7.3/ppc64le
baseurl=http://10.2.5.3//install/rhels7.3/ppc64le
enabled=1
gpgcheck=0
```

I never recall the name is `local-repository-0.repo` why is there a -0?

immarvin:
the fix will be available in xCAT 2.13.3 release.

If you hit this issue after migrate xCAT 2.12.3 or earlier releases to  xCAT releases between 2.12.4 and 2.13.2, the work around is to rerun copycds against the redhat isos.

immarvin:
hi @whowutwut ,  xCAT support user to specify multiple directories in the osimage.pkgdir of the diskfull osimage, there might be yum repos under each of these pkgdir directories, in this case, we will create /etc/yum.repos.d/local-repository-x.repo for each directory.

whowutwut:
[...]

PR #3163, opened 2017-06-01
Do operating system deployment in migration test cases

neo954:
Enhance existing migration test cases.
This pull request is for task #2856

### DIFF
--- a/xCAT-test/autotest/testcase/migration/redhat_migration
+++ b/xCAT-test/autotest/testcase/migration/redhat_migration
@@ -97,6 +97,8 @@ cmd:xdsh $$CN "lsdef"
 check:output=~node0001
 cmd:xdsh $$CN "noderm node0001"
 check:rc==0
+cmd:xdsh $$CN "/opt/xcat/share/xcat/tools/go-xcat smoke-test"
+check:rc==0
 cmd:latest_version_info=`lsxcatd -v`;xdsh $$CN "lsxcatd -v" | grep "$latest_version_info"
 check:rc==0
 cmd:if [[ -f /tmp/servicelabel ]];then chdef $$SN -p groups=service;rm -rf /tmp/servicelabel;fi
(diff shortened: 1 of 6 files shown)


hu-weihua:
@neo954 , I did not get the usage of ``go-xcat smoke-test`` in ``go-xcat`` usage information. But anyway, I don't think your change has achieved what we expected. I suggest we should do below steps:

* install lower version xcat
* Do basic definitions base on lower version xcat, such like makedns, makedhcp, node definition and osimage definition.
* upgrade to higher version xcat
* Do provision to verify all definitions still work in new version xcat

xCAT-test/autotest/testcase/migration/redhat_migration
```diff
@@ -93,12 +95,14 @@ check:rc==0
...
 check:rc==0
 check:output=~running
-cmd:xdsh $$CN "lsdef"
-check:output=~node0001
-cmd:xdsh $$CN "noderm node0001"
-check:rc==0
 cmd:latest_version_info=`lsxcatd -v`;xdsh $$CN "lsxcatd -v" | grep "$latest_version_info"
 check:rc==0
+cmd:ssh $$CN "nodeset $$SN osimage=__GETNODEATTR($$CN,os)__-__GETNODEATTR($$CN,arch)__-install-compute"
+cmd:rpower $$SN boot
```
hu-weihua: Maybe use ``rinstall`` is better option, this case need to be run on LPAR.

xCAT-test/autotest/testcase/migration/redhat_migration
```diff
@@ -93,12 +95,14 @@ check:rc==0
...
+check:rc==0
+cmd:ssh $$CN "a=0; while ! `lsdef -l $$SN grep status|grep booted >/dev/null`; do ((++a > 90)) && break; sleep 20; done"
+cmd:ssh $$CN "xdsh $$SN date"
+check:rc==0
```
hu-weihua: In fact, we did not check if the SN was installed successfully. right? If SN was not installed successfully,but enter initrd,  when line 103 is time out, line 104 still can pass.  So after line 103, we still need to check the ``booted`` again.

neo954:
@hu-weihua, Test cases updated based on your comments.

hu-weihua:
It is good for me, I agree to merge. thanks @neo954

PR #3165, opened 2017-06-01
Remove 'stop' label from all provision cases

hu-weihua:
Due to not find useful information and can not stop any more, so remove ``stop`` label from provision cases. Will add in the future if needed.


bybai:
I am OK to merge it.

PR #3170, opened 2017-06-01
rewrite if-clause in post.rhels7

fz883:
the original if clause does not check the interfaces correctly. moving the "then" of the if up a bit makes it a) nice to read and b) work as expected.
I had trouble with enabled (onboot=yes) interfaces, despite not having a link at all.

PR #3174, opened 2017-06-01
fix typo in if-clause in post.rhels7 

immarvin:
fix a typo in post.rhels7  to make it work as expected

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/post.rhels7
+++ b/xCAT-server/share/xcat/install/scripts/post.rhels7
@@ -9,11 +9,11 @@
 for i in $(find /etc/sysconfig/network-scripts/ifcfg-*|egrep -v ifcfg-lo )
 do 
   nicname=$(echo $i|awk -F 'ifcfg-' '{print $2}')
-  if ethtool $nicname|grep -E -i "Link detected.*yes" >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


whowutwut:
Duplicate to Community Pull Request #3170

whowutwut:
@hu-weihua This slipped through...  I added the label to it...  can we take a look at why we did not catch this issue? Could we have checked some return code and found errors?  The previous blame commit that touched this line seems from > 2 years ago?

hu-weihua:
@whowutwut , test case just check the output and return code of command. Up to now, there is not a case catch error from log or console. That means if some script have error and throw error just into log or console, test case can't catch such error. 

But from other standpoint, if there is such error, our xcat command should return error code and show error information in STDOUT.  It is little hard for test case to go through log.  Maybe ``xCAT Probe`` can do it, than test case  leverage the output of ``xCAT Probe`` to do judgement.

For script syntax problem, maybe the CI project which we are doing will helpful.

robin2008:
@hu-weihua we don't care the script of syntax, but the final result of the user requirement.  For example:   here xCAT want the Linked interfaces are onboot=yes.  Or for un-linked interfaces are not, etc.

BTW: the ifcfg file may be modified also by other postscripts, so our test case checking is focusing the e2e result,  and if possible to check some key checkpoints in the whole processing.

Issue #2939, opened 2017-04-25
[Feature request] Change `makeknownhosts` options to match other `make*` commands


samveen:
As of now, most `make*` commands use the option `-d` to `delete`/`remove` configuration for noderanges.

However, the `makeknownhosts` command doesn't follow that logic. Please add `-d` as an alias for `-r` for the `makeknownhosts` command:

`make*` commands:

```
[samveen@xcatmn1 upgrade]$ man makehosts |grep -- -d
       makehosts [-n] [noderange] [-l|--longnamefirst] [-d] [-m|--mactolinklocal]
       -d    Delete rather than create records. This will also delete any additional network interfaces (NICs) included in the node definitions.
[samveen@xcatmn1 upgrade]$ man makedns |grep -- -d
       makedns [-V|--verbose] [-e|--external] [-d|--delete noderange]
       -d|--delete
          makedns -d node1
[samveen@xcatmn1 upgrade]$ man makedhcp |grep -- -d
       makedhcp -a -d [-l|--localonly]
       makedhcp -d noderange [-l|--localonly]
       If you need to delete node entries from the DHCP configuration, use the -d flag.
       -d noderange
       -a -d       Delete all node entries, that were added by xCAT, from the DHCP server configuration.
[samveen@xcatmn1 upgrade]$ man makeconservercf |grep -- -d
       makeconservercf [-V|--verbose] [-d|--delete] [noderange]
       If -d is specified, makeconservercf will remove specified nodes from /etc/conserver.cf file. If noderange is not specified, all xCAT nodes
       -d|--delete
          makeconservercf -d node01
[samveen@xcatmn1 upgrade]$ man makeroutes |grep -- -d
       makeroutes [-d|--delete] [-r|--routenames r1[,r2...]]
       makeroutes noderange [-d|--delete] [-r|--routenames r1[,r2...]]
       -d|--delete
          makeroutes node1,node2 -d -r rr1,rr2
```
`makeknownhosts`:
```
[samveen@xcatmn1 upgrade]$ man makeknownhosts |grep -- -d
[samveen@xcatmn1 upgrade]$ man makeknownhosts |grep -- -r
       makeknownhosts noderange [-r|--remove] [-V|--verbose]
       -r|--remove
          makeknownhosts node02 -r
```

samveen:
@immarvin, @cxhong  If you want, I can submit a MR for this, as it is a trivial change.

immarvin:
[...]

PR #3180, opened 2017-06-03
[WIP]Fix for #2939: Add `-d` option for `makeknownhosts` as alias for `-r`

samveen:
This is an effort to fix issue #2939

@whowutwut I have not made changes to `docs/source/guides/admin-guides/references/man8/makeknownhosts.8.rst` as I found that the file was not in unix `fileformat` (`^M` characters showed up when editing with `vim`)

Should I leave that file out from this MR?

### DIFF
--- a/xCAT-server/lib/xcat/plugins/makeknownhosts.pm
+++ b/xCAT-server/lib/xcat/plugins/makeknownhosts.pm
@@ -65,13 +65,14 @@ sub process_request
     if   ($request && $request->{arg}) { @ARGV = @{ $request->{arg} }; }
     else                               { @ARGV = (); }
 
-    my $usage = "Usage: makeknownhosts <noderange> [-r] [-V]\n       makeknownhosts -h";
+    my $usage = "Usage: makeknownhosts <noderange> [-r|-d] [-V]\n       makeknownhosts -h";
 
     # print "argv=@ARGV\n";
     if (!GetOptions(
             'h|help'    => \$::opt_h,
             'V|verbose' => \$::opt_V,
             'r|remove'  => \$::opt_r
+            'd|delete'  => \$::opt_d
         ))
     {
         my $rsp = {};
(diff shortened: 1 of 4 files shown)


bybai:
It is a trivial change. As discussed in #2939, all agreed to add -d alias for -r.  I am OK to merge it. Hi @samveen ,For makeknownhosts.8.rst, I can edit it and summit PR.

xCAT-server/lib/xcat/plugins/makeknownhosts.pm
```diff
@@ -65,13 +65,14 @@ sub process_request
...
 
-    my $usage = "Usage: makeknownhosts <noderange> [-r] [-V]\n       makeknownhosts -h";
+    my $usage = "Usage: makeknownhosts <noderange> [-r|-d] [-V]\n       makeknownhosts -h";
 
     # print "argv=@ARGV\n";
     if (!GetOptions(
             'h|help'    => \$::opt_h,
             'V|verbose' => \$::opt_V,
             'r|remove'  => \$::opt_r
+            'd|delete'  => \$::opt_d
```
whowutwut: @samveen  I just updated and ran the code with `XCATBYPASS=1` and found a bug here, need a command before line 75...

```
String found where operator expected at /opt/xcat/lib/perl/xCAT_plugin/makeknownhosts.pm line 75, near "$::opt_r
            'd|delete'"
        (Missing operator before 'd|delete'?)
Error loading module /opt/xcat/lib/perl/xCAT_plugin/makeknownhosts.pm  ...skipping
```

whowutwut:
@hu-weihua  I don't know who else is looking at CI,  but can you pass along this msg.  The first thing we should do before even running any regression bucket is to get CI working with running a command with `XCATBYPASS=1` .  We can help find these types of "syntax" problems when pull request are created. 
This would be a really good first use case to solve

hu-weihua:
@whowutwut , Is running ``XCATBYPASS=1`` against one command helpful for this issue? Unless we run all command base on ``XCATBYPASS=1``, we can avoid this kind of error. We don't know which plugin has such problem ahead. If we run all command, it looks like a regression. It is good option for us to put it in CI? 

We can check syntax during CI, but if one issue has to be exposed during running, it is other story?

whowutwut:
@hu-weihua I think in the process of using bypass, xcatd will load the plugins and here is where we see these types of errros come out

PR #3185, opened 2017-06-05
Add rvitals test cases to pegas+p9 bundle

hu-weihua:
Add rvitals test cases to pegas+p9 bundle


junxiawang:
rvitals_leds need to add in bundle

junxiawang:
I agree

PR #3203, opened 2017-06-05
When using go-xcat, continue to prompt the user for y/n before aborting

whowutwut:
Minor usability issue, but when accidentally hitting a key (that is not y or n)  while go-xcat is reading the repositories will cause `go-xcat` to abort , so I added in a loop checking for the valid response before aborting. 

Here's the behavior with the changes: 

### Cancel case 
```
[root@fs2vm105 ~]#        /tmp/go-xcat --xcat-version devel install
Operating system:   linux
Architecture:       ppc64le
Linux Distribution: rhel
Version:            7.3

Reading repositories ...... done

xCAT is going to be installed.
Continue? [y/n] a
Unknown Response!
Continue? [y/n] b
Unknown Response!
Continue? [y/n] c
Unknown Response!
Continue? [y/n] d
Unknown Response!
Continue? [y/n] o
Unknown Response!
Continue? [y/n] n
Good-bye!
```

### Install case
```
[root@fs2vm105 ~]#        /tmp/go-xcat --xcat-version devel install
Operating system:   linux
Architecture:       ppc64le
Linux Distribution: rhel
Version:            7.3

Reading repositories ...... done

xCAT is going to be installed.
Continue? [y/n] a
Unknown Response!
Continue? [y/n] b
Unknown Response!
Continue? [y/n] y
Resolving Dependencies
--> Running transaction check
---> Package conserver-xcat.ppc64le 0:8.1.16-10 will be installed
---> Package elilo-xcat.noarch 0:3.14-4 will be installed
---> Package grub2-xcat.noarch 0:2.02-0.16.el7.snap201506090204 will be installed
---> Package ipmitool-xcat.ppc64le 0:1.8.17-1 will be installed
...
...
```

### DIFF
--- a/xCAT-server/share/xcat/tools/go-xcat
+++ b/xCAT-server/share/xcat/tools/go-xcat
@@ -2,7 +2,7 @@
 #
 # go-xcat - Install xCAT automatically.
 #
-# Version 1.0.19
+# Version 1.0.20
 #
 # Copyright (C) 2016 International Business Machines
 # Eclipse Public License, Version 1.0 (EPL-1.0)
--- a/xCAT-server/share/xcat/tools/go-xcat
+++ b/xCAT-server/share/xcat/tools/go-xcat
@@ -1567,14 +1567,22 @@ case "${GO_XCAT_ACTION}" in
 	then
 		echo
 		echo "xCAT is going to be ${GO_XCAT_ACTION/%e/}ed."
-		read -r -p "Continue? [y/n] "
-		case "${REPLY}" in
-		"Y"*|"y"*)
-			;;
-		*)
-			echo "Aborting ..."
-			exit 0
-		esac
+		while true; do
+
+			read -r -p "Continue? [y/n] "
+			case "${REPLY}" in
+			"Y"*|"y"*)
+				break	
+				;;
+			"N"*|"n"*)
+				echo "Good-bye!"
+				exit 0
+				;;
+			*)
+				echo "Invalid response!"
+				;;
+			esac
+		done
 	fi
 	# Use `-y' here. Since the STDOUT is redirect to tee.
 	# `yum' does not display the prompt message properly when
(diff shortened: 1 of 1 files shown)


whowutwut:
@neo954 Can you review this when you have a chance?

neo954:
Cool. Please also change line 5 and increase the version number. The version number will make our support works easier when external customers report issues to us.
BTW the version number shows when you run `go-xcat --help`.

neo954:
And also, the version number shows in the `User-Agent` header when `go-xcat` does HTTP query. If one day, we can get the apache access log of xcat.org, we can do log analysis. Currently, we cannot. We do not have read permission for files under `/var/log`.

whowutwut:
Sorry, too a while to get back to this.  Thanks! I bumped up the version number.

neo954:
I will merge this one.

xCAT-server/share/xcat/tools/go-xcat
```diff
@@ -1645,10 +1653,11 @@ case "${GO_XCAT_ACTION}" in
...
-		    for sh,
-		        \`source /etc/profile.d/xcat.sh'
-		    or csh,
-		        \`source /etc/profile.d/xcat.csh'
+		For sh: 
+			source /etc/profile.d/xcat.sh
+
+		For csh: 
+			source /etc/profile.d/xcat.csh
 		EOF
```
neo954: @whowutwut, I found one problem after the merge. We lost the indents of the outputs after your change. Since, the outputs was generated with here document, `<<-`, thus, the `<TAB>`s in front of the line will be omitted. In order to generate indents to the output, spaces must be placed after `<TAB>`s, like the old codes.

    xCAT has been installed!
    ========================

    If this is the very first time xCAT has been installed, run the following
    commands to set environment variables into your PATH:

    For sh:
    source /etc/profile.d/xcat.sh

    For csh:
    source /etc/profile.d/xcat.csh

whowutwut: @neo954  Yea, I intentionally removed the tabs, but I see you put back in #3370  .  That is ok.

PR #3208, opened 2017-06-06
add -y when run yum install git

xuweibj:
add -y when run yum install git


hu-weihua:
It is ok for me, agree to merge.

PR #3219, opened 2017-06-06
Modify one bug:Use of implicit split to @_ is deprecated at xcattest line 1059

hu-weihua:
Fix bug for ``xcattest`` : ``Use of implicit split to @_ is deprecated at xcattest line 1059.``

Found this bug on sles.  The OS of our control node in production automation environment is sles.


bybai:
I am OK to merge it to let autotest environment test it.

PR #3225, opened 2017-06-07
fix testcase clear_openbmc_simulator error

xuweibj:
1. modified path of change_ip.sh in testcase
2. add protection for kill simulation process


hu-weihua:
It's ok for me.

PR #3226, opened 2017-06-07
Revise all migration test cases

neo954:
Fix a few hardware control problems while do the operating system deployment.
Make the test cases wait longer for the operating system deployment.


robin2008:
@neo954 @hu-weihua 
Do we still need this enhancement for migration test cases? If yes, could we schedule it in our sprint plan?

PR #3227, opened 2017-06-07
fix 3224 nodeset create dracut.* file in sles statelite cost too much…

### LINKED ISSUE #3224: [customer]nodeset create dracut.* file in sles statelite cost too much time
This description is from Galid Berman lenovo mail:
because with 24 service nodes, we get 50 (!!) such files for each node (2 per service node and 2 per xcat master), the nodeset process is SO SLOW… removing the files will only makes it worse, because it will put extra load on that directory.

Under my investigate:
`genimage` and `nodeset` can create `rootimg/tmp/dracut.* ` files, `liteimg` will copy and pack the image, in `liteimg` enhancement, `liteimg` is to check and delete `rootimg/tmp/dracut.*` files first. Since genimage is used for both diskless and diskful, I do not delete these files in `nodeset` or `genimage`, enhance `liteimg` is enough. see issue #3142 

We use other code to replace dracut command in nodeset to check if dracut is used or not. So that there is no dracut.* files are created by nodeset, and the same time nodeset does not need to call shell command and system command, it may improve the nodeset performance.

We can update the function in code /opt/xcat/lib/perl/xCAT_plugin/sles.pm,  use this issue to trace enhancement;
```
sub using_dracut
{
    my $rootimgdir = shift;
    #my $chkcmd = "chroot $rootimgdir/rootimg dracut --list-modules";
    #my $rc = system($chkcmd);
    #if ($rc) {
    if ( -f "$rootimgdir/rootimg/etc/dracut.conf" ) {
        return 1;
    } else {
        return 0;
    }
}
```

bybai:
fix #3224 
Replace dracut command in sles statelite nodeset  to check if dracut is used or not. So that there is no dracut.* files are created by nodeset , and the same time nodeset does not need to call shell command and system command, it can improve sles statelite nodeset performance.

Unit test:
```
] # nodeset bybc0605 osimage=sles12.2-x86_64-statelite-compute
bybc0605: statelite sles12.2-x86_64-compute
] # ls /install/netboot/sles12.2/x86_64/compute/rootimg/tmp
] #
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/sles.pm
+++ b/xCAT-server/lib/xcat/plugins/sles.pm
@@ -1707,9 +1707,7 @@ erver, if so, stop it first and try again" ],
 sub using_dracut
 {
     my $rootimgdir = shift;
-    my $chkcmd = "chroot $rootimgdir/rootimg dracut --list-modules";
-    my $rc = system($chkcmd);
-    if ($rc) {
+    if ( -f "$rootimgdir/rootimg/etc/dracut.conf" ) {
         return 0;
     } else {
         return 1;
--- a/xCAT-server/lib/xcat/plugins/sles.pm
+++ b/xCAT-server/lib/xcat/plugins/sles.pm
@@ -1708,9 +1708,9 @@ sub using_dracut
 {
     my $rootimgdir = shift;
     if ( -f "$rootimgdir/rootimg/etc/dracut.conf" ) {
-        return 0;
-    } else {
         return 1;
+    } else {
+        return 0;
     }
 }
 


xCAT-server/lib/xcat/plugins/sles.pm
```diff
@@ -1707,9 +1707,7 @@ erver, if so, stop it first and try again" ],
 sub using_dracut
 {
     my $rootimgdir = shift;
-    my $chkcmd = "chroot $rootimgdir/rootimg dracut --list-modules";
-    my $rc = system($chkcmd);
-    if ($rc) {
+    if ( -f "$rootimgdir/rootimg/etc/dracut.conf" ) {
         return 0;
```
immarvin: so return value is 0 for dracut module? please compare the return value of ``using_dracut`` in anaconda.pm

daniceexi:
@bybai I tried hard to understand what happened but failed. Could you add more information for others to understand?

bybai:
Hi @daniceexi , we use `chroot $rootimgdir/rootimg dracut --list-modules` to check if `dracut` is used or not in `nodeset` command. `nodeset` call this command `dracut --list-modules` for sles statelite, this command will generate 1 dracut.* tmp files under rootimg/tmp files for one node, `nodeset` call this command twice, it will create 2 dracut.* files when executing `nodeset` for one node , these 2 files are 0 bytes and useless.  In Galid env, he has 24 service nodes, he gets 50 (!!) such dracut.* files for each node (2 per service node and 2 per xcat master), he has total 3.5K compute nodes, `nodeset` generate more or less 50 dracut.* files for each node so he get lots of dracut.* files,  the nodeset process is SO SLOW…   In my enhancement, I used `rootimg/etc/dracut.conf` to replace `dracut --list-modules` command to check if there is dracut or not, if there is dracut.conf, there is `dracut` and we will use dracut, and tmp dracut.* files will not be created, and `nodeset` will deduce call `system` command to execute `dracut` command twice, it can improve the performance. I looked into code for redhat, the function is totally different with sles, so this problem only existed in sles statelite `nodeset`. In my first pull request, I mis-return 0 and 1, based on my unit test, I commit the second code changes. That should work well. Thanks.

immarvin:
hi @daniceexi , another advantage of this PR is to reduce the time cost of nodeset, since invoking shell command "dracut --list-modules" cost significant time compared with the logic to check the existence of a file.

immarvin:
agree to merge

daniceexi:
@bybai thanks for your clarification, it's clear enough. Could we add this kind of information in the description part of the issue so that others can easily understand what happened in this issue?

PR #3229, opened 2017-06-07
Revert "Do operating system deployment in migration test cases"

neo954:
Reverts xcat2/xcat-core#3163, since these modified test cases cause auto test failures.


immarvin:
agree to merge

PR #3238, opened 2017-06-08
rm / from path of xcatpost folder when run rm command

xuweibj:
When run ``updatenode``, the folder ``/xcatpost`` and files under it did not be deleted. So change path of the folder to correct one.


immarvin:
agree to merge

PR #3240, opened 2017-06-08
modify rvitals and rinv testcase for issue:3184

junxiawang:
add case for rvitals altitude option for issue:#3240
1.rvitals_altitude
This one was missed in last sprint
If node supported this option there will be more than one output if could not there will be only one output"No attributes returned from the BMC."
2.rinv test case add and modified
Testcases need to add:
rinv_h
rinv_v
rinv_help
rinv_uuid
rinv_cpu
rinv_dimm
rinv_errorcommand
rinv_mixnode
rinv_wrongbmcpasswd

Testcases need to refine:
rinv_model
rinv_serial
rinv_firm
rinv_all
rinv_noderange_err
rinv_serial(ppc64le)
rinv_model(ppc64le)
rinv_firm(ppc64le)
rinv_deviceid(ppc64le)
rinv_vpd(ppc64le)
rinv_all(ppc64le)
rinv_noderange_err(ppc64le)


gurevichmark:
I was told that Witherspoon machines will not return any data for "altitude" attribute. It is a valid parameter for a OpenBMC controlled machine but the Witherspoon implementation will not have any sensors in it to measure altitude. In the future where might be a version of p9 machine build with such a sensor, but not the Witherspoon model.
So I think the testcase should set expected output of `No attributes returned from the BMC.`

whowutwut:
@junxiawang  Please edit the description and add the issue number to make the github links. 

I'm not sure I agree with this test case..  ... the way this testcase is written we are testing for a specific behavior of the `rvitals <> altitude`.  This means that this test case cannot be used if we wanted to test xCAT against other mgmt that support Power9 and higher... ... 

Why do we need to verify that "No attributes returned from the BMC." ?   My general question is how do we test this function so that we can ensure nothing is broken in xCAT.   No attributes returned is a valid behavior.    So rather,  I think we should count the values returned from the command, if this message is returned, there should NOT be anything else returned.  If there is some items returned, this message should NOT print out, and if nothing comes back from the command, there is a problem.   This comment applies to all the rvitals testcases written. 

The behavior or rvitas is such that we print back the information that is passed to us.  If we wanted to validate the information, we would need to run curl commands against the BMC itself, save that data and compare it against what xCAT reports.. i don't think we plan to do that right now.

junxiawang:
@whowutwut  I agree with your point.Too many output for rvitals on different machines for the same option ,we could not cover all scenarios for all machines if just using output to check.I will modify the testcases according to your comments. Thank your

junxiawang:
@whowutwut 
My testcase plan  will cover:
1.only one node
1.1.if command's return code !=0 command error
1.2. if return code=0
      1.2.1 if command's output only have 1 line and ":" only have 1 so output should be "No attributes returned from the BMC"
      1.2.2 if command's output have have 1 line and ":"  do not have 1 and besides output have no than 1 lines so output should not be contained "No attributes returned from the BMC"
2.have more than one node
  2.1 if two type node ,one is invalid so return code !=0
  2.2 if two type node ,command's return code =0 and all node's output should be more than one lines and should get right output
  2.3 if two type node,command's return code=0 and one node could get right output and another get only "No attributes returned from the BMC"

tingtli:
@junxiawang  Do we have cases for rinv_guid and rinv_mprom

tingtli:
@junxiawang  Please see the manpage. We do have guid and mprom.
   OpenPOWER (using openbmc) server specific:
       rinv noderange {model|serial|deviceid|uuid|guid|vpd|mprom|firm|cpu|dimm|all}
@xuweibj  Would you please confirm.

whowutwut:
@tingtli the man page is updated in #3145 but is waiting for @zet809 for approval before merging

gurevichmark:
I am ok with the changes. @xuweibj you can merge if you also agree.

xuweibj:
Looks good to me. I will merge it.

Issue #3239, opened 2017-06-08
Upgrade conserver to 8.2.1 version


chenglch:
Work items:
- Apply the patches from xcat-dep to conserver_8.2.1
- build new conserver-xcat packages for 11 systems
  rh7 x86_64, rh7 ppc64, rh7 ppc64le, rh6 ppc64, rh6 x86_64, ubuntu 14.04 x86_64, ubuntu
  14.04 ppc64le, sles 11.4 x86_64, sles12 x86_64, sles12 ppc64le, sles11.4 ppc64
- Convert sslauthority to sslcacertificatefile in xcat-core
- Migration config script to upgrade xcat from earlier version
- Ansible playbook to build conserver: https://github.com/chenglch/xcat-play/blob/master/ansible/conserver-xcat/README.rst

#3193 
xcat-dep: https://github.com/xcat2/xcat-dep/pull/22

PR #3252, opened 2017-06-13
Upgrade conserver to 8.2.1 version

chenglch:
This patch change the `sslauthority` to `sslcacertificatefile` to
support the latest version of conserver. It also add scripts to
update the `conserver.cf` and $HOME/.consolerc to make sure
migration works correctly.

implement: #3239

### DIFF
--- a/xCAT-server/lib/xcat/plugins/conserver.pm
+++ b/xCAT-server/lib/xcat/plugins/conserver.pm
@@ -207,7 +207,16 @@ sub docfheaders {
         {
             push @newheaders, "config * {\n";
             push @newheaders, "  sslrequired yes;\n";
-            push @newheaders, "  sslauthority /etc/xcat/cert/ca.pem;\n";
+            my $version = xCAT::Utils::get_conserver_version();
+            if (!$version) {
+                xCAT::SvrUtils::sendmsg([ 1, "Failed to get conserver version" ], $cb);
+                return;
+            }
+            if (xCAT::Utils::calc_conserver_version($version) < xCAT::Utils::calc_conserver_version("8.1.19")) {
+                push @newheaders, "  sslauthority /etc/xcat/cert/ca.pem;\n";
+            } else {
+                push @newheaders, "  sslcacertificatefile /etc/xcat/cert/ca.pem;\n";
+            }
             push @newheaders, "  sslcredentials /etc/xcat/cert/server-cred.pem;\n";
             push @newheaders, "}\n";
         }
--- a/xCAT-server/sbin/xcatconfig
+++ b/xCAT-server/sbin/xcatconfig
@@ -464,6 +464,10 @@ if ($::INITIALINSTALL || $::FORCE || $::UPDATEINSTALL || $::genCredentials)
 
 }
 
+if ($::UPDATEINSTALL) {
+    upgrade_conserver();
+}
+
 # more  config needed after xcatd start
 if ($::INITIALINSTALL || $::FORCE)
 {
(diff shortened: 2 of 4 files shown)


daniceexi:
@chenglch What if we keep both of the **sslauthority** and **sslcacertificatefile**? If this change can work for both old and new version that would make our change simpler.

chenglch:
@daniceexi User may be confused by the similar configurations from different versions, actually one of them does not work. 

I think we have two options:
1. Check `conserver` version while running xcatconfig or makeconservercf, if the version is not `8.2.1`, keep the original behavior, otherwise update the configurations.  This is what has been done in the current pull request.

2. If user hope to upgrade xcat-core to this version(2.13.5), he/she must upgrade the xcat-dep at the same time. We need to update the spec file to require the latest version of conserver-xcat. In this situation, we just need to keep the configuration for the latest conserver.  

I prefer the option 2 , any comments @zet809 @robin2008  ?

daniceexi:
@chenglch thanks for the explanation. 
For your two options: option2 is what xcat generally using. but option1 maybe more reasonable since it depends on the conserver version directly.

zet809:
Hi, @chenglch , I am good with option 2. For latest xcat-core, it will depend on latest xcat-dep. 
The only concern is user using older xcat-core with latest xcat-dep, it won't work, right? And we can advice them to update xcat-core. If the user want to use latest conserver without updating xcat-core, is it possbile? Any work around?

chenglch:
Hi @zet809 , for the scenario you mentioned, we can't give the advice message about upgrading xcat-core. Users must follow the rule: use the latest xcat-dep with the latest xcat-core.

robin2008:
In fact, I prefer the option 1 as it is more friendly to end user. 
But I'm also okay with option 2. We just need to announce we cannot support old conserve anymore.

chenglch:
Current conclusion is Option 1 which consider the compatibility with the old version. This patch is ready to review or merge now.

xCAT-server/sbin/xcatconfig
```diff
@@ -2338,3 +2342,45 @@ sub startnamedonboot
...
+#-----------------------------------------------------------------------------
+
+=head3    upgrade_conserver
+
+     Update conserver configuration files while upgrading xcat
+
+=cut
+
+#-----------------------------------------------------------------------------
+sub upgrade_conserver
```
robin2008: Just one concern,  is it possible to put the configuration upgrade into spec?

xCAT-server/lib/xcat/plugins/conserver.pm
```diff
@@ -207,7 +207,16 @@ sub docfheaders {
         {
             push @newheaders, "config * {\n";
             push @newheaders, "  sslrequired yes;\n";
-            push @newheaders, "  sslauthority /etc/xcat/cert/ca.pem;\n";
+            my $version = xCAT::Utils::get_conserver_version();
+            if (!$version) {
+                xCAT::SvrUtils::sendmsg([ 1, "Failed to get conserver version" ], $cb);
+                return;
+            }
+            if (xCAT::Utils::calc_conserver_version($version) < xCAT::Utils::calc_conserver_version("8.2.1")) {
```
robin2008: The new parameter is introduced in 8.1.19,  but for conserver-xcat, we don't have such package, we only have 8.1.16.  Maybe we just need to consider == 8.2.1. Or <=8.1.19

```
version 8.1.19 (Sep 26, 2013):
	- prevent select/read loop when EOF on non-pty input (console) -
	  reported by Chris Marget <chris@marget.com>
	- "!" syntax prefixing use of group names not honored - reported by
	  Zonker <consoleteam@gmail.com>
	- fixed memory leak using timestamps - patch by Karvendhan M.
	  <Karvendhan.M@netapp.com>
	- deprecated --with-cycladests (noop now) - cross-compilation should
	  work without it as autologin now expects setpgrp() to take two
	  arugments instead of testing for it
	- no automatic checks for an empty password when using PAM
	  authentication - based on discussion with Ryan Kirkpatrick
	  <linux@rkirkpat.net>
	- added 'sslcacertificatefile' and 'sslcacertificatepath' client
	  configuration options - based on patch by Aki Tuomi <cmouse@cmouse.fi>
	- added 'sslcacertificatefile' and 'sslreqclientcert' server
	  configuration options
```

xCAT-server/sbin/xcatconfig
chenglch: These configurations for conserver are created by xcat `makeconservercf` and the certification files like ca.pem is created by `xcatconfig`, actually, the package of conserver-xcat do not know the detail information about how to set the ssl or port information correctly.

robin2008: In fact, the configuration file is belong to the conserver package. 
```
# rpm -qf /etc/conserver.cf
conserver-xcat-8.1.16-10.x86_64
```
Normally, it is better to cover the configuration change in package itself.  
But this is not mandatory, we can decide what method to use. And if we use other method to cover the upgrade, we need to cover both the case: MN and SN, or using confluent instead of conserver.

PR #3253, opened 2017-06-13
update for won't fix issue

tingtli:
Update for confignics won't fix issue for nicaliases test case.


hu-weihua:
Due to we will use ``confignetworks`` to replace ``confignics``, so this test will be covered by new test cases against ``confignetworks``.  so just update this case to avoid failure in automation environment.

hu-weihua:
I agree to merge. thanks @tingtli .

whowutwut:
@tingtli can you update the issue title to have some meaning. The current title is missing some text

Issue #3254, opened 2017-06-13
Couldn't find ID in the discinfo database


cxhong:
I am trying to run copycds for rhels 7.0 iso.  it couldn't find ID in the discinfo database, but it was define in there.

```
# copycds /mnt/xcat/iso/redhat/7/RHEL-7.0-20140507.0-Server-ppc64-dvd1.iso
Error: copycds could not identify the ISO supplied, you may wish to try -n <osver>
````
captured logs from xcatd:
```
xCAT: Allowing copycds /mnt/xcat/iso/redhat/7/RHEL-7.0-20140507.0-Server-ppc64-dvd1.iso for root from localhost
distname =
copycd: GLOB(0x100162aa568), 1399449226.155578, RHEL-7.0 Server.ppc64, ppc64,
distname =
INFO - Could not find ID=1399449226.155578 in the discinfo database for OS=RHEL-7.0 Server.ppc64 ARCH=ppc64 NUM=
INFO - Attempting to auto-detect...
INFO - Could not auto-detect operating system.
```
```
# grep 155578 discinfo.pm
    "1399449226.155578" => "rhels7",        #ppc64
`````

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3256

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3258

cxhong:
Verified the fixes.  Thanks.

PR #3256, opened 2017-06-14
fix issue Couldn't find ID in the discinfo database #3254

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3254

the variables exported in modules should not be ``my``


zet809:
Hi, @immarvin , the fix looks good to me.
I found the code is changed on Jan 20, I think all the OS mentioned in the file will have the copycds issue mentioned in #3254 . Need a test case to cover that scenario.

Issue #2690, opened 2017-03-21
[Test case] refine test cases to check hostname in sles12.2 statelite


bybai:
add test case to check hostname in sles12.2 statelite #2616

junxiawang:
Testcases need to refine:
reg_linux_statelite_installation_flat
reg_linux_statelite_installation_hierarchy_by_nfs
reg_linux_statelite_installation_hierarchy_by_ramdisk

bybai:
Hi @junxiawang , yes, these 3 testcases should be enhanced, that is enough. Thanks.

PR #3257, opened 2017-06-14
 refine test cases to check hostname in sles12.2 statelite  #2690

junxiawang:
refine test cases to check hostname in sles12.2 statelite  #2690 
Will check hostname after statelite provision succeed


bybai:
I am OK to merge it.

PR #3258, opened 2017-06-14
fix issue Couldn't find ID in the discinfo database #3254

immarvin:
fix issue #3254

the variables exported in modules should not be ``my``


zet809:

Issue #3159, opened 2017-05-31
[openbmc]rflash support for firmware file upload


gurevichmark:
OpenBMC `rflash` command support - priority 4:

Firmware file upload, do not activate
`rflash <node> <file> -u|--upload`

gurevichmark:
Fixed by PR#3261

Issue #3173, opened 2017-06-01
[openbmc]rflash support for firmware upload+update


gurevichmark:
OpenBMC `rflash` command support - priority 1:

File upload and update.
`rflash <node> <file> --activate`

Update files are expected to be in `.tar` format.

gurevichmark:
There is no single action provided by openbmc to upload and activate a file in one step.
Possible algorithm might be:

1. Upload a file and look at its version in MANIFEST file.
2. Once file finishes upload run `/software/enumerate`
3. Look through all the output entries and find the match of "Version" attribute to the one from MANIFEST.
4. Activate the ID that matched in step 3 above

This assumes there can be no duplicate version strings returned from `/software/enumerate`

whowutwut:
@gurevichmark we would post this question to the firmware team and see if they can provide the function.  Or at least confirm that the process seems OK

gurevichmark:
I discussed this algorithm with Adriana before posting it here, and she agreed that it should work.

zet809:
Hi, @gurevichmark , can we close this task since the code had been included in 2.13.6 RC build?

gurevichmark:
I think this issue should remain open. I am not written any code to support this function.
Currently 2 steps are needed - upload file then activate the image.
This issue is to support upload and activation in 1 step.

gurevichmark:
PNOR file upload and activation implemented by PR #3750

zet809:
Thx, @gurevichmark , more friendly for users.

PR #3261, opened 2017-06-14
openbmc rflash command support for -u --upload option

gurevichmark:
Implements issue #3159 

Upload update `.tar` file to BMC using `curl` command.
Unable at this point to implement the upload using HTTP::Request(), so for now using `curl` command so as to not block the development progress.

Successful upload:
```
[root@stratton01 ~]# rflash p9euh02,p9euh01 /tmp/gurevich/pnor.squashfs.tar -u
p9euh01: Update file /tmp/gurevich/pnor.squashfs.tar successfully uploaded
p9euh02: Update file /tmp/gurevich/pnor.squashfs.tar successfully uploaded
[root@stratton01 ~]#
```

Login problems:
```
[root@stratton01 ~]# rflash p9euh02,p9euh01 /tmp/gurevich/pnor.squashfs.tar -u
p9euh01: Unable to login : 401 Unauthorized - Invalid username or password
p9euh02: Unable to login : 401 Unauthorized - Invalid username or password
[root@stratton01 ~]#
```

Upload problems:
```
[root@stratton01 ~]# rflash p9euh02,p9euh01 /tmp/gurevich/pnor.squashfs.tar -u
p9euh01: Failed to upload update file /tmp/gurevich/pnor.squashfs.tar : 403 Forbidden - The specified resource cannot be created: '/upload/imagea/pnor.squashfs.tar'
p9euh02: Failed to upload update file /tmp/gurevich/pnor.squashfs.tar : 403 Forbidden - The specified resource cannot be created: '/upload/imagea/pnor.squashfs.tar'
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -36,6 +36,7 @@ $::POWER_STATE_POWERING_OFF="powering-off";
 $::POWER_STATE_POWERING_ON="powering-on";
 $::POWER_STATE_QUIESCED="quiesced";
 $::POWER_STATE_RESET="reset";
+$::UPLOAD_FILE="";
 
 $::NO_ATTRIBUTES_RETURNED="No attributes returned from the BMC.";
 
@@ -124,6 +125,13 @@ my %status_info = (
         init_url       => "$openbmc_project_url/software/enumerate",
     },
     RFLASH_LIST_RESPONSE => {
+        process        => \&rflash_response,
+    },
+    RFLASH_FILE_UPLOAD_REQUEST  => {
+        method         => "PUT",
+        init_url       => "$openbmc_project_url/upload/image/",
+    },
+    RFLASH_FILE_UPLOAD_RESPONSE => {
         process        => \&rflash_response,
     },
 
@@ -217,6 +225,7 @@ my %status_info = (
 $::RESPONSE_OK                  = "200 OK";
 $::RESPONSE_SERVER_ERROR        = "500 Internal Server Error";
 $::RESPONSE_SERVICE_UNAVAILABLE = "503 Service Unavailable";
+$::RESPONSE_METHOD_NOT_ALLOWED  = "405 Method Not Allowed";
 
 #-----------------------------
 
@@ -677,6 +686,7 @@ sub parse_command_status {
             if ($update_file =~ /.*\.tar$/) {
                 # Filename ending on .tar was specified
                 $filename = $update_file;
+                $::UPLOAD_FILE = $update_file; # Save filename to upload
                 if ($check_version) {
                     # Display firmware version of the specified .tar file
                     my $firmware_version_in_file = `$grep_cmd $version_tag $filename`;
@@ -710,8 +720,9 @@ sub parse_command_status {
             return 1;
         }
         if ($upload) {
-            xCAT::SvrUtils::sendmsg("Upload option is not yet supported.", $callback);
-            return 1;
+            # Upload specified update file to  BMC
+            $next_status{LOGIN_RESPONSE} = "RFLASH_FILE_UPLOAD_REQUEST";
+            $next_status{"RFLASH_FILE_UPLOAD_REQUEST"} = "RFLASH_FILE_UPLOAD_RESPONSE";
         }
     }
 
@@ -832,7 +843,13 @@ sub gen_send_request {
     if ($method eq "GET") {
         $debug_info = "$node: DEBUG $method $request_url";
     } else {
-        $debug_info = "$node: DEBUG $method $request_url -d $content";
+        if ($::UPLOAD_FILE) {
+            # Slightly different debug message when doing a file upload
+            $debug_info = "$node: DEBUG $method $request_url -T " . $::UPLOAD_FILE;
+        }
+        else {
+            $debug_info = "$node: DEBUG $method $request_url -d $content";
+        }
     }
     print "$debug_info\n";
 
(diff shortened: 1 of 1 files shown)


zet809:
Hi, @gurevichmark , what are these 3 section output for? A little confusing.

gurevichmark:
@zet809 I just wanted to show what the output looks like for 3 different situations:
1. Everything worked well
2. Error output when there were problems logging into the BMC
3. Error output when the upload of the update file failed.

whowutwut:
I'm looking at the externals here and trying to play out the use case in my mind.  Normally without the `-u` option, the file will be uploaded and flashed.   My concern here is that it's very easy to leave out the -u option, and then the firmware will be uploaded and flashed, maybe there's no way around it  if we want to provide this function :-)

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -677,6 +686,7 @@ sub parse_command_status {
             if ($update_file =~ /.*\.tar$/) {
                 # Filename ending on .tar was specified
                 $filename = $update_file;
+                $::UPLOAD_FILE = $update_file; # Save filename to upload
```
whowutwut: Is there anywhere that we validate that the file provided actually does exists?  Or does the attempt to upload the file later break this?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1420,6 +1446,45 @@ sub rflash_response {
...
+        # TODO remove this block when proper request can be generated
+        if ($::UPLOAD_FILE) {
+            my $request_url_login  = "$http_protocol://" . $node_info{$node}{bmc} .  "/login";
+            my $request_url_logout = "$http_protocol://" . $node_info{$node}{bmc} .  "/logout";
+            my $request_url_upload = "$http_protocol://" . $node_info{$node}{bmc} .  "/upload/image/";
+            my $content_login = '{ "data": [ "' . $node_info{$node}{username} .'", "' . $node_info{$node}{password} . '" ] }';
+            my $content_logout = '{ "data": [ ] }';
+
+            # curl commands
+            my $curl_login_cmd  = "curl -c cjar -k -H 'Content-Type: application/json' -X POST $request_url_login -d '" . $content_login . "'";
```
whowutwut: Instead of defining the 3 variables above and only using it here, we could create a variable for `request_url` and then fill in the string on these `curl_login_cmd`, `curl_logout_cmd`, and `curl_upload_cmd` variables.

For example:
```
my $request_url = "$http_protocol://" . $node_info{$node}{bmc};

my $curl_login_cmd = "curl -c cjar -k -H 'Content-Type: application/json' -X POST $request_url/login -d '" . $content_login . "'";
my $curl_logout_cmd = "curl -c cjar -k -H 'Content-Type: application/json' -X POST $request_url/logout -d '" . $content_login . "'";
...
```

gurevichmark:
Perhaps we can add a `--activate` or `--commit` option to the `rflash` command, similar to `rflash` command format for HMC and FSP managed machines. That, however, will make it a bit different from ipmi managed machines where you just specify .hpm file with no flags to update.

whowutwut:
@gurevichmark  I do like the option to explicitly indicate what we want to happen with a --commit or --activate..   Maybe without any options, we should default to list.  This way the user will not accidentally update when they really want to show...   

I think there is enough change in the format  (.tar file vs .hpm..)  ability to store other versions on the BMC...   maybe enough changes here to justify a behavior change and make sure that it make sense usability wise, not just to sync with historical behavior.  

If I were to be very cautious during a update...  Perhaps I would do the following..
1) To prepare for the service or maintenance window, upload the new firmware to the machines to stage it. 
 This can be done ahead of time to save upload time during the maintenance window.  Maybe this time is not much... don't know. 
2) Now query against all the BMCs to make sure they all have the correct level of the firmware planned to be updated, now we have successfully staged the BMC to be ready to flash during the window... 
3) When the maintenance window is active, start flashing the BMCs in a small range, verify it looks OK...  or flash a range... 
4) Revert to an older level if bad?  

Without an extra flag, it seems easier to type but maybe more error prone?

gurevichmark:
Added new commit to implement suggested changes:
1. Check for upload file existence and readability
2. Collapse multiple `curl_*_cmd` lines into one.

I like the `--activate` flag to explicitly say that update is to be performed. There is noting to change for this PR, but the PR related to issues #3173 abd #3273 will be implement that flag.

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -687,6 +687,11 @@ sub parse_command_status {
                 # Filename ending on .tar was specified
                 $filename = $update_file;
                 $::UPLOAD_FILE = $update_file; # Save filename to upload
+                # Verify file exists and is readable
+                unless (-r $filename) {
+                    xCAT::SvrUtils::sendmsg("Error accessing update file $filename", $callback);
```
whowutwut: Can we send back the "E" for error in sendmsg?

whowutwut: Suggest changing this to be more standard error message, similar to: 
```
[root@c910f02c05p03 consoles]# ls /root/abc.tar
ls: cannot access /root/abc.tar: No such file or directory
```

Maybe a message like:   "Cannot access $filename" would be sufficient

gurevichmark:
Added a new commit with a better error message when invalid update file specified. Output:

```
[root@stratton01 xcat]# rflash p9euh02 /tmp/gurevich/pnor.squashfs5.tar -u
Error: Cannot access /tmp/gurevich/pnor.squashfs5.tar
[root@stratton01 xcat]# echo $?
1
[root@stratton01 xcat]#
```

PR #3267, opened 2017-06-15
add GPGSIGN =0 option to skip repo sign in ubuntu build script

immarvin:
for the travis CI:

SIGN=0 can be passed to avoid gpg sign 

Since travis CI will build xcat-core and xcat-dep in travis CI VM, it is not appropriate to transfer xcat private keys to it.

### DIFF
--- a/build-ubunturepo
+++ b/build-ubunturepo
@@ -22,6 +22,8 @@
 #                           When you are ready to release this build, use PROMOTE=1 without PREGA
 #               BUILDALL=1 - build all rpms, whether they changed or not.  Should be used for snap builds that are in
 #                            prep for a release.
+#               GPGSIGN=0 -  Do not sign the repo in the end of the build. The repo will be signed by default 
+#                            
 #               LOG=<filename> - provide an LOG file option to redirect some output into log file
 #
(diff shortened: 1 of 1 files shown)


whowutwut:
Not strong on this comment, so I'm okay leaving it as is or changing it..

But the `buildcore.sh` script uses `RPMSIGN` as the variable, should we do something similar, DEBSIGN or even GPGSIGN to make it more understandable.  Initially I didn't understand the title of this pull request: SIGN=0 until after i read the description.

immarvin:
hi @whowutwut , I prefer ``GPGSIGN`` and updated the code

whowutwut:
I think we need to update the autobuild scripts, or this change will break the daily build.  

Here's the changes that I made for RPMSIGN. 
https://github.ibm.com/xcat2/autobuild/search?q=RPMSIGN&type=Commits&utf8=%E2%9C%93

Seems fairly straightforward.. let me create a PR.

whowutwut:
I merged a PR to enable this option for gpg signing...  Let's merge this so we can test it in tomorrows daily build.

PR #3271, opened 2017-06-15
add hardware control bundle for different hardware type

tingtli:
This pull request is to put following bundles in test bundle directory. hdctrl_general.bundle is basic hardware control bundle which has the general test cases list. Other bundles are for different hardware type. These bundles will be used when we run hardware control verification in autotest. 
hdctrl_general.bundle
hdctrl_bmc.bundle
hdctrl_openpower_ipmi.bundle
hdctrl_openpower_openbmc.bundle
hdctrl_ppc_hmc.bundle


hu-weihua:
@tingtli , could you add description for all of bundles, we can support ``description:`` label in bundle file. 

Another question is if hardware control commands just include ``rpower``, ``rvitals`` and ``rinv``? what about command ``rsetboot``, ``rspconfig`` and ``reventlog``? Do we need to include these commands?

tingtli:
@hu-weihua  Yes , we have rsetboot and rspconfig test cases, but reventlog is a manually test case. Will update the  bundle files.

xCAT-test/autotest/bundle/hdctrl_bmc.bundle
```diff
@@ -0,0 +1,16 @@
...
+rvitals_wattage
+rvitals_fanspeed
+rvitals_power
+rvitals_leds
+rsetboot_net_statcheck
+rsetboot_cd_statcheck
+rsetboot_default_statcheck
+rsetboot_h
+rsetboot_help
+rsetboot_v
```
hu-weihua: Should ``restboot_h``, ``rsetboot_help`` and ``rsetboot_v`` belong to ``hdctrl_general.bundle``?

tingtli: Since rsetboot is only supported on bmc-based server. I didn't put rsetboot cases in general bundle.

hu-weihua: I think all cases against ``rsetboot`` can be added into `` hdctrl_general.bundle``.

tingtli: seem manpage of rsetboot. I don't think it could be applied to all hardware. It is only for BMC-based servers.

hu-weihua: Even now all cases against ``rsetboot`` just be verified on bmc-based server, but these cases were designed to cover all kinds of hardware control method. So they should belong to ``hdctrl_general.bundle``. If some of them can't work against other hcp, they are bug. But it is another story.

tingtli: They are not be verified on bmc-base servers. They are only supported on bmc-based servers. So, if they don't work on other hcp, they work as design. not bugs.

hu-weihua:
It is good for me, agree to merge it. thanks @tingtli

PR #3278, opened 2017-06-16
[DO NOT MERGE] travis CI: grant traivis permission to run xcat commands

immarvin:
For travis CI:

in the travis CI VM, the granted user is "travis" instead of "root", there is no chance to run tab/obj commands after xCAT installation, do this during xCAT installation


zet809:
Hi, @immarvin , why this issue is close?

immarvin:
this will be covered by the ``chtab`` command in travis script, not in xCAT

PR #3280, opened 2017-06-16
With mid June, 2017 firmware update, model number is moved from PartNumber -> Model

whowutwut:
It's unfortunate that we are bouncing back and forth on this.. but due to sheer luck, while Investigating a separate issue, I compared the json data being returned from one p9 machine to another which was recently updated to the following levels... 

```
[root@stratton01 openbmc]# ./show_firmware_level p9euh02
PRETTY_NAME="Phosphor OpenBMC (Phosphor OpenBMC Project Reference Distro) v1.99.6-130"
[                                                  ] 0%IBM-witherspoon-ibm-OP9_v1.17_1.20
```

When looking at the diff, the model information has been moved from PartNumber -> ModelNumber.  
image

Creating this PR to sync up...

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -1074,8 +1074,8 @@ sub bmcdiscovery_openbmc{
         my $serial;
 
         if (defined($response->{data})) {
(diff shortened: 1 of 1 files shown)


xuweibj:
Looks good to me.

Issue #3260, opened 2017-06-14
The Site table attribute snmpc for community string can't apply to all the switches in the cluster


cxhong:
This issue is reported from research system.  This system has different kinds of switches.  The site attribute for the snmp community string that xcat should use when communicating with the switches is
```
"snmpc","xxxxxxxxx",, 
```
but for Mellanox IB switch, the community string is "_public_"

the xcatprobe switch_macmap produces timeouts for the ib switches
```
[root@c460mgt01 ~]# xcatprobe switch_macmap
...
c460f04ib2 - Timeout                                                                          [FAIL]
c460f04ib1 - Timeout                                                                          [FAIL]
c460ibcore1 - Timeout                                                                         [FAIL]
c460f02ib2 - Timeout                                                                          [FAIL]
c460f03ib1 - Timeout                                                                          [FAIL]
c460f05ib1 - Timeout                                                                          [FAIL]
c460ibcore2 - Timeout                                                                         [FAIL]
c460f06ib1 - Timeout                                                                          [FAIL]
c460f01ib2 - Timeout                                                                          [FAIL]
c460f02ib1 - Timeout                                                                          [FAIL]
c460f03ib2 - Timeout                                                                          [FAIL]
c460f01ib1 - Timeout                                                                          [FAIL]
c460f06ib2 - Timeout                                                                          [FAIL]
```

zet809:
Which version of snmp used for Mellanox IB switch? All of those switches are using snmp v1?
If yes, they shall set "password" attribute of switches table as the correct community string for the specified switch node.

cxhong:
yes, set "password" attribute of switches table as the correct community string worked for this case.
I will go through the xcat code make sure we used snmppassword instead of community string.

cxhong:
verified different community string on research system.  If default community string is different for switches, need to add the community string to snmppassword attribute in the switches table, for example:
````
chdef switchname snmppassword=commpublic
```

zet809:
Hi, @cxhong , the fix is merged, will you pls verify?

PR #3282, opened 2017-06-16
Check community string in the switches table

cxhong:
side issue from #3260.
 In one cluster, if there are different switches with different community string,  need to check snmppassword in the switches table before default community string and site table attribute **"snmpc"**, and use that to send snmp command.


zet809:

Issue #3181, opened 2017-06-04
[FVT] Offer one bundle for new hardware control and implement the missed cases


hu-weihua:
what to do:

* Offer one bundle for new hardware control
* List the test cases we still missed now 
* implement the missed cases

The doc to refer to
https://w3-connections.ibm.com/wikis/home?lang=en#!/wiki/xCAT/page/New%20Hardware%20Support%20Checklist

tingtli:
Since PR# 3271 is merged. I will close this one.

PR #3285, opened 2017-06-19
add missed cases for hardware control commands. Fix bug 3269.

tingtli:
Add 5 cases in this pull request
This pull request is for task #3181.

[case 1]
Case Name:rpower_softoff
description:This case is to test softoff option could remote shutdown nodes

[case 2]
Case Name: rpower_onstandby
description:This case is to test onstandby option could bring server to standby state

[case 3]
Case Name: rinv_mprom
description:this case is to test mprom option for rinv on x86_64 servers.

[case 4]
Case Name: rinv_guid
description:this case is to test guid option for rinv on x86_64 servers.

[case 5]
Case Name:rinv_dimm
description:this case is to test dimm option for rinv on x86_64 servers.

### DIFF
--- a/xCAT-test/autotest/testcase/rinv/cases2
+++ b/xCAT-test/autotest/testcase/rinv/cases2
@@ -1,0 +1,26 @@
+start:rinv_mprom
+hcp:ipmi
+arch:x86_64
+cmd:rinv $$CN mprom 
+check:rc==0
+check:output=~BMC Firmware:\s*\w+.\w+
+end
+
+start:rinv_guid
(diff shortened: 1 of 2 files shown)


hu-weihua:
@tingtli , why did you delete the description of many test cases against ``rpower`` command?

tingtli:
@hu-weihua sorry. It seems i mixed up the base version. I've updated it.

hu-weihua:
It is good for me, will merge it. thanks @tingtli

Issue #3286, opened 2017-06-19
Conserver-xcat should be required to update to 8.2.1 after code change in PR #3252 


whowutwut:
With PR #3252 , we broke compatibility with older versions of conserver.   Using the new xCAT code with the older version of conserver results in this msg: 

```
console: SSL negotiation failed
70367486123328:error:14094418:SSL routines:SSL3_READ_BYTES:tlsv1 alert unknown ca:s3_pkt.c:1259:SSL alert number 48
```

The pull request should require a > 8.2.1 change in the spec file so we pull in the updated version of conserver when updating xCAT.

whowutwut:
@neo954 @hu-weihua  I don't think our migration testing catches this, does it?   Essentially, we are running a 2.13.4 base and pointing to the xcat-core and xcat-deps public repo.  Update xCAT and see if functions still run.  Here because conserver is not updated at the same time as #3252 , it breaks conserver...

hu-weihua:
@whowutwut , actually, we did not have any case against ``conserver`` and ``rflash``  in daily regression up to now. it belongs to totally manual case. Because it is hard for a automatic case to monitor console.  Not to mention the migration. You brought up a good question.  I will try to find out a way to automatically test it. maybe monitor ``/var/log/consoles/CN`` is a option

neo954:
@whowutwut,

Actually, xCAT team did not expect the end user to mix older version `xcat-core` and newer version `xcat-dep` or vice versa. xCAT team expect the end user use `xcat-core` and `xcat-dep` matching each other.

I do not think xCAT provides such forward/backward compatible by mixing old `xcat-dep` and new `xcat-core`.

In the old days, xCAT packages all have broken package dependency declarations. At that time, try to do xCAT upgrading with command `yum install xCAT` will get only one package, `xCAT` to be upgraded. That is before pull requests #2195 and #2198 were getting merged.

For this historical reason, all xCAT migration test cases covered this problem by not calling `yum install xCAT` along, but use a long list of `xcat-core` and `xcat-dep` package names. And `go-xcat` covered this problem as well.

We may need to improve our test case coverage if you do think this is a critical problem.

whowutwut:
@chenglch does the pull request fix the error message that is listed in the initial description?   It's not clear from the pull request information and I'm not familiar with that code. So does this mean any version of conserver should work?

So  in this issue I suggested to make a spec file change to pull in the updated version and your pull request Fixes a calculation errro. Is that the plan of action?

[...]

PR #3291, opened 2017-06-20
Fix the calculation error for conserver version number

chenglch:
This patch fix the error while calculating the version number.
Use int instead of ord to convert string to integer.

fix-issue: #3286


zet809:

PR #3292, opened 2017-06-20
If start simulator failed, will exit not mkdef nodes

xuweibj:
If start simulator failed, will exit not mkdef nodes.


hu-weihua:
It is ok for me. will merge it. thanks @xuweibj

PR #3061, opened 2017-05-16
use rpm inside chroot to query dracut version

aaronknister:
if the version of rpm on the host differs from the version of rpm in the image (e.g. sles11 host creating a sles12 image) the "rpm --root" will fail. instead if we the rpm from the chroot environment itself we can wor around this.

PR #3293, opened 2017-06-20
To replace pull request 3061: use rpm inside chroot to query dracut v…

zet809:
…ersion

To replace pull request #3061


immarvin:
agree to merge

Issue #3011, opened 2017-05-09
trace udevadmin LTC issue


bybai:
trace udevadmin LTC issue: https://bugzilla.linux.ibm.com/show_bug.cgi?id=153609
If this issue is fixed, update xcat getinstdisk code.

bybai:
Based on LTC comments, I uploaded unsuccessful and successful log.

bybai:
Hi @whowutwut , I got comments from LTC as following:
```
------- Comment From gduarte@redhat.com 2017-06-14 14:46:47 EDT-------
This looks like a duplicate of RHBZ 1438976 (LTC 153199).

Please update the firmware to skiboot-5.6.0 and retest.
```

bybai:
"udevadm info --attribute-walk --name=/dev/sda" in %pre section of kickst can work well when firmware is skiboot-5.6.0-rc2 on p9euh02.

bybai:
Fixed in #3204, will verify in autotest env.

robin2008:
Yuan have already verified in autotest env, and it works.

PR #3294, opened 2017-06-20
Fix and revert old udevadm code for 3011

bybai:
For issue #3011 

I tested this on p9euh02, the diskful installation can be successful. So revert getinstdisk to old code.


immarvin:
i am ok with the change

PR #3303, opened 2017-06-21
Revert "add GPGSIGN =0 option to skip repo sign in ubuntu build script"

immarvin:
Reverts xcat2/xcat-core#3267

This PR breaks the daily build on ubuntu, revert to avoid blocking daily test.

Will refine and create another PR


whowutwut:
Let's us thr autobuild system to build your pull request when it's ready. I should have done that earlier. Will take a look at the build output later

Issue #3264, opened 2017-06-15
bmcdiscovery error


cganglin:
Using xCAT 2.13.5.
Compute Nodes are Minsky 
All compute nodes are in the same state - bmc set to dhcp.  Dhcp address received from MN without issues.
When i do discovery I get the below message
```
[root@mn1 ~]# bmcdiscover --range 10.10.1.0/24
10.10.1.102,8335-GTB,7817BBA,,,mp,bmc
Error: Can not get status of 2200 port for ip 10.10.1.106.
```
If i run the command immediately after the node is discovered OK.  Wait a little and run bmcdiscover and the error is produced again.  Is this something to be concerned about?

zet809:
hi, @xuweibj , will you pls take a look at this issue? Thx!

xuweibj:
Hi @cganglin , could you help to run ``nmap 10.10.1.106 -p 2200 -Pn`` and show me the result?

whowutwut:
Closing this out, @cganglin  Please re-open this if you continue to see issues using the latest builds.

cganglin:
Sorry,was swamp for some time. Now 2.13.5 is official will try everything again and give feedback. Thanks again @whowutwut

PR #3311, opened 2017-06-22
Modified judging condition of openbmc in bmcdiscover, if not open enter bmcdiscover_ipmi

xuweibj:
To avoid #3264 .
If result of nmap is not 2200...open, consider the node as ipmi and enter bmcdiscovery_ipmi function.


whowutwut:
Aye, I didn't see this pull request, but it should be added into 2.13.5 release.  @cxhong  can you take a quick look at the pull request and if it looks OK, please merge.

whowutwut:
@xuweibj  If you provide some example with the before and after for the pull request, we can merge these things fairly quickly..   essentially proof of the UT and we don't need to be too careful about syntax.   And since openbmc support is beta, we want to push things in quickly with low risk...

cxhong:
the code looks fine.  I will merge it.

PR #3314, opened 2017-06-22
Format the ouput of getadapter result

chenglch:
This patch format the result of getadapter command and update
the man page to notice that `getadapter` will boot into genesis.

Fix-issue: #3046

Command example:
root@c910f05c01bc01k10:~# getadapter kvmguest1 -f
kvmguest1: Booting into genesis, this could take several minutes...
[kvmguest1] scan successfully below is result:
kvmguest1:[0]->mac=42:af:0a:05:65:6b|linkstate=UP
kvmguest1:[1]->pci=00:03.0|model=Red Hat, Inc Virtio network device

kvmguest1: nics talbe will not be updated as not any useful information could be found with udevadm command.

### DIFF
--- a/docs/source/guides/admin-guides/references/man1/getadapter.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/getadapter.1.rst
@@ -31,7 +31,7 @@ DESCRIPTION
 
 Traditionally, network interfaces in Linux are enumerated as eth[0123...], but these names do not necessarily correspond to actual labels on the chassis. \ **getadapter**\  help customer to get predictable network device name and some other network adapter information before provision or network configuration.
 
-\ **getadapter**\  use genesis to collect network adapters information, so that mean it need to restart the target node.
+\ **Since getadpter uses genesis to collect network adapters information, the target node will be restarted.**\ 
 
 \ **getadapter**\  For each node within the <noderange>, follows below scheme:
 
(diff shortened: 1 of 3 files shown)


gurevichmark:
@chenglch Few comments:
1. I do not think you should modify the Usage in getadapter.pm. In my view Usage output is just a quick summary of the command options, it should not give any descriptions or warnings.
2. The man page "Description" section already has a sentence about genesis. Instead of adding a new message, I think we can change the existing message to be more noticeable. Perhaps just change the existing sentence "getadapter use genesis to collect network adapters information, so that mean it need to restart the target node." to be "B<Warning>: Since B<getadapter> uses genesis to collect network adapters information, the target node will be restarted."
3. Since you are already making changes to the man page of getadapter, can you also:
a. Fix this sentence "Below are the possible information can be collect up to now:" to be "Collected information:".
b. The block of text under "Below are the possible information can be collect up to now:" was intended to be a list, but in the man page it gets displayed as a paragraph. Can you change the formatting so it appears as a list, one item per line ?

chenglch:
Thanks @gurevichmark , I have updated this patch with amend.

gurevichmark:
Thanks. Changes look good to me.

bybai:
I am OK to merge it.

Issue #3327, opened 2017-06-26
When run bmcdiscover, show bmcdiscover plugin bug


xuweibj:
------START::bmcdiscover_h::Time:Sun Jun 25 09:50:11 2017------

FILENAME:/opt/xcat/bin/../share/xcat/tools/autotest/testcase//bmcdiscover/cases0

RUN:bmcdiscover -h [Sun Jun 25 09:50:11 2017]
ElapsedTime:1 sec
RETURN rc = 1
OUTPUT:
Error: bmcdiscover plugin bug, pid 29886, process description: 'xcatd SSL: bmcdiscover for root@localhost: bmcdiscover instance' with error 'Not a HASH reference at /opt/xcat/lib/perl/xCAT_plugin/bmcdiscover.pm line 89.
'
CHECK:rc == 0   [Failed]

------END::bmcdiscover_h::Failed::Time:Sun Jun 25 09:50:12 2017 ::Duration::1 sec------

xuweibj:
Verified in automation.

PR #3328, opened 2017-06-26
Fix issue 3327, check environment of request type in openbmc and bmcdiscover

xuweibj:
#3327
 
If have exported ``XCAT_OPENBMC_DEVEL`` and did not export ``XCAT_DEV_WITHERSPOON``, when run bmcdiscover command, will print ``Error: bmcdiscover plugin bug, pid 29886, process description: 'xcatd SSL: bmcdiscover for root@localhost: bmcdiscover instance' with error 'Not a HASH reference at /opt/xcat/lib/perl/xCAT_plugin/bmcdiscover.pm line 89.``

So modified like below:

If ``$request->{environment}`` is 'ARRAY', use ``$request->{environment}->[0]->{XCAT_xxxx}`` to get env value.


zet809:

PR #2763, opened 2017-03-28
Reduce `grep` invocations by passing `-[46]` to `ip`

samveen:
The xCAT code has a lot of invocations of `ip` without paramters which outputs `IP` (both `v4` and `v6`) and `link` information,  when just one of the 3 pieces of information is needed (using `-[46]` and `-o`. 

I've traced down all such redundent filters using the pattern `'grep\s\+ipv6|grep\s\+-v\s\+ipv6` invocations and removed them, adjusting the `ip` and `awk` invocations appropriately.

PR #3353, opened 2017-06-27
The MAC is not got correct when doing hardware discovery

zet809:
An issue introduced by the pull request #2763 
Which will cause the incorrect MAC address got by hardware discovery.
The issue:
```
<nic>
<devname>enP1p1s0f0</devname>
<driver>tg3</driver>
<hwaddr>1000</hwaddr>   ==> shall be MAC address here
```
After fix:
```
<nic>
<devname>enP1p1s0f0</devname>
<driver>tg3</driver>
<hwaddr>98:BE:94:05:D7:80</hwaddr>
```


zet809:
@daniceexi had helped to review, so merge the code.

whowutwut:
@hu-weihua Is it possible to consider looking at running the do discovery script and checking the xml output at the end? Maybe we can add a --test option that does not send a request to xcat or gzips up the xml so we can test the output and at least verify that the MAC address is correct.  Without MAC address discovery is broken. So this would be the minimum we should test

zet809:
hi, @whowutwut , we shall modify the dodiscovery script to support the --test option if we don't want it send out the discovery request or gzip, then there can be a case to cover the checking.
My proposal is, check the format of MAC address, if it is invalid, send out a log to MN.

PR #3357, opened 2017-06-27
Improve the debug message when bmc value cannot be obtained from getbmcconfig

### LINKED ISSUE #3356: Discovery message for P9 node does not help debug User configuration error.
Running discovery on p9 ITC machines, In xCAT Genesis, the XML request returns bad response..

The XML request is the following: 
```
<xcatrequest>
<command>getbmcconfig</command>
<callback_port>300</callback_port>
<isopenbmc>1</isopenbmc>
<bmcmac></bmcmac>
</xcatrequest>
```

And in xCAT genesis, we get 
```
  <error>Invalid table configuration for bmcconfig</error>
  <errorcode>1</errorcode>
```

From `/var/log/xcat/cluster.log` the following error message is seen:
```
Jun 27 15:21:28 fs3 xcat[128856]: xCAT: Allowing getbmcconfig for mid05tor12cn02 from mid05tor12cn02
Jun 27 15:21:28 fs3 xcat[128859]: Unable to identify bmc for mid05tor12cn02, refusing to give config data
```

After some debug, I found the root cause is really that `mgt=ipmi` where `mgt=openbmc` should have been used.  So since from xCAT genesis, we set `<isopenbmc>1</isopenbmc>` We know this is an openbmc machine.   The message in the xcat.log does not help us debug what the real problem is, although we have enough information here to provide a better quality message.

Something like: 
> Received request from 'openbmc' host=mid05tor12cn02 but unable to determine it's BMC information from the 'openbmc' table.  Verify the mgmt configuration value for the node.

whowutwut:
Resolves #3356 

Improve the debug message when bmc value cannot be obtained from getbmcconfig
because the value is missing or the mgt= is incorrect for the machine being discovered. 

After the change the following message comes out in the cluster.log
```
Jun 27 16:18:12 fs3 xcat[130960]: xCAT: Allowing getbmcconfig for mid05tor12cn02 from mid05tor12cn02
Jun 27 16:18:12 fs3 xcat[130963]: Received request from host=mid05tor12cn02 but unable to determine openbmc.bmc value for the node.  Verify mgt attribute is configured correctly for the node and the BMC is defined.
```

Genesis message: 
```
  <error>Unable to detect BMC configuration value for bmcconfig</error>
  <errorcode>1</errorcode>
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcconfig.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcconfig.pm
@@ -105,7 +105,6 @@ sub process_request {
         return;
     }
 
-    #my $sitetable = xCAT::Table->new('site');
     my $ipmitable = xCAT::Table->new("$bmc_mgmt_type");
     my $tmphash;
     my $username;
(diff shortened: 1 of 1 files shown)


whowutwut:
@zet809 @robin2008 @hu-weihua I'd like to request that this goes into 2.13.5 because it will help with folks that maybe have a mix of ipmi and openbmc.  Would also apply to Boston nodes.

zet809:
Hi, @whowutwut , looks good to me, I agree to merge. Thx!

PR #3360, opened 2017-06-28
remove the log message "snooze for 1 second" in bmcsetup

zet809:


daniceexi:
@zet809 could give more description of why we need to make this change?

whowutwut:
@daniceexi , since @zet809 is on vacation so let me edit the description to provide some information about this PR.

daniceexi:
no problem. Thanks. @whowutwut

Issue #3368, opened 2017-06-29
noboot option does not seem to be working correctly to disable on the various interfaces 


whowutwut:
Opening this issue on behalf of @cganglin.  He mentioned to @cxhong  today that the noboot doesn't seem to be working and we investigated it a little.   I think there might be some problem.... 

Here's an example...
```
[root@fs3 ~]# ls -ltr /tftpboot/pxelinux.cfg/p/
total 16
-rw-r--r-- 1 root root 274 Jun 28 15:24 172.21.0.0_16
-rw-r--r-- 1 root root 274 Jun 28 15:24 172.20.0.0_16
-rw-r--r-- 1 root root 254 Jun 28 15:24 127.0.0.0_8
-rw-r--r-- 1 root root 264 Jun 28 15:24 10.0.0.0_8
```

```
[root@fs3 ~]# tabdump networks
#netname,net,mask,mgtifname,gateway,dhcpserver,tftpserver,nameservers,ntpservers,logservers,dynamicrange,staticrange,staticrangeincrement,nodehostname,ddnsdomain,vlanid,domain,mtu,comments,disable
"10_0_0_0-255_0_0_0","10.0.0.0","255.0.0.0","enP4p1s0f0","10.0.0.101",,"10.200.3.27",,,,,,,,,,,,,
"172_21_0_0-255_255_0_0","172.21.0.0","255.255.0.0","enP3p3s0d1","<xcatmaster>","172.21.253.27","172.21.253.27",,,,"172.21.253.100-172.21.253.200",,,,,,,,,
"172_20_0_0-255_255_0_0","172.20.0.0","255.255.0.0","enP3p3s0","<xcatmaster>","172.20.253.27","<xcatmaster>",,,,"172.20.253.100-172.20.253.200",,,,,,,"1500",,
```

So we have dhcp running on the 172.X network, 
```
[root@fs3 ~]# tabdump site | grep dhcpinter
"dhcpinterfaces","enP3p3s0,enP3p3s0d1",,
```

If i set noboot on enP3p3s0, it works.. if I set noboot on enP3p3s0d1, it does not, let's just set it on both for simple example..

```
[root@fs3 ~]# tabdump site | grep dhcpinter
"dhcpinterfaces","enP3p3s0:noboot,enP3p3s0d1:noboot",,

[root@fs3 ~]# mknb ppc64
Creating genesis.fs.ppc64.gz in /tftpboot/xcat
[...]

PR #3372, opened 2017-06-29
Fix 3368, noboot option does not seem to be working correctly to disable on the various interfaces

robin2008:
Merge fix for issue #3368 from master to 2.13


immarvin:
agree to merge

daniceexi:
Thanks for remember this.

We need to be sure that all the change in the release week must be cherry-pick to release branch. @robin2008 @zet809

Issue #3308, opened 2017-06-21
[openbmc] rspconfig - Support a way to send the ssh keys over to the BMC 


whowutwut:
Currently there's no REST command to upload a public key to the BMC. 

Rather than wait for this support, leverage the `rspconfig node sshcfg` option to use Expect for now to send over the ssh key so we can passwordless ssh to the BMC 

*Note* May need to create the `/root/.ssh` directory since that is missing from openbmc at the time of this writing

whowutwut:
This issue should be put on hold until we get confirmation on which milestone https://github.com/openbmc/openbmc/issues/1517 will be targeted for.  Should know more next week.

gurevichmark:
It appears that no REST API will be available in the near future, so for now will implement `scp`

gurevichmark:
Thinking of using `sshpass` command that should already be there. If not, can be installed with `yum -y install sshpass`. The `sshpass` allows passing of password on the command line. `BMC_PW` will be extracted from xCAT `passwd` table

1. Create a `.ssh` directory on BMC: 
   `sshpass -p <BMC_PW> ssh <BMC_IP> "mkdir -p ~/.ssh"`

2. Copy `id_rsa.pub` file from MN to BMC: 
   `sshpass -p <BMC_PW> scp ~/.ssh/id_rsa.pub <BMC_IP> ~/.ssh/id_rsa.pub"`

3. Append contents of copied `id_rsa.pub` to `authorized_keys`: 
   `sshpass -p <BMC_PW> ssh <BMC_IP> "cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys"`

4. Remove copied `id_rsa.pub`: 
   `sshpass -p <BMC_PW> ssh <BMC_IP> "rm -r ~/.ssh/id_rsa.pub`

whowutwut:
@gurevichmark So we don't have sshpass.  @cxhong  may have downloaded it...   I think if we really want to leverage more of this utility, we would want to consider shipping this in xcat-deps.  @chenglch any ideas about this?

whowutwut:
Maybe not in the normal OS 
```
[root@fs3 xcat]# yum install sshpass
Loaded plugins: product-id, search-disabled-repos, subscription-manager
This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.
No package sshpass available.
Error: Nothing to do
```

chenglch:
[...]

PR #3376, opened 2017-06-29
Copy ssh keys to BMC with rspconfig command

gurevichmark:
Implementing `rspconfig sshcfg` command to copy sshkeys to the BMC.
Issue #3308 

Success output:
```
[root@stratton01 xcat]# rspconfig p9euh02 sshcfg
p9euh02: ssh keys copied to 10.6.9.254
[root@stratton01 xcat]#
```

Invalid pw error output:
```
[root@stratton01 xcat]# rspconfig p9euh02 sshcfg
p9euh02: Error copying ssh keys to 10.6.9.254:
start SSH session...
Incorrect Password.

Warning: SSH failed, will try Telnet. Please set switches.protocol=telnet next time if you wish to use telnet directly.
start Telnet session...
problem connecting to "10.6.9.254", port 23: Connection refused

[root@stratton01 xcat]#
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -214,6 +214,13 @@ my %status_info = (
     RSPCONFIG_SET_RESPONSE => {
         process        => \&rspconfig_response,
     },
+    RSPCONFIG_SSHCFG_REQUEST => {
+        method         => "GET",
+        init_url       => "",
+    },
+    RSPCONFIG_SSHCFG_RESPONSE => {
+        process        => \&rspconfig_sshcfg_response,
+    },
     RVITALS_REQUEST => {
         method         => "GET",
         init_url       => "$openbmc_project_url/sensors/enumerate",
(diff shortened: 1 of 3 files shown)


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1351,6 +1367,56 @@ sub rspconfig_response {
 
 #-------------------------------------------------------
 
+=head3  rspconfig_sshcfg_response
```
whowutwut: Should we rename this to `rspconfig_sshcfg_response_expect` so that once we have a rest method, we would use this function for it and just deprecate the other one...  or maybe not necessary if we plan to just delete it all together.

whowutwut:
Ok let's get this code in and try it out

xCAT-server/lib/xcat/plugins/openbmc.pm
gurevichmark: Once we get the REST function, I will rename this one using "expect" one to something else. But till then, lets just keep this name.

Issue #3338, opened 2017-06-27
[openbmc]refine rsetboot based on latest RESTful interface and firmware


zet809:
Based on openbmc interface: https://github.com/openbmc/openbmc/issues/1548

PR #3378, opened 2017-06-30
Task 3338, refine rsetbot command for OpenBMC

xuweibj:
#3338 

Based on interface as below: (Wait for real interface, may need to modify)

To get boot source:
```
# curl -k -b cjar -X GET https://bmc/xyz/openbmc_project/control/boot/bootsource 
{"status": "ok", "message": "200 OK", "data": {“BootSource”: “xyz.openbmc_project.Control.Boot.Sources.Disk”}}
```
To set boot source:
```
# curl -k -b cjar -X PUT -H "Content-Type: application/json"  https://bmc/xyz/openbmc_project/control/boot/bootsource/attr/Sources/ -d '{"data": "xyz.openbmc_project.Control.Boot.Sources.Network”}’
{"status": "ok", "message": "200 OK"}
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -186,14 +186,15 @@ my %status_info = (
 
     RSETBOOT_SET_REQUEST => {
         method         => "PUT",
-        init_url       => "",
+        init_url       => "$openbmc_project_url/control/boot/bootsource/attr/Sources",
+        data           => "xyz.openbmc_project.Control.Boot.Sources.",
     },
     RSETBOOT_SET_RESPONSE => {
         process        => \&rsetboot_response,
     },
     RSETBOOT_STATUS_REQUEST  => {
         method         => "GET",
-        init_url       => "",
+        init_url       => "$openbmc_project_url/control/boot/bootsource",
     },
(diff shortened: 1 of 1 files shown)


whowutwut:
@xuweibj  Great! 👍  What happens if we execute it? it seems like the curl command returned 200 OK.  Does it take effect? If we set it to boot HDD and then we have a petitboot config file, would it ignore the network boot?

xuweibj:
@whowutwut @gurevichmark Now, we just get this interface, the code of firmware is not ready now. 
I think if we set it HD, it would ignore the network boot. 
For line 626 and 627, I think we should rm it when firmware is ready.

gurevichmark:
Ok, so this is not the final openbmc.pm change to support `rsetboot` ? When REST is ready, additional changes will be made to that file ?

whowutwut:
Let's see it today's code update exposes the out of band REST support for Network...

xuweibj:
@gurevichmark, I got this from firmware team. He said it seems right. But if the final REST is not same, we need to modify it again.

gurevichmark:
@xuweibj So with lines 626 and 627, `rsetboot` command can not be used ? Do you want to block `rsetboot` command until we have the final version of REST from firmware team ?

xuweibj:
@gurevichmark Yes, I want to keep it until the final REST.

PR #3380, opened 2017-06-30
Modify version to 2.13.6

tingtli:


hu-weihua:
it is good for me. thanks @tingtli

Issue #3347, opened 2017-06-27
[FVT] add test case for issue 1950: Setting the ipmi password incorrectly in the site table causes infinite loop in rpower: Unauthorized name


hu-weihua:
What to do:
* [ ] offer a test cases scope
* [ ] add test cases for issue #1950

junxiawang:
Add new testcase:
rpower_ipmi_wrongpasswd

whowutwut:
@hu-weihua @zet809 I'm still not convinced that we should be opening additional issues to track work on adding testcases that slipped through... .  I think by utilizing the "testcase:requested" and "testcase:added" labels, we can avoid having this extra issue in the middle.  The Pull request 3385 should be linking directly to 1950 and from 1950's perspective, we can see that a testcase was added directly instead of having this issue between....  1950 also contains all the information about the problem.  This intermediate issue does not and I think it's adding confusion

PR #3385, opened 2017-07-03
modify rpower testcase if ipmi using wrong passwd

junxiawang:
add test case for issue 1950: Setting the ipmi password incorrectly in the site table causes infinite loop in rpower: Unauthorized name #3347

### DIFF
--- a/xCAT-test/autotest/testcase/rpower/cases0
+++ b/xCAT-test/autotest/testcase/rpower/cases0
@@ -152,6 +152,34 @@ cmd:rpower $$CN onstandby
 cmd:a=0;while ! `rpower $$CN stat|grep "standby\|Standby" >/dev/null`; do sleep 5;((a++));if [ $a -gt 11 ];then break;fi done
 check:ouptut=~standby|Standby
 end
+start:rpower_suspend_OpenpowerBmc
+description:rpower openpowerbmc using suspend
+Attribute: $$CN-The operation object of rpower command
+cmd:rpower $$CN suspend
+check:output=~Error: unsupported command rpower suspend for OpenPOWER
+check:rc==1
+end
+start:rpower_softoff_OpenpowerBmc
+description:rpower openpowerbmc using softoff
+Attribute: $$CN-The operation object of rpower command
+cmd:rpower $$CN softoff
+check:output=~Error: unsupported command rpower softoff for OpenPOWER
+check:rc==1
+end
+start:rpower_wake_OpenpowerBmc
+description:rpower openpowerbmc using wake
+Attribute: $$CN-The operation object of rpower command
+cmd:rpower $$CN wake
+check:output=~Error: unsupported command rpower wake for OpenPOWER
+check:rc==1
+end
+start:rpower_errorcommand_OpenpowerBmc
+description:rpower openpowerbmc using errorcommand
+Attribute: $$CN-The operation object of rpower command
+cmd:rpower $$CN ddd
+check:output=~Unsupported command:
+check:rc==1
+end
 start:rpower_ipmi_wrongpasswd
 description:rpower ipmi and its password wrong
 Attribute: $$CN-The operation object of rpower command
--- a/xCAT-test/autotest/testcase/rpower/cases0
+++ b/xCAT-test/autotest/testcase/rpower/cases0
@@ -152,3 +152,13 @@ cmd:rpower $$CN onstandby
 cmd:a=0;while ! `rpower $$CN stat|grep "standby\|Standby" >/dev/null`; do sleep 5;((a++));if [ $a -gt 11 ];then break;fi done
 check:ouptut=~standby|Standby
 end
+start:rpower_ipmi_wrongpasswd
+description:rpower ipmi and its passwd wrong
+Attribute: $$CN-The operation object of rpower command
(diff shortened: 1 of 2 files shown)


xCAT-test/autotest/testcase/rpower/cases0
```diff
@@ -152,3 +152,13 @@ cmd:rpower $$CN onstandby
 cmd:a=0;while ! `rpower $$CN stat|grep "standby\|Standby" >/dev/null`; do sleep 5;((a++));if [ $a -gt 11 ];then break;fi done
 check:ouptut=~standby|Standby
 end
+start:rpower_ipmi_wrongpasswd
+description:rpower ipmi and its passwd wrong
+Attribute: $$CN-The operation object of rpower command
+cmd:tabdump passwd |grep ipmi>/tmp/ipmipasswd;Username=`cat /tmp/ipmipasswd |awk -F "\""  '{print $4}'`;Passwd=`cat /tmp/ipmipasswd |awk -F "\""  '{print $6}'`;`chtab key=ipmi passwd.password=$Passwd.wrong passwd.username=$Username`;tabdump passwd;
+check:rc==0
+cmd:rpower $$CN stat
+check:output=~Error: ERROR: Incorrect password provided
```
whowutwut: Is it enough to just check for "Error: ERROR:" so that if the message changes in the future, we don't need to update this testcase...

xCAT-test/autotest/testcase/rpower/cases0
```diff
@@ -152,3 +152,13 @@ cmd:rpower $$CN onstandby
 cmd:a=0;while ! `rpower $$CN stat|grep "standby\|Standby" >/dev/null`; do sleep 5;((a++));if [ $a -gt 11 ];then break;fi done
 check:ouptut=~standby|Standby
 end
+start:rpower_ipmi_wrongpasswd
+description:rpower ipmi and its passwd wrong
+Attribute: $$CN-The operation object of rpower command
+cmd:tabdump passwd |grep ipmi>/tmp/ipmipasswd;Username=`cat /tmp/ipmipasswd |awk -F "\""  '{print $4}'`;Passwd=`cat /tmp/ipmipasswd |awk -F "\""  '{print $6}'`;`chtab key=ipmi passwd.password=$Passwd.wrong passwd.username=$Username`;tabdump passwd;
```
whowutwut: I'd suggest that we make these temp filename more unique incase there is something left over or we can identify that we have left over due to autotest... something like 

`/tmp/xcat_test_case0_ipmipasswd`

Can this be a variable or needs to be hard coded?

xCAT-test/autotest/testcase/rpower/cases0
```diff
@@ -152,3 +152,13 @@ cmd:rpower $$CN onstandby
 cmd:a=0;while ! `rpower $$CN stat|grep "standby\|Standby" >/dev/null`; do sleep 5;((a++));if [ $a -gt 11 ];then break;fi done
 check:ouptut=~standby|Standby
 end
+start:rpower_ipmi_wrongpasswd
+description:rpower ipmi and its passwd wrong
```
whowutwut: Suggest  for description, use "password" instead of "passwd"

xCAT-test/autotest/testcase/rpower/cases0
whowutwut: Do we set the password in a previous test case?  Otherwise, this would fail if a brand new install of xCAT and ipmi is not set.    What happens if this testcase is run against something like openbmc... where the keyword is now `openbmc` instead of `ipmi`?

junxiawang: But we need to add new file:/tmp/xcat_test_case0_ipmipasswd in code.

junxiawang: If brand new test so the test may be not know the passwd and username and I do not think this case should cover brand new testcase.

xCAT-test/autotest/testcase/rpower/cases0
junxiawang: I do not know the changed messages in future may be we could add more message to check

xCAT-test/autotest/testcase/rpower/cases0
whowutwut: Why is `/tmp/ipmipasswd` added in code? it's doing `tabdump passwd | grep ipmi > /tmp/ipmipasswd`    It is redirecting to a file in `/tmp`, doesn't need to be shipped with xCAT-test rpm.  

The name of the file is unique so we know that it was xcat-test that created the temporary file. 
 You delete it later one, but if someone else copy this test case, and forgot to delete, we can identify this file as something generated by xCAT-test testcase

whowutwut: @hu-weihua  Why are we not installing xCAT on a brand new server when we are running the daily regression?  We should not re-use environment or we will not catch issues on a new install.   For this testcase, we should first set the ipmi password, verify that it is correct and we can get `rpower stat` , then set it incorrect and verify that we cannot.  Then we should also override the node definition with `bmcusername` and `bmcpassword` and verify that we can communicate to the node...

junxiawang: sorry my mistake about this comments

whowutwut:
@junxiawang  Some conflicts here that require manual resolution

whowutwut:
@junxiawang @hu-weihua  Should we use some different name for this file? 
Instead of `xCAT-test/autotest/testcase/rpower/cases0` we can call it something like `xCAT-test/autotest/testcase/rpower/openbmc_mgt_case0`.  This way we have the ability to select cases into bundles that are only applicable to certain management types without having to view the case0 file and decide depending on that....

whowutwut:
@junxiawang Please resolve the conflicts and I will take a review again.

junxiawang:
pull request again in 3443

PR #3389, opened 2017-07-04
"rinstall xxx" command should be valid and perform provision by default

immarvin:
``rinstall <node>`` should pass the syntax check, the default behavior equals ``rinstall <node> osimage``

### DIFF
--- a/perl-xCAT/xCAT/Usage.pm
+++ b/perl-xCAT/xCAT/Usage.pm
@@ -4,7 +4,6 @@
 package xCAT::Usage;
 use Getopt::Long;
 use xCAT::Utils;
-
 #-------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


daniceexi:
@immarvin I only saw the change for usage, do not we need to change some code?

immarvin:
hi @daniceexi ,  the subroutine xCAT::Usage::validateArgs is introduce for some pre syntax check, xCAT::Usage::validateArgs will take ``rinstall <node>`` invalid and complain, in fact,   ``rinstall <node>`` is supported before the PR https://github.com/xcat2/xcat-core/pull/3147/files

daniceexi:
@immarvin get it, thanks.

zet809:

Issue #3345, opened 2017-06-27
[FVT] add test case for issue 2948: rpower commands run against OpenPower BMC should be blocked if not supported


hu-weihua:
What to do:
* [x] offer a test cases scope
* [x] add test cases for issue #2948

junxiawang:
Need to add testcase:
rpower_softoff_OpenpowerBmc
rpower_wake_OpenpowerBmc
rpower_suspend_OpenpowerBmc
rpower_errorcommand_OpenpowerBmc

PR #3400, opened 2017-07-05
modify rpower testcase using unsurpported command for openpowerbmc

junxiawang:
add test case for issue 2948: rpower commands run against OpenPower BMC should be blocked if not supported #3345


DengShuaiSimon:
``CI CHECK RESULT`` : > **PR FORMAT CORRECT**

whowutwut:
Just a general comment, is there a way to have a list of commands and we just loop over it to generate these tests?  It seems like if we supported options in the future, we have to modify this instead of just removing a command from an array list.

junxiawang:
May be we should add the list using bundle

xuweibj:
Looks good to me. I will merge it.

whowutwut:
👍  @junxiawang  

I think we should have bundle lists for each type of management.  Instead of `testcase/rpower/cases0` , we should be able to tell from the name that it's openbmc test cases.. 


`testcase/rpower/openbmc0` ... Can we make that change? 

Otherwise, how do we control which test case will be executed depending on the machine types that we have? We have to look inside the files? That is not good..

PR #3407, opened 2017-07-06
Change bmcdiscover for openbmc to look at Part Number

whowutwut:
In latest ITC machine, the Model Type is in the PartNumber Field

```
[root@fs3 UT]# time curl -c cjar -b cjar -k -H "Content-Type: application/json" -X GET https://$IP/xyz/openbmc_project/inventory/system
{
  "data": {
    "BuildDate": "",
    "Cached": 0,
    "FieldReplaceable": 0,
    "Manufacturer": "",
    "Model": "2",
    "PartNumber": "8335-GTC",
    "Present": 1,
    "PrettyName": "",
    "SerialNumber": "1318C4A"
  },
  "message": "200 OK",
  "status": "ok"
}
real    0m0.386s
user    0m0.109s
sys     0m0.030s
```

This is still a problem with discovery to use this field because it's not AVAIL until the first time the host side boots...

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -1086,8 +1086,8 @@ sub bmcdiscovery_openbmc{
         my $serial;
 
         if (defined($response->{data})) {
-            if (defined($response->{data}->{Model}) and defined($response->{data}->{SerialNumber})) {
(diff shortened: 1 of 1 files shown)


whowutwut:
FYI @zet809 and @xuweibj , I didn't assign someone specifically to review so assign to @gurevichmark so that it's merged in today.  This is not really a solution for `bmcdiscover` because this information does not get populated until after the first IPL host boot.

PR #3412, opened 2017-07-06
Remove the plain text of user name and password of xcatbot

hu-weihua:
The purpose of this pull request is to remove the plain text of user name and password of xcatbot in CI log.


daniceexi:
OK for merge.

PR #3418, opened 2017-07-07
Support hybrid mode for different access vlan other than 1

cxhong:
Change mode to hybrid will change access vlan back to 1.  To avoid this,  the configMellanox scripts to setup vlan is required user pass in both access vlan and allowed vlan

Failed case:
```
[root@fs3 ~]# /opt/xcat/share/xcat/scripts/configMellanox --switches core01 --port 58 --vlan 30 --mode access
Tagging VLAN=30 for core01 port 58
```
after that, looks like this `Eth1/58         access       30`

```
[root@fs3 ~]# /opt/xcat/share/xcat/scripts/configMellanox --switches core01 --port 58 --vlan 31 --mode hybrid
Tagging VLAN=31 for core01 port 58
```
looks like this : `Eth1/58         hybrid       1                  31`

After script modification:
---------------------------
```
[root@fs3 ~]# /opt/xcat/share/xcat/scripts/configMellanox  --switches core01 --port 56 --accessvlan 20 --allowvlan 21 --mode hybrid
Tagging access vlan to 20 and allowed vlan to 21 with hybrid mode for core01 port 56
```
the vlan on port 56 will looks like this: `Eth1/56         hybrid       20                 21`

### DIFF
--- a/xCAT-server/share/xcat/scripts/configMellanox
+++ b/xCAT-server/share/xcat/scripts/configMellanox
@@ -43,7 +43,8 @@ if (
                 'ip'         => \$::IP,
                 'name'       => \$::NAME,
                 'snmp'       => \$::SNMP,
-                'vlan=s'     => \$::VLAN,
+                'accessvlan=s' => \$::PVID,
+                'allowvlan=s'  => \$::VID,
                 'port=s'     => \$::PORT,
                 'mode=s'     => \$::MODE,
                 'all'        => \$::ALL,
@@ -120,7 +121,7 @@ if (($::CONFIG) || ($::ALL)) {
     run_rspconfig();
 }
 
-if ($::VLAN) {
+if ( ($::PVID) || ($::VID) ) {
     config_vlan();
 }
 
@@ -363,6 +364,9 @@ sub config_vlan {
         exit(1);
     }
 
+    my $access_vlan = $::PVID;
+    my $allowed_vlan = $::VID;
+
     # will default to trunk mode
     if ($::MODE) {
        $mode = $::MODE;
@@ -372,11 +376,18 @@ sub config_vlan {
             &usage;
             exit(1);
        }
+       if ($mode =~ /hybrid/) 
+       {
+           if (!$access_vlan) {
+               print "NOTE: Hybrid mode will change access vlan back to 1\n"; 
+               print "please provide access vlan if other than 1\n"; 
+               $access_vlan = 1;
+           }
+       }
     } else {
        $mode = "access";
     }
 
-    $vlan = $::VLAN;
 
     foreach my $switch (@nodes) {
         my $devicetype;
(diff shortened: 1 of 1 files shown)


xCAT-server/share/xcat/scripts/configMellanox
```diff
@@ -372,11 +376,18 @@ sub config_vlan {
             &usage;
             exit(1);
        }
+       if ($mode =~ /hybrid/) 
+       {
+           if (!$access_vlan) {
+               print "NOTE: Hybrid mode will change access vlan back to 1\n";
```
whowutwut: Can we change `vlan` on print statements to `VLAN` in CAPS.   Why does it change it back to 1, is that a default?  What if we changed the default to something else...  would it change it back to 1 or the previous TRUNK VLAN?

xCAT-server/share/xcat/scripts/configMellanox
```diff
@@ -372,11 +376,18 @@ sub config_vlan {
             &usage;
             exit(1);
        }
+       if ($mode =~ /hybrid/) 
+       {
+           if (!$access_vlan) {
+               print "NOTE: Hybrid mode will change access vlan back to 1\n"; 
+               print "please provide access vlan if other than 1\n";
```
whowutwut: Let's not use "Please" in the output message.    How does the user call this?  Maybe we can print an example?  

If the user wanted to do Hybrid VLAN 11, and Access VLAN 10, how does the user call this?

xCAT-server/share/xcat/scripts/configMellanox
```diff
@@ -394,23 +405,31 @@ sub config_vlan {
             next;
         }
 
-        print "Tagging VLAN=$vlan for $switch port $port_input\n";
-
-        # create vlan
-        my $vlan_cmd = `xdsh $switch --devicetype $devicetype "enable;configure terminal;vlan $vlan;exit;exit" `;
+        print "Tagging access vlan to $access_vlan and allowed vlan to $allowed_vlan with $mode mode for $switch port $port_input\n";
```
whowutwut: Suggest: 

`access vlan` to `access VLAN` 
`allowed vlan` to `allowed VLAN`

xCAT-server/share/xcat/scripts/configMellanox
```diff
@@ -445,10 +464,10 @@ sub usage
         configMellanox --switches switchnames --config  
 
     To configure VLAN on a specified port (Mellanox Ethernet switch ONLY):
-        configMellanox --switches switchnames --port port --vlan vlan --mode mode
+        configMellanox --switches switchnames --port port --accessvlan vlan1 --allowvlan vlan2 --mode mode
```
whowutwut: Do we need to describe the new `--allowvlan` option somewhere in the usage?

xCAT-server/share/xcat/scripts/configMellanox
cxhong: somehow, it changed back to 1.  if change access vlan to something else,  after hybird, it will change to 1 again.

xCAT-server/share/xcat/scripts/configMellanox
cxhong: ```configMellanox --switches switchnames --port port --accessvlan vlan1 --allowvlan vlan2 --mode mode```

xCAT-server/share/xcat/scripts/configMellanox
cxhong: changed

xCAT-server/share/xcat/scripts/configMellanox
cxhong: no.  I think accessvlan and allowvlan should descriptor the meaning already.

Issue #3414, opened 2017-07-06
[CORAL]incorrect petitboot configuration for nodeset shell in flat hierarchical  


zet809:
The xcatd in petitboot configuration file shall be in the same subnet with CN.
[root@fs3 xcat]# lsdef mid05tor12cn04 -i ip -c
mid05tor12cn04: ip=172.20.140.4
[root@fs3 ~]# cat /tftpboot/petitboot/mid05tor12cn04
#shell
default xCAT Genesis shell
label xCAT Genesis shell
	kernel http://172.20.253.27:80/tftpboot/xcat/genesis.kernel.ppc64
	initrd http://172.20.253.27:80/tftpboot/xcat/genesis.fs.ppc64.gz
	append "quiet xcatd=172.21.253.27:3001 destiny=shell"  ===> The xcatd shall be 172.20.x.x
[root@fs3 ~]# lsdef -t site -i master
Object name: clustersite
    master=172.21.253.27

whowutwut:
@zet809  Let me sync up with @cxhong  today and look at the end to end steps.  Is this to avoid the need to set the master node attribute for the node definitions since we have different operating networks?   In reality, it should not be the master but <xcatmaster> for the network.. could be service node.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3419

PR #3419, opened 2017-07-10
fix issue [CORAL]incorrect petitboot configuration for nodeset shell in flat hierarchical #3414

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3414:

the ``noderes.xcatmaster`` takes precedence, then the ip address facing the node, the ``site.master`` takes last precedence


robin2008:
@immarvin What's the behavior if there are two IPs facing to CN? if one of them are site.master, better to use it.

PR #3425, opened 2017-07-10
Support rpower softoff for openbmc 

### LINKED ISSUE #3422: OpenBMC support for softoff breaks xCAT rpower command
In recent OpenbBMC builds, the activation of softoff support for controlling the servers breaks xCAT rpower support.  

The following needs to be done, and tracking with this issue.  
1)  Change the rpower off to be "hard off" to the chassis instead of the host 
2)  rpower stat needs to be aware of the chassis power off 
```
}[root@fs3 mid05tor12cn04]# curl -c cjar -b cjar -k -H "Content-Type: application/json" -X GET https://$IP/xyz/openbmc_project/state/enumerate
{
  "data": {
    "/xyz/openbmc_project/state/bmc0": {
      "CurrentBMCState": "xyz.openbmc_project.State.BMC.BMCState.Ready",
      "RequestedBMCTransition": "xyz.openbmc_project.State.BMC.Transition.None"
    },
    "/xyz/openbmc_project/state/chassis0": {
      "CurrentPowerState": "xyz.openbmc_project.State.Chassis.PowerState.Off",
      "RequestedPowerTransition": "xyz.openbmc_project.State.Chassis.Transition.Off"
    },
    "/xyz/openbmc_project/state/host0": {
      "CurrentHostState": "xyz.openbmc_project.State.Host.HostState.Running",
      "RequestedHostTransition": "xyz.openbmc_project.State.Host.Transition.Reboot"
    }
  },
  "message": "200 OK",
  "status": "ok"
```
However rpower state is not looking at chassis state and returns "on" 
```
[root@fs3 mid05tor12cn04]# rpower mid05tor12cn04 state
mid05tor12cn04: on
```

3) Open up the `softoff` function for rpower and implement this

The time of softoff gives the Host 5 minutes to power off, before it sends the hard power off (integrated 6/25 via https://gerrit.openbmc-project.xyz/#/c/4895/) so depending on the firmware levels, it's either 5 minutes or 45 minutes. 

Asked some questions here: https://github.com/openbmc/openbmc/issues/1933

whowutwut:
Partially resolves #3422 

Based on the [README](https://github.com/openbmc/phosphor-dbus-interfaces/blob/master/xyz/openbmc_project/State/README.md) for the State API for OpenBMC,

Hardoff should be done against the chassis0 object and softoff is done against the host0 object.   The xCAT  rpower command had this backwards but prior to 1724B firmware, but sending OFF status on the host object powered off the chassis , so we didn't see an issue at the time.

### DIFF
--- a/docs/source/guides/admin-guides/references/man1/rpower.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/rpower.1.rst
@@ -45,7 +45,7 @@ OpenPOWER OpenBMC:
 ==================
 
 
-\ **rpower**\  \ *noderange*\  [\ **off | on | reset | boot | stat | state | status**\ ]
+\ **rpower**\  \ *noderange*\  [\ **off | on | softoff | reset | boot | stat | state | status**\ ]
 
 
 PPC (with IVM or HMC) specific:
--- a/xCAT-client/pods/man1/rpower.1.pod
+++ b/xCAT-client/pods/man1/rpower.1.pod
@@ -22,7 +22,7 @@ B<rpower> I<noderange> [B<pduon>|B<pduoff>|B<pdustat>|B<pdureset>]
 
 =head2 OpenPOWER OpenBMC:
 
-B<rpower> I<noderange> [B<off>|B<on>|B<reset>|B<boot>|B<stat>|B<state>|B<status>]
+B<rpower> I<noderange> [B<off>|B<on>|B<softoff>|B<reset>|B<boot>|B<stat>|B<state>|B<status>]
 
 =head2 PPC (with IVM or HMC) specific:
 
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -168,6 +168,11 @@ my %status_info = (
     RPOWER_OFF_RESPONSE => {
         process        => \&rpower_response,
     },
+    RPOWER_SOFTOFF_REQUEST  => {
+        method         => "PUT",
+        init_url       => "$openbmc_project_url/state/host0/attr/RequestedHostTransition",
+        data           => "xyz.openbmc_project.State.Host.Transition.Off",
+    },
     RPOWER_RESET_REQUEST  => {
         method         => "PUT",
         init_url       => "$openbmc_project_url/state/host0/attr/RequestedHostTransition",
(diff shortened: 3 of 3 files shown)


whowutwut:
@xuweibj This code has not really been exercised,  If you want me to test this out a little before we merge, that is OK,  Or since we have been issuing power off on the node using this, it should work.  

I think some of the states would report differently.   If we soft off, it should go into "powering-off" for at least 5 minutes to allow the OS to shutdown.  I know there has been some issues against the openbmc project when the Host is shutting down, so we might want to just merge this to allow function into xCAT so that we can exercise this more in the upcoming weeks.

xuweibj:
Hi @whowutwut ,  The code is OK.
I have tested it. 
After command ``# rpower mid05tor12cn04 softoff``
```
# XCATBYPASS=1 rpower mid05tor12cn04 stat
mid05tor12cn04: DEBUG - State CurrentPowerState=xyz.openbmc_project.State.Chassis.PowerState.Off
mid05tor12cn04: DEBUG - State RequestedPowerTransition=xyz.openbmc_project.State.Chassis.Transition.Off
mid05tor12cn04: DEBUG - State CurrentHostState=xyz.openbmc_project.State.Host.HostState.Off
mid05tor12cn04: DEBUG - State RequestedHostTransition=xyz.openbmc_project.State.Host.Transition.Off
mid05tor12cn04: off
```
After command ``# rpower mid05tor12cn04 off``
```
# XCATBYPASS=1 rpower mid05tor12cn04 stat
mid05tor12cn04: DEBUG - State CurrentPowerState=xyz.openbmc_project.State.Chassis.PowerState.Off
mid05tor12cn04: DEBUG - State RequestedPowerTransition=xyz.openbmc_project.State.Chassis.Transition.Off
mid05tor12cn04: DEBUG - State CurrentHostState=xyz.openbmc_project.State.Host.HostState.Running
mid05tor12cn04: DEBUG - State RequestedHostTransition=xyz.openbmc_project.State.Host.Transition.On
mid05tor12cn04: off
```

Did you meet that wait at least 5 minutes to shutdown? But I got it less than 1 minute.

whowutwut:
@xuweibj the softoff should invoke the OS shutdown. The firmware we have installed has a 45 min time out but in the newer firmware it waits 5 minutes to allow the OS to shutdown before it does a hard off.  It could only take 1 min. Would have to have a console to the node and verify the OS does a graceful shutdown.   Rpower off now immediately pulls power to the host

daniceexi:
if it's a graceful shutdown, there should be specific logs in the syslog of the host.

PR #3427, opened 2017-07-11
refine confignetwork cases

tingtli:
Refine confignetwork cases.
1) fix failed match for MTU
2)remove checkrc=0 for the environment recovery since it is not for result check


bybai:
I am OK to merge it.

Issue #3150, opened 2017-05-30
xCAT should better support the service group (servicenode postscript) to avoid errors when using Hierarchy


whowutwut:
With hierarchy support and the configuration of Service Nodes, the service group is a special group for xCAT, yet the users are able to fall into many configuration traps and prone to user errors.  

Let me try and describe some problems: 

### Issue:  If node is not added to "service" group, `servicenode` postscript not run 

Create a fake node: 
```
[root@fs2vm110 ~]# mkdef sn01 groups=all netboot=petitboot mac=00:00:00:00:00:00
1 object definitions have been created or modified.
[root@fs2vm110 ~]# lsdef sn01
Object name: sn01
    groups=all
    mac=00:00:00:00:00:00
    netboot=petitboot
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
```

nodeset a "service node" osimage and see what happens....

```
[root@fs2vm110 ~]# nodeset sn01 osimage=rhels7.3-ppc64le-install-service
sn01: install rhels7.3-ppc64le-service
Error: The sn01 can not be resolved
Error: dhcp server is not running.  please start the dhcp server.
[root@fs2vm110 ~]# lsdef sn01
Object name: sn01
    arch=ppc64le
    currchain=boot
    currstate=install rhels7.3-ppc64le-service
    groups=all
    mac=00:00:00:00:00:00
    netboot=petitboot
    os=rhels7.3
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    profile=service
    provmethod=rhels7.3-ppc64le-install-service
```

The profile is set to `service` in the node definition, but the `servicenode` postscript is not added to this node definition, causing errors when provisioning service nodes. 

### Issue: service group can be deleted, once deleted it does not work... 

Multiple Issues here.... not sure how to re-create it..... seems like once the service group is deleted, it doesn't behave well anymore. 

Create a group service, and servicenode postscript is there... 
```
[root@fs2vm110 ~]# lsdef -t group -o service
Error: Could not find an object named 'service' of type 'group'.
[root@fs2vm110 ~]# mkdef -t group -o service 
Warning: Cannot determine a member list for group 'service'.
1 object definitions have been created or modified.
[root@fs2vm110 ~]# lsdef -t group -o service
Object name: service
    grouptype=static
    members=
    postscripts=servicenode
```

Once we remove it, can't re-create it...

```
[...]

PR #3442, opened 2017-07-12
fix xCAT should better support the service group (servicenode postscript) to avoid errors when using Hierarchy #3150

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3150:

add postscript "servicenode"  to the postscripts attributes of ``xxxx-service-xxxx`` osimages

Verified with sles12.2 and rhels7.3:
````
[root@c910f03c05k21 ~]# lsdef -t osimage sles12.2-ppc64le-install-service
Object name: sles12.2-ppc64le-install-service
    imagetype=linux
    osarch=ppc64le
    osdistroname=sles12.2-ppc64le
    osname=Linux
    osvers=sles12.2
    otherpkgdir=/install/post/otherpkgs/sles12.2/ppc64le
    otherpkglist=/opt/xcat/share/xcat/install/sles/service.sles12.ppc64le.otherpkgs.pkglist
    pkgdir=/install/sles12.2/ppc64le
    pkglist=/opt/xcat/share/xcat/install/sles/service.sles12.pkglist
    **postscripts=servicenode**
    profile=service
    provmethod=install
    template=/opt/xcat/share/xcat/install/sles/service.sles12.tmpl
````

### DIFF
--- a/xCAT-server/lib/perl/xCAT/SvrUtils.pm
+++ b/xCAT-server/lib/perl/xCAT/SvrUtils.pm
@@ -742,6 +742,13 @@ sub update_tables_with_templates
                     osarch       => $arch,
                     synclists    => $synclistfile,
                     osdistroname => $osdistroname);
+
+                #for service node osimage, add service node to the postscripts attributes
+                if($profile eq "service"){
+                    $tb_cols{postscripts}="servicenode";
+                }
+
+
                 if ($args{description}) {
                     $tb_cols{description} = $args{description};
                 }


immarvin:
The postage.pm has processed the duplicate PS or PBS in node.postscript and osimage.postscript, so when a node has been assigned to the "service" group, and it is associated with a "xxx-service-xxx" osimage generated by copycds, the "servicenode" PS will only be run once. I think it is ok

daniceexi:
Even the duplicate has been handled, we still need to consider the order changes when remove the duplicated one.

immarvin:
hi @daniceexi , the code logic to generate mypostscript is in https://github.com/xcat2/xcat-core/blob/master/xCAT-server/lib/perl/xCAT/Postage.pm#L1914, the PS in osimage.postscripts will be first appended to the mypostscript, then the node.postscripts will be appended, if the "servicenode" PS has been processed as  osimage.postscripts, it will not be appended to mypostscript.

Issue #1950, opened 2016-10-10
Setting the ipmi password incorrectly in the site table causes infinite loop in rpower: Unauthorized name


whowutwut:
Error injection, a wrong password into the `passwd` table for the ipmi row causes rpower to loop with `Unauthorized name`.    I don't know if it ever comes back but for the 30seconds to 1 minute that I left it, it keeps printing the same msg. 

```
[root@fs1 ~]# tabdump passwd | grep ipmi
"ipmi","USERID","PASSW0RD",,,,
```

```
[root@fs1 ~]# rpower frame45cn08 stat
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
frame45cn08: Error: Unauthorized name
...
```

Setting the `bmcusername` to the node allows for more expected behavior with the wrong password:
```
[root@fs1 ~]# rpower frame45cn08 stat
frame45cn08: Error: Incorrect password provided
```

@tingtli FYI, Please look into creating some error injection testcases for all hw control commands

zet809:
@penguhyang Hi, Yang peng, will you pls take a look at this issue?

whowutwut:
Verified using the 10/25 build of xCAT

* Wrong user/password
```
[root@fs4 ~]# tabdump passwd | grep ipmi
"ipmi","USERID","PASSW0RD",,,,
[root@fs4 ~]# rpower frame23cn18 stat
frame23cn18: Error: Unauthorized name
```

* Wrong password
```
[root@fs4 ~]# tabdump passwd | grep ipmi
"ipmi","*****","PASSW0RD",,,,
[root@fs4 ~]# rpower frame23cn18 stat
[...]

PR #3443, opened 2017-07-12
modify rpower testcase if ipmi using wrong passwd

junxiawang:
add test case for issue #1950: Setting the ipmi password incorrectly in the site table causes infinite loop in rpower: Unauthorized name #3347


whowutwut:
@junxiawang Please add the `#` before the issue numbers so that GitHub can create the references.

whowutwut:
Also, please fill in the right side bar, milestone, and labels...

xCAT-test/autotest/testcase/rpower/cases0
```diff
@@ -180,4 +180,16 @@ cmd:rpower $$CN ddd
 check:output=~Unsupported command:
 check:rc==1
 end
+start:rpower_ipmi_wrongpasswd
```
whowutwut: The file name for this is `xCAT-test/autotest/testcase/rpower/cases0` 

What bundles are these included in?    If this is a generic rpower test case, then `start:rpower_ipmi_wrongpasswd` should not be included in this file.  Because what if we wanted to run this on `mgt=openbmc`, or `mgt=fsp`   or `mgt=hmc`?

whowutwut: Currently this is probably a testcase applicable for ipmi/openbmc, so my recommendation is to create two testcase file... 

`testcase/rpower/case1_ipmi` 
`testcase/rpower/case2_openbmc` 

In each one, we will call the rpower test script and pass in the mgmt time, see comments below.  The mgmt type will then be a variable so we can re-use this test for both ipmi and openbmc.

xCAT-test/autotest/testcase/rpower/rpower_ipmi_wrongpasswd_test.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
```
whowutwut: Suggest name change for this file to be `rpower_wrongpassword_test.sh`  and pass in the option for `ipmi` or `openbmc`   .  Then all the references to "ipmi" should be a variable so that we can re-use this test to test both openbmc and ipmi.....

xCAT-test/autotest/testcase/rpower/rpower_ipmi_wrongpasswd_test.sh
```diff
@@ -0,0 +1,118 @@
...
+function check_passwd_table(){
+tabdump passwd |grep ipmi
+    if [ $? -eq 0 ];then
+        `tabdump passwd |grep ipmi>/tmp/xcat_test_rpower_ipmi_wrongpasswd`
+    else
+        return 1;
+    fi
+}
+function modify_passwd_table(){
+Username=`cat /tmp/xcat_test_rpower_ipmi_wrongpasswd |awk -F "\""  '{print $4}'`;
```
whowutwut: I think we should dynamically generate these files based on the name of the file instead of hard coding a path...  

Here's an example:

```
[root@c910f02c05p02 ~]# touch testcase/rpower/rpower_wrongpassword_test.sh
[root@c910f02c05p02 ~]# vi testcase/rpower/rpower_wrongpassword_test.sh
[root@c910f02c05p02 ~]# vi testcase/rpower/rpower_wrongpassword_test.sh
[root@c910f02c05p02 ~]# chmod +x testcase/rpower/rpower_wrongpassword_test.sh

[root@c910f02c05p02 ~]# cat testcase/rpower/rpower_wrongpassword_test.sh
#!/bin/bash

echo "0 ARG: $0"

SCRIPT=$(readlink -f $0)
echo "SCRIPT=$SCRIPT"

TMPFILE="/tmp/xcat-test-`basename $SCRIPT`.tmp"
echo "TMPFILE=$TMPFILE"

```

Running it from /root....
```
[root@c910f02c05p02 ~]# testcase/rpower/rpower_wrongpassword_test.sh
0 ARG: testcase/rpower/rpower_wrongpassword_test.sh
SCRIPT=/root/testcase/rpower/rpower_wrongpassword_test.sh
TMPFILE=/tmp/xcat-test-rpower_wrongpassword_test.sh.tmp

```

Running it from rpower directory 
```
[root@c910f02c05p02 ~]# cd testcase/rpower/
[root@c910f02c05p02 rpower]# ./rpower_wrongpassword_test.sh
0 ARG: ./rpower_wrongpassword_test.sh
SCRIPT=/root/testcase/rpower/rpower_wrongpassword_test.sh
TMPFILE=/tmp/xcat-test-rpower_wrongpassword_test.sh.tmp
```

xCAT-test/autotest/testcase/rpower/rpower_ipmi_wrongpasswd_test.sh
junxiawang: I have modify code ,the test case could cover both ipmi and openbmc

xCAT-test/autotest/testcase/rpower/cases0
junxiawang: The testcase will not be supported for fsp adn hmc:
1.For fsp will be dropped in future as talked with ertao.
2.For mgt=hmc ,the management for lpar need hmc node to handle and so username and passwd need to be wrote in hmc node definition.
So the testcase here only support ipmi and openbmc.

PR #3450, opened 2017-07-13
To support hierarchy for openbmc

xuweibj:
1. Set ``$callback`` value in process_request function.
2. Get subcommand by ``$request->{arg}`` instead of ``@ARGV`` in parse_command_status function.


zet809:

PR #3454, opened 2017-07-14
Fix test case mkdef_regex_kvm

neo954:
Fixed the wrongly spelled hostname.


hu-weihua:
It is good for me, agree to merge. thanks @neo954

PR #3456, opened 2017-07-14
Fix #3088, Using 'disjointdhcps' site attribute to make nodeset request will only be handled on service node for nodes it manages 

### LINKED ISSUE #3088: [xCAT performance] nodeset improvement - define new method to reduce the managed node number under service nodes
Now for nodeset in hierarchical topology (sharedftp=0) , it will broadcast the request to service nodes,  and service node will handle all the nodes which are in the same network. So for a CORAL like flat networking, the SNs need handle all node range per requesting. However, the execution on SN has bad performance than it run on MN as the remote accessing DB and DNS.  

This is an idea to let user define a 'disjoint' mode, in such mode, the SN only care about its direct children (explicitly defined in node's "servicenode"). Then we can balance the workload to SN even in a flat networking. For MGMT node whose dhcp server is serving dynamic range,  it is required to  handle all requesting noderange as it may already have the dynamic lease information for all nodes(even if the CN is not managed by this server).

Note:
Here is only providing a quick check for not explicitly managed CNs on SN. The CN will still  be checked again to see if the CN is on the same network with current SN. If not, the CN will be ignored again.

In a separate subnet  hierarchical cluster, it won't bring big performance improvement as only same network  checking is omitted. But for FLAT networking cluster, the whole nodeset processing are omitted for unmanaged nodes.

robin2008:
Fix #3088
Now for nodeset in hierarchical topology (sharedftp=0) , it will broadcast the request to service nodes, and service node will handle all the nodes which are in the same network. So for a flat networking, the SNs need handle all node range per requesting.

This is an idea to let user define a 'strict' mode, in such mode, the SN only cares about its direct children (explicitly defined in node's "servicenode"). Then we can balance the workload to SN even in a flat networking. For MGMT node,  it still handle all requesting nodes.

How to enable this mode:
```
chdef -t site sharedtftp=0 disjointdhcps=1
```

Note:
Here is only providing a quick check for not explicitly managed CNs on SN. The CN will still  be checked again to see if the CN is required to be handled by checking if it is on the same network with the SN. If not, the CN will be ignored when sharedtftp is disabled.

In a separate subnet  hierarchical cluster, it won't bring big performance improvement as only same network  checking is omitted. But for FLAT networking cluster, the whole nodeset processing are omitted for unmanaged nodes.

### DIFF
--- a/perl-xCAT/xCAT/Scope.pm
+++ b/perl-xCAT/xCAT/Scope.pm
@@ -1,8 +1,8 @@
 package xCAT::Scope;
 
-#use Data::Dumper;
 use xCAT::Utils;
 use xCAT::Table;
+use xCAT::TableUtils;
 use xCAT::ServiceNodeUtils qw(getSNList);
 
 
@@ -148,22 +148,31 @@ sub get_broadcast_scope_with_parallel {
     if ($req->{_xcatpreprocessed}->[0] == 1) { return [$req]; }
 
     my $snlist = shift;
+    my @reqs = ();
+    $req->{_xcatpreprocessed}->[0] = 1;
+    my $isMN = xCAT::Utils->isMN();
     #Handle the one for current management/service node
-    my $reqs = get_parallel_scope($req); 
+    if ( $isMN ) {
+        $reqs = get_parallel_scope($req);
+    } else {
+        # get site.master and broadcast to it as MN will not be in SN list or defined in noderes.servicenode.
+        my @entries = xCAT::TableUtils->get_site_attribute("master");
+        my $master = $entries[0];
+        my $reqcopy = {%$req};
+        $reqcopy->{'_xcatdest'} = $master;
+        $reqs = get_parallel_scope($reqcopy);
+    }
     my @requests = @$reqs;
 
     #Broadcast the request to other management/service nodes
     foreach (@$snlist) {
         my $xcatdest = $_;
-        if (xCAT::NetworkUtils->thishostisnot($xcatdest)) {
-            my $reqcopy = {%$req};
-            $reqcopy->{'_xcatdest'} = $xcatdest;
-            $reqcopy->{_xcatpreprocessed}->[0] = 1;
+        my $reqcopy = {%$req};
+        $reqcopy->{'_xcatdest'} = $xcatdest;
 
-            $reqs = get_parallel_scope($reqcopy);
-            foreach (@$reqs) {
-                push @requests, {%$_};
-            }
+        $reqs = get_parallel_scope($reqcopy);
+        foreach (@$reqs) {
+            push @requests, {%$_};
         }
     }
     return \@requests;
@@ -197,12 +206,22 @@ sub get_broadcast_disjoint_scope_with_parallel {
 
     my $sn_hash = shift;
     my @reqs = ();
+    $req->{_xcatpreprocessed}->[0] = 1;
+    my $isMN = xCAT::Utils->isMN();
     #Handle the one for current management/service node
-    if ( xCAT::Utils->isMN() ) {
-        $reqs = get_parallel_scope($req); 
+    if ( $isMN ) {
+        $reqs = get_parallel_scope($req);
+    } else {
+        # get site.master and broadcast to it as MN will not be in SN list or defined in noderes.servicenode.
+        my @entries = xCAT::TableUtils->get_site_attribute("master");
+        my $master = $entries[0];
+        my $reqcopy = {%$req};
+        $reqcopy->{'_xcatdest'} = $master;
+        $reqs = get_parallel_scope($reqcopy);
     }
     my @requests = @$reqs;
 
+    my $handled4me = 0;
     #Broadcast the request to other management/service nodes
     foreach (keys %$sn_hash) {
         my $xcatdest = $_;
@@ -210,15 +229,26 @@ sub get_broadcast_disjoint_scope_with_parallel {
             my $reqcopy = {%$req};
             $reqcopy->{'_xcatdest'} = $xcatdest;
             $reqcopy->{'node'} = $sn_hash->{$xcatdest};
-            $reqcopy->{_xcatpreprocessed}->[0] = 1;
 
             $reqs = get_parallel_scope($reqcopy);
             foreach (@$reqs) {
                 push @requests, {%$_};
             }
+        } elsif ($isMN == 0) {
+            # avoid handle myself multiple times when different IP address used in `noderes.servicenode`.
+            next if $handled4me;
+            my $reqcopy = {%$req};
+            $reqcopy->{'_xcatdest'} = $xcatdest;
+            $reqcopy->{'node'} = $sn_hash->{$xcatdest};
+
+            $reqs = get_parallel_scope($reqcopy);
+            foreach (@$reqs) {
+                push @requests, {%$_};
+            }
+            $handled4me = 1;
         }
     }
-    #print Dumper(\@requests);
+
     return \@requests;
 }
 
--- a/perl-xCAT/xCAT/Scope.pm
+++ b/perl-xCAT/xCAT/Scope.pm
@@ -126,7 +126,8 @@ sub get_parallel_scope {
 =head3 get_broadcast_scope_with_parallel
 
     Convert a request object to an array of multiple requests according to the 
-    splitted node range.
+    splitted node range. Also it replicates the requests to all required service
+    nodes or management node.
 
     Arguments:
        Reference of request
@@ -155,7 +156,8 @@ sub get_broadcast_scope_with_parallel {
     my @requests = (); # The request array will be return.
     push @requests, @$reqs;
 
-    # get site.master from DB and dispatch to it as MN will not be in SN list.
+    # when this method is called on service node, it is required to broadcast to MN too.
+    # get site.master from DB in order to dispatch to MN ( MN will not be added in servicenode table)
     if ( xCAT::Utils->isServiceNode() ) {
         my @entries = xCAT::TableUtils->get_site_attribute("master");
         my $master = $entries[0];
@@ -184,8 +186,10 @@ sub get_broadcast_scope_with_parallel {
 
 =head3 get_broadcast_disjoint_scope_with_parallel
 
-    Convert a request object to an array of multiple request objects according to the 
-    service node management scope. (Work under disjoint mode)
+    Convert a request object to an array of multiple requests according to the 
+    splitted node range. Also it replicates the requests to all required service
+    nodes or management node, but the request to a service node will only contains 
+    the node range it manages.
 
     Arguments:
        Reference of request
@@ -245,9 +249,6 @@ sub get_broadcast_disjoint_scope_with_parallel {
 
             $reqs = get_parallel_scope($reqcopy);
             push @requests, @$reqs;
-            #foreach (@$reqs) {
-            #    push @requests, {%$_};
-            #}
         } elsif ($handled4me == 0) {
             my $reqcopy = {%$req};
             $reqcopy->{'node'} = $sn_hash->{$xcatdest};
--- a/perl-xCAT/xCAT/Scope.pm
+++ b/perl-xCAT/xCAT/Scope.pm
@@ -146,33 +146,35 @@ sub get_broadcast_scope_with_parallel {
     }
     #Exit if the packet has been preprocessed in its history
     if ($req->{_xcatpreprocessed}->[0] == 1) { return [$req]; }
+    $req->{_xcatpreprocessed}->[0] = 1;
 
     my $snlist = shift;
-    my @reqs = ();
-    $req->{_xcatpreprocessed}->[0] = 1;
-    my $isMN = xCAT::Utils->isMN();
-    #Handle the one for current management/service node
-    if ( $isMN ) {
-        $reqs = get_parallel_scope($req);
-    } else {
-        # get site.master and broadcast to it as MN will not be in SN list or defined in noderes.servicenode.
+
+    my $reqs = get_parallel_scope($req);
+
+    my @requests = (); # The request array will be return.
+    push @requests, @$reqs;
+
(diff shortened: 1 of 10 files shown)


robin2008:
UT:

3 compute nodes, and their parent as below:
```
lsdef mid08tor31cn0[3-5] -i servicenode
Object name: mid08tor31cn03
    servicenode=
Object name: mid08tor31cn04
    servicenode=sn02
Object name: mid08tor31cn05
    servicenode=sn01
```
[sharedtftp=0]

If not set disjointnetboot, the default value is 0. After nodeset,  we can see
all MN and SN are handling all 3 CNs.
```
xdsh sn[01-02] ls /tftpboot/petitboot/mid08tor31*
sn01: /tftpboot/petitboot/mid08tor31cn03
sn01: /tftpboot/petitboot/mid08tor31cn04
sn01: /tftpboot/petitboot/mid08tor31cn05
sn02: /tftpboot/petitboot/mid08tor31cn03
sn02: /tftpboot/petitboot/mid08tor31cn04
sn02: /tftpboot/petitboot/mid08tor31cn05
ls /tftpboot/petitboot/mid08tor31* -1
/tftpboot/petitboot/mid08tor31cn03
/tftpboot/petitboot/mid08tor31cn04
/tftpboot/petitboot/mid08tor31cn05
```
set **disjointnetboot=1**

```
chdef -t site disjointnetboot=1

# nodeset mid08tor31cn0[3-5] osimage=rhels7.3-ppc64le-netboot-compute
mid08tor31cn03: netboot rhels7.3-ppc64le-compute
mid08tor31cn04: netboot rhels7.3-ppc64le-compute
mid08tor31cn05: netboot rhels7.3-ppc64le-compute
mid08tor31cn04: netboot rhels7.3-ppc64le-compute
mid08tor31cn05: netboot rhels7.3-ppc64le-compute
# xdsh sn[01-02] ls /tftpboot/petitboot/mid08tor31*
sn01: /tftpboot/petitboot/mid08tor31cn05
sn01: ls: cannot access /tftpboot/petitboot/mid08tor31cn03: No such file or directory
sn01: ls: cannot access /tftpboot/petitboot/mid08tor31cn04: No such file or directory
sn02: /tftpboot/petitboot/mid08tor31cn04
sn02: ls: cannot access /tftpboot/petitboot/mid08tor31cn03: No such file or directory
sn02: ls: cannot access /tftpboot/petitboot/mid08tor31cn05: No such file or directory
# ls /tftpboot/petitboot/mid08tor31* -1
/tftpboot/petitboot/mid08tor31cn03
/tftpboot/petitboot/mid08tor31cn04
/tftpboot/petitboot/mid08tor31cn05
```
we can see that only MN has all config files, but sn01, sn02 only contains its managed nodes.

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -528,27 +527,56 @@ sub process_request {
...
+                foreach (@{ $sn_hash->{$cur_xmaster} }) { $managed{$_} = 1; }
+                last;
+            }
+        }
 
         foreach (@rnodes) {
+            # For MN, the scope is all CN, for SN, the scope is the nodes it managed if disjointnetboot is set.
+            my $req2manage = exists($managed{$_}) || xCAT::Utils->isMN();
+            if ($mynodeonly == 1 && $req2manage != 1) {
+                push @unmanagednodes, $_;
```
zet809: Do we need to log the warning message that the node ... is not managed by SN/MN ...?
Otherwise, why do we need to have @unmanagednodes?

xCAT-server/lib/xcat/plugins/petitboot.pm
```diff
@@ -408,27 +408,57 @@ sub process_request {
         return;
     }
 
-    #if not shared tftpdir, then filter, otherwise, set up everything
+    my @hostinfo = xCAT::NetworkUtils->determinehostname();
+    $::myxcatname = $hostinfo[(scalar @hostinfo) - 1];
```
zet809: Same comments as grub2.pm

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -528,27 +527,56 @@ sub process_request {
...
+
+        # Get managed node list under current server
+        # The node will be under under 'site.master' if no 'noderes.servicenode' is defined
         my $sn_hash = xCAT::ServiceNodeUtils->getSNformattedhash(\@rnodes, "xcat", "MN");
-        my %managed = {};
-        foreach (@{ $sn_hash->{$cur_xmaster} }) { $managed{$_} = 1; }
+        #print Dumper($sn_hash);
+        my %managed = ();
+        foreach (keys %$sn_hash) {
+            if (exists($iphash{$_})) {
```
robin2008: During the UT, found that the sn_hash key might be name or IP,  we need to go through the all the iphash to see if it is matched.

xCAT-server/lib/xcat/plugins/grub2.pm
robin2008: unmanagednodes is reserved for next action: to offline unmanaged nodes on this SN.
And I think no warning message is required, as it is normal behavior. But I can add a debug message there (debugmode=2).

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -528,27 +527,56 @@ sub process_request {
         return;
     }
 
-    #if not shared tftpdir, then filter, otherwise, set up everything
+    my @hostinfo = xCAT::NetworkUtils->determinehostname();
+    $::myxcatname = $hostinfo[(scalar @hostinfo) - 1];
```
robin2008: I can polish it.

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -528,27 +527,56 @@ sub process_request {
...
+                my $cur_xmaster = $_;
+                foreach (@{ $sn_hash->{$cur_xmaster} }) { $managed{$_} = 1; }
+                last;
+            }
+        }
 
         foreach (@rnodes) {
+            # For MN, the scope is all CN, for SN, the scope is the nodes it managed if disjointnetboot is set.
+            my $req2manage = exists($managed{$_}) || xCAT::Utils->isMN();
+            if ($mynodeonly == 1 && $req2manage != 1) {
```
immarvin: ``$req2manage `` takes a bool value,  should better be ``($mynodeonly == 1 && $req2manage)``

immarvin:
The code logic looks good,  I suggest the process of "makedhcp" and ``nodeset offline`` should better be considered with this PR together, otherwise, once this PR is merged, there might be issues.

robin2008:
Updated the codes according to the comments, and big changes on how to handle the error message when checking if in the same network.

robin2008:
Couples of changes after discussion (https://github.com/xcat2/xcat-core/issues/3088#issuecomment-316317305), waiting for new Change set.

xCAT-server/lib/xcat/plugins/grub2.pm
zet809: Seems ok, the servicenode attribute can be set with SN ip address. Thx!

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -528,34 +527,80 @@ sub process_request {
...
+        }
+
+        my $notsamenet_nodes_err  = '';
+        my $notsamenet_nodes_warn = '';
 
         foreach (@rnodes) {
+            # For MN, the scope is all CN, for SN, the scope is the nodes it managed if disjointnetboot is set.
+            my $req2manage = exists($managed{$_});
+            if ($req2manage) {
+                push @unmanagednodes, $_;
```
zet809: If $req2manage is true, it means $_ can be managed, why it is also put to @unmanagednodes array?

xCAT-server/lib/xcat/plugins/petitboot.pm
```diff
@@ -408,34 +408,80 @@ sub process_request {
...
+        }
+
+        my $notsamenet_nodes_err  = '';
+        my $notsamenet_nodes_warn = '';
 
         foreach (@rnodes) {
+            # For MN, the scope is all CN, for SN, the scope is the nodes it managed if disjointnetboot is set.
+            my $req2manage = exists($managed{$_});
+            if ($req2manage) {
+                push @unmanagednodes, $_;
```
zet809: If $req2manage is true, it means $_ can be managed, why it is also put to @unmanagednodes array?

xCAT-server/lib/xcat/plugins/xnba.pm
```diff
@@ -450,36 +451,80 @@ sub process_request {
...
+        }
+
+        my $notsamenet_nodes_err  = '';
+        my $notsamenet_nodes_warn = '';
 
-        # Whatever the node managed by this xcatmaster explicitly, if the node in same subnet, we need to handle its boot configuration files
         foreach (@rnodes) {
+            # For MN, the scope is all CN, for SN, the scope is the nodes it managed if disjointnetboot is set.
+            my $req2manage = exists($managed{$_});
+            if ($req2manage) {
```
zet809: If $req2manage is true, it means $_ can be managed, why it is also put to @unmanagednodes array?

perl-xCAT/xCAT/Scope.pm
```diff
@@ -1,5 +1,6 @@
 package xCAT::Scope;
 
+#use Data::Dumper;
```
samveen: Maybe remove this?

xCAT-server/lib/xcat/plugins/xnba.pm
```diff
@@ -437,73 +467,116 @@ sub process_request {
...
+    # Filter those nodes which have bad DNS: not resolvable or inconsistent IP
+    my %failurenodes = ();
+    my %preparednodes = ();
+    foreach (@rnodes) {
+        my $ipret = xCAT::NetworkUtils->checkNodeIPaddress($_);
+        my $errormsg = $ipret->{'error'};
+        my $nodeip = $ipret->{'ip'};
+        if ($errormsg) {# Add the node to failure set
+            xCAT::MsgUtils->trace(0, "E", "xnba: IP address of $_ is $nodeip. $errormsg");
+            unless ($nodeip) {
```
zet809: If the $errormsg return can not resolve ip for a node, will it confusing that the tracing message return "IP of ... is ..."? Why shall we continue if error returned?

robin2008: 1, If there is an IP defined in DB,  but only DNS error.  nodeset/makedhcp could still go and report warning or error. (Depends on its existing business logic in each plugin).  For example, offline, shutdown, boot, also for osimage=xxx,  we should keep the capability as is.

2, Maybe "Defined IP address ..." is better here.

perl-xCAT/xCAT/Scope.pm
robin2008: Thanks, I'll remove it.

perl-xCAT/xCAT/Scope.pm
```diff
@@ -146,24 +147,78 @@ sub get_broadcast_scope_with_parallel {
...
+    if ($req =~ /xCAT::Scope/) {
+        $req = shift;
+    }
+    #Exit if the packet has been preprocessed in its history
+    if ($req->{_xcatpreprocessed}->[0] == 1) { return [$req]; }
+
+    my $sn_hash = shift;
+    my @reqs = ();
+    #Handle the one for current management/service node
+    if ( xCAT::Utils->isMN() ) {
```
immarvin: might need to consider the scenario that sn forward ``nodeset`` command to MN

perl-xCAT/xCAT/Scope.pm
```diff
@@ -146,24 +147,78 @@ sub get_broadcast_scope_with_parallel {
...
             $reqcopy->{_xcatpreprocessed}->[0] = 1;
-            #Apply callback to filter the node range in future.
+
             $reqs = get_parallel_scope($reqcopy);
             foreach (@$reqs) {
                 push @requests, {%$_};
             }
         }
     }
+    #print Dumper(\@requests);
```
samveen: Maybe remove this too

xCAT-server/lib/xcat/plugins/petitboot.pm
```diff
@@ -346,23 +349,55 @@ sub preprocess_request {
...
+            }
+
+            my $sn_hash = xCAT::ServiceNodeUtils->getSNformattedhash(\@CN, "xcat", "MN");
+            my @dhcpsvrs = ();
+            my $ntab = xCAT::Table->new('networks');
+            if ($ntab) {
+                foreach (@{ $ntab->getAllEntries() }) {
+                    next unless ($_->{dynamicrange});
+                    # if dynamicrange specified but dhcpserver was not - issue error message
+                    push @dhcpsvrs, $_->{dhcpserver} if ($_->{dhcpserver})
```
zet809: If the dhcpserver is set to the facing CN IP of SN, the nodeset request can not be dispatched to it from MN since the MN can not access it directly, how to deal with it?

robin2008: Check the dhcp.pm codes, not required to set the facing IP for networks.dhcpserver, we can just use service node name here. Then MN and all SN could access it.  If using IP, admin must ensure MN and SN could access the IP address.

perl-xCAT/xCAT/Scope.pm
```diff
@@ -129,12 +130,12 @@ sub get_parallel_scope {
 
     Arguments:
        Reference of request
-       Callback: TODO, Optional, the Callback will be used to filter the nodes
+       SN list: Array of target service nodes
```
robin2008: updated.

xCAT-server/lib/xcat/plugins/petitboot.pm
zet809: OK, Thx, looks good to me.

zet809:

PR #3461, opened 2017-07-14
[DO NOT MERGE]Add new test case to daily regression bundles

hu-weihua:
From 2.13.4sprint1 to now, 5 sprints, FVT have added 97 new cases. 35 cases need to be added into each daily regression bundles. Due to some cases is unsuitable for one specific scenario, so the number of updated cases for specific bundle is less than 35.


tingtli:
@hu-weihua  Since we have switch to rhels7.4. Do you think we also need to update rhels7.4 bundle files.

tingtli:
i have reviewed and agree to merge

samveen:
@tingtli Please update the milestone of this ticket from `2.13.6` to `2.13.7`

PR #3462, opened 2017-07-14
Kickstart improvements for Centos templates

samveen:
The following changes:
* Improve Centos7 `compute` template
* Add Centos7 `kvm` template.


immarvin:
I am ok with the change

PR #3465, opened 2017-07-17
remove confignetworks cases since daily regression environment does not support the cases

tingtli:
@bybai Please review. Thanks!


bybai:
I am OK to merge.

Issue #3297, opened 2017-06-21
[fvt]2.13.5 rinstall could not set runcmd and runimg's currchain if using “ rinstall noderange "runcmd=cmdtest,shell"”


junxiawang:
env:all os
build:
c910f04x35v05:/opt/xcat/share/xcat/tools/autotest/testcase/genesis # lsxcatd -v
Version 2.13.5 (git commit 6720675b21440e30d56f7b4e491352f4587c6617, built Tue Jun 20 06:15:47 EDT 2017)
c910f04x35v05:/opt/xcat/share/xcat/tools/autotest/testcase/genesis #
How to reproduce:
```
1. rinstall "c910f04x35v07" "runcmd=cmdtest,shell"
2.# tabdump chain
#node,currstate,currchain,chain,ondiscover,comments,disable
"c910f04x35v06","boot","boot",,,,
"c910f04x35v07","runcmd=cmdtest",,,,,
------------------> could not set currchain using rinstall
```

whowutwut:
👍 @junxiawang is this manually caught or was this caught in automation. . ?

junxiawang:
@whowutwut  as autotest sometimes failed so I find the root cause manually.

whowutwut:
@junxiawang If the environment is ALL OS, we should put the examples from the PPC64 LE related OS, to prove the issue and ensure we are focused on the correct priority.

zet809:
Fixed in pull request #3491

PR #3491, opened 2017-07-18
fix issue 3297: rinstall could not set runcmd and runimg's currchain

zet809:
Fix issue #3297 

```
root@c910f03c05k10:/tmp# tabdump chain | grep cn1
"cn1",,,,,,
root@c910f03c05k10:/tmp# rinstall cn1 shell,runcmd=runcmd,osimage=rhels7.3-ppc64le-install-compute
Provision node(s): cn1
...
root@c910f03c05k10:/tmp# tabdump chain | grep cn1
"cn1","shell","runcmd=runcmd,osimage=rhels7.3-ppc64le-install-compute",,,,
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/rinstall.pm
+++ b/xCAT-server/lib/xcat/plugins/rinstall.pm
@@ -132,9 +132,6 @@ sub rinstall {
         else {
             unless ($state =~ /-/) {
                 $STATES = $state;
-                if ($RESTSTATES) {
-                    $STATES .= ",$RESTSTATES";
-                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/rinstall.pm
```diff
@@ -132,6 +132,9 @@ sub rinstall {
         else {
             unless ($state =~ /-/) {
                 $STATES = $state;
+                if ($RESTSTATES) {
+                    $STATES .= ",$RESTSTATES";
+                }
```
robin2008: Suggest to move these codes to line 259-261 to make the codes a bit more consistent.  After that all appending for $RESTSTATES is in the same if-elseif-else.

Issue #3496, opened 2017-07-18
OpenBMC debug not available when we are in configured for HIERARCHY


whowutwut:
For OpenBMC, we have used XCATBYPASS=1 to print out debug information and the rest calls... with Hierarchy we can't use it. 

```
[root@fs3 xcat]# XCATBYPASS=1 rpower mid05tor12cn02 state
Warning: XCATBYPASS is set, skipping hierarchy call to sn01
```

Can we convert the openbmc code to use the `site.xcatdebugmode=1`, similar to how we are doing things for `[ipmi_debug]` calls? 

Need a way to get some information about to deubg the openbmc code in a hierarchy environment.

PR #3506, opened 2017-07-19
fix 3496, depends on xcatdebugmode to print debug msg

xuweibj:
#3496
1. If ``xcatdebugmode`` is ``1`` or ``2``, will print debug msg.
```
[root@fs3 ~]# rpower mid05tor12cn02 stat
[OpenBMC development support] Using this version of xCAT, ensure firware level is at v1.99.6-0-r1, or higher.
mid05tor12cn02: [openbmc_debug] POST https: //172.20.140.102/login -d { "data": [ "root", "0penBmc" ] }
mid05tor12cn02: [openbmc_debug] login_response 200 OK
mid05tor12cn02: [openbmc_debug] GET https: //172.20.140.102/xyz/openbmc_project/state/enumerate
mid05tor12cn02: [openbmc_debug] rpower_status_response 200 OK
mid05tor12cn02: [openbmc_debug] State CurrentBMCState=xyz.openbmc_project.State.BMC.BMCState.Ready
mid05tor12cn02: [openbmc_debug] State RequestedBMCTransition=xyz.openbmc_project.State.BMC.Transition.None
mid05tor12cn02: [openbmc_debug] State CurrentPowerState=xyz.openbmc_project.State.Chassis.PowerState.On
mid05tor12cn02: [openbmc_debug] State RequestedPowerTransition=xyz.openbmc_project.State.Chassis.Transition.Off
mid05tor12cn02: [openbmc_debug] State CurrentHostState=xyz.openbmc_project.State.Host.HostState.Running
mid05tor12cn02: [openbmc_debug] State RequestedHostTransition=xyz.openbmc_project.State.Host.Transition.Reboot
mid05tor12cn02: on
```

2. For hierarchy env, ``GetOptions`` could not get option value. So modified to use parameters input.

3. If ``-V`` or ``--verbose`` is not the last parameter, will print error msg
```
# rpower mid05tor12cn02 -V state
[OpenBMC development support] Using this version of xCAT, ensure firware level is at v1.99.6-0-r1, or higher.
Error parsing arguments.
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -485,7 +485,7 @@ sub parse_args {
 
     if (scalar(@ARGV) >= 2 and ($command =~ /rpower|rinv|rsetboot|rvitals/)) {
         return ([ 1, "Only one option is supported at the same time for $command" ]);
-    } elsif (scalar(@ARGV) == 0 and $command =~ /rpower|rsetboot|rspconfig|rflash/) {
+    } elsif (scalar(@ARGV) == 0 and $command =~ /rpower|rspconfig|rflash/) {
         return ([ 1, "No option specified for $command" ]);
     } else { 
         $subcommand = $ARGV[0];
@@ -511,10 +511,15 @@ sub parse_args {
         # disable function until fully tested
         #
         $check = unsupported($callback); if (ref($check) eq "ARRAY") { return $check; }
+        $subcommand = "stat" if (!defined($ARGV[0]));
         unless ($subcommand =~ /^net$|^hd$|^cd$|^def$|^default$|^stat$/) {
             return ([ 1, "Unsupported command: $command $subcommand" ]);
         }
     } elsif ($command eq "reventlog") {
+        #
+        # disable function until fully tested
+        #
+        $check = unsupported($callback); if (ref($check) eq "ARRAY") { return $check; }
         my $option_s = 0;
         unless (GetOptions("s" => \$option_s,)) {
             return ([1, "Error parsing arguments." ]);
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -29,10 +29,10 @@ use xCAT::SvrUtils;
 use xCAT::GlobalDef;
 use xCAT_monitoring::monitorctrl;
 
-$::VERBOSE = 0;
+$::VERBOSE                  = 0;
 # String constants for rbeacon states
-$::BEACON_STATE_OFF="off";
-$::BEACON_STATE_ON="on";
+$::BEACON_STATE_OFF         = "off";
+$::BEACON_STATE_ON          = "on";
 # String constants for rpower states
 $::POWER_STATE_OFF          = "off";
 $::POWER_STATE_ON           = "on";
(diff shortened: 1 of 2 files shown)


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -460,24 +467,28 @@ sub parse_args {
...
-    } elsif (scalar(@ARGV) == 2) {
-        # Check if one is calling for Verbose output
-        foreach (@ARGV) {
-           if ($_ =~ /V|verbose/) {
-              $::VERBOSE=1;
-           } else {
-               $subcommand = $_
-           }
-        }
+    my $verbose    = undef;
```
gurevichmark: Do we need so many different "verbose" variables ?
It seems we have $verbose, $::VERBOSE and $xcatdebugmode.
Are they the same, controlling "[openbmc_debug]" output. Or is there a difference between them ?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -460,24 +467,28 @@ sub parse_args {
...
+
+    # If command includes '-V', it must be the last one prarmeter. Or print error message.
+    if ($verbose) {
+        my $option = $$extrargs[-1];
+        return ([ 1, "Error parsing arguments." ]) if ($option !~ /V|verbose/);
+    }
+
+    if (scalar(@ARGV) >= 2 and ($command =~ /rpower|rinv|rsetboot|rvitals/)) {
+        return ([ 1, "Only one option is supported at the same time for $command" ]);
+    } elsif (scalar(@ARGV) == 0 and $command =~ /rpower|rsetboot|rspconfig|rflash/) {
```
gurevichmark: It looks like `rsetboot` allows not specifying a subcommand. It defaults to `stat` for ipmi controlled nodes. Do we want the openbmc controlled nodes to do the same ?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -460,24 +467,28 @@ sub parse_args {
...
-        foreach (@ARGV) {
-           if ($_ =~ /V|verbose/) {
-              $::VERBOSE=1;
-           } else {
-               $subcommand = $_
-           }
-        }
+    my $verbose    = undef;
+    unless (GetOptions(
+        'V|verbose'  => \$verbose,
```
gurevichmark: I think man pages and Usage.pm need to be modified. Not all of the commands handled in this module list `-V|--verbose` as a valid option. Some, like `rsetboot` show it in usage output but not in man page.

xCAT-server/lib/xcat/plugins/openbmc.pm
xuweibj: I saw that @whowutwut added ``verbose`` for rinv firm. I thought that it could be used by users to get more firm information. If not, I think we could mv ``$::VERBOSE`` and use ``xcatdebugmode`` to show more as debug information. What is your option? @whowutwut

whowutwut: Yes, the --verbose option for rinv is to display more output than running the command normally.  For the firmware case, it will display all software versions that have been uploaded to the BMC.  Without it, it will display only the active one.    In this case, it's not for DEBUG but to show more output.  

xcatdebugmode should be used to print debug (similar to XCATBYPASS) since we cannot use that in hierarchy environments.

whowutwut:
Not a change that is needed...   @xuweibj There's a couple places that we use `print Dumper` I think those just won't get printed in a hierarchy environment now...    @gurevichmark one place is in the flash code, is that dumper stuff necessary?

xCAT-server/lib/xcat/plugins/openbmc.pm
xuweibj: For openbmc, only rinv command has ``-V``, I saw that it has been in Usage.pm. And for ``rsetboot``, I think we could delete ``-V`` in Usage.

whowutwut:
@xuweibj  When I run the openbmc.pm code in XCATBYPASS mode: 

```
Use of uninitialized value in pattern match (m//) at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 654.
```

Caused by this pull request... 

I think we need to change this to something like:
```
    if (defined($$subcommands[-1]) and $$subcommands[-1] =~ /V|verbose/) {
        $::VERBOSE = 1;
        pop(@$subcommands);
    }
```

Can you update this the next time you touch the file?

xuweibj:
Hi @whowutwut, I have modified it in pull request #3677 .

PR #3509, opened 2017-07-19
doxcat improvements

samveen:
This PR adds a check if the `cpufreq` governor being applied is available or not.


immarvin:
i am ok with the change

Issue #3513, opened 2017-07-20
The document about build genesis locally with latest RH MN


zet809:
As mentioned in #2933 
To support mlx4_en driver for genesis base OS newer than Fedora 24, there is no tar ball available to build from.
This task is used to trace the issue that build xcat-genesis-base locally on the RH MN.
We plan to only support the latest RH OS to minimize our maintance effort.

zet809:
Before the build process is available,  the steps below can be used to install xCAT-genesis-base based on Fedora 24 with mlx-en driver verson 3.2-1.0.1.1.
```
rpm -e xCAT-genesis-base-ppc64-2.13.6-snap201707190550.noarch --nodeps
yum install https://xcat.org/files/xcat/xcat-dep/2.x_Linux/beta/xCAT-genesis-base-ppc64-2.13.2-snap201702150049.noarch.rpm
mknb ppc64
```

zet809:
We have the xCAT-genesis-build rpm on https://xcat.org/files/xcat/xcat-dep/2.x_Linux/beta/, so close this task.

PR #3514, opened 2017-07-20
remove the build mlx4-en and i40e section when building xcat-genesis-…

zet809:
…base

* Can not find suitable mlx4-en driver for Fedora26.
* Fedora26 had shipped newer i40e driver than 1.5.16

So, remove the section to build those 2 drivers.

### DIFF
--- a/xCAT-genesis-builder/buildrpm
+++ b/xCAT-genesis-builder/buildrpm
@@ -5,7 +5,7 @@
 #            this whole dir into it somewhere (like /tmp).
 # Then run this script.  The optional 1st arg should be mcp if you are building against mcp.
 
-# Currently, *Fedora 23* is the only OS supported to build genesis-base for ppc64, and Centos 6.5 for x86_64
+# Currently, *Fedora 26* is the only OS supported to build genesis-base for ppc64, and Centos 6.5 for x86_64
 
 HOSTOS="$1"
 DIR=`dirname $0`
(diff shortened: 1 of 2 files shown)


whowutwut:
@zet809  How does this affect existing customers if they update it.. for ppc64 and x86_64? 
 Perhaps we should look at creating a special genesis base for rh7.4? until we can close out the gaps?

zet809:
The only issue is if customer have Mellanox EN adapter, these adapter won't be show in the discovery request. 
And for the next 2 question, I think yes we shall at least figure out the steps to create genesis-base locally based on Latest RH, had opened issue #3513 to trace. But this is a little low priority, right?

whowutwut:
@zet809  I think we just need to create a directory for `rh7.4` and then the respective scripts should understand this ...   

* `go-xcat` if installing on rhels 7.4, or detected it, then it will use the rh7.4 directory. 
* the user needs to `cd rh7.4/ppc64le` and run `./mklocalrepo`

zet809:
Hi, @whowutwut , from xCAT-genesis-base point of view, it doesn't care about which OS it is installed, but the object node it will be deployed on. For example, it the MN is redhat 7.3, it still shall be able to discovery witherspoon server.
So, before the mlx-en driver is available on Fedora 26, the best way to support it is build genesis-base on MN if the OS is latest Redhat release. The task #3513 had been opened to trace it.

whowutwut:
delegate to @immarvin

immarvin:
hi @zet809 , the kernel-4.11.8-300.fc26.src.rpm can be downloaded from

http://dl.fedoraproject.org/pub/fedora/linux/releases/26/Everything/source/tree/Packages/k/kernel-4.11.8-300.fc26.src.rpm

seems it contains the mlx drivers.

you might need to rebuild  it

zet809:
It doesn't mean there is no mlx4_en.ko, it means the file is not new enough to support new hardware. 
Maybe the best method is build genesis on OS that can build latest mlx4_en driver.

immarvin:
talked with @zet809 , if the user want to enable Mellanox EN adapter in genesis for witherspoon, he might need to wait for the  latest mlx driver and rebuild xcat genesis base before we officially pack the latest mlx driver in.

The newly built xcat genesis based on fedora 26 also include the 2.2-1 mlx_en and mlx_core driver, it cannot support mlx en adaper on witherspoon, but might work for the old servers with mlx en adaper, we did not verify this because we do not have this kind hareware.

PR #3519, opened 2017-07-20
xCAT-client:  allow pgsqlsetup(1) to accept additional listen_addresses

khm:
On AIX, pgsqlsetup(1) only configures the management node IP in PostgreSQL's configuration. On Linux, it only configures localhost and the management node IP. This change creates a new option in pgsqlsetup(1) that allows the user to specify an additional address to be added in the listen_addresses line in PostgreSQL's configuration. If the user specifies this option more than once, all specified addresses are added to the line.

This is useful in cases where the management node has more than one IP, such as in HA configurations where an address is shared between multiple management nodes. In these cases, pgsqlsetup(1) may use the system's primary address, when it is preferred to listen on the shared address as well. This is also useful in situations where the cluster is fragmented across different subnets, and the management node needs to talk to service nodes that are on disparate networks.


immarvin:
i am ok with the change

PR #3521, opened 2017-07-21
fix bundles for confignetwork and confignics

bybai:
Since confignics test cases does not backup and restore network nics configuration files, delete all confignics test cases; Add new confignetwork test cases in 
```
sles12.2_ppc64le.bundle
sles12.2_x86_64.bundle
rhels7.3_ppc64le.bundle
rhels7.3_x86_64.bundle
```


hu-weihua:
It is ok for me to merge. thanks @bybai

Issue #3319, opened 2017-06-22
noderange in xCAT commands might be broken by shell expansion


ChristopherWalker:
Having just tried to reprovision two nodes (dn22 and dn23), I've reprovisioned dn2 instead.

The command I ran was:

[root@mgt.apocrita consoles]# rinstall dn[22-23] --osimage="apocrita-centos7.3-x86_64-install-compute" -u
Provision node(s): dn2

Needless to say, this hasn't been popular with the user running jobs on the node. This of course would work correctly with differently named files in the directory.

Perhaps http://xcat-docs.readthedocs.io/en/stable/guides/admin-guides/references/man3/noderange.3.html

could warn about the necessity of quoting these on the command line (and even better give an example). 

The problem is caused by the presence of a file named dn2 in the directory. 

cd /var/-bash-4.1$ cd /var/log/consoles/
-bash-4.1$ echo dn[22-23]
dn2 dn3
-bash-4.1$ ls -l dn2 
-rw-r--r-- 1 root root 1677108 Jun 21 23:02 dn2
-bash-4.1$ ls -l dn3
-rw-r--r-- 1 root root 2035512 Jun 17 23:06 dn3
-bash-4.1$ cd
-bash-4.1$ echo dn[22-23]
dn[22-23]

immarvin:
The steps to recreate:
````
[root@c910f03c05k21 xcat-core]# cd /var/log/consoles/
[root@c910f03c05k21 consoles]# ls
nd2  nd3
[root@c910f03c05k21 consoles]# ls nd[22-23]
nd2  nd3
[root@c910f03c05k21 consoles]# lsdef nd[22-23]
$VAR1 = [
          'nd2',
          'nd3'
        ];
$VAR1 = {
          'arg' => [
                     'nd2',
                     'nd3'
                   ],
          'cwd' => [
                     '/var/log/consoles'
                   ],
          'command' => [
                         'lsdef'
                       ]
        };
Error: noderange can not contain spaces.
````

this is caused by the shell expansion, ``nd[22-23]`` will be expanded to "nd2" and "nd3" under the working directory while running xCAT commands before it passed to ARGV of xCAT-client bins

Turn off the shell expansion might help. but it will modify the user environment and can not be accepted.  Modify the manpage and Documentation to make user aware of this

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3524

zet809:
Some changed need to document, so move back to backlog.

PR #3524, opened 2017-07-21
fix issue noderange in xCAT commands might be broken by shell expansion #3319

immarvin:
for issue: https://github.com/xcat2/xcat-core/issues/3319

modify the documentation, suggest user to quote noderange string in xCAT commands to avoid shell expansion

Compiled Doc: http://10.3.5.21/xcat-doc/html/guides/admin-guides/references/man3/noderange.3.html


daniceexi:
We may add the exception example in the map page for which needs to quote with quotation, otherwise I have to use quotation for all the noderange since I don't know which one has problem without quotation.

Issue #3406, opened 2017-07-06
Clean up the discovered node definition work around for P9 discovery


whowutwut:
Is it possible to modify the node discovery to look for the `node-<IP -> HEX>` discovered node definitions as part of hardware discovery?  

The issue here is that I'm not 100% confident that we will be able to rely on the model type serial number of these OpenBMC boxes because the VPD information is not avail until you boot the compute side. Then only the serial number appears.   The Model number is not yet burned in... 

So while we have this work around, if we can't find a node definition for `node-mt-sn` could we then look for the `node-<ip>` and delete it if it's found? 

The task include: 
* For bmcdiscover: 
- [x] Use MTM/Serial, if there
- [x] Fall back to MAC address (REST, if 00:00:00... then nmap value 
- [x] Temporary BMC node like node-<MTM>-<Serial> or node-<MAC of BMC>(In case MTMS is not available) will be created
* For dodiscovery in genesis:
- [x] Collect BMC MAC information through 'ipmitool lan print 1'
* For hardware discovery request matching process
- [x] Search corresponding BMC node with name like node-<MTM>-<Serial> or node-<MAC of BMC>(In case MTMS is not available)

whowutwut:
@zet809  I assigned this to you for now.  If you agree and think someone else should work on it , please reassign.

zet809:
OK, I see, it is a good idea to use the IP to match the temporary BMC node.

whowutwut:
You mentioned on the call that we need to be able to figure this out from the host.   So the MT/Serial could be done...   This temporary DHCP IP address will change after we made discovery and restarted network... , is that a problem in the removal?

zet809:
The Dynamic ip for the host will be changed during the node is booting into petitboot and genesis kernel, but the Dynamic IP for the BMC won't be changed until expired. So, i think it won't be a problem to remove the temporary BMC node.

zet809:
hi, @whowutwut , any idea?

whowutwut:
@zet809 do you mean use the MAC address to generate a temp bmc node definition name? We don't have that with bmc discover. 

I wonder if there is any unique ID we can use on the openbmc. Haven't figured any out yet..

whowutwut:
I think this should not block #3464  but the priority implementation should be... 

- [ ] Use MTM/Serial, if there
- [ ] Fall back to MAC address (REST, if 00:00:00... then nmap value 
- [ ] Ensure MAC address is cleaned up 
[...]

PR #3525, opened 2017-07-21
Task 3406, add BMC MAC in bmcdiscover and discovery process

xuweibj:
#3406 

1. For bmcdiscover.pm : 
    1.1 Get BMC MAC address when using ``nmap`` to get live IPs, and save in %ipmac hash.
    1.2 If can not get mtm or serial, define bmc node by ``node-$mac``.
    1.3 If can get mtm and serial, but is string like ``0000000000`` (all 0s), consider it as null. Define bmc node by ``node-$mac``.
    1.4 If can get valid mtm and serial, still use ``node-$mtm-$serial`` as bmc node name.

The output is when mtm is ``000000``.
```
# bmcdiscover --range 10.3.5.99 -w -z
node-42a5c12b1c3f:
	objtype=node
	groups=all
	bmc=10.3.5.99
	cons=openbmc
	mgt=openbmc
	serial=00TEST00
```

2. For dodiscovery:
    2.1 If mtm or serial is like ``0000000``(all 0s), will not be added into packet.
    2.2 Add BMC MAC into packet, will be sent to xcatd.

3. For discovery (switch.pm typemtm.pm seqdiscovery.pm):
    After checking whether there is bmc node in mp table by mtm and serial. Judge whether ``node-$mac`` in mp table. If has, it is the bmc node and remove it.

### DIFF
--- a/xCAT-genesis-scripts/bin/dodiscovery
+++ b/xCAT-genesis-scripts/bin/dodiscovery
@@ -141,10 +141,12 @@ if [ "$UUID" != "unknown" ]; then
         echo "<uuid>$UUID</uuid>" >> /tmp/discopacket
 fi
 
-if [ "$MTM" != "unknown" ]; then
+flag_mtm=`echo '$MTM' | sed 's/0//g'`
+if [ $flag_mtm ] && [ "$MTM" != "unknown" ]; then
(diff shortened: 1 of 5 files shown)


whowutwut:
@xuweibj  thank you for getting this in asap

Issue #3204, opened 2017-06-05
xCAT upgrade changes boot order settings


gurevichmark:
This problem was reported by `vsolasa@us.ibm.com - Veera s Solasa/Poughkeepsie/IBM`

1. All cluster nodes had their boot order set to "hd". After upgrade from xCAT 2.12 to 2.13.5 the boot order was changed to "default".  Same was observed when upgrading 2.13.3 to 2.13.4

2. The `rsetboot stat` returns `frame5u39: boot override inactive` message, which sounds like an error, but in fact just a message that default order is in effect. Perhaps we should change the message to something that sounds less like an error? Maybe something like "default boot order is in effect" ?

3. Should it be possible to provision a node in default boot order ? Currently provisioning gets stuck in genesis when boot order is default.

immarvin:
hi @zet809 , would you please take a look at this issue?

hi @gurevichmark , can you provide the node definition of frame5u39?

gurevichmark:
Yes. frame5u39 is defined on stratton01 as:
```
bject name: frame5u39
    arch=ppc64le
    bmc=50.5.39.1
    cons=ipmi
    cpucount=160
    cputype=POWER8 (raw), altivec supported
    currchain=boot
    currstate=boot
    disksize=sda:1000GB,sdb:1000GB
    groups=all,compute
    installnic=mac
    ip=10.5.39.1
    mac=98:be:94:59:ee:a7
    memory=261482MB
    mgt=ipmi
    mtm=8335-GTA
    netboot=petitboot
    nodetype=mp
    os=rhels7.4-pegas-alpha
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    profile=compute
    provmethod=rhels7.4-pegas-alpha-ppc64le-install-compute
    serial=2100EEA
    status=powering-on
    statustime=05-17-2017 11:25:43
```

zet809:
Hi, @gurevichmark , for the questions, the answer are:
For 1. need more investigate
       2.I agree with you that we need to modify the output message
       3. For OpenPOWER server, to do netboot (OS deployment or run genesis), 'rsetboot net' is needed to have the server boot from network otherwise it will choose the default boot item(maybe the harddisk). So, it is impossible to deploy a node with default boot order.

gurevichmark:
[...]

PR #3537, opened 2017-07-26
modify rsetboot depending on URL & data

xuweibj:
Interface: (Version v1.99.8-79-g246c6f9-dirty)
```
 # curl -k -b cjar -X PUT -H "Content-Type: application/json" -d '{"data" : "xyz.openbmc_project.Control.Boot.Source.Sources.Default"}' https://9.3.185.161/xyz/openbmc_project/control/host0/boot_source/attr/BootSource
{
  "data": null,
  "message": "200 OK",
  "status": "ok"
}

# curl -k -b cjar -X GET https://9.3.185.161/xyz/openbmc_project/control/host0/boot_source
{
  "data": {
    "BootSource": "xyz.openbmc_project.Control.Boot.Source.Sources.Default"
  },
  "message": "200 OK",
  "status": "ok"
}
```

Output:
```
# rsetboot redfishtest hd
[OpenBMC development support] Using this version of xCAT, ensure firware level is at v1.99.6-0-r1, or higher.
redfishtest: Hard Drive
# rsetboot redfishtest stat
[OpenBMC development support] Using this version of xCAT, ensure firware level is at v1.99.6-0-r1, or higher.
redfishtest: boot override inactive
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -1364,10 +1364,18 @@ sub rsetboot_response {
     my $response_info = decode_json $response->content;    
 
     if ($node_info{$node}{cur_status} eq "RSETBOOT_STATUS_RESPONSE") {
-        xCAT::SvrUtils::sendmsg("Hard Drive", $callback, $node) if ($response_info->{'data'}->{BootSource} =~ /Disk$/);
-        xCAT::SvrUtils::sendmsg("Network", $callback, $node) if ($response_info->{'data'}->{BootSource} =~ /Network$/);
-        xCAT::SvrUtils::sendmsg("CD/DVD", $callback, $node) if ($response_info->{'data'}->{BootSource} =~ /ExternalMedia$/);
-        xCAT::SvrUtils::sendmsg("Default", $callback, $node) if ($response_info->{'data'}->{BootSource} =~ /Default$/);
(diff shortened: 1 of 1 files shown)


gurevichmark:
Is it correct that when we get "Default" back we should display "boot override inactive" ?
From the comment that @zet809 added to Issue #3204, "Default" and "boot override inactive" are 2 different states:

```
[root@fs1 xcat]# rsetboot frame45cn04,frame45cn05,frame45cn07,frame45cn02 state
frame45cn02: boot override inactive
frame45cn05: BIOS default
frame45cn04: Network
frame45cn07: Network
```

zet809:
Hi, @gurevichmark , yes, "Default" shall be right output for default setting. "boot override inactive" means invalid or failed setting.

xuweibj:
@gurevichmark Thank you for your comments. I made a mistake there. I have modified, please help to review again.

gurevichmark:
Do you think it is possible for `boot_source/attr/BootSource` to return "boot override inactive" state ? 
What will we display in that case ?
Maybe, more generally, there should be a message we display if the string returned by `($response_info->{'data'}->{BootSource}` is something we do not expect.

xuweibj:
Hi @gurevichmark , from the design doc https://github.com/openbmc/phosphor-dbus-interfaces/blob/master/xyz/openbmc_project/Control/Boot/Source.interface.yaml, the ``data`` only can be ``Disk``, ``ExternalMedia``, ``Network`` and ``Default``. I think if we get other value, it will be a bug.

PR #3538, opened 2017-07-26
remove unsuitable cases from hardward control bundle

hu-weihua:
Remove ``rinv_mprom`` and ``rinv_guid`` from ``hdctrl_openpower_ipmi.bundle`` since these 2 cases is just available in x86.  remove ``rinv_deviceid``  since it does not exist any more.


bybai:
I am OK to merge it.

PR #3547, opened 2017-07-27
Fix 3474, [Usability] The nodeset output cannot tell where the message it coming from

### LINKED ISSUE #3474: [Usability] The nodeset output cannot tell where the message it coming from
The message is bad, can't tell where the message it coming from. my assumptions below...
```
[root@fs3 common]# nodeset mid05tor12cn02 offline
mid05tor12cn02: offline
mid05tor12cn02: offline
Error: The mid05tor12cn02 can not be resolved
mid05tor12cn02: offline
Error: The mid05tor12cn02 can not be resolved
```
Note:  need to cover other actions too, for example, stat

Trace https://github.ibm.com/vhu/c910env/issues/77

robin2008:
This PR is to fix #3474, to introduce a new method to show the message is coming from which xcatd. It would be helpful when there are many output from many SNs. For example, in a FLAT network with 16 SN (and not in disjoint mode)

makedhcp/nodeset will have 17 output for one node,  each SN will have one line. And if one of the SN report error, we can easily tell where is the problem.

Highlevel:
1, Every response object will be rendered with the name of running Mn or Sn.
2, Each xCAT Plugin could have the capability to disable it when it does not want to show the name.
3, In CLI side, to show the server name prefix when given option is specified.

Detail:
1, "xcatdsource" is added into each response object by default in xcatd, the value is the name of current server.
2, If user run CLI with environment variable "XCATSHOWSVR", then the output will be added server name prefix. The format as below.
3, If a response with "NoSvrPrefix" key (plugin could set it explicitly), then no extra server name will be shown even if user run it with "XCATSHOWSVR"

UT:
```
XCATSHOWSVR=1 nodeset fake1,fake4 osimage=rhels7.3-ppc64le-install-compute
fake1: [c910f03c05k29]: install rhels7.3-ppc64le-compute
fake1: [c910f03c05k20]: install rhels7.3-ppc64le-compute
fake4: [c910f03c05k20]: install rhels7.3-ppc64le-compute
Error: [c910f03c05k20]: The fake4 can not be resolved
```
```
nodeset fake1,fake4 osimage=rhels7.3-ppc64le-install-compute
fake1: install rhels7.3-ppc64le-compute
fake1: install rhels7.3-ppc64le-compute
fake4: install rhels7.3-ppc64le-compute
Error: The fake4 can not be resolved
```

Note:  Not only show the name of SN, but also for MN if the environment variable is set.

```
[root@c910f03c05k20 ~]# XCATSHOWSRV=1 copycds /mnt/iso/redhat/7.3/GA/RHEL-7.3-20161019.0-Server-x86_64-dvd1.iso
[c910f03c05k20]: Copying media to /install/rhels7.3/x86_64
10.46%
```

```
# XCATSHOWSRV=1 xdsh c910f03c05k29 uptime
[c910f03c05k20]: c910f03c05k29:  04:24:22 up 78 days,  6:18,  2 users,  load average: 0.00, 0.01, 0.05
```


robin2008:
manpage is updated, please review.

gurevichmark:
Was there good reason why we chose a environment variable to control this output instead of adding a verbose option `-V`. I think verbose option is more intuitive than a long env like `XCATSHOWSVR`.

xCAT-client/pods/man8/nodeset.8.pod
```diff
@@ -34,6 +40,8 @@ B<nodeset> only sets the next boot state, but does not reboot.
 B<nodeset>  is  called  by B<rinstall> and B<winstall> and is also called by the
 installation process remotely to set the boot state back to "boot".
 
+In a hierarchical cluster which managed by service nodes, the boot stat of the nodes should be kept in consistence on the mn and on the service nodes. Then B<nodeset> command is required to be dispatched to the service nodes for handling the boot configuration files. There are multiple messages for one node from service nodes, it is hard to tell where is the problem from. To make the output more clear, user could run the command with "XCATSHOWSVR=1" environment variable, and then the output message will be appended with the server information.
```
gurevichmark: I would suggest simplifying this paragraphs like this:

"In a hierarchical cluster managed by service nodes, B<nodeset> command is used to make sure compute node states are consistent on service and management nodes. When errors are reported, run the B<nodeset> command with "XCATSHOWSVR=1" environment variable. The command will display additional service node information, which might be useful in identifying the problem.

robin2008:
@gurevichmark I've updated the Changeset based your comments.

robin2008:
close and reopen it to trigger a new CI check.

PR #3556, opened 2017-07-28
modify hardware control testcase for physical machine

junxiawang:
I modify hardware control testcase for physical machine test.


xuweibj:

Issue #2529, opened 2017-02-22
[fvt]2.13.2  Database handle not defined after xcat installed on rhels7.3 ppc64le 


junxiawang:
env: rhels 7.3 ppc64le 
build:[root@c910f03c17k27 ~]# lsxcatd -v
Version 2.13.2 (git commit bf47335dccb7dd365fb175e219c2c325b6bd46e0, built Tue Feb 21 07:16:03 EST 2017)
[root@c910f03c17k27 ~]#

How to reproduce:
```
Feb 21 21:30:44 c910f03c17k28 xcat[2195]:  Could not connect to the database. Database handle not defined.
Feb 21 21:30:45 c910f03c17k28 xcat[2195]:  Could not connect to the database. Database handle not defined.
Feb 21 21:30:46 c910f03c17k28 xcat[2195]:  Could not connect to the database. Database handle not defined.
Feb 21 21:30:46 c910f03c17k28 xcat[2195]:  ERROR: Unable to open basic site table for configuration
Feb 21 21:31:32 c910f03c17k27 xcat: c910f03c17k29 status: netbooting statustime: 02-21-2017 21:31:32
Feb 21 21:31:45 c910f03c17k27 xcat[1569]: xCAT: Allowing getpostscript from c910f03c17k29
Feb 21 21:31:46 c910f03c17k27 xcat: getpostscript: sending data
Feb 21 21:31:46 c910f03c17k29 xcat:  Wed Feb 22 02:19:10 UTC
```

chenglch:
I looked into the environment, this problem can not be recreated. @junxiawang Can you check the database connection manually when this error happens?

junxiawang:
could not reproduce the issue

dspatil7768:
To solve this error we need to define `xcatdport` in site table.

`XCATBYPASS=1 chdef -t site xcatdport=3001`

whowutwut:
@dspatil7768 Did you hit the same issue recently? @junxiawang @chenglch did you see the comment above?

chenglch:
@whowutwut , I think the error that dspatil7768  encountered is not the same one described by @junxiawang. Junxia's problem is the database connection can not be setup with the connection string which causes the failure to read or write the site table. I don't know the running environment at that time(like: is the database started?). So I can not guess how it happened.

whowutwut:
Closing issue for now.

zet809:
hi, @chenglch , why you reopen this issue?

[...]

PR #3557, opened 2017-07-28
Retry 3 times if database connection can not be established

chenglch:
As connection may fail due to the limit or error from database side,
this patch add retry mechanism to reduce probability of failure.

fix-issue: #2529


zet809:

PR #3558, opened 2017-07-28
Add bunlds for rhels7.4

hu-weihua:
In order to support rhels7.4 in daily regression. Add bundles(ppc64le,ppc64,x86) for rhels7.4.

Now, the bundles against rhels7.4 are same as rhels7.3


bybai:
I am OK to merge it.

PR #3564, opened 2017-07-31
modify genesis testcase for if xcatmaster=nodename

junxiawang:
Modify genesis test case if node's definition's  xcatmaster=nodename. For in genesis environment there could not have  dns server and could not resolve nodename .


zet809:

Issue #3561, opened 2017-07-31
Failed to generate computes.log in rhels7.4


hu-weihua:
In redhat, xcat redirects 2 log files, ``cluster.log`` and ``computes.log``.
But in rhels7.4, xcat failed to generate ``computes.log``.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3566

immarvin:
verified by autotest

PR #3566, opened 2017-07-31
fix issue [FVT] Failed to generate computes.log in rhels7.4GA #3561

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3561

rh7.4 ships rsyslog 8.24.0, which does not contain or generate /etc/rsyslog.d/remote.conf, the settings to open udp and tcp 514 port listen should be done on  /etc/rsyslog.conf


bybai:
I am OK to merge it.

PR #3568, opened 2017-07-31
fix 13.1.0.100 is existed in some unknown system

bybai:
In auto test env,   there is one case give out failure with following error, since 13.1.0.0 network is only used in test case `confignetwork_secondarynic_thirdnic_multiplevalue_updatenode`, it is virtual network, I guess some other test CN running the testcase and  using the same ip, and did not clear the test env. We cannot find which system.  After I replaced this ip with a un-used ip, the test case is passed. And in this testcase, the test env will be deleted after the case is finished, so modify this ip in testcase.
```
c910f03c11k14: [I]: configeth on c910f03c11k14: eth1 changed, modify the configuration files
c910f03c11k14: bring up ip
c910f03c11k14: ERROR     : [/etc/sysconfig/network-scripts/ifup-eth] Error, some other host (DA:08:4C:8F:85:04) already uses address 13.1.0.100.
c910f03c11k14: [E]: ifup eth1 failed.
```


hu-weihua:
It is ok for me, agree to merge. thanks @bybai

Issue #3562, opened 2017-07-31
The line 71 of postbootscript "setbootfromnet" is wrong in rhels7.4GA


hu-weihua:
OS: rhels7.4 GA
xcat version: Version 2.13.6 (git commit 3139353de14074526135936490453597cd7d95c0, built Sat Jul 29 06:15:35 EDT 2017)

The line 71 of postbootscript ``setbootfromnet``  is wrong in rhels7.4.

    c910f02c08p07: Sun Jul 30 17:04:55 UTC 2017 Running postbootscript: setbootfromnet
    c910f02c08p07: ./setbootfromnet: line 71: [: too many arguments
    c910f02c08p07: ./setbootfromnet: The network interface NIC on the host 10.2.8.7 is not retrieved, pls check.\n

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3569

immarvin:
verified by autotest

PR #3569, opened 2017-07-31
fix issue The line 71 of postbootscript setbootfromnet is wrong in rhels7.4GA #3562

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3562

in rhels7.4:
````
[root@c910f02c08p05 xcat]# ip route
default via 10.0.0.103 dev eth0
10.0.0.0/8 dev eth0 proto kernel scope link src 10.2.8.5  #1 space between "eth0" and "proto"
````

in rhels7.3 or earlier version:
````
[root@c910f03c05k21 xcat-core]# ip route
default via 10.3.5.10 dev eth0
10.0.0.0/8 dev eth0  proto kernel  scope link  src 10.3.5.21  #2 spaces between "eth0" and "proto"
````


bybai:
I am ok to merge it.

PR #3573, opened 2017-08-01
Add 'Error' label when unable to get information

xuweibj:
```
# rpower testnode stat
[OpenBMC development support] Using this version of xCAT, ensure firmware level is at v1.99.6-0-r1, or higher.
testnode: Error: Unable to get information from openbmc table
```


zet809:

PR #3574, opened 2017-08-01
Remove the openbmc version information when running xCAT commands

whowutwut:
When running commands that invoke the openbmc plugin, a message prints out: 
```
[OpenBMC development support] Using this version of xCAT, ensure firmware level is at v1.99.6-0-r1, or higher.
```

This was intended to inform the user that at a specific version of xCAT, the firmware must be at a certain version due to function being integrated into the openbmc firmware.  As we get closer to GA, removing this message. It's also caused some confusion in our internal teams, so removing the message.


xuweibj:
I agreed to merge.

Issue #3567, opened 2017-07-31
`genimage` failed against xCAT shipped default osimage definition ubuntu16.04.2-ppc64el-netboot-compute


neo954:
xCAT 2.13.5 GA edition is used. Architecture is ppc64el.

    # genimage ubuntu16.04.2-ppc64el-netboot-compute 2>&1 | tee /tmp/genimage.out
    ... <<< omit many lines >>> ...
    Processing triggers for libc-bin (2.23-0ubuntu3) ...
    Processing triggers for initramfs-tools (0.122ubuntu8.8) ...
    bash
    nfs-common
    openssl
    isc-dhcp-client
    libc-bin
    linux-image-generic
    openssh-server
    openssh-client
    wget
    vim
    ntp
    ntpdate
    rsync
    busybox-static
    gawk
    dnsutils
    tar
    gzip
    xz-utils
     DEBIAN_FRONTEND=noninteractive chroot /install/netboot/ubuntu16.04.2/ppc64el/compute/rootimg apt-get -y  install bash nfs-common openssl isc-dhcp-client libc-bin openssh-server openssh-client wget vim ntp ntpdate rsync busybox-static gawk dnsutils tar gzip xz-utils
    Reading package lists...
    Building dependency tree...
    Reading state information...
    bash is already the newest version (4.3-14ubuntu1.1).
    tar is already the newest version (1.28-2.1ubuntu0.1).
    gzip is already the newest version (1.6-4ubuntu1).
    isc-dhcp-client is already the newest version (4.3.3-5ubuntu12).
    libc-bin is already the newest version (2.23-0ubuntu3).
    Some packages could not be installed. This may mean that you have
    requested an impossible situation or if you are using the unstable
    distribution that some required packages have not yet been created
    or been moved out of Incoming.
    The following information may help to resolve the situation:
    
    The following packages have unmet dependencies:
     vim : Depends: vim-common (= 2:7.4.1689-3ubuntu1.2) but 2:7.4.1689-3ubuntu1 is to be installed
           Depends: vim-runtime (= 2:7.4.1689-3ubuntu1.2) but it is not going to be installed
           Depends: libpython3.5 (>= 3.5.0~b1) but it is not going to be installed
    E: Unable to correct problems, you have held broken packages.
    Failed to install packages bash nfs-common openssl isc-dhcp-client libc-bin openssh-server openssh-client wget vim ntp ntpdate rsync busybox-static gawk dnsutils tar gzip xz-utils.

[...]

PR #3576, opened 2017-08-01
One symbolic link was added for ubuntu 16.04.2 support

neo954:
This pull request fix issue #3567


immarvin:
i am ok with the change

Issue #3535, opened 2017-07-25
rinstall/winstall -c doesn't open console


kcgthb:
Hi!

I'd like to report that `rinstall -c <node>` or `winstall -c <noderang>` don't seem to open any console once the provision starts, contrary to the man pages.

Example:
```
# rinstall sh-101-60
Provision node(s): sh-101-60
#
```
It's immediately back to the prompt, no console is opened. `rcons` and `wcons` work correctly when invoked individually.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3577

immarvin:
Verify result:
````
[root@c910f03c05k21 xcat-core]# lsdef testd -s
c910f03c05k27  (node)
c910f03c17k42  (node)
[root@c910f03c05k21 xcat-core]# rinstall testd shell --console
Error: rinstall -c/--console can only be run against one node! Please use winstall -c/--console for multiple nodes.

[root@c910f03c05k21 ~]# rinstall c910f03c17k42 shell -c
Provision node(s): c910f03c17k42
[Enter `^Ec?' for help]
Installing QEMU fb



Scanning USB
  OHCI: initializing
    USB Keyboard
````

PR #3577, opened 2017-08-01
fix issue rinstall/winstall -c doesn't open console #3535

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3535

correct the rinstall code logic to support -c

### DIFF
--- a/xCAT-client/bin/rinstall
+++ b/xCAT-client/bin/rinstall
@@ -21,6 +21,7 @@ use Getopt::Long;
 use xCAT::MsgUtils;
 use xCAT::Utils;
 use xCAT::Client;
+use xCAT::NodeRange;
 use Cwd;
 use strict;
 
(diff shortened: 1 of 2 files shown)


xCAT-client/bin/rinstall
```diff
@@ -82,9 +84,7 @@ if ($xCAT::Client::EXITCODE == 0)             # no errors
             exec("rcons $noderange");
```
chenglch: I'm a little curious about this line, does `rcons` support noderange ? Can we get the terminal for a bunch of nodes?

immarvin: the check of this is in rinstall.pm, https://github.com/xcat2/xcat-core/pull/3577/files?diff=unified#diff-51ffe05704f3e49d08346be47b480638R595

chenglch: OK, found it, one more question, if there are failure nodes and one successful node, what is the return code received by the client before running `rcons`?

immarvin: should be nonzero

chenglch: If so, why don't you check the size of the noderange in the plugins file instead of the client file?

immarvin: code updated

xCAT-server/lib/xcat/plugins/rinstall.pm
```diff
@@ -355,21 +356,17 @@ sub rinstall {
...
         my @lines = @$res;
         foreach my $line (@lines) {
             $rsp->{data}->[0] = $line;
-            if (($line =~ /: install/) or ($line =~ /: netboot/)) {
-                my $successnode;
-                my $restline;
-                ($successnode, $restline) = split(/:/, $line, 2);
-                $nodes{$successnode} = 0;
-                push @successnodes, $successnode;
+            if($line =~ /The (\S+) can not be resolved/){
```
chenglch: Not sure if this line can cover all of the error message from `nodeset`.  I have no idea for this.

xCAT-client/bin/rinstall
```diff
@@ -65,15 +66,23 @@ while ($arg =~ /^-/) {
 $cmdref->{noderange}->[0] = $arg;
 push(@{ $cmdref->{arg} }, @ARGV);
 
+my $startconsole=0;
+if(grep m/^-c$/,@ARGV){
```
chenglch: How about `--console`? Seems man page support this.

xCAT-server/lib/xcat/plugins/rinstall.pm
immarvin: i have no idea either, it is not a good idea to list and parse all possible error messages here, it might be coupled tightly between the nodeset code logic and rinstall.

immarvin: currently, ``install: xxxx`` or ``netboot; yyyy`` will be in the ``nodeset``  output, so this is not reliable to determine the successful nodes

xCAT-client/bin/rinstall
immarvin: code updated

Issue #3579, opened 2017-08-01
[fvt] "pgsqlsetup -i " does not work well in rhels7.3 and rhels7.4


hu-weihua:
``xcat version:``  Version 2.13.6 (git commit ea6498a638c31b1eb49167856e3fea6c0c164070, built Mon Jul 31 06:15:45 EDT 2017)
``OS:`` rhels 7.4 rc 1

``pgsqlsetup -i `` does not work well in rhels7.3 and rhels7.4.   The return code of "pgsqlsetup" is no-zore, but there is not error in the log.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3580

whowutwut:
@hu-weihua what is the error in the log ? 

@zet809 shouldn't we set this pending and allow @hu-weihua to verify and close out???

zet809:
hi, @whowutwut , this fix will be verified by automation regression test, it passed yesterday, so I close it.

hu-weihua:
@whowutwut, The output of this defect is a little strange, I could not get any error information from the output of ``pgsqlsetup -i`` , just the return code is non-zore.  After @immarvin did some debug, he thought this was a bug and need to change some code.

whowutwut:
Thanks for the info!

PR #3580, opened 2017-08-01
[fvt] 'pgsqlsetup -i ' does not work well in rhels7.3 and rhels7.4 #3579

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3579:

correct a typo


bybai:
I am OK to merge it.

PR #3582, opened 2017-08-01
Install Capture::Tiny for Z in CI test environment

hu-weihua:
Install ``Capture::Tiny`` for Z in CI test environment. 

Related pull request of z is #2927


zet809:
Agree to merge.

Issue #3583, opened 2017-08-01
Bug introduced in `bmcsetup` by previous commit


samveen:
Commit bb460dc5ceaa906c485a523de307840e4c639363 introduced a minor bug, where the value of `LOCKEDUSERS` was no longer considered while handling existing entries in the user list.

Line 432 is
```
if [ "$USERSLOT" == 0 ]; then USERSLOT=2; fi
```
while it should be
```
if [ "$USERSLOT" == 0 -o -z "$LOCKEDUSERS" ]; then USERSLOT=2; fi
```

zet809:
Hi, @samveen , do you know why need to lock user for  ""$IPMIMFG" == 19046 -a "$XPROD" == 13616"?

samveen:
@zet809 The code change I introduced only triggers in case values of $IPMICFG and $XPROD are not recognized from among the listed values.

For us, we want the `bmcsetup` script to overwrite the factory pre-configured user instead of disabling it. In xCAT `2.9.1`, the overwrite code-path triggered when `LOCKEDUSERS` was not defined (unrecognized $IPMICFG and $XPROD). However, now that code-path is missing, and from the description of  commit bb460dc5ceaa906c485a523de307840e4c639363 ` Provide a more adaptive approach to 'LOCKEDUSERS' in bmcsetup`, I can see that instead of an adaptive approach, the value LOCKEDUSERS became unused: 
```
[samveen@development xcat-core]$ git describe
2.13.5
[samveen@development xcat-core]$ grep LOCKEDUSERS xCAT-genesis-scripts/bin/bmcsetup 
        LOCKEDUSERS=1
        LOCKEDUSERS=1
   LOCKEDUSERS=1
        LOCKEDUSERS=1
    LOCKEDUSERS=1
```

Compare this to xCAT 2.11.1, for example:
```
[samveen@development xcat-core]$ git describe
2.11.1
[samveen@development xcat-core]$ grep LOCKEDUSERS xCAT-genesis-scripts/bin/bmcsetup 
        LOCKEDUSERS=1
        LOCKEDUSERS=1
        LOCKEDUSERS=1
    LOCKEDUSERS=1
if [ ! -z "$LOCKEDUSERS" ]; then
```

zet809:
OK, thx!

zet809:
I will close this issue since the code had been merged and will be included in 2.13.6 release.

PR #3584, opened 2017-08-01
Fix bug introduced into `bmcsetup` by commit `bb460dc`

samveen:
Commit bb460dc introduced a minor bug, where the value of LOCKEDUSERS
was no longer considered while handling existing entries in the user
list. This fixes that by adding back the check. Fixes Issue #3583 .


whowutwut:
@jjohnson42  @zet809  could you review this PR?

zet809:
Looks good to me, I will merge, thx @samveen

PR #3593, opened 2017-08-02
Optimize the performance for `getNodesAttribs` and `getAllNodeAttribs`

chenglch:
As `get_site_attribute` subroutine could not cache the value for
hierarchicalattrs configuration, this commit move the check logic of
`hierarchicalattrs` from `getNodeAttribs_nosub_returnany` to `getNodesAttribs`
and `getAllNodeAttribs` to optimize the the performance of bulk query.

After testing, this change will bring 50% performance improvement for
`getNodesAttribs` subroutine.


immarvin:
i am ok with the change

PR #3600, opened 2017-08-03
modify rinv testcase for different arch has different output

junxiawang:
Modify rinv testcase for different arch has different output if using wrong passwd .


xuweibj:

Issue #3408, opened 2017-07-06
dodiscovery sending findme packet 4 times for each node 


whowutwut:
In genesis, we see the following messages: 

```
<166>Jul  5 20:47:21 xcat.genesis.dodiscovery: Network configuration complete, commencing transmit of discovery packets
<166>Jul  5 20:47:21 xcat.genesis.dodiscovery: Beginning echo infomation to discovery packet file...
<166>Jul  5 20:47:22 xcat.genesis.dodiscovery: Discovery packet file is ready.
<166>Jul  5 20:47:22 xcat.genesis.dodiscovery: Sending the discovery packet to xCAT (172.20.253.27:3001)...
<166>Jul  5 20:47:22 xcat.genesis.dodiscovery: Sleeping 5 seconds...
<166>Jul  5 20:47:27 xcat.genesis.dodiscovery: Beginning echo infomation to discovery packet file...
<166>Jul  5 20:47:28 xcat.genesis.dodiscovery: Discovery packet file is ready.
<166>Jul  5 20:47:28 xcat.genesis.dodiscovery: Sending the discovery packet to xCAT (172.20.253.27:3001)...
<166>Jul  5 20:47:28 xcat.genesis.dodiscovery: Sleeping 5 seconds...
<166>Jul  5 20:47:34 xcat.genesis.dodiscovery: Beginning echo infomation to discovery packet file...
<166>Jul  5 20:47:35 xcat.genesis.dodiscovery: Discovery packet file is ready.
<166>Jul  5 20:47:35 xcat.genesis.dodiscovery: Sending the discovery packet to xCAT (172.20.253.27:3001)...
<166>Jul  5 20:47:35 xcat.genesis.dodiscovery: Sleeping 5 seconds...
<166>Jul  5 20:47:40 xcat.genesis.dodiscovery: Beginning echo infomation to discovery packet file...
<166>Jul  5 20:47:42 xcat.genesis.dodiscovery: Discovery packet file is ready.
<166>Jul  5 20:47:42 xcat.genesis.dodiscovery: Sending the discovery packet to xCAT (172.20.253.27:3001)...
<166>Jul  5 20:47:42 xcat.genesis.dodiscovery: Sleeping 5 seconds...
<166>Jul  5 20:47:47 xcat.genesis.dodiscovery: Restart network interfaces...
Killed old client process
```

Then on xCAT `cluster.log`, we also see 4 request obtained: 
```
[...]

PR #3601, opened 2017-08-03
fix issue dodiscovery sending findme packet 4 times for each node #3408

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3408

1. notify the node when the processing of its "findme" start and finish.
2. the node whose "findme" request is under processing won't send new "findme" request until:
     (1). its request has been processed but it does not receive the "restart" response.
     (2). its request is still under processing but the wait time exceeds the timeout value(180s by default)

### DIFF
--- a/xCAT-server/lib/xcat/plugins/zzzdiscovery.pm
+++ b/xCAT-server/lib/xcat/plugins/zzzdiscovery.pm
@@ -19,14 +19,32 @@ sub process_request {
     my $cb    = shift;
     my $doreq = shift;
     if ($req->{command}->[0] eq 'findme') {
+
         if (!defined($req->{discoverymethod}) or !defined($req->{discoverymethod}->[0]) or ($req->{discoverymethod}->[0] eq 'undef')) {
             my $rsp = {};
             $rsp->{error}->[0] = "The discovery request can not be processed";
             $cb->($rsp);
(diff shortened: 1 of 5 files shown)


xCAT-server/lib/xcat/plugins/zzzdiscovery.pm
```diff
@@ -19,14 +19,32 @@ sub process_request {
...
+        #now, notify the node that its findme request has been processed
+        xCAT::MsgUtils->message("S","Notify $client_ip that its findme request has been processed");
+        my $sock = new IO::Socket::INET(
+            PeerAddr => $client_ip,
+            PeerPort => '3001',
+            Timeout  => '1',
+            Proto    => 'tcp'
+        );
+        unless ($sock) { 
+            xCAT::MsgUtils->message("S", "Failed to notify $client_ip that its findme request has been processed");
```
zet809: If no sock is generated, shall return immediately.
The nodediscover will be dealt immediate after the findme request is resolved. Then, the IP of client may have been changed to desired static IP. 
So, I think the processed flag is useless.

zet809:
hi, @immarvin , it seems the issue that findme send out multiple times is not resolved. As the first findme request won't be dealt until the switch cache is updated. During this time, no `processing` flag is able to send back to client, then, the client will send out other findme requests.

zet809:

PR #3604, opened 2017-08-03
mknb is not generated on service node (ppc64le) after the initial installation

### LINKED ISSUE #3603: mknb is not generated on service node (ppc64le) after the initial installation.
After ppc64le service node installed,  we found that the genesis netboot is not generated correctly.
Check the log (/var/log/xcat/commands.log), and found that:
```
[Date]       2017-08-02 23:14:58
[ClientType] cli
[Request]    mknb ppc64le
[Response]
The arch:ppc64le is not supported at present, pls use "ppc64" instead
[ElapsedTime] 0 s
```

robin2008:
Fix #3603, If using ppc64le/ppc64el for mknb, just use 'ppc64' as a replacement 

UT:
```
# mknb ppc64le
The arch:ppc64le is not supported at present, use "ppc64" instead
Creating genesis.fs.ppc64.gz in /tftpboot/xcat
```


zet809:

PR #3616, opened 2017-08-07
Fix typo in xcatprobe discovery precheck

xuweibj:
If xcat-genesis-base-ppc64 is not built in fedora environment, will print warning msg.
```
# xcatprobe discovery -n test_node -m mtms -c
Genesis files are avaliable                                                                                       [ OK ]
xcat-genesis-base-ppc64 is not built in environment fedora on ppc64.                                                    [WARN]
```


zet809:

PR #3629, opened 2017-08-08
Support rspconfig command to change PDU's ip address, netmask and hostname

cxhong:
Add support to use rspconfig command to change PDU's ip address/netmask and hostname
Example of rspconfig command:
````
rspconfig coralpdu hostname=coralpdu
rspconfig coralpdu ip=1.1.1.1,netmask=255.0.0.0
````
This PR also supports pdudiscover command to discover PDU and add --setup options will config PDU: setup passwordless, change hostname and change ip address/netmask.

````
pdudiscover --range dhcpip_dynamic_range --setup

````
NOTE:  Since we don't have coralpdu on the lab,  the function is not fully running on real hardware (set up ip address).

### DIFF
--- a/xCAT-server/lib/xcat/plugins/pdu.pm
+++ b/xCAT-server/lib/xcat/plugins/pdu.pm
@@ -177,6 +177,8 @@ sub process_request
         my $subcmd = $exargs[0];
         if ($subcmd eq 'sshcfg') {
             process_sshcfg($noderange, $subcmd, $callback);
+        }elsif ($subcmd =~ /=/) {
+            process_netcfg($request, $subreq, $subcmd, $callback);
         } else {
             $callback->({ errorcode => [1],error => "The input $command $subcmd is not support for pdu"});
         }
@@ -467,6 +469,87 @@ sub connectTopdu {
 
 }
 
+#-------------------------------------------------------
+
+=head3  process_netcfg 
+   
+    Config hostname of PDU
+    Config ip/netmask of PDU via PduManager command 
+      PduManager is a tool for CoralPdu to manager the PDU.
+      * /dev/shm/bin/PduManager -h
+          '-i' set PDU system IP
+          '-n' set system ip netmask. e.g.:PduManager -i xxx.xxx.xxx.xxx -n xxx.xxx.xxx.xxx
+
+    example:  rspconfig coralpdu hostname=coralpdu
+              rspconfig coralpdu ip=1.1.1.1,netmask=255.0.0.0
+
+=cut
+
+#-------------------------------------------------------
+sub process_netcfg {
+    my $request = shift;
+    my $subreq    = shift;
+    my $subcmd    = shift;
+    my $callback = shift;
+    my $hostname;
+    my $ip;
+    my $netmask;
+    my $args;
+
+    my $nodes = $request->{node};
+    my $rsp = {};
+
+    xCAT::MsgUtils->message("I", "set hostname ");
+    my @cmds = split(/,/,$subcmd);
+    foreach my $cmd (@cmds) {
+        my ($key, $value) = split(/=/, $cmd);
+        if ($key =~ /hostname/) {
+            $hostname = $value;
+            $args = "echo $hostname > /etc/hostname;/etc/init.d/hostname.sh";
+            xCAT::Utils->runxcmd({ command => ['xdsh'],
+                                   node => $nodes,
+                                   arg => [ $args ] }, $subreq, 0, 1);
+            if ($::RUNCMD_RC != 0) {
+               xCAT::MsgUtils->message("I", "xdsh command to set hostname failed");
+            }
+        }elsif ($key =~ /ip/) {
+            $ip = $value;
+        } elsif ($key =~ /netmask/) {
+            $netmask = $value;
+        } else {
+            xCAT::MsgUtils->message("I", "rspconfig $cmd is not support yet");
+        }
+    }
+
+    $args = "/dev/shm/bin/PduManager ";
+    my $opt;
+    if ($ip) {
+        $opt = "-i $ip ";
+    }
+    if ($netmask) {
+        $opt = $opt . "-n $netmask";
+    }
+    if ($opt) {
+        my $dshcmd = $args . $opt ;
+        #comment this for now, coralPDU on the lab is not support PduManager yet
+        my $output = xCAT::Utils->runxcmd({ command => ['xdsh'], node => $nodes, arg => [ $dshcmd ] }, $subreq, 0, 1);
+        if ($::RUNCMD_RC != 0) {
+            xCAT::SvrUtils::sendmsg("@$output", $callback);
+        }
+    }
+}
+
+#-------------------------------------------------------
+
+=head3  process_sshcfg 
+   
+    Config passwordless for coralpdu 
+
+    example:  rspconfig coralpdu sshcfg 
+
+=cut
+
+#-------------------------------------------------------
 sub process_sshcfg { 
     my $noderange = shift;
     my $subcmd = shift;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: xCAT-server/lib/xcat/plugins/pdu.pm [hunk 1]
```diff
@@ -467,6 +469,87 @@ sub connectTopdu {
...
+        }elsif ($key =~ /ip/) {
+            $ip = $value;
+        } elsif ($key =~ /netmask/) {
+            $netmask = $value;
+        } else {
+            xCAT::MsgUtils->message("I", "rspconfig $cmd is not support yet");
+        }
+    }
+
+    $args = "/dev/shm/bin/PduManager ";
```
zet809: Do you need to add PduManager as a pre required software? Or, any plan to update the document?

xCAT-server/lib/xcat/plugins/pdu.pm
```diff
@@ -467,6 +469,87 @@ sub connectTopdu {
...
+    my $hostname;
+    my $ip;
+    my $netmask;
+    my $args;
+
+    my $nodes = $request->{node};
+    my $rsp = {};
+
+    xCAT::MsgUtils->message("I", "set hostname ");
+    my @cmds = split(/,/,$subcmd);
```
zet809: The behavior is a little different with our normal logic. We will use the args array directly. Why don't keep it consistent?

xCAT-server/lib/xcat/plugins/pdu.pm
```diff
@@ -177,6 +177,8 @@ sub process_request
         my $subcmd = $exargs[0];
         if ($subcmd eq 'sshcfg') {
             process_sshcfg($noderange, $subcmd, $callback);
+        }elsif ($subcmd =~ /=/) {
```
zet809: So, for configuring ssh, there is no "enable or disable" need to be specified, but for ip/netmask..., it need a '=' to specify the parameter, feeling a little inconsistent. 
And why only the first arg is dealt with? If the input is 'rspconfig node ip=xxx,net=xxx hostname=xxx', there will nothing returned for what happened about hostname=xxx?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -1436,6 +1439,35 @@ sub switchsetup {
...
+            my $cmd = "rspconfig $pdu sshcfg"; 
+            xCAT::Utils->runcmd($cmd, 0);
+            my $ip = $nodehash->{$pdu}->[0]->{otherinterfaces};
+            my $mask;
+            foreach my $net (@nets) {
+                if (xCAT::NetworkUtils::isInSameSubnet( $net->{'net'}, $ip, $net->{'mask'}, 0)) {
+                    $mask=$net->{'mask'};
+                }
+            }
+            $cmd = "rspconfig $pdu hostname=$pdu,ip=$otherinterfaces,netmask=$mask";
```
zet809: Where does the $otherinterfaces define?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -1436,6 +1439,35 @@ sub switchsetup {
...
+            xCAT::Utils->runcmd($cmd, 0);
+            if ($::RUNCMD_RC == 0) {
+                xCAT::Utils->runxcmd({ command => ['chdef'], arg => ['-t','node','-o',$pdu,"ip=$ip","otherinterfaces="] }, $sub_req, 0, 1);
```
zet809: Hi, @cxhong , a little confusing here:
1. For rspconfig sshcfg, it will try to ping ip or otherinterface to make sure which IP is available.
2. But for rspconfig hostname=xxx, it will use 'xdsh pdu' directly, how does it know either ip or otherinterface shall be used?
    From my point of view, it shall use the $otherinterface to connect to pdu and set $ip to it, doesn't it?

whowutwut:
One concern here is that we are mixing commands. rspconfig stands for "remote service processor config"

We have to make sure we are consistent or at least define the scope of what is planned to be supported. How would users know which objects rspconfig can run against? We need to make sure if it's some nodetype that is not supported that we block the command

xCAT-server/lib/xcat/plugins/pdu.pm
cxhong: PduManager is a tool to manage the PDU,  it already installed in the coral pdu .

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: discover ip (or dhcp ip from switchdiscover command) address is stored to otherinterfaces attribute

xCAT-server/lib/xcat/plugins/pdu.pm
cxhong: yeh, I like to change it to keep consistent.  I didn't find a good example for this.
how the array args will look like?

xCAT-server/lib/xcat/plugins/pdu.pm
cxhong: on the rspconfig man page,  there are two options: 
````
 sshcfg={enable|disable}
           Enable or disable SSH

       sshcfg
           Copy SSH keys.
`````
I only need to copy SSH keys to pdu,that's why I chose no "enable or disable"

we can change rspconfig to array args

File: xCAT-server/lib/xcat/plugins/switchdiscover.pm (hunk 5 above)
Comment (cxhong) [inline xCAT-server/lib/xcat/plugins/switchdiscover.pm]: switchdiscover --setup will always run sshcfg first, then change hostname and ip.

if user just issue rspconfig command,  hopefully they configured ssh passwordless already,   I can add error message to ask them to run "rspconfig ssh" first.

File: xCAT-server/lib/xcat/plugins/pdu.pm (hunk 1 above)
Comment (zet809) [inline xCAT-server/lib/xcat/plugins/pdu.pm]: Oh, OK, I see, Thx

File: xCAT-server/lib/xcat/plugins/switchdiscover.pm (hunk 4 above)
Comment (zet809) [inline xCAT-server/lib/xcat/plugins/switchdiscover.pm]: I see, the for this variable, it is not defined yet.

File: xCAT-server/lib/xcat/plugins/pdu.pm (hunk 2 above)
Comment (zet809) [inline xCAT-server/lib/xcat/plugins/pdu.pm]: It looks like this:
```
    my $extrargs  = $request->{arg};
    my @exargs    = ($request->{arg});
    if (ref($extrargs)) {
        @exargs = @$extrargs;
    }
```
The @exargs here store all the options passed in.

zet809:
hi, @whowutwut , you are right that for rspconfig, it is used to configure the management module of servers, switch...
For for PDU, I think it is OK to support configuring IP, hostname, sshcfg since these function are useful to manage or operate the PDU.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
zet809: Hi, @cxhong , in line 1455, it had read out the 'otherinterfaces' attribute to variable $ip, so, my question is, how does the $otherinterfaces variable get value in this line?

zet809: The option of rspconfig had been changed ',' to ' ', you shall modify here also.

cxhong: thanks

zet809:
Hi, @cxhong , the pull request haven't been changed since last discussion, and the PDU using for CORAL will be a little different, will this PR still valid? If not, we can close it.

zet809:
Looks good to me, I will merge it after the CI is finished. Thx @cxhong 
For the refinement for the output of rinv and rvitals, @cxhong will fix it in next release.

PR #3651, opened 2017-08-09
Revise the man page of noderange. Describe the special characters need to be escaped more clear

neo954:
This pull request fix issue #3319


immarvin:
i am ok with the change

daniceexi:
Mostly this issue happens on which shell?

neo954:
@daniceexi, I think on all popular used shells, user will be affected with issue #3319. The list will including original Bourne shell, Bash, Ksh, csh, Tcsh, Z shell. All these shells use `*` and `~` for pathname pattern matching. Except csh, all other shells use `[` and `]` for filename globbing.

whowutwut:
This one i think missed running `create_man_pages.py` to update the .rst

PR #3654, opened 2017-08-09
using sshpass (if availble) connect to openbmc console

cxhong:
connecting to openbmc console require user manually send SSH key over to openbmc.  

Sshpass is a tool for non-interactivly performing password authentication with SSH's so called "interactive keyboard password authentication".   xCAT will use this tool to connect openbmc console without user enter password.   currently, if user like to use this tool, needs to download sshpass from online http://rpmfind.net/linux/rpm2html/search.php?query=sshpass
 , then install to conserver server.  for near future,  we may add this package to xcat-dep.

### DIFF
--- a/xCAT-server/share/xcat/cons/openbmc
+++ b/xCAT-server/share/xcat/cons/openbmc
@@ -91,7 +91,12 @@ if ($ENV{SSHCONSOLEPORT}) {
     $sshport= $ENV{SSHCONSOLEPORT};
 }
 
-print "If the console cannot connect, please verify whether ssh keys has been configured on the bmc for $username user\n";
+#user needs to download sshpass rpm if need to use it
+if (-x '/usr/bin/sshpass') {
+    exec "/usr/bin/sshpass -p $password ssh -p $sshport -l $username $bmcip";
+} else {
+    exec "ssh -p $sshport -l $username $bmcip";
+}
 
-exec "ssh -p $sshport -l $username $bmcip";
+
 
--- a/xCAT-server/share/xcat/cons/openbmc
+++ b/xCAT-server/share/xcat/cons/openbmc
@@ -91,10 +91,13 @@ if ($ENV{SSHCONSOLEPORT}) {
     $sshport= $ENV{SSHCONSOLEPORT};
 }
 
-#user needs to download sshpass rpm if need to use it
+# To automatically connect to the console without the need to send over the ssh keys, 
+# ensure sshpass is installed on the Management and/or Service Nodes.
 if (-x '/usr/bin/sshpass') {
     exec "/usr/bin/sshpass -p $password ssh -p $sshport -l $username $bmcip";
 } else {
+    # This will prompt the user to enter the password for the BMC 
+    # if ssh keys are not sent and sshpass is not being used
     exec "ssh -p $sshport -l $username $bmcip";
 }
 


xCAT-server/share/xcat/cons/openbmc
```diff
@@ -91,7 +91,12 @@ if ($ENV{SSHCONSOLEPORT}) {
     $sshport= $ENV{SSHCONSOLEPORT};
 }
 
-print "If the console cannot connect, please verify whether ssh keys has been configured on the bmc for $username user\n";
+#user needs to download sshpass rpm if need to use it
+if (-x '/usr/bin/sshpass') {
+    exec "/usr/bin/sshpass -p $password ssh -p $sshport -l $username $bmcip";
```
whowutwut: Suggest putting comment here:  

```
# To automatically connect to the console without the need to send over the ssh keys, ensure sshpass is installed on the Management and/or Service Nodes. 
```

xCAT-server/share/xcat/cons/openbmc
```diff
@@ -91,7 +91,12 @@ if ($ENV{SSHCONSOLEPORT}) {
...
+} else {
+    exec "ssh -p $sshport -l $username $bmcip";
```
whowutwut: Add comment here 

```
# This will prompt the user to enter the password for the BMC if ssh keys are not sent and sshpass is not being used
```

xCAT-server/share/xcat/cons/openbmc
```diff
@@ -91,7 +91,12 @@ if ($ENV{SSHCONSOLEPORT}) {
     $sshport= $ENV{SSHCONSOLEPORT};
 }
 
-print "If the console cannot connect, please verify whether ssh keys has been configured on the bmc for $username user\n";
+#user needs to download sshpass rpm if need to use it
```
whowutwut: Suggest removing this command and replacing with the below suggestions.

xCAT-server/share/xcat/cons/openbmc
cxhong: modified

xCAT-server/share/xcat/cons/openbmc
cxhong: modified

xCAT-server/share/xcat/cons/openbmc
cxhong: ok

zet809:
Hi, @cxhong , if we use `rspconfig <node> sshcfg` to copy public key to OpenBMC, do we still need to pass password to ssh?

whowutwut:
@zet809 In order for rcons to work seamlessly we have to send over the ssh keys.  However, an alternative to this is to use the `sshpass` command.  Before we decide if we want to add another binary into our xcat-dep, the intent here is that if the admin has already downloaded `sshpass`  onto his system ... we will attempt to use that to send the password.    If not, then we will default to just normal ssh, and if the keys are not sent, then it will prompt the user for the password before opening up the console.

whowutwut:
@zet809 If no objections will merge this Monday.

zet809:

gurevichmark:
I might be too late, since the pull request is merged, but for openbmc command `rspconfig sshcfg` I was able to use existing xCAT function to copy sshkeys without the need for `sshpass`. 
I used `xCAT::RShellAPI::run_remote_shell_api()` in PR #3376 :
```
# Login and append content of the read in id_rsa.pub file to the authorized_keys file on BMC
+        my $output = xCAT::RShellAPI::run_remote_shell_api($bmcip, $userid, $userpw, 0, 0, "mkdir -p ~/.ssh; echo \"$id_rsa_pub_contents\" >> ~/.ssh/authorized_keys");
```

PR #3662, opened 2017-08-10
Adding some packages that would help debug xCAT issues from service nodes

whowutwut:
By default some of the xcatprobe utilities and diagnostics to validate configuration should be added to the service node package list so they do not need to be installed after the fact...

### DIFF
--- a/xCAT-server/share/xcat/install/rh/service.rhels7.ppc64le.pkglist
+++ b/xCAT-server/share/xcat/install/rh/service.rhels7.ppc64le.pkglist
@@ -34,3 +34,6 @@ perl-DBD-Pg
 wget
 util-linux-ng
 net-tools
+pciutils
+bind-utils
+tcpdump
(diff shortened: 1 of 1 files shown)


zet809:
Thx, @whowutwut , I can not think out more about it.
@immarvin and @xuweibj , any idea?

immarvin:
``strace``

whowutwut:
@immarvin What a scenario we would use  `strace` for to debug xCAT?

immarvin:
@whowutwut , ``strace`` is quite useful while debugging issues such as the child processes of xcat hang or die due to some critical issues. I once encountered some PMR that the "install monitor" service died for some reason, ``strace -p <install monitor pid>`` proved to be a sharp tool in this scenario.

whowutwut:
Thanks @immarvin I added it to the commit.

whowutwut:
If no issues, can we merge this? @immarvin @zet809 ?

zet809:
OK, will merge it. Thx!

PR #3664, opened 2017-08-10
Remove workaround for OpenBMC discovery to provide a unique discover node name

whowutwut:
With PR #3525 no longer need this workaround since we are uisng the MAC address to create a unique name.


xuweibj:

Issue #3682, opened 2017-08-14
bmcdiscover exit with socketpair error msg when against over 1000 nodes


xuweibj:
command: time bmcdiscover --range 10.100.71-90.1-100 (2000 nodes)

```
Error: bmcdiscover plugin bug, pid 20170, process description: 'xcatd SSL: bmcdiscover for root@localhost: bmcdiscover instance' with error 'socketpair: Too many open files at /opt/xcat/lib/perl/xCAT_plugin/bmcdiscover.pm line 657.
```

xuweibj:
Verified. Close this issue.

PR #3685, opened 2017-08-15
Fix the error caused by too many open file descriptors

chenglch:
The code in bmcdiscover do not close socket pair correctly which
causes `Too many open files` error from system. This commit maintains
a hash map to locate the file descriptors for child processes, close
these files when the child processes exit.

fix-issue: #3682


whowutwut:
@chenglch  :+1:  How did we find this problem?

whowutwut:
I see the issue linked now...  :)

xuweibj:
Looks good and have tested, agree to merge.

PR #3696, opened 2017-08-16
Set xcatmaster to facing IP in priority when the node is managed by service node.

### LINKED ISSUE #3428: kickstart file is not correct in hierarchy mode
In an hierarchy mode with service pool:
site.master = 172.21.

For nodes under sn01 and sn02,  and with below settings:
    noderes.servicenode=sn01,sn02
    noderes.xcatmaster=''

In the design (not using share /install),  the kickstart file generated on sn01 and sn02 should using the CN facing IP address (172.20.x.x),  but now  it contains the IP of site.master.  It will cause installation failed and loop install again and again.

BTW:
This could be workaround by set noderes.xcatmaster='172.20.x.x' explicitly.

robin2008:
…ananged by service node.

The previous priority:  noderes.xcatmaster -> site.master -> facing IP
In such design, it will bring issue when xcatmaster is not set and hierarchy layout (if CN cannot access MN),  here we make some changes on it  as below:

Fix #3428, set xcatmaster to facing IP in priority when the node is managed by service node.

```
    #the logic to determine the $ENV{XCATMASTER} confirm to the following priority(from high to low):
    ## 1, the "xcatmaster" attribute of the node
    #Route A: the node is managed by service node
    ## A.2, the ip address of the mn facing the compute node
    ## A.3, the site.master
    #Route B: the node is not managed by service node
    ## B.2, the site.master
    ## B.3, the ip address of the mn facing the compute node
```

Finally, we decide that only using one option,  always using facing IP in high priority.
```
    #the logic to determine the $ENV{XCATMASTER} confirm to the following priority(from high to low):
    ## 1, the "xcatmaster" attribute of the node
    ## 2, the ip address of the mn/sn facing the compute node
    ## 3, the site.master
```

### DIFF
--- a/xCAT-server/lib/perl/xCAT/Template.pm
+++ b/xCAT-server/lib/perl/xCAT/Template.pm
@@ -86,7 +86,6 @@ sub subvars {
     ## 2, the ip address of the mn/sn facing the compute node
     ## 3, the site.master
     my $master;
-    my $sitemaster;
 
     #the "xcatmaster" attribute of the node
     my $noderestab = xCAT::Table->new('noderes');
@@ -112,7 +111,7 @@ sub subvars {
         my @masters = xCAT::TableUtils->get_site_attribute("master");
         my $tmp     = $masters[0];
         if (defined($tmp)) {
-            $sitemaster = $tmp;
+            $master = $tmp;
         }
     }
 
--- a/xCAT-server/lib/perl/xCAT/Template.pm
+++ b/xCAT-server/lib/perl/xCAT/Template.pm
@@ -82,16 +82,23 @@ sub subvars {
     close($inh);
 
     #the logic to determine the $ENV{XCATMASTER} confirm to the following priority(from high to low):
-    #the "xcatmaster" attribute of the node
-    #the site.master
-    #the ip address of the mn facing the compute node
+    ## 1, the "xcatmaster" attribute of the node
+    #Route A: the node is managed by service node
+    ## A.2, the ip address of the mn facing the compute node
+    ## A.3, the site.master
+    #Route B: the node is not managed by service node
+    ## B.2, the site.master
+    ## B.3, the ip address of the mn facing the compute node
     my $master;
+    my $sitemaster;
+    my $parents;
 
     #the "xcatmaster" attribute of the node
     my $noderestab = xCAT::Table->new('noderes');
-    my $et = $noderestab->getNodeAttribs($node, ['xcatmaster']);
-    if ($et and $et->{'xcatmaster'}) {
+    my $et = $noderestab->getNodeAttribs($node, ['servicenode','xcatmaster']);
+    if ($et) {
         $master = $et->{'xcatmaster'};
+        $parents = $et->{'servicenode'};
     }
 
     unless ($master) {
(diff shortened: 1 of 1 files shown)


robin2008:
UT: (/install is not shared on SN)
```
# lsdef -t site -o clustersite| grep master
    master=10.3.5.20

# chdef c910f03c05k24 servicenode= xcatmaster=
# nodeset c910f03c05k24 osimage=rhels7.3-ppc64le-install-compute -V
c910f03c05k24: [c910f03c05k20]: install rhels7.3-ppc64le-compute
c910f03c05k24: [c910f03c05k29]: install rhels7.3-ppc64le-compute

Check the /install/autoinst/c910f03c05k24,  only the 10.3.5.20 inside.
#chdef c910f03c05k24 servicenode= xcatmaster=
# nodeset c910f03c05k24 osimage=rhels7.3-ppc64le-install-compute -V
Check the /install/autoinst/c910f03c05k24,  only the 10.3.5.29 inside.

```

daniceexi:
@robin2008 I remember we discussed the possibility to cache the facing ip in certain place. Is it possible to have a key-value DB to accelerate this kind of operation?

robin2008:
1,  to use "getipaddr" is to using its cache to accelerate the IP resolution. (it will reduce time greatly when most of the nodes using the same xcatmaster)

2,  For "my_ip_facing",  as the key is the node IP address, each node has different IP (not netmask for it), we cannot use cache for it.  Normally one plugin, we will call "my_ip_facing" for a node once, and for different plugin, as they are not in same process, we cannot share the cache.

The possible caching is that the result of "ip addr", we can get the result before the foreach node loop and pass it into "my_ip_facing".

immarvin:
hi @robin2008 , I think Route#A and #B should be the same logic. if the nodelist.xcatmaster is not set, the CN facing ip should always takes the precedence, what case do you think that site.master is different with the CN facing ip of MN?

immarvin:
the site.master is initialized on xcat installation or "xcatconfig -i/-d", the code slice is:
````
#cat `which xcatconfig`
...
1070 sub initDB
1071
1072 {
1073
1074     my $hname = `hostname`;
1075     chomp $hname;
1076
1077     # some values common to both AIX & Linux
...
1082     my $master;
1083     $master = xCAT::NetworkUtils->getipaddr($hname);
1084     if (!$master) {
1085         xCAT::MsgUtils->message("E", "Hostname resolution for $hname failed, setting the site.master to NORESOLUTION for now,      after the xCAT installation is finished, fix this hostname resolution problem, change the site.master to the correct value a     nd then resart xcatd.");
1086         $master = "NORESOLUTION";
1087     }
````

there is no guarantee that the site.master will be the CN facing ip of MN.  it might be "NORESOLUTION", and will pass the logic in this PR and be written to osimage template file

robin2008:
For global shared mode,  we may want a central cluster service for CN. In that case,  all CN will not set `servicenode` attribute.

So using two router here is to make a compatible method to meet  some different requirements.

immarvin:
if all cn does not set the ``servicenode`` attribute,  the MN will provide installation service, the MN's CN facing IP will work. If user expect another installation server instead of MN, the nodelist.xcatmaster instead of site.master should be set. 

if site.master is "NORESOLUTION", the string will be written to kickstart file literally and there is no error message

robin2008:
Updated that facing IP is in high priority always.

whowutwut:
In last weeks working session with LBS, there are some talks about clearly defining the cause/effects of the servicenode attribute and the xcatmaster attribute when we are using disjointdhcp=true and when disjointdhcp=false and how this relates to service node dhcp handling and next server in tftpboot 

Is there a good way we can take this offline and create a table that clearly defines it.. @cxhong is out this week...

immarvin:
i am ok with the change

immarvin:
hi @whowutwut , the behavior is a little confusing in the current documentation, xCAT developers might be clear with the details, but it will be helpful if there are some doc to make xCAT users clear about this

I will merge this PR first, I created a ticket on the DOC in https://github.com/xcat2/xcat-core/issues/3738, the discussion can be continued there.

Issue #3690, opened 2017-08-15
rflash support for Supermicro based firmware. New product id and relative path for data dir


gurevichmark:
1. Dave Shapiro was not able to do `rflash` of Supermicro nodes on his system.
After some debugging turns out his system returns 2437 for product id. Our code was only checking for 0 or 2355 for product id. After some discussion, turns out Chassis ID could also be 9006-12C and 5104-22C. Our code currently only checks for 8001-22C and 9006-22C

2. Need to support relative path for `-d=<data directory>` flag of `rflash` command.

whowutwut:
@david-shapiro FYI, Mark opened this issue...

whowutwut:
@gurevichmark What is 2437 product ID?

gurevichmark:
I think 2437 is yet another version of Boston machine.

gurevichmark:
After talking to Dave I think Product ID for Boston nodes is 0 on our systems:
```
[root@c910f02c05p03 xcat]# rinv boston01,boston02 | grep "Product ID"
boston01: Product ID: 0
boston02: Product ID: 0
[root@c910f02c05p03 xcat]#
```

However after the VPD update, it should return 2437:
```
[root@c685mgt02 xcat]# rinv c685f1n[01-09] | grep "Product ID"
c685f1n09: Product ID: 2437
c685f1n07: Product ID: 2437
c685f1n02: Product ID: 0
c685f1n08: Product ID: 2437
c685f1n05: Product ID: 2437
c685f1n06: Product ID: 2437
c685f1n04: Product ID: 2437
c685f1n03: Product ID: 2437
c685f1n01: Product ID: 2437
[root@c685mgt02 xcat]#
```

PR #3702, opened 2017-08-16
handle rflash relative path and more Boston product ids

gurevichmark:
This fixes issue #3690 

Allow relative path specification for `-d=<directory>`

Handle additional Boston Product ID and Chassis Part Numbers.


whowutwut:
Thanks @gurevichmark Let's merge this.  We are seeing more people hitting this issue with Supermicro servers.

PR #3707, opened 2017-08-17
Skip built-in MAC addresses on switches

mattaezell:
Newer Cumulus versions show a MAC address on each port. Those MACs aren't discoverable devices, so skip them to clean up the 'xcatprobe switch_macmap' output.

### DIFF
--- a/perl-xCAT/xCAT/MacMap.pm
+++ b/perl-xCAT/xCAT/MacMap.pm
@@ -734,6 +734,9 @@ sub refresh_switch {
         xCAT::MsgUtils->message("I", "MTU information is not availabe for this switch $switch");
     }
 
+    # get port state
+    my $mactostate = walkoid($session, '.1.3.6.1.2.1.17.7.1.2.2.1.3', silentfail => 1, verbose => $self->{show_verbose_info}, switch => $switch, callback => $self->{callback});
+
     #Above is valid without community string indexing, on cisco, we need it on the next one and onward
     my $iftovlanmap = walkoid($session, '.1.3.6.1.4.1.9.9.68.1.2.2.1.2', silentfail => 1, verbose => $self->{show_verbose_info}, switch => $switch, callback => $self->{callback}); #use cisco vlan membership mib to ascertain vlan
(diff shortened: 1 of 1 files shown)


mattaezell:
I've only tested this against a Cumulus switch and a Mellanox SX1410 (though not much was plugged into it), so I hope you guys have some other hardware in the lab you can try this against.

perl-xCAT/xCAT/MacMap.pm
```diff
@@ -807,8 +810,11 @@ sub refresh_switch {
                 my $vlan      = @tmp[0];
                 my @mac       = @tmp[ -6 .. -1 ];
                 my $macstring = sprintf("%02x:%02x:%02x:%02x:%02x:%02x", @mac);
-                push @{ $index_to_mac{$index} }, $macstring;
-                push @{ $index_to_vlan{$index} }, $vlan;    
+                # Skip "permanent" ports
+                if (!defined($mactostate->{$_}) || $mactostate->{$_} != 4) {
```
zet809: Hi, @mattaezell , thx very much for the enhance, it seems 4 means self address, and 3 means learned address, do you think using 3 is more reasonable?

mattaezell: Hi @zet809.  I wrote it using the logic "I'm seeing 4/self entries and I don't want them" instead of "I only want 3/learned entries" because I wasn't sure if we would want to show the MACs from other states (1/other, 2/invalid, or 5/mgmt).  I only see 3 or 4 on my switches, so I'm open to either method.  Let me know if you want me to update the PR

cxhong:
I verified this PR on our test system, for the BNT, Mellanox sx1410 and sx1410 switches.  it returns state 3 or 4.  I think  this PR looks fine.  Thanks @mattaezell .
@zet809 , if you don't have other issue, can u merge it?

zet809:
Looks good to me, thx @mattaezell and @cxhong , I agree to merge.

perl-xCAT/xCAT/MacMap.pm
zet809: OK, I agree with you, @mattaezell , thx!

PR #3715, opened 2017-08-18
fix some typo in case makedhcp_remote_network

tingtli:
fix some typo in case makedhcp_remote_network


hu-weihua:
It is ok for me, agree for me, thanks @tingtli

Issue #3295, opened 2017-06-20
[openbmc] rspconfig - support setting the ip address to DHCP for the BMC


whowutwut:
This task tracks the support for adding to rspconfig to set the BMC IP address to DHCP. 

* ~~Power off compute hosts~~  - Currently for OpenPower, we do not power off the compute node when we reset the BMC IP address back to DHCP.  Should we consider this? 
* Reset the BMC IP address back to DHCP

zet809:
Hi, @whowutwut , I don't think we need to power off the compute node when reset BMC back to DHCP. 1. If we keep host is powering on, we can check the dynamic IP BMC got through the host.
2. From my point of view, the changing of BMC ip shall not affect the host.

whowutwut:
@zet809 Thanks for the comments, that is a valid point.  In the re-discovery scripts that I have written for P8 OpenPower nodes,  I normally just use raw commands  directly to BMC to reset to DHCP, but I will change those to use `rpower <> off` then `rspconfig`

xuweibj:
https://github.com/openbmc/docs/blob/master/dbus-interfaces.md#orgopenbmcnetworkmanager

whowutwut:
@xuweibj did you look into that URL? If you think you have a command to execute I can help execute it if we think we may mess up the configuration where we will not be able to recover IP  address.  I have a serial cable plugged into the box elder board on mid05tor12cn02 so that would be the machine we should try this on

xuweibj:
Hi @whowutwut ,
1. To set BMC to DHCP:
```
curl -b cjar -k -H "Content-Type: application/json" -X POST -d '{"data": ["eth0"]}' https://172.20.140.102/org/openbmc/NetworkManager/Interface/action/EnableDHCP
```
It will not get any response and hung there. 
Now ``mid05tor12cn02`` bmc ip is ``172.20.253.105``.

2. Check Address Type:
```
curl -k -b cjar -H "Content-Type: application/json" -X POST -d '{"data": ["eth0"]}' https://172.20.140.102/org/openbmc/NetworkManager/Interface/action/GetAddressType
{
  "data": "DHCP",
  "message": "200 OK",
  "status": "ok"
}
```

whowutwut:
Nice!! Thanks!  That's unfortunate that it hangs. The ipmitool command returned so we could get back out .. :-(. Can we put that curl cmd into ip=dhcp command or does xcat not handle that hang well?

xuweibj:
[...]

PR #3716, opened 2017-08-18
rspconfig ip to dhcp for openbmc

xuweibj:
#3295
Output as below:
```
# rspconfig frame6u05 ip=dhcp
frame6u05: DHCP
frame6u05: Will reset BMC...
frame6u05 BMC: reboot
```

When xcatdebugmode is 1:
```
# rspconfig frame6u05 ip=dhcp
frame6u05: [openbmc_debug] POST https: //10.6.5.253/login -d { "data": [ "root", "0penBmc" ] }
frame6u05: [openbmc_debug] login_response 200 OK
frame6u05: [openbmc_debug] POST https: //10.6.5.253/xyz/openbmc_project/network/action/Reset -d {"data":[]}
frame6u05: [openbmc_debug] rspconfig_dhcp_response 200 OK
frame6u05: DHCP
frame6u05: Will reset BMC...
frame6u05: [openbmc_debug] PUT https: //10.6.5.253/xyz/openbmc_project/state/bmc0/attr/RequestedBMCTransition -d {"data":"xyz.openbmc_project.State.BMC.Transition.Reboot"}
frame6u05: [openbmc_debug] rpower_reset_response 200 OK
frame6u05 BMC: reboot
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -561,6 +561,7 @@ sub parse_args {
         #
         # disable function until fully tested
         #
+        $check = unsupported($callback); if (ref($check) eq "ARRAY") { return $check; }
         my $setorget;
         foreach $subcommand (@ARGV) {
             if ($subcommand =~ /^(\w+)=(.*)/) {
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -1669,8 +1669,8 @@ sub rspconfig_response {
 
     if ($node_info{$node}{cur_status} eq "RSPCONFIG_DHCP_RESPONSE") {
         if ($response_info->{'message'} eq $::RESPONSE_OK) {
-            xCAT::SvrUtils::sendmsg("DHCP", $callback, $node);
-            xCAT::SvrUtils::sendmsg("Will reset BMC...", $callback, $node);
+            my $bmc_node = "$node BMC";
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1649,6 +1667,13 @@ sub rspconfig_response {
         xCAT::SvrUtils::sendmsg("$_", $callback, $node) foreach (@output);
     }
 
+    if ($node_info{$node}{cur_status} eq "RSPCONFIG_DHCP_RESPONSE") {
+        if ($response_info->{'message'} eq $::RESPONSE_OK) {
+            xCAT::SvrUtils::sendmsg("DHCP", $callback, $node);
```
whowutwut: I think for the externals, IPMI does not print any messages but I think we should print something like you did... but suggest the following external message: 

```
# rspconfig frame6u05 ip=dhcp
frame6u05 BMC: Setting IP to DHCP...
frame6u05 BMC: reboot

File: xCAT-server/lib/xcat/plugins/openbmc.pm [hunk 2]
```diff
@@ -553,7 +561,6 @@ sub parse_args {
         #
```
whowutwut: @xuweibj  If we remove 556 below, should also remove these comments. 

@gurevichmark  If we expose this command, do we have the proper blocks for the other attributes that are not supported?

xuweibj: @whowutwut sorry I forgot to put it back after test. I have put it back now.

Issue #3698, opened 2017-08-16
[DEV] Callback message may hang at xcatd side


chenglch:
We encountered this problem while running bmc discovery for a large cluster (more than 2 thousand nodes). The child processes may hang at the end if  `xCAT::MsgUtils->message("I" , ...` is called. This is a general error of xcatd while redirecting the message from pipe to the client.

chenglch:
After running some self scripts, the root cause is from the code logic of xcatd while handling communication between the socket pairs if multiple child processes exist.

Below is the code block that has race condition if child processes send the message back to the client at the same time. `relay_fds` is running in `immidiate` process and `convey_response` is running in child plugin processes.
```
sub relay_fds { # Relays file descriptors from pipes to children to the SSL socket
    my $fds        = shift;
    my $replyqueue = shift;
    my $goneclient = 0;
    my $collate    = (scalar @_ > 0 ? shift : 0);
    my @readyset   = $fds->can_read(1);
    my $rfh;
    my $rc = @readyset;
    my $text;

    # A PIPE signal might be received when run fd_retrieve from the plugin sub processors
    # This mostly happens when there are multiple plugins are called for certain command
    # So spkit the pipe error handle
    $pipeexpected = 1;
    foreach $rfh (@readyset) { # go through each child, extract a complete, atomic message
        my $line;
        my $resp;
        eval {
            $resp = fd_retrieve($rfh);
        };
        if ($@ and $@ =~ /^Magic number checking on storable file/) { # this most likely means we ran over the end of available input
            $fds->remove($rfh);
            close($rfh);
        } else {
            push @$replyqueue, $resp;
            print $rfh "nfin\n";
        }
    }
   ....
```

```
sub convey_response {
    my $resp = shift;

    # TODO: This is where the following will/may happen:
    # -Track transaction id
    # -Save output for deferred commands
    unless ($parent_fd) {
        build_response($resp);
        return;
    }
    unless ($resp) { return; }
    $pipeexpected = 1;

    #$resp = XMLout($resp,KeyAttr=>[], NoAttr=>1,RootName=>'xcatresponse');
    # sanitize the response, to avoid being killed by non-printable bytes
    #$resp =~ tr/\011-\177/?/c;
    # seeing if using utf-8 offloads potential issues to client terminal, it didn't
    store_fd($resp, $parent_fd);
    yield; # parent must get timeslice anyway before an ack could possibly return
    my $parsel = new IO::Select;
[...]

PR #3719, opened 2017-08-18
Fix the bug due to the race condition about the pipe usage in xcatd

chenglch:
This commit delete the code logic that may cause the race condiction
if multiple child processes are waiting for the response.

fix-issue: #3698

### DIFF
--- a/xCAT-server/sbin/xcatd
+++ b/xCAT-server/sbin/xcatd
@@ -2449,35 +2449,12 @@ sub convey_response {
     # sanitize the response, to avoid being killed by non-printable bytes
     #$resp =~ tr/\011-\177/?/c;
     # seeing if using utf-8 offloads potential issues to client terminal, it didn't
-    store_fd($resp, $parent_fd);
-    yield; # parent must get timeslice anyway before an ack could possibly return
-    my $parsel = new IO::Select;
-    $parsel->add($parent_fd);
-    my $selbits = $parsel->bits;
-    my $rsp;
-    while ($selbits && ($rsp = select($selbits, undef, undef, 5))) { # block up to five seconds
-
-        if ($quit) {    # Obey quit flag
-            xexit 0;
-        }
-        if ($rsp == 0) {    # This means the filedescriptor was removed
-            last;
-        }
-        if ($rsp < 0) {     # A signal caused select to skip out, do-over
-            next;
-        }
-
-        # At this point, the only possibility is a positive return, meaning parent_fd requires attention of some sort
-        $rsp = <$parent_fd>;
-        if ($rsp) { # If data actually came in, last, otherwise, remove it from the IO::Select, but both should amount to the same thing
-            last;
-        } else {
-            $parsel->remove($parent_fd);
-            last;
-        }
+    eval {
+        store_fd($resp, $parent_fd);
+    };
+    if ($@) {
+        xCAT::MsgUtils->message("S", "Failed to store into socketpair: $@");
     }
-    yield; # If still around, it means a peer process still hasn't gotten to us, so might as well yield
-    $selbits = $parsel->bits;
 }
 
 sub build_response {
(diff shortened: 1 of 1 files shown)


daniceexi:
@chenglch Looks like this part is code is used to send an ACK to the message sender, it may used for flow control. But the problem is there's no retry.

chenglch:
@daniceexi, I have described the race-condition phenomenon cased by the multi-child processes environment in #3698. I don't know how could these code implement the flow-control, in my opinion, if there is one child process, it is not necessary to import flow-control, if there are multi-child processes, the flow control could be implemented by the kernel as all of the IO operations are in block mode.

daniceexi:
For the flow control, the current code logic is to make sure that each time that a child only can send one message. If no this ACK, then the child process could send as much as possible until the buffer is full. I am not sure whether it's a problem. But I think this kind of code might have goodness in the case that there are tons of messages come from child process.

chenglch:
@daniceexi , I understand your point, but I think `store_fd` which send the message to the parent process will be block if  the pipe is full.

This flow control also could not work well if multiple child processes send a large amount of message to the pipe at the  same time as the operation may not be atomic within kernel.

I hope `store_fd` could do something to provide the atomic feature, but I don't know how it implement inside.

robin2008:
Here is some comments after the review meeting:
 - The root cause is that some plugins using xfork to handle business in parallel, but not using plugin internal socketpair to handle its inter-process messaging first before giving the response to xCATd.
When there are multiple child process send the response (using global callback in xCAT::MsgUtils->message) to xcat in the same time,  xCATd give back the ACK and only 1 child process could be released.

Now the fix is to disable the ACK mechanism, it could workaround above issue. But we does not know if it will bring some issue when handling lots of response, need a sanity testing to check if the response is still heath when bursting message coming in same time. 

Also, we may need to check with Jarrod to see if anything important missing.

xCAT-server/sbin/xcatd
```diff
@@ -2449,35 +2449,12 @@ sub convey_response {
...
-    store_fd($resp, $parent_fd);
-    yield; # parent must get timeslice anyway before an ack could possibly return
-    my $parsel = new IO::Select;
-    $parsel->add($parent_fd);
-    my $selbits = $parsel->bits;
-    my $rsp;
-    while ($selbits && ($rsp = select($selbits, undef, undef, 5))) { # block up to five seconds
-
-        if ($quit) {    # Obey quit flag
-            xexit 0;
```
jjohnson42: I'm wondering if removing this has side effect of failure to abort the child by the parent.

jjohnson42:
So this is a bit beyond my recollection...  Nearly 10 years ago some of that code was fleshed out...
commit 5075adf2694da37b73787e27e40bc30ca859c89a
Author: jbjohnso <jbjohnso@8638fb3e-16cb-4fca-ae20-7b5d299a9bcd>
Date:   Fri May 16 18:12:46 2008 +0000

    -Fix problem where xCATd would terminate prematurely because a grandchild exits at an inopportune time

In terms of the original wait for ack behavior, it *might* have been preparing for a long ditched goal of having stored responses, it *might* have been something about the state of Linux distros 10 years ago, it *might* have been me not knowing or being confident in behavior of the pipe, this behavior has been there since before it was even on sourceforge, so can't see the change history far enough back.

chenglch:
Thanks @jjohnson42 , from my self test , the `$quit` behavior is not affected.

I tested this fix with my own script which fork about 3000 processes and send response back to the xcatd. If the size of message is not larger than 8000 bytes, this changes works well as the `store_fd` and `fd_retrieve` always call the `write` or `read` system calls with the `8192` buffer size.   If the message size is larger than 8192, the processes may hang at `store_fd` or `fd_retrieve` stage as they call the `write` or `read` in a loop.

Below is the test conclusion:
- This fix do not bring extra problems in a concurrent environment. 
- It is not advocated to reuse the `unix socket with stream type` to send the callback message from the forked child processes concurrently as the `write` and `read` on socket can not be treat as atomic operation within the kernel.

zet809:
Hi, @chenglch , it is possible to add a warning message that the message size larger than 8000 will cause potential issues?

chenglch:
@zet809, we'd better not to do that for some reasons:
- $resp is a pointer, it's not easy to calculate the size for this structure.
- the subroutine itself do not aware of if it is in concurrent mode, if only one process to send the message, we should allow this, it is the correct way.
- it is not recommended to send message in concurrent mode.

Issue #3718, opened 2017-08-18
[FVT]:refine case mkdef_regex_kvm to reduce the false failure


tingtli:
In case mkdef_regex_kvm, there is some error  option for command chdef
```
start:mkdef_regex_kvm
description:The fourth example in xCAT document Using Regular Expressions in the xCAT Tables
cmd:chtab node=kvms 'vm.host=|\D+(\d+)\D+(\d+)\D+(\d+)\D+(\d+)|c($1)f($2)x($3)|' 'vm.storage=|\D+(\d+)\D+(\d+)\D+(\d+)\D+(\d+)|dir:///install/vms/vm($4+0)|' vm.memory=3072 vm.cpus=2 vm.nics=virbr2 vm.nicmodel=virtio
check:rc==0
cmd:mkdef -t node -o c01f02x03v04 groups=kvms mgt=kvm
check:rc==0
cmd:chdef c01f02x03v04 -i vmcpus | grep 'vmcpus=2$'
check:rc==0
cmd:chdef c01f02x03v04 -i vmhost | grep 'vmhost=c01f02x03$'
check:rc==0
cmd:chdef c01f02x03v04 -i vmmemory | grep 'vmmemory=3072$'
check:rc==0
cmd:chdef c01f02x03v04 -i vmnicnicmodel | grep 'vmnicnicmodel=virtio$'
check:rc==0
cmd:chdef c01f02x03v04 -i vmnics | grep 'vmnics=virbr2$'
check:rc==0
cmd:chdef c01f02x03v04 -i vmstorage | grep 'vmstorage=dir:///install/vms/vm4$'
check:rc==0
cmd:rmdef -t node c01f02x03v04
check:rc==0
cmd:chtab -d node=kvms vm
check:rc==0
end
```
-i should only be used by lsdef

tingtli:
Fixed in #3720

PR #3720, opened 2017-08-18
fix bug 3718, refine case mkdef_regex_kvm to reduce the false failure

tingtli:
fix bug #3718, refine case mkdef_regex_kvm to reduce the false failure


hu-weihua:
It is good for me, agree to merge. thanks @tingtli for finding such mistake.

PR #3731, opened 2017-08-21
Install the cumulus license from cumulusztp

mattaezell:
If running ZTP from a fresh install, the license must be installed before _configinterface_ can be run.


immarvin:
i am ok with the change
thanks @mattaezell

PR #3732, opened 2017-08-21
Allow the cumulus discovery listener to accept multiple messages

mattaezell:
By default, _socat_ exits after it has received a message.  If that message is "_processing_", the switch won't realize that it has been discovered.  This change allows socat to receive multiple messages from the xCAT master so that it can receive the "_restart_" directive that indicates that discovery completed.

### DIFF
--- a/xCAT/postscripts/documulusdiscovery
+++ b/xCAT/postscripts/documulusdiscovery
@@ -17,7 +17,8 @@ EOF
 
 #listen to the 3001 port for any response from xcatd on MN
 #the message in the response will be written to /tmp/result.socat.out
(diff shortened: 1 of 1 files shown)


mattaezell:
Note: cumulusztp discovery doesn't work correctly in xCAT 2.13.6 due to this bug

immarvin:
I am ok with the change

agree to merge first, need to refine the documulusdiscovery according to the PR https://github.com/xcat2/xcat-core/pull/3601/files#diff-4033787c6534fab1cd85faa4b7e30b4f for the newly introduced notification from xcatd to prevent duplicate findme requests while the client's findme request is processing

PR #3739, opened 2017-08-22
refine cases to reduce false error

tingtli:
refine cases to reduce false error


hu-weihua:
It is good for me, thinks @tingtli

PR #3742, opened 2017-08-22
Send MTM in documulusdiscovery to support MTMS discovery

### LINKED ISSUE #3730: Cumulus discovery does not work with MTMS
Eventually we want to use switch-based switch discovery, but for early verification I tried to use MTMS-based discovery.  The _documulusdiscovery_ script does not send a MTM, so MTMS-based discovery cannot work.  It's sending the MTM in the _vendor_ attribute:
https://github.com/xcat2/xcat-core/blob/2679d56b730d0f70ef73225b714580bc664d8277/xCAT/postscripts/documulusdiscovery#L56

In my original [draft](https://gist.github.com/mattaezell/f385baa5bdccf6f63145489f3c172da2), MTM was sent as mtm, not vendor.  Why was this changed @immarvin?

mattaezell:
Fixes #3730


immarvin:
agree to merge, thanks @mattaezell

PR #3533, opened 2017-07-25
expose reventlog and modify time format

xuweibj:
#3497 
1. Expose reventlog in openbmc.pm

2. Change ``Timestamp`` format to UTC time.
```
# XCATBYPASS=1 reventlog openbmc156 5 -s
[OpenBMC development support] Using this version of xCAT, ensure firware level is at v1.99.6-0-r1, or higher.
openbmc156: Entry 67: 07/26/2017 19:13:59 org.open_power.Host.Error.WatchdogTimedOut
openbmc156: Entry 66: 07/26/2017 19:13:48 org.open_power.Host.Error.Checkstop
openbmc156: Entry 65: 07/26/2017 19:12:49 org.open_power.Host.Error.Checkstop
openbmc156: Entry 64: 07/26/2017 19:11:51 org.open_power.Host.Error.Checkstop
openbmc156: Entry 63: 07/26/2017 19:10:54 org.open_power.Host.Error.Checkstop
```

* [ ]  For option ``clear``, it is to delete all logs. There is no interface now, it will be done in phase 5 depending on comments in https://github.com/openbmc/openbmc/issues/2001.

PR #3744, opened 2017-08-23
modify reventlog testcase for code change

junxiawang:
I modify revnetlog testcase for code chage:
https://github.com/xcat2/xcat-core/pull/3533 
So the output for testcase's check need to modify


xuweibj:

PR #3746, opened 2017-08-23
Fix test case xdcp_nonroot_user

neo954:
It seems xCAT-test cannot handle command line use sub-shell. This is a fix.


hu-weihua:

PR #3750, opened 2017-08-23
OpenBMC PNOR upload and activate firmware image file

gurevichmark:
Implements `rflash` upload and activate image file in one step for issue #3173 

Output:
```
[root@stratton01 xcat]# rflash p9euh02 -a /tmp/gurevich/witherspoon.20170713s.pnor.squashfs.tar
p9euh02: Uploading /tmp/gurevich/witherspoon.20170713s.pnor.squashfs.tar ...
p9euh02: Successful, use -l option to list.
p9euh02: rflash started, please wait...
p9euh02: Activating firmware update. 10%
p9euh02: Activating firmware update. 60%
p9euh02: Activating firmware update. 60%
p9euh02: Firmware update successfully activated
[root@stratton01 xcat]#
```

```
[root@stratton01 xcat]# rflash p9euh02 -l
p9euh02: ID       Purpose State      Version
p9euh02: -------------------------------------------------------
p9euh02: 2a1022fe Host    Active(*)  IBM-witherspoon-sequoia-ibm-OP9_v1.17_1.67
p9euh02: a66e68f6 BMC     Active     v1.99.9-30-gaaafab8
p9euh02: 39885ab3 Host    Active     IBM-witherspoon-redbud-ibm-OP9_v1.18_1.37
p9euh02:
[root@stratton01 xcat]#
```

### DIFF
--- a/docs/source/guides/admin-guides/references/man1/rflash.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/rflash.1.rst
@@ -59,7 +59,7 @@ OpenPOWER OpenBMC specific :
 
 \ **rflash**\  \ *noderange*\  {[\ **-c | -**\ **-check**\ ] | [\ **-l | -**\ **-list**\ ]}
 
-\ **rflash**\  \ *noderange*\  \ *tar_file_path*\  {[\ **-c | -**\ **-check**\ ] | [\ **-u | -**\ **-upload**\ ]}
+\ **rflash**\  \ *noderange*\  \ *tar_file_path*\  {[\ **-c | -**\ **-check**\ ] | [\ **-a | -**\ **-activate**\ ] | [\ **-u | -**\ **-upload**\ ]}
 
 \ **rflash**\  \ *noderange*\  \ *image_id*\  {[\ **-a | -**\ **-activate**\ ] | [\ **-d | -**\ **-delete**\ ]}
 
@@ -208,7 +208,7 @@ The command will update firmware for OpenPOWER OpenBMC when given an OpenPOWER n
 
 \ **-a|-**\ **-activate**\ 
  
- Activate update image. Image id must be specified.
+ Activate update image. Image id or update file must be specified.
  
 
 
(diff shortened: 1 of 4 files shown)


docs/source/guides/admin-guides/references/man1/rflash.1.rst
```diff
@@ -322,7 +322,17 @@ The command will update firmware for OpenPOWER OpenBMC when given an OpenPOWER n
  
  .. code-block:: perl
  
-   rflash briggs01 -d=/root/supermicro/OP825
+   rflash briggs01 -d /root/supermicro/OP825
```
whowutwut: :+1:

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -845,6 +853,7 @@ sub parse_command_status {
         my $upload = 0;
         my $activate = 0;
         my $update_file;
+        my $upload_and_activate = 0;
```
whowutwut: I am trying to see where `upload` variable is being used?  It seems like it's not...... 

What happens if update_file is provided, then we set upload=1, then we don't need another variable...

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -940,6 +963,13 @@ sub parse_command_status {
             $next_status{"RFLASH_SET_PRIORITY_REQUEST"} = "RFLASH_SET_PRIORITY_RESPONSE";
             $next_status{"RFLASH_SET_PRIORITY_RESPONSE"} = "RFLASH_UPDATE_CHECK_STATE_REQUEST";
         }
+        if ($upload_and_activate) {
```
whowutwut: Instead of creating a new variable , couldn't we just test for `if ($upload && $activate)` and go into this clause?

xCAT-server/lib/xcat/plugins/openbmc.pm
gurevichmark: The `upload` variable gets set on line 866 and then checked on line 948.

xCAT-server/lib/xcat/plugins/openbmc.pm
gurevichmark: Yes, if we use `if ($upload && $activate)` we will execute the block on line 966, but we would also execute the block on line 948 (because $upload is set) and the block on line 953 (because $activate is set). 

This can be fixed by changing those lines to be `if ($upload && !$activate)` and `if ($activate && !$upload)` respectively. But I think it would make the code a little harder to follow, is it worth it to save one variable ?

whowutwut: Ok, fair enough.  This is fine.

zet809:
Hi, @gurevichmark , thx for the work, it will make it easier to use rflash.

PR #3754, opened 2017-08-24
Remove test cases about vlan, bond, bridge from unsupported platform

hu-weihua:
Remove test cases about vlan, bond, bridge from unsupported platform.
There is not expected physical configuration in PPC64 platform.
xCAT has not supported vlan, bond, bridge in sles and ubuntu.


bybai:
I am OK to merge it.

PR #3775, opened 2017-08-25
[xCAT-genesis-builder]Fix `sed` leaving `dracut_install` without parameters

samveen:
The `sed` transformation to `install` in the `buildrpm` script leaves an unparamterized `dracut_install` for the pattern `\/usr\/sbin\/iprconfig` as it is the only option to `dracut_install`. The line should deleted instead of applying a substitution to the parameter.


zet809:
Hi, @samveen , looks good to me, thx a lot!

PR #3777, opened 2017-08-25
Cherry-pick the Fix 3749 from master, handle '+' and '-' in id when merging /etc/passwd 

### LINKED ISSUE #3749: [PCM] The merge function of updatenode cannot work with special charcharacters like '+'
when special characters are used in order to enable NIS authentication in /etc/passwd file as show below:
+@sysadmins::::::
+@p8users::::::
+:*:::::/sbin/nologin

After running updatenode computenode -F, will sync the passwd to CN from passwd.merge file. However, if we run the command  again, it creates duplicate entries of the above line plus few 2-3 users to CN.  Run the command third time, it will triplicate those line in passwd in CN and so on.

robin2008:
* Fix #3749, handle '+' and '-' in id when merging /etc/passwd

* Add some comments

### DIFF
--- a/xCAT-server/share/xcat/scripts/xdcpmerge.sh
+++ b/xCAT-server/share/xcat/scripts/xdcpmerge.sh
@@ -76,7 +76,9 @@ for i in $*; do
     delim="|"
     for u in $removeusers
     do
-      userlist=$userlist$u$delim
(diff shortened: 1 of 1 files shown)


robin2008:
The branch is old, new CI is not suitable.

PR #3780, opened 2017-08-25
Comment KeyFile instead of replacing as key is included in CertFile

samveen:
Quoting `ssl.conf`:
```
#   Server Private Key:
#   If the key is not combined with the certificate, use this
#   directive to point at the key file.  Keep in mind that if
#   you've both a RSA and a DSA private key you can configure
#   both in parallel (to also allow the use of DSA ciphers, etc.)
```
As the xCAT host SSLCertificateFile already contains the key, the substitution for `SSLCertificateKeyFile` can be changed to simpler one: which just comments it out.


zet809:
HI, @chenglch , will you pls take a look at this PR? Thx!

PR #3784, opened 2017-08-28
Add debug information for isntallation failed case in automation

hu-weihua:
In sles11.4+ppc64 project in automation, sn provision test case always failed recently.
After debug, I found MYSQL was installed twice, but I could not found where triggered the second time MYSQL installation. So add some information for further debug.


xuweibj:
Looks good to me. Agree to merge.

PR #3786, opened 2017-08-28
Change the order of confignetworks test case in rhels7.4_x86_64.bundle

hu-weihua:
Due to test case ``nodeset_runimg`` boots CN into genesis, although this case will reboot CN from disk before it return, but CN needs a while to finish reboot.  When the ``confignetworks`` test case run, the CN is still rebooting, so the ``confignetworks`` test case failed to connect CN. 

So move all ``confignetworks`` test case ahead.


bybai:
I am OK to merge it.

Issue #3589, opened 2017-08-01
xcat-dep repos are not in sync after devel repo created 


whowutwut:
@neo954 manually created the yum repos on xcat.org, and I'm thinking we probably do not want to continue to use this because the complete picture is not implemented. I think we will be worse off when the devel repos are not kept in sync with the latest.  

There's too many pieces that must be completed for this to work
* https://github.com/xcat2/xcat-core/blob/master/builddep.sh
* https://github.com/xcat2/xcat.org/blob/master/promote_build.py
* GSA location where we are storing the files 

I think the flow should be.... 
1. Put the xcat-dep packages in GSA 
2. Running build-dep should ALWAYS upload the file to http://xcat.org/files/xcat/xcat-dep/2.x_Linux/ or http://xcat.org/files/xcat/xcat-dep/2.x_Ubuntu/ and the rsync to xcat.org yum repo should always go here: http://xcat.org/files/xcat/repos/yum/devel/xcat-dep 
3. When we are ready to promote a release, the promote script should move the yum repo from the devel area to the public: http://xcat.org/files/xcat/repos/yum/xcat-dep/ 

Without this in place, after @immarvin built a new xcat-dep and published it, the devel repo is out of date.  

Here's what happened: 
1. Installed 8/1 build of xCAT using `go-xcat` 
2. Realized the xCAT-genesis-base ppc64 is NOT updated, but it IS there in the non devel repo 
```
[root@briggs01 yum.repos.d]# cat xcat-dep.repo
[xcat-dep]
name=xcat-dep
baseurl=http://xcat.org/files/xcat/repos/yum/devel/xcat-dep/rh7/ppc64le
enabled=1
gpgcheck=0
```
Looking at what files are provided by the repos....
```
[root@briggs01 yum.repos.d]# yum list   |grep xCAT-genesis
xCAT-genesis-base-ppc64.noarch          2:2.13.2-snap201702150049  @xcat-dep <------
xCAT-genesis-base-x86_64.noarch         2:2.13.2-snap201702200120  @xcat-dep
xCAT-genesis-scripts-ppc64.noarch       1:2.13.6-snap201708010615  @xcat-core
xCAT-genesis-scripts-x86_64.noarch      1:2.13.6-snap201708010615  @xcat-core
xCAT-genesis-x86_64.noarch              1:2.7.7-snap201301100842   xcat-dep
```

This is dated from Feb.... 

[...]

PR #3791, opened 2017-08-28
Place xcat-dep to the devel directory of xcat.org

neo954:
rsync xcat-dep to `http://xcat.org/files/xcat/repos/yum/devel/xcat-dep/`


neo954:
This pull request partly solve xcat2/xcat-core#3589

PR #3795, opened 2017-08-29
Add missing dependency package for pre_deploy_sn used by SN setup test case

hu-weihua:
In last day fix #3784, missed dependency package. Add it in this pull request.


xuweibj:
Agree to merge.

PR #3789, opened 2017-08-28
fix issue [FVT]:nodeset against diskless osimage output is not consistent on rhels and sles/ubuntu #3783

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3783:
 
refine the message of mknetwork

PR #3798, opened 2017-08-29
Modify the matching condition of test case rmimage_diskless

hu-weihua:
Due to the output of ``nodeset`` are different between SLES, RHEL and UBUNTU. 
It is lower priority for developer to unified the output, so change the matching condition of test case ``rmimage_diskless`` first.


hu-weihua:
It seems developer has unified the output between different OS in pr #3789. So close this change first.

PR #3799, opened 2017-08-29
refine passwd test cases

tingtli:
refine passwd test cases


hu-weihua:
It is ok for me, agree to merge. thanks

Issue #3783, opened 2017-08-28
[FVT]:"nodeset" against diskless osimage output is not consistent on rhels and sles/ubuntu


tingtli:
xCAT 2.13.6
When image is delete by rmimage, then, user run nodeset, the error on sles and ubuntu is as follows
```
root@c910f04x35v02:/opt/xcat/share/xcat/tools/autotest/result# nodeset c910f04x35v04 osimage=ubuntu16.04.1-x86_64-netboot-compute
Error: Did you run "genimage" before running "packimage"? kernel cannot be found
```
But on rhels, the error message is as follows
```
#nodeset c910f02c08p07 osimage=rhels7.4-ppc64-netboot-compute [Sun Aug 27 09:52:02 2017]
ElapsedTime:2 sec
Error: Did you run "genimage" before running "packimage"? kernel cannot be found at /install/netboot/rhels7.4/ppc64/compute/kernel on c910f02c08p05
```

immarvin:
HI @bybai ，would you please take a look at this?

thanks

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3789

immarvin:
verified by autotest

PR #3805, opened 2017-08-29
fix issue [FVT]:nodeset against diskless osimage output is not consistent on rhels and sles/ubuntu #3783

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3783:

refine the message on nodeset against diskless/statelite osimages when kernel cannot be found


bybai:
I saw your commit updates for debian and sles, where are changes for RH?

immarvin:
the message for ubuntu and sles should be kept consistent with redhat

bybai:
I am OK to merge it.

Issue #3792, opened 2017-08-28
Discovery does not start running chain command because mac address NOIP....


whowutwut:
I've been seeing lots of people get hung up on this issue, so much so that I immediately know the answer...   which is not good....

 @ralphbellofatto brought this up to my attention today and it's difficult for users to understand the problem here...  

So they are re-discovering a node and it sits in genesis looping, can't continue... 
```
[Mon Aug 28 14:28:28 2017]<166>Aug 28 14:28:27 xcat.genesis.doxcat: The destiny=discover, destiny parameters=discover
```

The chain attribute is: `chain=runcmd=bmcsetup,standby` but it never gets to running bmcsetup. 

The mac is: `mac=70:e2:84:14:29:2b!*NOIP*`

So the reason for this is makehosts/makedhcp is NOT run. 

@zet809 I thought you said it would automatically run in discovery?? 

* So how can we better confirm that makedhcp is automatically run?
* What is the missing step that causes us to get hung up on this so much? Reproduce this... 
* Is there a way to improve the logging? 
* Does xCAT probe find it? Can it be used during discovery to find problems to resolve?

zet809:
1. We can use `makedhcp -q <node>` to check the DHCP configuration for a node.
2. Have him run 'makedns <pre-defined-node>' before starting hardware discovery?
3. Yes, we can enhance the log to report the error configuration. Updated in PR #3806
4. Yes, this error can be checked by `xcatprobe discovery -c` to do pre-check before start hardware discovery.

zet809:
Hi, @whowutwut , with PR #3806, I think we can close this issue, what is your idea?

whowutwut:
Sure, this one is more difficult for me to re-produce right now, so let's close and see if anything comes up during testing.

PR #3806, opened 2017-08-29
add log for hardware discovery when no ip address can be resolved for…

zet809:
… the pre-defined node

The following log will be added in /var/log/xcat/cluster.log with this fix:
```
Aug 31 05:59:14 briggs01 xcat[22442]: xcat.discovery.nodediscover: Can not resolve IP for the matching node:<pre-defined-node>, please makesure "makehost <pre-defined-node>" and "makedns <pre-defined-node>" have run before starting hardware discovery.
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/nodediscover.pm
+++ b/xCAT-server/lib/xcat/plugins/nodediscover.pm
@@ -313,7 +313,7 @@ sub process_request {
                             push @hostnames_to_update, $hosttag;
                         }
                         elsif (!inet_aton($node)) {
-                            xCAT::MsgUtils->message("S", "xcat.discovery.nodediscover: Can not resolve IP for the matching node:$node, please makesure \"makehost $node\" and \"makedns $node\" have run before starting hardware discovery.");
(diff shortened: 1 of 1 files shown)


whowutwut:
@zet809 Can you provide the UT output and what the external messages look like in which log will it appear.   I want to make sure that it makes sense from a debug point of view...  would it be in the cluster.log?

xCAT-server/lib/xcat/plugins/nodediscover.pm
```diff
@@ -312,6 +312,9 @@ sub process_request {
                             $hosttag = "$node-$ifinfo[1]";
                             push @hostnames_to_update, $hosttag;
                         }
+                        elsif (!inet_aton($node)) {
+                            xCAT::MsgUtils->message("S", "xcat.discovery.nodediscover: Can not resolve IP for the matching node:$node, please makesure \"makehost $node\" and \"makedns $node\" have run before starting hardware discovery.");
```
whowutwut: some spelling errors and `makehost` needs an s..   Suggest changing to this message:

`...  matching node:$node.  Make sure \"makehosts\" and \"makedns\" have been run for $node.");`

zet809:
Hi, @whowutwut and @gurevichmark , modified, pls help to review. Thx!

whowutwut:
Resolves #3792

PR #3807, opened 2017-08-29
update for confignetwork issue

tingtli:
update for confignetwork issue


hu-weihua:
It is ok for me, agree to merge. thanks @tingtli

Issue #3796, opened 2017-08-29
[OpenBMC] Add trace log to record request and response


xuweibj:
If ``xcatdebugmode`` is 1/2, will print debug info on the screen and add into /var/log/xcat/cluster.log.
Output:
```
# rpower openbmctest stat
openbmctest: [openbmc_debug] curl -k -c cjar -H "Content-Type: application/json" -d '{ "data": ["root", "xxxxxx"] }' https://9.3.185.29/login
openbmctest: [openbmc_debug] login_response 200 OK
openbmctest: [openbmc_debug] curl -k -b cjar -X GET -H "Content-Type: application/json" https://9.3.185.29/xyz/openbmc_project/state/enumerate
openbmctest: [openbmc_debug] rpower_status_response 200 OK
openbmctest: on
```

Log:
```
Aug 29 23:04:59 c910f03c05k04 xcat[3507]: DEBUG [openbmc_debug] openbmctest curl -k -c cjar -H "Content-Type: application/json" -d '{ "data": ["root", "xxxxxx"] }' https://9.3.185.29/login
Aug 29 23:04:59 c910f03c05k04 xcat[3507]: DEBUG [openbmc_debug] openbmctest login_response 200 OK
Aug 29 23:04:59 c910f03c05k04 xcat[3507]: DEBUG [openbmc_debug] openbmctest curl -k -b cjar -X GET -H "Content-Type: application/json" https://9.3.185.29/xyz/openbmc_project/state/enumerate
Aug 29 23:05:00 c910f03c05k04 xcat[3507]: DEBUG [openbmc_debug] openbmctest rpower_status_response 200 OK
```

whowutwut:
@zet809 @xuweibj Before you spend too much time on this. What does the response look like when it's printed into the logs. The best output is really what is printed by Dumper() but that doesn't directly go into the logs (esp for hierarchy...)  So I don't want to waste time trying to format JSON data into xCAT logs...  

My suggestion, and I also mentioned this on the call...  is to just generate the curl command for the user to execute when xcatdebugmode=1.  

This would save us a lot of time when we need to ask the user to capture more data.  Instead of trying to piece together the curl command, they can just copy/paste it on the command line.   So our response would be: 
1.  Set `xcatdebugode=1` , give us that output 
2.  Run the curl command that's printed when debugmode is on and send us that output. 

then we also are not printing large amounts of data, it would affect performance.

xuweibj:
[...]

PR #3813, opened 2017-08-30
add trace log in openbmc

xuweibj:
#3796 
xcatdebugmode=1
```
# rpower openbmctest stat
openbmctest: [openbmc_debug] curl -k -c cjar -H "Content-Type: application/json" -d '{ "data": ["root", "xxxxxx"] }' https://9.3.185.29/login
openbmctest: [openbmc_debug] login_response 200 OK
openbmctest: [openbmc_debug] curl -k -b cjar -X GET -H "Content-Type: application/json" https://9.3.185.29/xyz/openbmc_project/state/enumerate
openbmctest: [openbmc_debug] rpower_status_response 200 OK
openbmctest: off
```
in /var/log/xcat/cluster.log
```
Aug 30 03:02:56 c910f03c05k04 xcat[4289]: DEBUG [openbmc_debug] openbmctest curl -k -c cjar -H "Content-Type: application/json" -d '{ "data": ["root", "xxxxxx"] }' https://9.3.185.29/login
Aug 30 03:02:56 c910f03c05k04 xcat[4289]: DEBUG [openbmc_debug] openbmctest login_response 200 OK
Aug 30 03:02:56 c910f03c05k04 xcat[4289]: DEBUG [openbmc_debug] openbmctest curl -k -b cjar -X GET -H "Content-Type: application/json" https://9.3.185.29/xyz/openbmc_project/state/enumerate
Aug 30 03:02:57 c910f03c05k04 xcat[4289]: DEBUG [openbmc_debug] openbmctest rpower_status_response 200 OK
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -949,7 +949,6 @@ sub parse_command_status {
         }
     }
 
-    print Dumper(\%next_status) . "\n";
     return;
 }
 
@@ -1239,9 +1238,6 @@ sub rpower_response {
     xCAT_monitoring::monitorctrl::setNodeStatusAttributes(\%new_status, 1) if (%new_status);
 
     if ($node_info{$node}{cur_status} eq "RPOWER_STATUS_RESPONSE" and !$next_status{ $node_info{$node}{cur_status} }) { 
-
-        print Dumper(\%{ $response_info->{data} }) . "\n";
-
         my $bmc_state = "";
         my $bmc_transition_state = "";
         my $chassis_state = "";
@@ -1333,8 +1329,6 @@ sub rinv_response {
     my $response = shift;
 
     my $response_info = decode_json $response->content;
-
-    print Dumper(\%{ $response_info->{data} }) . "\n";
 
     my $grep_string;
     if ($node_info{$node}{cur_status} eq "RINV_FIRM_RESPONSE") {
@@ -1570,9 +1564,6 @@ sub reventlog_response {
             xCAT::SvrUtils::sendmsg("clear", $callback, $node);
         }
     } else {
-
-        print Dumper(\%{ $response_info->{data} }) . "\n";
-
         my ($entry_string, $option_s) = split(",", $status_info{REVENTLOG_RESPONSE}{argv});
         my $content_info; 
         my %output = ();
@@ -1790,12 +1781,6 @@ sub rvitals_response {
     my $src;
     my $content_info;
     my @sorted_output;
-    
-    if ($xcatdebugmode) {
-        my $debug_info = "Processing command: rvitals $grep_string";
-        process_debug_info($node, $debug_info);;
-    }
-    print Dumper(%{$response_info->{data}});
 
     foreach my $key_url (keys %{$response_info->{data}}) {
         my %content = %{ ${ $response_info->{data} }{$key_url} };
@@ -1876,8 +1861,6 @@ sub rflash_response {
     my $response = shift;
 
     my $response_info = decode_json $response->content;
-
-    print Dumper(%{$response_info->{data}});
 
     my $update_id;
     my $update_activation = "Unknown";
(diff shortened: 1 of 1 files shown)


gurevichmark:
I thought the idea for #3796 was to print to the console and the log what data comes back in the response. So for example `curl software/enumerate` would put the Dumper() style response objects into the log file.

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1067,19 +1068,20 @@ sub gen_send_request {
...
-        }
-        else {
-            $debug_info = "$method $request_url -d $content";
+    if ($xcatdebugmode) {
+        my $debug_info;
+        if ($method eq "GET") {
+            $debug_info = "curl -k -b cjar -X $method -H \"Content-Type: application/json\" $request_url";
+        } else {
+            if ($::UPLOAD_FILE) {
+                # Slightly different debug message when doing a file upload
```
whowutwut: @gurevichmark can you take a look here?  On the upload, would we want to print the command that we used to upload, for debug?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1215,6 +1239,9 @@ sub rpower_response {
     xCAT_monitoring::monitorctrl::setNodeStatusAttributes(\%new_status, 1) if (%new_status);
 
     if ($node_info{$node}{cur_status} eq "RPOWER_STATUS_RESPONSE" and !$next_status{ $node_info{$node}{cur_status} }) { 
+
+        print Dumper(\%{ $response_info->{data} }) . "\n";
```
whowutwut: What's the purpose of adding the dumper out put here? And does it impact the performance of the commands? @robin2008 Is there any way to tell?   It's not wrapped in debugmode, so on all calls, we would be running this, regardless of bypass or not....  

And the output would only come out in non-hierarchy environments, we cannot use XCATBYPASS on hierarchy mode....

xCAT-server/lib/xcat/plugins/openbmc.pm
gurevichmark: @whowutwut You are right.
@xuweibj The $request_url for the file upload on the next line should be:
`$request_url  = "$http_protocol://" . $node_info{$node}{bmc} . "/upload/image/"`

whowutwut:
@gurevichmark I'm not sure we want to hold that much debug information in the log files.  

Here's the scenario that I'm trying to help with debugging... 
When customer reports some incorrect data in xCAT output on commands that gather data from the BMC.  We need to know what the BMC is sending back to us to verify whether it's a processing error on xCAT or the BMC data is incorrect. 

Debug would be: 
- As the user to rerun the command with `xcatdebugmode=1` in the site table

The problem here is that it only shows the request and not the response.  Then to ask the user to provide us with the response, we have to spend time to give them a correctly formatted curl command, Even for developers to come up with the command, it takes some time.  So to reduce this time spent formatting a correct CURL command, I suggested that we just print this out with the debug, so the code generates the command for the user to run, saving everyone time. 


Then the next question is whether we should only print this for debugmode=2... ?

xCAT-server/lib/xcat/plugins/openbmc.pm
xuweibj: Yes, it's just for ourself to debug when using XCATBYPASS=1 in non-hierarchy environments. I think if there is some changes for ``rinv``, we could use it to debug quickly.

xCAT-server/lib/xcat/plugins/openbmc.pm
xuweibj: @gurevichmark , sorry I did not get your point. Is the url defined below is wrong? If not, I think the url on the next line is OK. Or if it is wrong, we should modify the hash. On line 1065, we have added http and bmc in the string.
```
RFLASH_FILE_UPLOAD_REQUEST  => {
        method         => "PUT",
        init_url       => "$openbmc_project_url/upload/image/",
    },
```

xCAT-server/lib/xcat/plugins/openbmc.pm
robin2008: We can use profile tool to see if the Dumper take more time,  but I think it should be less performance impact unless it is very large.
So if it is only for debugging,  we need check the environment XCATBYPASS or foreground mode before running it. Then it won't impact the customer and be convenient for our debugging.

xCAT-server/lib/xcat/plugins/openbmc.pm
gurevichmark: @xuweibj It is a little hard to explain. Lets just merge this pull request as is, and I will do another pull request on top if it later to display correct info for "file upload"

Issue #3543, opened 2017-07-26
Need to fully understand rcons in openbmc environment


whowutwut:
There's something that we don't quite understand with conserver and how it interacts with the openbmc ssh port...   I need someone to really understand if we are doing things correctly and when we need to re-run makeconservercf ....  

With OpenBMC ssh port to expose the console, we should be able to have multiple console sessions opened.   Here are some snips of what I have seen some issues with but really don't have the time to investigate.. 

Some questions: 
* [ ] what causes the: `console: connect(): 782@localhost: Connection refused`
* [x] When do we need to run makeconservercf: 
    It seems we don't need to run makeconsercf with sshpass or `-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null` fix
* [x] What happens when firmware is updated , can the console re-connect
* [ ] Can we do better in `rspconfig <> sshcfg` it seems to always append... 

### connection refused until after makeconservercf, can't re-create this... 
```
[root@fs2vm110 ~]# rspconfig p9euh01 sshcfg
[OpenBMC development support] Using this version of xCAT, ensure firware level is at v1.99.6-0-r1, or higher.
p9euh01: ssh keys copied to 10.6.7.254
[root@fs2vm110 ~]# rcons p9euh01
console: connect(): 782@localhost: Connection refused
[root@fs2vm110 ~]# rcons p9euh01
console: connect(): 782@localhost: Connection refused
[root@fs2vm110 ~]# makeconservercf p9euh01
[root@fs2vm110 ~]# rcons p9euh01
[Enter `^Ec?' for help]
[disconnect]
```

### After copying the ssh keys, still required password when doing ssh.... ???! 
```
[root@fs2vm110 ~]# rspconfig p9euh01 sshcfg
[OpenBMC development support] Using this version of xCAT, ensure firware level is at v1.99.6-0-r1, or higher.
p9euh01: ssh keys copied to 10.6.7.254
[root@fs2vm110 ~]# rcons p9euh01
[Enter `^Ec?' for help]



^A^?[disconnect]
[root@fs2vm110 ~]# ssh 10.6.7.254
Warning: Permanently added '10.6.7.254' (RSA) to the list of known hosts.
root@10.6.7.254's password:
```

### Does multiple rcons work from different servers defining the same machine?

whowutwut:
@zet809  who should look into this? We probably should understand this when we are doing firmware updates...    maybe @gurevichmark or @xuweibj ?

[...]

PR #3815, opened 2017-08-30
Disable the ssh prompt when connecting the openbmc console

chenglch:
This patch force the ssh client to use the ssh key only. It also
disable the known host check to avoid of the error after rflash.

Fix-issue: #3543

### DIFF
--- a/xCAT-server/share/xcat/cons/openbmc
+++ b/xCAT-server/share/xcat/cons/openbmc
@@ -93,12 +93,11 @@ if ($ENV{SSHCONSOLEPORT}) {
 
 # To automatically connect to the console without the need to send over the ssh keys, 
 # ensure sshpass is installed on the Management and/or Service Nodes.
+print "If unable to open the console, ensure sshpass is installed or ssh keys have been configured on the BMC.\n";
 if (-x '/usr/bin/sshpass') {
-    exec "/usr/bin/sshpass -p $password ssh -p $sshport -l $username $bmcip";
+    exec "/usr/bin/sshpass -p $password ssh -p $sshport -l $username -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $bmcip";
 } else {
-    # This will prompt the user to enter the password for the BMC 
-    # if ssh keys are not sent and sshpass is not being used
-    exec "ssh -p $sshport -l $username $bmcip";
+    exec "ssh -p $sshport -l $username -o PasswordAuthentication=no -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $bmcip";
 }
 
 


whowutwut:
@chenglch What about this issue? 
```
[root@briggs01 20170829_VLAN_DEV]# rcons node-70e28414280d
console: SSLVerifyCallback(): error with certificate at depth: 0
console: SSLVerifyCallback():  issuer  = /CN=xCAT CA
console: SSLVerifyCallback():  subject = /CN=briggs01.pok.stglabs.ibm.com
console: SSLVerifyCallback():  error #7: certificate signature failure
console: SSL negotiation failed
70367064202656:error:04067084:rsa routines:RSA_EAY_PUBLIC_DECRYPT:data too large for modulus:rsa_eay.c:738:
70367064202656:error:0D0C5006:asn1 encoding routines:ASN1_item_verify:EVP lib:a_verify.c:249:
70367064202656:error:14090086:SSL routines:ssl3_get_server_certificate:certificate verify failed:s3_clnt.c:1264:
```
Maybe this is not something that we can handle without running makeconservercf again. 

If not, then we need to at least have a documented process for this.  Or maybe after the flash of the firmware, we run `makeconservercf <node>` but this only works once the BMC is active...

chenglch:
@cxhong , if ssh key is not there, the user should run `rspconfig sshcfg` to setup the ssh key. If ssh prompt is enabled, ssh session will hang and wait for the input, this stop the conserver from reconnecting automatically.  This may help resolve some problem when setting `consoleondemand=no` as the user should not use `rcons` to enter the password manually.

@whowutwut , how about running `rspconfig sshcfg` before the `rcons`, with this patch and if the `sshpass` is not enabled, conserver is always trying to connect to the bmc, once the ssh key is ready, the session will setup.  

`makeconserver <node>` is not needed as it affects all of the node under conserver.

whowutwut:
@chenglch Is it better to require ssh keys to be sent over to the BMC or for xCAT to ship `sshpass`? 
 
Right now the ssh keys do not persist on the BMC with a flash, I don't know when that will get into the persistent memory of the firmware, so until then... this would have to be run each time we flash the firmware...

whowutwut:
I don't think we should disable the password prompting because we don't say that ssh keys must be required. 

The removal of the known hosts is good. 

But thinking about this more. Maybe we should ship sshpass-xcat and handle this that way.. 

I think it would work for both bmc discovered nodes and xcat managed nodes without the sshcfg being used

chenglch:
@whowutwut , I think disabling the password prompting is not conflict with requiring the ssh keys as  the way with `sshpass` is still kept.  In my opinion I prefer the way with `ssh-key` as it is a normal way, we could add  some message for user after `rflash` or `discovery` to tell the user to configure the `sshcfg` for the BMC.
If `sshcfg` is not our default option, we have to build packages for `sshpass-xcat` for multiple platforms, it will take too much effort.

cxhong:
I just ran into this issue yesterday.   Console moved on after I enter the password. I prefer this way.  

with this PR,  it will keep running "permission denied" message on the console window  and there is no message indicated user needs to send ssh-key to the BMC.

whowutwut:
I'm not sure why the password prompt does not happen automatically.   When we `ssh -p 2200` we get the password prompt all the time, but under rcons, is it because the session is already open previously by conserver? and so it's waiting there? 

The danger here is that we won't be capturing log messages to the files unless ssh keys are sent or ssh-pass is used....   

What happens in the case when we have ssh keys sent, but the firmware is updated, would it automatically recover?  I think we have to decide a method and go with it... then start testing with firmware updates...

chenglch:
@cxhong , if conserver is running under the setting of `consoleondemand=no`, the user may not run `rcons` for every node and then enter the password. 

Conserve can rerun the command automatically if the command ends, but the prompt prohibit this.

My opinion is that: 
 As the ssh setting is lost on the BMC side after `rflash`, we could document the steps to run `rspconfig sshcfg` , as this is the limitation of current openbmc implementation, upgrade the firmware and then configure it, this is reasonable.

sshpass is another way we can accept, but it takes too much effort to build the package like `sshpass-xcat`.

whowutwut:
Some discussion here on the openbmc GitHub https://github.com/openbmc/openbmc/issues/1517

xCAT-server/share/xcat/cons/openbmc
```diff
@@ -94,11 +94,9 @@ if ($ENV{SSHCONSOLEPORT}) {
 # To automatically connect to the console without the need to send over the ssh keys, 
 # ensure sshpass is installed on the Management and/or Service Nodes.
 if (-x '/usr/bin/sshpass') {
-    exec "/usr/bin/sshpass -p $password ssh -p $sshport -l $username $bmcip";
+    exec "/usr/bin/sshpass -p $password ssh -p $sshport -l $username -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $bmcip";
 } else {
-    # This will prompt the user to enter the password for the BMC 
-    # if ssh keys are not sent and sshpass is not being used
-    exec "ssh -p $sshport -l $username $bmcip";
+    exec "ssh -p $sshport -l $username -o PasswordAuthentication=no -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null $bmcip";
```
whowutwut: Maybe before we exec this section starting at line  97, we need to inform the user incase they do not have sshpass or ssh keys sent to the BMC, they will not know why the console does not respond.  

Today the password prompt will come up and it's intuitive and no one has complained about it, but I think if we constantly get Permission Denied, no one will know how to recover this 

Some thing like:
```
If unable to open the console, ensure sshpass is installed or ssh keys have been configured on the BMC.
```

chenglch:
@whowutwut , I have updated the patch based on you comments. For the password prompt, my concern is the console session running in the background, there is no chance for the user to enter the password. The result is the console session hangs, no any output in `/var/log/console/<node>` log file.

immarvin:
according to the comments from @whowutwut and talked with @chenglch , merge this PR first, any refinement can performed in the following release

whowutwut:
@chenglch I'm not sure I understand the password prompt issue.  I did work with some admin on Friday that updated xcat and we saw this message scrolling on the screen..  I'm working on getting sshpass into xcat dep but need approval.  Meeting on Monday regarding this...

chenglch:
I understand the phenomenon you mentioned, if the BMC connection is lost, `conserver` is always trying to reconnect to the BMC until the connection can be established.  `sshpass` could solve some part of this, but if the BMC is not reachable, message scrolling may be still exist in the log file.

chenglch:
@whowutwut , the password prompt may have some issue when the `ssh` session under the conserver is running in the background. For example, the console is running in `consoleondemand=no` mode for 500 nodes, the console log(especially the booting message) should be kept in the console log files automatically. But if there is password prompt, it will hangs and wait for the input with `rcons`, if not, the logs from console is missing.

Issue #3811, opened 2017-08-29
[OpenBMC] rspconfig sshcfg not working


whowutwut:
@gurevichmark This doesn't seem to be working quite as expected....

```
[root@briggs01 xcat]# rspconfig node-70e28414280d sshcfg
node-70e28414280d: ssh keys copied to 172.12.253.119
[root@briggs01 xcat]# ssh root@172.12.253.119
root@172.12.253.119's password:
root@xyz123:~# ls -ltr .ssh/authorized_keys
ls: .ssh/authorized_keys: No such file or directory
root@xyz123:~# ls -ltr .ssh
total 0
root@xyz123:~#
```

Can you validate?

gurevichmark:
Fixed by #3817

zet809:
Hi, @whowutwut , can we close this issue if it is closed?

PR #3817, opened 2017-08-30
Make sure authorized_keys file was created

gurevichmark:
#3811 

For some unknown reason, the `echo` command to create `~/.ssh/authorized_keys` file fails when running `rspconfig sshcfg` command. This happens only when creating this file for the first time after `.ssh` directory is created. Occurs about 1 in 5 attempts.

This pull request adds a check to make sure `~/.ssh/authorized_keys` file did get created, and reports an error to the user if this check fails:

```
[root@stratton01 gurevich]# rspconfig p9euh02 sshcfg
p9euh02: Error copying ssh keys to 10.6.9.254 Rerun rspconfig command.
```

```
[root@stratton01 gurevich]# rspconfig p9euh02 sshcfg
p9euh02: ssh keys copied to 10.6.9.254
[root@stratton01 gurevich]#
```


whowutwut:
Let's merge this so we have better reporting when we hit these failures.

PR #3819, opened 2017-08-31
Remove the call to logger before rsyslog is started in Genesis doxcat 

whowutwut:
This was introduced in PR #1366, but removing the lines to clean up the messages that appear in Genesis

The output when Genesis runs is the following: 
```
logger: socket /dev/log: No such file or directory
initramfs  lock  log  mount  screen  udev         
<166>Aug 30 22:17:21 xcat.genesis.doxcat: Beginning doxcat process...
Done                                                      
```

* The first is caused by logger being called before syslog is started
* The second is listing out /var/run.  I'm not sure why I added that, but I don't think there's any value to printing that out...


zet809:
Looks good to me, I agree to merge. Thx @whowutwut

Issue #3737, opened 2017-08-22
[FVT]:updatenode still return 0 when syncfile does not exist.


tingtli:
updatenode still return 0 when syncfile does not exist.
The steps to reproduce
```
1. mkdir -p /install/custom/install/rhels7.3/
2. echo "/tmp/non-existent -> /etc/motd" > /install/custom/install/rhels7.3/booboo.synclist
3. chdef -t osimage -o rhels7.3-ppc64-install-compute synclists=/install/custom/install/rhels7.3/booboo.synclist
4. updatenode c910f02c01p23 -F
[root@c910f02c01p22 /]# updatenode c910f02c01p23 -F
c910f02c01p23: rsync: link_stat "/tmp/non-existent" failed: No such file or directory (2)
c910f02c01p23: rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1039) [sender=3.0.6]
File synchronization has completed for nodes.
[root@c910f02c01p22 /]# echo $?
0
```

zet809:
hi, @bybai , will you pls help to take a look at this issue? Thx!

bybai:
This is not high priority, will look at it in next sprint or else. Thx.

hu-weihua:
In rhels6.9+ppc64, except above problem, there is below problem too.  using the same steps, the output of ``updatenode`` is:

    [root@c910f02c01p13 autotest]# cat /tmp/updatenode.F.out
    There were no syncfiles defined to process. File synchronization has completed.

It does not make sense.

robin2008:
updatenode will call xdcp to do the file sync, and it will return error.
```
xdcp sn02 --nodestatus -F /tmp/abc.lst
sn02: Remote_command_failed, error_code=23
sn02: rsync: link_stat "/tmp/non-existent" failed: No such file or directory (2)
sn02: rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1052) [sender=3.0.9]
```

updatenode should handle the error and set the return code accordingly.

robin2008:
Fix it by PR #3816,  when some nodes are failed,  the "buildnodestatus" method will fill the failed nodes into @::FAILEDNODES

robin2008:
And it might require more wide discussion.  

Maybe the existing design is the right:
- the return code of updatenode is only to response the updatenode operation is triggered successfully or not.
- but the operation result for each node should be got from the DB (updatestatus)

[...]

PR #3831, opened 2017-09-01
Fix bug #3737 in automation environment

hu-weihua:
For bug #3737.

In automation environment, test case ``updatenode_diskful_syncfiles_failing`` is run after test case ``rmimage_diskless``. 

In test case ``rmimage_diskless``, it changed `` provmethod`` of ``$$CN`` to another value, not the ``__GETNODEATTR($$CN,os)__-__GETNODEATTR($$CN,arch)__-install-compute``. So when ``updatenode_diskful_syncfiles_failing`` run, it does not work as expected. 

It is also a defect of ``updatenode_diskful_syncfiles_failing`` itself, so fix defect in this pull request.


neo954:
Okay, will merge this one.

PR #3844, opened 2017-09-05
Adjust the order of nodeset_disjointdhcps_petitboot in daily regression bundles

hu-weihua:
Due to ``nodeset_disjointdhcps_petitboot`` needs hierarchy environment, so move this case after SN setup. Due to there are not hierarchy in ubuntu environments, so delete this case from ubuntu.


bybai:
I am OK to merge it.

PR #3847, opened 2017-09-05
Ensure mypostscript has permission 700

mattaezell:
Only root needs to execute this file

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/post.debian
+++ b/xCAT-server/share/xcat/install/scripts/post.debian
@@ -77,7 +77,7 @@ do
          done
 
             
-         chmod +x /xcatpost/mypostscript
+         chmod 700 /xcatpost/mypostscript
 			GOTIT=1
 			break
 		fi
--- a/xCAT-server/share/xcat/install/scripts/post.rhel5.s390x
+++ b/xCAT-server/share/xcat/install/scripts/post.rhel5.s390x
@@ -68,7 +68,7 @@ do
             done
 
             # Make executable
-            chmod +x /xcatpost/mypostscript
+            chmod 700 /xcatpost/mypostscript
             GOTIT=1
             break
         fi
--- a/xCAT-server/share/xcat/install/scripts/post.rhel6.s390x
+++ b/xCAT-server/share/xcat/install/scripts/post.rhel6.s390x
@@ -68,7 +68,7 @@ do
             done
 
             # Make executable
-            chmod +x /xcatpost/mypostscript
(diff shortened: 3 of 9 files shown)


whowutwut:
@immarvin can you review the changes here?

xCAT-server/share/xcat/install/scripts/post.debian
```diff
@@ -60,6 +60,7 @@ do
          rm -rf $i
          chmod +x /xcatpost/*
          /xcatpost/getpostscript.awk |egrep '<data>' |sed -e 's/<[^>]*>//g'|egrep -v '^ *$'|sed -e 's/^ *//' | sed -e 's/&lt;/</g' -e 's/&gt;/>/g' -e 's/&amp;/\&/g' -e 's/&quot;/"/g' -e "s/&apos;/'/g" > /xcatpost/mypostscript
+         chmod 700 /xcatpost/mypostscript 2>/dev/null
```
immarvin: hi @mattaezell , I think this PR should only replace the strings like ``chmod +x mypostscript`` or ``chmod 777 mypostscript`` to ``chmod 700 mypostscript``, do not need to chmod the requested /xcatpost/mypostscript

mattaezell: My goal was to reduce the amount of time that the _mypostscript_ file has 755 permissions.  It should be very brief, so I guess I can remove these lines if you think they are superfluous.

immarvin:
hi @mattaezell , get your point. My though is that before mypostscript is ready run, the content of requested mypostscript is either the error response or raw data,  I suggest to remove the lines, sorry for the inconvenience, thanks

xCAT-server/share/xcat/install/scripts/post.debian
neo954: @mattaezell, for this purpose, you may add `umask 0077` in somewhere in this file.
Maybe, change line 62 to `( umask 0077 ; /xcatpost/getpostscript.awk |egrep '<data>' |sed -e 's/<[^>]*>//g'|egrep -v '^ *$'|sed -e 's/^ *//' | sed -e 's/&lt;/</g' -e 's/&gt;/>/g' -e 's/&amp;/\&/g' -e 's/&quot;/"/g' -e "s/&apos;/'/g" > /xcatpost/mypostscript )`
Thus, only one line is affected with `umak 0077`. Then line 63 can be omitted.

immarvin:
I am ok with the change, agree to merge

Issue #3822, opened 2017-08-31
[cumulus] refine the syslog script for cumulus switch


immarvin:
* [ ] refine the syslog script to make the cumulus switch log forwarding consistent with the compute nodes
* [ ] make syslog compatible with configonie

immarvin:
hi @cxhong ,  i noticed that there is strings like "for Cumulus OS inside ONIE switches" in syslog script,  have we already  finished the syslog support for cumulus ?

cxhong:
No, I don't think so

immarvin:
made a simple test, seems that syslog has support cumulus. but it's not included in configonie

immarvin:
hi @cxhong , I think we need add "syslog","configinterface","enablesnmp",config passwordless root ssh, "configntp" as the default operation in ``switchdiscover --setup``, would you please take a look at this?

thanks

cxhong:
the PR is merged, will close this issue.

PR #3850, opened 2017-09-05
Support syslog for Cumulus Switch

cxhong:
For issue #3822 
Add syslog postscript to default options of ```switchdiscover --setup``` for Cumulus switch


immarvin:
I am ok with the change

PR #3858, opened 2017-09-06
Modify the order of nodeset_cmdline nodeset_shell and nodeset_runimg three test cases in bundle

hu-weihua:
Due to these 3 test cases ``nodeset_cmdline nodeset_shell  nodeset_runimg`` will trigger CN enter into genesis, although before exit, these cases try to recover CN to original status, but the code is not good enough, they can not guarantee CN back to original status. So these three cases always effect other test cases which need to operate CN. So move these 3 cases to a place where won't effect other test cases.


bybai:
I am ok to merge it.

Issue #3487, opened 2017-07-18
[FVT] add test case for issue 3088 : nodeset improvement


hu-weihua:
Add test case for issue #3456 
what to do:

* [ ] offer test case list and description
* [ ] implement these test cases
* [ ] verify issue #3088 before 2.13.6 release

hu-weihua:
@neo954, could you list your test plan in the comment, you have delayed than origin plan one week. One more thing, please discuss your test plan with @robin2008  first before you paste you plan in comment. thanks

hu-weihua:
@neo954, TWO weeks have passed of this sprint. You still didn't update anything for this task. Could you update what is you plan for this task?

robin2008:
Disjoint mode Design is already finalized. The test case need to cover below scenario with SNs at least: 

sharedtftp=0, disjointdhcps=1 :  
   - w/ dhcpserver/dynamic range defined in networks
   - w/o dhcpserver/dynamic range defined in networks
dhcpserver will handled whole node range, and other sn will handle its own children.


 sharedtftp=0, disjointdhcps=0: 
The nodeset request will be dispatched to all MN and SNs

 sharedtftp=1 (Assume that old cases should already cover it)

Extra for "query" action;
nodeset xxx stat -a in disjoint mode.

PR #3871, opened 2017-09-07
Fix test case nodeset_disjointdhcps_petitboot

neo954:
Force to create the fake osimage directory on SN, in case it could cover the case /install is not mounted. (The feature is #3088, and the test case issue is #3487)


robin2008:
@neo954 
I think you got the wrong root cause,  the fake sn is added after those `nodeset` command.

The real cause might be as below:
```
Error: Install image not found in /install/rhels7.99/ppc64le
testnode1: install rhels7.99-ppc64le-compute
CHECK:rc == 0   [Failed]
```

PR #3872, opened 2017-09-07
Fix the servicenode return nonzero on stateful service node

robin2008:
Solution:
Not restart xcatd in 'servicenode' when the node is not stateless and statelite as it will fail in postscript stage.


immarvin:
i am ok with the change

PR #3881, opened 2017-09-08
Remove mess information form ssh keys and credential files

### LINKED ISSUE #3856: [DEV] Found mess information <xcatdsource> in ssh keys on CN
Although there is not found function impact, but it is not right behavior.

```
~/.ssh # grep -r "xcatdsource" ./*
./authorized_keys:  <xcatdsource>c910f03c11k09</xcatdsource>
./id_rsa:  <xcatdsource>c910f03c11k09</xcatdsource>
./id_rsa.pub:  <xcatdsource>c910f03c11k09</xcatdsource>
c910f03c11k10:~/.ssh # grep -r "xcatdsource" /etc/*
/etc/ssh/ssh_host_dsa_key:  <xcatdsource>c910f03c11k09</xcatdsource>
/etc/ssh/ssh_host_rsa_key:  <xcatdsource>c910f03c11k09</xcatdsource>
/etc/ssh/ssh_host_ecdsa_key:  <xcatdsource>c910f03c11k09</xcatdsource>
/etc/xcat/hostkeys/ssh_host_dsa_key:  <xcatdsource>c910f03c11k09</xcatdsource>
/etc/xcat/hostkeys/ssh_host_ecdsa_key:  <xcatdsource>c910f03c11k09</xcatdsource>
/etc/xcat/hostkeys/ssh_host_rsa_key:  <xcatdsource>c910f03c11k09</xcatdsource>
```

robin2008:
Fix #3856 
Remove the mess info `<xcatdsource>` from the content when call `getcredentials.awk`

Root cause: 
 we introduce `<xcatdsource>` in xcat response, but some awk script does not handle it.
Then it will be written to the final file on CN.

Now just filter the line with `<xcatdsource>` in such awk scripts.

UT: Passed.
1, Patch the awk script on MN
2, To check if there mess information on MN's files (if there are, we need to clean up first)
3, reprovision CN
4, grep "xcatdsource"  in /root/.ssh,  /etc/

There should be no `<xcatdsource>` line.
```
# cd ~/.ssh
# grep -r xcatdsource ./*
# cd /etc
# grep -r xcatdsource ./*
```

### DIFF
--- a/xCAT/postscripts/getcredentials.awk
+++ b/xCAT/postscripts/getcredentials.awk
@@ -16,7 +16,9 @@ BEGIN {
         print "</xcatrequest>" |& server
 
         while (server |& getline) {
-                print $0 
+                if (match($0,"<xcatdsource>") == 0) {
+                  print $0
+                }
                 if (match($0,"<serverdone>")) {
                   quit = "yes"
                 }
--- a/xCAT/postscripts/getpostscript.awk
+++ b/xCAT/postscripts/getpostscript.awk
@@ -22,7 +22,9 @@ BEGIN {
                   start = 1
                 }
                 if (start == 1) {
-                  print $0
+                  if (match($0,"<xcatdsource>") == 0) {
+                    print $0
+                  }
                 }
 
                 if (match($0,"<serverdone>")) {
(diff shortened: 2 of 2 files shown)


immarvin:
i am ok with the change

Issue #3873, opened 2017-09-07
rflash recovery support for Supermicro based firmware


chenglch:
Add command `rflash <node> --recover <bmc_image>` to recover the Supermicro firmware which is in the Brick Protection state.

whowutwut:
@chenglch This is required for hardware GA, so we could set a later milestone.. FYI.

zet809:
Hi, @whowutwut , shall we move this task to 2.13.8 for the verification? It is very hard for us to do the FVT.

whowutwut:
Yea I think the hardware GA is later this year

zet809:
Let's remove the milestone and add one when the server is ready to verify.

whowutwut:
@zet809 Was some activity planned here for this sprint? 2.13.9?

zet809:
Sorry, @whowutwut , not planed in 2.13.9. Shall we move to 2.13.10?

zet809:
hi, @whowutwut , not sure whether this interface has been used? If not ,we shall verify it by ourselves.

whowutwut:
I think we still have to do something here.... but our hardware is a little limited at the moment.

zet809:
hi, @whowutwut , do you think we can use #3374 to cover this issue?

gurevichmark:
Also message displayed if node is not in "Brick Protection state" by #4998

whowutwut:
Will need some more time here to validate the process of recovering the SMC node before we attempt to brick it.  Don't want to end up with a bricked server...  Will try and do during 2.14.11

whowutwut:
I was able to validate this brick protection recovery on a machine that is able to brick by flashing firmware almost every time.  It works.  This is good news! Great job @chenglch! 

When the BMC failed to flash... the BMC becomes unresponsive:
```
# rpower boston_brick state
boston_brick: [xcat1]: Error: ERROR: timeout
```

There's no way to query in this state whether it's in brick protection, and unless you  have a serial cable directly plugged to the node, there's no activity that will be seen remotely. 

Attempt to upload the BMC firmware again in this recovery mode: `rflash <node> --recover <path_to_BIN_file>` 
 
Example:
```
# rflash boston_brick --recover /firmware/Boston/P9_123_BMC_01302018pnor/SMT_P9_123.bin
boston_brick: Successfully updated recovery image. BMC is restarting and will not be reachable for 5-10 minutes.Recover the BMC firmware 
```

[...]

PR #3882, opened 2017-09-08
Add rflash recovery support for Supermicro based firmware

chenglch:
This commit add recover option for rflash command to upload
bmc image with tftp when the BMC is in Brick Protection state.

Example:
```
rflash <node> --recover <image>
```
implement #3873

### DIFF
--- a/xCAT-client/pods/man1/rflash.1.pod
+++ b/xCAT-client/pods/man1/rflash.1.pod
@@ -26,6 +26,8 @@ B<rflash> I<noderange> I<http_directory>
 
 B<rflash> I<noderange> [I<hpm_file_path> | B<-d> I<data_directory>] [B<-c>|B<--check>] [B<--retry=>I<count>] [B<-V>]
 
+B<rflash> I<noderange> B<--recover> I<bmc_file_path>
+
 =head2 OpenPOWER OpenBMC specific :
 
 B<rflash> I<noderange> {[B<-c>|B<--check>] | [B<-l>|B<--list>]}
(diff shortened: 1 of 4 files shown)


xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -2494,6 +2495,38 @@ sub do_rflash_process {
...
+                        errorcode => 1 });
+                exit(1);
+            }
+            # Sometimes tftp command retrun error message but without nonzero error code
+            if($output) {
+                $callback->({ error => "Running tftp command \'$cmd\' failed. Output: $output",
+                        errorcode => 1 });
+                exit(1);
+            }
+            $callback->({ data => "$node: Success to update the recovery image. BMC is restarting and is not reachable within 5-10 minutes."});
```
gurevichmark: Better message:
"Successfully updated recovery image. BMC is restarting and will not be reachable for 5-10 minutes."

xCAT-client/pods/man1/rflash.1.pod
```diff
@@ -135,6 +135,10 @@ Used to commit the flash image in the temporary side of the chip to the permanen
 
 Used to recover the flash image in the permanent side of the chip to the temporary side for both managed systems and power subsystems.
 
+OpenPOWER BMC specific (using IPMI):
+
+Used for IBM Power S822LC for Big Data systems only. Specifies the BMC image downloaded from FixCentral.
+
```
gurevichmark: 1. I think a new Options entry should be created instead of reusing existing `--recover`. The existing entry does not take any parameters, but this new one does and should be displayed as `--recover I<image>`. It might be useful to add one or two sentences explaining what this option will actually do, just so that a user has some idea what will happen if they choose this option.

2. This new `--recover` option should be added to the "Synopsis" section under "OpenPOWER BMC specific (using IPMI):". Maybe add a new line like: `B<rflash> I<noderange> B<--recover> I<image>` ?

3. Usage.pm needs to be modified to add this new option.

chenglch: @gurevichmark , I modified the usage of `--recover`, currently, this option do not take any parameters, it works as a boolean flag to indicate the command type. The format like`rflash <noderange> <image> --recover` also works.

xCAT-server/lib/xcat/plugins/ipmi.pm
chenglch: Done

PR #3884, opened 2017-09-08
change testcase fake ip to fix 12.1.0.100 is used by other host 

bybai:
In auto test rhels6.9,  `confignetwork_2eth_bridge_br0` is failed, and ip 12.1.0.100 is used by other host.
Since this ip is used by different test cases, we have several auto test environments, 12.1.0.100 may be configured in other CN. Here, I changed it in this failed case, and try to see if it is fixed next day.
```
c910f04x35v04: bring up ip
c910f04x35v04: Determining if ip address 12.1.0.100 is already in use for device eth2...
c910f04x35v04: Error, some other host (FF:FF:FF:FF:FF:FF) already uses address 12.1.0.100.
c910f04x35v04: [E]:Error: ifup eth2 failed.
```


hu-weihua:
It is ok for me, agree to merge, thanks @bybai .

Issue #3823, opened 2017-08-31
[cumulus] cumulus ZTP process enhancement


immarvin:
* [ ] add a switch in site table to disable ZTP 
* [x] refine the documulusdiscovery script to reflect the changes introduced in https://github.com/xcat2/xcat-core/pull/3601

immarvin:
https://github.com/xcat2/xcat-core/pull/3886

whowutwut:
@immarvin what is the intent of having the site table value to disable ztp? Would it also enable it? Why not just have a command?

immarvin:
hi @whowutwut , since we are not sure whether ZTP is enabled in cumulus switches shipped in CRL, if the ZTP is well configured on MN, the switches will walk through the discovery and configuration 1st time on power on, then when we discover switches with "switchdiscover --setup", the switches will be configured for the second time, this might cause some problem. what do you think? @whowutwut @cxhong

cxhong:
if switch discovered and configured 1st time using ZTP,   the ip address should change to static ip address already.   The user will pass in range of DHCP ip address to run switchdiscover --setup command,  so I don't think there will have problems.

PR #3886, opened 2017-09-08
refine cumulus discovery process

immarvin:
For task https://github.com/xcat2/xcat-core/issues/3823:

refine cumulus discovery procedure to handle the different response from the MN

### DIFF
--- a/xCAT/postscripts/documulusdiscovery
+++ b/xCAT/postscripts/documulusdiscovery
@@ -96,22 +96,37 @@ if [ -z "$XCATMASTER" ] || [ -z "$XCATPORT" ]; then
 fi
 
 
-#send the find me request ot xcatd on MN
-MAX_RETRY=10
+#send the find me request to xcatd on MN
+MAX_RETRY=1000
 RETRY=0
 DISCOVERED=0
 
 while [ $RETRY -lt $MAX_RETRY ]; do
     logger -s -t $log_label -p local4.info "Sending the discovery packet to xCAT ($XCATMASTER:$XCATPORT) [ RETRY= $RETRY ]  ..."
+    echo >/tmp/result.socat.out
     (cat /tmp/discopacket.gz | socat STDIN UDP:$XCATMASTER:$XCATPORT,sourceport=301 ) &
 
     WAITRETRY=0
-    while [ $WAITRETRY -lt 100 ]; do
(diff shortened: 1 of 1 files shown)


immarvin:
hi @cxhong , would you please take a look at this?

xCAT/postscripts/documulusdiscovery
```diff
@@ -97,29 +97,43 @@ fi
 
 
 #send the find me request ot xcatd on MN
```
cxhong: ot ----should be to

xCAT/postscripts/documulusdiscovery
```diff
@@ -97,29 +97,43 @@ fi
...
-MAX_RETRY=10
+MAX_RETRY=1000
```
cxhong: that number seems very long...

xCAT/postscripts/documulusdiscovery
```diff
@@ -97,29 +97,43 @@ fi
...
+        if [ -f "/tmp/result.socat.out" ];then
+            if grep "restart" /tmp/result.socat.out; then
+                DISCOVERED=1
+                break;
+            elif grep "processing" /tmp/result.socat.out; then 
+                echo "My findme request is still under processing, do not send new request"
+                logger -s -t $log_label -p local4.info  "My findme request is still under processing, do not send new request"
+                sleep 5
+            elif grep "processed" /tmp/result.socat.out; then
+                echo "Fail to discover me, retry to send findme request 1 second later!"
```
cxhong: do u want to wait for 1 second or 10 second?

xCAT/postscripts/documulusdiscovery
immarvin: regarding to node discovery, the findme request will be resent forever. Since cumulus discovery is performed in ZTP process, we'd better try to resend adequate number of findme requests, while quit on fail after a reasonable time.

xCAT/postscripts/documulusdiscovery
immarvin: 10, corrected

xCAT/postscripts/documulusdiscovery
immarvin: corrected

immarvin:
hi @cxhong , thanks for your comments, code updated

PR #3891, opened 2017-09-11
Add regex helpers

mattaezell:
This adds some helper functions that can make regular expressions shorter.

a2idx: Turns an ASCII character into an index (a=1, b=2, etc)
a2zidx: Turns an ASCII character into a 0-index (a=0, b=1, etc)
dim2idx: Turns dimensions (eg row, column, rack-unit) into an index
skip: Return an index where some items are 'skipped'
ipadd: Add a value to an IP address

### DIFF
--- a/docs/source/guides/admin-guides/references/man5/xcatdb.5.rst
+++ b/docs/source/guides/admin-guides/references/man5/xcatdb.5.rst
@@ -157,6 +157,27 @@ Before you panic, let me explain each column:
 See http://www.perl.com/doc/manual/html/pod/perlre.html for information on perl regular expressions.
 
 
+Regular Expression Helper Functions
+============================
+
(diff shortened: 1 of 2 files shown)


mattaezell:
These were developed to support the Summit CORAL machine at ORNL.  Some background... Cabinets are named with a row (a-h) and column (01-36), so it's clearer and more convenient to use `a2idx($1)` than `ord($1)-96`.  There are several 'holes' in the numbering due to non-compute racks in the middle of the machine and facility obstructions, so the `skip()` function makes it so we don't have to make a separate group for each contiguous run of cabinets.   `dim2idx` is more compact than doing it native.  `ipadd` is helpful because we need to cross /24 boundaries, so we don't have to calculate the index for the division and the modulus.

PR #3892, opened 2017-09-11
fix missing graphics label

### LINKED ISSUE #3769: Virtual Machine cannot power on due to xCAT Plugin Bug around graphics code
Opening on behalf of Zhibao Hu

```
[root@ems6 bin]# rpower essio61-vm on
Error: kvm plugin bug, pid 8711, process description: 'xcatd SSL: rpower to essi
o61-vm for root@localhost: kvm instance' with error 'Can't call method "setAttri
bute" on an undefined value at /opt/xcat/lib/perl/xCAT_plugin/kvm.pm line 1403.
' while trying to fulfill request for the following nodes: essio61-vm
```

So looking quickly at this code, I modified line 1403 to check if graphics is defined.    I think this is for VNC password... 
```
1398     my $parseddom = $parser->parse_string($xml);
1399     my ($graphics) = $parseddom->findnodes("//graphics");
1400     if ($confdata->{vm}->{$node}->[0]->{vidpassword}) {
1401         $graphics->setAttribute("passwd", $confdata->{vm}->{$node}->[0]->{vidpassword});
1402     } else {
1403         if (defined($graphics)) {
1404             $graphics->setAttribute("passwd", genpassword(20));
1405         }
1406     }
``` 

With this patch , we are able to get the VM to power up.  

However, what I don't understand is what causes the `//graphics` entry to be inserted into the XML data.  Any wouldn't we hit this in our testing? @hu-weihua ? There's nothing about `vidpassword` either in the node... 
```
[root@ems6 ~]# lsdef essio62-vm | grep vidpass
[root@ems6 ~]#
```

So maybe the better fix is to wrap the graphics check at line 1400 and cover both the if and else case.....

Assign to @bybai  to investigate and create a better fix.

bybai:
fix #3769 

If there is no `graphics` label, it affects `rpower` and `wcons` command. Based on my investigate, this fix is for `rpower` command, if there is no `graphics` label in kvm xml, `wcons` will not be broken, it will print out `$DISPLAY  not set`, so it is no need to change anything for `wcons` command.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/kvm.pm
+++ b/xCAT-server/lib/xcat/plugins/kvm.pm
@@ -1397,12 +1397,14 @@ sub makedom {
     }
     my $parseddom = $parser->parse_string($xml);
     my ($graphics) = $parseddom->findnodes("//graphics");
-    if ($confdata->{vm}->{$node}->[0]->{vidpassword}) {
-        $graphics->setAttribute("passwd", $confdata->{vm}->{$node}->[0]->{vidpassword});
-    } else {
-        $graphics->setAttribute("passwd", genpassword(20));
+    if (defined($graphics)) {
+        if ($confdata->{vm}->{$node}->[0]->{vidpassword}) {
+            $graphics->setAttribute("passwd", $confdata->{vm}->{$node}->[0]->{vidpassword});
+        } else {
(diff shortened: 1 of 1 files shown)


immarvin:
i am ok to merge

PR #3905, opened 2017-09-12
CORAL Reference Documentation, initial skeleton

whowutwut:
Attempt to create a reference documentation for our End to End steps standing a CORAL like environment


whowutwut:
I've commented out the skeleton function for now and it looks only like the following: 

image

On recommendation of @cxhong , we should merge this so a place is avail for us to put some short known issues and resolutions as they come up for the benefits of those doing deployment, and then I can work on the other sections on the side.

whowutwut:
@robin2008 @hu-weihua Please merge this so that the skeleton is avail and I'll add more content and think about the format in additional pull requests in the coming weeks.

Issue #3908, opened 2017-09-12
[OpenBMC] Cannot configure more than 1 node at a time


whowutwut:
@xuweibj Just noticed this, is there a technical reason for this? 

```
[root@briggs01 910.1736.20170828r-1736B]# rspconfig mid05tor12cn[09-10] ip=dhcp
Can not configure more than 1 nodes' ip at the same time
```

Can perform the actions 1 at a time...
```
[root@briggs01 910.1736.20170828r-1736B]# rspconfig mid05tor12cn09 ip=dhcp
mid05tor12cn09: BMC Setting IP to DHCP...
mid05tor12cn09: BMC reboot
[root@briggs01 910.1736.20170828r-1736B]# rspconfig mid05tor12cn10 ip=dhcp
mid05tor12cn10: BMC Setting IP to DHCP...
mid05tor12cn10: BMC reboot
[root@briggs01 910.1736.20170828r-1736B]#
```

xuweibj:
For ``rspconfig ip``, to avoid users set the same IP address for different nodes, added this error msg. I think for ``ip=dhcp`` it's not needed. I will fix this issue.

xuweibj:
Hi @whowutwut , could we close this issue?

whowutwut:
Let me attempt to verify and I will close out...

immarvin:
can this be closeD?

whowutwut:
Yes, let's close this.

PR #3895, opened 2017-09-11
BMC update image check and upload+activate rflash options

gurevichmark:
Small change to enable BMC update images to be correctly handled by `rflash -c` and `rflash -a` options.

Extract version number from the BMC MANIFEST update file using the same code as PHOR MANIFEST file.

PR #3909, opened 2017-09-13
expose rspconfig get options and rflash command

xuweibj:
Expose rspconfig ip/netmask/gateway/vlan (get), hostname, ip=dhcp.

#3908 
Only when set same IP address for more than 1 node, print error msg "Can not configure more than 1 nodes' ip at the same time".

Expose rflash command based on #3895 merged.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -581,10 +581,10 @@ sub parse_args {
 
                 my $nodes_num = @$noderange;
                 return ([ 1, "Invalid parameter for option $key" ]) unless ($value);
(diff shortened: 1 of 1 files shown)


whowutwut:
I like how we are exposing the rflash command, we need to do this for 2.13.7.  What about rspconfig, are we ready yet to expose this? I think Ratan was looking at implementing a way for us to modify an existing ipv4 obeject instead of creating a new one, that's in phase 5.1 (9/29) , should we hold off for that before we expose all the functions for rspconfig?

xuweibj:
Hi @whowutwut , in this pull request, just expose getting ip/netmask/vlan/gateway information. The command for setting is not exposed. @zet809 received the requirement that need to use rspconfig to read IP's information. So expose the command of getting.

PR #3915, opened 2017-09-14
Delete genesis base built check in xcatprobe discovery pre-check

xuweibj:
The newest xcat-genesis-base is not built in fedora environment, so delete this check.


zet809:
The genesis-base based on Fedora 26 didn't support p9, so I think it is OK to remove the checking. Thx, @xuweibj

Issue #3552, opened 2017-07-28
xCAT Logging is inconsistent between MN and SN


whowutwut:
The logs are inconsistent in where we are storing them in the MN and SN.  This is just an example and we should ensure that other logs are correctly placed. 

On the MN, the discovery logs appear in `/var/log/xcat/cluster.log` 

```
[root@boston01 xcat]# grep xcat.discovery /var/log/xcat/cluster.log  | more 
Jul 25 21:20:04 boston01 xcat[11494]: xcat.discovery.aaadiscovery: (70:e2:84:14:0a:71) Got a discovery request, attempting to discover the node...
Jul 25 21:20:05 boston01 xcat[11494]: xcat.discovery.blade: (70:e2:84:14:0a:71) Warning: Could not find any nodes using blade-based discovery
Jul 25 21:20:08 boston01 xcat[11494]: xcat.discovery.switch: (70:e2:84:14:0a:71) Found node: sn02
Jul 25 21:20:09 boston01 xcat[11494]: xcat.discovery.nodediscover: Removing discovered node definition: node-8335-gtb-100465a...
Jul 25 21:20:10 boston01 xcat[11494]: xcat.discovery.nodediscover: node-8335-gtb-100465a definition removed from xCAT
Jul 25 21:20:12 boston01 xcat[11494]: xcat.discovery.nodediscover: sn02 has been discovered
```

On SN,  they are in `/var/log/messages`  .  

```
[root@sn02 log]# grep xcat.discovery /var/log/messages
Jul 27 21:15:04 sn02 xcat[14929]: xcat.discovery.aaadiscovery: (70:e2:84:14:29:10) Got a discovery request, attempting to discover the node...
Jul 27 21:15:04 sn02 xcat[14929]: xcat.discovery.blade: (70:e2:84:14:29:10) Warning: Could not find any nodes using blade-based discovery
Jul 27 21:15:16 sn02 xcat[14929]: xcat.discovery.switch: (70:e2:84:14:29:10) Found node: mid05tor12cn11
Jul 27 21:15:17 sn02 xcat[14929]: xcat.discovery.nodediscover: Removing discovered node definition: node-70e28414290f...
Jul 27 21:15:17 sn02 xcat[14929]: xcat.discovery.nodediscover: node-70e28414290f definition removed from xCAT
```

whowutwut:
@immarvin @zet809 FYI, is there something we can do here..... 

[...]

PR #3921, opened 2017-09-14
Support configuration of local syslog files on the Service Node

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3552

1)ship logrotate and log filter rule files in xCATsn;
2)refine syslog script to enable log filter and rotate rules

### DIFF
--- a/xCAT/postscripts/syslog
+++ b/xCAT/postscripts/syslog
@@ -243,6 +243,14 @@ config_rsyslog_V8()
                 s/#module(load="imtcp")/module(load="imtcp")/;
                 s/#input(type="imtcp" port="514")/input(type="imtcp" port="514")/' $conf_file
     fi
+    
+    if [ $isSN -eq 1 ] ; then
+        [ -d "/etc/xcat/rsyslog.conf" ] && cp -a /etc/xcat/rsyslog.conf/* /etc/rsyslog.d/
+        if [ -d "/etc/xcat/logrotate.conf" ];then 
+            mkdir -p "/etc/logrotate.d/"
+            cp -a /etc/xcat/logrotate.conf/* /etc/logrotate.d/
+        fi
+    fi
 
     # Mark the end of xCAT section
     echo "# $xCATSettingsEND" >> $conf_file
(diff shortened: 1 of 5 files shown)


whowutwut:
@immarvin  FYI I changed the title to be more descriptive of the changes contained in the PR

xCATsn/xCATsn.spec
```diff
@@ -93,6 +93,8 @@ tar -xf license.tar
 %install
 %ifos linux
 mkdir -p $RPM_BUILD_ROOT/etc/xcat/conf.orig
+mkdir -p $RPM_BUILD_ROOT/etc/xcat/rsyslog.conf.orig
+mkdir -p $RPM_BUILD_ROOT/etc/xcat/logrotate.conf.orig
```
robin2008: suggest to use rsyslog.conf.d and logrotate.conf.d

xCAT/postscripts/syslog
```diff
@@ -261,6 +269,13 @@ config_rsyslog_V8()
 
         [ -f "$conf_file" ] && sed -i '/#\$ModLoad \+imudp\|imtcp\|imudp.so\|imtcp.so/s/^#//;
                        /#\$InputTCPServerRun\|UDPServerRun.*/s/^#//' $conf_file
+
+        [ -f "/etc/rsyslog.d/xcat-cluster.conf" ] && rm -rf "/etc/rsyslog.d/xcat-cluster.conf" ;
```
robin2008: not suggest to use `rm -rf` here for a plain text file.

xCATsn/xCATsn.spec
```diff
@@ -102,6 +104,8 @@ cp %{SOURCE1} $RPM_BUILD_ROOT/etc/httpd/conf.d/xcat.conf
 cp %{SOURCE3} $RPM_BUILD_ROOT/etc/xCATSN
 cp %{SOURCE1} $RPM_BUILD_ROOT/etc/xcat/conf.orig/xcat.conf.apach22
 cp %{SOURCE6} $RPM_BUILD_ROOT/etc/xcat/conf.orig/xcat.conf.apach24
+\cp -rf  etc/rsyslog.d/* $RPM_BUILD_ROOT/etc/xcat/rsyslog.conf.orig/
```
samveen: Maybe use `cp -a` instead of `cp -rf` ?

xCAT/postscripts/syslog
```diff
@@ -243,6 +243,14 @@ config_rsyslog_V8()
                 s/#module(load="imtcp")/module(load="imtcp")/;
                 s/#input(type="imtcp" port="514")/input(type="imtcp" port="514")/' $conf_file
     fi
+    
+    if [ $isSN -eq 1 ] ; then
+        [ -d "/etc/xcat/rsyslog.conf.orig" ] && \cp -rf /etc/xcat/rsyslog.conf.orig/* /etc/rsyslog.d/
+        if [ -d "/etc/xcat/logrotate.conf.orig" ];then 
+            mkdir -p "/etc/logrotate.d/"
+            \cp -rf /etc/xcat/logrotate.conf.orig/* /etc/logrotate.d/
```
robin2008: Maybe we need to specify which logrotate conf file be copied, instead of using '*' as we might have the case to cover non-rsyslog service in future.

robin2008: And do we need to reload logrotated here,  or it is already covered in existing codes.

immarvin:
hi @robin2008 @samveen , thanks for the comments, code updated

Issue #3910, opened 2017-09-13
The default configuration of rsyslog may not provide the log data integrity for xcat log locally


chenglch:
I encounter this problem on redhat7.3, 
Version 2.13.7 (git commit 436f69e85fefd2325738eac0770c801fc51e31bf)

The method from `xCAT::MsgUtils->message("S"...)` subroutine send the log message to the syslog, but some of log may be lost as the option of `$OmitLocalLogging on` from the configuration of `rsyslog`.

Some description for this option
```
# Turn off message reception via local log socket;
# local messages are retrieved through imjournal now.
```
After changing `$OmitLocalLogging` from `on` to `off`, all of the log can be kept.

neo954:
The default `rsyslog.conf` has the following configurations.
```
$ModLoad imuxsock # provides support for local system logging (e.g. via logger command)
$ModLoad imjournal # provides access to the systemd journal
$ModLoad imudp
$UDPServerRun 514
$ModLoad imtcp
$InputTCPServerRun 514
$WorkDirectory /var/lib/rsyslog
$ActionFileDefaultTemplate RSYSLOG_TraditionalFileFormat
$IncludeConfig /etc/rsyslog.d/*.conf
$OmitLocalLogging on
$IMJournalStateFile imjournal.state
authpriv.*                                              /var/log/secure
mail.*                                                  -/var/log/maillog
cron.*                                                  /var/log/cron
*.emerg                                                 :omusrmsg:*
uucp,news.crit                                          /var/log/spooler
local7.*                                                /var/log/boot.log
```

Due to [1], [2]. module `imjournal` reads systemd socket "SystemLogSocketName", which is `/run/systemd/journal/syslog` on RHEL7 by default, and modules `imuxsock` reads `/dev/log`. That is why `$OmitLocalLogging on` is used.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1459633
[2] http://www.rsyslog.com/doc/v8-stable/configuration/modules/imuxsock.html

After set `$OmitLocalLogging off` on RHEL7, local log messages will be logged twice.

neo954:
In [3], it mentioned, the performance of module `imjournal` may be notably slower than when using `imuxsock`.

[3] http://www.rsyslog.com/doc/v8-stable/configuration/modules/imjournal.html

Thus, I guess, when use module `imjournal`, some logs were missing due to poor performance.

neo954:
[...]

PR #3922, opened 2017-09-14
Add description for the issue that may lose log

chenglch:
This commit add descritpion for `dbtracelevel` entry in site table
to help the user find the detail solution if the log is lost due
to the issue from rsyslog side.

partial-issue: #3910


neo954:
I will merge this one.

PR #3925, opened 2017-09-14
Handle TIMEOUT error message for change pdu ip address

cxhong:
Able to test rspconfig command to change ip address on Coral PDU.  
````
# rspconfig coral-pdu ip=9.114.82.124 netmask=255.255.0.0
Failed to run /dev/shm/bin/PduManager -i 9.114.82.124 -n 255.255.0.0, error=1: TIMEOUT
````
Actually ip address is changed.  The error message logged because  the xdsh command was hung after ip changes.   Add code to check if ip address is pingabled.

```
# rspconfig coral-pdu ip=9.114.82.95 netmask=255.255.0.0
9.114.82.95 is reachable
/dev/shm/bin/PduManager -i 9.114.82.95 -n 255.255.0.0 ran successfully
````


zet809:
Looks good, I agree to merge.

PR #3937, opened 2017-09-15
Fix the documentation from #3891

mattaezell:
Add more examples of using the regex helpers.  Previous commit modified a man page directly instead of the db2man script.


mattaezell:
@whowutwut @chenglch Would one of you be willing to review this?  The current version of the doc (http://xcat-docs.readthedocs.io/en/stable/guides/admin-guides/references/man5/xcatdb.5.html) will get overwritten next time someone generates the man pages.

PR #3950, opened 2017-09-18
Treat 5104-22C as a ESS Boston node for rflash command

gurevichmark:
Add another Chassis ID for Boston node: 5104-22C


whowutwut:
Cross reference with initial issue that was opened Internally. https://github.ibm.com/vhu/c910env/issues/166

Issue #3951, opened 2017-09-18
[OpenBMC] discovery leaves around "node-MAC" if "node-MTM-SN" exists as well


whowutwut:
In a error recovery scenario, we have a bmc node definition created using the MTM-SN format `node-MTM-SN`.  Then after recovery of the BMC to repair MAC, it gets a different MAC address and doing `bmcdiscover` causes it to create a discovered node definition in the format of `node-mac`.   

I left both node definitions around to see what would happen.... 

After xcat.discovery is run , it makes a match: 

```
[root@briggs01 xcat]# grep xcat.discovery cluster.log
Sep 18 15:54:46 briggs01 xcat[56208]: xcat.discovery.aaadiscovery: (70:e2:84:14:28:7d) Got a discovery request, attempting to discover the node...
Sep 18 15:54:46 briggs01 xcat[56208]: xcat.discovery.blade: (70:e2:84:14:28:7d) Warning: Could not find any nodes using blade-based discovery
Sep 18 15:54:49 briggs01 xcat[56208]: xcat.discovery.switch: (70:e2:84:14:28:7d) Found node: mid05tor12cn13
Sep 18 15:54:49 briggs01 xcat[56208]: xcat.discovery.nodediscover: Removing discovered node definition: node-8335-gtc-1318daa...
Sep 18 15:54:50 briggs01 xcat[56208]: xcat.discovery.nodediscover: node-8335-gtc-1318daa definition removed from xCAT
Sep 18 15:54:50 briggs01 xcat[56208]: xcat.discovery.nodediscover: mid05tor12cn13 has been discovered
Sep 18 15:54:50 briggs01 xcat[56208]: xcat.discovery.zzzdiscovery: (70:e2:84:14:28:7d) Successfully discovered the node using switch discovery method.
```

So it actually removed the `node-MTM-SN` and does not remove the `node-MAC`.  So in the discovery code, we should check for both and clean up the xCAT database accordingly...

immarvin:
hi @zet809 ,would you please take a look at this?

whowutwut:
or @xuweibj Are you familiar with this code?  If @zet809 is busy, could you take a look at this?

xuweibj:
In cluster.log
```
ep 19 04:20:10 briggs01 xcat[113616]: xcat.discovery.nodediscover: Removing discovered node definition: node-8335-gtc-1318c4a,node-70e28414280d...
[...]

PR #3962, opened 2017-09-19
Fix issue 3951, rm all bmc nodes when discovered

xuweibj:
#3951 

Add all bmc nodes in ARRAY @bmc_nodes, rm all when node is discovered.
If BMC that doesn't support in-band configuration, show warning msg that rm bmc nodes manually.


zet809:
Looks good to me, I agree to merge. Thx! @xuweibj

Issue #3972, opened 2017-09-20
[fvt] rflash does not work against 8335-GTA physical machine


hu-weihua:
XCAT version: 2.13.7 (git commit df8c223afba0daaa53bcbcaf571a8fbc5b823fbf, built Tue Sep 19 07:15:37 EDT 2017)

``rflash`` has problem to resolve command line and to do upgrade job against 8``335-GTA`` physical machine

```
[root@c910f03c05k25 firmware]# ll
total 688260
-rw-r--r-- 1 root root 100663905 Mar 27 13:34 8335GTB_820.1642.20170327n.hpm
-rw-r--r-- 1 root root 100663905 Mar 27 13:34 8335GTB_820.1642.20170327n_pass1.hpm
-rw-r--r-- 1 root root  67109473 Mar 27 13:34 8335GTB_820.1642.20170327n_update.hpm
-rw-r--r-- 1 root root  67109473 Mar 27 13:34 8335GTB_820.1642.20170327n_update_pass1.hpm
-rw-r--r-- 1 root root 100663905 Mar 27 12:58 8335_820.1642.20170327n.hpm
-rw-r--r-- 1 root root    110137 Sep 20 04:33 8335_820.1642.20170327n.html
-rw-r--r-- 1 root root  67109473 Mar 27 12:58 8335_820.1642.20170327n_update.hpm
......

[root@c910f03c05k25 firmware]# pwd
/tmp/firmware

[root@c910f03c05k25 firmware]# rflash fs4 8335_820.1642.20170327n_update.hpm
Error: fs4: Data directory must be specified.

[root@c910f03c05k25 firmware]# rflash fs4 /tmp/firmware/8335_820.1642.20170327n_update.hpm
Error: fs4: Data directory must be specified.

[root@c910f03c05k25 firmware]#   rflash fs4 -d /tmp/firmware/
Error: fs4: The image file doesn't match this machine

[root@c910f03c05k25 firmware]# rflash fs4 /tmp/firmware/8335_820.1642.20170327n_update.hpm -d /tmp/firmware
Error: fs4: Unable to determine firmware version currently installed. Verify that "/opt/xcat/bin/ipmitool-xcat -H 50.3.25.1 -I lanplus -U ADMIN -P admin fru print 47 | grep OP8_v" command returns a version.

[...]

PR #3973, opened 2017-09-20
Do not check the directory if pUpdate is not used

chenglch:
fix-issue: #3972

With this patch, the directory issue maybe solved, but `firestone_update_version` may be wrong as the digit is `0` which could not help us to determine the version. 
```
fs4: System Firmware Product Version: IBM-firestone-ibm-OP8_PFD_v1.6_0.11
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -1970,24 +1970,6 @@ sub do_firmware_update {
                 }
             }
         }
-
-        if (defined $directory_name) {
-            unless (File::Spec->file_name_is_absolute($directory_name)) {
-               # Directory name was passed in as relative path, prepend current working dir
-               $directory_name = xCAT::Utils->full_path($directory_name, $::cwd);
-            }
-            # directory was passed in, verify it is valid
-            if (-d $directory_name) {
-               # Passed in directory name exists
-               $pUpdate_directory = $directory_name;
-            }
-            else {
-                $exit_with_error_func->($sessdata->{node}, $callback, "Can not access data directory $directory_name");
-            }
-        }
-        else {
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -1970,37 +1970,31 @@ sub do_firmware_update {
...
+                # Passed in directory name exists
+                $pUpdate_directory = $directory_name;
             }
             else {
                 $exit_with_error_func->($sessdata->{node}, $callback, "Can not access data directory $directory_name");
             }
         }
         else {
-            $exit_with_error_func->($sessdata->{node}, $callback, "Data directory must be specified.");
+            $exit_with_error_func->($sessdata->{node}, $callback, "Directory name is required to update Boston or Briggs machines.");
```
gurevichmark: Suggest not to use "Boston" or "Briggs". Better to use "IBM Power S822LC for Big Data"

chenglch: Done

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -2148,7 +2142,7 @@ sub do_firmware_update {
                 "Running ipmitool command $cmd failed: $output");
         }
         # Check what firmware version is currently running on the machine
-        if ($output =~ /OP8_v\d\.\d+_(\d+)\.\d+/) {
+        if ($output =~ /OP8_.*v\d\.\d+_(\d+)\.\d+/) {
```
chenglch: As the firmware product version is OP8_PFD_v1.6_0.11 on fs4, the `frs_version` is always 0. I'm not sure if it is correct to add `.*` to match the `PFD` characters. For some quick test on `fs4`, `ipmitool-xcat hpm upgreade ...` command return failure, no idea about this.

PR #3980, opened 2017-09-21
Use unified attribute in rpower_wrongpasswd test case

hu-weihua:
There are unified attribute ``$$bmcpasswd $$bmcusername`` in xcattest configuration file, do not need to add new attribute.


bybai:
I am ok to merge it.

Issue #3979, opened 2017-09-21
fatal error in openbmc.pm terminates `nodeset <node> next` processing when node status switching from "installing" to "booting"


immarvin:
steps to reproduce:
1. run xcatd with `xcatd -f` on mn
2. on cn:
````
cd /xcatpost
./updateflag.awk <mn ip> 3002
````
3. from the MN console same with step #1:
````
...
monitorctrl::setNodeStatusAttributes called
HTTP::Async object destroyed but still in use at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 460.
HTTP::Async INTERNAL ERROR: 'id_opts' not empty at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 460.
Use of uninitialized value in hash element at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 477.
Use of uninitialized value in hash element at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 477.
Use of uninitialized value in hash element at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 1179.
Use of uninitialized value in subroutine entry at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 1179.
Use of uninitialized value in subroutine entry at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 1179.
Can't use string ("") as a subroutine ref while "strict refs" in use at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 1179.

Died at /opt/xcat/sbin/xcatd line 2010.

xcatd: possible BUG encountered by xCAT install monitor service: Died at /opt/xcat/sbin/xcatd line 2010.
````

which prevents the following process of changing petitboot configuration

**Comment by zet809** [comment]
Hi, @immarvin , this issue is fixed with PR #3983 , the test looks like this, so will you pls close this issue? Thx!
```
# cat /tftpboot/petitboot/mid05tor12cn05
#install rhels7.4-snap2-ppc64le-compute
default xCAT
label xCAT
	kernel http://172.12.253.27:80/tftpboot/xcat/osimage/rhels7.4-snap2-ppc64le-install-compute/vmlinuz
	initrd http://172.12.253.27:80/tftpboot/xcat/osimage/rhels7.4-snap2-ppc64le-install-compute/initrd.img
[...]

PR #3983, opened 2017-09-21
Fix issue 3979, do not set next status if command is getopenbmccons

xuweibj:
#3979


zet809:

PR #3990, opened 2017-09-21
Setting BMC hostname message fix

gurevichmark:
With OpenBMC 1738 build, reboot of BMC is not required to change hostname


whowutwut:
Will merge after 2.13.7 releases

whowutwut:
Actually, let me merge this because it's good to not say we need to reboot because the current FW applies the hostname immediately.

Issue #3993, opened 2017-09-21
[OpenBMC] seeing strange MTM and SN values from OpenBMC


whowutwut:
Something new with the VPD is causing this to appear from the REST API. .

```
[root@briggs01 ~]# bmcdiscover --range 172.12.253.100-176
172.12.253.168,8335-GTC........,1318C4A.........,,,mp,bmc,,
[root@briggs01 ~]# bmcdiscover --range 172.12.253.100-176 -w
Writing node-8335-gtc---------1318c4a--------- (172.12.253.168,8335-GTC........,1318C4A.........,,,mp,bmc,,) to database...
[root@briggs01 ~]# 
```

We need to fix 
* dodiscovery
* bmcdiscover

to remove those dots and follow up with FW team.

zet809:
With fix in #3994 , the output looks like this, so I close this issue. Thx @whowutwut and @xuweibj 
```
# bmcdiscover --range  172.12.139.105
172.12.139.105,8335-GTC,1318C4A,,,mp,bmc,,
# lsdef mid05tor12cn05 -i mtm,serial
Object name: mid05tor12cn05
    mtm=8335-GTC
    serial=1318C4A
```

whowutwut:
This is good for xCAT, but petitboot still shows... 
```
 Petitboot (v1.5.1-p36336db)                  8335-GTC........ 1318C4A.........
 ──────────────────────────────────────────────────────────────────────────────
  [Disk: sda2 / d3b8847a-d933-4611-b06d-21f753633851]
    Red Hat Enterprise Linux Server (0-rescue-66afe22050bd41e187b8c8124d721158)
 *  Red Hat Enterprise Linux Server (4.11.0-30.el7a.ppc64le) 7.4 (Pegas)       
  [Network: enP5p1s0f0 / 70:e2:84:14:28:0e]
    xCAT Genesis shell

  System information
  System configuration
  System status log
  Language
  Rescan devices
  Retrieve config from URL    
  Exit to shell






 ──────────────────────────────────────────────────────────────────────────────
 Enter=accept, e=edit, n=new, x=exit, l=language, g=log, h=help```

So now we are hiding this potential problem. I wonder how it got to this in the first place...   It's not firmware related i don't think..

PR #3994, opened 2017-09-22
Delete all '.' in mtm and serial string for OpenBMC

xuweibj:
#3993
For current version of firmware, mtm and serial like this:
```
# rinv mid05tor12cn05 model
mid05tor12cn05: SYSTEM Model : 8335-GTC........
# rinv mid05tor12cn05 serial
mid05tor12cn05: SYSTEM SerialNumber : 1318C4A.........
```

Before modify:
```
# bmcdiscover --range 172.12.139.105
172.12.139.105,8335-GTC........,1318C4A.........,,,mp,bmc,,
```
After modify:
```
# bmcdiscover --range 172.12.139.105
172.12.139.105,8335-GTC,1318C4A,,,mp,bmc,,
```

### DIFF
--- a/xCAT-genesis-scripts/bin/dodiscovery
+++ b/xCAT-genesis-scripts/bin/dodiscovery
@@ -159,10 +159,12 @@ fi
 
 flag_mtm=`echo '$MTM' | sed 's/0//g'`
 if [ $flag_mtm ] && [ "$MTM" != "unknown" ]; then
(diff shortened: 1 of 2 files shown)


zet809:

whowutwut:
This is good.  I don't think it's something in the firmware but something burned into the VPD. the firmware passes the value through. Regardless this will help avoid this output in xcat! 👍🏼 Will try to understand why this happens...

Issue #1270, opened 2016-06-06
Ran discovery without setting host information correctly


markprez:
Switch based discovery, I successfully setup a garrison node. 
When I ran the nodeset command and got the error:
[root@c712ems4 TRY3]# nodeset c712f7n04-pvt  osimage=test
c712f7n04-pvt: install rhels7.2-ppc64le-compute
Error: The c712f4n04-pvt can not be resolved

I found I did not have the /etc/hosts entry for the OS' private network. Added /etc/hosts entry and makedns, makedhcp, nodeset, reboot.

I then found the node booting into genesis and found the lease entry:
host c712f7n04-pvt-noip70e284140a5f {
  dynamic;
  hardware ethernet 70:e2:84:14:0a:5f;
        supersede server.allow-booting = 00;
}

Found the mac address from discovery had _NOIP_
mac=70:e2:84:14:0a:5f!_NOIP_

I removed the "!_NOIP_" from the mac and reran makedhcp/nodeset, the node installed.

I realize the problem was the result of an admin error, however how to recover was not clear. Is fixing resolution and changing the mac the best solution?

daniceexi:
@markprez if you know the correct mac address, remove the ``!NOIP`` from the mac is OK to fix this.
Another way is to fix the configuration and do the discovery again.

xCAT team also is working on to tool which can be used to check the validity of predefined node.

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/3995

PR #3995, opened 2017-09-22
enhance nodediscovery process: don't write "NOIP" if the node can not be resloved to an IP

zet809:
The fix for #1270

enhance_nodediscover_process, don't write mac table and don't generate dhcp lease entry

In the client, it will be in the sending findme request loop, after admin run "makehosts <pre-defined-node>", the hardware discovery can update MAC address successfully after send out the next findme request.

The /var/log/xcat/cluster.log will show this right now:
```
Nov 15 11:16:34 briggs01 xcat[47212]: xcat.discovery.aaadiscovery: (70:e2:84:14:28:0e) Got a discovery request, attempting to discover the node...
Nov 15 11:16:34 briggs01 xcat[47212]: DEBUG xcatd: call plugin <blade> to handle command <findme>
Nov 15 11:16:34 briggs01 xcat[47212]: xcat.discovery.blade: (70:e2:84:14:28:0e) Warning: Could not find any nodes using blade-based discovery
Nov 15 11:16:34 briggs01 xcat[47212]: DEBUG xcatd: call plugin <profilednodes> to handle command <findme>
Nov 15 11:16:34 briggs01 xcat[47212]: DEBUG xcatd: call plugin <seqdiscovery> to handle command <findme>
Nov 15 11:16:34 briggs01 xcat[47212]: DEBUG xcatd: call plugin <switch> to handle command <findme>
Nov 15 11:16:34 briggs01 xcat[47212]: xcat.discovery.switch: (70:e2:84:14:28:0e) Found node: mid05tor12cn05
Nov 15 11:16:34 briggs01 xcat[47212]: DEBUG xcatd: call plugin <nodediscover> to handle command <discovered>
Nov 15 11:16:35 briggs01 xcat[47212]: xcat.discovery.nodediscover: Can not resolve IP for the matching node:mid05tor12cn05. Make sure "makehosts" and "makedns" have been run for mid05tor12cn05.
Nov 15 11:16:35 briggs01 xcat[47212]: xcat.discovery.zzzdiscovery: (70:e2:84:14:28:0e) Failed for node discovery: Can not resolve IP for the matching node:mid05tor12cn05.
Nov 15 11:16:35 briggs01 xcat[47212]: xcat.discovery.zzzdiscovery: Notify 172.12.139.5 that its findme request has been processed
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/nodediscover.pm
+++ b/xCAT-server/lib/xcat/plugins/nodediscover.pm
@@ -274,6 +274,7 @@ sub process_request {
         my @hostnames_to_update = ();
         my %bydriverindex;
         my $forcenic = 0; #-1 is force skip, 0 is use default behavior, 1 is force to be declared even if hosttag is skipped to do so
+        my $macs_withip = 0;
         foreach (@{ $request->{mac} }) {
             @ifinfo = split /\|/;
 
@@ -308,6 +309,7 @@ sub process_request {
             if ($ifinfo[3]) {
                 (my $ip, my $netbits) = split /\//, $ifinfo[3];
                 if ($ip =~ /\d+\.\d+\.\d+\.\d+/) {
+                    $macs_withip++;
                     my $ipn = unpack("N", inet_aton($ip));
                     my $mask = 2**$netbits - 1 << (32 - $netbits);
                     my $netn = inet_ntoa(pack("N", $ipn & $mask));
@@ -348,6 +350,9 @@ sub process_request {
             } else {
                 if ($forcenic == 1) { $macstring .= $currmac . "|"; }
             }
+        }
+        if ($macs_withip == 1 and $macstring =~ /\*NOIP\*/) {
+            return;
         }
         $macstring =~ s/\|\z//;
         $mactab->setNodeAttribs($node, { mac => $macstring });
--- a/xCAT-server/lib/xcat/plugins/nodediscover.pm
+++ b/xCAT-server/lib/xcat/plugins/nodediscover.pm
@@ -267,6 +267,12 @@ sub process_request {
 
     my $macstring = "";
     if (defined($request->{mac})) {
+        if (!inet_aton($node)) {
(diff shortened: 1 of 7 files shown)


whowutwut:
@zet809 Are you planning on refining the last message since the discovery has failed? OR that will come in another PR?  

Should we be adding a message in that xcat.discovery logs based on the  changes here? I'm not following the changes very well, the title says don't write NOIP if it can't resolve IP address. so what would the MAC address show if it can't resolve IP?

zet809:
hi, @whowutwut , yes, you are right that the discovery is failed in the end. I need to modify it.

For the original situation, if the node can not be resolved to an IP address, there will be a "NOIP" appended to the mac address and then set "<mac_address!*NOIP*>" to the mac attribute of the pre-defined node. Then, when the CN restart network server, it won't be able to get a dynamic IP any more since the "*NOIP*" will cause a "allow-booting = 00" entry in the dhcp lease file.
So, in this fix, we will report the error and don't set "NOIP" entry in order not to cause the CN failed to get a dynamic again.

whowutwut:
@zet809  Ok, so the NOIP was actually helping us debug before, now with this change you're saying it caused issues unless we remove the NOIP from the mac string? So to avoid that, we just do not add it?

xCAT-server/lib/xcat/plugins/nodediscover.pm
```diff
@@ -349,6 +351,9 @@ sub process_request {
                 if ($forcenic == 1) { $macstring .= $currmac . "|"; }
             }
         }
+        if ($macs_withip == 1 and $macstring =~ /\*NOIP\*/) {
+            return;
```
robin2008: Why not just remove `*NOIP*` here, but continue with the other parts in this function.
I see lots of logic there, for example, makedhcp, updateswitch, etc.

zet809:
Hi, @whowutwut and @robin2008 , I have update the PR, will you pls help to take a look at again? Thx!

immarvin:
i am ok with the change

Issue #3977, opened 2017-09-20
switch discovery for ib switches is NOT scalable if we use our own community string.


ralphbellofatto:
we are doing the following process:
following this set of directions:
https://github.ibm.com/vhu/c910env/blob/master/notes/coral/end2end/edge_switch.md

template and groups are different here...
```
mkdef c699ibs[1-2]_[00] groups="switch,ib_switch" 
```
if we have not already done so add to group ib_switch
```chdef -t group nodetype=switch ib_switch password=admin     username=admin```

Then the individual discover gorp.
```
chdef c699ibs1_00 switch=c699eag01 switchport=16 ip=10.5.100.1
chdef c699ibs2_00 switch=c699eag01 switchport=18 ip=10.5.100.2
makehosts -a
```

then do a non write trial of the switch discover:
```
switchdiscover --range 10.5.254.1-250
```

no workee:
```
[root@c699mgt00 ~]# switchdiscover --range 10.5.254.1-250
Discovering switches using nmap for 10.5.254.1-250. It may take long time...
 No switch found                                                            
```

using nmap -sn we got:
```
[root@c699mgt00 ~]# nmap -sn 10.5.254.1-250

Starting Nmap 6.40 ( http://nmap.org ) at 2017-09-20 16:03 EDT
Nmap scan report for 10.5.254.1
Host is up (0.00023s latency).
MAC Address: 24:8A:07:B1:C1:CC (Unknown)
Nmap scan report for 10.5.254.2
Host is up (0.00025s latency).
MAC Address: 24:8A:07:B1:C1:5C (Unknown)
Nmap scan report for 10.5.254.5
Host is up (0.0013s latency).
MAC Address: A4:BF:01:36:38:C3 (Unknown)
Nmap scan report for 10.5.254.7
Host is up (0.00023s latency).
MAC Address: 24:8A:07:20:B5:EE (Unknown)
Nmap done: 250 IP addresses (4 hosts up) scanned in 2.06 seconds
```

10.5.254.1 and 10.5.254.2 are the switches we defined...

**Comment by ralphbellofatto** [comment]
doing this to debug:
```
switchdiscover --range 10.5.254.1-2 -s snmp -V
[c699mgt00]: Discovering switch using snmpwalk for 10.5.254.1-2 .....
[...]

PR #3818, opened 2017-08-30
Support User defined community string for enable snmp on Cumulus Switches

cxhong:
For Issue #3614.  
This PR will support User defined community string from following way:
1) if switch is specified in switches table, and switches.snmpversion != "3", take switches.password if specified as the community; otherwise goto step 2
2) take site.snmpc as the community if specified; otherwise goto step 3
3) take "public" as the community string

this PR also support to config SNMPv3 if user defined  authentication and encryption attributes in the switches table.

PR #4001, opened 2017-09-22
Add -c option for switchdiscover command to pass in community string

cxhong:
For issue #3977,

The default community string for switch and pdu will be **public**.  if user define different community other than default on those device already,  need to pass in the user defined community to switchdiscover command to be able to discover the device.  otherwise, snmpwalk command will failed and nothing will be discovered.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/switchdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/switchdiscover.pm
@@ -25,7 +25,7 @@ use xCAT::data::switchinfo;
 
 #global variables for this module
 my $device;
-my $community;
+my $community="public";
 my %globalopt;
 my @filternodes;
 my @iprange;
@@ -144,7 +144,7 @@ sub parse_args {
     # Process command-line flags
     #############################################
     if (!GetOptions( \%opt,
-            qw(h|help V|verbose v|version x z w r n range=s s=s setup pdu))) {
+            qw(h|help V|verbose v|version x z w r n range=s s=s c=s setup pdu))) {
         return( usage() );
     }
 
@@ -251,6 +251,14 @@ sub parse_args {
     if ( exists( $opt{n} )) {
         $globalopt{n} = 1;
     }
+
+    #########################################################
+    # Accept the community string from user
+    #########################################################
+    if ( exists( $opt{c} )) {
+        $community=$opt{c};
+    }
+
 
     #########################################################
     # setup discover switch
@@ -892,13 +900,6 @@ sub snmp_scan {
     }
     my @lines = split /\n/, $result;
   
-    #set community string for switch
-    $community = "public";
-    my @snmpcs = xCAT::TableUtils->get_site_attribute("snmpc");
-    my $tmp    = $snmpcs[0];
-    if (defined($tmp)) { $community = $tmp }
-
-
     foreach my $line (@lines) {
         my @array = split / /, $line;
         if ($line =~ /\b(\d{1,3}(?:\.\d{1,3}){3})\b/)
--- a/xCAT-server/lib/xcat/plugins/switchdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/switchdiscover.pm
@@ -21,11 +21,12 @@ no strict;
 use Data::Dumper;
 use Socket;
 use Expect;
+use SNMP;
 use xCAT::data::switchinfo;
 
 #global variables for this module
 my $device;
-my $community="public";
+my $community;
 my %globalopt;
 my @filternodes;
 my @iprange;
@@ -258,7 +259,6 @@ sub parse_args {
     if ( exists( $opt{c} )) {
         $community=$opt{c};
     }
-
 
     #########################################################
     # setup discover switch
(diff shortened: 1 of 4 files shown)


mattaezell:
I'm not a big fan of entering passwords on the command line.  They end up in the history file.  What if the choice was between `default` or defined in the site/switches table?

cxhong:
This is for switchdiscover command to find the switches. Each cluster may have different kind of switches, some switch have default community string , but some switches are reused, and community string is user defined.  During the discover phase,  we can't use community string define in the switch table because at that time we don't know which switch yet.  If user knows some of switches have community string other than default, they can use -c options to find it.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -892,13 +900,6 @@ sub snmp_scan {
     }
     my @lines = split /\n/, $result;
   
-    #set community string for switch
```
immarvin: I have a little concern here, the modification here changed the default behavior. We might need to evaluate whether this will affect the user behavior

immarvin:
manpage needs to be updated

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -253,6 +253,14 @@ sub parse_args {
     }
 
     #########################################################
+    # Accept the community string from user
+    #########################################################
+    if ( exists( $opt{c} )) {
+        $community=$opt{c};
```
immarvin: besides the community string to access the switch via snmp, will the `$community `  variable here be used to program the community in the switch on `--setup`?

mattaezell:
@cxhong Understood.  What I was trying to propose is avoiding providing the community string on the command line (so it doesn't get shoved in the shell history).  By default, it would use `public`; but if `-c` is specified, it would look up the string in the database.  Or even better, maybe it could try `public` and automatically fall back to the password in the site table if `public` fails (no extra command line option needed).

immarvin:
hi @mattaezell , in PR https://github.com/xcat2/xcat-core/pull/3818, the site.snmpc might be used to program the community inside the switch, if it will also used as the initial snmp community on ``switchdiscover`` , will the two usage conflict in some scenario?

cxhong:
well, they will work if only two community string (one is default and other one is define in the site table) for the switches on this cluster.
maybe we can change snmpc attribute to accept multiple value.

immarvin:
hi @mattaezell , since on `switchdiscover -c `, the community string is just a initial value which will be later modified to value defined in the DB(in https://github.com/xcat2/xcat-core/pull/3818) on `--setup` or `enablesnmp`, i think it might not be a big deal to display the plain "-c <community string>" text in bash history

immarvin:
hi @cxhong , "change snmpc attribute to accept multiple value." i think this will be quite confusing for user

cxhong:
PR #3818 is only for cumulus switch.  xCAT needs to deal with other switches too,  like Mellanox Ethernet, Mellanox IB, BNT....

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: I think we should.   currently, switch table can be defined  snmpv1/v2 with community  or snmpv3.   --setup will set up snmpv3, but if snmpv3 is set up, we will not use community string any more for any snmp communication.

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: the comment maybe mislead here.   I just use community to communicate with switch, didn't set it to the switch.

immarvin:
hi @cxhong , on my last 2 comments, I think there might be some communication issue between us. my thought on this is:

1. the logic to determine the community string used in `switchdiscover -s snmp` should better be:
 **if `-c <string>` is specified, use <string>; otherwise, use  `site.snmpc` if defined;otherwise, use "public"**

2. **the community string specified with `-c` should only be used in `switchdiscover -s snmp`, it **should not** be programmed into the switch. **, seems you opened a ticket  https://github.com/xcat2/xcat-core/issues/4002 to trace the task to implement the same function on non-cumulus switches with PR https://github.com/xcat2/xcat-core/pull/3818

cxhong:
@immarvin  for the first comment, I implement as we discussed in the team meeting.  I can make changes as you suggested.
for the second comment,  not sure if we need to do now.   we took care cumulus switch in the PR#3818, for Mellanox switch, there is rspconfig command to set community string already.  we can talk about in the issue #4002 for other switches

immarvin:
thanks @cxhong , okay,  I suggest this PR to cover the logic in my first comment, we'd better keep the logic for `site.snmpc` to keep backward compatibility if no "-c" specified.

for https://github.com/xcat2/xcat-core/issues/4002, since rspconfig has covered this, what is this ticket created for?

cxhong:
@immarvin , code modified, can u review it again?

xCAT-server/lib/xcat/plugins/switchdiscover.pm
```diff
@@ -954,20 +955,47 @@ sub get_snmpvendorinfo {
     my $request = shift;
     my $ip  = shift;
     my $snmpwalk_vendor;
```
immarvin: `snmpwalk_vendor` is defined but not used?

immarvin:
hi @cxhong , please take a look at my last comment. I do not have any other comment, please ask @whowutwut to merge it once you process the last comment, thanks

xCAT-server/lib/xcat/plugins/switchdiscover.pm
cxhong: It used to return undefined value if in the error case.

whowutwut:
Merging on behalf of @immarvin.  Out of office on holiday.

PR #3982, opened 2017-09-21
modify rsetboot url based on new interface for 1738

xuweibj:
Based on issue https://github.com/openbmc/openbmc/issues/2338, modify URL of command rsetboot. 
Support 1.99.10 or above version.

For 1.99.9 or old version, support ``rsetboot <node> xxxx -o`` .

PR #4003, opened 2017-09-23
[OpenBMC] Support older endpoint path before OP910 1738 driver, fix rsetboot, 403 message.

whowutwut:
To support machines with older firmware than OP910 1738, we need to allow for an environment variable that would change the URL endpoints back to what it was prior to 1738.  

The issue here is that on https://github.com/xcat2/xcat-core/pull/3982 made a change to the endpoint that is used to select boot source based on what we had seen with FW 1738...  however, many of our machines are not upgraded yet to that FW level and so we are unable to control them correctly.  

Machines with firmware level > 1738 should NOT have any impact because of PR #3982, but prio firmware levels would see the following error: 
```
[root@briggs01 ~]# rsetboot mid05tor12cn18 net
mid05tor12cn18: Error: 403 Forbidden - This function is not yet available in OpenBMC firmware.
```

This is because the endpoint is not correct so we get a 403 error code. 

To work around with the PR, you would do the following: 
```
[root@briggs01 ~]# export XCAT_OPENBMC_FIRMWARE=1736
[root@briggs01 ~]# rsetboot mid05tor12cn18 net
mid05tor12cn18: Network
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -44,6 +44,7 @@ $::POWER_STATE_RESET        = "reset";
 $::POWER_STATE_REBOOT       = "reboot";
 $::UPLOAD_FILE              = "";
 $::UPLOAD_FILE_VERSION      = "";
+$::RSETBOOT_URL_PATH        = "boot";
 
 $::NO_ATTRIBUTES_RETURNED   = "No attributes returned from the BMC.";
 
@@ -253,7 +254,7 @@ my %status_info = (
 
     RSETBOOT_SET_REQUEST => {
         method         => "PUT",
-        init_url       => "$openbmc_project_url/control/host0/boot/attr/BootSource",
+        init_url       => "$openbmc_project_url/control/host0/$::RSETBOOT_URL_PATH/attr/BootSource",
         data           => "xyz.openbmc_project.Control.Boot.Source.Sources.",
     },
     RSETBOOT_SET_RESPONSE => {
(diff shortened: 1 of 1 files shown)


whowutwut:
This code was used by @gurevichmark and myself yesterday, so after the Travis CI checking is complete, I plan to merge it so get the fix into the development build so our internal teams can pick it up asap...

xuweibj:
Hi @whowutwut , for pull request #3982 option ``-o`` is used to support firmware and other old version, I think if you modify like this, we do not need ``-o `` option.

whowutwut:
@xuweibj , I see...  we were not  sure what that `-o` option was for but did not have time to take a closer look last Friday.   I will take a look at this -o option later this week...

whowutwut:
Actually, I just decided to try this out...

```
[root@briggs01 ~]# rsetboot mid05tor12cn18 net 
mid05tor12cn18: Error: 403 Forbidden - This function is not yet available in OpenBMC firmware.
[root@briggs01 ~]# rsetboot -o mid05tor12cn18 net
Unsupported command: rsetboot -o
[root@briggs01 ~]# rsetboot mid05tor12cn18 net -o
mid05tor12cn18: Error: 403 Forbidden - This function is not yet available in OpenBMC firmware.

```

How is it invoked? 

Here is the change with the PR to show that this machine is actually running the 1736 FW...
```
[root@briggs01 ~]# export XCAT_OPENBMC_FIRMWARE=1736
[root@briggs01 ~]# rsetboot mid05tor12cn18 net
mid05tor12cn18: Network
[root@briggs01 ~]# unset XCAT_OPENBMC_FIRMWARE
[root@briggs01 ~]# rsetboot mid05tor12cn18 net
mid05tor12cn18: Error: 403 Forbidden - This function is not yet available in OpenBMC firmware.
[root@briggs01 ~]# 
```

PR #4005, opened 2017-09-25
Improvements to Cumulus ZTP

mattaezell:
- Use the full path to cl-license
- Allow configinterface to download files
- Run enablesnmp in mypostscript so we have the correct community string

### DIFF
--- a/xCAT/postscripts/cumulusztp
+++ b/xCAT/postscripts/cumulusztp
@@ -104,21 +104,11 @@ done
 mkdir -p /root/.ssh
 mv _ssh/authorized_keys /root/.ssh/authorized_keys 
 
-#enable and config snmpd
-./enablesnmp
-rc=$?
-if [ "$rc" != "0" ];then
-    logger -s -t "xcat.cumulusztp" -p local4.err "enablesnmp: failed to enable SNMP"
-    echo "enablesnmp: failed to enable SNMP"
-else
-    logger -s -t "xcat.cumulusztp" -p local4.info "SNMP enabled!"
-fi
-
 #install license, if needed
-cl-license > /dev/null 2>&1
+/usr/cumulus/bin/cl-license > /dev/null 2>&1
 rc=$?
 if [ "$rc" != "0" ];then
-    if cl-license -i http://$server_ip/install/custom/sw_os/cumulus/licensefile.txt; then
+    if /usr/cumulus/bin/cl-license -i http://$server_ip/install/custom/sw_os/cumulus/licensefile.txt; then
         logger -s -t "xcat.cumulusztp" -p local4.info "installed Cumulus license"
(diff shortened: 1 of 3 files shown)


xCAT/postscripts/cumulusztp
```diff
@@ -104,21 +104,11 @@ done
...
-    logger -s -t "xcat.cumulusztp" -p local4.info "SNMP enabled!"
-fi
-
 #install license, if needed
-cl-license > /dev/null 2>&1
+/usr/cumulus/bin/cl-license > /dev/null 2>&1
 rc=$?
 if [ "$rc" != "0" ];then
-    if cl-license -i http://$server_ip/install/custom/sw_os/cumulus/licensefile.txt; then
+    if /usr/cumulus/bin/cl-license -i http://$server_ip/install/custom/sw_os/cumulus/licensefile.txt; then
```
immarvin: hi @mattaezell , what is the license policy for cumulus switches? one license per switch? for a specific license, is it bound to a specific switch?

mattaezell: Hi @immarvin.  According to https://support.cumulusnetworks.com/hc/en-us/articles/216422657-Perpetual-Licenses-General-Questions-Current-Licensing-Model-#are_cumulus_linux_licenses_tied_to_specific_hardware you just have to make sure you don't install licenses on more switches than you paid for.  If you would like, I could implement logic to allow it to try to download a hostname-specific file first.

cxhong:
Can we decide a common directory for cumulus files?  

for the full install,  xcat expected the onie bin in the location:
```
www-server = "http://129.40.41.221/install/onie/onie-installer";

```

mattaezell:
@cxhong Indeed, having those two directories is suboptimal.  The docs instruct users to put the license in /install/custom ([here](http://xcat-docs.readthedocs.io/en/stable/advanced/networks/onie_switches/os_cumulus/install.html) and [here](http://xcat-docs.readthedocs.io/en/stable/advanced/networks/onie_switches/os_cumulus/prepare.html)).  What I'm doing in my lab is symlinking `/install/onie/onie-installer` to `/install/custom/sw_os/cumulus/cumulus-prod.img` which is a symlink to the actual versioned image.  It probably makes sense to have the DHCP onie-installer path pulled from the site table

immarvin:
hi @mattaezell , thanks for you PR, this PR works well as a solution. 

I think we need to design and implement a formal method on the management of license file and switch configuration files , I  think we might need to discuss with the team on this, created https://github.com/xcat2/xcat-core/issues/4025 to trace the work.  It will be appreciated if you have some idea on this.

cxhong:
I like to merge this PR to allow configinterface to download xCAT.intf file from MN

Issue #4011, opened 2017-09-26
The detection of OpenBMC servers in xCAT common code has some flaws


whowutwut:
In common code between ipmi based BMC and OpenBMC based BMC, xCAT is using port 2200 as a way to differentiate a IPMI vs OpenBMC.  

However, the check is flawed, and we saw this when running `bmcdiscover`.  

The test is whether the port 2200 is **open** via `nmap`.  In the server we are looking at, the port status is `filtered`, NOT `open` so we go down IPMI code path... 

```
[root@p10xcat1 ~]# nmap 129.40.41.236 -p 2200 -Pn

Starting Nmap 6.40 ( http://nmap.org ) at 2017-09-26 09:46 EDT
Nmap scan report for 129.40.41.236
Host is up (0.00039s latency).
PORT     STATE    SERVICE
2200/tcp filtered ici
MAC Address: 70:E2:84:14:31:D3 (Wistron InfoComm(Zhongshan))

Nmap done: 1 IP address (1 host up) scanned in 0.86 seconds
```

When it drops into IPMI code, it's very confusing for both admins and xCAT team debugging these issues..... "Why is this code in IPMI?"

So if we determine this machine to be OpenBMC, then we should always go down the correct code path and fail later on, if the port is /filtered/closed/unfiltered/open, we should always go into the right code path and potentially fail later on. 

Similar issue and complaint is here... https://github.com/xcat2/xcat-core/issues/3975 where we should first detect machine type, then go down the correct path so we can detect unsupported systems, instead of going into wrong code paths.

This is just one place, (and maybe that's all we have..) but have to ensure that we covered in all places with this issue.

immarvin:
hi @xuweibj ,would you please take a look at this?thx

whowutwut:
This one we have encountered several times on various cluster and patching the files seems to make this better.. closing.

PR #4012, opened 2017-09-26
Assume that any bmc that has nmap response on the openbmc_port is a openbmc server

whowutwut:
If the openbmc console port is returned by nmap, assume that it's openbmc

Resolves the bmcdiscover command for #4011 

Otherwise if the nmap state is anything other than `open` then we will fall into IPMI code, which is confusing...

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -670,8 +670,8 @@ sub scan_process {
 
                 my $nmap_cmd = "nmap ${$live_ip}[$i] -p $openbmc_port -Pn";
                 my $nmap_output = xCAT::Utils->runcmd($nmap_cmd, -1);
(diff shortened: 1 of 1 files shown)


xuweibj:
Hi @whowutwut , I think this modify will cause another issue. Even node that mgt is ipmi will return info.
```
# nmap 50.3.17.1 -p 2200 -Pn

Starting Nmap 6.40 ( http://nmap.org ) at 2017-09-27 03:10 EDT
Nmap scan report for 50.3.17.1
Host is up (0.071s latency).
PORT     STATE  SERVICE
2200/tcp closed ici
```
I think maybe we could use closed to filter, if ``closed``, consider it as ipmi.

whowutwut:
👍 Thanks for that comment, @xuweibj  I will make some changes tomorrow.

whowutwut:
@xuweibj  I've updated this pull request.

cxhong:
one of system hit this issue, I modified their code with  this PR.  worked fine.  will merge it.

PR #4018, opened 2017-09-27
Have xcoll take a try at sorting output groups

jjohnson42:
It isn't natural sort, but in many cases it should work.

Natural sort in perl is a pain or a dependency.


immarvin:
i am ok with the change

PR #4019, opened 2017-09-27
Workaround switches that report ifindex in bridge

jjohnson42:
Some switches report ifindex instead of bridge index.

This is not compliant behavior.  However, we can detect
that every last mac was a dead end, and assume this is
the case to try to get better result.  This won't work
if there is overlap in interface and bridge indexes, so
in that case, the spec compliant behavior is assumed.


immarvin:
agree to merge

Issue #4007, opened 2017-09-26
Switch-based node discovery failed on cumulus switch


zet809:
For a cumulus switch haven't have ssh passwordless configured, our current code logic will fail to get mac table from it. 

Our code will use command `ssh -o StrictHostKeyChecking=no -o PasswordAuthentication=no c685f2u26vc 'bridge fdb show|grep -i -v permanent|tr A-Z a-z  2>/dev/null' 2>/dev/null` to get mac table from cumulus switch.
```
# ssh -o StrictHostKeyChecking=no -o PasswordAuthentication=no c685f2u26vc 'bridge fdb show|grep -i -v permanent|tr A-Z a-z  2>/dev/null' 2>/dev/null
```
But when using this command with xcat subroutine 'runcmd', `2>&1` will be appended to the command above and will cause the command have output like below:
```
# ssh -o StrictHostKeyChecking=no -o PasswordAuthentication=no c685f2u26vc 'bridge fdb show|grep -i -v permanent|tr A-Z a-z  2>/dev/null' 2>/dev/null 2>&1
Permission denied (publickey,password).
```
Then, the `refresh_switch` subroutine will get nothing from the cumulus switches.

immarvin:
encountered some issue on the return value of `runcmd`, in the MacMap.pm, the command is success  but `$?` and `$::RUNCMD_RC` are "-1", might be a issue in runcmd, need to take a look at this

immarvin:
thanks for the help of @neo954 and @robin2008 , we found a fatal issue in MacMap.pm,  this pm rewrite the signal handler for SIGCHLD
````
534     $SIG{CHLD} = sub { while (waitpid(-1, WNOHANG) > 0) { $children-- } };
````
which introduced 2 problems:
1). breaks the return value of ``xCAT::Utils->runcmd()``, since `qx` relies on SIGCHLD to pass back the return value of shell commands
2)if the child processes of `refresh_switch` forks their own children process, the SIGCHLD signal will be passed upward to `refresh_switch` and break the children count down

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/4022

PR #4022, opened 2017-09-28
fix issue Switch-based node discovery failed on cumulus switch #4007

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/4007:
 
retore the default SIG{CHLD} in the child processes

### DIFF
--- a/perl-xCAT/xCAT/MacMap.pm
+++ b/perl-xCAT/xCAT/MacMap.pm
@@ -543,11 +543,19 @@ sub refresh_table {
         pipe my $child, my $parent;
         $child->autoflush(1);
         $parent->autoflush(1);
+
         $children++;
         my $cpid = xCAT::Utils->xfork;
-        unless (defined $cpid) { die "Cannot fork" }
+        unless (defined $cpid) { 
+            $children--;
+            close($child);
+            close($parent);
+            xCAT::MsgUtils->message("S", "refresh_table: failed to fork refresh_switch process for $entry->{switch},skip..."); 
+            next;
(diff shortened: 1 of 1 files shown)


perl-xCAT/xCAT/MacMap.pm
```diff
@@ -548,6 +548,7 @@ sub refresh_table {
         unless (defined $cpid) { die "Cannot fork" }
 
         if ($cpid == 0) {
+            $SIG{CHLD} = 'DEFAULT';
```
chenglch: Suggest moving the line 546 `children ++` behind the `xfork`,  if xfork fails, there may be endless cycle to wait for the child process actually the child process is not exist.

immarvin: updated

perl-xCAT/xCAT/MacMap.pm
```diff
@@ -543,11 +543,19 @@ sub refresh_table {
...
+
         $children++;
         my $cpid = xCAT::Utils->xfork;
-        unless (defined $cpid) { die "Cannot fork" }
+        unless (defined $cpid) { 
+            $children--;
+            close($child);
+            close($parent);
+            xCAT::MsgUtils->message("S", "refresh_table: failed to fork refresh_switch process for $entry->{switch},skip..."); 
+            next;
```
robin2008: Failed to fork normally means there is not enough resource, not sure `next` here is good. Maybe give 1 sec delay here.

Issue #3998, opened 2017-09-22
[DEV] rpower will hang for long time when operating large amount nodes


robin2008:
2.13.7

Negative testing for fake node with non-accessing BMC IP.
For example, to operate 576 nodes 
`time rpower mid07tor[01-32]cn[01-18] on`

It will hang for long time.

robin2008:
This issue share the same root cause with https://github.com/xcat2/xcat-core/issues/4192
It is expected that the new fix could cover both the ticket.

Close this first. To track it in #4192

PR #4026, opened 2017-09-28
Fix issue 3998, check node's status before send login request

xuweibj:
#3998 

rpower on against 500 nodes (Unavailable)
Before modify: **48 minutes**
After modify:    22~24s

Test result (against openbmc simulator)
```
Nodes/rpower       100/on     100/status   200/on     200/status  500/on     500/status   1000/on     1000/status   1000(500 Unavailable)
Befor modify:      9.838s     9.015s       19.545s    17.2s       47.205s    42.925s      1m33.809s    1m25.232s         43m28s
After modify:      10.238s    9.364s       19.916s    18.06s      48.12s     44.04s       1m35.388s    1m26.558s         46.195s
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -362,6 +362,8 @@ my $xcatdebugmode = 0;
 
 my $flag_debug = "[openbmc_debug]";
 
+my @node_bmc_ips = ();
+
 #-------------------------------------------------------
 
 =head3  preprocess_request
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -482,12 +484,24 @@ sub process_request {
     $wait_node_num = keys %node_info;
     my @donargs = ();
 
+    my $all_bmc = join(" ", @node_bmc_ips);
```
robin2008: I think we cannot put `nmap` into preprocess handler as xcatd will dispatch the target nodes to SN first if hierarchy.

whowutwut:
@xuweibj @robin2008 Doesn't seem to be that much improvement (around + or - 1 second for each case?) , what's the driving force behind making this change?

robin2008:
@whowutwut  without the enhancement, if will take 2-3 sec for one unavailable node.  It means if we have 500 unavailable nodes, it will take about 20min.

@xuweibj just append the result when the node is up, we need to see if there are impact for normal case.

robin2008:
The fix can only cover the IP is not accessible issue, but not for the bmc restart case (say ip is pingable, but openbmc is not serviceable )

So, here we may not merge this, but waiting for a total solution for openbmc parallel enhance.

robin2008:
close it without merge, refer to https://github.com/xcat2/xcat-core/issues/4192 for new fix.

PR #4030, opened 2017-09-28
Support config snmpv1/2 community string for Mellanox switches during discovery phase

cxhong:
If pre-defined switch has snmpv1/2 community string defined, the command **switchdiscover --setup** options  will config it.  Due to limitation of switches table, user can only define either snmpv1/2 or snmpv3.
If user like to config both version snmp during discovery phase, we suggest:
1) define snmpv3 configuration for the predefine switch
2) switchdiscover --setup will setup snmpv3
3) then run another command **rspconfig $switch community=$community_string** to set up snmpv1


immarvin:
i am ok with the change ,thanks

PR #4045, opened 2017-10-04
reduce `awk` invocations in `dodiscovery`

samveen:
This MR reduces the invocations of `awk` in `dodiscovery`.


whowutwut:
@zet809 Could you take a review when you have a chance?

immarvin:
i am ok with the change

PR #4046, opened 2017-10-04
Improvements to RH/Centos Kickstart templates.

samveen:
Effort to clean up RH/Centos templates and remove divergence and errors that has crept in over time.


immarvin:
i am ok with the change

Issue #3991, opened 2017-09-21
[OpenBMC] rpower should detect when the BMC is in a service state 


whowutwut:
I'm not sure whether this will happen in production, but we should detect when the bmc is in a "service" state, for example when a CRONUS box is connected to it for debug. 

It would be good to do this anyway so that in our development, we know when a machine is in service mode or whether the compute side is ON. 

Debug mode is when the following occurs...  Chassis is ON and Host is OFF.. 

```
root@frame6u11-bmc:~# obmcutil state 
CurrentBMCState:    xyz.openbmc_project.State.BMC.BMCState.Ready
CurrentPowerState:  xyz.openbmc_project.State.Chassis.PowerState.On
CurrentHostState:   xyz.openbmc_project.State.Host.HostState.Off
root@frame6u11-bmc:~# 
```

gurevichmark:
Do you mean we just want to return something like:
```
[root@stratton01 xcat]# rpower p9euh01 bmcstate
p9euh01: BMC Service
[root@stratton01 xcat]#
```

Or should we always check for BMC state and not allow some `rpower` commands to go through when BMC is in service state ?

whowutwut:
@gurevichmark No, the rpower command should detect when it's in this "debug" state when query `state|stat`

Currently, when it's in that state, `rpower <node> state` returns the server is "On".  But the compute side is not booted.  So this is misleading because we will think that all your compute nodes are Running, but when you ssh to the compute node, it won't work and console will not have anything...  because the host is not booted, but we report "On" as the state.

zet809:
hi, @whowutwut and @gurevichmark , we have `rpower bmcstate` to check the status of BMC like this, so can we close this issue?
```
[root@briggs01 ~]# rpower p9up bmcstate
mid05tor12cn18: Error: BMC did not respond within 10 seconds, retry the command.
mid05tor12cn13: BMC Ready
mid05tor12cn16: BMC Ready
mid05tor12cn11: BMC Ready
mid05tor12cn15: BMC Ready
mid05tor12cn02: BMC Ready
mid05tor12cn05: BMC Ready
```

whowutwut:
This was resolved by  #4047 using commit 5b6074c9b490047169729b8aea1816f0a6b0be16

whowutwut:
@zet809 This is actually a state in the BMC when the Host is OFF, but the Chassis is ON.  It's a debug state and represented by `on (Chassis)`

PR #4047, opened 2017-10-04
Support transition states for OpenBMC - phase1

whowutwut:
This is a little interesting here.   The problem comes because OpenBMC is using a RESTful interface. and so commands are non blocking, we send a command to the BMC And it returns.  

For power on and power off, there's a period of time where the power states do not yet reflect the intended transition state.  Take Power off for example: 

1.  Machine is on 
2. Power off requested 
3. Transition to Off,  At this point, if you use xCAT to query state, it will return on.  

So it looks like the following: 
```
# rpower cn01 state
cn01: on
# rpower cn01 off
cn01: off     <--- here we just print target state as a response
# rpower cn01 state. 
cn01: on
# rpower cn01 state
cn01: off     <--- after some time, it will be off.  
```

So this is misleading because the user does not know if the command took or not.  So with this, xCAT attempted to take the transition state that we are able to query from OpenBMC and create `powering-off` and `powering-on`.   However, with some testing of this, it turns out that doing a hard off on the server, transition states are not reflected correctly.  

The first phase of this PR is to scaffold in some code to be able to print out transition state code. 

## Example: 

Set the VARIABLE
```
[root@briggs01 vhu]# export XCAT_OPENBMC_POWER_TRANSITION=YES
```

Only accurate with softoff
```
[root@briggs01 vhu]# rpower mid05tor12cn[15-17] softoff
mid05tor12cn16: powering-off
mid05tor12cn15: powering-off
mid05tor12cn17: powering-off
[root@briggs01 vhu]# rpower mid05tor12cn[15-17] stat
mid05tor12cn16: powering-off
mid05tor12cn15: powering-off
mid05tor12cn17: powering-off
```

If we unset the variable, it will default to NOT showing transition state, but actual state:
```
[root@briggs01 vhu]# unset XCAT_OPENBMC_POWER_TRANSITION
[root@briggs01 vhu]# rpower mid05tor12cn[15-17] stat
mid05tor12cn16: on
mid05tor12cn17: on
mid05tor12cn15: on
```

Eventually, once it powered off, the state will reflect correctly:
```
[root@briggs01 vhu]# export XCAT_OPENBMC_POWER_TRANSITION=YES
[root@briggs01 vhu]# rpower mid05tor12cn[15-17] stat
mid05tor12cn16: powering-off
mid05tor12cn17: powering-off
mid05tor12cn15: powering-off
[root@briggs01 vhu]# rpower mid05tor12cn[15-17] stat
mid05tor12cn16: powering-off
mid05tor12cn17: powering-off
mid05tor12cn15: powering-off
[root@briggs01 vhu]# rpower mid05tor12cn[15-17] stat
mid05tor12cn16: off
mid05tor12cn15: powering-off
mid05tor12cn17: powering-off
[root@briggs01 vhu]# rpower mid05tor12cn[15-17] stat
mid05tor12cn16: off
mid05tor12cn15: off
mid05tor12cn17: off
```

Same idea with powering on...
```
root@briggs01 vhu]# rpower mid05tor12cn17 on
mid05tor12cn17: powering-on
[root@briggs01 vhu]# rpower mid05tor12cn17 stat
mid05tor12cn17: powering-on
[root@briggs01 vhu]# rpower mid05tor12cn17 stat
mid05tor12cn17: on
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -391,6 +391,14 @@ sub preprocess_request {
         $::OPENBMC_FW = $request->{environment}->{XCAT_OPENBMC_FIRMWARE};
     }
 
+    # Provide a way to turn on and off transition state processing, default to off
+    if (ref($request->{environment}) eq 'ARRAY' and ref($request->{environment}->[0]->{XCAT_OPENBMC_POWER_TRANSITION}) eq 'ARRAY') {
+        $::OPENBMC_PWR = $request->{environment}->[0]->{XCAT_OPENBMC_POWER_TRANSITION}->[0];
+    } elsif (ref($request->{environment}) eq 'ARRAY') {
+        $::OPENBMC_PWR = $request->{environment}->[0]->{XCAT_OPENBMC_POWER_TRANSITION};
+    } else {
+        $::OPENBMC_PWR = $request->{environment}->{XCAT_OPENBMC_POWER_TRANSITION};
+    }
     ##############################################
 
     $callback  = shift;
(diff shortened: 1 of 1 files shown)


whowutwut:
@xuweibj  Thinking about phase 2, perhaps if we are able to know whether a hard off or soft off was issued against the machine (perhaps with the status field) , we would either show transition states or NOT show transition states.    There's currently a `setNodeStatusAttributes` but there is no get function for that...

xuweibj:
Hi @whowutwut , for ``setNodeStatusAttributes`` (here or in ipmi, kvm or others), ``powering_on`` means that run ``rpower on`` and received the "OK" from CN, maybe the CN is on successful or not we don't care here. And the same with ``powering_off``, it's just means the command exe successfully.

whowutwut:
Commit 5b6074c9b490047169729b8aea1816f0a6b0be16 address #3991

PR #4049, opened 2017-10-05
Update setupntp and call it from cumulusztp

mattaezell:
Make sure cumulus switches come up with NTP

### DIFF
--- a/xCAT/postscripts/setupntp
+++ b/xCAT/postscripts/setupntp
@@ -40,31 +40,10 @@ pmatch ()
   return 1          # non-zero return code means string not matched by pattern
 }
 
-# is_lsb_ubuntu exit status indicates whether system appears to be Ubuntu.
-# Using required /etc/lsb-release file, instead of optional lsb_release command.
-
-is_lsb_ubuntu ()
+is_lsb_cumulus ()
 {
-    awk '
-        (match($0, "^[ \t]*DISTRIB_ID=") == 1) {    # A DISTRIB_ID line
-            id = substr($0, RLENGTH + 1)            # Save its value
-        }
-
-        END {
-            # Examine last DISTRIB_ID value to see if Ubuntu indicated
-
-            if (match(id, "^(Ubuntu|\"Ubuntu\")[ \t]*$") == 1) {
-                exit 0    # Ubuntu
-            }
-            exit 1        # Not Ubuntu
-        }
-
-    ' /etc/lsb-release >/dev/null 2>&1
-
-    # Routine exit status is exit status of the last command -- the awk script.
-    #
-    #   Note: if /etc/lsb-release does not exist, the exit status indicates
-    #         failure (not Ubuntu), which is the correct outcome.
+  grep -qs Cumulus /etc/lsb-release
+  return $?
 }
 
 logger -t xcat "Install: Setup NTP"
@@ -97,6 +76,14 @@ fi
 
 OS_TYPE=`uname`
 if [ $OS_TYPE = Linux ]; then
+    # Set the timezone
+    if [ -n $TIMEZONE ]; then
+        echo $TIMEZONE > /etc/timezone
+        if which timedatectl > /dev/null; then
+            timedatectl set-timezone $(cat /etc/timezone)
+        fi
+    fi
+
     mkdir -p /var/lib/ntp
     chown ntp /var/lib/ntp
     if ( pmatch $OSVER "sles*" ) || ( pmatch $OSVER "suse*" ) || [ -f /etc/SuSE-release ];then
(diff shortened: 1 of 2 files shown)


mattaezell:
A couple comments:
- The function is_lsb_ubuntu wasn't being used, so I removed it
- 'ntpd -gq' seems to work for me on RHEL7, SLES11, SLES12, and Cumulus Linux. I don't think it's necessary to have the different code path for SLES
- Added setting the timezone (currently only works on systemd-based systems, ignored on others)

mattaezell:
@cxhong Would you please review this?

whowutwut:
@immarvin Could you take a review of this?

xCAT/postscripts/setupntp
```diff
@@ -97,6 +76,14 @@ fi
 
 OS_TYPE=`uname`
 if [ $OS_TYPE = Linux ]; then
+    # Set the timezone
+    if [ -n $TIMEZONE ]; then
+        echo $TIMEZONE > /etc/timezone
+        if which timedatectl > /dev/null; then
```
neo954: I suggest use
    `if type timedatectl >/dev/null`

xCAT/postscripts/setupntp
```diff
@@ -107,40 +94,19 @@ if [ $OS_TYPE = Linux ]; then
...
-        logger -t xcat "ntpdate -t5 $master "
-        ntpdate -t5 $master
-        if [ "$?" != "0" ];then
-            echo "  ntpdate -t5 $master failed"
-            logger -t xcat "ntpdate -t5 $master failed"
-        fi
-    fi
+
+    logger -t xcat "setting current time"
+    ntpd -gq > /dev/null 2>&1 || ntpdate -t5 $master > /dev/null 2>&1 || echo "WARNING: NTP Sync Failed!!"
```
neo954: Add another line with `logger -t xcat` when failing may help debugging.

immarvin: thanks @mattaezell , the code looks good, but using `if-else` logic with `logger` on each operation will help debugging

cxhong:
@mattaezell , verify the setupntp postscript from updatenode commands.   it looks like updatnode command stuck here and never returns back..
````
]# updatenode mid05tor25 setupntp
mid05tor25: xcatdsklspost: downloaded postscripts successfully
mid05tor25: Sun Jan 11 05:09:54 UTC 1970 Running postscript: setupntp
.................................
mid05tor25: + echo systemctl stop ntp
mid05tor25: systemctl stop ntp
mid05tor25: + '[' -z 'systemctl stop ntp' ']'
mid05tor25: + local retmsg
mid05tor25: ++ systemctl stop ntp
mid05tor25: + retmsg=
mid05tor25: + retval=0
mid05tor25: + '[' 0 = 0 ']'
mid05tor25: + grep -i 'Running in chroot,\s*ignoring request.*'
mid05tor25: + echo ''
mid05tor25: + return 0
mid05tor25: + logger -t xcat 'setting current time'
mid05tor25: + ntpd -gq
``````
the process on the cumulus switches:
````
root@mid05tor25:/etc# ps -ef | grep ntpd
root     10768 10716  0 05:09 ?        00:00:00 ntpd -gq
root     11028 10911  0 05:28 pts/0    00:00:00 grep ntpd
# service ntp status
● ntp.service - NTP - Network Time Protocol daemon
   Loaded: loaded (/lib/systemd/system/ntp.service; enabled)
   Active: inactive (dead) since Sun 1970-01-11 05:09:55 UTC; 19min ago
     Docs: man:ntpd(8)
 Main PID: 1360 (code=exited, status=0/SUCCESS)
````

**Comment by cxhong** [comment]
found issue here,  cumulus switch is on different network than master
````
root@mid05tor25:/xcatpost# ifconfig eth0
eth0      Link encap:Ethernet  HWaddr a8:2b:b5:59:ac:40
          inet addr:172.11.205.25  Bcast:172.11.255.255  Mask:255.255.0.0
          inet6 addr: fe80::aa2b:b5ff:fe59:ac40/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
`````
the ```MASTER='172.10.253.27'```,  it couldn't reach from cumulus switch.  
```
root@mid05tor25:/etc# ping 172.10.253.27
connect: Network is unreachable
````
the ntp server has to change to 172.11 network
````
root@mid05tor25:/xcatpost# cat /etc/ntp.conf
server 172.11.253.27
driftfile /var/lib/ntp/drift
disable auth
restrict 127.0.0.1
interface listen eth0
````

**Comment by cxhong** [comment]
if ntpserver is not reachable, the ntpd -g will take 1000s to exit.
```
-g     Normally, ntpd exits with a message to the system log if the offset exceeds the panic threshold, which  is 1000s by default.
```
we should check the ntpserver before issue ntpd command.  
@mattaezell , can u make this changes?  and comments from @immarvin and @neo954

cxhong:
@mattaezell , not sure if you have time to work that.  I can merge your PR first.  then i can make changes on top of your code.  please let me know..  thanks.

mattaezell:
Hi @cxhong.  Yes, sorry, I have been focusing all of my time on CORAL deployment efforts.  We are running this patch in our local install so I haven't had the need to get it pushed upstream.  Feel free to either merge it as-is and then fix it with another commit, reject this MR and just fix it the way you want, or wait until some of my time frees up.  Whatever is most convenient for you.

Thanks for staying on top of this.

cxhong:
I will merge it then create another PR.

PR #4060, opened 2017-10-09
fix xHRM fails when installnic is not set

### LINKED ISSUE #4058: xHRM fails when installnic is not set
Hi!

When `installnic` is not set in the `noderes` table, the documentation states that the value will default to `primarynic`, and then to the `mac` keyword in case `primarynic` is not defined either.

Turns out this is not really the case, as the default diskful install template (`mypostscript.tmpl`) does [this](https://github.com/xcat2/xcat-core/blob/master/xCAT-server/share/xcat/mypostscript/mypostscript.tmpl#L43):
```
:INSTALLNIC=#TABLE:noderes:$NODE:installnic#
```
So if `installnic` is not defined in the `noderes` table, an empty `INSTALLNIC` environment variable is exported to the postinstall scripts.

As a consequence, the `xHRM` postinstall script fails when called without any argument, when `installnic` is not set, and exits with ["Incorrect usage"](https://github.com/xcat2/xcat-core/blob/master/xCAT-server/share/xcat/scripts/xHRM#L141):

```
# lsdef -c sh-hv01 -i installnic
sh-hv01: installnic=

# updatenode sh-hv01 -P "xHRM bridgeprereq"
sh-hv01: xcatdsklspost: downloaded postscripts successfully
sh-hv01: Fri Oct  6 15:07:24 PDT 2017 Running postscript: xHRM
sh-hv01: Incorrect usage
sh-hv01: postscript: xHRM exited with code 1
sh-hv01: Running of postscripts has completed.
```

bybai:
fix #4058 
enhance xHRM based on doc:
"When installnic is not set in the noderes table, the documentation states that the value will default to primarynic, and then to the mac keyword in case primarynic is not defined either."

### DIFF
--- a/xCAT-server/share/xcat/scripts/xHRM
+++ b/xCAT-server/share/xcat/scripts/xHRM
@@ -1,4 +1,4 @@
-#!/bin/bash 
+#!/bin/bash  
 if [ "$(uname -s|tr 'A-Z' 'a-z')" = "linux" ];then
    str_dir_name=`dirname $0`
    . $str_dir_name/xcatlib.sh
@@ -127,15 +127,23 @@ elif [ "bridgeprereq" = "$1" ]; then
     # get the port for installation
     if [ -n "$INSTALLNIC" ]; then
         if [[ "$INSTALLNIC" = mac ]] || [[ "$INSTALLNIC" = MAC ]]; then
-            #INSPORT=`ifconfig -a|grep -v inet6| grep -i 'HWaddr '$MACADDRESS|head -n 1|awk '{print $1}'`
+            #INSPORT=`ifconfig -a|grep -v inet6| grep -i 'HWaddr '$MACADDRESS|head -n 1|awk '{print $1}'`            
             INSPORT=`ip -oneline link show|grep -i ether|grep -i $MACADDRESS|awk -F ':' '{print $2}'|grep -o "[^ ]\+\( \+[^ ]\+\)*"`
         else
             INSPORT=$INSTALLNIC
         fi
+    elif [ -n "$PRIMARYNIC" ]; then
+        if [[ "$PRIMARYNIC" = mac ]] || [[ "$PRIMARYNIC" = MAC ]]; then
+            INSPORT=`ip -oneline link show|grep -i ether|grep -i $MACADDRESS|awk -F ':' '{print $2}'|grep -o "[^ ]\+\( \+[^ ]\+\)*"`
+        else
+            INSPORT=$PRIMARYNIC
+        fi
+    elif [ -n "$MACADDRESS" ]; then      
+            INSPORT=`ip -oneline link show|grep -i ether|grep -i $MACADDRESS|awk -F ':' '{print $2}'|grep -o "[^ ]\+\( \+[^ ]\+\)*"`
     fi
 
     if [ -z "$NETDESC" ]; then
-        if [ -n "$INSTALLNIC" ]; then
+        if [ -n "$INSPORT" ]; then
           NETDESC=$INSPORT:default
         else
             echo "Incorrect usage"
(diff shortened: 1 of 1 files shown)


xCAT-server/share/xcat/scripts/xHRM
```diff
@@ -127,15 +127,23 @@ elif [ "bridgeprereq" = "$1" ]; then
     # get the port for installation
     if [ -n "$INSTALLNIC" ]; then
         if [[ "$INSTALLNIC" = mac ]] || [[ "$INSTALLNIC" = MAC ]]; then
```
immarvin: hi @bybai ,the logic here seems a little redundant, is it ok to simplify the logic like this?
````
if [ -n "$INSTALLNIC" ]; then
    INSPORT=$INSTALLNIC
elif [ -n "$PRIMARYNIC" ]; then
    INSPORT=$PRIMARYNIC
fi

if ([ -z "$INSPORT" ] || [[  "$INSPORT" =~ ^(mac|MAC)$ ]]) && [ -n "$MACADDRESS" ] ; then
     INSPORT=$MACADDRESS;
fi

if [[ "$INSPORT"  =~ ^([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2}$ ]] ; then
    INSPORT=`ip -oneline link show|grep -i ether|grep -i $INSPORT |awk -F ':' '{print $2}'|grep -o "[^ ]\+\( \+[^ ]\+\)*"`
fi
````

Comment (immarvin) [reply xCAT-server/share/xcat/scripts/xHRM]: btw, the installnic and primarynic might be the value of mac address

**Comment by bybai** [comment]
The UT:
.......
bybc0609: + '[' -n '' ']'
bybc0609: + '[' -n mac ']'
bybc0609: + INSPORT=mac
bybc0609: + '[' -z mac ']'
bybc0609: + [[ mac =~ ^(mac|MAC)$ ]]
bybc0609: + '[' -n 42:f5:0a:05:6a:09 ']'
bybc0609: + INSPORT=42:f5:0a:05:6a:09
bybc0609: + [[ 42:f5:0a:05:6a:09 =~ ^([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2}$ ]]
bybc0609: ++ grep -i ether
bybc0609: ++ grep -i 42:f5:0a:05:6a:09
bybc0609: ++ awk -F : '{print $2}'
bybc0609: ++ ip -oneline link show
bybc0609: ++ grep -o '[^ ]\+\( \+[^ ]\+\)*'
bybc0609: + INSPORT=eth0
.........

File: xCAT-server/share/xcat/scripts/xHRM [hunk 2]
```diff
@@ -126,16 +126,21 @@ elif [ "bridgeprereq" = "$1" ]; then
...
-            INSPORT=`ip -oneline link show|grep -i ether|grep -i $MACADDRESS|awk -F ':' '{print $2}'|grep -o "[^ ]\+\( \+[^ ]\+\)*"`
-        else
-            INSPORT=$INSTALLNIC
-        fi
+        INSPORT=$INSTALLNIC
+    elif [ -n "$PRIMARYNIC" ]; then
+        INSPORT=$PRIMARYNIC
+    fi
+
+    if ([ -z "$INSPORT" ] || [[  "$INSPORT" =~ ^(mac|MAC)$ ]]) && [ -n "$MACADDRESS" ] ; then
```
immarvin: might need to consider the scenario when primarynic or installnic is "mac", but the $MACADDRESS is blank,

bybai:
If $MACADDRESS is blank, ut:
bybc0609: + '[' -n '' ']'
bybc0609: + '[' -n mac ']'
bybc0609: + INSPORT=mac
bybc0609: + '[' -z mac ']'
bybc0609: + [[ mac =~ ^(mac|MAC)$ ]]
bybc0609: + '[' -n '' ']'
bybc0609: + echo 'should configure mac in bybc0609 definition.'
bybc0609: should configure mac in bybc0609 definition.
bybc0609: + exit 1

immarvin:
i am ok with the change

Issue #4057, opened 2017-10-06
[OpenBMC] Enhance framework to accept node ranges 


whowutwut:
If I have a node group defined for OpenBMC nodes, the framework is such that when we are parsing the args, we do not understand the node range...

Here's a p9 group that contains the following nodes:
```
[root@stratton01 xcat]# nodels p9
frame6u11
p9euh01
p9euh01-bmc
p9euh02
p9euh02-bmc
```

When we execute a invalid command ,we only get a response for all..
```
[root@stratton01 xcat]# rbeacon p9 state
Only 'on' or 'off' is supported for OpenBMC managed nodes.
```

But on IPMI managed node, 
```
[root@stratton01 xcat]# nodels boston
boston01
boston02
```

We get a response for each node in the node range...
```
[root@stratton01 xcat]# rbeacon boston stat
boston02: Error: Only 'on' or 'off' is supported for IPMI managed nodes.
boston01: Error: Only 'on' or 'off' is supported for IPMI managed nodes.
```

xuweibj:
The issue was fixed in pull request #4061 .

whowutwut:
Thanks I verified with the dev changes

PR #4061, opened 2017-10-09
Fix issue 4057, enhance framework to print node range when unsupport

xuweibj:
#4057 

```
# rbeacon p9 state
boston01: Error: Only 'on' or 'off' is supported for OpenBMC managed nodes.
boston02: Error: Only 'on' or 'off' is supported for OpenBMC managed nodes.
```


whowutwut:
@xuweibj How did you run that example above? The boston nodes are IPMI managed?

xuweibj:
@whowutwut The code is for OpenBMC, I just created two nodes that mgt=openbmc and groups=p9 (without bmc ip), and then run ``rbeacon stat`` against them.

zet809:
Let's merge this PR.

Issue #4062, opened 2017-10-09
bmcdiscover cases failed in automation (IPMI nodes but enter OpenBMC branch)


xuweibj:
OS:  RH 7.4 
ARCH: ppc64

```
------START::bmcdiscover_range_w::Time:Sun Oct  8 09:43:39 2017------

FILENAME:/opt/xcat/bin/../share/xcat/tools/autotest/testcase//bmcdiscover/cases0

RUN:bmcdiscover --range  10.4.30.254 -u USERID -p PASSW0RD -w [Sun Oct  8 09:43:39 2017]
ElapsedTime:1 sec
RETURN rc = 0
OUTPUT:
Warning: HTTP::Response=HASH(0x10038404b70)->status_line for 10.4.30.254
CHECK:rc == 0   [Pass]
CHECK:output =~ Writing node    [Failed]

------END::bmcdiscover_range_w::Failed::Time:Sun Oct  8 09:43:40 2017 ::Duration::1 sec------
------START::bmcdiscover_range_z::Time:Sun Oct  8 09:43:40 2017------

FILENAME:/opt/xcat/bin/../share/xcat/tools/autotest/testcase//bmcdiscover/cases0

RUN:bmcdiscover --range  10.4.30.254 -u USERID -p PASSW0RD -z [Sun Oct  8 09:43:40 2017]
ElapsedTime:1 sec
RETURN rc = 0
OUTPUT:
Warning: HTTP::Response=HASH(0x10038404b40)->status_line for 10.4.30.254
CHECK:rc == 0   [Pass]
CHECK:output =~ bmc=10.4.30.254 [Failed]

------END::bmcdiscover_range_z::Failed::Time:Sun Oct  8 09:43:41 2017 ::Duration::1 sec------
```

```
# nmap 10.4.30.254 -p 2200 -Pn

Starting Nmap 6.40 ( http://nmap.org ) at 2017-10-09 02:05 EDT
Nmap scan report for 10.4.30.254
Host is up (0.00043s latency).
PORT     STATE    SERVICE
2200/tcp filtered ici
MAC Address: 34:40:B5:B9:BA:D7 (IBM)

Nmap done: 1 IP address (1 host up) scanned in 7.24 seconds
```

whowutwut:
I think this code change probably caused it...  https://github.com/xcat2/xcat-core/pull/4012

whowutwut:
But actually, I think this why is this port `filtered`?  What is listening on this port? 
[...]

PR #4066, opened 2017-10-10
Fix issue 4062, modify method of node type check in bmcdiscover

xuweibj:
#4062 

Using `` ipmitool-xcat -I lanplus mc info`` to get "Manufacturer ID" and "Product ID" to check whether it‘s OpenBMC. 
For OpenBMC node, `` ipmitool-xcat -I lanplus -H x.x.x.x -P 0penBmc mc info``. If is "42817" and "16975", consider it as OpenBMC node.
For IPMI node, ``ipmitool-xcat -I lanplus -H x.x.x.x -U USERID -P PASSW0RD mc info``. If not"42817" and "16975", IPMI node.   
If can't get by using openbmc and ipmi password, consider it as OpenBMC node(For OpenBMC nodes, may not support ipmi service).

Output is the same with before.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -47,6 +47,8 @@ my $bmc_user;
 my $bmc_pass;
 my $openbmc_user;
 my $openbmc_pass;
+$::P9_WITHERSPOON_MFG_ID     = "42817";
+$::P9_WITHERSPOON_PRODUCT_ID = "16975";
 
 #-------------------------------------------------------
 
@@ -667,14 +669,19 @@ sub scan_process {
                 # Set child process default, if not the function runcmd may return error
                 $SIG{CHLD} = 'DEFAULT';
 
-                my @mc_cmds = ("/opt/xcat/bin/ipmitool-xcat -I lanplus -H ${$live_ip}[$i] -P $openbmc_pass mc info",
-                              "/opt/xcat/bin/ipmitool-xcat -I lanplus -H ${$live_ip}[$i] -U $bmc_user -P $bmc_pass mc info");
+                my $bmcusername;
+                my $bmcpassword;
+                $bmcusername = "-U $bmc_user" if ($bmc_user);
+                $bmcpassword = "-P $bmc_pass" if ($bmc_pass);
+                
+                my @mc_cmds = ("/opt/xcat/bin/ipmitool-xcat -I lanplus -H ${$live_ip}[$i] -P $openbmc_pass mc info -N 1 -R 1",
+                              "/opt/xcat/bin/ipmitool-xcat -I lanplus -H ${$live_ip}[$i] $bmcusername $bmcpassword mc info -N 1 -R 1");
                 my $mc_info;
                 my $flag;
                 foreach my $mc_cmd (@mc_cmds) {
                     $mc_info = xCAT::Utils->runcmd($mc_cmd, -1);
                     if ($mc_info =~ /Manufacturer ID\s*:\s*(\d+)\s*Manufacturer Name.+\s*Product ID\s*:\s*(\d+)/) {
-                        if ($1 eq "42817" and $2 eq "16975") {
+                        if ($1 eq $::P9_WITHERSPOON_MFG_ID and $2 eq $::P9_WITHERSPOON_PRODUCT_ID) {
                             bmcdiscovery_openbmc(${$live_ip}[$i], $opz, $opw, $request_command);
                             $flag = 1;
                             last; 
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -668,14 +667,26 @@ sub scan_process {
...
-                } else {
-                    bmcdiscovery_ipmi(${$live_ip}[$i], $opz, $opw, $request_command);
+                my @mc_cmds = ("/opt/xcat/bin/ipmitool-xcat -I lanplus -H ${$live_ip}[$i] -P $openbmc_pass mc info",
+                              "/opt/xcat/bin/ipmitool-xcat -I lanplus -H ${$live_ip}[$i] -U $bmc_user -P $bmc_pass mc info");
+                my $mc_info;
+                my $flag;
+                foreach my $mc_cmd (@mc_cmds) {
+                    $mc_info = xCAT::Utils->runcmd($mc_cmd, -1);
+                    if ($mc_info =~ /Manufacturer ID\s*:\s*(\d+)\s*Manufacturer Name.+\s*Product ID\s*:\s*(\d+)/) {
+                        if ($1 eq "42817" and $2 eq "16975") {
```
whowutwut: I'd like these to be globals or at least variables that give us some information...
```
$::P9_WITHERSPOON_MFG_ID="42817"
$::P9_WITHERSPOON_PRODUCT_ID="16975" 
```
Incase in the future we support other types of OpenBMC servers

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -668,14 +667,26 @@ sub scan_process {
...
                 $SIG{CHLD} = 'DEFAULT';
 
-                my $nmap_cmd = "nmap ${$live_ip}[$i] -p $openbmc_port -Pn";
-                my $nmap_output = xCAT::Utils->runcmd($nmap_cmd, -1);
-                if ($nmap_output !~ /$openbmc_port(.+)closed/) {
-                    # If the openbmc_port is anything execpt 'closed' assume it's OpenBMC server
-                    bmcdiscovery_openbmc(${$live_ip}[$i], $opz, $opw, $request_command);
-                } else {
-                    bmcdiscovery_ipmi(${$live_ip}[$i], $opz, $opw, $request_command);
+                my @mc_cmds = ("/opt/xcat/bin/ipmitool-xcat -I lanplus -H ${$live_ip}[$i] -P $openbmc_pass mc info",
```
whowutwut: This seems quite inefficient... @zet809 Do you recall the conversation with OpenBMC team regarding why -u is not passed into the ipmitool command . Do all other IPMI management accept this?  

Do we know what the performance is between the old code vs this new code, when we scale up?

whowutwut:
Conversation about this in Slack to OpenBMC team: https://ibm-systems-power.slack.com/archives/C0Q6TQP5Z/p1507640467000387

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
xuweibj: I modified the ipmitool command as the command in bmcdiscovery_ipmi function. I think it's OK for all IPMI management. If it's not work, I think we could not get mtms or other info in bmcdiscovery_ipmi function too.

I tested on briggs01, against 172.11.139.2-18, the performance is just like before. 
And I tested against 2000 IPs, 500 Hosts up(simulator, do not support ipmi), the performance is better than before. But I tested this on my MN c910f03c05k04, maybe different with on physical machine.
For the nodes that do not support ipmi, it will spend almost 2s to wait for rsp, and then enter bmcdiscovery_openbmc, so I thought it would spend more time than before. But the test result is not like that.

zet809: Hi, @whowutwut , no idea of why -u is not used. As I know, the **only** BMC that didn't support -U is FSP simulated BMC(switch to OPAL mode from PowerVM mode), it use userid 1 which have empty username. 
root@c910f03c05k10:/tmp# ipmitool-xcat -I lanplus -H 50.3.19.2 -P passw0rd user list
ID  Name	     Callin  Link Auth	IPMI Msg   Channel Priv Limit
1                    true    false      true       ADMINISTRATOR
2   test002          true    false      true       ADMINISTRATOR
3   user001          true    false      true       ADMINISTRATOR

And for the performance, based on the code logic that a new process will be created for an IP, I don't think it will be a problem.

whowutwut:
Since some folks on openBMC seem to think this would be unique to identify, let's check this in  to see if it will help with some of the issues with other BMCs and go from there.  @xuweibj Thanks!

PR #4078, opened 2017-10-11
'makehosts -n' Improvement

mattaezell:
- 'makehosts -n' will leave the server without a hosts file while it is
  generating the new one, but there's really no reason for that.
- While /etc/hosts should never be missing, 'makehosts -n' should
  include the loopback addresses no matter what.
- makehosts with and without '-n' should use the same locking to prevent
  multiple concurrent execution.


bybai:
I am ok to merge it.

Issue #3888, opened 2017-09-08
If service node is down, control to OpenBMC node never returns


whowutwut:
I use never, but I didn't wait forever for this command to come back, maybe 2-3 minutes before I `ctrl-c` out.  

```
[root@briggs01 ~]# rpower mid05tor12cn02 state
^C^C^C
```

turns out the service node is not installed and in anaconda...
```
[root@briggs01 ~]# ssh sn02
ssh: connect to host sn02 port 22: Connection refused
```

Changing to a responding service node resolves it...
```
[root@briggs01 ~]# chdef mid05tor12cn02 servicenode=fsn01
1 object definitions have been created or modified.

[root@briggs01 ~]# rpower mid05tor12cn02 state
mid05tor12cn02: [openbmc_debug] curl -k -c cjar -H "Content-Type: application/json" -d '{ "data": ["root", "xxxxxx"] }' https://172.12.140.102/login
mid05tor12cn02: [openbmc_debug] login_response 200 OK
mid05tor12cn02: [openbmc_debug] curl -k -b cjar -X GET -H "Content-Type: application/json" https://172.12.140.102/xyz/openbmc_project/state/enumerate
mid05tor12cn02: [openbmc_debug] rpower_status_response 200 OK
mid05tor12cn02: off
```

I thought there was something added @xuweibj to time out...?? Maybe in this case it's something inside xCAT that holds up the service node communication?

xuweibj:
I tried to reproduce this issue, but I got the response quickly.
```
[root@briggs01 ~]# time rpower mid05tor12cn02 status
Error: Unable to dispatch hierarchical sub-command to sn02:3001.  Error: Connection failure: IO::Socket::INET: connect: No route to host at /opt/xcat/lib/perl/xCAT/Client.pm line 248.
.

real	0m3.290s
user	0m0.087s
sys	0m0.009s
```
Test env:
```
# rpower sn02 status
sn02: off
# lsdef mid05tor12cn02 -i servicenode -c
mid05tor12cn02: servicenode=sn02
```
Hi @whowutwut , what else should I do to reproduce it?

zet809:
Hi, @whowutwut , can we close this issue if it can not reproduced?

immarvin:
hi @whowutwut , I made some investigation on this, I guess the root cause is like this:
[...]

PR #4083, opened 2017-10-11
fix issue If service node is down, control to OpenBMC node never returns #3888

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/3888:
use port 3054 instead of 3001 to run debug commands inside installer


immarvin:
TODO: check whether it is possible to add some timeout mechanism in submit_request while waiting for response

Issue #4067, opened 2017-10-10
[CORAL]pgsqlsetup support adding more hosts in pg_hba.conf


bybai:
This requirement is coming from the following xcat MN HA:
1, In xCAT MN HA, when using virtual ip in postgresql, pgsqlsetup will add site.master ip into pg_hba.conf, if the site.master ip is alias virtual ip, `pgsql` cannot connect `postgresql` as following error, so `pgsqlsetup` should add physical ip into pg_hba.conf for HA.
```
[root@bybc0609 pgsql]# su - postgres
Last login: Tue Sep 26 04:34:46 EDT 2017 on pts/1
-bash-4.2$ psql -h 10.5.106.100 -U xcatadm -d xcatdb
psql: FATAL:  no pg_hba.conf entry for host "10.5.106.9", user "xcatadm", database "xcatdb", SSL off
```
2.  In HA, both primary MN ip and standby MN ip should be added into pg_hba.conf;

immarvin:
seems this feature has been implemented in PR https://github.com/xcat2/xcat-core/pull/3519/files

robin2008:
It seems difference,  one is for the IP which postgresql daemon is listening, and the asking is for extra host authentication in `pg_hba.conf`

bybai:
@immarvin , Just like @robin2008 said, the enhancement need add IPv4 local connections in `pg_hba.conf` like:
```
host    all          all        10.5.106.100/32    md5
host    all          all        10.5.106.5/32      md5
```

bybai:
After reading pgsqlsetup code, I found the option "-a|--access" is to add IPv4 local connection in `pg_hba.conf`, but the `pgsqlsetup` manpage says "-a" is for adding service node access, so `pgsqlsetup` code is no need to be changed, man page should be enhanced.

whowutwut:
Helps resolve #3938

PR #4084, opened 2017-10-11
enhance pgsqlseup manpage about access addresses

bybai:
for #4067 

UT : after executing `pgsqlsetup -i -a 10.5.106.5 -a 10.5.106.9`, these 2 nodes can access the postgresql DB.


robin2008:
You still need to update the docs for the manpage

Issue #4065, opened 2017-10-09
bmcsetup as postscript hangs due to background process


mattaezell:
I discovered a couple boston nodes with an old firmware level, and the BMCs failed to configure with `bmcsetup`.  Once they booted to OS, I tried `updatenode <mynode> -P bmcsetup`, but it hangs at the end:
```
...
gw1: bmcsetup: Enabling SOL for channel 1: OK
gw1: bmcsetup: Enabling SOL for ADMIN
gw1: bmcsetup: Enabling SOL for ADMIN: OK
gw1: bmcsetup: Lighting Identify Light
```
It looks like `mypostscript` is waiting on the backgrounded process that blinks the LED
```
# pstree -p|grep mypos
           |-mypostscript(5706)---sleep(10102)
           |-sshd(5247)---bash(5253)---file7gGC8F.dsh(5258)---mypostscript(5335)---tee(5347)
```

I guess when run inside `genesis` the chain scripts don't care and the node just gets rebooted?  I certainly don't want the beacon left on forever.  Is there a way to detect if this is running in `genesis` or in a running osimage?

whowutwut:
@zet809  Can you take a look?

zet809:
Fixed in #4089

immarvin:
hi @mattaezell , can this issue be closed?

mattaezell:
We haven't moved to a snapshot with this fix to verify, but I'm OK to close.

PR #4089, opened 2017-10-12
fix issue 4065: bmcsetup as postscript hangs due to background process

zet809:
Fix issue #4065 

The current output will be:
```
...
bmcsetup: Enabling SOL for channel 1
bmcsetup: Enabling SOL for channel 1: OK
bmcsetup: Enabling SOL for ADMIN
bmcsetup: Enabling SOL for ADMIN: OK
bmcsetup: Lighting Identify Light
Chassis identify interval: 250 seconds
Wed Oct 11 04:24:34 EDT 2017 postscript bmcsetup return with 0
returned from postscript
```

### DIFF
--- a/xCAT-genesis-scripts/bin/bmcsetup
+++ b/xCAT-genesis-scripts/bin/bmcsetup
@@ -640,6 +640,16 @@ while [ $idev -gt 0 ]; do
 
     logger -s -t $log_label -p local4.info "Lighting Identify Light"
     if [ "$XPROD" = "43707" -a "$IPMIMFG" = '0' ]; then
+        ISOPENPOEWR=1
+    elif [ "$IPMIMFG" = "10876" ];then
+        # MFG=10876 and PROD=2437 is for BOSTON server and MFG=10876 and PROD=2355 is for B&S server
+        if [ "$XPROD" = "2437" -o "$XPROD" = "2355" ]; then
+            ISOPENPOEWR=1
+        fi
+    fi
+ 
+    
+    if [ "$ISOPENPOEWR" = '1' ]; then
         # OpenPOWER BMC specific, turn on the LED beacon light.  
         #   - default interval, # ipmitool chassis identify
         #                         Chassis identify interval: default (15 seconds)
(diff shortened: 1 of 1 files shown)


whowutwut:
@zet809 Seems like the reason for the hang is because we did not detect Boston's as an OpenPower server?

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -640,6 +640,16 @@ while [ $idev -gt 0 ]; do
 
     logger -s -t $log_label -p local4.info "Lighting Identify Light"
     if [ "$XPROD" = "43707" -a "$IPMIMFG" = '0' ]; then
+        ISOPENPOEWR=1
+    elif [ "$IPMIMFG" = "10876" ];then
+        # MFG=10876 and PROD=2437 is for BOSTON server and MFG=10876 and PROD=2355 is for B&S server
```
whowutwut: I would suggest changing comments to: 
```
# Handle Supermicro Servers (MFG=10876)
# Boston (PROD=2437), Briggs/Stratton (PROD=2355)
```
Because we might not remember what B&S stands for after some time...

zet809:
Hi, @whowutwut , modified, pls help to review.

xCAT-genesis-scripts/bin/bmcsetup
```diff
@@ -640,6 +640,17 @@ while [ $idev -gt 0 ]; do
 
     logger -s -t $log_label -p local4.info "Lighting Identify Light"
     if [ "$XPROD" = "43707" -a "$IPMIMFG" = '0' ]; then
+        ISOPENPOEWR=1
```
mattaezell: s/ISOPENPOEWR/ISOPENPOWER/  ?

zet809:
Thx @mattaezell , modified.

mattaezell:
So, will probably fix it for my `Boston` nodes (haven't tested yet), but would this hang still occur on generic x86 hardware?

zet809:
Hi, @mattaezell , you are right, but the bmcsetup is not designed to be run in OS, it won't hung in genesis with the original code. 
So, the fix here is focusing on OpenPOWER

whowutwut:
@mattaezell is that a general question about x86 or did you see it actually hang?

mattaezell:
Hi @whowutwut - just a general question.  Our only x86 xCAT cluster did not use `bmcsetup` (the BMCs were setup manually).

zet809:
The original code won't hung in genesis kernel. As when it works OK for OpenPOWER servers(Boston, S822LC), @whowutwut and @mattaezell , can we deal with x86_64 support in the future?

zet809:
hi, @whowutwut , I have modified based on your and @mattaezell comments, will you pls help to review again? Thx!

Issue #4053, opened 2017-10-05
[OpenBMC] Improvements needed output of flashing process and option parsing


whowutwut:
This is just a placeholder for adding better protection on the rflash command or enhance rflash to support multiple options...

```
[root@briggs01 ~]# rflash mid05tor12cn05 -l 
mid05tor12cn05: ID       Purpose State      Version
mid05tor12cn05: -------------------------------------------------------
mid05tor12cn05: 9e55358e BMC     Ready      ibm-v1.99.10-0-r11-0-g9c65260
mid05tor12cn05: 221d9020 Host    Active(*)  IBM-witherspoon-redbud-ibm-OP9_v1.19_1.33
mid05tor12cn05: 776f6a74 Host    Ready      IBM-witherspoon-ibm-OP9_v1.19_1.52
mid05tor12cn05: 0        BMC     Active(*)  ibm-v1.99.10-0-r7-0-gcce0f1b
mid05tor12cn05: f7f626b8 Host    Active     IBM-witherspoon-sequoia-ibm-OP9_v1.19_1.37
mid05tor12cn05: 
[root@briggs01 ~]# rflash mid05tor12cn05 -a 9e55358e 776f6a74
mid05tor12cn05: rflash started, please wait...
mid05tor12cn05: Activating firmware update. 10%
mid05tor12cn05: Activating firmware update. 10%
mid05tor12cn05: Activating firmware update. 10%
mid05tor12cn05: Activating firmware update. 10%
mid05tor12cn05: Activating firmware update. 10%
mid05tor12cn05: Activating firmware update. 10%
mid05tor12cn05: Firmware update successfully activated
[root@briggs01 ~]# rflash mid05tor12cn05 -l 
mid05tor12cn05: ID       Purpose State      Version
mid05tor12cn05: -------------------------------------------------------
mid05tor12cn05: 9e55358e BMC     Ready      ibm-v1.99.10-0-r11-0-g9c65260
mid05tor12cn05: 221d9020 Host    Active     IBM-witherspoon-redbud-ibm-OP9_v1.19_1.33
mid05tor12cn05: 776f6a74 Host    Active(*)  IBM-witherspoon-ibm-OP9_v1.19_1.52
mid05tor12cn05: 0        BMC     Active(*)  ibm-v1.99.10-0-r7-0-gcce0f1b
mid05tor12cn05: 
```

1. We can loop for each ID and activate/delete (or block this code from running) 
[...]

Issue #4055, opened 2017-10-05
Need to change rflash to scale by uploading firmware in parallel  


whowutwut:
When flashing two nodes to 1740A, one is at 1738A and one is at 1740A already...  I start to see some issues. 

```
[root@briggs01 ~]# rflash gh4048 -l
mid05tor12cn13: ID       Purpose State      Version
mid05tor12cn13: -------------------------------------------------------
mid05tor12cn13: 9e55358e BMC     Active(*)  ibm-v1.99.10-0-r11-0-g9c65260
mid05tor12cn13: 221d9020 Host    Active(*)  IBM-witherspoon-redbud-ibm-OP9_v1.19_1.33
mid05tor12cn13: 7a9ef57c Host    Ready      IBM-witherspoon-sequoia-ibm-OP9_v1.19_1.35
mid05tor12cn13: f7f626b8 Host    Active     IBM-witherspoon-sequoia-ibm-OP9_v1.19_1.37
mid05tor12cn13: 3ab50d91 BMC     Active     ibm-v1.99.10-0-r7-0-gcce0f1b
mid05tor12cn13: 
mid05tor12cn11: ID       Purpose State      Version
mid05tor12cn11: -------------------------------------------------------
mid05tor12cn11: 221d9020 Host    Active(*)  IBM-witherspoon-redbud-ibm-OP9_v1.19_1.33
mid05tor12cn11: 3ab50d91 BMC     Active(*)  ibm-v1.99.10-0-r7-0-gcce0f1b
mid05tor12cn11: 
```

### Issue 1 

The upload is being done sequentially

```
[root@briggs01 ~]# rflash gh4048 -a /mnt/xcat/iso/openbmc/910.1740.20171003b_1740A/obmc-phosphor-image-witherspoon.ubi.mtd.tar
mid05tor12cn13: Uploading /mnt/xcat/iso/openbmc/910.1740.20171003b_1740A/obmc-phosphor-image-witherspoon.ubi.mtd.tar ...
mid05tor12cn13: Successful, use -l option to list.
mid05tor12cn11: Uploading /mnt/xcat/iso/openbmc/910.1740.20171003b_1740A/obmc-phosphor-image-witherspoon.ubi.mtd.tar ...
mid05tor12cn11: Successful, use -l option to list.
```

It's not until Successful comes back where we see mid05tor12cn11 start the upload...

whowutwut:
@chenglch could you help take a look at implementing this?

[...]

PR #4097, opened 2017-10-13
Enhance openbmc framework with multi-processes

chenglch:
This patch integrate the process worker with the state machine
to support upload firmware in parallel.

partial-issue: #4055

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -29,6 +29,7 @@ use xCAT::Usage;
 use xCAT::SvrUtils;
 use xCAT::GlobalDef;
 use xCAT_monitoring::monitorctrl;
+use POSIX qw(WNOHANG);
 
 $::VERBOSE                  = 0;
 # String constants for rbeacon states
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -2257,10 +2244,57 @@ sub rflash_response {
...
+        if ($node_info{$node}{method} || $status_info{ $node_info{$node}{cur_status} }{method}) {
+            gen_send_request($node);
+        }
     } else {
         $wait_node_num--;
     }
     return;
 }
+
+sub rflash_upload {
```
zet809: I think we shall deal with the *state machine* in this subroutine. Both "RFLASH_FILE_UPLOAD_REQUEST" and "RFLASH_FILE_UPLOAD_RESPONSE".

chenglch: This is just a subroutine running in the child process as I'd like to split the subroutine into small parts especially with multi-processes. It's impossible to put the `state` transformation in the child process as the main process is responsible for maintaining of the state machine.  The change of the state in the child process could not reflect to the main process.

zet809:
Looks great to me, I agree to merge, thx @chenglch

whowutwut:
Also resolves: #4053

PR #4101, opened 2017-10-13
Add netboot osimage to csm rpm to easily install the correct packages for compute node

whowutwut:
I'm not sure if this will really be leveraged in the final product, but internally as I am provisioning compute nodes, I am going to start using this rpm as a way to quickly bring up a base image without having to create osimage individually on new management nodes diskless to deploy compute nodes with correct pkglist for CORAL


whowutwut:
Since this is really only for dev use right now and RPM is optional, I don't think there's any risk in merging so I can continue to make small changes as I use the RPM.

cxhong:
Merged

PR #4102, opened 2017-10-13
Fix erroneous 'noping' in various circumstances

jjohnson42:
nmap's raw IP facility is prone to mistakes in various
environments.  Disable the raw IP handling by
telling nmap to run as if it didn't have privilege
Note this was already done back in:

commit fb02c18853dd2004ed936c77afabd310d7f115dd
Author: jjohnson2 <jjohnson2@lenovo.com>
Date:   Fri Mar 20 10:58:26 2015 -0400

But it was broken since then..


immarvin:
i am ok with the change

PR #4103, opened 2017-10-13
Add pkglist sample file specific for cuda-7.9 in xCAT-csm rpm

whowutwut:
Ship a sample pkglist file for CUDA to easily create a starting point osimage, cuda repo needs to be installed first.

### DIFF
--- a/xCAT-csm/netboot.rh.cuda/cuda-7.9.otherpkgs.pkglist
+++ b/xCAT-csm/netboot.rh.cuda/cuda-7.9.otherpkgs.pkglist
@@ -1,0 +1,2 @@
+cuda-7.9/ppc64le/cuda-deps/dkms
+cuda-7.9/ppc64le/cuda-core/cuda
--- a/xCAT-csm/xCAT-csm.spec
+++ b/xCAT-csm/xCAT-csm.spec
@@ -33,11 +33,14 @@ xCAT-csm provides Packages for installation of CSM nodes
 rm -rf %{buildroot}
 
 mkdir -p $RPM_BUILD_ROOT/install/postscripts/
+mkdir -p $RPM_BUILD_ROOT/install/custom/netboot/rh/
+mkdir -p $RPM_BUILD_ROOT/install/custom/netboot/rh/cuda/
 mkdir -p $RPM_BUILD_ROOT/%{prefix}/share/xcat/install/rh/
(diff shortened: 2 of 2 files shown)


immarvin:
hi @whowutwut , by traditional design, `/install/custom/` is a customized directory created by user to place some customized pkglist,otherpkglist, etc, xCAT won't ship any file under this directory bu default.
what about place the pkglist under `/opt/xcat/share/xcat/netboot/rh/`?  in fact, we have already shipped several stuff like this : 
````
[root@c910f03c05k21 xCAT-csm]# find /opt/xcat/share/xcat/netboot/rh/ -name "*otherpkg*"
/opt/xcat/share/xcat/netboot/rh/service.rhels7.ppc64le.otherpkgs.pkglist
/opt/xcat/share/xcat/netboot/rh/cudafull.rhels7.ppc64le.otherpkgs.pkglist
/opt/xcat/share/xcat/netboot/rh/cudaruntime.rhels7.ppc64le.otherpkgs.pkglist
/opt/xcat/share/xcat/netboot/rh/service.rhels5.ppc64.otherpkgs.pkglist
/opt/xcat/share/xcat/netboot/rh/service.rhels7.x86_64.otherpkgs.pkglist
/opt/xcat/share/xcat/netboot/rh/service.rhels5.x86.otherpkgs.pkglist
/opt/xcat/share/xcat/netboot/rh/service.rhels6.ppc64.otherpkgs.pkglist
/opt/xcat/share/xcat/netboot/rh/service.rhels5.x86_64.otherpkgs.pkglist
````

whowutwut:
@immarvin Thanks for those comments.  

Let me think about this some more... 

xCAT-csm is an optional package and I'm not sure that it will be actually used to deploy... but currently as we re-build our own xCAT dev environment, we either have to manage our own osimage, or put it into some "source control" and I was just leveraging this rpm for now as a base where if xCAT is freshly installed, we could more quickly create the osimage required to have mlnx and cuda installed for the diskless compute nodes... 

Perhaps it's not the right place to put this...  I don't know that we want to build into xCAT default templates version specific pkglist files...    It's not rhels7 specific but references cuda 7.9 version... maybe this could be a variable.   Will think some more here..

immarvin:
hi @whowutwut , is it ok to close this?

whowutwut:
Yea,  let's close, thanks @immarvin

Issue #4138, opened 2017-10-20
the node status cannot be updated on normal system reboot


immarvin:
this issue is reported by user, when they provisioned a redhat7 diskful node,  then reboot the node on mn with `rpower <node> off` and `rpower <node> on`, the node status keep "powering-on" even after the node is booted

immarvin:
hi @hu-weihua , we might need a test case to cover this.

PR #4139, opened 2017-10-20
fix issue the node status cannot be updated on normal system reboot #4138

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/4138

UT verified on redhat7 on ppc64


bybai:
I am ok to merge it.

PR #4141, opened 2017-10-20
remove the redundancy codes to avoid report mac missing warning  twice

### LINKED ISSUE #4023: [DEV] makedhcp will report duplicate warning when no mac defined for this node
It reproduced when generate the dhcp lease for a node without MAC

```
makedhcp fake3 -V
Warning: [c910f03c05k20]: Unable to find mac address for fake3
Warning: [c910f03c05k20]: Unable to find mac address for fake3
```

it might be caused when generate the lease info, it delete and re-add it.

robin2008:
Fix #4023, remove the redundancy codes to report mac missing warning, and adjust message line make it better show in IDE

UT: after patched, and remove the mac attribute then run makedhcp, only one warning now.
```
makedhcp fake3
Warning: Unable to find mac address for fake3
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -627,7 +627,7 @@ sub addnode
             $callback->(
                 {
                     warning => [
-"The hostname $hname of node $node could not be resolved."
+                        "The hostname $hname of node $node could not be resolved."
                       ]
                 }
             );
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/dhcp.pm
```diff
@@ -2013,6 +2013,16 @@ sub process_request
         }
         my $mactab = xCAT::Table->new('mac');
         $machash = $mactab->getNodesAttribs($req->{node}, ['mac']);
+        unless ($opt{n} or $machash)
+        {
+            $callback->(
+                {
+                    error => ["Unable to open mac table, it may not exist yet"],
```
immarvin: suggest the message to be `Unable to get mac address of $req->{node} in mac table, please make sure it is specified`

immarvin:
i am ok with it

PR #4144, opened 2017-10-20
Modify setupntp postscripts

cxhong:
Enhance PR #4049 , checking ntp server before starts `ntpd` command.
`ntpd -gq` command will be hung if ntpserver is not reachable.

### DIFF
--- a/xCAT/postscripts/setupntp
+++ b/xCAT/postscripts/setupntp
@@ -6,8 +6,6 @@
 #
 #---------------------------------------------------------------------------
 
-
-
 if [ "$(uname -s|tr 'A-Z' 'a-z')" = "linux" ];then
    str_dir_name=`dirname $0`
    . $str_dir_name/xcatlib.sh
@@ -25,7 +23,6 @@ fi
 
 master=$MASTER
 setup=0
-sitemaster=$SITEMASTER
 conf_file="/etc/ntp.conf"
 conf_file_org="/etc/ntp.conf.org"
 conf_file_backup="/etc/ntp.conf.postbackup"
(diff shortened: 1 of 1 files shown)


xCAT/postscripts/setupntp
```diff
@@ -66,20 +63,32 @@ if [ $NTPSERVERS ]; then
...
         done 
+        if [ $setup -eq 0 ]; then
+	    echo "server $master" >$conf_file
+        fi
     fi
 else
 	echo "server $master" >$conf_file
 fi
 
+if ! ping $master -c 1 > /dev/null 2>&1 ; then
```
mattaezell: Maybe move this down to line 117.  Even if the master isn't listening yet or you have the wrong IP address for the master, it's important to leave the system with a (mostly) valid ntp config file.  The `driftfile`, `restrict`, and `interface` settings below still need to go in the file in case the admin manually fixes the master IP address later and tries to start NTP back up.

immarvin: agree with @mattaezell , in most cases,  `setupntp` only generate correct ntp config files, this script does not guarantee the ntp is synced

xCAT/postscripts/setupntp
```diff
@@ -99,14 +103,24 @@ if [ $OS_TYPE = Linux ]; then
         echo "interface listen eth0" >>$conf_file
     fi
 
+    # ntpd will be hung if ntp service is not reachable
+    if ! ping $master -c 1 > /dev/null 2>&1 ; then
+        echo "Error: ntpserver $master is not reachable, will not setup NTP"
+        exit 1
```
immarvin: should it exit here? what about the configuration after this line?

immarvin:
discussed with @cxhong offline, agree to merge this PR first, will take a look at whether the configuration after `ntpd -qg` depends on the result of `ntpd -qg`

PR #4156, opened 2017-10-23
Manage /var/log/xcat/rflash permissions

mattaezell:
Set /var/log/xcat/rflash to `0700`


mattaezell:
While we only intend to have trusted users on our management node, this file will contain our IPMI password and we should attempt to protect it.

PR #4171, opened 2017-10-24
Sort nodels attribute names

jjohnson42:
The ordering of attribute names out of nodels
is inconsistent.  Use sort to provide consistent
output


chenglch:
Close/reopen for CI to run again..

Issue #3955, opened 2017-09-19
Stateless SN enhancement


robin2008:
- [x] 1, Add the RTD for stateless SN
- [x] 2, localdisk support sync option to allow the sync the required contents to local directory at the first time.

robin2008:
Task 2 is done in https://github.com/xcat2/xcat-core/pull/4152

robin2008:
Task1 is covered in #4174

PR #4174, opened 2017-10-25
Stateless service node RTD update

robin2008:
For #3955 

1, the previous document is using rhel6.4 ppc64 as example, now replaced with rhel7.3 ppc64le, and update the steps to reflect the new xCAT release.

2,add the `localdisk` option here
  - remove the limitation to sync the directory manually from osimage rootfs
  - using the right file `source/guides/admin-guides/manage_clusters/ppc64le/diskless/customize_image/localdisk.rst`, and remove the old one.
  - Fix a typo `getpartitioin` there

For new generate document, you could access below link:
http://10.3.5.29/install/html/advanced/hierarchy/provision/diskless_sn.html


robin2008:
document is updated.

zet809:
looks good to me, I agree to merge, thx @robin2008

Issue #4172, opened 2017-10-24
Reduce the amount of logs generated into the commands.log file for console


whowutwut:
Each node is creating a entry into the commands.log file for console and will quickly fill up. 

```
====================================================
[Date]       2017-10-23 03:34:02
[ClientType] cli
[Request]    getopenbmccons mid05tor12cn08 text
[Response]
mid05tor12cn08: openbmc, getopenbmccoms
mid05tor12cn08
[ElapsedTime] 0 s
```

* Is there a way to reduce this by? 
* Why do we need this command to get logged, what benefits does it provide?

hu-weihua:
@whowutwut, When design ``commands.log `` long time ago,  the purpose is to log the output of all command from CLI or Restapi.  But due to xcat itself may trigger some command which also send to xcatd like CLI during it's lift cycle, there is no flag to let ``commands.log `` to distinguish which command really come from CLI and which come from xcat itself.  So  ``commands.log `` log almost all of them.   But when I implemented ``commands.log ``, I have noticed some command logs are  useless but with high rate， I have filtered them out, such like below

```
   3507     $tmpreq =~ s/\[Request\]\s+(.+)/$1/g;
   3508     if ($tmpreq =~ /getipmicons/) { return 1; }
   3509     if ($tmpreq =~ /getcons/)     { return 1; }
```
``getopenbmccons`` was developed later than ``commands.log``, so it was not added into filter list.   If need, we can add it.

hu-weihua:
@whowutwut  @zet809 @chenglch,  do you think we should add ``getopenbmccons`` into filter list as ``getipmicons``? If we agree that, I will add it ASAP. thanks.

mattaezell:
@hu-weihua 👍 Please!

whowutwut:
@hu-weihua Yes, please add it into the filter list, I don't think there's any value to that log message!

hu-weihua:
@whowutwut, the fix has been merged. Could you try again? thanks

whowutwut:
@hu-weihua  Updated to todays build and the logs are no longer there in commands.log. This is good 👍 

```
[root@briggs01 xcat]# lsxcatd -v
Version 2.13.8 (git commit 2353abb48a6ea72affc2ecacafb353bec02b3d79, built Wed Oct 25 06:15:45 EDT 2017)
[root@briggs01 xcat]#
[root@briggs01 xcat]# date
Wed Oct 25 11:23:41 EDT 2017
[...]

PR #4176, opened 2017-10-25
Add getopenbmccons to commands.log 's filter list

hu-weihua:
For issue #4172 
Due to ``getopenbmccons`` will generate amount of useless log in ``commands.log``, to filter its output out.


zet809:
OK, Looks good to me, I agree to merge. Thx @hu-weihua

Issue #3974, opened 2017-09-20
Power9 NVME Bolt adapter update 


markprez:
The NVME bolt cards on the witherspoon power9 will require a BIOS update. This is a request to provide a method through xCAT to update the firmware. I will update this issue with the process from Matt Scheckel.

whowutwut:
I've also been looking into this based on the information here: http://ausgsa.ibm.com/projects/s/scsi_sas/de/bolt/bolt.htm

immarvin:
@whowutwut @zet809 can this one be closed?

zet809:
Let me close this task as the code have been merged in.

PR #4184, opened 2017-10-26
add the script to update nvme firmware

zet809:
For task #3974 
The script can be used to update nvme firmware:

```
[root@briggs01 postscripts]# updatenode mid05tor12cn17 -P "update_nvme_firmware /install/MN12MN12.img"
mid05tor12cn17: xcatdsklspost: downloaded postscripts successfully
mid05tor12cn17: Thu Oct 26 08:53:56 UTC 2017 Running postscript: update_nvme_firmware
mid05tor12cn17: ==> NVMe Firmware File PATH: /install/MN12MN12.img
mid05tor12cn17: ==> NVMe Firmware File: MN12MN12.img
mid05tor12cn17: ==> Retrieving file from http://172.12.253.27//install/MN12MN12.img
mid05tor12cn17: -rw-r--r-- 1 root root 3198976 Oct 18 09:18 //MN12MN12.img
mid05tor12cn17: Firmware download success
mid05tor12cn17: Success activating firmware action:1 slot:1
mid05tor12cn17: NVME Admin command error:FIRMWARE_IMAGE(107)
mid05tor12cn17: NVME Admin command error:FIRMWARE_IMAGE(107)
mid05tor12cn17: NVME Admin command error:FIRMWARE_SLOT(106)
mid05tor12cn17: Update Succeed to MN12MN12
mid05tor12cn17: postscript: update_nvme_firmware exited with code 0
mid05tor12cn17: Running of postscripts has completed.
```

### DIFF
--- a/xCAT/postscripts/update_nvme_firmware
+++ b/xCAT/postscripts/update_nvme_firmware
@@ -1,0 +1,81 @@
+#!/bin/sh
+# 
+# If NVMe firmware file is located here: 
+# /install/MN12MN12.img
+#
+# Call the script as follows to update firmware for device /dev/nvme0: 
+#    updatenode <noderange> -P "update_nvme_firmware /install/MN12MN12.img [/dev/nvme0]"
+#
+
+if [[ -z ${1} ]]; then 
+   echo "Error: you must provide the path of the NVMe Firmware file" 
+   exit 1
+fi
+NVME_FILE_PATH=${1}
+if ! which nvme > /dev/null; then 
+    echo "Error: No nvme tool find, pls install nvme-cli first"
+    exit 1
+fi
+if [[ -z ${2} ]]; then
+    if ! ls /dev/nvme* 2>1 >/dev/nul; then 
+        echo "No NVMe device find"; 
+        exit 1
+    else 
+        TARGET_DEV=`ls /dev/nvme*| egrep "nvme[0-9]$"`
+    fi
+else
+    TARGET_DEV=${2}
+    if ! ls $TARGET_DEV 2>1 >/dev/null; then 
+        echo "Not find nvme device $TARGET_DEV"; 
+        exit 1
+    fi
+fi
+
+WORKING_DIR="/"
+NVME_FIRM_FILE=`basename ${NVME_FILE_PATH}`
+TARGET_FIRM_FILE="${WORKING_DIR}/${NVME_FIRM_FILE}"
+
+echo "==> NVMe Firmware File PATH: ${NVME_FILE_PATH}"
+echo "==> NVMe Firmware File: ${NVME_FIRM_FILE}" 
+
+if [[ -e ${TARGET_FIRM_FILE} ]]; then 
+   rm -f ${TARGET_FIRM_FILE}
+fi 
+
+echo "==> Retrieving file from http://${MASTER}/${NVME_FILE_PATH}"
+wget -q http://${MASTER}/${NVME_FILE_PATH} -O ${TARGET_FIRM_FILE}
+if [[ $? ]]; then
+    echo "Error: Download ${NVME_FILE_PATH} failed from ${MASTER}"
+    exit 1
+fi
+
+ls -ltr ${TARGET_FIRM_FILE}
+
+if ! nvme fw-download $TARGET_DEV --fw=${TARGET_FIRM_FILE}; then
+    echo "Firmware download to NVMe device $TARGET_DEV failed"
+    exit 1
+fi
+
+nvme fw-activate $TARGET_DEV -a 1 -s 1
+nvme fw-activate $TARGET_DEV -a 1 -s 2
+nvme fw-activate $TARGET_DEV -a 1 -s 3
+nvme fw-activate $TARGET_DEV -a 1 -s 4
+
+nvme reset $TARGET_DEV
+
+TARGET_VERSION=`nvme id-ctrl $TARGET_DEV | grep "fr " | awk '{ print $3}'`
+
+RC=0
+if [[ x"$NVME_FIRM_FILE" == x"${TARGET_VERSION}.img" ]]; then
+    echo "Update Succeed to $TARGET_VERSION"
+else
+    echo "Update failed, the current version is $TARGET_VERSION"
+    RC=1
+fi
+#
+# Clean up
+#
+sleep 1
+rm -f ${TARGET_FIRM_FILE}
+
+exit $RC
--- a/xCAT/postscripts/update_nvme_firmware
+++ b/xCAT/postscripts/update_nvme_firmware
@@ -6,32 +6,36 @@
 # Call the script as follows to update firmware for device /dev/nvme0: 
 #    updatenode <noderange> -P "update_nvme_firmware /install/MN12MN12.img [/dev/nvme0]"
 #
-
+function usage() {
+    echo "Usage: $0 </path_to/NVME.img> [nvme device]"
+}
 if [[ -z ${1} ]]; then 
    echo "Error: you must provide the path of the NVMe Firmware file" 
+   usage
    exit 1
 fi
 NVME_FILE_PATH=${1}
 if ! which nvme > /dev/null; then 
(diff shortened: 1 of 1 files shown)


zet809:
hi, @whowutwut , shall we merge this script first and include it in the 2.13.8 release?

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+# If NVMe firmware file is located here: 
+# /install/MN12MN12.img
+#
+# Call the script as follows to update firmware for device /dev/nvme0: 
+#    updatenode <noderange> -P "update_nvme_firmware /install/MN12MN12.img [/dev/nvme0]"
+#
+
+if [[ -z ${1} ]]; then 
+   echo "Error: you must provide the path of the NVMe Firmware file" 
+   exit 1
```
whowutwut: To help user,  maybe create a usage function: 
```
function usage {
    echo "Usage: $0 </path_to/NVME.img> [nvme device]"
}
```

Then call here before `exit 1`

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+fi
+NVME_FILE_PATH=${1}
+if ! which nvme > /dev/null; then 
+    echo "Error: No nvme tool find, pls install nvme-cli first"
```
whowutwut: Suggest: 

`Error: nvme utility not found, install 'nvme-cli' package and retry.`

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+    exit 1
+fi
+if [[ -z ${2} ]]; then
+    if ! ls /dev/nvme* 2>1 >/dev/nul; then
```
whowutwut: Spelling issue with `/dev/null`   

Also, should we sync up the if clause with brackets or not brackets.  I'm not sure what to do here.. we list `/bin/sh` and on RHEL systems, that redirects to bash.   So not sure...

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+   exit 1
+fi
+NVME_FILE_PATH=${1}
+if ! which nvme > /dev/null; then 
+    echo "Error: No nvme tool find, pls install nvme-cli first"
+    exit 1
+fi
+if [[ -z ${2} ]]; then
+    if ! ls /dev/nvme* 2>1 >/dev/nul; then 
+        echo "No NVMe device find";
```
whowutwut: If we want to support running this script on ANY node, regardless of whether it has NVME or not, maybe we don't want to return 1, because then the user will get an error in their postscript. 

Suggest: 

```
echo "INFO: No NVMe devices found, nothing to do." 
exit 0
```

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+if ! which nvme > /dev/null; then 
+    echo "Error: No nvme tool find, pls install nvme-cli first"
+    exit 1
+fi
+if [[ -z ${2} ]]; then
+    if ! ls /dev/nvme* 2>1 >/dev/nul; then 
+        echo "No NVMe device find"; 
+        exit 1
+    else 
+        TARGET_DEV=`ls /dev/nvme*| egrep "nvme[0-9]$"`
```
whowutwut: space after `*` for better readability.

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+    fi
+else
+    TARGET_DEV=${2}
+    if ! ls $TARGET_DEV 2>1 >/dev/null; then 
+        echo "Not find nvme device $TARGET_DEV";
```
whowutwut: * This one should indicate Error in the message
*  no need semicolon at end

`Error: Could not find nvme device specified: $TARGET_DEV`

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+    fi
+else
+    TARGET_DEV=${2}
+    if ! ls $TARGET_DEV 2>1 >/dev/null; then 
+        echo "Not find nvme device $TARGET_DEV"; 
+        exit 1
+    fi
+fi
+
+WORKING_DIR="/"
```
whowutwut: I don't think we should ever use `/` as the working dir, we should put in `/tmp` or even something to identify that it was xCAT postscript that created it... . 

If we do the following, need to create the dir... 
```
WORKING_DIR=`/tmp/xcat.nvme` 
/usr/bin/mkdir -p ${WORKING_DIR}
```

The danger if we have an error in the script and WORKING_DIR is `/`, and we do `rm -rf $WORKING_DIR` it would wipe out the entire filesystem.

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+echo "==> Retrieving file from http://${MASTER}/${NVME_FILE_PATH}"
+wget -q http://${MASTER}/${NVME_FILE_PATH} -O ${TARGET_FIRM_FILE}
+if [[ $? ]]; then
+    echo "Error: Download ${NVME_FILE_PATH} failed from ${MASTER}"
+    exit 1
+fi
+
+ls -ltr ${TARGET_FIRM_FILE}
+
+if ! nvme fw-download $TARGET_DEV --fw=${TARGET_FIRM_FILE}; then
```
whowutwut: Does this work if we have multiple TARGET_DEV?    I think we need this in a loop....

Up above, you do `ls /dev/nvme*| egrep "nvme[0-9]$` , wouldn't we need something like....
```
for DEV in $TARGET_DEV; do
    echo $DEV
done
```

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+nvme fw-activate $TARGET_DEV -a 1 -s 3
+nvme fw-activate $TARGET_DEV -a 1 -s 4
+
+nvme reset $TARGET_DEV
+
+TARGET_VERSION=`nvme id-ctrl $TARGET_DEV | grep "fr " | awk '{ print $3}'`
+
+RC=0
+if [[ x"$NVME_FIRM_FILE" == x"${TARGET_VERSION}.img" ]]; then
+    echo "Update Succeed to $TARGET_VERSION"
```
whowutwut: Here this needs to be in a loop and we should print something like....

`${DEV}: Firmware successfully updated to ${TARGET_VERSION}`

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+else
+    echo "Update failed, the current version is $TARGET_VERSION"
```
whowutwut: This should also be in loop and should be like: 

`${DEV}: Firmware update failed, current version is ${TARGET_VERSION}`

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+    RC=1
+fi
+#
+# Clean up
+#
+sleep 1
```
whowutwut: What's the purpose of the sleep here?

xCAT/postscripts/update_nvme_firmware
```diff
@@ -0,0 +1,81 @@
...
+rm -f ${TARGET_FIRM_FILE}
```
whowutwut: If we create a dir, then also add `rmdir ${WORKING_DIR}`

xCAT/postscripts/update_nvme_firmware
zet809: Good point, modify it. Thx!

zet809:
Hi, @whowutwut, updated based on your comments, pls help to review again. Thx!

PR #4188, opened 2017-10-26
z/VM IVP should tolerate slow xcatd startup.

StevenGessner:
The IVP expects xcatd to be fully operational by the
time is begins a run.  It is possible for xcatd to
be delayed.  The delay should not cause a corruption
of the zvmnotify property in the site table.
In addition, tests for tabdump failures should be
corrected and the log file should be set up sooner
so that debug information is available for similar
problems.

Change-Id: I2b0e3e875cc28f740a35b348faa0cc74f5acb7bf


zet809:
Hi, @chuckbrazie , will you pls review this PR ASAP. We plan to freeze code merge from the end of this week for 2.13.8 release. Thx!

Issue #4169, opened 2017-10-24
Diskfull installation does not contain NODE= in /opt/xcat/xcatinfo


tgooding:
I tried a diskfull node installation and the resulting installation produced an `/opt/xcat/xcatinfo` file that did not contain this line:
```
    NODE=<hostname>
```

We have other pieces of software are depending on xcatinfo for a consistent node name to reference back to xCAT.  

Running xCAT release `snap201710110615`.

Let me know if you need other info from the system.  Thanks!  Tom

whowutwut:
@cxhong Were you going to fix this one?

whowutwut:
Closing this out as it was merged into 2.13.8, @tgooding please open a new issue if you see problems after upgrading to the latest build.

PR #4190, opened 2017-10-26
Add Node name to the /opt/xcat/xcatinfo for diskfull installation

cxhong:
for issue #4169 

After diskfull node provision, file /opt/xcat/xcatinfo will contain Node name
```
# cat /opt/xcat/xcatinfo
XCATSERVER=172.12.253.27
INSTALLDIR=/install
REBOOT=TRUE
NODE=mid05tor12cn18
````


immarvin:
i am ok with this change

Issue #4192, opened 2017-10-26
[OpenBMC] rpower (or framework) calls is not sent from xCAT in parallel 


whowutwut:
I think this is a fundamental problem in the framework, @chenglch did some improvements for rflash, I think @xuweibj has some data about the performance of the calls in the OpenBMC perl module. 

With changes I made to #4187 I'm able to show the issues that I have been seeing.  It seems to spawn when the BMC connection is not good.  But because of this, the framework blocks sending all the rest calls to the BMCs, I think this should NOT block. 

Here's an example, I reboot the BMC, so that it won't respond correctly to force this situation: 
```
[root@briggs01 openbmctoolfordist]# time rpower mid05tor12cn[02,11] bmcreboot
Thu Oct 26 16:50:27 2017 mid05tor12cn11: [openbmc_debug] curl -k -c cjar -H "Content-Type: application/json" -d '{ "data": ["root", "xxxxxx"] }' https://172.11.139.11/login
Thu Oct 26 16:50:28 2017 mid05tor12cn02: [openbmc_debug] curl -k -c cjar -H "Content-Type: application/json" -d '{ "data": ["root", "xxxxxx"] }' https://172.11.139.2/login
Thu Oct 26 16:50:28 2017 mid05tor12cn11: [openbmc_debug] login_response 200 OK
Thu Oct 26 16:50:28 2017 mid05tor12cn11: [openbmc_debug] curl -k -b cjar -X PUT -H "Content-Type: application/json" -d '{"data":"xyz.openbmc_project.State.BMC.Transition.Reboot"}' https://172.11.139.11/xyz/openbmc_project/state/bmc0/attr/RequestedBMCTransition
Thu Oct 26 16:50:28 2017 mid05tor12cn02: [openbmc_debug] login_response 200 OK
Thu Oct 26 16:50:28 2017 mid05tor12cn02: [openbmc_debug] curl -k -b cjar -X PUT -H "Content-Type: application/json" -d '{"data":"xyz.openbmc_project.State.BMC.Transition.Reboot"}' https://172.11.139.2/xyz/openbmc_project/state/bmc0/attr/RequestedBMCTransition
Thu Oct 26 16:50:29 2017 mid05tor12cn11: [openbmc_debug] rpower_reset_response 200 OK
mid05tor12cn11: BMC reboot
Thu Oct 26 16:50:29 2017 mid05tor12cn02: [openbmc_debug] rpower_reset_response 200 OK
[...]

PR #4207, opened 2017-10-31
Fix issue 4192, fork to login before send real login request

xuweibj:
#4192 
Fork process to send ``curl`` command to login, if login failed or timeout(10s), will not run other command for that node.

Output:
```
# rpower simulation_test stat
simulation_test: Error: 503 Service Unavailable
```

Output when xcatdebugmode is 1:
```
# rpower simulation_test stat
Tue Oct 31 02:06:17 2017 simulation_test: [openbmc_debug] LOGIN Failed using curl command
simulation_test: Error: 503 Service Unavailable
```

* Performance test:

MN: briggs01
CN: simulator run on mid05tor12cn13
Command: ``rpower status``

500 nodes (all not available and wait for timeout): 1m24s
500 nodes (all available) : 25s

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -386,6 +386,8 @@ my $xcatdebugmode = 0;
 
 my $flag_debug = "[openbmc_debug]";
 
+my %login_pid_node; # used in process_request, record login fork pid map
+
 #-------------------------------------------------------
 
 =head3  preprocess_request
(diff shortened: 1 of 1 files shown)


zet809:
Looks good, I agree to merge. Thx @xuweibj

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -532,37 +534,68 @@ sub process_request {
...
+                my $node = shift @nodes_login;
+                my $rst_fork = fork_process_login($node);
+                $child_num++ unless($rst_fork);
+            }
+        }
+    }
+
+    foreach my $node (keys %node_info) {
+        if (!$valid_nodes{$node}) {
+            xCAT::SvrUtils::sendmsg([1, $::RESPONSE_SERVICE_UNAVAILABLE], $callback, $node);
```
whowutwut: @xuweibj Should have caught this one but fell behind on code review, we need to change this so we are not sending back a REST issue unless it's directly coming from the BMC.  Here we are deciding what is valid or not, so making it look like a REST API error is misleading.

PR #4212, opened 2017-11-01
Quiet some debug messages during discovery

mattaezell:
When there are a large number of nodes and switches, the discovery process can be quite verbose.  This commit changes two messages to trace by default (so they show up if xcatdebugmode is set).  It also adds the word 'switch' to a couple messages to make it more clear.

### DIFF
--- a/perl-xCAT/xCAT/MacMap.pm
+++ b/perl-xCAT/xCAT/MacMap.pm
@@ -512,7 +512,7 @@ sub refresh_table {
         my $ntype = $typehash->{$entry->{node}}->[0]->{nodetype};
         if ( (($discover_switch) and ( $ntype ne "switch"))
             or ( !($discover_switch) and ( $ntype eq "switch")) ){
-            xCAT::MsgUtils->message("S", "refresh_table: skip $entry->{node} and $entry->{switch}, $discover_switch , $ntype\n");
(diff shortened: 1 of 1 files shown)


mattaezell:
On the second message I changed to trace (refresh_switch elapsed timer), I'm not seeing the debug messages (<= 10 seconds) when `xcatdebugmode` is set (though I am seeing the warning messages when it takes more than 10 seconds).  Is this because `MsgUtils->trace()` needs to talk to the database to see if `xcatdebugmode` is enabled, but it can't access the database in a forked process?  Do I need to check before fork and pass in `$verbose` if `xcatdebugmode` is set?

immarvin:
hi @mattaezell , I do not think you need to check before fork and pass in $verbose if xcatdebugmode is set. why DB cannot be accessed in a forked process?  suggest you add some debug `print` in `xCAT::MsgUtils->trace` to see what happened

zet809:
Hi, @mattaezell , not every findme request will cause the refresh of switch mac table. For the matching MAC address, the timeout of the recording will be 5 minutes. So, it means if you try with the same CN, the next log about refresh_switch will be at least 5 minutes later. I have verified your code with less timeout interval, it works. Thx!

zet809:
So, let me merge this.

PR #4219, opened 2017-11-01
dodiscovery: better disksize - ordered by major,minor and no `\n`

samveen:
* Minor improvement to do `dodiscovery` to generate the disksize field in sorted order (sorted by major,minor numbers of the disks) and remove unnecessary newlines from the field
* Corresponding change in `nodediscover` to remove redundant `disksize` correction code.

### DIFF
--- a/xCAT-genesis-scripts/bin/dodiscovery
+++ b/xCAT-genesis-scripts/bin/dodiscovery
@@ -134,8 +134,8 @@ fi
 # The MEMORY will look like this: 32868920
 MEMORY=`cat /proc/meminfo |grep MemTotal|awk '{printf "%.0fMB\n", $2/1024}'`
 
-# The DISKSIZE will look like this: sdb:250GB,sda:250GB
-DISKSIZE=`cat /proc/partitions |grep -e "sd.\>" |awk -F' ' '{printf "%s:%.0fGB\n", $4, $3*0.000001024}' |sed 'N;s/\n/,/'`
+# The DISKSIZE will look like this: sda:960GB,sdb:960GB,sdc:480GB,sdd:480GB (sorted by major,minor)
+DISKSIZE="$(cat /proc/partitions |sort -n |awk 'BEGIN{sep=""} /sd.\>/{printf("%s%s:%.0fGB", sep, $4, $3*0.000001024) ; sep=","}')"
 
 logger -s -t $log_label -p local4.info "Beginning echo information to discovery packet file..."
 echo '<xcatrequest>' > /tmp/discopacket
--- a/xCAT-genesis-scripts/bin/dodiscovery
+++ b/xCAT-genesis-scripts/bin/dodiscovery
@@ -69,7 +69,7 @@ while [ ! -r /restart ]; do
 
     #the existence of "/processing" indicates that my  findme request is under processing
     if [ -f "/processing"  ]; then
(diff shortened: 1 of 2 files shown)


samveen:
@whowutwut For your kind review (I'm hoping to get this into the `2.13.8` release

immarvin:
i am ok with the change

immarvin:
hi @samveen , thanks for your contribution. We will review this PR ASAP. 

This PR cannot be included in 2.13.8 release, since we are already at the end of xCAT 2.13.8 release process this week, the code is in invite mode, only the critical fix can be merged, sorry for that.

xCAT-genesis-scripts/bin/dodiscovery
```diff
@@ -134,8 +134,8 @@ fi
 # The MEMORY will look like this: 32868920
 MEMORY=`cat /proc/meminfo |grep MemTotal|awk '{printf "%.0fMB\n", $2/1024}'`
 
-# The DISKSIZE will look like this: sdb:250GB,sda:250GB
-DISKSIZE=`cat /proc/partitions |grep -e "sd.\>" |awk -F' ' '{printf "%s:%.0fGB\n", $4, $3*0.000001024}' |sed 'N;s/\n/,/'`
+# The DISKSIZE will look like this: sda:960GB,sdb:960GB,sdc:480GB,sdd:480GB (sorted by major,minor)
+DISKSIZE="$(cat /proc/partitions |sort -n |awk 'BEGIN{sep=""} /sd.\>/{printf("%s%s:%.0fGB", sep, $4, $3*0.000001024) ; sep=","}')"
```
immarvin: `/sd.\>/` , should `\>` be `$` here?

neo954: I prefer use `sort -g -k 1,2`, which is straight forward.

And the `/sd.\>/` part in the `awk` script looks not good. What if there are more than 26 disks, or more than 26 + 26 * 26 =702 disks? I prefer use `sd[a-z]+`

Plus, I believe `$3*0.000001024` did not do the math correctly. The correct math is `$3/2^20`.

Thus, I prefer use ...
```
sort -g -k 1,2 </proc/partitions |awk 'BEGIN{sep=""} /sd[a-z]+$/ {printf("%s%s:%.0fGB", sep, $4, $3/2^20) ; sep=","}')"
```

samveen: @neo954 I agree that this is weird: However would you prefer Disk Manufacturer math (powers of 10^3) or computer science math (power of 2^10)?

samveen: @immarvin both work, but differently: however I'm leaning towards `/[^0-9]$/`. can you check against a large enough test group? I dont have access to that large a test group.

samveen:
@neo954 About `0.000001024`, agreed that this is weird: However would you prefer Disk Manufacturer math (powers of 10^3) or Computer Scientist math (powers of 2^10)?

neo954:
@samveen, Technically, when use Disk Manufacturer math (powers of 10^3), the units should be KB, MB, GB. And when use Computer Scientist math (powers of 2^10), the units should be KiB, MiB and GiB.

BTW, I prefer to use powers of two. When I read a disk size in powers of ten, I had to do the conversion in my mind.

immarvin:
i am ok with the change

PR #4223, opened 2017-11-01
Add documentation for xcatsslversion and xcatsslconfig to RTD

whowutwut:
Commit Log: 
> Add a section into RTD that explains to the users how to configure
xcatsslversion and xcatsslconfig.  Currently the only documentation
is in the security bulletins

Content was taken from here: https://sourceforge.net/p/xcat/wiki/SSLv3%20POODLE%20Attack%20CVE-2014-3566/

Recently had to help someone configure xCAT to pass a vulnerability scan and didn't like how the documentation to configure and validate was not somewhere inside xCAT RTD.  Adding it.

### DIFF
--- a/docs/source/advanced/security/index.rst
+++ b/docs/source/advanced/security/index.rst
@@ -6,4 +6,5 @@ The security of a system covers a wide range of elements, from the security of s
 .. toctree::
    :maxdepth: 2
 
-   security
+   ssl_config.rst
+   security.rst
--- a/docs/source/advanced/security/ssl_config.rst
+++ b/docs/source/advanced/security/ssl_config.rst
@@ -1,0 +1,56 @@
+OpenSSL Configuration
+=====================
+
+xCAT does not ship OpenSSL RPMS nor does it statically link to any OpenSSL libraries.  Communication between the xCAT client and daemon utilizes OpenSSL and the administrator can configure SSL_version and SSL_cipher that should be used by xCAT daemons.
+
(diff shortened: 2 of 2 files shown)


whowutwut:
No plans to address in this pull request, but seems like with the vulnerabilities recently in OpenSSL,  should we consider configuration of xCAT to have stricter configuration by **default**?   

TLSv1.2 and disable SSLv2 and SSLv3 as acceptable communication methods.  @immarvin @zet809 Any implications here?

daniceexi:
I am OK with changing default ssl cfg to be more strict. But the doc also need to be kept up for user to set. A risk here is the old OS support for the strict ssl cfg like rh 7.1 or even rh 6.x.

robin2008:
Do we already have auto-detect method for the OS version and give the suitable default value there?

whowutwut:
@daniceexi we have version control in our documentation, so this one is going into 2.13.8  and will be in the doc moving forward..  we should leverage this to keep our documentation in sync with release ..  but I'm not planning on creating old doc in the older branch at this time, nor am I planning to change that for now, I just wanted to have this doc in the RTD as reference....  We can open another issue if we agree that some change may be better.. and then think about it there. 

Having the configuration steps in a security issue isn't the best place for it.

whowutwut:
I'm not sure why the delay in merging this, but I will merge it since it's documentation on how to configure SSL.  I think it's important to have this as part of the docs...  please open other issues as the scope of this is just to move the documentation into RTD,  not to change any code behavior.

PR #4247, opened 2017-11-06
Remove the "access on" for OpenBMC and use raw command instead

### LINKED ISSUE #4242: [OpenBMC] Change the bmcsetup code back to use raw ipmi commands for "access on"
Revert the `bmcsetup` code back to raw commands from `access on`

I'm not sure the history about why we had changed the raw command back to `access on` , it seemed like it was working at some later build of OpenBMC... 
* We first set it here: https://github.com/xcat2/xcat-core/pull/2990
* Then undid it here: https://github.com/xcat2/xcat-core/pull/3833

But looking at https://github.com/openbmc/openbmc/issues/2492, it seems that certain commands will not take effect without the raw command (Setting IPSRC to DHCP)

whowutwut:
Resolves #4242 

Flip flopping this back to using the raw command since some issues have arisen and seems like even though the command does take effect one some cases, not all of them work without issues. 

There still is an error message that comes out from issuing the raw command: 
```
Unable to send RAW command (channel=0x0 netfn=0x6 lun=0x0 cmd=0x40 rsp=0xce): Command response could not be provided
```

But the changes do take effect...

### DIFF
--- a/xCAT-genesis-scripts/bin/bmcsetup
+++ b/xCAT-genesis-scripts/bin/bmcsetup
@@ -396,7 +396,11 @@ fi
 # update the node status to 'bmcready' for openbmc, no more configuration is needed.
 if [ ! -z "$ISOPENBMC" ]; then
     # To enable network configuration for openbmc
-    ipmitool -d 0 lan set $LANCHAN access on
+    # 
+    # For OpenBMC, FW team still suggest running the raw command instead of access on, use raw for now
+    #
+    # ipmitool -d 0 lan set $LANCHAN access on
+    ipmitool -d 0 raw 0x06 0x40 $LANCHAN 0x42 0x44
     # update the node status to 'bmcready'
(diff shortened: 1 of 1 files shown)


whowutwut:
@zet809 please review this and merge ASAP, some of the issues we are seeing in bmcsetup today is caused by the `access on` vs `raw` command.

zet809:
It is very strange, but let's merge it.

whowutwut:
Here's a patch file: 
[bmcsetup.4242.patch.txt](https://github.com/xcat2/xcat-core/files/1451181/bmcsetup.4242.patch.txt)

To patch your system running 2.13.8 

1.  Download patch file to: `/tmp/bmcsetup.4242.patch.txt`
2. `cd /opt/xcat/share/xcat/netboot/genesis/ppc64/fs/usr/bin`
3. `patch -p0 < /tmp/bmcsetup.4242.patch.txt`

PR #3683, opened 2017-08-14
Add switch_type lookup via MAC for Penguin Computing switches

ryanbowen:
switchdiscover fails to discover Penguin Computing bare-metal switches on current branch due to not being able to identify vendor. This adds Penguin Computing's vendor MAC to the global_mac_identity hash to allow the discovery of Penguin hardware.

PR #4253, opened 2017-11-07
To replace pr 3683: Add switch_type lookup via MAC for Penguin Comput…

zet809:
…ing switches

Replace PR #3683


immarvin:
agree to merge

PR #4255, opened 2017-11-07
Fix the typo in openbmc rflash

chenglch:
Add $ before node to indicate the variable.


xuweibj:
Looks good to me. Thanks @chenglch .

Issue #4240, opened 2017-11-06
[OpenBMC] bmcdiscover --check does not seem to work


whowutwut:
```
[root@briggs01 ~]# bmcdiscover -i 172.12.139.117 --check -u root -p 0penBmc
Warning: Wrong BMC password
```

If this is not supported from BMC, we need to block this function and say it's not supported. This is misleading and indicates that the password is incorrect. 

@zet809 We should add this as a known issue for 2.13.8

zet809:
Hi, @whowutwut , it is a little hard to distinguish between BMC and OpenBMC, and not much user is using these 2 options, so I disable them in PR #4258 .

zet809:
After code merged, I will add restriction in 2.13.8 release notes. Thx!

zet809:
Add restriction in 2.13.8 release notes, so, @whowutwut , can we close this issue? Thx!

zet809:
The `--check` and `--ipsource` option have been disabled, so close this issue. 
```
[root@briggs01 ~]# bmcdiscover -i 172.12.139.117 --check -u root -p 0penBmc
Error: The option '--check' is not supported
```

PR #4258, opened 2017-11-07
Disable option --check and --ipsource for bmcdiscover

zet809:
For #4240 

Disable option `--check` and `--ipsource` for bmcdiscover.

```
[root@briggs01 xCAT_plugin]# bmcdiscover -i 172.11.139.2 -p 0penBmc --ipsource
Error: The option '--ipsource' is not support at present
[root@briggs01 xCAT_plugin]# echo $?
1
[root@briggs01 xCAT_plugin]# bmcdiscover -i 172.11.139.2 -p 0penBmc --check
Error: The option '--check' is not support at present
[root@briggs01 xCAT_plugin]# echo $?
1
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -107,11 +107,11 @@ sub preprocess_request {
         }
         return \@requests;
     } elsif (grep /--check/, @ARGV) {
-        $callback->({ error => ["The option '--check' is not supported at present"], errorcode=>[1]});
+        $callback->({ error => ["The option '--check' is not supported"], errorcode=>[1]});
         $request = ();
         return;
     } elsif (grep /--ipsource/, @ARGV) {
-        $callback->({ error => ["The option '--ipsource' is not supported at present"], errorcode=>[1]});
+        $callback->({ error => ["The option '--ipsource' is not supported"], errorcode=>[1]});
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -106,6 +106,14 @@ sub preprocess_request {
             push @requests, $reqcopy;
         }
         return \@requests;
+    } elsif (grep /--check/, @ARGV) {
+        $callback->({ error => ["The option '--check' is not supported at present"], errorcode=>[1]});
```
whowutwut: @zet809 nit picky, but  would just rather say: `The '--check' option is not supported.` 

I don't think we have any ETA or plan right now to support it

xCAT-server/lib/xcat/plugins/bmcdiscover.pm
```diff
@@ -106,6 +106,14 @@ sub preprocess_request {
...
+        $request = ();
+        return;
+    } elsif (grep /--ipsource/, @ARGV) {
+        $callback->({ error => ["The option '--ipsource' is not supported at present"], errorcode=>[1]});
```
whowutwut: Same thing as above... nit picky `The '--ipsource' option is not supported.`

zet809:
Thx, @whowutwut , I have modified.

PR #4260, opened 2017-11-07
Copy in hidden files to mknb

jjohnson42:
mknb was ignoring hidden files (e.g. .screenrc), fix this.


zet809:

PR #4262, opened 2017-11-07
Add some known issues that are tied to the OpenBMC firmware

whowutwut:
Add in a section for CORAL reference for some known issues as we have a bunch of hardware at different FW levels and to allow internal admins to see work arounds that may be needed.


cxhong:
looks good to me.  I will merge it.

PR #4263, opened 2017-11-07
Skip connection to BMC on rpower reseat

jjohnson42:
rpower reseat gets directed at the chassis manager.  Skip the
superfluous connection to the BMC and keep it from interfering
with the reseat.


zet809:
Agree to merge.

Issue #4264, opened 2017-11-07
rpower command intermittently reports 503 Service Unavailable Error


belldi:
**Problem description**
I intermittently see error messages for " Error: 503 Service Unavailable" while attempting to power on compute node.  After failure message, my next attempt is generally successful.

**Attempt to power on compute node**
```
# rpower cn06 on
cn06: Error: 503 Service Unavailable
# echo $?
1
# rpower cn06 state
cn06: off
# rpower cn06 on
cn06: on
# rpower cn06 state
cn06: on
```

**xCAT version**
```
# lsxcatd -a
Version 2.13.8 (git commit e4f452a44d417e2f9c64b6dd064e3fc316046ac4, built Sat Nov  4 07:15:42 EDT 2017)
This is a Management Node
cfgloc=Pg:dbname=xcatdb;host=192.168.95.1|xcatadm
dbengine=Pg
dbname=xcatdb
dbhost=192.168.95.1
dbadmin=xcatadm
```

whowutwut:
@belldi  I formatted your above issue. 

We are seeing the 503 come back the first time that we call a REST command after rebooting the BMC.  I logged into the cn06 BMC and it looks like it's been up for around 20 minutes. 

Let me try and make sure this is a FW issue by taking xCAT out of the equation...

whowutwut:
Uh o... After rebooting the BMC, we are able to log into it without 503.  This error seems to be in our code.... 
```
[root@briggs01 p9]# ./login.sh 
==> IP=172.11.139.5
+ date
Tue Nov  7 17:29:23 EST 2017
+ curl -c cjar -k -X POST -H 'Content-Type: application/json' -d '{"data": [ "root", "0penBmc" ] }' https://172.11.139.5/login
{
  "data": "User 'root' logged in", 
  "message": "200 OK", 
  "status": "ok"
}
real	0m13.744s
user	0m0.021s
sys	0m0.065s
```

whowutwut:
@xuweibj  I think perhaps #4207 caused this so I installed 10/30 build of xCAT: 

```
[root@briggs01 xcat]# lsxcatd -v
Version 2.13.8 (git commit 0133577acf2482e30650fb87ce6f35a7dc3c2b4c, built Mon Oct 30 06:15:41 EDT 2017)
```
Reboot BMC:
```
[root@briggs01 openbmc]# ./run_cmd_bmc.sh  mid05tor12cn05 uptime
[...]

PR #4271, opened 2017-11-08
Update the known issues to remove invalid FW issue and replace with #4264…

whowutwut:
Turns out that SW406960 is actually an issue that we introduced in xCAT, see #4264.

Update known issues list to correctly reflect this.


whowutwut:
Merge to keep known issues list as current as possible.

PR #4274, opened 2017-11-09
enhance my_ip_facing when multiple ips found for xcatmaster

### LINKED ISSUE #4059: Incorrect IP address for xcatmaster
Our xcatmaster has legs into multiple networks.  xCAT uses the `my_ip_facing` subroutine to pick the appropriate IP address to talk to clients, but not for generating `mypostscript`.  That seems to just look at the `xcatmaster` attribute on the node, if it exists, or use the `master` attribute from the `site` table.  To make progress, I've just set `xcatmaster` to the alternate IP address for the groups that should use the alternate network.

Should this work 'automatically' or is this step required?

bybai:
fix #4059 

The logic to determine the $ENV{XCATMASTER} confirm to the following priority(from high to low):
* Using noderes.xcatmaster
* Facing_IP on MN or SN to the CN (If multiple and the SITEMASTER is also there, using SITEMASTER). If the CN is using SN, should configure xcatmaster for the CN, this should be included in doc.
* using SITEMASTER

When there is no xcatmaster configured in node attribute, If multiple my facing ips are found and the SITEMASTER is also there, using SITEMASTER;

UT results:
In HA MN, there are 2 ips for management network, before and after code changes, the MASTER and MASTER_IP are different in mypostscript, vip 10.5.106.100 is the master ip;
```
In xCAT MN
]# ip addr show dev eth0| grep inet
    inet 10.5.106.9/8 brd 10.255.255.255 scope global eth0
    inet 10.5.106.100/8 brd 10.255.255.255 scope global secondary eth0:0
```

Before the code changes
```
]# grep "MASTER" mypostscript
......
SITEMASTER='10.5.106.100'
export SITEMASTER
MASTER='10.5.106.9'
export MASTER
MASTER_IP=10.5.106.9
export MASTER_IP
MONMASTER='10.5.106.100'
export MONMASTER
```
After the code changes
```
]# grep MASTER mypostscript
SITEMASTER='10.5.106.100'
export SITEMASTER
MASTER='10.5.106.100'
export MASTER
MASTER_IP=10.5.106.100
export MASTER_IP
MONMASTER='10.5.106.100'
export MONMASTER
```

### DIFF
--- a/xCAT-server/lib/perl/xCAT/Postage.pm
+++ b/xCAT-server/lib/perl/xCAT/Postage.pm
@@ -417,22 +417,18 @@ sub makescript {
             #the ip address of the mn facing the compute node
             my @ipfnd = xCAT::NetworkUtils->my_ip_facing($node);
             my $ipfndscalar = @ipfnd;
-            if ($ipfnd[0]) {
-                if ($ipfndscalar ==2) {
-                    $::GLOBAL_TAB_HASH{noderes}{$node}{xcatmaster} = $ipfnd[1]; 
-                    $master = $ipfnd[1];
-                } elsif ($ipfndscalar > 2) {
+            unless ($ipfnd[0]) {
+                $master = $ipfnd[1];
+                if ($ipfndscalar > 2) {
                     foreach my $ipinfnd (@ipfnd) {
                         if ($::XCATSITEVALS{master} and $ipinfnd eq $::XCATSITEVALS{master}) {
-                            $::GLOBAL_TAB_HASH{noderes}{$node}{xcatmaster} = $ipinfnd;
                             $master = $ipinfnd;
                             last;
                         }
-                    }    
-                    unless ($master) {
-                        $::GLOBAL_TAB_HASH{noderes}{$node}{xcatmaster} = $ipfnd[1];
-                        $master = $ipfnd[1];
                     }
+                }
+                if ($master) {
+                    $::GLOBAL_TAB_HASH{noderes}{$node}{xcatmaster} = $master;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/perl/xCAT/Postage.pm
```diff
@@ -413,8 +413,33 @@ sub makescript {
             $noderesent = $::GLOBAL_TAB_HASH{noderes}{$node};
         }
 
+        unless ($master) {
+            #the ip address of the mn facing the compute node
+            my @ipfnd = xCAT::NetworkUtils->my_ip_facing($node);
+            my $ipfndscalar = @ipfnd;
+            if ($ipfnd[0]) {
+                if ($ipfndscalar ==2) {
```
robin2008: there might be more than 2 ips.

xCAT-server/lib/perl/xCAT/Postage.pm
```diff
@@ -413,8 +413,33 @@ sub makescript {
             $noderesent = $::GLOBAL_TAB_HASH{noderes}{$node};
         }
 
+        unless ($master) {
+            #the ip address of the mn facing the compute node
+            my @ipfnd = xCAT::NetworkUtils->my_ip_facing($node);
+            my $ipfndscalar = @ipfnd;
+            if ($ipfnd[0]) {
```
robin2008: The right value should be return when `ipfnd[0]` be zero , like [0, ip1, ip2]

xCAT-server/lib/perl/xCAT/Postage.pm
robin2008: code style: use blank  space after the '=='

xCAT-server/lib/perl/xCAT/Postage.pm
```diff
@@ -413,8 +413,33 @@ sub makescript {
...
+                if ($ipfndscalar ==2) {
+                    $::GLOBAL_TAB_HASH{noderes}{$node}{xcatmaster} = $ipfnd[1]; 
+                    $master = $ipfnd[1];
+                } elsif ($ipfndscalar > 2) {
```
robin2008: suggest modified to below:
```
unless ($ipfnd[0]) {
    $master = $ipfnd[1];
    if ($ipfndscalar > 2)
       ...
       last
     }
    if ($master) $::GLOBAL_TAB_HASH{noderes}{$node}{xcatmaster} = $master
```

PR #4277, opened 2017-11-09
[OpenBMC] Adding known issue for SW406771 

whowutwut:
This is regarding potential issues with the ID of the fw going from 1740A level of firmware and up


whowutwut:
Merge to update documentation

PR #4281, opened 2017-11-10
Add all modules when build xCAT-genesis-base

zet809:
When build xCAT-genesis-base, include all the kernel modules include in the host kernel. 

For ppc64 based on Fedora26, the package size is increase from 49M to 92M. And after installed on MN and run mknb,  the genesis rootfs size is increase from 55M to 123M. 
So, it seems OK to include all the kernel modules. 
```
[root@c910f03c05k03 noarch]# ls -lh xCAT-genesis-base-ppc64*
-rw-r--r-- 1 root root 49M Nov 10 03:09 xCAT-genesis-base-ppc64-2.13.7-snap201711100308.noarch.rpm
-rw-r--r-- 1 root root 92M Nov 10 05:10 xCAT-genesis-base-ppc64-2.13.7-snap201711100509.noarch.rpm
```
```
[root@fs1 ~]# ls -lh /tftpboot/xcat/genesis.fs.ppc64.gz
-rw-r--r-- 1 root root 55M Nov 10 05:18 /tftpboot/xcat/genesis.fs.ppc64.gz
[root@fs1 ~]# ls -lh /tftpboot/xcat/genesis.fs.ppc64.gz
-rw-r--r-- 1 root root 123M Nov 10 04:26 /tftpboot/xcat/genesis.fs.ppc64.gz
```

### DIFF
--- a/xCAT-genesis-builder/buildrpm
+++ b/xCAT-genesis-builder/buildrpm
@@ -35,6 +35,16 @@ else
     DRACUTMODDIR=/usr/lib/dracut/modules.d/97xcat
 fi
 
+# get all modules in the kernel
+mv ./installkernel ./default_module_list
+echo "#!/bin/bash" > ./installkernel
+#instmods
+for line in `cat /lib/modules/4.11.8-300.fc26.ppc64/modules.dep | awk -F: '{print \$1}'`; do 
+    basename $line >> ./installkernel; 
+done
+sed -i 's/\(.*\)\.ko/instmods \1/g' ./installkernel
+chmod +x ./installkernel
+
 mkdir -p $DRACUTMODDIR
 cp $DIR/* $DRACUTMODDIR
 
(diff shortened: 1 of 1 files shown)


zet809:
Hi, @whowutwut , if you think it is OK to include all kernel modules from the builder hosts, I will upload the xCAT-genesis-base-ppc64 rpm package onto GSA and try to build x86_64 ones. Thx!

whowutwut:
@zet809 Why is your test build using version 2.13.7?  You should use the rpm created with the daily build with this change.  

Any planned verification before including this into xcat-dep?  I think we should make sure again that it doesn't install over the xCAT-genesis-builder rpm that we ship in the beta directory on xcat.org. 

I also think we should build a version for 2.13.8 and package it with the 2.13.8 GA version of xcat-dep, we missed this and which included some changes for the mlnx drivers.     It would be strange for people installing xCAT 2.13.8 and get a xCAT xCAT-genesis-base-2.13.9... rpm

Youl

zet809:
Hi, @whowutwut , an enhance of this PR can be seen in PR #4292 , pls feel free to add comments there. Thx!

PR #4254, opened 2017-11-07
Task 3339, rspconfig ip/netmask/gateway/vlan

xuweibj:
#3339 
Mini design: https://github.com/xcat2/xcat-core/wiki/rspconfig-network

UT:

* Static -> Static
```
rspconfig mid05tor12cn11 ip=172.11.139.11 netmask=255.255.0.0 gateway=0.0.0.0 vlan=11
Mon Nov  6 21:55:40 2017 mid05tor12cn11: [openbmc_debug] curl -k -c cjar -H "Content-Type: application/json" -d '{ "data": ["root", "xxxxxx"] }' https://172.11.139.222/login
Mon Nov  6 21:55:41 2017 mid05tor12cn11: [openbmc_debug] login_response 200 OK
Mon Nov  6 21:55:41 2017 mid05tor12cn11: [openbmc_debug] curl -k -b cjar -X GET -H "Content-Type: application/json" https://172.11.139.222/xyz/openbmc_project/network/enumerate
Mon Nov  6 21:55:42 2017 mid05tor12cn11: [openbmc_debug] rspconfig_get_response 200 OK
Mon Nov  6 21:55:42 2017 mid05tor12cn11: [openbmc_debug] curl -k -b cjar -X POST -H "Content-Type: application/json" -d '{"data":["eth0","11"]}' https://172.11.139.222/xyz/openbmc_project/network/action/VLAN
Mon Nov  6 21:55:43 2017 mid05tor12cn11: [openbmc_debug] rspconfig_vlan_response 200 OK
Mon Nov  6 21:55:45 2017 mid05tor12cn11: [openbmc_debug] curl -k -b cjar -X POST -H "Content-Type: application/json" -d '{"data":["xyz.openbmc_project.Network.IP.Protocol.IPv4","172.11.139.11",16,"0.0.0.0"]}' https://172.11.139.222/xyz/openbmc_project/network/eth0_11/action/IP
Mon Nov  6 21:55:50 2017 mid05tor12cn11: [openbmc_debug] rspconfig_ipobject_response 200 OK
Mon Nov  6 21:55:52 2017 mid05tor12cn11: [openbmc_debug] curl -k -b cjar -X GET -H "Content-Type: application/json" https://172.11.139.222/xyz/openbmc_project/network/enumerate
Mon Nov  6 21:55:58 2017 mid05tor12cn11: [openbmc_debug] rspconfig_check_response 200 OK
[...]

Issue #4235, opened 2017-11-02
[OpenBMC] Enable function to trigger BMC dumps


whowutwut:
As we start looking at ways to capture dump to debug issues with the BMC, there's a dump API: https://github.com/openbmc/phosphor-dbus-interfaces/tree/master/xyz/openbmc_project/Dump

However, the API is lacking some information.   Externally, from xCAT, we have the following things that we would want to do: 

* List available dumps
   ```
   curl -c cjar -b cjar -k https://<BMC_IP>/xyz/openbmc_project/dump/list
   ```
* Manually trigger a generation of a dump 
   ```
   curl -c cjar -b cjar -k -H "Content-Type: application/json" -d "{\"data\": []}" \
   -X POST  https://<BMC_IP>/xyz/openbmc_project/dump/action/CreateDump
   ```
* Download the dump 
   ```
   curl -O -J -c cjar -b cjar -k -X GET https://<BMC_IP>/download/dump/<ID>
   ```
* Clear the dumps from the BMC 
   * Single
      ```
      curl -c cjar -b cjar -k -H "Content-Type: application/json" -d "{\"data\": []}" \
      -X POST  https://<BMC_IP>/xyz/openbmc_project/dump/entry/<ID>/action/Delete
      ```
   * All
      ```
      curl -c cjar -b cjar -k -H "Content-Type: application/json" -d "{\"data\": []}" \
      -X POST  https://<BMC_IP>/xyz/openbmc_project/dump/action/DeleteAll
      ```

Suggested xCAT Externals: 
* List:  `rspconfig <noderange> dump --list` 
* Generate:  `rspconfig <noderange> dump --generate`
* Download: `rspconfig <noderange> dump --download <ID>`
* Clear
    * single: `rspconfig <noderange> dump -clear <id>`
    * all:  `rspconfig <noderange> dump -clear all`

When we download, openbmc names it the following: `obmcdump_<ID>_<Elapsed Time>.tar.xz`, So already we will have issues scaling this out and identifying which BMC this file comes from. 

Create some use cases and will open questions to OpenBMC team...

**Use Case 1:** 
As an administrator, I detect issues on 100 BMCs and I'd like to generate and capture the dump across all these BMCs...

whowutwut:
Some more ideas... If nothing changes in the request to openbmc team, we could do the following: 

[...]

PR #4289, opened 2017-11-14
Task 4235, rspconfig BMC dump for OpenBMC

xuweibj:
#4235 

* rspconfig -l|--list
```
[root@briggs01 dump]# rspconfig mid05tor12cn02 dump --list
mid05tor12cn02: [1] Elapsed: 09/22/2017 14:41:05, Size: 4624
mid05tor12cn02: [2] Elapsed: 09/22/2017 14:46:29, Size: 3608
mid05tor12cn02: [3] Elapsed: 10/05/2017 22:46:59, Size: 180976
mid05tor12cn02: [4] Elapsed: 10/09/2017 22:10:46, Size: 8184
mid05tor12cn02: [5] Elapsed: 10/09/2017 22:10:51, Size: 8404
mid05tor12cn02: [6] Elapsed: 10/09/2017 22:10:57, Size: 8384
mid05tor12cn02: [7] Elapsed: 10/26/2017 10:29:16, Size: 3692
mid05tor12cn02: [8] Elapsed: 10/26/2017 10:29:16, Size: 3716
mid05tor12cn02: [9] Elapsed: 10/26/2017 10:29:16, Size: 3760
mid05tor12cn02: [10] Elapsed: 10/26/2017 10:29:16, Size: 3676
```

* rspconfig -g|--generate
```
[root@briggs01 dump]# rspconfig mid05tor12cn02 dump -g
mid05tor12cn02: [11] success
```
When testing, found that need to wait several minutes could saw it in the list.

* rspconfig -c|--clear
```
[root@briggs01 dump]# rspconfig mid05tor12cn02 dump --clear 11
mid05tor12cn02: [11] clear
```

* rspconfig -d|--download
```
[root@briggs01 dump]# rspconfig mid05tor12cn02 dump --download 2
mid05tor12cn02: Saved dump 2 as /var/log/xcat/dump/mid05tor12cn02_dump_2.tar.xz
```
I modified some code in the same line with #4254, may has conflicts.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -329,6 +329,41 @@ my %status_info = (
     RSPCONFIG_SSHCFG_RESPONSE => {
         process        => \&rspconfig_sshcfg_response,
     },
+    RSPCONFIG_DUMPLIST_REQUEST => {
+        method         => "GET",
+        init_url       => "$openbmc_project_url/dump/enumerate",
+    },
+    RSPCONFIG_DUMPLIST_RESPONSE => {
+        process        => \&rspconfig_dump_response,
+    },
+    RSPCONFIG_DUMPCRT_REQUEST => {
+        method         => "POST",
+        init_url       => "$openbmc_project_url/dump/action/CreateDump",
+        data           => "[]",
+    },
+    RSPCONFIG_DUMPCRT_RESPONSE => {
+        process        => \&rspconfig_dump_response,
+    },
+    RSPCONFIG_DUMPCLR_REQUEST => {
+        method         => "POST",
+        init_url       => "$openbmc_project_url/dump/entry/#ID#/action/Delete",
+        data           => "[]",
+    },
+    RSPCONFIG_DUMPCLRA_REQUEST => {
+        method         => "POST",
+        init_url       => "$openbmc_project_url/dump/action/DeleteAll",
+        data           => "[]",
+    },
+    RSPCONFIG_DUMPCLR_RESPONSE => {
+        process        => \&rspconfig_dump_response,
+    },
+    RSPCONFIG_DUMPDWLD_REQUEST => {
+        init_url       => "download/dump/#ID#",
+        process        => \&rspconfig_dump_response,
+    },
+    RSPCONFIG_DUMPDWLD_RESPONSE => {
+        process        => \&rspconfig_dump_response,
+    },
     RVITALS_REQUEST => {
         method         => "GET",
         init_url       => "$openbmc_project_url/sensors/enumerate",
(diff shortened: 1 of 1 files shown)


whowutwut:
@xuweibj What happens if we issue `dump` by itself?  

Looking at usability scenario here.. I think most of the time we would want to call this to generate and download...  From the lessons learned in rflash... let me propose a use case here. 

`rspconfig <noderange> dump` 
   * Triggers a dump to be generated, the return of the create will give us the ID
   * Save the ID and then wait here for several minutes... 
   * When the ID is seen in the BMC, download it to the proposed directory 

The normal output could be.... 
```
Capturing BMC Dump information, this will take some time...
Generating Dump....
Downloading Dump...
```

If we have --verbose option, then we could do: 
```
Capturing BMC Dump information, this will take some time...
Generating BMC Dump....
Generating BMC Dump on <node>...
....
Downloading BMC Dump...
Downloading BMC Dump to /var/log/xcat/dump/<file>
```

The reason I state the above... is because the solution we implement does not scale or help the admin.  The above provides an xCAT interface into the REST but it does not help the user generate and capture Dumps across 100, 1000, 2000, 5000 nodes...  There's no way to guarantee the next dump ID is the same unless we do a clear before...  So we need to have the create pass back the ID, and then we have to wait until the ID is generated and capture the output...

xuweibj:
@whowutwut , I want to make sure I understand your point well. Do you mean for command ``rspconfig dump``, we send a request to create a new one, and if could list it download it? So if do this, do we also need the options ``--list, --generate, --download``? Or is the requirement added the code for ``rspconfig dump`` itself in this pull request?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -781,10 +816,22 @@ sub parse_args {
                 $setorget = "get";
             } elsif ($subcommand =~ /^sshcfg$/) {
                 $setorget = ""; # SSH Keys are copied using a RShellAPI, not REST API
+            } elsif ($subcommand eq "dump") {
+                my $option = $ARGV[1];
+                if ($option =~ /^-d$|^--download$/) {
+                    return ([ 1, "Invalid parameter for $command $option" ]) unless ($ARGV[2]);
```
zet809: The error msg is not making sense. 
What about 'No dump file ID specified'?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -781,10 +816,22 @@ sub parse_args {
...
+                    return ([ 1, "Invalid parameter for $command $option $ARGV[2]" ]) if ($ARGV[2] !~ /^\d*$/);
+                } elsif ($option =~ /^-c$|^--clear$/) {
+                    return ([ 1, "Invalid parameter for $command $option" ]) unless ($ARGV[2]);
```
zet809: Same comment with line 822.

zet809:
hi, @whowutwut , you are right that this interface may not make sense for groups of nodes, we need to provide a better interface quickly.
But maybe sometime, if user want to download a dump file days before, he can use the interface right now.(Just as `rflash <node> <image_id> -a` to active a specified version of firmware, right?)
So, I think we can merge it right now.

PR #4311, opened 2017-11-16
Delete bmcdiscover_check_paswd,bmcdiscover_check_passwd_wrong,bmcdiscover_get_ipsource from bundles

hu-weihua:
Depending on @zet809 's latest information, we won't support ``--check`` and ``--ipsource`` option in discovery, so delete these related 3 cases ``bmcdiscover_check_paswd``,``bmcdiscover_check_passwd_wrong``,``bmcdiscover_get_ipsource`` from daily bundles.


whowutwut:
Adding `component:test` for any PR that touches testcase file only.

Issue #4299, opened 2017-11-14
`xdsh` return with no output when "site.master" is not set


mattaezell:
update:

steps to reproduce:
````
 chdef <cn> servicenode=
 chdef -t site -o clustersite master=
 xdsh <cn> hostname
````

==============================================

I'm not getting any output from `xdsh` unless I turn on bypass mode.

```
[root@mgmt1 ~]# lsdef -s compute
h41n01  (node)
h41n02  (node)
h41n03  (node)
h41n04  (node)
h41n05  (node)
h41n06  (node)
h41n07  (node)
h41n08  (node)
h41n09  (node)
h41n10  (node)
h41n11  (node)
h41n12  (node)
h41n13  (node)
h41n14  (node)
h41n15  (node)
h41n16  (node)
h41n17  (node)
h41n18  (node)
[root@mgmt1 ~]# xdsh compute hostname
[root@mgmt1 ~]# xdsh compute -T -v hostname
[root@mgmt1 ~]# ssh h41n01 hostname
h41n01
[root@mgmt1 ~]# xdsh compute -B hostname
h41n03: h41n03
h41n02: h41n02
h41n11: h41n11
h41n01: h41n01
h41n10: h41n10
h41n14: h41n14
h41n04: h41n04
h41n06: h41n06
h41n07: h41n07
h41n08: h41n08
h41n12: h41n12
h41n15: h41n15
h41n17: h41n17
h41n18: h41n18
h41n16: h41n16
h41n13: h41n13
```
Any ideas?

```
[root@mgmt1 ~]# lsxcatd -v
Version 2.13.8 (git commit fe7405f1702f63758e0b65ae7999bd457ea0cd5a, built Thu Nov  2 07:15:44 EDT 2017)
```

**Comment by cxhong** [comment]
@mattaezell , can u check `nslookup` output?  
if you get this:
```
[root@mgmt1 ~]# nslookup h41n01
Server:         160.91.203.61
Address:        160.91.203.61#53

** server can't find h41n01: NXDOMAIN
````
check /etc/resolv.conf, add mgmt1's ip address , then run 'makedns' again.

mattaezell:
Hello @cxhong.  We don't have our compute nodes in DNS, they are in the /etc/hosts file.  Notice the ssh above worked fine.

whowutwut:
@chenglch @zet809 any ideas here?

immarvin:
[...]

PR #4313, opened 2017-11-16
fix issue #4299 `xdsh` return with no output when "site.master" is not set

immarvin:
fix issue:
https://github.com/xcat2/xcat-core/issues/4299

UT:
````
[root@c910f03c05k21 ~]# lsdef -t site -o clustersite -i master
Object name: clustersite
[root@c910f03c05k21 ~]# xdsh c910f03c05k27   hostname
c910f03c05k27: c910f03c05k27
[root@c910f03c05k21 ~]# tail -n2 /var/log/xcat/cluster.log
Nov 16 01:33:27 c910f03c05k21 xcat[17388]: xCAT: Allowing xdsh to c910f03c05k27 hostname for root from localhost
Nov 16 01:33:27 c910f03c05k21 xcat[17389]: site.master is not set!
[root@c910f03c05k21 ~]#
````

### DIFF
--- a/perl-xCAT/xCAT/ServiceNodeUtils.pm
+++ b/perl-xCAT/xCAT/ServiceNodeUtils.pm
@@ -509,8 +509,11 @@ sub get_ServiceNode
 
     # get site.master this will be the default
     my $master = xCAT::TableUtils->get_site_Master();
-    $noderestab = xCAT::Table->new('noderes');
+    unless($master){
+        xCAT::MsgUtils->message('SW',"site.master is not set!\n");
+    }
 
+    $noderestab = xCAT::Table->new('noderes');
     unless ($noderestab)    # no noderes table, use default site.master
     {
         xCAT::MsgUtils->message('I',
@@ -548,7 +551,7 @@ sub get_ServiceNode
                     my $key = $rec->{$snattribute};
                     push @{ $snhash{$key} }, $node;
                 }
-                else                                  # use site.master
+                elsif($master)                                  # use site.master
                 {
                     push @{ $snhash{$master} }, $node;
                 }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by robin2008** [comment]
the caller already handled the case when sn is empty, it will directly to use MN.

**Comment by immarvin** [comment]
@robin2008 , yes, site.master is not mandatory for xdsh

**Comment by whowutwut** [comment]
@robin2008 @immarvin Sorry for the really late review here.. 
 
Shouldn't this be an ERROR message in the log?
Also the log format is inconsistent.... 

Should it look something like:
```
Nov 16 01:33:27 c910f03c05k21 xcat[17388]: xCAT: Allowing xdsh to c910f03c05k27 hostname for root from localhost
Nov 16 01:33:27 c910f03c05k21 xcat[17389]: xCAT: Error: site.master is not configured.
```

If we could even do better....where we can say what command failed, it would open up better ways to grep for the error... 

```
Nov 16 01:33:27 c910f03c05k21 xcat[17388]: xCAT: Allowing xdsh to c910f03c05k27 hostname for root from localhost
Nov 16 01:33:27 c910f03c05k21 xcat[17389]: xCAT: Error: xdsh failed. site.master is not configured.
```

Thinking what happens if there are lots and lots of nodes.... would first want to grep Error., and then trickle down.. looks like the host is there to grep on too..

immarvin:
hi @whowutwut , `site.master` is not mandatory for xdsh, `xdsh` works well when `site.master` is not set. The issue we found on Ralph's env was proved a bug , this PR is to fix it.  

so it is not a error that`site.master` is not set , so just add a warning message in syslog to reminder user on this.

PR #4307, opened 2017-11-15
Introduce a natural compare func for sorting array and using it for rinv/rvital

robin2008:
To fix #4298 

1， add a new compare func in xCAT::Utils
2,   in openbmc plugin to use it for sorting the output by natural.

Note:  it did not cover other place to use the natural sorting 

UT:
```
[root@briggs01 xcat]# rinv mid05tor12cn05 all
mid05tor12cn05: BMC BuildDate :
mid05tor12cn05: BMC FieldReplaceable : 0
mid05tor12cn05: BMC Manufacturer : IBM
mid05tor12cn05: BMC Model :
mid05tor12cn05: BMC PartNumber : 01DH118
mid05tor12cn05: BMC Present : 1
mid05tor12cn05: BMC PrettyName : BMC PLANAR
mid05tor12cn05: BMC SerialNumber : 000000000000
mid05tor12cn05: BMC UUID :
mid05tor12cn05: CPU0 BuildDate : 1996-01-01 - 00:00:00
mid05tor12cn05: CPU0 CORE0 Functional : 1
mid05tor12cn05: CPU0 CORE0 Present : 0
mid05tor12cn05: CPU0 CORE0 PrettyName :
mid05tor12cn05: CPU0 CORE1 Functional : 1
mid05tor12cn05: CPU0 CORE1 Present : 0
mid05tor12cn05: CPU0 CORE1 PrettyName :
mid05tor12cn05: CPU0 CORE2 Functional : 1
mid05tor12cn05: CPU0 CORE2 Present : 0
mid05tor12cn05: CPU0 CORE2 PrettyName :
mid05tor12cn05: CPU0 CORE3 Functional : 1
mid05tor12cn05: CPU0 CORE3 Present : 0
mid05tor12cn05: CPU0 CORE3 PrettyName :
mid05tor12cn05: CPU0 CORE4 Functional : 1
mid05tor12cn05: CPU0 CORE4 Present : 1
mid05tor12cn05: CPU0 CORE4 PrettyName :
mid05tor12cn05: CPU0 CORE5 Functional : 1
mid05tor12cn05: CPU0 CORE5 Present : 1
mid05tor12cn05: CPU0 CORE5 PrettyName :
mid05tor12cn05: CPU0 CORE6 Functional : 1
mid05tor12cn05: CPU0 CORE6 Present : 0
mid05tor12cn05: CPU0 CORE6 PrettyName :
mid05tor12cn05: CPU0 CORE7 Functional : 1
[...]

PR #4314, opened 2017-11-16
natural_sort_cmp: recursion to iterative implementation

samveen:
Improve the implementation of `natural_sort_cmp` to be iterative rather than recursive (speedup for large output sets).

This is a follow up to #4307


robin2008:
@samveen Thanks for your contribution, the `while` has a bit better than recursive (I've tested with /var/log/messages with 50k lines)  

but I found that your changes in `openbmc.pm` will case the error.

samveen:
@robin2008 checking `openbmc.pm`

samveen:
@robin2008 Can you give me more details of the error you are seeing? I cannot see the error.

Issue #4302, opened 2017-11-14
[OpenBMC] If BMC is not pinging, return message after 3 seconds, that waited for 10.


whowutwut:
This error message is misleading...   It looks like when the BMC is not pingable
```
[root@stratton01 ~]# ping_bmc p9euh02
+ lsdef p9euh02 -i bmc -c
+ cut -d= -f2
+ xargs ping -c 1
PING 10.6.9.254 (10.6.9.254) 56(84) bytes of data.
From 10.6.29.1 icmp_seq=1 Destination Host Unreachable

--- 10.6.9.254 ping statistics ---
1 packets transmitted, 0 received, +1 errors, 100% packet loss, time 0ms

```

We should not have a time in the error message: 
```
[root@stratton01 ~]# time rpower p9euh02 state
p9euh02: Error: BMC did not respond within 10 seconds, retry the command.

real	0m3.315s
user	0m0.121s
sys	0m0.015s
```

Says 10 seconds but really 3... . Can we differentiate between the two errors and return something saying BMC is not pinging?  

I really don't like this change in behavior from #4207 because it's confusing me as a user about what is really going on with the BMC, maybe nothing is wrong but we choose not to wait for the BMC to respond... and in this case... the bmc is not pinging but I kept issuing the command because it "normally " fails the first time on a bmc reboot... so i thought it would work on the 2nd or 3rd attempt... 

So this is not good at all..

xuweibj:
#4320

whowutwut:
Using today's build, I'm ok with the output currently issued. 

```
[root@briggs01 910.1746.20171116b_1742Ed]# rpower mid05tor12cn18 bmcreboot
mid05tor12cn18: BMC reboot
[root@briggs01 910.1746.20171116b_1742Ed]# rpower mid05tor12cn18 state
mid05tor12cn18: Error: BMC did not respond. Verify BMC is in BMCReady state and retry the command.
[root@briggs01 910.1746.20171116b_1742Ed]#
```

PR #4320, opened 2017-11-16
Modify timeout of login by curl command for OpenBMC

xuweibj:
#4264 #4302 
Modify timeout to 20s, and modify error msg.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -18,6 +18,7 @@ use warnings "all";
 use JSON;
 use HTTP::Async;
 use HTTP::Cookies;
+use LWP::UserAgent;
 use File::Basename;
 use File::Spec;
 use File::Copy qw/copy cp mv move/;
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -591,7 +591,7 @@ sub process_request {
 
     foreach my $node (keys %node_info) {
         if (!$valid_nodes{$node}) {
-            xCAT::SvrUtils::sendmsg([1, "BMC did not respond within 10 seconds, retry the command."], $callback, $node);
+            xCAT::SvrUtils::sendmsg([1, "BMC could not be connected or did not respond within 20 seconds, retry the command."], $callback, $node);
```
whowutwut: @xuweibj Thanks!  

Do we think we would hit this often for BMC that are really up without error?  
If not, we can change this to be more generic....

`BMC did not respond. Verify BMC is in BMCReady state and retry the command.`

whowutwut:
@xuweibj Did you run a "bmcreboot" and then issue the command and see if we get the early timeout?

zet809:

Issue #4206, opened 2017-10-31
[Performance] Integrate simulation tool into baseline testing case


robin2008:
Here we need add below scope into base line testing:

- [x]  Need to integrated openbmc simulation tool for remote control baseline testing (500)
- [x]  Using container as live fake node to testing updatenode/pping/xdsh

robin2008:
In xCAT autotest environment, there are 3 nodes as a test group, 1MN and 1SN, 1CN.
The previous performance testing only leverage the MN to do those test without a real CN or an BMC.

And now, it is planed to use run  openbmc simulation on SN to mimic 1000 openbmc nodes, so that it could test the HW commands based on those simulators.
And to run 500 containers on both SN and CN to mimic 500-1000 CN with ssh alive, so that it could 
test the updatenode commands based on those simulators.

Note:  For simple, it will use the bridge networking on the 2nd interface (for example, eth1).

robin2008:
PR #4295 is for this issue.

PR #4325, opened 2017-11-17
Put all required performance data into one report

robin2008:
Another PR for #4206 
 * Minor enhance on `xcatperftest` have the capability to put all logs into one file
 * In test case, all performance data (db only, hw ctl, and node related) is logged into `perfreport-5000.log.daily`
 * Fix a bug in simulatorctl.sh, and it cause the script cannot be found


bybai:
I am ok to merge it.

Issue #101, opened 2015-08-20
[FVT]: updatenode returns some strange error message 


tingtli:
xCAT 2.10 20150714 build.

[root@c910f04x35 postscripts]# updatenode c910f04x36 "confignics -s"
 c910f04x36: xcatdsklspost: downloaded postscripts successfully
 c910f04x36: chmod: cannot access 'exited': No such file or directory <== please remove it.
 c910f04x36: chmod: cannot access 'with': No such file or directory
 c910f04x36: chmod: cannot access 'code': No such file or directory
 c910f04x36: chmod: cannot access '0': No such file or directory
 c910f04x36: Wed Jul 15 02:10:37 EDT 2015 Running postscript: confignics -s
 c910f04x36: confignics on c910f04x36: config install nic:1, remove: 0, iba ports :
 c910f04x36: enp0s26u1u5u5!50.4.36.1
 c910f04x36: confignics on c910f04x36: call 'configeth enp0s26u1u5u5 50.4.36.1 50 _0_0_0-255_0_0_0|'
 c910f04x36: configeth on c910f04x36: os type: redhat
 c910f04x36: configeth on c910f04x36: old configuration: 2: enp0s26u1u5u5: <BROAD CAST,MULTICAST,UP,LOWER_UP=""> mtu 1500 qdisc pfifo_fast state UNKNOWN qlen 1000
 c910f04x36: link/ether 36:40:b5:b9:11:19 brd ff:ff:ff:ff:ff:ff
 c910f04x36: inet 50.4.36.1/8 brd 50.255.255.255 scope global enp0s26u1u5u5
 c910f04x36: valid_lft forever preferred_lft forever
 c910f04x36: inet6 fe80::3440:b5ff:feb9:1119/64 scope link
 c910f04x36: valid_lft forever preferred_lft forever
 c910f04x36: configeth on c910f04x36: new configuration
 c910f04x36: 50.4.36.1, 50.0.0.0, 255.0.0.0, 50.0.0.103
 c910f04x36: 2: enp0s26u1u5u5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc p fifo_fast state UNKNOWN mode DEFAULT qlen 1000
 c910f04x36: enp0s26u1u5u5
 c910f04x36: configeth on c910f04x36: enp0s26u1u5u5 no changed, return directly.
 c910f04x36: confignics on c910f04x36: call 'configeth -s eno1
[...]

PR #4328, opened 2017-11-17
Fix merging in xdcpmerge.sh

mrmhodak:
Two fixes:
1. The grep pattern when finding duplicate usernames is missing ":" at the end. So, for example user "test" would also match "test2, etc.". Adding the ":" delimiter fixes the issue.
2. Another issue happens when the file to be merged is a superset of the files on the nodes. For example, if a new user is added and entire passwd file (that is otherwise identical) is sent to be merged. In this case, the $filebackup.nodups file, i.e. the original file with duplicates removed, becomes empty and the condition "if [ -s "$filebackup.nodups" ]" does not execute. Then the merged file ends up being the original file with the merge file fully appended, clearly not what was intended.

This is solved by changing the condition to check for file existence "-a" rather then for size. Additionally, I also turn the logic around so that the duplicates are removed from the merge file and then added to the original file. I think this makes logic a bit cleaner and also ensures that existing entries are not reordered or changed in any way.

### DIFF
--- a/xCAT-server/share/xcat/scripts/xdcpmerge.sh
+++ b/xCAT-server/share/xcat/scripts/xdcpmerge.sh
@@ -91,14 +91,13 @@ for i in $*; do
     # now run it
     eval $grepcmd
   fi 
-  # Now update the currentfile  
-  cat $filebackup > $curfile
+  
   # add new entries from mergefile, if any 
   if [ -a "$mergefile.nodups" ]; then
     cat $mergefile.nodups >> $curfile
     rm $mergefile.nodups
   fi
-  #echo "cat $filebackup $mergefile > $curfile"
+  
   # now cleanup
   rm $filebackup.userlist 
   #  echo "rm $filebackup.userlist"
--- a/xCAT-server/share/xcat/scripts/xdcpmerge.sh
+++ b/xCAT-server/share/xcat/scripts/xdcpmerge.sh
@@ -90,15 +90,14 @@ for i in $*; do
     #echo "grepcmd=$grepcmd"
     # now run it
     eval $grepcmd
-    # if no dups file created
-    if [ -a "$mergefile.nodups" ]; then
-      cp -p $mergefile.nodups $mergefile
-      #echo "cp -p $mergefile.nodups $mergefile" 
-      rm $mergefile.nodups
-    fi 
   fi 
   # Now update the currentfile  
-  cat $filebackup $mergefile > $curfile
+  cat $filebackup > $curfile
+  # add new entries from mergefile, if any 
+  if [ -a "$mergefile.nodups" ]; then
+    cat $mergefile.nodups >> $curfile
+    rm $mergefile.nodups
+  fi
   #echo "cat $filebackup $mergefile > $curfile"
   # now cleanup
   rm $filebackup.userlist 
--- a/xCAT-server/share/xcat/scripts/xdcpmerge.sh
+++ b/xCAT-server/share/xcat/scripts/xdcpmerge.sh
@@ -82,19 +82,19 @@ for i in $*; do
     done
     # remove the last delimiter
     userlisttmp="${userlist%?}"
-    listend=")'"
+    listend="):'"
     userlist=$userlisttmp$listend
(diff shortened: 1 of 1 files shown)


zet809:
hi, @immarvin , will you pls help to review this PR? Thx!

immarvin:
hi @mrmhodak . have you signed the CLA? if not , please follow the steps here http://xcat-docs.readthedocs.io/en/stable/developers/license/contributors.html

mrmhodak:
@immarvin: I am contributing as a Lenovo employee, presumably they have already signed the CCLA.
Is this sufficient, or do I still need to sign the individual agreement as well?

whowutwut:
@mrmhodak thanks for letting us know, Yes no need to sign an individual CLA

immarvin:
hi @mrmhodak , see my comments

xCAT-server/share/xcat/scripts/xdcpmerge.sh
```diff
@@ -82,19 +82,19 @@ for i in $*; do
...
     #echo "grepcmd=$grepcmd"
     # now run it
     eval $grepcmd
     # if no dups file created
-    if [ -s "$filebackup.nodups" ]; then
-      cp -p $filebackup.nodups $filebackup
-      #echo "cp -p $filebackup.nodups $filebackup" 
-      rm $filebackup.nodups
+    if [ -a "$mergefile.nodups" ]; then
+      cp -p $mergefile.nodups $mergefile
```
immarvin: suggest preserve the original `mergefile`,  a backup file for `mergefile` should be created and be overwritten here, the concatenation in line #101 should use the backup file, remeber to remove the backup merge file at the end

immarvin:
hi @mrmhodak , thanks for contribution,  the code logic seems ok, some minor changes required, see my comments

xCAT-server/share/xcat/scripts/xdcpmerge.sh
mrmhodak: @immarvin: We can do it that way, but the conditional at line 94 needs to be eliminated, because we are running into the issue of the backup file being created in the conditional block, but used later unconditionally.

What do you think? The conditional statement is redundant anyway, $mergefile.nodups gets created even if it is empty.

immarvin: hi @mrmhodak , agree

mrmhodak:
@immarvin: When implementing your suggestions, I realized that it is not necessary to overwrite mergefile at all, because what is needed is already in mergfile.nodups. So I am simply using that file to add to the original file. I think it makes logic a bit cleaner.

xCAT-server/share/xcat/scripts/xdcpmerge.sh
```diff
@@ -82,23 +82,22 @@ for i in $*; do
...
-    # if no dups file created
-    if [ -s "$filebackup.nodups" ]; then
-      cp -p $filebackup.nodups $filebackup
-      #echo "cp -p $filebackup.nodups $filebackup" 
-      rm $filebackup.nodups
-    fi 
   fi 
   # Now update the currentfile  
-  cat $filebackup $mergefile > $curfile
+  cat $filebackup > $curfile
```
immarvin: 1. if `$filebackup` is the same as `$curfile`, why bother overwrite `$curfile` here?
2. the changes should better be applied to `$curfile` once

immarvin:
hi @mrmhodak , see my comments, sorry for inconvenience

xCAT-server/share/xcat/scripts/xdcpmerge.sh
mrmhodak: Good point, let me fix that

mrmhodak: @immarvin Take a look now, removing the line where mergefile ovewrites curfile, the rest should work and the changes are applied in one shot now

mrmhodak:
@immarvin: have a look now, removing the unnecessary copy, the rest should work fine

immarvin:
i am ok with the change

PR #4330, opened 2017-11-18
Use IO::Socket to check BMC console port

mattaezell:
nmap's `-Pn` option instructs it to assume the host is up, but xCAT only pauses if 0 are up.  In my environment, nmap was always showing the host as up, making the nmap check ineffective.  I generally prefer native commands to shelling out anyway, and this patch makes my service nodes that don't have their computes attached yet usable.

### DIFF
--- a/xCAT-server/share/xcat/cons/openbmc
+++ b/xCAT-server/share/xcat/cons/openbmc
@@ -3,6 +3,7 @@
 use Fcntl qw(:DEFAULT :flock);
 use Time::HiRes qw(sleep);
 use File::Path;
+use IO::Socket;
 BEGIN {
     $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : -d '/opt/xcat' ? '/opt/xcat' : '/usr';
 }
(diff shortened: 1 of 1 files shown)


whowutwut:
@cxhong Can you please take a look at this PR?

cxhong:
thanks @mattaezell

xCAT-server/share/xcat/cons/openbmc
```diff
@@ -92,15 +93,25 @@ if ($ENV{SSHCONSOLEPORT}) {
...
     sleep(0.1);
     release_lock();
-    $nmap_output = `/usr/bin/nmap $bmcip -p $sshport -Pn`;
+    $sock = IO::Socket::INET->new(
+        PeerAddr => $bmcip,
+        PeerPort => $sshport,
+        Proto => "tcp",
+        Timeout => 10
+        );
 }
```
neo954: Resource leaking. One file descriptor, `$sock` is leaked here. And opened file descriptor will be passed across `exec()` system call in line 122 or 124. Thus, one file descriptor is leaked forever here.

Issue #4360, opened 2017-11-22
[automation] "makentp -a" failed in sles12.2 hierarchical environment


hu-weihua:
``makentp -a`` failed in sles12.2 hierarchical environment.

The output of ``makentp -a``  is:
```
RUN:makentp -a [Wed Nov 22 01:50:53 2017]
ElapsedTime:139 sec
RETURN rc = 1
OUTPUT:
configuring management node: c910f03c09k09.
configuring service nodes: c910f03c09k10
c910f03c09k10: vpdupdate is not supported on the pSeries KVM Guest platform
c910f03c09k10: xcatdsklspost: updating VPD database
c910f03c09k10: xcatdsklspost: downloaded postscripts successfully
c910f03c09k10: Wed Nov 22 06:50:55 EST 2017 Running postscript: setupntp
c910f03c09k10: inactive
c910f03c09k10: Created symlink from /etc/systemd/system/multi-user.target.wants/ntpd.service to /usr/lib/systemd/system/ntpd.service.
c910f03c09k10: postscript: setupntp exited with code 1
c910f03c09k10: Running of postscripts has completed.
```
It seems sync time timeout. 

```
Nov 22 01:50:56 c910f03c09k10 xcat:  setting current time
Nov 22 01:50:56 c910f03c09k10 ntpd[7761]:  ntpd 4.2.8p8@1.3265-o Mon Jun  6 08:12:00 UTC 2016 (1): Starting
Nov 22 01:50:56 c910f03c09k10 ntpd[7761]:  Command line: ntpd -gq
Nov 22 01:50:56 c910f03c09k10 ntpd[7761]:  proto: precision = 0.082 usec (-23)
Nov 22 01:50:56 c910f03c09k10 ntpd[7761]:  Listen and drop on 0 v6wildcard [::]:123
Nov 22 01:50:56 c910f03c09k10 ntpd[7761]:  Listen and drop on 1 v4wildcard 0.0.0.0:123
Nov 22 01:50:56 c910f03c09k10 ntpd[7761]:  Listen normally on 2 lo 127.0.0.1:123
Nov 22 01:50:56 c910f03c09k10 ntpd[7761]:  Listen normally on 3 eth2 10.3.9.10:123
Nov 22 01:50:56 c910f03c09k10 ntpd[7761]:  Listen normally on 4 lo [::1]:123
[...]

PR #4362, opened 2017-11-22
setupntp: do not set exit_code to 1 when the clock is not synced before the timeout

immarvin:
fix issue https://github.com/xcat2/xcat-core/issues/4360:

do not set exit_code to 1 when the clock is not synced before the timeout
add several verbose message and log message

UT:
````
c910f03c09k09:~ # makentp -a
configuring management node: c910f03c09k09.
configuring service nodes: c910f03c09k10
c910f03c09k10: vpdupdate is not supported on the pSeries KVM Guest platform
c910f03c09k10: xcatdsklspost: updating VPD database
c910f03c09k10: xcatdsklspost: downloaded postscripts successfully
c910f03c09k10: Wed Nov 22 04:13:51 EST 2017 Running postscript: setupntp
c910f03c09k10: active
c910f03c09k10: systemctl stop ntpd
c910f03c09k10: syncing the clock ...
c910f03c09k10: WARNING: NTP Sync Failed before timeout. ntp server will try to sync...
c910f03c09k10: postscript: setupntp exited with code 0
c910f03c09k10: Running of postscripts has completed.


c910f03c09k09:~ # echo $?
0
````

### DIFF
--- a/xCAT/postscripts/setupntp
+++ b/xCAT/postscripts/setupntp
@@ -116,11 +116,14 @@ if [ $OS_TYPE = Linux ]; then
          stopservice ntpserver
     fi
 
-    logger -t xcat "setting current time"
+    msg='syncing the clock ...'
+    logger -t xcat $msg 
+    echo $msg
     if ! timeout 120 ntpd -gq  > /dev/null 2>&1 ; then
         if ! ntpdate -t5 $master > /dev/null 2>&1; then
-            logger -t xcat "WARNING: NTP Sync Failed!!"
-            exit_code=1
+            msg='WARNING: NTP Sync Failed before timeout. ntp server will try to sync...'
+            logger -t xcat $msg
+            echo $msg
         fi
     fi
 

### CODE REVIEW

File: xCAT/postscripts/setupntp [hunk 1]
```diff
@@ -116,11 +116,12 @@ if [ $OS_TYPE = Linux ]; then
...
 
-    logger -t xcat "setting current time"
+    logger -t xcat "syncing the clock ..."
+    echo "syncing the clock ..."
     if ! timeout 120 ntpd -gq  > /dev/null 2>&1 ; then
         if ! ntpdate -t5 $master > /dev/null 2>&1; then
-            logger -t xcat "WARNING: NTP Sync Failed!!"
-            exit_code=1
+            logger -t xcat "WARNING: NTP Sync Failed before timeout. ntp server will try to sync..."
+            echo "WARNING: NTP Sync Failed before timeout. ntp server will try to sync... "
```
robin2008: The multiple lines to output to console and syslog is not good.
Not sure if `logger -s` could cover it on all the supported OS.

Or maybe we could have an common func for it, just like
```
log() {
  echo "$@"
  logger -t xcat "$@"
}
```

immarvin: updated

PR #4376, opened 2017-11-27
Add some new cases into and delete outdated test cases from hdctrl_openpower_openbmc.bundle

hu-weihua:
Add some new cases into and delete outdated test cases from ``hdctrl_openpower_openbmc.bundle``.


xuweibj:
Looks good to me.

Issue #4365, opened 2017-11-22
site.master changing


mattaezell:
It seems like every time we upgrade xCAT, our `site.master` gets changed to a different interface on our management node.  The interface is our bmc-in-os port, and it doesn't even have an entry in /etc/hosts or DNS.  It looks like various RPM pre/post install scripts call `xcatconfig`, and I suspect that's where this is happening.
- How can we prevent this from changing when upgrading xCAT?, or
- How do we make it pick the correct address?

whowutwut:
@zet809 @robin2008 can you investigate this issue?

robin2008:
It should be a bug, `xcatconfig` should always use the existing `site.master` first.

robin2008:
In `xcatconfig`, the site table will only be populated with below conditions.
```
    # create basic site definition, if they do not exist reset request
    if (((!-r "/etc/xcat/site.sqlite") && (!-r "/etc/xcat/cfgloc"))
        || $::FORCE
        || $::initDB)
    {
```
The only things there might cause the issue might be `/etc/xcat/cfgloc` and `/etc/xcat/site.sqlite` is not there.

And I have tried that update site.master and run `xcatconfig -u` on an non-HA environment,  it looks the site.master will not be changed during the running of `xcatconfig -u`.

@mattaezell  I assume that you are running on an HA environment, right? 
What's your step to do the xCAT upgrade?  yum update on both MNs in the same time or 
update in on active MN, and then failover to the standby, and update it accordingly. (In HA, the standby Node might not mount /etc/xcat in standby mode)

But I thought this will only impact the local Sqlite DB, if using the PostgreSQL, the site.master should be still the former one.

robin2008:
For the default site.master in `xcatconfig`, it will get the name by `hostname` command, and try to get its IP.  Could you check your MN to see the name resolution is what you expected?

mattaezell:
Hello @robin2008
We have two management nodes on this system, but only one is active.  We use Postgres.

```
[root@mgmt1 db]# hostname
mgmt1.summit.olcf.ornl.gov
[root@mgmt1 db]# host mgmt1.summit.olcf.ornl.gov
mgmt1.summit.olcf.ornl.gov has address 10.10.0.4
[root@mgmt1 db]# host 10.10.0.4
4.0.10.10.in-addr.arpa domain name pointer mgmt1.summit.olcf.ornl.gov.
```

That is the correct IP (.4), and the one that should be in the database.

[...]

PR #4389, opened 2017-11-29
Leave a log file there to debug issue while upgrading

robin2008:
It is related to #4365, and now cannot know the root cause. Here to add a logfile `/var/log/xcat/upgrade.log` to record the `xcatconfig -u -V` output.

UT:
1, manual built xCAT packages
2,run `yum update xCAT` with the new repository
3, check `/var/log/xcat`
```
tail /var/log/xcat/upgrade.log

Imported TERM=xterm-256color into systemd.
Running command on c910f03c05k20: type systemctl >/dev/null 2>&1 && systemctl daemon-reload 2>&1

Restarting xcatd (via systemctl):  [  OK  ]
Running command on c910f03c05k20: /usr/sbin/conserver -V 2>&1

Running command on c910f03c05k20: /bin/cat /etc/conserver.cf | grep sslauthority 2>&1

Running '/opt/xcat/sbin/mknb', triggered by the installation/update of xCAT-genesis-scripts ...
```

### DIFF
--- a/xCAT/debian/postinst
+++ b/xCAT/debian/postinst
@@ -35,7 +35,9 @@ case "$1" in
             if [ -r "/tmp/xcat/mainservice.pid" ]; then
                 mv /tmp/xcat/mainservice.pid /var/run/xcat/mainservice.pid
             fi
-            xcatconfig -u
+            mkdir -p /var/log/xcat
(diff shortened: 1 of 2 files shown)


immarvin:
i am ok with the change

PR #4390, opened 2017-11-29
Listen on 0.0.0.0 instead of the hostname for goconserver

chenglch:
This patch modify the configuration of `makegocons` and `rcons`
for goconserver.

`cat /etc/goconserver/server.conf`
```
global:
  host: 0.0.0.0
  ssl_key_file: /etc/xcat/cert/server-key.pem
  ssl_cert_file: /etc/xcat/cert/server-cert.pem
  ssl_ca_cert_file: /etc/xcat/cert/ca.pem
  logfile: /var/log/goconserver/server.log
api:
  port: 12429
console:
  port: 12430
```

The modification for environment variable used by`congo` from `goconserver` is not included (usually, this is not used by the end user), user should edit it by hand, below is an example:
```
CONGO_PORT=12430
CONGO_URL=https://c910f05c01bc02k74:12429
CONGO_SERVER_HOST=c910f05c01bc02k74
CONGO_SSL_CA_CERT=/root/.xcat/ca.pem
CONGO_SSL_CERT=/root/.xcat/client-cred.pem
CONGO_SSL_KEY=/root/.xcat/client-key.pem
```

### DIFF
--- a/xCAT-client/bin/rcons
+++ b/xCAT-client/bin/rcons
@@ -149,7 +149,8 @@ elif [ $USE_GOCONSERVER == "1" ]; then
                CONGO_SSL_CERT=$HOME/.xcat/client-cred.pem \
                CONGO_SSL_CA_CERT=$HOME/.xcat/ca.pem \
                CONGO_PORT=12430 \
-               CONGO_SERVER_HOST=$CONSERVER"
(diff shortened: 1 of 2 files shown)


neo954:
The patch looks good. Will merge this one.

Issue #4394, opened 2017-11-30
rflash failed when trying to update PNOR on Witherspoon 1742C'' -> 1742Ef


moralesj:
When updating a rack of Witherspoons to 1742Ef, rflash got stuck for ~10 minutes and then failed. 
 Node a01n14 got held up with defect SW387409, which held rflash hostage.  All Witherspoons except for a01n14 were updated correctly.  After a power cycle I was able to rflash the PNOR for that node.

``` # rflash -a a01 witherspoon.pnor.squashfs.tar 
Attempting to upload witherspoon.pnor.squashfs.tar, please wait...
a01n04: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.88 (ID: 78abd177)
a01n08: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.88 (ID: 78abd177)
a01n10: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.88 (ID: 78abd177)
a01n03: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.88 (ID: 78abd177)
a01n05: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.88 (ID: 78abd177)
a01n11: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.88 (ID: 78abd177)
a01n06: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.88 (ID: 78abd177)
a01n07: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.88 (ID: 78abd177)
a01n12: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.88 (ID: 78abd177)
a01n13: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.88 (ID: 78abd177)
a01n15: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.88 (ID: 78abd177)
[...]

PR #4396, opened 2017-11-30
Print out error message if receive nothing when uploading firmware file

xuweibj:
#4394 (currently, not sure whether this is the root cause)

If do not receive response when uploading file, will print out the error msg:

``Error: openbmc plugin bug, pid 25027, process description: 'xcatd SSL: rflash to mid05tor12cn05 for root@localhost: openbmc instance' with error 'malformed JSON string, neither array, object, number, string or atom, at character offset 0 (before "(end of string)") at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 3208.
' while trying to fulfill request for the following nodes: mid05tor12cn05``

After modified:
```
# rflash mid05tor12cn05 obmc-phosphor-image-witherspoon.ubi.mtd.tar -u
Attempting to upload obmc-phosphor-image-witherspoon.ubi.mtd.tar, please wait...
 mid05tor12cn02: Error: Out own error message for command 'curl -b /tmp/_xcat_cjar.mid05tor12cn02 -k -H 'Content-Type: application/octet-stream' -X PUT -T /mnt/xcat/iso/openbmc/910.1746.20171128a_1742Ef/witherspoon.pnor.squashfs.tar https://172.11.139.2/upload/image/'
 mid05tor12cn02: Error: Out own error message ''
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -3218,7 +3218,12 @@ sub dump_download_process {
     my $curl_dwld_cmd = "curl -J -b $cjar_id -k -H 'Content-Type: application/octet-stream' -X GET $request_url/$down_url -o $file_name";
 
     my $curl_login_result = `$curl_login_cmd -s`;
-    my $h = from_json($curl_login_result);
+    my $h;
+    eval { $h = from_json($curl_login_result) };
+    if ($@) {
+        xCAT::SvrUtils::sendmsg([1, "Out own error message for command '$curl_login_cmd'"], $callback, $node);
+        return 1;
+    }
     if ($h->{message} eq $::RESPONSE_OK) {
         xCAT::SvrUtils::sendmsg("Dump $dump_id generated. Downloading to $file_name", $callback, $node);
         my $curl_dwld_result = `$curl_dwld_cmd -s`;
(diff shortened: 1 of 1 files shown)


whowutwut:
@xuweibj We saw some issue with Dump exiting when 1 of the nodes in the noderange exit with error, what happens here when you try to upload and activate across 2 nodes... and there is this failure on one.. does the 2nd one succeed or does it also stop once this first node fails?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -3271,6 +3271,10 @@ sub rflash_upload {
             process_debug_info($node, $debugmsg);
         }
         my $curl_upload_result = `$curl_upload_cmd`;
+        if (!$curl_upload_result) {
```
whowutwut: What do we get back when this fails? 
```
[root@briggs01 Software]# curl -b ./cjar -k -H 'Content-Type: application/octet-stream' -X PUT -T /mnt/xcat/iso/openbmc/910.1746.20171128a_1742Ef/wiherspoon.pnor.squashfs.tar https://172.11.139.16/upload/image/
{
  "data": null,
  "message": "200 OK",
  "status": "ok"
}
```

There's no return code that we are checking? Checking ! on the result is enough?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -3271,6 +3271,10 @@ sub rflash_upload {
...
+            xCAT::SvrUtils::sendmsg("Did not receive response from OpenBMC when uploading $::UPLOAD_FILE ...", $callback, $node);
```
whowutwut: The reason I ask is because this is a similar error message where we cannot say anything to the user about what RC or anything happened? i guess in the above ,we have no data... ? 

Does it make sense to print the command we used? 

If this is an error, should we pass back `(1, ["Did not...` so that it flags Error in the message?

xCAT-server/lib/xcat/plugins/openbmc.pm
xuweibj: I did never encountered it fails. But if it fails, line 3289 will handle it (else branch). But if it's blank, we did not have code to handle, will print out msg like:
```
Error: openbmc plugin bug, pid 25027, process description: 'xcatd SSL: rflash to mid05tor12cn05 for root@localhost: openbmc instance' with error 'malformed JSON string, neither array, object, number, string or atom, at character offset 0 (before "(end of string)") at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 3208. ' while trying to fulfill request for the following nodes: mid05tor12cn05
```

xCAT-server/lib/xcat/plugins/openbmc.pm
xuweibj: I have modified the error msg, do you think it's helpful?

whowutwut:
@xuweibj How did we re-create this on the node, FW dev team needs a dump to see what is going on and how we get an empty string.

xuweibj:
@whowutwut I just force the code to go into the error path, actually I did not reproduce it.

whowutwut:
@xuweibj Thanks,  I'd @gurevichmark  to take a quick look at this one since he write some of that upload code. And can merge tomorrow.

xCAT-server/lib/xcat/plugins/openbmc.pm
gurevichmark: Message like that would make it really hard to debug. Is it possible to test `$curl_upload_result` to see if it is in json format before attempting to parse it with `from_json($curl_upload_result)` ?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -3271,6 +3271,10 @@ sub rflash_upload {
             process_debug_info($node, $debugmsg);
         }
         my $curl_upload_result = `$curl_upload_cmd`;
+        if (!$curl_upload_result) {
+            xCAT::SvrUtils::sendmsg("Error: Did not receive response from OpenBMC when run command '$curl_upload_cmd'", $callback, $node);
```
gurevichmark: 1. Suggest changing "when run" to "after running" in the message.
2. Would it be better to make this an error message instead of adding "Error:" to the message text. Something similar to `xCAT::SvrUtils::sendmsg([1, "Did not receive...."]....`

xCAT-server/lib/xcat/plugins/openbmc.pm
xuweibj: @gurevichmark I tried to find the method to check the format whether is JSON, but did not find. So just check whether empty here.

gurevichmark: What if we do something like:
```
eval { from_json(...) };
if ($@) {
    xCAT::SvrUtils::sendmsg([1, "Out own error message"], ...);
}
```

OR 

I noticed many placed in openbmc.pm module, we use `decode_json`, is that better than `from_json` ?

whowutwut:
I think some merge conflicts resolution made the changes in this PR really hard to read and looks like there is a lot more changes than there really is with this PR?  

Is it better to close this and re-create another PR ?

xuweibj:
@whowutwut I just modified format in "rflash_upload" more.

gurevichmark:
@xuweibj Few more comments:
1. I think we should still check that the string returned from the `curl` command is not empty and display the original message `xCAT::SvrUtils::sendmsg([1, "Did not receive response from OpenBMC after running command '$curl_xxxxx_cmd'", $callback, $node);`, then, if something is returned, we can do the `eval` check and display the new `xCAT::SvrUtils::sendmsg([1, "Received wrong format response from..."]` new message.

2. I do not like the two `xCAT::SvrUtils::sendmsg()`, one after the other when the `eval` for upload command fails. Perhaps they can be combined into a single message, like
`"Received wrong format response from command '$upload_cmd': '$curl_upload_result'"`

gurevichmark:
@xuweibj The section for processing the upload command looks good to me now. But I think you should do the same for the 2 login commands also - `my $curl_login_result = $curl_login_cmd -s;`. check first if anything is returned, then, if we do get something back, check it with `eval` and display what we get back in the error message by including `$curl_login_result`, similar to upload command.

gurevichmark:
Looks good to me. I am ok to merge if no additional comments from @whowutwut

whowutwut:
@gurevichmark I don't have any additional comments, The description still contains the old output, but that should not stop the merge.

Issue #4404, opened 2017-11-30
rspconfig dump handling of timeout 


gurevichmark:
Continuation of work for PR #4382 

10 min is probably too long to wait for the dump to be generated. Perhaps 5 min is better, and every 2 min message should be displayed that we are still waiting, otherwise user might think the command hung.

PR #4410, opened 2017-11-30
OpenBMC rspconfig dump timeout fixes

gurevichmark:
Fixes issue #4404 

1. Reduce timeout to 5 min
2. Print "Still processing..." message every 2 min.

```
[root@briggs01 xCAT_plugin]# rspconfig mid05tor12cn16 dump
Capturing BMC Diagnostic information, this will take some time...
mid05tor12cn16: Dump requested. Target ID is 32, waiting for BMC to generate...
mid05tor12cn16: Still waiting for dump 32 to be generated...
mid05tor12cn16: Still waiting for dump 32 to be generated...
mid05tor12cn16: Still waiting for dump 32 to be generated...
mid05tor12cn16: Error: Could not find dump 32 after waiting 300 seconds.
[root@briggs01 xCAT_plugin]#
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -65,7 +65,7 @@ $::UPLOAD_WAIT_TOTALTIME    = int($::UPLOAD_WAIT_ATTEMPT*$::UPLOAD_WAIT_INTERVAL
 $::RPOWER_CHECK_INTERVAL    = 2;
 $::RPOWER_MAX_RETRY         = 30;
 
-$::RSPCONFIG_DUMP_INTERVAL  = 30;
+$::RSPCONFIG_DUMP_INTERVAL  = 15;
 $::RSPCONFIG_DUMP_MAX_RETRY = 20;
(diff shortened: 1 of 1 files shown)


whowutwut:
@gurevichmark Just for information for those monitoring.... On a success case, the Dump is generated within 1 minute, so we don't see these messages, correct?

whowutwut:
delegate to @xuweibj

gurevichmark:
The message "Still processing..." gets displayed every 2 min, so successful generation takes less than that.

xuweibj:
The initial value is 20, so the first time print out this message maybe after 1 minute?

whowutwut:
@xuweibj from the comments in the change, the interval changed to 15 seconds * 8 iterations so  120 seconds.   The idea of this PR is when there is problem downloading the dumps, there's nothing printed by xCAT for 10 minutes, so we wanted to pick an interval where most of the Dumps will already be ready and downloaded, and then the error ones will print out each 2 minutes and inform the user sitting there that we are still waiting on something (normally in this case there is a problem on the BMC)

PR #4419, opened 2017-12-01
Use short hostname in rcons for goconserver

chenglch:
As the certificate of xcat is signed with short hostname, this
commit force to use the short hostname in  the environment variable for
`congo console`.


chenglch:
More changes is required, close it at first

PR #4428, opened 2017-12-02
NODE attribute didn't populate in /opt/xcat/xcatinfo after reboot

cxhong:
This is for issue https://github.ibm.com/DCS-research/WSC-coral/issues/697

After node reboot, the NODE attribute didn't populate in the /opt/xcat/xcatinfo
````
]# xdsh c699wrk01 cat /opt/xcat/xcatinfo
c699wrk01: XCATSERVER=10.3.100.41
c699wrk01: INSTALLDIR=/install
c699wrk01: REBOOT=TRUE
c699wrk01: NODE=
c699wrk01: USEFLOWCONTROL=NO
````
in the `xcatdsklspost` postscripts,  it looks for /proc/cmdline to get NODE info, in most case, the /proc/cmdline didn't contain the NODE info.  so in the PR, will check NODE variable before add to /opt/xcat/xcatinfo, make sure NODE variable is not empty.

### DIFF
--- a/xCAT/postscripts/xcatdsklspost
+++ b/xCAT/postscripts/xcatdsklspost
@@ -24,6 +24,7 @@
 if [ -f /xcatpost/mypostscript.post ]; then
     XCATDEBUGMODE=`grep 'XCATDEBUGMODE=' /xcatpost/mypostscript.post | cut -d= -f2 | tr -d \'\" | tr A-Z a-z`
     MASTER_IP=`grep '^MASTER_IP=' /xcatpost/mypostscript.post |cut -d= -f2|sed s/\'//g`
+    NODE=`grep '^NODE=' /xcatpost/mypostscript.post |cut -d= -f2|sed s/\'//g`
 else
     for param in `cat /proc/cmdline`; do
             key=`echo $param|awk -F= '{print $1}'`
--- a/xCAT/postscripts/xcatdsklspost
+++ b/xCAT/postscripts/xcatdsklspost
@@ -381,6 +381,9 @@ else # for common mode  MODE=1,2,3,5 (updatenode,moncfg,node deployment)
             break
         fi
      done
+     if [ -z "$NODE" ]; then
+         NODE=`hostname -s`
+     fi
      grep 'NODE' /opt/xcat/xcatinfo > /dev/null  2>&1
      if [ $? -eq 0 ]; then
         sed -i "s/NODE=.*/NODE=$NODE/" /opt/xcat/xcatinfo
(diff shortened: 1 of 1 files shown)


zet809:
Hi, @cxhong , the `hostname -s` is used here to get the nodename of the CN, do you think we shall get the nodename of CN from mypostscript on CN?

cxhong:
when I ran test,  I found mypostscripts file was not there sometime.  that's why used `hostname -s`.

zet809:
So, do you think we shall use the nodename in mypostscript first, if there is no mypostscript, then we can use `hostname -s`.

immarvin:
hi @cxhong , what is this PR for? diskless or diskful?

cxhong:
@zet809 , I will make the change
@immarvin , it's for diskfull

immarvin:
hi @cxhong , I think the purpose of this PR is not clear
1. the "NODE=" will be populated in /opt/xcat/xcatinfo file on diskful provision
2. the "NODE=" will be populated in  /opt/xcat/xcatinfo file on updatenode against the diskful node

which one is the expected behavior of this PR?

cxhong:
@immarvin , we are seeing this issue in our test system and WSC cluster.  reboot diskfull nodes will cause node name in the /opt/xcat/xcatinfo change to `NODE=`

immarvin:
hi @cxhong , the value of `NODE` should be obtained from the newly generated `/xcatpost/mypostscript`, after line 647 in the following code slice in `xcatdsklspost`
````
639     fi
640     /$xcatpost/getpostscript.awk | egrep  '<data>' | sed  -e 's/<[^>]*>//g'|egrep -v '^ *$'|sed -e 's/^ *//' | sed -e 's/&lt;/    </g' -e 's/&gt;/>/g' -e 's/&amp;/\&/g' -e 's/&quot;/"/g' -e "s/&apos;/'/g" >  /$xcatpost/mypostscript;
641     MYCONT=`grep MASTER /$xcatpost/mypostscript`
642     if [ ! -z "$MYCONT" ]; then
643         break;
644     fi
645  done
646 fi
647
648 #save the MASTER into the xcatinfo file for node deployment case,
649 #for updatenode case, only save it when -M is specified
650 if [ $NODE_DEPLOYMENT -eq 1 ] || [ "$MODE" = "4" ]; then
651
````

cxhong:
currently, after line 647, the NODE is updated only in the mode 1 and mode 2.   do u think we should update NODE info for all the mode?

immarvin:
what is this PR for?  fix the issue that "NODE=" is blank after `updatenode -P/-S`?

if yes, I think we should just get ''NODE" from the newly obtained /xcatpost/mypostscript after line 647 and leave the other places untouched

cxhong:
@immarvin , this issue is not caused by updatenode or provision.    it is caused by diskfull reboot.

PR #4429, opened 2017-12-02
Add dhcp-client-identifier to lease block

russauld:
Machines that use Infiniband for PXE booting need to have the
dhcp-client-identifier set in the lease block.
Without it, they will not get the lease from the server.

This change adds dhcp-client-identifier along side the hardware address.
The line appears as "uid <mac>" in the lease file.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -747,6 +747,7 @@ sub addnode
             print $omshell "new host\n";
             print $omshell "set name = \"$hostname\"\n";
             print $omshell "set hardware-address = " . $mac . "\n";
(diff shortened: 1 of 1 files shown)


zet809:
Hi, @immarvin , will you pls take a look at this PR? Thx!

immarvin:
i am ok with the change

whowutwut:
@russauld  I know you have been using xCAT from some time and seen many comments on the user list! Thanks! 

Trying to be good about this, we do require contributors to sign a CLA http://xcat-docs.readthedocs.io/en/stable/developers/license/contributors.html and trying to do our best to keep this up to date.

If you are covered by a Corporate CLA, could you let us know the company and I can add you into the list...email address is on the pdf.

whowutwut:
@russauld FYI, we got the CLA, thank you very much.

PR #4432, opened 2017-12-04
Use short hostname in rcons for goconserver

chenglch:
As the certificate of xcat is signed with short hostname, this
commit force to use the short hostname in  the environment variable for
`congo console`.

### DIFF
--- a/xCAT-client/bin/rcons
+++ b/xCAT-client/bin/rcons
@@ -148,13 +148,26 @@ elif [ $USE_GOCONSERVER == "1" ]; then
     CONGO_ENV="CONGO_SSL_KEY=$HOME/.xcat/client-cred.pem \
                CONGO_SSL_CERT=$HOME/.xcat/client-cred.pem \
                CONGO_SSL_CA_CERT=$HOME/.xcat/ca.pem \
-               CONGO_PORT=12430 \
-               CONGO_SERVER_HOST=$CONSERVER \
-               CONGO_URL=https://$CONSERVER:12429"
-
+               CONGO_PORT=12430"
     if [ "$CONSERVER" == `hostname` ]; then
+        host=`hostname -s`
+        if [ $? -ne 0 ]; then
+            echo "Error: could not get the short hostname for $CONSERVER."
+            exit 1
+        fi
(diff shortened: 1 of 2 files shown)


xCAT-client/bin/rcons
```diff
@@ -148,13 +148,26 @@ elif [ $USE_GOCONSERVER == "1" ]; then
     CONGO_ENV="CONGO_SSL_KEY=$HOME/.xcat/client-cred.pem \
                CONGO_SSL_CERT=$HOME/.xcat/client-cred.pem \
                CONGO_SSL_CA_CERT=$HOME/.xcat/ca.pem \
-               CONGO_PORT=12430 \
-               CONGO_SERVER_HOST=$CONSERVER \
-               CONGO_URL=https://$CONSERVER:12429"
-
+               CONGO_PORT=12430"
     if [ "$CONSERVER" == `hostname` ]; then
```
zet809: Shall here be `hostname -s` also?

xCAT-server/lib/perl/xCAT/Goconserver.pm
```diff
@@ -26,7 +27,7 @@ sub http_request {
             SSL_cert_file   => xCAT::Utils->getHomeDir() . "/.xcat/client-cred.pem",
             SSL_ca_file     => xCAT::Utils->getHomeDir() . "/.xcat/ca.pem",
             SSL_use_cert    => 1,
-            SSL_verify_mode => 'SSL_VERIFY_PEER',  }, );
+            SSL_verify_mode => SSL_VERIFY_PEER,  }, );
```
zet809: What is the difference here?

chenglch: The original one can not work on ubuntu 16.04 system, this is more standard way.

xCAT-client/bin/rcons
chenglch: No, $CONSERVER has another responsibility,   it get the value above and will be used by ssh to login. The short name is only used for `congo` as the xcat certificate is registered with the short name.

zet809:

PR #4434, opened 2017-12-04
Fix confignetwork bond nic_type detection with multiple bonds

### LINKED ISSUE #4433: confignetwork does not support multiple bonds of different nic type
Our login nodes have an Ethernet and Infiniband bond.

I'm getting the error:
```
login1: [E]:Error: [E]: different nic device types in ib0@ib1.
```

I think I see the problem, patch coming soon

mattaezell:
When there is more than one bond, the `temp_base_nic_type_one` variable needs to be reset when evaluating each bond.
Closes #4433


whowutwut:
Assigned to @bybai to review

bybai:
I am ok to merge it.

Issue #4300, opened 2017-11-14
[Coral] issues for ntp configuration on Accton switches


cxhong:
Issues found from lab:
https://w3.rchland.ibm.com/projects/bestquest/?defect=SW408125
   --- incorrect sets NTP server
https://w3.rchland.ibm.com/projects/bestquest/?defect=SW408126
    -- incorrect iburst entry
https://w3.rchland.ibm.com/projects/bestquest/?defect=SW408127
   --didn't update timezone

for configonie --ntp,  we should replace code to use `setupntp` postscripts, or get rid of this command and update the documentation to suggest  user to use `setupntp` postscripts to set ntp on the accton switches.

immarvin:
how to export the environment variables to setupntp during `configonie`? maybe we can consider leverage getmyposttscript.cumulus to generate the mypostscript and run the specified script, just like what we do for the updatenode against node, will look into this...

PR #4436, opened 2017-12-04
Modified configonie --ntp command

cxhong:
for issue #4300 .
1) pull in timezone from site table.
2) use the ntpservers as this order:
     ---- use ntpserver defined in the network table 
     ---- use ntpservers defined in the site table
     ---- use master as ntpservers defined in the site table


immarvin:
i am ok with the change

Issue #4426, opened 2017-12-01
makedhcp -d not working


mattaezell:
I haven't tried this before today, so maybe I'm doing something wrong.  Should this work?

```
[root@mgmt1 ~]# makedhcp -d a05n13
[root@mgmt1 ~]# echo $?
0
[root@mgmt1 ~]# grep a05n13 /var/lib/dhcpd/dhcpd.leases
host a05n13 {
        supersede server.ddns-hostname = "a05n13";
        supersede host-name = "a05n13";
        supersede conf-file = "http://10.10.0.11/tftpboot/petitboot/a05n13";
```

mattaezell:
Maybe we have a larger issue with dhcp...
```
[root@mgmt1 ~]# chdef a05n13 mac=aa:bb:cc:dd:ee:aa
1 object definitions have been created or modified.
[root@mgmt1 ~]# lsdef -i mac a05n13
Object name: a05n13
    mac=aa:bb:cc:dd:ee:aa
[root@mgmt1 ~]# nodeset a05n13 osimage=compute
a05n13: netboot rhels7.4-ppc64le-compute
[root@mgmt1 ~]# grep aa:bb:cc:dd:ee:aa /var/lib/dhcpd/dhcpd.leases
[root@mgmt1 ~]# 
```

whowutwut:
@mattaezell For the 2nd comment , you have to run a `makedhcp a05n13` after setting the mac for it to get added into the lease file.   This happens automatically under the covers when we are doing discovery and fill in the mac address on a predefined node. 

```
chdef a05n13 mac=aa:bb:cc:dd:ee:aa
makedhcp a05n13
```

Then grep for it.

mattaezell:
`nodeset` calls `makedhcp` also.  We've been running with `disjointdhcps=1` from back when we thought we were going to run DHCP on all the service nodes.  It seems like now that's preventing this from working.  It's sending the `makedhcp` commands to the service node, but they aren't doing anything or logging any errors.

mattaezell:
I think things are working OK with `disjointdhcps=0`.  I can't explain why/how this was working previously.  Is `disjointdhcps=0` the recommendation on our system?

whowutwut:
@mattaezell humm.. ok I had a suspicion that something with hierarchy and disjoint may be causing a problem, but wanted to talk to @robin2008 about it..   @robin2008 could you chime in? 

[...]

PR #4440, opened 2017-12-05
makedhcp does not work well when disjointdhcps=1 

robin2008:
makedhcp does not work well when all service nodes not running dhcp but disjointdhcps=1 (#4426)

- if all service nodes not running dhcp, to treat it as disjointdhcps=0
- nodeset will send request to MN by default even if disjointdhcps=1
- Move out of the dhcp service checking from opts pre-check, and do it just before real makedhcp handling.

### DIFF
--- a/perl-xCAT/xCAT/Scope.pm
+++ b/perl-xCAT/xCAT/Scope.pm
@@ -219,6 +219,10 @@ sub get_broadcast_disjoint_scope_with_parallel {
     my $reqs = get_parallel_scope($req);
 
     my $handled4me = 0;  # indicate myself is already handled.
+    if (xCAT::Utils->isMN()) { # For MN, add itself always.
+        push @requests, @$reqs;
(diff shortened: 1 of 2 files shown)


immarvin:
i am ok with the change

robin2008:
disjointdhcps=1, sharedtftp=0, all SNs no dhcp running:

```
nodeset mid05tor12cn01 osimage -V
mid05tor12cn01: [briggs01]: install rhels7.4-rc1-ppc64le-compute
mid05tor12cn01: [sn02]: install rhels7.4-rc1-ppc64le-compute
.
[root@briggs01 xcat]# makedhcp -q mid05tor12cn01 -V
[briggs01]: mid05tor12cn01: ip-address = 172.12.139.1, hardware-address = 11:22:33:44:55:66

# makedhcp -d mid05tor12cn01
# makedhcp -q mid05tor12cn01 -V
#

```

PR #4445, opened 2017-12-05
Add support for the "file -> (noderange) file" syntax in synclist with ServiceNodes

### LINKED ISSUE #4425: Hierarchical mode support with nodenames in synclist
Hi!

It looks like having nodenames in synclist doesn't work in hiearchical mode.

I have a syncfile that contains:
```
/tmp/a   -> (sh-101-60)     /tmp/a
/tmp/b   -> (sh-6-34)       /tmp/b
```

`sh-101-60` is directly managed by the management node, while `sh-6-34` is managed by a service node:
```
# lsdef -c sh-101-60,sh-6-34 -i servicenode
sh-101-60: servicenode=
sh-6-34: servicenode=sh-hn03
```

Distributing files on `sh-101-60` (direct mode) works normally:
```
# updatenode sh-101-60 -F
# md5sum /tmp/a && ssh sh-101-60 md5sum /tmp/a
c563744e1d6d090285937d352e3128db  /tmp/a
c563744e1d6d090285937d352e3128db  /tmp/a
```
but it fails on `sh-6-34` (hierarchical mode):
```
# updatenode sh-6-34 -f
File synchronization has completed for service nodes.
# updatenode sh-6-34 -F
sh-6-34: Permission denied (publickey,hostbased).
sh-6-34: /tmp/rsync_sh-6-34: line 3: syntax error near unexpected token `('
sh-6-34: /tmp/rsync_sh-6-34: line 3: `/usr/bin/rsync --rsync-path /usr/bin/rsync -Lprogtz  /var/xcat/syncfiles/tmp/b sh-hn01.SUNet@sh-6-34:(sh-6-34)/tmp'
```

The problem is that the `rsync` script generated for the hierarchical mode fails to filter synclist entries based on hostnames and includes them all, with an erroneous syntax:
```
# ssh sh-hn03 cat /tmp/rsync_sh-6-34
#!/bin/sh
/usr/bin/ssh  sh-hn01.SUNet@sh-6-34 '/bin/mkdir -p (sh-6-34)/tmp (sh-101-60)/tmp'
/usr/bin/rsync --rsync-path /usr/bin/rsync -Lprogtz  /var/xcat/syncfiles/tmp/b sh-hn01.SUNet@sh-6-34:(sh-6-34)/tmp
/usr/bin/rsync --rsync-path /usr/bin/rsync -Lprogtz  /var/xcat/syncfiles/tmp/a sh-hn01.SUNet@sh-6-34:(sh-101-60)/tmp
```
Note how the generated `rsync` command still includes the destination hostname inside parenthesis as the destination path for the file.

Could you please provide a fix for this? This is currently making the hierarchical mode pretty unusable for us. Thanks!

kcgthb:
Fixes #4425

This patch ensures that:
1. the synclist is correctly parsed when running on a Service Node
2. all files from the synclist are synchronized to the SN in hierarchical mode

See #4425 for more details about the original issue.

A more elegant solution may be to only synchronize files that are meant to be copied on nodes that a given ServiceNode services, but that seems to require adding a lot of overhead in the code.

### DIFF
--- a/perl-xCAT/xCAT/DSHCLI.pm
+++ b/perl-xCAT/xCAT/DSHCLI.pm
@@ -5181,8 +5181,10 @@ sub parse_rsync_input_file_on_MN
 
                 foreach my $target_node (@dest_host)
                 {
-                    # skip the node if it's NOT in the permitted list
-                    if ($dest_node && !grep /^$target_node$/, @dest_nodes) {
+                    # skip the node if it's NOT in the permitted list and if
+                    # it's not a SN
+                    if ($dest_node && !(grep /^$target_node$/, @dest_nodes) &&
+                        !(xCAT::Utils->isSN($target_node)) ) {
                         next;
                     }
                     $$options{'destDir_srcFile'}{$target_node} ||= {};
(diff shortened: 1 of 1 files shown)


kcgthb:
Argh, there's a problem when multiple node-specific files have the same destination on service nodes, like this:

```
/path/src.sn1 -> (sn1) /path/dest
/path/src.sn2 -> (sn2) /path/dest
```

In that case, with this #PR, `/path/dest` gets overwritten with `src.dst2` on both service nodes...

kcgthb:
Last commit addresses the issue mentioned above, so this PR should be good to review.

immarvin:
hi @kcgthb , sorry for late response, I am still looking into the DSHCLI code and reviewing this PR

immarvin:
i am ok with the change

PR #4453, opened 2017-12-06
Support multiple bonds on bring-up

mattaezell:
If there are multiple bonds enabled, the previous code would fail to detect that the second bond device had been created.


whowutwut:
@bybai can you review this?

bybai:
I am ok to merge it. Thanks @mattaezell

PR #4459, opened 2017-12-07
Ignore syslog error in monitorctrl when setNodeStatusAttributes

xuweibj:
Issue:
```
Error: openbmc plugin bug, pid 30251, process description: 'xcatd SSL: rpower for mgmt1.summit.olcf.ornl.gov@mgmt1: openbmc instance' with error 'no connection to syslog available        - unix dgram connect: Connection refused        - stream can't open /dev/log: No such device or address at /opt/xcat/lib/perl/xCAT_monitoring/monitorctrl.pm line 707.' while trying to fulfill request for the following nodes: b01n01
--
```

Solution:
If 'syslog' failed, do not exit.

### DIFF
--- a/xCAT-server/lib/xcat/monitoring/monitorctrl.pm
+++ b/xCAT-server/lib/xcat/monitoring/monitorctrl.pm
@@ -704,9 +704,7 @@ sub setNodeStatusAttributes {
                 $updates{'status'}     = $_;
                 $updates{'statustime'} = $currtime;
                 my $nodestate = "@$nodes status: $updates{'status'} statustime: $updates{'statustime'}";
-                openlog('xcat', 'ndelay', 'local4');
-                syslog('local4|info', '%s', $nodestate);
-                closelog();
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/monitoring/monitorctrl.pm
```diff
@@ -705,7 +705,7 @@ sub setNodeStatusAttributes {
                 $updates{'statustime'} = $currtime;
                 my $nodestate = "@$nodes status: $updates{'status'} statustime: $updates{'statustime'}";
                 openlog('xcat', 'ndelay', 'local4');
```
immarvin: the logic should be 
```
eval { openlog,syslog,closelog }; 
if ($@) { print "$@" } 
```

immarvin: this logic has been covered in 
```
xCAT::MsgUtils->message('S', "xxxx");
```
please use this instead

immarvin:
i am ok with the change

Issue #4301, opened 2017-11-14
[Coral]  issues for install license file on accton switches.


cxhong:
https://w3.rchland.ibm.com/projects/bestquest/?defect=SW408131
https://w3.rchland.ibm.com/projects/bestquest/?defect=SW408130

after upgrade the cumulus OS,  xcat suggest to install license file via xdsh command.  if user invoke `switchdiscover --setup` or `configonie --all` command,  the configuration of interface will failed since license is not install yet at that time.

immarvin:
Specifying the licenses for switches in a network range on `switchdiscover ` is difficult for us.  If user upgrade cumulus os with fresh installation, we can leverage ztp to install the license, I will draft a design on this

PR #4460, opened 2017-12-07
issues for install license file on accton switches

cxhong:
for issue #4301 .
1) checking installed license  via cl-license command
2) restart switch and reload interface after install license files.


immarvin:
i am ok with the change

PR #4461, opened 2017-12-07
modify response for bmcdiscover when error

xuweibj:
Before modify:
```
# bmcdiscover --range 10.3.5.99
Warning: HTTP::Response=HASH(0x100269b6fc8)->status_line for 10.3.5.99
```

After modified:
```
# bmcdiscover --range 10.3.5.99
Warning: Received response 500 read timeout for 10.3.5.99
```


zet809:

Issue #4443, opened 2017-12-05
[OpenBMC]  Clearing BMC Diagnostics message should only be printed when reboot BMC and there is a pending update for BMC (not PNOR)


whowutwut:
Running build, maybe with some dev code on top.... 
```
[root@briggs01 910.1746.20171128a_1742Ef]# lsxcatd -a
Version 2.13.9 (git commit 0743b0d82644eb2196284aa330f90b3a16e8923c, built Fri Dec  1 06:15:42 EST 2017)
This is a Management Node
```

I AM not flashing BMC but flashing PNOR...

```
[root@briggs01 910.1746.20171128a_1742Ef]# rflash mid05tor12cn02 -l
Tue Dec  5 12:05:19 2017 mid05tor12cn02: [openbmc_debug] rflash_list_response 200 OK
mid05tor12cn02: ID       Purpose State      Version
mid05tor12cn02: -------------------------------------------------------
mid05tor12cn02: 6e71e1af BMC     Active     ibm-v1.99.10-113-g65edf7d-r10-0-gcdf7635
mid05tor12cn02: 3a7f1407 BMC     Active(*)  ibm-v2.0-0-r13-0-g3a160ac
mid05tor12cn02: 78abd177 Host    Active(+)  IBM-witherspoon-ibm-OP9_v1.19_1.88    <---- pending HOST
mid05tor12cn02: 8a15b806 Host    Active(*)  IBM-witherspoon-ibm-OP9_v1.19_1.73
mid05tor12cn02:
```

But when doing a bmcreboot, the message comes out regarding Clear of Diagnostics...

```
[root@briggs01 910.1746.20171128a_1742Ef]# rpower mid05tor12cn02 bmcreboot
Tue Dec  5 12:05:25 2017 mid05tor12cn02: [openbmc_debug] curl -k -c cjar -H "Content-Type: application/json" -d '{ "data": ["root", "xxxxxx"] }' https://mid05tor12cn02-bmc/login
Tue Dec  5 12:05:26 2017 mid05tor12cn02: [openbmc_debug] login_response 200 OK
Tue Dec  5 12:05:26 2017 mid05tor12cn02: [openbmc_debug] curl -k -b cjar -X GET -H "Content-Type: application/json" https://mid05tor12cn02-bmc/xyz/openbmc_project/software/enumerate
Tue Dec  5 12:05:26 2017 mid05tor12cn02: [openbmc_debug] rinv_firm_response 200 OK
[...]

PR #4470, opened 2017-12-07
Do not display message for clearing dumps when only PNOR

gurevichmark:
#4443 

Check setting of the flag `$to_clear_dump` before printing the message. When only PNOR update, that flag is not set.


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -2464,7 +2464,9 @@ sub rinv_response {
         }
     } else {
         if ($status_info{RINV_FIRM_RESPONSE}{check}) {
-            xCAT::MsgUtils->message("I", { data => ["$node: Firmware will be flashed on reboot, deleting all BMC diagnostics..."] }, $callback);
+            if ($to_clear_dump) {
```
whowutwut: Ah, so this flag was there, but we never had a check to print the message? I didn't check the debug logs carefully, did it actually send any clear dump command?

gurevichmark: Correct, the "dump delete" was never sent, we just displayed the message by mistake. However, my attempts to recreate this problem discovered a functional problem with `@sorted_output`. @xuweibj provided the fix, which I included in this PR.

PR #4478, opened 2017-12-08
Modify the default consoleondemand based on the global setting

chenglch:
This commit fix the bug that consoleondemand works incorrectly.


zet809:
Looks good to me, agree to merge, thx!

whowutwut:
@zet809 just a friendly reminder to ensure that milestones are set when the PR is merged (at a minimum), this would help us more easily track changelog for each release...  I just set it.

Issue #4477, opened 2017-12-08
[OpenBMC] bmcdiscover should only create temporary node definitions using node-<mac> format. 


whowutwut:
As we start monitoring some of the bring up going on with the OpenPower servers running OpenBMC. I think we should NOT support creating temporary node definitions for OpenBMC servers in the format of `node-<mode>-<sn>`.  

Before the host is booted for the very first time, the boxelder board doesn't have any information about the VPD model/serial... and so the only way we could detect a unique node is creating the temporary node based on the MAC address.  

This all works fine, but if the user runs `bmcdiscover` on a node range where the host has been booted, then writing those to the xCAT database would result in a definition called `node-<model>-<serial>` and now this is confusing.  There's 2 node definitions pointing to the same machine. 

To prevent this, we should be consistent and only support 1 format.  
@zet809 Also would affect clean up of the temporary node definition when discovery is made.

zet809:
hi, @xuweibj , will you pls help to take a look at this issue?
 [ ] check whether both node-$mac and node-$mtm-$sn can be removed during the hardware discovery process
 [ ] update or create node-$mac only for OpenBMC node. If no $mac found (such as remote OpenBMC), if $mtm and $sn found, create node-$mtm-$sn. Otherwise, report error that no MAC or $mtms found for the IP(Such as a remote OpenBMC haven't been booted yet).

whowutwut:
I'm not sure how to validate this because our nodes are powered on and the BMCs see this information already there, need to think about this... . 

Just updated to todays build: 
```
Version 2.13.10 (git commit 730d4550413c8851dfb7d8681c26872cf3ad4917, built Wed Dec 20 06:16:03 EST 2017)
```

and seems to still work:

```
[root@briggs01 yum.repos.d]# bmcdiscover --range 172.12.253.100
172.12.253.100,8335-GTC,1318C4A,,,mp,bmc,,
[root@briggs01 yum.repos.d]# bmcdiscover --range 172.12.253.100 -w
Writing node-8335-gtc-1318c4a (172.12.253.100,8335-GTC,1318C4A,,,mp,bmc,,) to database...
[root@briggs01 yum.repos.d]# rsetboot node-8335-gtc-1318c4a net
node-8335-gtc-1318c4a: Network
[root@briggs01 yum.repos.d]# rpower node-8335-gtc-1318c4a reset
```

whowutwut:
[...]

PR #4496, opened 2017-12-12
Fix issue 4477, if has node-<mac> will not create node-<mtms> for the same node

xuweibj:
#4477 

If has node named "node-<mac>", update node-<mac> data but not create new node definition.

```
# bmcdiscover --range 172.11.139.2 -w
172.11.139.2: Detected openbmc, attempting to obtain system information...
Writing node-70e284142897 (172.11.139.2,8335-GTC,1318D3A,,,mp,bmc,,) to database...
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -50,6 +50,8 @@ my $openbmc_pass;
 $::P9_WITHERSPOON_MFG_ID     = "42817";
 $::P9_WITHERSPOON_PRODUCT_ID = "16975";
 
+my %node_in_list = ();
+
 #-------------------------------------------------------
 
 =head3  handled_commands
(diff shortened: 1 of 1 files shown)


zet809:

whowutwut:
@xuweibj Can you clarify the behavior... 

If there is already a `node-mac` in the xCAT database, we would update that with the information from bmcdiscover? 

What if there is no `node-mac` and we have MT and SN information?  It would write `node-MT-SN` instead?

xuweibj:
@whowutwut 
The login as below:

1. There is no defined mtms_node and mac_node
    1.1 we get all information including mtms and mac, create mtms_node
    1.2 we get only mtms or mac information, create mtms_node or mac_node
2. There is defined mtms_node, only update mtms_node
3. There is defined mac_node, update the mac_node even we get mtms information
4. There are defined mtms_node and mac_node, update mtms_node

whowutwut:
@xuweibj  
I think for 1.2, you mean "If we only get mac information, create mac_node"? 

Then for 4. I guess we will not fix this scenario right?  How will the clean up happen? 
If there is mtms_node and mac_node, and we update mtms_node and discovery will clean up mtms_node, we will still be left with mac_node right?


 I guess this is OK... :)  hopefully moving forward, this will prevent both nodes from showing up on the xCAT node definition.

xuweibj:
@whowutwut  
Sorry, you are right, I made typo in 1.2.
For 4, if discovery success, will clean up all mtms_node and mac_node. This have been solved before.

whowutwut:
Thanks!

Issue #2423, opened 2017-01-19
Customize /etc/named.conf


kcgthb:
Hi!

I'm wondering if there's any way in `makedns` to customize the generated `named.conf` file? 

I'm thinking including custom, site-specific directives, such as custom configuration options, logging directives, or zone delegation? It could be done by having a `/etc/named.conf.local` file whose contents would be appended to the generated `/etc/named.conf` file.

The same principle could ideally be also applied to the generation of `/etc/hosts` or `/etc/dhcpd.conf`

Any thoughts?

kcgthb:
Actually, a mechanism to include external configuration is kind of required, as the generated `named.conf` file currently misses some of the required configuration.

For instance, Redhat provides a `/etc/named.rfc1912.zones` file which provides zone configuration for zones recommended by RFC 1912. This is missing from the `named.conf` file generated by `makedns`. As a result, resolving `localhost` is broken:
```
$ host 127.0.0.1 127.0.0.1
Using domain server:
Name: 127.0.0.1
Address: 127.0.0.1#53
Aliases:

Host 1.0.0.127.in-addr.arpa. not found: 3(NXDOMAIN)
```
With the proper `include "/etc/named.rfc1912.zones";` line in `/etc/named.conf`, `localhost` resolves properly:
```
host 127.0.0.1 127.0.0.1
Using domain server:
Name: 127.0.0.1
Address: 127.0.0.1#53
Aliases:

1.0.0.127.in-addr.arpa domain name pointer localhost.
```

So, could you please provide a mechanism in `makedns` to include external configuration files?

bybai:
Hi @kcgthb , we plan to add this feature into next release. Thanks.

whowutwut:
@bybai  Please set a milestone when you are ready to look at this issue.

kcgthb:
Just checking to see if any plans have been made to implement customization of `/etc/named.conf`?

zet809:
Let's close this issue.

kcgthb:
Sorry if I missed it, but has this feature been implemented?

PR #4497, opened 2017-12-12
Crude attempt at including external configuration files in named.conf

kcgthb:
This is a first try at including external configuration files in the `/etc/named.conf` file that xCAT generates, as described in #2423 .

It takes the `namedincludes` attributes from the `site` table, split it on `,`, and adds an 
```
include "path/to/file";
```
line in `/etc/named.conf` for each entry. 
No include line is added if the attribute doesn't exist in the `site` table.

For instance, with:
```
# tabdump site | grep namedincludes
"namedincludes","/etc/named/local.conf,/etc/named/more.conf",,
```
`makedns` generates the following `named.conf` file:
```
options {
[...]
};

include "/etc/named/local.conf";
include "/etc/named/more.conf";

zone "[...]
```

No check is done to verify that the files in `site.namedincludes` actually exists, and it assumes the same includes on all the DNS servers (master, forwarding server, slaves).

There's clearly much improvement to do:
- [ ] checking include files existence
- [ ] allowing different includes on master and slaves
- [ ] generalizing the mechanism to allow inclusion of additional files in other generated files, such as `/etc/hosts` or `/etc/dhcpd.conf`
- [ ] documentation

But that's a first step. And it works for me. :)

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ddns.pm
+++ b/xCAT-server/lib/xcat/plugins/ddns.pm
@@ -1280,6 +1280,19 @@ sub update_namedconf {
         push @newnamed, "};\n\n";
     }
 
+    # include external configuration file(s) if present in site.namedincludes
+    my @entries = xCAT::TableUtils->get_site_attribute("namedincludes");
+    my $site_entry = $entries[0];
+    if (defined($site_entry)) {
+        my @includes = split /[ ,]/, $site_entry;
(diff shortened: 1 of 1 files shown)


bybai:
Hi @kcgthb, Thanks your contribution. I will merge this pull request first. If you have new commit, please create new pull request.

PR #4503, opened 2017-12-13
Do not restart conserver if goconserver was started

chenglch:
If goconserver was enabled, do not start conserver when restart
xcatd on service node.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/AAsn.pm
+++ b/xCAT-server/lib/xcat/plugins/AAsn.pm
@@ -488,8 +488,14 @@ sub setup_CONS
 {
     my ($nodename) = @_;
     my $rc = 0;
-
-    my $cmd;
+    if (-x "/usr/bin/goconserver") {
+        my $cmd = "ps axf | grep -v grep | grep \/usr\/bin\/goconserver";
+        xCAT::Utils->runcmd($cmd, 0);
+        if ($::RUNCMD_RC == 0) {
+            xCAT::MsgUtils->message("S", "INFO: goconserver was started, do nothing for conserver.");
+            return 0;
+        }
+    }
     my $cmdref;
     $cmdref->{command}->[0] = "makeconservercf";
     $cmdref->{arg}->[0]     = "-l";


xCAT-server/lib/xcat/plugins/AAsn.pm
```diff
@@ -488,8 +488,14 @@ sub setup_CONS
 {
     my ($nodename) = @_;
     my $rc = 0;
-
-    my $cmd;
+    if (-x "/usr/bin/goconserver") {
+        my $cmd = "ps axf | grep -v grep | grep \/usr\/bin\/goconserver";
+        xCAT::Utils->runcmd($cmd, 0);
+        if ($::RUNCMD_RC == 0) {
+            # goconserver is started, do nothing.
```
whowutwut: Should we inform the user that nothing is happening with conserver because goconserver is running?

chenglch: done

xCAT-server/lib/xcat/plugins/AAsn.pm
```diff
@@ -488,8 +488,14 @@ sub setup_CONS
 {
     my ($nodename) = @_;
     my $rc = 0;
-
-    my $cmd;
+    if (-x "/usr/bin/goconserver") {
+        my $cmd = "ps axf | grep -v grep | grep \/usr\/bin\/goconserver";
```
robin2008: Could we use `chkconfig` or `systemctl` to check if the service is enabled and do such thing?

chenglch: Check the process status is simplest way as we do not need to check the service type of the system. Currently goconserver do not support sysv directly,  this kind of check is safe if the user just want to try the binary manually.

robin2008: Maybe it is in our next sprint for the enhancement as `xcatd` now will try to bring its dependent service up. (For example, if goconserver is down occasionally, restart xcatd will bring it up again)

Issue #4513, opened 2017-12-13
[OpenBMC] reventlog option parsing regressed between 12/12 and 12/13 build


whowutwut:
@hu-weihua @xuweibj We need to add some UT tests for reventlog. 
 **Edit**: Or let's see if our automation results catch this issue... 

On the 12/12 build... 
```
[root@briggs01 ~]# reventlog mid05tor12cn13 10 -s
mid05tor12cn13: 12/13/2017 04:46:09 [42] xyz.openbmc_project.Software.Version.Error.ManifestFileFailure
mid05tor12cn13: 12/13/2017 04:39:45 [41] xyz.openbmc_project.Software.Version.Error.ManifestFileFailure
mid05tor12cn13: 12/13/2017 03:32:28 [40] xyz.openbmc_project.Software.Version.Error.ManifestFileFailure
mid05tor12cn13: 12/13/2017 02:59:20 [39] xyz.openbmc_project.Software.Version.Error.ManifestFileFailure
mid05tor12cn13: 12/06/2017 20:36:11 [38] xyz.openbmc_project.Software.Version.Error.ManifestFileFailure
mid05tor12cn13: 12/06/2017 12:05:06 [37] org.open_power.Host.Boot.Error.Checkstop
mid05tor12cn13: 12/06/2017 11:58:08 [36] xyz.openbmc_project.Software.Version.Error.ManifestFileFailure
mid05tor12cn13: 12/06/2017 11:46:57 [35] org.open_power.Host.Boot.Error.Checkstop
mid05tor12cn13: 12/06/2017 11:39:43 [34] xyz.openbmc_project.Software.Version.Error.ManifestFileFailure
mid05tor12cn13: 12/05/2017 20:55:15 [33] xyz.openbmc_project.Common.Error.InternalFailure
[root@briggs01 ~]#
```

and after updating to 12/13 build
```
[root@briggs01 ~]# reventlog mid05tor12cn13 10 -s
mid05tor12cn13: Error: Only one option is supported at the same time for reventlog
```

So we regressed..  we need testcases to detect this kind of stuff and run it in UT before we submit pull requests.. 

@xuweibj Can you target a fix for 2.13.9?

xuweibj:
@whowutwut , we have discussed about this, decided to remove ``-s`` option. In the describe of pull request #4366, said that "Delete -s option, will sort for all. If run reventlog {number-of-entries}, will print out the last {number-of-entries} logs."

[...]

PR #4524, opened 2017-12-14
Modify or add openbmc test cases or bundle

hu-weihua:
This pull request modified some openbmc rflash test cases depending on the latest xcat code change, add one reventlog test case (for issue #4513) and update related openbmc bundle.


hu-weihua:
For task #4547

PR #4527, opened 2017-12-14
[RFC] Do not comment out rsyslog /var/log/messages on MN

sebth:
Since xCAT 2.11, the rule `*.debug /var/log/messages` is no longer added to rsyslog.conf on the MN, but all pre-existing rules ending with "/var/log/messages" are still commented out. Is this a mistake? I cannot imagine that xCAT is supposed to disable all logging to /var/log/messages (though, it enables the daemon and ftp facilities in the config file `/etc/rsyslog.d/xcat-debug.conf`, still leading to a "/var/log/messages" that is missing a lot of messages).

This PR has a commit that solves the problem, though I am not sure how to properly solve duplicate messages coming from `/etc/rsyslog.d/xcat-debug.conf`, hence the RFC tag. Still, duplicate messages are better than missing messages!

### DIFF
--- a/xCAT/etc/rsyslog.d/xcat-debug.conf
+++ b/xCAT/etc/rsyslog.d/xcat-debug.conf
@@ -1,3 +1,3 @@
-# By default, dhcpd use the ``daemon'' facility. Thus dhcp logs will go here.
-# ftp.* is for tftp log.
-daemon.debug;ftp.*		/var/log/messages
+# Make sure dhcpd requests and responses (daemon.=debug) and
+# tftp messages (ftp.*) are logged.
+daemon.=debug;ftp.*		/var/log/messages
--- a/xCAT/postscripts/syslog
+++ b/xCAT/postscripts/syslog
@@ -117,11 +117,8 @@ config_Rsyslog_C3()
 
   #enable to send the logging to master
   if [ $isLocal -eq 1 ]; then
-      #enable the MN to log all the logs from CN and MN itself
-      sed -i 's/\(^[^\#].*\/var\/log\/messages\)/\#\1/g' $2
-      echo "#*.debug  /var/log/messages" >> $2
-      touch  /var/log/messages  
       if ( pmatch $OSVER "ubuntu*" ) || ( is_lsb_ubuntu ) || ( pmatch $OSVER "debian*" ); then
+        touch /var/log/messages
         chown syslog:adm /var/log/messages
         sed -i 's/\$PrivDropToGroup syslog/\$PrivDropToGroup adm/' $2
         chown syslog:adm /var/log/xcat
@@ -200,7 +197,7 @@ config_rsyslog_V8()
 
   if [ $goLocal -eq 1 ]; then
     #logging is local, do not forward
-    #making sure all the messages goes to /var/log/messages
+    #making sure /var/log/messages is setup for xcat-debug
     touch  /var/log/messages
     if ( pmatch $OSVER "ubuntu*" ) || ( is_lsb_ubuntu ) || ( pmatch $OSVER "debian*" ); then
       chown syslog:adm /var/log/messages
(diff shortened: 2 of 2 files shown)


whowutwut:
@sebth Thanks for your interest in the xCAT project and your pull request! We appreciate it.  Before we are able to accept contributions, we do require a CLA to be signed.  You can find the document and instructions here: 
http://xcat-docs.readthedocs.io/en/stable/developers/license/contributors.html

CLA or CCLA (corporate).. If you are covered under a corporate CLA, please let us know the company and we will add your ID into the list. Thanks!

immarvin:
hi @sebth ,  xCAT ships some log filter rules, such as "xcat-compute.conf" and "xcat-cluster.conf" and  "xcat-debug.conf",  the logs which pass the filters will go to separate log files to facilitate debug

Just as you mentioned, commenting out the log entry for `/var/log/messages` is to avoid duplicate logs.  The duplicate logs introduce some trouble when these logs are processed by some log analysis tools, such as ELK and `xcatprobe` shipped by xCAT.

of course, this was just our idea on this, it is open to discuss and determine whether this is reasonable from the user's perspective

sebth:
@whowutwut Thanks, I will look into the CLA if the discussion on this PR leads to a commit that would be accepted.

@immarvin I think that if xCAT needs to take control over `/var/log/messages`, it should set a reasonable rule. As of now, if you install xCAT on a RHEL 7 system, `/var/log/messages` will suddenly only show the daemon and ftp facilities. For instance, kernel messages will no longer be logged anywhere. As a user, I think this is unexpected and a bit dangerous. The previous default of sending "*.debug" to `/var/log/messages` was more reasonable.

Though I would prefer if xCAT did not touch `/etc/rsyslog.conf` at all, and redirected the things it needed to analyze to something like `/var/log/xcat/debug.log` by a rule in `/etc/rsyslog.d/xcat-debug.conf`, I think the most important thing here is that xCAT does not set a too minimal rule for `/var/log/messages` like it does right now.

whowutwut:
@immarvin @neo954  I do agree with @sebth ...  Why do we need to touch the main rsyslog config file `/etc/rsyslog.conf` file if we have the option to provide our own user defined ones in `/etc/rsyslog.d`?  

`/etc/rsyslog.conf` has this config setting:
```
$IncludeConfig /etc/rsyslog.d/*.conf
```

In the past we did have some issues during updates and messing with other config in that file, so we now wrap with our own tags to search and replace only between those...
```
# xCAT settings START
# Entries between the START and END lines will be replaced on the next xCAT install or update
#*.debug   /var/log/messages
# xCAT settings END
```

neo954:
@whowutwut, we don't really need to touch `/etc/rsyslog.conf`. Those codes were there just because they are historical remains. I think just comment out line 121 and line 213 in file `xCAT/postscripts/syslog` will satisfy @sebth's need.

And for this change, we still need some extra lines to handle the situation of xCAT migration. For example upgrade from xCAT 2.10 to 2.11, or 2.10 to 2.13. Uncomment the line commented out by postscript `syslog` prior to xCAT 2.11. That line should look like the following line in RHEL7. But it shows differently in RHEL6, RHEL5, RHEL4, SLES 12, and Ubuntu.
```
*.info;mail.none;authpriv.none;cron.none                /var/log/messages
```

immarvin:
i am ok with the change

robin2008:
Known issue:
 This PR works fine with new installed xCAT on fresh MN.  But if you upgrade from previous xCAT version, you must take some manual actions. Or /var/log/messages won't have the logs as expected.

It is caused that the `/etc/rsyslog.conf` is already modified by xCAT,  so all default rules to /var/log/messages are disabled.  And after the fix,  the rule in `/etc/rsyslog.d/xcat-debug.conf` is not working too.  You have to recover the default rules in `/etc/rsyslog.conf`.

whowutwut:
@immarvin Seems like this was merged without a CLA sent to xCAT Team...  @sebth if possible could you fill out a CLA for a our records. 

> @whowutwut Thanks, I will look into the CLA if the discussion on this PR leads to a commit that would be accepted. 

http://xcat-docs.readthedocs.io/en/stable/developers/license/contributors.html

sebth:
@whowutwut I have emailed a signed CLA to 'xcat-legal' just now.

whowutwut:
@sebth Got it, Thank you! I have sent an invite to be part of the xCAT contributor team, you can accept or not, no requirement to do so but just sent it incase.  Welcome!

sebth:
@whowutwut Thank you!

PR #4529, opened 2017-12-15
Support custom udev rules in the initrd

### LINKED ISSUE #4489: Need a way to set ethtool parameters early in boot
Due to #4458 we have a need to set some `ethtool` parameters on our boot nic.  We've seen it fail before we've even downloaded the image, so these settings need to happen in the initrd.  Unfortunately, the `tg3` module doesn't take any relevant parameters, so it looks like something will need to call `ethtool` very soon after the device is discovered.

Idea 1: Create udev rules so that when any device using the `tg3` driver comes online it runs `ethtool`.  Problem: you can't set `rx-jumbo` if the interface MTU is 1500.  Workaround: set the MTU to 9000, run `ethtool`, set the MTU back to 1500.

Idea 2: Use `dhclient`'s hook infrastructure to run scripts whenever an interface is configured.  Problem: dracut's version of `dhclient-script` doesn't call any hooks.  Workaround: patch `dhclient-script.sh` inside dracut.

Idea 3: Put something in the xcat dracut module.  Downside: there's a non-zero time between when the interface comes up and when the xcat module runs.  Might be trivial, but the interface *could* fail in that window.

Thoughts?

mattaezell:
This allows any udev rules in `/etc/udev/rules.d` that have the case-insensitive word `xcat` in them to be pulled into the initrd.  Admins should create their own udev rules from an RPM or using a postinstall script, or modify an existing rule by appending a comment
```
echo "# Please install me xCAT" >> $IMG_ROOTIMGDIR/etc/udev/rules.d/10-preexisting-rule.rules
```
This also pulls in ethtool, which is currently required on WSP nodes.
Closes #4489

### DIFF
--- a/xCAT-server/share/xcat/netboot/rh/dracut_033/install.netboot
+++ b/xCAT-server/share/xcat/netboot/rh/dracut_033/install.netboot
@@ -4,6 +4,10 @@ dracut_install wget tar cpio gzip modprobe touch echo cut wc xz
 dracut_install grep ifconfig hostname awk egrep grep dirname expr
 dracut_install mount.nfs
 dracut_install parted mke2fs bc mkswap swapon chmod
+dracut_install ethtool
 inst "$moddir/xcat-updateflag" "/tmp/updateflag"
 inst "$moddir/xcatroot" "/sbin/xcatroot"
 inst_hook cmdline 10 "$moddir/xcat-cmdline.sh" 
+for file in /etc/udev/rules.d/*;do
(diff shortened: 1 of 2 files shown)


mattaezell:
Please review @immarvin and/or @robin2008

immarvin:
hi @mattaezell , `ethtool` should be added to the rhels6&7 pkglist under `netboot/rh/`, created  PR https://github.com/xcat2/xcat-core/pull/4532/files for this

immarvin:
agree to merge

Issue #4538, opened 2017-12-15
confignetwork fails when BMC interfaces are defined in the 'nics' table


kcgthb:
The `confignetwork` postscript returns an error when BMC interfaces are defined in the 'nics` table.

For instance, with  the following node definition:
```
Object name: sh-06-34
    nicaliases.bmc=sh-06-34
    nicaliases.ib0=sh-06-34
    nicextraparams.ib0=MTU=65520 CONNECTED_MODE=yes
    nicips.bmc=10.11.06.34
    nicips.ib0=10.8.06.34
    nicnetworks.bmc=net_ipmi
    nicnetworks.ib0=net_ib_56
    nictypes.bmc=bmc
    nictypes.ib0=infiniband
```

```
# updatenode sh-06-34 -P confignetwork
sh-06-34: xcatdsklspost: downloaded postscripts successfully
sh-06-34: Fri Dec 15 09:23:49 PST 2017 Running postscript: confignetwork
sh-06-34: [I]: All valid nics and device list:
sh-06-34: [I]: bmc
sh-06-34: [I]: ib0
sh-06-34: [I]: NetworkManager is inactive.
sh-06-34: ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
sh-06-34: configure nic and its device : bmc
sh-06-34: [E]:Error: Check the NIC data in the 'nics' table.
sh-06-34: ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
sh-06-34: configure nic and its device : ib0
sh-06-34: [I]: Call configib for IB nics: ib0, ports:
sh-06-34: [I]: NIC_IBNICS=ib0 NIC_IBAPORTS= configib
sh-06-34: em1!MTU=9000
sh-06-34: ib0!MTU=65520 CONNECTED_MODE=yes
sh-06-34: array_nic_params 0=MTU=65520 CONNECTED_MODE=yes
sh-06-34:   0: name=MTU value=65520
sh-06-34:   1: name=CONNECTED_MODE value=yes
sh-06-34: postscript: confignetwork exited with code 1
sh-06-34: Running of postscripts has completed.
```

The script produces an error for the `bmc` interface:
```
[E]:Error: Check the NIC data in the 'nics' table.
```
[...]

PR #4539, opened 2017-12-15
ignore bmc interfaces defined in the nics table

kcgthb:
workaround for #4538 

If BMC interfaces are defined in the `nics` table, just ignore them, so `confignetwork` doesn't exit with an error.

BMC interfaces may need to be defined in `nics` to generate proper DNS records, but `confignetwork` can't actually configure them. Their configuration is handled in `bmcsetup`.


bybai:
I am ok to merge it. Thanks.

PR #4551, opened 2017-12-18
Use all name servers from site.nameservers in mkresolvconf

kcgthb:
`mkresolvconf` configures a single name server, even if several are defined in `site.nameservers`.

This patch uses all the name servers defined in `site` when generating `/etc/resolv.conf` on nodes.


immarvin:
hi @bybai ,would you please take a look at this?

bybai:
I am ok to merge it. Thanks @kcgthb

PR #4552, opened 2017-12-18
Check if the include line is already there before adding it to the file

kcgthb:
Fix for #4497: includes were added each time `makedns` was called. 

This patch checks that the `include` line doesn't already exist in the file before adding it.


bybai:
I am ok to merge it. Thanks

Issue #4474, opened 2017-12-07
[OpenBMC] Allow the ability to clear GARD via REST on the BMCs


whowutwut:
This is maybe more for development ...  when we hit a GARD stop, it is nice to be able to clear the GARD file and reboot the node...  

The file to clear is: 
`/var/lib/phosphor-software-manager/pnor/prsv/GUARD` 

There's a REST interface, but OpenBMC team will change the team will change the path to xyz, so until that happens trying to track down the org/openbmc interface if we want to support this quicker... 

## Proposed External: 
```
rspconfig <noderange> reset_gard
```

What I don't understand is the file is spelled `GUARD` but the REST API endpoing is `gard`... oh well.

whowutwut:
I don't think this is ready yet to use yet... , they have a typo in the org/openpower endpoint that needs to be resolved..    If we get a clear endpoint defined, we could do the development and just error on endpoint incorrect...  will track this and update if I find out more info.

whowutwut:
This function is in the `r21` build of OP910 driver, https://w3.austin.ibm.com/afs/austin/projects/gfw/images/openpower9/910.1746.20171215a/ship/ so this work can begin...

whowutwut:
I verified on the r21 build today that it works... 

The following script just does a `ls -ltr` on  `/var/lib/phosphor-software-manager/pnor/prsv/GUARD` on the BMC.... 

```
root@briggs01 Control]# ./list_gard.sh
==> BMCIP=172.11.139.18
-rw-rw-r--    1 root     root         20480 Dec 13 07:46 /var/lib/phosphor-software-manager/pnor/prsv/GUARD

Then using the UT scripts: 

```
[root@briggs01 Control]# ./login.sh
==> BMCIP=172.11.139.18
+ date
Mon Dec 18 21:43:14 EST 2017
+ curl -c cjar -k -X POST -H 'Content-Type: application/json' -d '{"data": [ "root", "0penBmc" ] }' https://172.11.139.18/login
{
  "data": "User 'root' logged in",
  "message": "200 OK",
  "status": "ok"
}
[root@briggs01 Control]# ./clear_gard.sh
+ date
Mon Dec 18 21:43:33 EST 2017
+ curl -c cjar -b cjar -k -H 'Content-Type: application/json' -X POST -d '{"data":[]}' https://172.11.139.18/org/open_power/control/gard/action/Reset
{
  "data": null,
  "message": "200 OK",
  "status": "ok"
}
```

[...]

PR #4573, opened 2017-12-20
rspconfig gard -c|--clear for OpenBMC

xuweibj:
#4474 

```
# rspconfig mid05tor12cn18 gard -c
mid05tor12cn18: GARD cleared
```

Run when xcatdebugmode is 1:
```
# rspconfig mid05tor12cn18 gard -c
Wed Dec 20 03:23:39 2017 OpenBMC: [openbmc_debug]
Wed Dec 20 03:23:39 2017 mid05tor12cn18: [openbmc_debug] curl -k -c cjar -H "Content-Type: application/json" -d '{ "data": ["root", "xxxxxx"] }' https://172.11.139.18/login
Wed Dec 20 03:23:40 2017 mid05tor12cn18: [openbmc_debug] login_response 200 OK
Wed Dec 20 03:23:40 2017 mid05tor12cn18: [openbmc_debug] curl -k -b cjar -X POST -H "Content-Type: application/json" -d '{"data":[]}' https://172.11.139.18/org/open_power/control/gard/action/Reset
Wed Dec 20 03:23:40 2017 mid05tor12cn18: [openbmc_debug] rspconfig_clear_gard_response 200 OK
mid05tor12cn18: GARD cleared
```

UT:
1. create a non-empty GUARD file on mid05tor12cn18-bmc-test "/var/lib/phosphor-software-manager/pnor/prsv/GUARD"
2. run ``rspconfig mid05tor12cn18 gard -c``
3. ``ls: /var/lib/phosphor-software-manager/pnor/prsv/GUARD: No such file or directory``

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -449,12 +449,12 @@ my %status_info = (
     RSPCONFIG_SSHCFG_RESPONSE => {
         process        => \&rspconfig_sshcfg_response,
     },
-    RSPCONFIG_RESET_GARD_REQUEST => {
+    RSPCONFIG_CLEAR_GARD_REQUEST => {
         method         => "POST",
         init_url       => "/org/open_power/control/gard/action/Reset",
         data           => "[]", 
     }, 
-    RSPCONFIG_RESET_GARD_RESPONSE => {
+    RSPCONFIG_CLEAR_GARD_RESPONSE => {
         process        => \&rspconfig_response,
     },
     RSPCONFIG_DUMP_LIST_REQUEST => {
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1096,6 +1104,9 @@ sub parse_args {
             } elsif ($subcommand =~ /^sshcfg$/) {
                 return ([ 1, "Configure sshcfg must be issued without other options." ]) if ($num_subcommand > 1);
                 $setorget = ""; # SSH Keys are copied using a RShellAPI, not REST API
+            } elsif ($subcommand =~ /^reset_gard$/) {
```
whowutwut: I am thinking about this some more,  i came up with `reset_gard` because of the Reset endpoint but it does not make sense because when firmware team talks to us about it, it's always "Need to clear the GARD and reboot the server"...  I think Reset is just a generic endpoint.  

This gets me thinking about another option that would make more sense.... Thinking "out loud" 

* rspconfig <range> gard --clear 
* rspconfig <range> clear_gard
* rspconfig <range> cleargard
* rspconfig <range> clear --gard <-- would we have other options

Trying to think about other options we have clear for and whether we could synchronize them up... 
* rspconfig <range> dump -c all 
* reventlog <range> clear

Anything stand out as better?

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1096,6 +1104,9 @@ sub parse_args {
...
+                return  ([ 1, "Reset GARD must be issued without other options."]) if ($num_subcommand > 1);
```
whowutwut: Suggest: `Clear GARD cannot be issued with any other options.`

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -3226,6 +3242,12 @@ sub rspconfig_response {
         return;
     }
 
+    if ($node_info{$node}{cur_status} eq "RSPCONFIG_RESET_GARD_RESPONSE") {
+        if ($response_info->{'message'} eq $::RESPONSE_OK) {
+            xCAT::SvrUtils::sendmsg("GARD reset", $callback, $node);
```
whowutwut: If we change the name would suggest: `GARD file removed` or `GARD removed` or `GARD cleared` instead of reset

PR #4574, opened 2017-12-20
new perl plugin draft for simple review [DO NOT MERGE]

xuweibj:
Just for cooperative development.


whowutwut:
@xuweibj Is there a way to toggle this? I think we would stage this in.... so supporting both at the same time... `mgt=openbmc` would remain the perl version....  maybe `mgt=openbmc.v2` would be this new one? 

I'd like to have both running because i think this will be somewhat long term as we ensure the functions are identical before we can deprecate one over another.

xuweibj:
Has be contained in #4585.

PR #4577, opened 2017-12-20
Modify some openbmc test cases depending on the latest output of command

hu-weihua:
Due to some output of ``rflash`` and ``rspconfig`` have been changed, so update the test cases depending these change.


xuweibj:
Looks good to me. Thanks @hu-weihua .

Issue #4561, opened 2017-12-19
[OpenBMC] Ceate communication between python plugin and perl plugin


zet809:
* [ ] the pipeline/socket pair used to contact python plugin and perl plugin.

PR #4585, opened 2017-12-26
Add agent example with gevent to work with xcat

chenglch:
This patch add a framework to communicate betwwen python and perl
modules. Orignal xcat plugin run a python agent, then send the request
info and the data from db to the python agent with unix domain socket.

Below is the UT example, some overhead to fork the process, but not too
much.
```
# time rpower simulator_test0,simulator_test1 status
simulator_test0: rpower called info={u'bmcip': u'10.5.102.73', u'bmc': u'10.5.102.73', u'username': u'root', u'password': u'cluster'} args=[u'status']
simulator_test1: rpower called info={u'bmcip': u'10.5.102.73', u'bmc': u'10.5.102.73', u'username': u'root', u'password': u'cluster'} args=[u'status']
real   	0m0.310s
user   	0m0.071s
sys    	0m0.016s
```

implement #4561 
```
# /opt/xcat/lib/perl/xCAT_plugin/agent/agent.py --help
usage: xcatagent [--standalone] [--sock SOCK]

Optional arguments:
  --standalone  Start xcat agent as a standalone service, mostly work for test
                purpose.
  --sock SOCK   The unix domain sock file to communicate with the client

# /opt/xcat/lib/perl/xCAT_plugin/agent/client.py --help
usage: agentclient [--sock SOCK]

Optional arguments:
  --sock SOCK  The unix domain sock file to communicate with the server
```

### DIFF
--- a/xCAT-server/debian/dirs
+++ b/xCAT-server/debian/dirs
@@ -23,6 +23,9 @@ opt/xcat/share/xcat/ib/netboot/sles
 opt/xcat/share/xcat/ib/netboot/rh
 opt/xcat/lib
 opt/xcat/lib/perl
+opt/xcat/lib/python
+opt/xcat/lib/python/agent
+opt/xcat/lib/python/agent/xcatagent
 opt/xcat/lib/perl/Confluent
 opt/xcat/lib/perl/xCAT
 opt/xcat/lib/perl/xCAT_plugin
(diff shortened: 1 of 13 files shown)


whowutwut:
@chenglch @xuweibj Late review to this,(holiday...)  but I don't know if I like to have this code integrated into the same packaging as xCAT2.... there are talks about the "next gen" of xcat being Python based, why would we integrate like this? Wouldn't we want the pieces to be separate so we can more easily make a break.. or build upon the python version of code separate from xcat2?

Issue #4554, opened 2017-12-19
[OpenBMC] rpower support with Python and coroutine


xuweibj:
* [x] Read attributes "bmc", "bmcusername", "bmcpassword" from file
* [x] Send request and receive response to/from OpenBMC
* [x] Run ``rpower`` command success on Python code

xuweibj:
The task was finished in pull request #4587

PR #4587, opened 2017-12-27
rpower support utilizing the python framework for OpenBMC

xuweibj:
#4554 

UT:
```
# rpower f6u17 state
f6u17: on
```
UT with "xcatdebugmode" is 1:
```
# rpower f6u17 bmcreboot
Wed Dec 27 02:08:47 2017 f6u17: [openbmc_debug] login curl -k -c cjar -b cjar -X POST -H "Content-Type: application/json" https://10.6.17.100/login -d '{"data": ["root", "xxxxxx"]}'
Wed Dec 27 02:08:47 2017 f6u17: [openbmc_debug] login 200 OK
Wed Dec 27 02:08:47 2017 f6u17: [openbmc_debug] rpower_bmcreboot curl -k -c cjar -b cjar -X PUT -H "Content-Type: application/json" https://10.6.17.100/xyz/openbmc_project/state/bmc0/attr/RequestedBMCTransition -d '{"data": "xyz.openbmc_project.State.BMC.Transition.Reboot"}'
Wed Dec 27 02:08:47 2017 f6u17: [openbmc_debug] rpower_bmcreboot 200 OK
f6u17: BMC reboot
```
UT when password is wrong:
```
# rpower openbmctest state
openbmctest: Error: [401] Invalid username or password
```

### DIFF
--- a/xCAT-server/lib/python/agent/client.py
+++ b/xCAT-server/lib/python/agent/client.py
@@ -41,8 +41,8 @@ class ClientShell(object):
 
         s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         nodes = ['node1', 'node2']
-        nodeinfo = {'node1':{'bmc':'9.3.185.161', 'bmcip':'9.3.185.161', 'username':'root', 'password': '0penBmc1'},
-                    'node2':{'bmc':'10.6.17.100', 'bmcip': '10.6.17.100', 'username':'root', 'password': '0penBmc'}}
+        nodeinfo = {'node1':{'bmc':'10.0.0.1', 'bmcip':'10.0.0.1', 'username':'root', 'password': 'xxxxxx'},
+                    'node2':{'bmc':'10.0.0.2', 'bmcip':'10.0.0.2', 'username':'root', 'password': 'xxxxxx'}}
 
         s.connect(options.sock)
         req = {'module': 'openbmc',
@@ -51,7 +51,7 @@ class ClientShell(object):
                'cwd': os.getcwd(),
                'nodes': nodes,
                'nodeinfo': nodeinfo,
-               'debugmode': 1}
+               'envs': {'debugmode': 1}}
 
         buf = json.dumps(req)
         s.send(utils.int2bytes(len(buf)))
--- a/xCAT-server/lib/python/agent/client.py
+++ b/xCAT-server/lib/python/agent/client.py
@@ -40,18 +40,18 @@ class ClientShell(object):
             return 0
 
         s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-        nodes = ['node%s' % i for i in range(100)]
-        nodeinfo = {node: {'username': 'admin', 'password': 'cluster'} for node
-                    in
-                    nodes}
+        nodes = ['node1', 'node2']
+        nodeinfo = {'node1':{'bmc':'9.3.185.161', 'bmcip':'9.3.185.161', 'username':'root', 'password': '0penBmc1'},
+                    'node2':{'bmc':'10.6.17.100', 'bmcip': '10.6.17.100', 'username':'root', 'password': '0penBmc'}}
 
         s.connect(options.sock)
         req = {'module': 'openbmc',
                'command': 'rpower',
-               'args': ['-d', 'firmware'],
+               'args': ['state'],
                'cwd': os.getcwd(),
                'nodes': nodes,
-               'nodeinfo': nodeinfo}
+               'nodeinfo': nodeinfo,
+               'debugmode': 1}
 
         buf = json.dumps(req)
         s.send(utils.int2bytes(len(buf)))
--- a/xCAT-server/lib/python/agent/xcatagent/openbmc.py
+++ b/xCAT-server/lib/python/agent/xcatagent/openbmc.py
@@ -223,11 +223,11 @@ class OpenBMC(base.BaseDriver):
 
 
 class OpenBMCManager(base.BaseManager):
-    def __init__(self, messager, cwd, nodes, debugmode):
+    def __init__(self, messager, cwd, nodes, envs):
         super(OpenBMCManager, self).__init__(messager, cwd)
         self.nodes = nodes
         global DEBUGMODE
-        DEBUGMODE = debugmode
+        DEBUGMODE = envs['debugmode']
 
     def rpower(self, nodeinfo, args):
         super(OpenBMCManager, self).process_nodes_worker('openbmc', 'OpenBMC', self.nodes, nodeinfo, 'rpower', args)
--- a/xCAT-server/lib/python/agent/xcatagent/openbmc.py
+++ b/xCAT-server/lib/python/agent/xcatagent/openbmc.py
@@ -1,26 +1,233 @@
 from xcatagent import base
-import gevent
+import time
+import sys
+
+import xcat_exception
+import rest
+
+HTTP_PROTOCOL = "https://"
+PROJECT_URL = "/xyz/openbmc_project"
+
+RESULT_OK = 'ok'
+
+DEBUGMODE = False
+
+POWER_SET_OPTIONS = ('on', 'off', 'bmcreboot', 'softoff')
+POWER_GET_OPTIONS = ('bmcstate', 'state', 'stat', 'status')
+
+RPOWER_URLS = {
+    "on"        : {
+        "url"   : PROJECT_URL + "/state/host0/attr/RequestedHostTransition",
+        "field" : "xyz.openbmc_project.State.Host.Transition.On",
+    },
+    "off"       : {
+        "url"   : PROJECT_URL + "/state/chassis0/attr/RequestedPowerTransition",
+        "field" : "xyz.openbmc_project.State.Chassis.Transition.Off",
+    },
+    "softoff"   : {
+        "url"   : PROJECT_URL + "/state/host0/attr/RequestedHostTransition",
+        "field" : "xyz.openbmc_project.State.Host.Transition.Off",
+    },
+    "bmcreboot" : {
+        "url"   : PROJECT_URL + "/state/bmc0/attr/RequestedBMCTransition",
+        "field" : "xyz.openbmc_project.State.BMC.Transition.Reboot",
+    },
+    "state"     : {
+        "url"   : PROJECT_URL + "/state/enumerate",
+    },
+}
+
+RPOWER_STATE = {
+    "on"        : "on",
+    "off"       : "off",
+    "Off"       : "off",
+    "softoff"   : "softoff",
+    "boot"      : "reset",
+    "reset"     : "reset",
+    "bmcreboot" : "BMC reboot",
+    "Ready"     : "BMC Ready",
+    "NotReady"  : "BMC NotReady",
+    "chassison" : "on (Chassis)",
+    "Running"   : "on",
+    "Quiesced"  : "quiesced",
+}
+
+class OpenBMC(base.BaseDriver):
+
+    headers = {'Content-Type': 'application/json'}
+
+    def __init__(self, messager, name, node_info):
+        super(OpenBMC, self).__init__(messager)        
+        self.node = name
+        for key, value in node_info.items() :
+            setattr(self, key, value)
+        global DEBUGMODE
+        self.client = rest.RestSession(messager, DEBUGMODE)
+
+
+    def _login(self) :
+        """ Login
+        :raise: error message if failed
+        """
+        url = HTTP_PROTOCOL + self.bmcip + '/login'
+        data = { "data": [ self.username, self.password ] }
+        self.client.request('POST', url, OpenBMC.headers, data, self.node, 'login')
+        return RESULT_OK
+
+
+    def _set_power_onoff(self, subcommand) :
+        """ Set power on/off/softoff/bmcreboot
+        :param subcommand: subcommand for rpower
+        :returns: on/off/softoff/BMC reboot if success
+        :raise: error message if failed
+        """
+        url = HTTP_PROTOCOL + self.bmcip + RPOWER_URLS[subcommand]['url']
+        data = { "data": RPOWER_URLS[subcommand]['field'] }
+        try :
+            response = self.client.request('PUT', url, OpenBMC.headers, data, self.node, 'rpower_' + subcommand)
+        except (xcat_exception.SelfServerException,
+                xcat_exception.SelfClientException) as e :
+            if subcommand != 'bmcreboot':
+                result = e.message
+            return result
+
+        return RESULT_OK
+
+
+    def _get_power_state(self, subcommand) :
+        """ Get power current state
+        :param subcommand: state/stat/status/bmcstate
+        :returns: current state if success
+        :raise: error message if failed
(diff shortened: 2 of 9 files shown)


xCAT-server/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -1,26 +1,223 @@
...
+        "init_url" : openbmc_project_url + "/state/host0/attr/RequestedHostTransition",
+        "data"     : "xyz.openbmc_project.State.Host.Transition.Off",
+    },
+    "rpower_state" : {
+        "init_url" : openbmc_project_url + "/state/enumerate",
+    }
+}
+
+subcommand_tup_dict = {
+    "rpower_set_onoff" : ['on', 'off', 'bmcreboot', 'softoff'],
```
chenglch: 1. upper case
2. tuple is better than list here.

No need to use subcommand_tup_dict, you are writing openbmc module, a class to provide service, not command handler,  I prefer the following: 
``` 
POWER_STATE = ('on', 'off', 'bmcreboot', 'softoff')
```

xCAT-server/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -1,26 +1,223 @@
...
+import gevent.monkey
+gevent.monkey.patch_all()
+import time
+
+import openbmc_rest
+
+http_protocol="https://";
+openbmc_project_url = "/xyz/openbmc_project";
+
+command_dict = {
```
chenglch: Only method, no command here.
I think it's better to put the url and the data field in the method.
```
def _power_on():
    url = PROJECT_URL + "/state/host0/attr/RequestedHostTransition"
    field = "xyz.openbmc_project.State.Host.Transition.On"
```
You could also rename this dict like RPOWER_URLS = { on: {...}, off:...}, `init_url` is not a proper name here.

xCAT-server/lib/python/agent/xcatagent/openbmc_rest.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python
```
chenglch: I think the file name should be rest.py,  make it as a common module. Add get, post, put, delete, patch method may be more friendly.

xCAT-server/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -1,26 +1,223 @@
...
         self.nodes = nodes
 
-    def rpower(self, nodeinfo, args):
-        driver = OpenBMCDriver(self.messager)
-        lst = [gevent.spawn(driver.rpower, node, nodeinfo[node],
-                            args) for node in self.nodes]
-        gevent.joinall(lst)
-
+    def rpower(self, nodeinfo, args, debugmode):
+        glist = [ gevent.spawn(self.handler, node, nodeinfo[node], 'rpower', args, debugmode) for node in self.nodes ]
```
chenglch: Actually your OpenBMC class is worked like the original OpenBMC driver.
You should create a list of driver(OpenBMC) objects, then create gevent objects to call the driver method(rpower).  Like below:
```
objs = [OpenBMC(node, nodeinfo[node]) for node in self.nodes]
glist = [ gevent.spawn(obj.rpower, args, debugmode) for obj in objs]
```

xCAT-server/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -1,26 +1,223 @@
...
+        gevent.joinall(glist)
 
-class OpenBMCDriver(base.BaseDriver):
-    def __init__(self, messager):
-        super(OpenBMCDriver, self).__init__(messager)
+    def handler(self, node, node_info, command, args, debugmode): 
+        openbmc = OpenBMC(node, node_info, self.messager, debugmode)
+        getattr(openbmc,command)(args)
```
chenglch: Call the rpower method directly, handler is not a meaningful method name compare with the behavior like `rpower`

xCAT-server/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -1,26 +1,223 @@
...
+            if key_type == 'bmc0' :
+                bmc_current_state = response['data'][key]['CurrentBMCState'].split('.')[-1]
+            if key_type == 'chassis0' :
+                chassis_current_state = response['data'][key]['CurrentPowerState'].split('.')[-1]
+            if key_type == 'host0' :
+                host_current_state = response['data'][key]['CurrentHostState'].split('.')[-1]
+
+        if subcommand == 'bmcstate' :
+            if bmc_current_state == 'Ready' :
+                return 'BMC Ready'
```
chenglch: Put the message in a constant dict, the return the message based on that dict.

xCAT-server/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -1,26 +1,223 @@
...
+    '''
+
+    handle rpower command
+
+    subcommand : on/off/bmcreboot/softoff/state/stat/status/bmcstate
+
+    '''
+    def rpower(self, args) :
+        subcommand = args[0]
+        self._login()
```
chenglch: I think _login() method should return something, it could not be success all the time.

xCAT-server/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -1,26 +1,223 @@
...
+
+    send login request
+
+    '''
+    def _login(self) :
+        login_url = http_protocol + self.bmcip + '/login'
+        login_data = { "data": [ self.username, self.password ] }
+        self.client.request('POST', login_url, OpenBMC.headers, login_data, self.node, 'login')
+
+    '''
```
chenglch: Take the code from pyghmi as a example, 
```
        def register_keepalive(self, cmd, callback):
        '''Register  custom keepalive IPMI command

        This is mostly intended for use by the console code.
        calling code would have an easier time just scheduling in their
        own threading scheme.  Such a behavior would naturally cause
        the default keepalive to not occur anyway if the calling code
        is at least as aggressive about timing as pyghmi
        :param cmd: A dict of arguments to be passed into raw_command
        :param callback: A function to be called with results of the keepalive

        :returns: value to identify registration for unregister_keepalive
        '''
        regid = random.random()
        if self._customkeepalives is None:
            self._customkeepalives = {regid: (cmd, callback)}
        else:
            while regid in self._customkeepalives:
                regid = random.random()
            self._customkeepalives[regid] = (cmd, callback)
        return regid
```

xCAT-server/lib/python/agent/xcatagent/xcat_exception.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python
+
+class SelfServerException(Exception) :
```
chenglch: Define the exception like this is meaningless, you could refer the design of openstack nova https://github.com/openstack/nova/blob/master/nova/exception.py  to make the error message more standard.
When the exception is captured, we could call `messager.error(exception.msg)` to send the error message back to the user directly.

xCAT-server/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -1,26 +1,230 @@
...
+        return RESULT_OK
+
+
+    def _set_power_onoff(self, subcommand) :
+        """ Set power on/off/softoff/bmcreboot
+        :param subcommand: subcommand for rpower
+        :returns: on/off/softoff/BMC reboot if success
+        :raise: error message if failed
+        """
+        request_url = HTTP_PROTOCOL + self.bmcip + RPOWER_URLS[subcommand]['url']
```
chenglch: I prefer url, data directly, the prefix of `request_` is meaningless here.

xCAT-server/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -1,26 +1,230 @@
...
-    def __init__(self, messager):
-        super(OpenBMCDriver, self).__init__(messager)
-
-    def rpower(self, node, info, args):
-        if node == 'node1':
-            self.messager
-            gevent.sleep(3)
-        self.messager.info(
-            "%s: rpower called info=%s args=%s" % (node, info, args))
+        super(OpenBMCManager, self).process_nodes_worker('openbmc', 'OpenBMC', self.nodes, nodeinfo, 'rpower', args, self.debugmode)
```
chenglch: You misunderstand again.
OpenBMCManager is to manage the collection of nodes. It call the method of OpenBMC inside the coroutine.
OpenBMC class is for one node, you should make it work without the OpenBMCManager.

xCAT-server/lib/python/agent/xcatagent/xcat_exception.py
xuweibj: But I do not want to show the message to the user directly. Whether show message should method decide.

chenglch: If you do not know how to design the exception, this file is not needed, just catch Exception instead, this approach is not suggested, but better than the meaningless self defined exceptions.

xCAT-server/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -1,26 +1,233 @@
...
-    def __init__(self, messager):
-        super(OpenBMCDriver, self).__init__(messager)
-
-    def rpower(self, node, info, args):
-        if node == 'node1':
-            self.messager
-            gevent.sleep(3)
-        self.messager.info(
-            "%s: rpower called info=%s args=%s" % (node, info, args))
+        super(OpenBMCManager, self).process_nodes_worker('openbmc', 'OpenBMC', self.nodes, nodeinfo, 'rpower', args)
```
chenglch: My concern is that you use reflect again which is unnecessary.  Why not to just create the openbmc objects, and trasfer the rpower func directly?

xCAT-server/lib/python/agent/client.py
```diff
@@ -40,18 +40,18 @@ def main(self, argv):
             return 0
 
         s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-        nodes = ['node%s' % i for i in range(100)]
-        nodeinfo = {node: {'username': 'admin', 'password': 'cluster'} for node
-                    in
-                    nodes}
+        nodes = ['node1', 'node2']
+        nodeinfo = {'node1':{'bmc':'9.3.185.161', 'bmcip':'9.3.185.161', 'username':'root', 'password': '0penBmc1'},
+                    'node2':{'bmc':'10.6.17.100', 'bmcip': '10.6.17.100', 'username':'root', 'password': '0penBmc'}}
```
robin2008: Don't put the the real IP into public repository, even it is for UT.

xCAT-server/lib/python/agent/client.py
```diff
@@ -40,18 +40,18 @@ def main(self, argv):
...
         s.connect(options.sock)
         req = {'module': 'openbmc',
                'command': 'rpower',
-               'args': ['-d', 'firmware'],
+               'args': ['state'],
                'cwd': os.getcwd(),
                'nodes': nodes,
-               'nodeinfo': nodeinfo}
+               'nodeinfo': nodeinfo,
+               'debugmode': 1}
```
robin2008: better to use `envs` to cover debugmode,  as we may have more variables which could passed into python client.

xCAT-server/lib/python/agent/xcatagent/server.py
```diff
@@ -86,9 +86,9 @@ def _handle(self, sock, address):
                 messager.error("Could not find manager for %s" % req['module'])
                 return
             nodes = req.get("nodes", None)
-            manager = manager_func(messager, req['cwd'], nodes)
+            manager = manager_func(messager, req['cwd'], nodes, req['envs'])
```
robin2008: @chenglch Maybe we can cache the whole `req` dict in manager,  I wonder we may have more variables in req from perl.

chenglch: @robin2008 , dict is too flexible, maybe we need a context class to help store the global attributes like this in the future.

xCAT-server/lib/python/agent/client.py
```diff
@@ -40,18 +40,18 @@ def main(self, argv):
             return 0
 
         s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-        nodes = ['node%s' % i for i in range(100)]
-        nodeinfo = {node: {'username': 'admin', 'password': 'cluster'} for node
-                    in
-                    nodes}
+        nodes = ['node1', 'node2']
+        nodeinfo = {'node1':{'bmc':'10.0.0.1', 'bmcip':'10.0.0.1', 'username':'root', 'password': 'xxxxxx'},
```
whowutwut: What are this used for? `nodes` and `nodeinfo` ?

whowutwut:
Before we allow this to merge this have a few questions: 

1. The error message is not the same with the Perl support: 
```
[root@stratton01 ~]# rpower p9euh01 state
p9euh01: Error: [401] Invalid username or password
```
2. How do we toggle between currently `mgt=openbmc` and this new one? 
3. How does this fit into RedFish support... I see a lot is defined in openbmc.py, could we create a base class that OpenBMC and RedFish could both be derived from...  
3. Can you show me how we can call this utility outside of xCAT?  It would make it dual purpose... 
   * This would lead us to package this python module in another rpm and the user would install it and get openbmc python support instead of having it combined into xCAT, or install it stand alone and be able to do everything xCAT can do via command line

    Then some more internal discussion questions based on the above...
    * integrate option parsing using docopt? http://docopt.org/
     * Consider using python package management or structure the layout so we could package with pip? 

    The initial idea I had was to have this as a "stand alone" python application that xCAT would call it's API with `--raw` to get JSON data returned and consume it as a JSON parser... Then users could code to it or we could use it stand alone outside xCAT. Let's discuss a little tonight.

xCAT-server/lib/python/agent/client.py
xuweibj: The client.py is used to debug python part code, "nodes" and "nodeinfo" are got from perl script and used in python script. We could modify here instead of running perl code to generate  data for test.

xuweibj:
Hi @whowutwut , 
For 1, I have modified the error msg. But now python will not return 503 or 504 error, so just print msg like "Error: Unable to connect to server", "Error: Timeout to connect to server".

For 2, For "rpower" command, before release, we will check mgt if "mgt=openbmc2", run python script. If release, will not run perl version again.

For 3, we can create "RedFish" class if we needed. In __init__.py will check the "module" to mapping class.  

For 4, we just could run python script for debug.
Open 2 terminals:
one run ``./agent.py --standalone  --sock agent.sock``
one run ``./client.py --sock agent.sock``
 all under "/opt/xcat/lib/python/agent"

whowutwut:
@robin2008 whomever merges, please set the current milestone at the time of the merge so that  it accurately reflects the correct release that the code went into and leverage GitHub queries

robin2008:
Now it is still in the dev branch,  and we will see the test result to decide if it is okay to in master branch. And for how to be delivered in master (just in xcat-server or separate package), it is still under discussion.

PR #4590, opened 2017-12-28
Add warning method for communication messager

chenglch:
This patch send the length of the buffer and the buffer bytes
in one system call.

partial: #4561


whowutwut:
@xuweibj please set milestone when merging pull requests so it matches the release that the change goes into

PR #4599, opened 2018-01-03
Remove deprecated SSH option KeyRegenerationInterval from remoteshell

mattaezell:
Avoid a syslog message like:
`/etc/ssh/sshd_config line 144: Deprecated option KeyRegenerationInterval`
That option is only for SSHv1, which has already been deprecated and is being removed from OpenSSH

### DIFF
--- a/xCAT/postscripts/remoteshell
+++ b/xCAT/postscripts/remoteshell
@@ -56,8 +56,6 @@ then
         #delete all occurance of the attribute and then add xCAT settings
         sed -i '/X11Forwarding /'d /etc/ssh/sshd_config
         echo "X11Forwarding yes" >>/etc/ssh/sshd_config
-        sed -i '/KeyRegenerationInterval /'d /etc/ssh/sshd_config
-        echo "KeyRegenerationInterval 0" >>/etc/ssh/sshd_config
         sed -i '/MaxStartups /'d /etc/ssh/sshd_config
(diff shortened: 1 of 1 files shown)


immarvin:
hi @mattaezell , is it possible to check the version of ssh version and keep the option for SSHv1?

mattaezell:
@immarvin I suppose, but doing that reliably in a platform-independent way could be tricky.  Does anyone know why that statement is there in the first place?  One might argue that `remoteshell` shouldn't be modifying sshd_config at all and any required configs should be handled in `genimage` or `packimage`/`syncfiles`

What actually brought this to light is we have customized sshd_config for our `login` image using `syncfiles` (and it gets built into the image when we run `packimage`).  We have a `Match` block at the end, but `remoteshell` puts some global options  after the match statement.

immarvin:
@mattaezell , I think it is there to keep consistent between diskful and diskless provision. 

the postscripts and postbootscripts specified in "xcatdefaults" will be run on the nodes provisioned by xCAT by default.
```
#tabdump postscripts
"xcatdefaults","syslog,remoteshell,syncfiles","otherpkgs",,
```

the behavior of postscripts `syslog` and `remoteshell` are similar, one advantage of modifying them instead of syncing is that we can only apply the options required by xCAT while  keeping the other parts in the default config file shipped in different os distribution and versions.

immarvin:
agree to merge

Issue #2258, opened 2016-12-05
Genesis image should now be centos7


arif-ali:
Hi chaps,

We've had numerous issues the past year, that would have helped to have the genesis image to be el7, I would suggest that we should do this, as at the moment, we have build a runimage image in el6 first, and find an el6 machine to do this on.

We should also add IB drivers into it, so that we are able to discover via 100G ethernet as well as 10G and 1G

thanks,
Arif

whowutwut:
Hi @arif-ali   Interesting timing....  I spent a good amount of time today re-creating the build machine to xCAT-genesis-base using CentOS 6.6...   I think @zet809 added IB drivers into the ppc64 genesis kernel.   Let me give it a quick try tonight and discuss with the team.  We are 1 week away from the 2.13 GA and not sure if we want to do it now.. but maybe it's the best time...  .  Would you have a chance to try out the xCAT-genesis-base for x86_64 if we provided a test rpm?

whowutwut:
@arif-ali  which version of centos are you using 7.0, 7.1, 7.2?

arif-ali:
Well, I have a workaround at the moment, that will hopefully work for the customer, which they will try tomorrow, i.e. 

```
mkdir mlx5_test
wget http://vault.centos.org/6.6/os/x86_64/Packages/kernel-2.6.32-504.el6.x86_64.rpm
rpm2cpio kernel-2.6.32-504.el6.x86_64.rpm | cpio -i --make-directories
mkdir -p /opt/xcat/share/xcat/netboot/genesis/x86_64/fs/lib/modules/2.6.32-504.el6.x86_64/kernel/drivers/net/mlx5/core
cp ./lib/modules/2.6.32-504.el6.x86_64/kernel/drivers/net/mlx5/core/mlx5_core.ko /opt/xcat/share/xcat/netboot/genesis/x86_64/fs/lib/modules/2.6.32-504.el6.x86_64/kernel/drivers/net/mlx5/core/
mknb x86_64
```

But no doubt, if we gave the customer to test the genesis image with a el7, then they would be happy to test. I doubt much that the version of CentOS will matter much in this specific case, as for runimage, the shared libraries would just work. The key things I think, that we we need in the genesis image should include, Mellanox 100G driver (mlx5_core), and the Intel OPA driver (hfi1), in addition to what we already have for mlx4 and ipath

whowutwut:
[...]

PR #4609, opened 2018-01-08
Improvements to xcat init unit for dracut

samveen:
Changes:
- Explicit shebang (make it explicit). It not required but clarifies syntax used.
- ShellCheck suggested fixes.
- Correct mirroring of `doxcat` to first physical cons if serial cons is primary.
- Add protection for accidental closure of primary screen.

### DIFF
--- a/xCAT-genesis-builder/xcat-cmdline.sh
+++ b/xCAT-genesis-builder/xcat-cmdline.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 root=1
 rootok=1
 netroot=xcat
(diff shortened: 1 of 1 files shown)


xCAT-genesis-builder/xcat-cmdline.sh
```diff
@@ -35,34 +38,30 @@ mkdir -p /var/lib/dhclient/
 mkdir -p /var/log
 ip link set lo up
 echo '127.0.0.1 localhost' >> /etc/hosts
-if grep console=ttyS /proc/cmdline > /dev/null; then
-	while :; do sleep 1; screen -x console < /dev/tty1 > /dev/tty1 2>&1; clear; done &
+if grep -q console=ttyS /proc/cmdline; then
+        while :; do sleep 1; screen -S console -ln screen -x doxcat </dev/tty1 &>/dev/tty1; clear &>/dev/tty1 ; done &
```
immarvin: hi @samveen , what's the modification in this line for?

samveen: @immarvin 
- `grep -q`  gets rid of the `>/dev/null`
- With this change to `screen`, The same console is now available in multi-display mode on both `TTYS1` and `tty1` : Earlier, `screen -x console < /dev/tty1 > /dev/tty1 2>&1` was always failing as there was no `screen` named `console` but due to I/O redirection, no error was displayed ever. This caused constant screen blanking with Centos 7 genesis if you remember (because the loop continuously run on failed screen).
- Redirection of clear clears the tty1 instead of the `rcons` console.

samveen: @immarvin 
- `grep -q`  gets rid of the `>/dev/null`
- With this change to `screen`, The same console is now available in multi-display mode on both `TTYS1` and `tty1` : Earlier, `screen -x console < /dev/tty1 > /dev/tty1 2>&1` was always failing as there was no `screen` named `console` but due to I/O redirection, no error was displayed ever. This caused constant screen blanking with Centos 7 genesis if you remember (because the loop continuously run on failed screen).
- Redirection of clear clears the tty1 instead of the `rcons` console.

zet809:
hi, @samveen , my testing shows that the screen is cleared every second issue mentioned in #2258 is fixed with this PR, thx very much and I will do a new xCAT-genesis-base build for both x86_64 and ppc64. Thx again.

xCAT-genesis-builder/xcat-cmdline.sh
zet809: Hi, @samveen , since we don't sure why here using '-x console' before, and no issues encountered, I suggest switch back to `screen -S console -ln screen -x console` rather than `screen -S console -ln screen -x doxcat`.

immarvin: hi @jjohnson42 , would you please take a look at the modification on the `screen` stuff in this PR? thanks

Issue #4600, opened 2018-01-04
[OpenBMC] reventlog logging Resolve attribute should only show up with CALLOUT properties in the entry


whowutwut:
Resolved a bunch of logging events and the LED fault lights turned off on the Witherspoon boxes, but when I checked there was 1 log message with a CRITICAL error that was not cleared.  So I asked FW team why the LED lights turned off, response was: 

> only logs with a callout are involved in LEDs 

So it looks like we display `Resolved: ` on all log messages and it might make the Admin think they need resolve all issues to clear LED fault lights.    I think we (xCAT) should only display "Resolved: " when we have CALLOUT in the Additional Data. 

```
    "/xyz/openbmc_project/logging/entry/15": {
      "AdditionalData": [
        "CALLOUT_DEVICE_PATH=/sys/devices/platform/gpio-fsi/fsi0/slave@00:00/raw",
        "CALLOUT_ERRNO=0",
        "_PID=1766"
      ],
      "Id": 15,
      "Message": "org.open_power.Proc.FSI.Error.MasterDetectionFailure",
      "Resolved": 1,
      "Severity": "xyz.openbmc_project.Logging.Entry.Level.Error",
      "Timestamp": 1512622364888,
      "associations": []
    }
```

whowutwut:
Thinking about this some more, it might be useful to have `Resolved: <>` still printed out incase Admins want to record issues that are fixed in the logging events on the BMC. 

So it might be better to indicate which logs will contribute to LED faults....

Suggest something like: 

`Resolved: <value> [LED]` so that it's grep-able....

gurevichmark:
An easy was to create an entry with CALLOUT is to run on BMC:
`busctl set-property xyz.openbmc_project.Inventory.Manager /xyz/openbmc_project/inventory/system/chassis/motherboard/fan3 xyz.openbmc_project.State.Decorator.OperationalStatus Functional b 0`

whowutwut:
Found a node... 
```
[root@briggs01 Logging]# rvitals $NODE leds
mid05tor12cn05: Front . . . . . : Power:On Fault:On Identify:Off
mid05tor12cn05: Front Fans  . . : No LEDs On
mid05tor12cn05: Rear  . . . . . : Power:On Fault:On Identify:Off
```

```
[root@briggs01 Logging]# reventlog $NODE | grep LED
[...]

PR #4616, opened 2018-01-09
Add LED tag to the output of reventlog

gurevichmark:
#4600 

Add LED tag to the output of reventlog if the log entry contains a CALLOUT in additional data.


gurevichmark:
Output for the log entry with CALLOUT data:
`Resolved: <event log entry> [LED]`

xuweibj:
Looks good to me.

Issue #4597, opened 2018-01-03
 logger plugin support in goconserver


zet809:
* [x] interface and configuration framework to support different logger plugins
* [x] design a pipeline structure
* [x] construct the json message from console lines. 
* [x] tcp logger , ssl support
* [x] udp logger
* [ ] support redirect the console message to multiple targets
~~* [ ] makegocons -f <configuration_file> to accept the configuration customized by user. Failed to add support for this item due to the noderange design of xcat.~~

ref: https://github.com/chenglch/goconserver/issues/14

PR #4633, opened 2018-01-15
Modify the default configuration for gocons

chenglch:
As output plugins are suported by goconserver, this patch
modify the configuration file accordingly.

partial: #4597


zet809:

Issue #3724, opened 2017-08-18
[OpenBMC] Implement a new command to control OpenPOWER P9 behavior when PSU is cycled


whowutwut:
I found out that the OpenBMC firmware, in later versions, will remember the last state of the Host power when the power is physically cycled (power cable pulled) .  So if the host was booted before power was lost, and then applied again, it will boot up.  

This differs from the older OpenPOWER servers behavior where if power is pulled and re-applied, the BMC will boot up ONLY, but the host will be powered off until we run `rpower <> on/boot`  

From a conversation... 

OpenBMC Dev:
> Newer code remembers the last requested state, and after a PDU cycle, if the last requested state was a power on, it will auto power on.  so be sure to just power off before doing bmc reboots if you don’t want to auto power back on.  There’s also a setting you can set to disable that

Me: 
> Thanks will look into this..  We would have to disable this and continue to disable on each update of the FW?

OpenBMC Dev:
> it will be a persistent setting, so will be saved accross firmware updates.  With the new BMC code update coming, it may be reset with updates until we get that in.

So,  need to keep this item open and find out the flag and provide a new external that allows the user to be able to switch this behavior on or off.. 

Should target the 2.13.7 release...

whowutwut:
Here's the documentation for the OpenBMC API: [RestorePolicy](https://github.com/openbmc/phosphor-dbus-interfaces/blob/master/xyz/openbmc_project/Control/Power/RestorePolicy.interface.yaml)

whowutwut:
I've opened issue https://github.com/openbmc/openbmc/issues/2186 to change back the policy to default AlwaysOff.  But we should come up with this function support....  

External Design needed for this to start discussion and move it forward.   It's lower priority, but if there are cycles, we can implement it now.

gurevichmark:
To query:

```
[root@briggs01 ~]#curl -k -b cjar -X GET -H "Content-Type: application/json" https://172.11.139.13/xyz/openbmc_project/control/host0/power_restore_policy
{
  "data": {
    "PowerRestorePolicy": "xyz.openbmc_project.Control.Power.RestorePolicy.Policy.AlwaysOn"
  },
  "message": "200 OK",
  "status": "ok"
}
```

To set:
```
[...]

PR #4634, opened 2018-01-15
Support for OpenBMC PowerRestorePolicy attribute

gurevichmark:
#3724 

Implement support for OpenBMC PowerRestorePolicy using attribute definition.

```
[root@briggs01 xCAT_plugin]# rspconfig mid05tor12cn15 powerrestorepolicy
mid05tor12cn15: PowerRestorePolicy: AlwaysOn
[root@briggs01 xCAT_plugin]# rspconfig mid05tor12cn15 powerrestorepolicy=always_off
mid05tor12cn15: BMC Setting PowerRestorePolicy...
[root@briggs01 xCAT_plugin]# rspconfig mid05tor12cn15 powerrestorepolicy
mid05tor12cn15: PowerRestorePolicy: AlwaysOff
[root@briggs01 xCAT_plugin]#
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -539,6 +539,14 @@ my %status_info = (
     RSPCONFIG_API_CONFIG_OFF_RESPONSE => {
         process        => \&rspconfig_api_config_response,
     },
+    RSPCONFIG_API_CONFIG_ATTR_REQUEST  => {
+        method         => "PUT",
+        init_url       => "$openbmc_project_url",
+        data           => "false",
+    },
+    RSPCONFIG_API_CONFIG_ATTR_RESPONSE => {
+        process        => \&rspconfig_api_config_response,
+    },
     RSPCONFIG_API_CONFIG_QUERY_REQUEST  => {
         method         => "GET",
         init_url       => "$openbmc_project_url",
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -3415,11 +3452,21 @@ sub rspconfig_api_config_response {
                         last;
                     }
                 }
-                if (scalar($value) >= 0) {
+                if (($value eq "0") || ($value eq "1")) {
```
zet809: Hi, @gurevichmark , shall we translate 0/1 to a string?

gurevichmark: If we do not treat it as a string, then something like:
`if (($value == 0) || ($value == 1)) {` or 
`if ((scalar($value) = 0) || (scalar($value) = 1)) {`
will get evaluated to be true when value is `abc`. I think that is because any setting of $value is evaluated to be 1 when treated as integer.
In this code I would like to treat boolean values, like 0 or 1 differently for display purposes.

zet809: OK, here, that is no problem, but for the output in line 3457, shall we print enable/disable rather than 1/0?

gurevichmark: I am not sure if "Enable/Disable" will make sense for all boolean attribute setting. For some "On/Off" might make more sense. Also, I am thinking it might be confusing if when doing a query, the output is "Enable" or "Disable", but then to set the value, user has to use `attribute=1` or `attribute=0`. They might attempt to use `attribute=Enable` and will get an error.

zet809:
OK, let's merge and try.

PR #4643, opened 2018-01-16
Fix IMM MAC data check

jjohnson42:
The IMM firmware is not guaranteed to offer up data.  As a result, protect the code from missing data.


zet809:

PR #4647, opened 2018-01-17
From 2.13.10, the genesis scripts is moved to usr/bin or usr/sbin

zet809:
From 2.13.10, the genesis scripts will be moved to usr/bin or usr/sbin from bin or sbin.


neo954:

Issue #4629, opened 2018-01-12
need empty-zones-enable no; in named.conf


adambertsch:
For sites that would like to forward to their own DNS servers, the site table includes the 'forwarders' attribute to allow for this.  If the site internal DNS resolves RFC1918 addresses, xCAT dns doesn't work.  The reason is that empty-zones-enable in BIND defaults to yes, which creates by default empty reverse zones for all rfc1918 address blocks.

Please allow for a site option to add empty-zones-enable no; to options {} in named.conf for xCAT master node.

bybai:
hi @adambertsch ， thanks your requirements, we will consider your mentioned.

bybai:
Hi @adambertsch , could you provide your failed examples? for excample, 'forwarders', master, nameservers ,domain from site table, /etc/resolv.conf, /etc/hosts,  related network entry from networks table,  makedns outputs etc. Let me reproduce this in my environment. Thanks.

adambertsch:
The issue is fairly simple.  Current makedns generates a named.conf that looks like this:
 
#generated by xCAT: /opt/xcat/sbin/makedns command 
options {
        directory "/var/named/";
        allow-recursion { any; };
        forwarders {
                x.x.1.x;
                x.x.2.x;
        };
};

And I need to generate this one:

#generated by xCAT: /opt/xcat/sbin/makedns command 
options {
        directory "/var/named/";
        allow-recursion { any; };
        forwarders {
                x.x.1.x;
                x.x.2.x;
        };
        empty-zones-enable no;
};

bybai:
@adambertsch , thanks your response.

gzjfn4:
@adambertsch @bybai. PMR: 36104,227,000 was created to track this issue.

bybai:
Hi @gzjfn4,  I am working on this. Thanks

bybai:
Hi @adambertsch and @gzjfn4 , could you confirm your OS?

gzjfn4:
@bybai - the OS is Red Hat Enterprise Linux version 7.3 on Power 8 Little Endian and 7.4 on Power 9 Little Endian

bybai:
Hi @adambertsch and @gzjfn4 ,
New fix for this request will be in xCAT 2.13.10 build that will be release next week.
Now you can pick up new code and do the following to work around your problem, if any problem, contact me.

The steps:

[...]

PR #4655, opened 2018-01-19
format emptyzonesenable

bybai:
For #4629 

Format the emptyzonesenable description.
The same UT in #4629


zet809:
The fix looks good to me, let's merge it to try RH build. @bybai pls help to continue look at the CI failed problem. Thx a lot!

PR #4657, opened 2018-01-19
Add gocons setup doc on service node

chenglch:
Add console state 2 in the servicenode table to help set up
goconserver daemon.


zet809:
Looks good, I agree to merge. thx!

Issue #4658, opened 2018-01-19
[openbmc]rpower state return error message


bybai:
openbmc in python, reproduce:
```
[root@c910f03c11k08 rpm]# rpower f6u17 state
f6u17: Error: Only one option is supported at the same time for rpower

[root@c910f03c11k08 rpm]# lsxcatd -v
Version 2.13.10 (git commit 179b1cdf6c9a642eafc2d53e3f9b92d7fb559ecb, built Thu Jan 18 00:59:12 EST 2018)
```

PR #4660, opened 2018-01-19
fix issue 4658, parameter check for rpower openbmc

xuweibj:
#4658


bybai:
agree to merge. Thx @xuweibj

Issue #4143, opened 2017-10-20
[fvt] configure ssh in cumulus switch is not stable


hu-weihua:
Configuring ssh in cumulus switch by ``switchdiscover``  is not stable.

After predefining a switch, run ``switchdiscover --range 172.11.205.25 -w --setup`` first time, the ssh is configured unsuccessfully, but try again without any change, it works.
```
[root@briggs01 hwh]# switchdiscover --range 172.11.205.25 -w --setup
Discovering switches using nmap for 172.11.205.25. It may take long time...
ip          	name                	vendor                                            	mac
------------	------------        	------------                                      	------------
172.11.205.25	mid05tor25          	Edgecore Networks Switch                          	A8:2B:B5:59:AC:40
switch discovered and matched: mid05tor25 to mid05tor25
Writing the data into xCAT DB....
Call to config mid05tor25: /opt/xcat/share/xcat/scripts/configonie --switches mid05tor25 --all

output =

[root@briggs01 hwh]# lsdef mid05tor25
Object name: mid05tor25
    groups=switch,edge_switch
    ip=172.11.205.25
    mac=A8:2B:B5:59:AC:40
    mgt=switch
    nodetype=switch
    otherinterfaces=172.11.205.25
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    snmpauth=sha
    snmppassword=xcatpassw0rd
    snmpprivacy=DES
    snmpusername=xcatadmin
    snmpversion=3
    status=Matched
    switch=mgmtsw01
    switchport=25
    switchtype=onie
    usercomment=Edgecore Networks Switch

[root@briggs01 hwh]# switchdiscover --range 172.11.205.25 -w --setup
Discovering switches using nmap for 172.11.205.25. It may take long time...
ip          	name                	vendor                                            	mac
------------	------------        	------------                                      	------------
172.11.205.25	mid05tor25          	Edgecore Networks Switch                          	A8:2B:B5:59:AC:40
switch discovered and matched: mid05tor25 to mid05tor25
Writing the data into xCAT DB....
Call to config mid05tor25: /opt/xcat/share/xcat/scripts/configonie --switches mid05tor25 --all

output = ssh configured for mid05tor25
switch: mid05tor25 configured
```

hu-weihua:
Related test task #4069

cxhong:
couldn't re-created.  I added some print statement on the scripts. if you ran into this again,  can u show me the output?

[...]

PR #4722, opened 2018-01-25
remove remote host key from known_hosts file

cxhong:
For issue #4143.

The remote host keys for hostname/ip needs to remove first when configure the onie switch via **configonie --all**  script.


immarvin:
i am ok with the change

Issue #4718, opened 2018-01-24
[OpenBMC] Add support for modifying BootMode settings


whowutwut:
With the framework set in place, we can easily add support for things like BootMode: https://github.com/openbmc/phosphor-dbus-interfaces/blob/master/xyz/openbmc_project/Control/Boot/Mode.interface.yaml and allow the user to leverage xCAT to set this attributes instead of running CURL commands manually.  (similar to `RSPCONFIG_POWERRESTORE_POLICY`)

I'm not sure what the use case is for this ,we should discuss with Firmware team to understand when we would want to set this ..

gurevichmark:
Implemented by #4727

Issue #4715, opened 2018-01-24
[OpenBMC] Update PowerSupplyRedundancy to use the correct endpoints 


whowutwut:
Talking to the Firmware team today, the implementation of Power Supply is using the old interface and the correct way to get/set the value is using the `setValue` and `getValue` endpoints: 

> I think the PS sensor, even though the object is in the xyz/ namespace, implements the old PS interface. The right way to set a value is by calling 'SetValue' or something equivalent.

Those are these commands:

```
curl -b cjar -k -H "Content-Type: application/json" -X POST \
https://$BMC_IP/xyz/openbmc_project/sensors/chassis/PowerSupplyRedundancy/action/getValue \
-d '{"data": []}'

curl -b cjar -k -H "Content-Type: application/json" -X POST \
https://$BMC_IP/xyz/openbmc_project/sensors/chassis/PowerSupplyRedundancy/action/setValue \
-d '{"data": ["Enabled"]}'

curl -b cjar -k -H "Content-Type: application/json" -X POST \
https://$BMC_IP/xyz/openbmc_project/sensors/chassis/PowerSupplyRedundancy/action/setValue \
-d '{"data": ["Disabled"]}'
```

Update the xCAT code to use these correct endpoints

zet809:
Verified, so close this issue.
```
[root@stratton01 ~]# rspconfig f6u17 powersupplyredundancy | grep PowerSupply
Tue Jan 30 00:48:05 2018 f6u17: [openbmc_debug] curl -k -b cjar -X POST -H "Content-Type: application/json" -d '{"data":[]}' https://10.6.17.100/xyz/openbmc_project/sensors/chassis/PowerSupplyRedundancy/action/getValue
f6u17: BMC PowerSupplyRedundancy : Disabled
[root@stratton01 ~]# rspconfig f6u17 powersupplyredundancy=enabled | grep PowerSupply
Tue Jan 30 00:48:41 2018 f6u17: [openbmc_debug] curl -k -b cjar -X POST -H "Content-Type: application/json" -d '{"data":["Enabled"]}' https://10.6.17.100/xyz/openbmc_project/sensors/chassis/PowerSupplyRedundancy/action/setValue
f6u17: BMC Setting BMC PowerSupplyRedundancy...
```

PR #4727, opened 2018-01-25
[OpenBMC] Support bootmode generically

whowutwut:
This function is used to view/set the host bootmode.  This is really added to help debug some strange issues that we are seeing where certain machines are stopping at Petitboot and not booting off the network on first discovery. 

The 4th commit is related to PowerSupplyRedundancy (tag #4715) in that the value takes effect on a IPL, not a bmc reboot.  A bmcreboot actually removes the setting. 

I'd like to get this into the 2.13.10 release tomorrow. 

UT:
```
[root@stratton01 Control]# rspconfig p9euh01 powersupplyredundancy=disablEd
p9euh01: BMC Setting BMC PowerSupplyRedundancy...
[root@stratton01 Control]# rspconfig p9euh01 bootmode
p9euh01: BMC BootMode : Regular
[root@stratton01 Control]# rspconfig p9euh01 bootmode=setup
p9euh01: BMC Setting BMC BootMode...
[root@stratton01 Control]# rspconfig p9euh01 bootmode
p9euh01: BMC BootMode : Setup
[root@stratton01 Control]# rspconfig p9euh01 bootmode=regular
p9euh01: BMC Setting BMC BootMode...
[root@stratton01 Control]# rspconfig p9euh01 bootmode
p9euh01: BMC BootMode : Regular
[root@stratton01 Control]#
```

On errror, we would see the following msg: 
```
[root@stratton01 Control]# rspconfig p9euh01 bootmode
p9euh01: BMC BootMode : Regular
p9euh01: WARNING: Unexpected value set: xyz.openbmc_project.Control.Boot.Mode.Modes.Regular
p9euh01: WARNING: Valid values: xyz.openbmc_project.Control.Boot.Mode.Safe,xyz.openbmc_project.Control.Boot.Mode.Regular,xyz.openbmc_project.Control.Boot.Mode.Setup
```


zet809:
hi, @whowutwut , why not implement the boot mode interface to rsetboot as I suggest in https://ibm-systems-storage.slack.com/archives/C68N2UK0X/p1516868331000070?thread_ts=1516820529.000339&cid=C68N2UK0X?
For `rsetboot`, it is used to set boot order, and also can be used to set boot mode, right?

whowutwut:
@zet809 What would be the suggested interface here if we tagged to rspconfig? 

The rsetboot is temporarily setting the next boot of the server, where this bootmode in my opinion is really a debug attribute.  The use case here is to run across a large number of systems and make sure that the value is "Regular" I don't think customer will be messing with this value at all.  MFG uses it when pre-loading the server, but should return the value to Regular when ship out. 

So really, adding this attribute is really to help debug and allow xCAT to display this OpenBMC configuration value more easily

zet809:
Hi, @whowutwut , I think you are right that bootmode mainly useful for debugging.
But `rspconfig` is mostly configuration for BMCs, but bootmode works for host, right?

whowutwut:
Resolves #4718

whowutwut:
@zet809 Yes, bootmode controls the behavior I believe up to Petitboot... when the host IPLs..

PR #4737, opened 2018-01-26
reminder user to rebuild xcat-genesis-base to update xCAT newer than 2.13.10

zet809:
Reminder to rebuild xcat-genesis-base to update xCAT2.13.10 or newer.


hu-weihua:
It is ok for me, thinks @zet809

Issue #4739, opened 2018-01-26
the incompatible between "pdu" table schema and "pdu" object


immarvin:
this is introduced the a PR from @cxhong 
```
commit a1efbea5e3c46426dbfb206ab7d4d9887c4215e7
Author: Casandra Qiu <cxhong@us.ibm.com>
Date:   Wed Nov 29 14:53:42 2017 -0500

    Relay action and snmp configuration support for Coral PDU

diff --git a/perl-xCAT/xCAT/Schema.pm b/perl-xCAT/xCAT/Schema.pm
index cf2f6a0..65c69ec 100755
--- a/perl-xCAT/xCAT/Schema.pm
+++ b/perl-xCAT/xCAT/Schema.pm
@@ -694,17 +694,25 @@ passed as argument rather than by table value',
         },
     },
     pdu => {
-        cols => [qw(node nodetype outlet machinetype modelnum serialnum comments disable)],
+        cols => [qw(node nodetype pdutype outlet username password snmpversion community snmpuser authtype authkey privtype privk
         keys         => [qw(node)],
         nodecol      => "node",
         table_desc   => 'Parameters to use when interrogating pdus',
         descriptions => {
             node => 'The hostname/address of the pdu to which the settings apply',
             nodetype => 'The node type should be pdu ',
+            pdutype => 'The type of pdu ',
             outlet => 'The pdu outlet count',
-            machinetype => 'The pdu machine type',
-            modelnum    => 'The pdu model number',
-            serialnum   => 'The pdu serial number',
+            username => 'The remote login user name',
+            password => 'The remote login password',
+            snmpversion => 'The version to use to communicate with switch.  SNMPv1 is assumed by default.',
+            community => 'The community string to use for SNMPv1/v2',
+            snmpuser => 'The username to use for SNMPv3 communication, ignored for SNMPv1',
+            authtype => 'The authentication protocol(MD5|SHA) to use for SNMPv3.',
+            authkey => 'The authentication passphrase for SNMPv3 ',
+            privtype => 'The privacy protocol(AES|DES) to use for SNMPv3.',
+            privkey => 'The privacy passphrase to use for SNMPv3.',
+            seclevel => 'The Security Level(noAuthNoPriv|authNoPriv|authPriv) to use for SNMPv3.',
             comments => 'Any user-written notes.',
             disable  => "Set to 'yes' or '1' to comment out this row.",
         },
@@ -2950,24 +2958,64 @@ my @nodeattrs = (
[...]

PR #4750, opened 2018-01-29
Remove not existing table fields in PDU table

cxhong:
For issue #4739 

Remove not existing attributes for pdu table. machine type, model number and serial number.  Those information can get from rinv command.

The current attributes for PDU table are:
```
# tabdump pdu
#node,nodetype,pdutype,outlet,username,password,snmpversion,community,snmpuser,authtype,authkey,privtype,privkey,seclevel,comments,disable
````


immarvin:
agree to merge

Issue #4683, opened 2018-01-23
goconserver packaging from binary tar ball


zet809:
Investigate:
* [x] in xcat-core, shall include packaging script based on prebuild goconserver binary tar ball ==> The build script to build binary, packaging have been put in xcat2/goconserver repo.
* [x] to have xCAT, xCATsn recommand on goconserver

zet809:
Let's close this task as the items are all finished.

PR #4768, opened 2018-01-31
Add package dependency for goconserver

neo954:
This is for task #4683


whowutwut:
@neo954 @zet809  can we please set milestones for the issues as well as labels as best you can when we merge pull requests, this helps us be able to run queries leveraging GitHub filters...   

I will let you modify it for this PR.

PR #4772, opened 2018-01-31
Add code to support timesyncmethod for OpenBMC using rspconfig

### LINKED ISSUE #4771: [OpenBMC] Additional support needed to enable NTP time server sync in BMC
Issue https://github.com/xcat2/xcat-core/issues/4472 initially added the ability to set/get a list of NTP servers for the BMC to configure and sync time against. 

Are we are testing this, turns out there are some issues that cause this to not work.  Lack of documentation here: https://github.com/openbmc/phosphor-dbus-interfaces/tree/master/xyz/openbmc_project/Time 

What we have found out in the last few days: 
* For the settings to be applied, the HOST needs to be power cycled https://github.com/openbmc/phosphor-time-manager#special-note-on-host-on
* The Sync Method needs to be manually set to NTP
```
curl -c cjar -b cjar -k -H 'Content-Type: application/json' -X PUT \
-d '{"data": "xyz.openbmc_project.Time.Synchronization.Method.NTP" }' \
https://$BMCIP/xyz/openbmc_project/time/sync_method/attr/TimeSyncMethod
```
* Still some FW issues exists because it seems like `/etc/systemd/timesyncd.conf` is not modified by any of the RestAPI commands and needs to be restarted for the NTP to sychronize. 
Internal Issue: https://w3.rchland.ibm.com/projects/bestquest/?defect=SW415613

It seems like from xCAT's perspective, the following should be done. 
1. `rspconfig <noderange> time_sync_method=NTP`
2. `rspconfig <noderange> ntpservers=IP1,IP2,etc`
3. Something needs to be fixed on the BMC to set this correctly and restart timesyncd (pending response from FW team)
4. Then power cycle the host for it to apply...

We did see that rebooting the bmc seems to sync up the time, but need to gather the output of `timedatectl status` and `journalctl` to see if we actually sync'ed with a time server or not

whowutwut:
Resolves #4771 

UT: 
```
[root@fs2vm110 Time]# rspconfig f6u17 timesyncmethod
f6u17: BMC TimeSyncMethod : NTP
[root@fs2vm110 Time]# rspconfig f6u17 timesyncmethod=manual
f6u17: BMC Setting BMC TimeSyncMethod...
[root@fs2vm110 Time]# rspconfig f6u17 timesyncmethod
f6u17: BMC TimeSyncMethod : Manual
[root@fs2vm110 Time]# rspconfig f6u17 timesyncmethod=ntp
f6u17: BMC Setting BMC TimeSyncMethod...
[root@fs2vm110 Time]# rspconfig f6u17 timesyncmethod
f6u17: BMC TimeSyncMethod : NTP
```

### DIFF
--- a/xCAT-client/pods/man1/rspconfig.1.pod
+++ b/xCAT-client/pods/man1/rspconfig.1.pod
@@ -26,6 +26,14 @@ B<rspconfig> I<noderange> B<garp>=I<time>
 
 B<rspconfig> I<noderange> {B<ipsrc>|B<ip>|B<netmask>|B<gateway>|B<hostname>|B<vlan>|B<sshcfg>}
 
+B<rspconfig> I<noderange> B<autoreboot>
+
+B<rspconfig> I<noderange> B<autoreboot={0|1}>
+
+B<rspconfig> I<noderange> B<bootmode>
+
+B<rspconfig> I<noderange> B<bootmode={safe|regular|setup}>
+
 B<rspconfig> I<noderange> B<dump> [B<-l>|B<--list>] [B<-g>|B<--generate>] [B<-c>|B<--clear> {I<id> | B<all>}] [B<-d>|B<--download> {I<id> | B<all>}]
 
 B<rspconfig> I<noderange> B<powerrestorepolicy>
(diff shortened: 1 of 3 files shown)


xCAT-client/pods/man1/rspconfig.1.pod
```diff
@@ -344,23 +348,23 @@ The subnet mask.
...
 
 =item B<bootmode>
 
-Display or control BMC Boot Mode attribute setting.
+Display or control BMC Boot Mode attribute setting. [OpenBMC]
 
 =item B<dump>
 
-Manage OpenBMC system dumps. If no sub-option is provided, will generate, wait, and download the dump.
+Generate BMC system dump. If no sub-option is provided, will generate, wait, and download the dump. [OpenBMC]
```
gurevichmark: Probably not very important, but the "dump" subcommand, in addition to generating dumps, also has options to "list", "remove" and "download". Maybe "Generate and manage" will covers them all ?

whowutwut: Let me fix it, I was trying to alphabetize the list but then it just became a big mess so decided not to :-)

PR #4775, opened 2018-02-01
print go-xcat.log each time when we run go-xcat

tingtli:
print go-xcat.log each time when we run go-xcat


hu-weihua:

PR #4778, opened 2018-02-01
Add boot nvme support and filter small devices

jjohnson42:
nvme devices were not autosensed as install candidates.

Also, devices smaller than 128 MB are filtered out, as they are likely
to be some media or similar.

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/getinstdisk
+++ b/xCAT-server/share/xcat/install/scripts/getinstdisk
@@ -37,13 +37,19 @@ if [ -z "$install_disk" ]; then
 
     # Get all partitions and disks from /proc/partitions file
     if [ -z "$has_awk" ]; then
-        entries=$(cat /proc/partitions | sed 's/  */ /g' | cut -d " " -f5 | grep -v "name" | grep -e "[s|h|v]d.*$")
+        entries=$(cat /proc/partitions | sed 's/  */ /g' | cut -d " " -f5 | grep -v "name" | grep -E '^[s|h|v]d|nvme')
     else
-        entries=$(awk -F ' '  '{print $4}' /proc/partitions | grep -v "name" | grep -e "[s|h|v]d.*$")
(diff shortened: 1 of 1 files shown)


whowutwut:
@zet809 @immarvin could you take a look at this?

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -37,13 +37,19 @@ if [ -z "$install_disk" ]; then
 
     # Get all partitions and disks from /proc/partitions file
     if [ -z "$has_awk" ]; then
-        entries=$(cat /proc/partitions | sed 's/  */ /g' | cut -d " " -f5 | grep -v "name" | grep -e "[s|h|v]d.*$")
+        entries=$(cat /proc/partitions | sed 's/  */ /g' | cut -d " " -f5 | grep -v "name" | grep -E '^[s|h|v]d|nvme')
```
immarvin: if the NVME device names follows the format `nvmeXnYpZ`, should this be `grep -E '^([s|h|v]d|nvme)'`?

xCAT-server/share/xcat/install/scripts/getinstdisk
```diff
@@ -37,13 +37,19 @@ if [ -z "$install_disk" ]; then
...
     else
-        entries=$(awk -F ' '  '{print $4}' /proc/partitions | grep -v "name" | grep -e "[s|h|v]d.*$")
+        entries=$(awk -F ' '  '{print $4}' /proc/partitions | grep -v "name" | grep -E '^[s|h|v]d|nvme')
```
immarvin: same as above

xCAT-server/share/xcat/install/scripts/getinstdisk
jjohnson42: That sounds like it should make sense

immarvin:
agree to merge

PR #4781, opened 2018-02-02
dodiscovery: fix syntax error in case flag_serial is not defined

kcgthb:
If `$flag_serial` is empty, `dodiscovery` generates an error:

```
[...]
xcat.genesis.dodiscovery: Beginning echo information to discovery packet file...
/usr/bin/dodiscovery: line 157: [: too many arguments
[...]
```

This PR fixes this.


zet809:
Looks good, I agree to merge, thx!

Issue #4674, opened 2018-01-23
[OpenBMC]rspconfig in python


zet809:
rspconfig support
* [x] get ip, ipsrc, netmask, 'gateway','vlan','hostname', 'ntpserver' ==> #4827 
* [x] get/set 'bootmode','autoreboot','powersupplyredundancy','powerrestorepolicy' ==> #4786 
* [x] sshcfg, bmc hostname, ip=dhcp ==> #4827 
* [x] set ntpserver, admin_passwd, gard ==> #4847 
* [x] dump  ==> #4838 
* [x] set bmc ip/netmask/gateway/vlan ==> #4884

whowutwut:
I vote this up on the backlog since we may need to run commands against the BMC at scale.

zet809:
* [x] get ip, ipsrc, netmask, 'gateway','vlan','hostname', 'ntpserver' ==> #4827 
* [x] get/set 'bootmode','autoreboot','powersupplyredundancy','powerrestorepolicy' ==> #4786 
* [x] sshcfg, bmc hostname, ip=dhcp ==> #4827 
* [x] set ntpserver, admin_passwd, gard ==> #4847 
* [x] dump  ==> #4838 
* [ ] set bmc ip/netmask/gateway/vlan

zet809:
Move to 2.14 since the rspconfig set ip/netmask/gateway/vlan will be merged then.

zet809:
Let's close this EPIC since all the subitems are finished. Thx @xuweibj and @whowutwut

PR #4786, opened 2018-02-04
Rspconfig framework for openbmc in python

zet809:
For task #4674 

The latest usage for rspconfig in python is:
```Handle rspconfig operations.

Usage:
       rspconfig -h|--help
       rspconfig -v|--verbose
       rspconfig dump [[-l|--list] | [-g|--generate] | [-c|--clear <arg>] | [-d|--download <arg>]]
       rspconfig sshcfg
       rspconfig ip=dhcp
       rspconfig get [<args>]...
       rspconfig set [<args>]...


Options:
  -l,--list           List are dump files
  -g,--generate       Trigger a new dump file
  -c,--clear <arg>    The id of file to clear or all if specify 'all'
  -d,--download <arg> The id of file to download or all if specify 'all'

The supported attributes to get are: %s

The supported attributes and its values to set are:
   ip=<ip address> netmask=<mask> gateway=<gateway> [vlan=<vlanid>]
   hostname=*|<string>
   autoreboot={0|1}
   powersupplyredundancy={enabled|disabled}
   powerrestorepolicy={always_on|always_off|restore}
```

### DIFF
--- a/xCAT-openbmc-py/lib/python/agent/hwctl/bmcconfig.py
+++ b/xCAT-openbmc-py/lib/python/agent/hwctl/bmcconfig.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python
+###############################################################################
+# IBM(c) 2018 EPL license http://www.eclipse.org/legal/epl-v10.html
+###############################################################################
+# -*- coding: utf-8 -*-
+#
+
+class BmcConfigInterface(object):
+    """Interface for bmc configuration related actions."""
+    interface_type = 'bmcconfig'
+    version = '1.0'
+
+    def dump_list(self, task):
+        return task.run('dump_list')
+
+    def dump_generate(self, task):
+        return task.run("dump_generate")
+
+    def dump_clear(self, task, id):
+        return task.run("dump_clear", id)
+
+    def dump_download(self, task, id):
+        return task.run("dump_download", id)
+
+    def dump_process(self, task):
+        return task.run("dump_process")
+
+    def set_sshcfg(self, task):
+        return task.run("set_sshcfg")
+
+    def set_ipdhcp(self, task):
+        return task.run("set_ipdhcp")
+
+    def get_attributes(self, task, attributes):
+        return task.run("get_attributes", attributes)
+
+    def set_attributes(self, task, attributes):
+        return task.run("set_attributes", attributes)
+
+class DefaultBmcConfigManager(BmcConfigInterface):
+    """Interface for BmcConfig actions."""
+    pass
--- a/xCAT-openbmc-py/lib/python/agent/hwctl/executor/openbmc_bmcconfig.py
+++ b/xCAT-openbmc-py/lib/python/agent/hwctl/executor/openbmc_bmcconfig.py
@@ -18,49 +18,6 @@ logger = logging.getLogger('xcatagent')
 RSPCONFIG_GET_NETINFO=['ip', 'netmask', 'gateway', 'vlan', 'ipsrc', 'hostname']
 RSPCONFIG_SET_NETINFO=['ip', 'netmask', 'gateway', 'vlan']
 
-RSPCONFIG_APIS = {
-    'autoreboot' : {
-        'baseurl': "/control/host0/auto_reboot/",
-        'set_url': "attr/AutoReboot",
-        'get_url': "attr/AutoReboot",
-        'display_name': "BMC AutoReboot",
-    },
-    'powersupplyredundancy':{
-        'baseurl': "/sensors/chassis/PowerSupplyRedundancy/",
-        'set_url': "/action/setValue",
-        'get_url': "/action/getValue",
-        'get_method': 'POST',
-        'get_data': '[]',
-        'display_name': "BMC PowerSupplyRedundancy",
-        'attr_values': {
-            'disabled': "Disables",
-            'enabled': "Enabled",
-        },
-    },
-    'powerrestorepolicy': {
-        'baseurl': "/control/host0/power_restore_policy/",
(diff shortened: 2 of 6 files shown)


whowutwut:
@zet809 Need to resolve some conflicts here.

xCAT-openbmc-py/lib/python/agent/hwctl/bmcconfig.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
+###############################################################################
+# IBM(c) 2018 EPL license http://www.eclipse.org/legal/epl-v10.html
+###############################################################################
+# -*- coding: utf-8 -*-
+#
+
+class BmcConfigInterface(object):
+    """Interface for power-related actions."""
```
robin2008: modify the comment bmc configuration relation actions.

xCAT-openbmc-py/lib/python/agent/hwctl/bmcconfig.py
```diff
@@ -0,0 +1,42 @@
...
+    interface_type = 'bmcconfig'
+    version = '1.0'
+
+    def bmcconfig_dump_list(self, task):
```
robin2008: better not using 'bmcconfig_*' here as they are already in the class namespace.

xCAT-openbmc-py/lib/python/agent/hwctl/executor/openbmc_bmcconfig.py
```diff
@@ -0,0 +1,125 @@
...
+import logging
+logger = logging.getLogger('xcatagent')
+
+RSPCONFIG_GET_NETINFO=['ip', 'netmask', 'gateway', 'vlan', 'ipsrc', 'hostname']
+RSPCONFIG_SET_NETINFO=['ip', 'netmask', 'gateway', 'vlan']
+
+class OpenBMCBmcConfigTask(ParallelNodesCommand):
+        
+    def bmcconfig_dump_list(self, **kw):
+        return self.callback.info('bmcconfig_dump_list')
```
robin2008: It is just a fake method, right?
Maybe give brief comments to demonstrate the expected logic

xCAT-openbmc-py/lib/python/agent/hwctl/executor/openbmc_bmcconfig.py
```diff
@@ -0,0 +1,125 @@
...
+    def _bmcconfig_set_powersupplyredundancy(self, powersupplyredundancy, **kw):
+        return self.callback.info("bmcconfig_set_powersupplyredundancy: %s" % powersupplyredundancy)
+
+    def _bmcconfig_get_powersupplyredundancy(self, **kw):
+        return self.callback.info("bmcconfig_get_powersupplyredundancy")
+
+    def _bmcconfig_set_autoreboot(self, autoreboot, **kw):
+        return self.callback.info("bmcconfig_set_autoreboot: %s" % autoreboot)
+
+    def _bmcconfig_get_autoreboot(self, **kw):
```
robin2008: For such general attributes,  better to define a map in openbmc_client.py to contains the path, payload suffix, etc. And using a general function to handle it. So it could be easy to implement other general attributes for getter/setter.

xCAT-openbmc-py/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -623,7 +666,60 @@ def rpower(self, nodesinfo, args):
             DefaultPowerManager().reboot(runner, optype=action)
         else:
             DefaultPowerManager().set_power_state(runner, power_state=action)
+    def rspconfig(self, nodesinfo, args):
+        try:
+            new_args=[]
+            for a in args:
+               new_a = a.encode('utf-8')
```
robin2008: we may need handle it in server.py

xCAT-openbmc-py/lib/python/agent/xcatagent/openbmc.py
```diff
@@ -75,6 +78,46 @@
     }
 }
 
+RSPCONFIG_GET_OPTIONS = ['ip','ipsrc','netmask','gateway','vlan','hostname','bootmode','autoreboot','powersupplyredundancy','powerrestorepolicy']
```
gurevichmark: Is there any way to dynamically generate RSPCONFIG_GET_OPTIONS and RSPCONFIG_SET_OPTIONS ? Ideally, once the commands are defined in RSPCONFIG_APIS structure in openbmc_client.py, it would be best if nothing else would need to be modified.

gurevichmark:
Need to think about what to do with man page. Since the usage for `rspconfig` is different  in python than it is in perl, should we list both usage formats in man page, or somehow dynamically change the man page when python version of the command is enabled ?

xuweibj:
Looks good to me. Merge it in first. We can modify any details for subcommand later.

zet809:
Hi, @gurevichmark , even through the python usage is different with perl interface, but from rspconfig command point of view, the only interface is perl, so for manpage of rspconfig, we don't need to modify anything. 
But if we provide a independent tool to call python directly, we can output a different Usage than perl.

xCAT-openbmc-py/lib/python/agent/xcatagent/openbmc.py
zet809: It is true we can, but a little tricky.
We have 3 layers, 1: OpenBMCManager, the cli parser, 2. DefaultBmcConfigManager: the abstracted interface, 3: OpenBMCBmcConfigTask: the logic of doing rspconfig, 4: OpenBMCRest: the rest call operation.

Issue #4742, opened 2018-01-26
Validate that xCAT can support PDU control when passwords are changed from the shipped defaults


whowutwut:
Opening this issue to make sure that xCAT PDU support can operate when default passwords are changed for the following PDUs. 

* CORAL compute rack PDU 
* Infrastructure rack PDU (IBM PDU Manager)

zet809:
hi, @whowutwut , shall we make sprint for this task? ^_^

whowutwut:
@cxhong ^ ?

cxhong:
should able to make this sprint.

whowutwut:
Closing out

PR #4801, opened 2018-02-07
Support user defined password from pdu table

cxhong:
For issue #4742

Support user defined authentication from PDU table for each PDUs. it included community string for snmpV1,  user name and password for login  and snmpV3 authentication.
````
]# tabdump pdu
#node,nodetype,pdutype,outlet,username,password,snmpversion,community,snmpuser,authtype,authkey,privtype,privkey,seclevel,comments,disable
"f6pdu15","pdu","irpdu","9","ADMIN","1001",,"public",,,,,,,,
"f6pdu16","pdu","irpdu","9",,,,"public",,,,,,,,
"coral-pdu","pdu","crpdu",,,,"3","public","admin","MD5","password1","DES","password2","authPriv",,````

### DIFF
--- a/xCAT-server/lib/xcat/plugins/pdu.pm
+++ b/xCAT-server/lib/xcat/plugins/pdu.pm
@@ -450,6 +450,7 @@ sub powerstat {
     foreach my $pdu (@$noderange) {
         if ($pduhash->{$pdu}->[0]->{pdutype} eq 'crpdu') {
             my $snmpversion = $pduhash->{$pdu}->[0]->{snmpversion};
+            my $snmpcmd;
             if ($snmpversion =~ /3/) {
                 my $snmpuser = $pduhash->{$pdu}->[0]->{snmpuser};
                 my $seclevel = $pduhash->{$pdu}->[0]->{seclevel};
@@ -469,7 +470,6 @@ sub powerstat {
                             $privkey=$authkey;
                         }
                     }
-                    my $snmpcmd;
                     if ($seclevel eq "authNoPriv") {
                         $snmpcmd = "snmpwalk -v3 -u $snmpuser -a $authtype -A $authkey -l $seclevel"; 
                     } elsif ($seclevel eq "authPriv") {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by robin2008** [comment]
@cxhong 

Could you help to describe the simple steps for how to test this PR?
So that others could take over the verification if possible.

**Comment by cxhong** [comment]
attributes in the pdu table:
````
community:    1) used for **pdudiscover** command, will check site.snmpc first, then default value
                        2) used for **irpdu** commands (rinv, rvitals and rpower) to connect PDU via snmp, will check pdu node attribute first, then default.
username/password:   1) used for all the crpdu command, check pdu node attributes first, then default
                                     2) used for irpdu rspconfig command, check pdu node attributes first, then default
snmpv3 attributes:  used for the crpdu rpower command.  check pdu node attributes first, then default
````

File: xCAT-server/lib/xcat/plugins/pdu.pm [hunk 1]
```diff
@@ -469,24 +470,26 @@ sub powerstat {
...
                     } else {   #default to notAuthNoPriv
                         $snmpcmd = "snmpwalk -v3 -u $snmpuser -l $seclevel"; 
                     }
-
-                    for (my $relay = 1; $relay <= 3; $relay++) {
-                        relaystat($pdu, $relay, $snmpcmd, $callback);
-                    }
+                } else {
+                    xCAT::SvrUtils::sendmsg("ERROR: No snmpuser or Security level defined for snmpV3 configuration", $callback,$pdu);
+                    xCAT::SvrUtils::sendmsg("    use chdef command to add pdu snmpV3 attributes to pdu table", $callback,$pdu);
```
zet809: Is it possible to give out a command example that how to add snmpV3 attribute with lsdef?

xCAT-server/lib/xcat/plugins/pdu.pm
```diff
@@ -469,24 +470,26 @@ sub powerstat {
...
+                    xCAT::SvrUtils::sendmsg("    then run 'rspconfig $pdu snmpcfg' command ", $callback,$pdu);
                     next;
                 }
-           } 
-            xCAT::SvrUtils::sendmsg("please config snmpv3 to be able to query pdu relay status", $callback,$pdu);
-            xCAT::SvrUtils::sendmsg("    use chdef command to add pdu snmpv3 attributes to pdu table", $callback);
-            xCAT::SvrUtils::sendmsg("    then run 'rspconfig $pdu snmpcfg' command ", $callback);
+            } else {
```
zet809: What happen if user set snmp v1/v2c information?
Shall use v3 with default username/password?

xCAT-server/lib/xcat/plugins/pdu.pm
```diff
@@ -773,10 +786,17 @@ sub process_sshcfg {
...
     my $static_ip   = shift;
     my $discover_ip   = shift;
+    my $userid = shift;
+    my $password = shift;
 
     #default password for coral pdu
-    my $password = "password8";
-    my $userid = "root";
+    if (!defined $userid) {
+        $userid = "root";
```
zet809: Can we put the default username/password at the front of the file for easier maintenance.?

zet809:
hi, @cxhong , an overall comment is, can we read table pdu before go into every function? Then we don't need to read pdu table for every command, right?

xCAT-server/lib/xcat/plugins/pdu.pm
cxhong: added

xCAT-server/lib/xcat/plugins/pdu.pm
cxhong: will use default snmpv3 , snmpv1/v2c will not work for this command

xCAT-server/lib/xcat/plugins/pdu.pm
cxhong: I don't think that's necessary.  Those are default values, only used one place in the code and will never make changes.

cxhong:
@zet809,  users should issue the command one at a time,  we get the info from pdu table depends which command and noderanges users specified.

zet809:
If we put the section to query 'pdu' table before go into every command, we can have an unique process to deal with username/password, ip, ....

zet809:

Issue #4743, opened 2018-01-27
setroute doesn't work for DHCP interfaces


kcgthb:
I'm trying to define a default route on my CentOS 7 nodes, on the install NIC, via the `setroute` postscript. It works when the script i s executed, but since the NIC are configured via DHCP, the route settings are ignored when the node reboots or the network is restarted.

For instance, I want to deploy `sh-101-60` with `10.10.0.62` as its default gateway, via `em1` (its installation NIC):
```
# lsdef -t route gw_ln02
Object name: gw_ln02
    gateway=10.10.0.62
    mask=0.0.0.0
    net=default
    usercomment=route all traffic through sh-ln02

# lsdef sh-101-60 -i routenames
Object name: sh-101-60
    routenames=gw_ln02

# lsdef sh-101-60 --nics
Object name: sh-101-60
    nicaliases.bmc=sh-101-60
    nicaliases.ib0=sh-101-60
    nicextraparams.em1=MTU=9000
    nicextraparams.ib0=MTU=65520 CONNECTED_MODE=yes
    nicips.bmc=10.11.101.60
    nicips.ib0=10.9.101.60
    nicnetworks.bmc=net_ipmi
    nicnetworks.ib0=net_ib_100
    nictypes.bmc=bmc
    nictypes.ib0=infiniband

# lsdef sh-101-60 -i postscripts
Object name: sh-101-60
    postscripts=syslog,remoteshell,syncfiles,otherpkgs,confignetwork,setroute replace
```

After deployment, `sh-101-60` is configured with `em1` (the install NIC) in DHCP, and the static route is correctly defined:
```
# cat /etc/sysconfig/network-scripts/ifcfg-em1
# Generated by parse-kickstart
DEVICE="em1"
IPV6INIT="yes"
BOOTPROTO="dhcp"
UUID="44756a91-8596-4eea-85dc-1d3dfd711304"
ONBOOT="yes"

# cat /etc/sysconfig/network-scripts/route-em1
default/0 via 10.10.0.62
```
But since the interface is configured for DHCP, the static route is ignored.

After a reboot:
```
mn # ssh sh-101-60 ip route \| grep default
default via 10.10.0.1 dev em1
```

`setroute` works when it's executed manually:
```
mn # updatenode sh-101-60 -P setroute
sh-101-60: xcatdsklspost: downloaded postscripts successfully
sh-101-60: Fri Jan 26 16:18:06 PST 2018 Running postscript: setroute
[...]

PR #4807, opened 2018-02-08
routeop: support remove GATEWAY from /etc/sysconfig/network

bybai:
refer to #4803, it add GATEWAY in /etc/sysconfig/network in redhat without configuring nic for default route. Since routeop also support delete default route, GATEWAY in /etc/sysconfig/network should be removed when executing delete option.

related to #4743 

UT:
```
bybc0605: + filename=/etc/sysconfig/network
bybc0605: + '[' default = default -a -f /etc/sysconfig/network ']'
bybc0605: + grep GATEWAY /etc/sysconfig/network
bybc0605: + '[' 0 -eq 0 ']'
bybc0605: + sed -i /GATEWAY=/d /etc/sysconfig/network
bybc0605: + echo 'Persistent GATEWAY=10.0.0.101 removed from /etc/sysconfig/network.'
bybc0605: + exit 0
```

### DIFF
--- a/xCAT/postscripts/routeop
+++ b/xCAT/postscripts/routeop
@@ -758,7 +758,7 @@ rm_persistent_route()
         fi
         filename="/etc/sysconfig/network";
         if [ "$net" = "default" -a -f $filename ]; then
-            grep "GATEWAY" $filename >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


immarvin:
i am ok with the change

Issue #4670, opened 2018-01-23
enhance rcons when goconserver is running


zet809:
* [x] Add env in cmd plugin of goconserver
* [x] Move the reconnecting logic from command script to goconserver
* [x] enhance rcons script for ipmi, openbmc, kvm... in case goconserver is running.

PR #4810, opened 2018-02-08
Move sleep retry from xcat to goconserver

chenglch:
If cons script is waiting, goconserver could not aware of
the status of the console connection. This patch check
the environment when running the script, if it is gocons
and retrying is needed, exit immediately.

implement: #4670

### DIFF
--- a/xCAT-server/share/xcat/cons/kvm
+++ b/xCAT-server/share/xcat/cons/kvm
@@ -1,5 +1,6 @@
 #!/usr/bin/env perl
 # IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
+use strict;
 use Fcntl qw(:DEFAULT :flock);
 
 sub get_lock {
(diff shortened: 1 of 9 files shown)


xCAT-server/share/xcat/cons/kvm
```diff
@@ -65,8 +65,7 @@ xCAT::Client::submit_request($cmdref, \&getans);
 until ($dsthost and $speed and $dstty) {
     release_lock();
     $sleepint = int(rand(30)) + 60;
-    print "Console not ready, retrying in $sleepint seconds (Ctrl-C to skip delay)\n";
-    exec "sleep $sleepint";
+    xCAT::Utils::console_sleep($sleepint, "Console not ready, retrying in $sleepint seconds (Ctrl-C to skip delay)\n");
```
zet809: Pls modify Ctrl-C to Ctrl-E c o

PR #4815, opened 2018-02-08
Allow choosing individual openbmc commands to use Python

gurevichmark:
Allow setting `XCAT_OPENBMC_PYTHON=` to be the list of commands to run in Python.
If `XCAT_OPENBMC_PYTHON=YES` all openbmc commands will run in Python
If `XCAT_OPENBMC_PYTHON=NO` or not set all openbmc commands will run in Perl
If `XCAT_OPENBMC_PYTHON=command,command,...` commands listed will run in Python, commands not listed will run in Perl.
If `XCAT_OPENBMC_PYTHON=EXCEPT:command,command,...` commands listed will run in Perl, commands not listed will run in Python.

### DIFF
--- a/xCAT-server/lib/perl/xCAT/OPENBMC.pm
+++ b/xCAT-server/lib/perl/xCAT/OPENBMC.pm
@@ -201,7 +201,13 @@ sub wait_agent {
 sub is_openbmc_python {
     my $environment = shift;
     $environment = shift if (($environment) && ($environment =~ /OPENBMC/));
-    # If XCAT_OPENBMC_PYTHON is YES, will run openbmc2.pm. If not, run openbmc.pm
+    # If XCAT_OPENBMC_PYTHON is YES, 
+    # will return "ALL" and caller will run openbmc2.pm. 
+    # If XCAT_OPENBMC_PYTHON is not set or is set to NO, return "NO" and caller
+    # will run openbmc.pm
+    # If XCAT_OPENBMC_PYTHON is a list of commands, return that list and caller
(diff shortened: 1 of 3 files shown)


whowutwut:
@gurevichmark Trying to see how we would use this in production... right now we have a smaller set of commands that are in Python, but the Python framework is faster in performance so this is the direction we want to take...  

I think we should have `XCAT_OPENBMC_PYTHON=YES` to control the use of python until we fully change over by default.  But we need a way to revert back to Perl if we find some bug during this transition process, but only for the problem commands. 

If we have 15 commands, and we find a single one has an issue, we have to export a variable export with 14 commands names to turn off the bad one, i think it's a little backwards.

gurevichmark:
How about this for completed flexibility:

* `XCAT_OPENBMC_PYTHON=YES` - all openbmc commands go through Python
* `XCAT_OPENBMC_PYTHON=NO` - all openbmc commands go through Perl
* `XCAT_OPENBMC_PYTHON=command,command,...` commands listed will go through Python, commands not listed will run in Perl.
* `XCAT_OPENBMC_PYTHON=EXCEPT:command,command,...` commands listed will go through Perl, commands not listed will run in Pyhon.

For example:
Setting `XCAT_OPENBMC_PYTHON=rpower,rspconfig` will make only `rpower` and `rspconfig` go through Python, all others through Perl.
Setting `XCAT_OPENBMC_PYTHON=EXCEPT:rflash,rinv` will make only `rflash` and `rinv` go through Perl, all others through Python.

gurevichmark:
New commit added to reverse listed command logic with "EXCEPT:" prefix

PR #4826, opened 2018-02-13
Add test cases for configuration  management first version

hu-weihua:
These test cases are used to test configuration management (first version).


bybai:
As discussed in scrum meeting, merge xcat inventory test case first, let xcat-test package all test cases. Thanks @hu-weihua

PR #4829, opened 2018-02-14
Modify command message for makeconservercf and makegocons

chenglch:
Make the command message more friendly.

parital-issue: https://github.com/xcat2/goconserver/issues/26


chenglch:
Close this PR at first until we make agreement about how to deal with the conserver, confluent and goconserver together.

PR #4834, opened 2018-02-20
Get xCAT MASTER from dhcpclient lease file for configinterface postscripts

cxhong:
the configinterface postscripts gets xcat master from mypostscripts, but sometime,  mypostscripts didn't contain MASTER.  The configinterface postscripts will failed if there are no xCAT MASTER defined.

### DIFF
--- a/xCAT/postscripts/configinterface
+++ b/xCAT/postscripts/configinterface
@@ -9,7 +9,7 @@ fi
 
 xcat_intf="/etc/network/interfaces.d/xCAT.intf"
 
-MASTER=$(cat /var/lib/dhcp/dhclient.eth0.leases |grep cumulus-provision-url|tail -1|awk -F/ '{print $3}')
+MASTER=$(cat /var/lib/dhcp/dhclient.eth0.leases|sed -n 's/.*cumulus-provision-url.*http:\/\+\([^\/]\+\)\/.*/\1/p'|tail -1)
 if [ -z "$MASTER" ]; then
     echo "xCAT Master unset! Cannot download interface description"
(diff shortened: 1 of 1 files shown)


cxhong:
the lease file contains 
```
option cumulus-provision-url "http://172.11.253.27/install/postscripts/cumulusztp";
  option dhcp-message-type 5;
  option domain-name-servers 172.10.253.27;
  option dhcp-server-identifier 172.11.253.27;
````
@immarvin , should we use dhcp-server-identfier instead of  cumulus-provision-url?

**Comment by immarvin** [comment]
hi @cxhong , `cumulus-provision-url` is a customized dhcp option(code 239) for ZTP mentioned in cumulus provision guide, I do not think it can be replaced with  `dhcp-server-identfier`

File: xCAT/postscripts/configinterface [hunk 1]
```diff
@@ -9,6 +9,7 @@ fi
 
 xcat_intf="/etc/network/interfaces.d/xCAT.intf"
 
+MASTER=$(cat /var/lib/dhcp/dhclient.eth0.leases |grep cumulus-provision-url|tail -1|awk -F/ '{print $3}')
```
immarvin: hi @cxhong , extract the xcatmaster for cumulus-provision-url is ok, but the command should better be `MASTER=$(cat /var/lib/dhcp/dhclient.eth0.leases|sed -n  's/.*cumulus-provision-url.*http:\/\+\([^\/]\+\)\/.*/\1/p')`

immarvin:
ok with the change

Issue #4811, opened 2018-02-08
not set domain in site table will cause dhcp service failed to start


zet809:
In my environment, I unset site.domain wrongly. After running `makedhcp -n`, there will be lines like below generated in dhcpd.conf file which will cause dhcp server failed to start.
And no warning message or error message output if *no 'domain' set* or *set as empty*
So, we need  output error message since the dhcp won't work.
```
    zone . {
       primary <MN_IP>; key xcat_key;
    }
```
The site.domain entry:
```
# lsdef -t site -i domain -c
#
```
The result of dhcpd:
```
# service isc-dhcp-server status
* isc-dhcp-server.service - ISC DHCP IPv4 server
   Loaded: loaded (/lib/systemd/system/isc-dhcp-server.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Thu 2018-02-08 01:18:12 EST; 50s ago
     Docs: man:dhcpd(8)
  Process: 20013 ExecStart=/bin/sh -ec      CONFIG_FILE=/etc/dhcp/dhcpd.conf;      if [ -f /etc/ltsp/dhcpd.conf ]; then CONFIG_FILE=/etc/ltsp/dhcpd.conf; fi;      [ -e
 Main PID: 20013 (code=exited, status=1/FAILURE)

Feb 08 01:18:12 c910f03c05k10 sh[20013]: Configuration file errors encountered -- exiting
Feb 08 01:18:12 c910f03c05k10 sh[20013]: If you think you have received this message due to a bug rather
Feb 08 01:18:12 c910f03c05k10 sh[20013]: than a configuration issue please read the section on submitting
...
```

PR #4839, opened 2018-02-22
fix not set domain in site table will cause dhcp service failed to start

bybai:
for #4811 

When there is no domain in site table, all networks entry should have domain configured, or else, dhcp service will be failed. So I modified the original warning as error and return code is 1.

UT covered the following scenarios:
1, refresh install xcat will not generate new dhcp configuration file, domain in site table has non-empty value. 
2, if `domain` is not set in `site` table, all networks entry have domains, `makedhcp -n` works, dhcp service works.
3, if `domain` is set in `site` table, there is no domain configured in `networks` table, `makedhcp -n` works, dhcp service works.
4, if `domain` is not set in `site` table, there are multiple networks entries in `networks` table, all or one networks entry has no domain configured, `makedhcp -n` return error message , will not generate new dhcp configuration file, return code is 1, example is as following.
```
[root@rhmn xCAT_plugin]# makedhcp -n
Renamed existing dhcp configuration file to  /etc/dhcp/dhcpd.conf.xcatbak

Error: No domain defined for 10.0.0.0 entry in networks table, and no domain defined in site table.
Warning: No dynamic range specified for 10.0.0.0. If hardware discovery is being used, a dynamic range is required.
Error: No domain defined for 30.0.0.0 entry in networks table, and no domain defined in site table.
Warning: No dynamic range specified for 30.0.0.0. If hardware discovery is being used, a dynamic range is required.
Error: above error fail to generate new dhcp configuration file, restore dhcp configuration file /etc/dhcp/dhcpd.conf
```
In cluster.log:
```
Feb 26 22:43:21 rhmn xcat[27022]: DEBUG dhcp: Renamed existing dhcp configuration file to  /etc/dhcp/dhcpd.conf.xcatbak
Feb 26 22:43:21 rhmn xcat[27022]: DEBUG dhcp: Restore dhcp configuration file to  /etc/dhcp/dhcpd.conf
```
5, when there is no `domain` in `site` table, `makedns` will return error message.
6, when there is no `domain` in `site` table, no domain in `networks` table, `makehosts` generate only short hostname for node.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -2441,7 +2441,7 @@ sub addnet
                 $callback->(
                     {
                         error => [
-                            "No $net specific entry for domain, and no domain defined in site table."
+                            "No domain defined for $net entry in networks table, and no domain defined in site table."
                           ],
                         errorcode => [1]
                     });
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -2440,9 +2440,10 @@ sub addnet
             } else {
                 $callback->(
                     {
-                        warning => [
+                        error => [
                             "No $net specific entry for domain, and no domain defined in site table."
-                          ]
+                          ],
+                        errorcode => [1]
                     });
             }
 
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/dhcp.pm
```diff
@@ -2440,9 +2440,10 @@ sub addnet
             } else {
                 $callback->(
                     {
-                        warning => [
+                        error => [
```
zet809: The error message is very strange, pls refine it.

zet809:

Issue #2962, opened 2017-04-28
Syncfiles on Cumulus switches


mattaezell:
Normally syncfiles are attached to an osimage and deployed to compute/login nodes, but it would be nice to be able to use syncfiles to place configuration on Cumulus switches.

immarvin:
hi @mattaezell , the running of Postscripts on cumulus deployment is not supported yet, which is traced in https://github.com/xcat2/xcat-core/issues/2662 and is a requisite to support syncing files during cumulus deployment.

Currently, you can run ``xdcp `` against the cumulus nodes to place configuration files on cumulus switches. please see http://xcat-docs.readthedocs.io/en/stable/advanced/networks/onie_switches/os_cumulus/install.html#configure-xcat-remote-commands for details

zet809:
Move to next release.

mattaezell:
Was this implemented as part of #2662, which was recently closed for the 2.13.6 release?

cxhong:
no,  not part of #2662.  will work on for next release

whowutwut:
@cxhong @immarvin is this still an open issue?  I want to close out 2.13.9 milestone ... so i'll move it to 2.13.10 for now.

zet809:
Hi, @cxhong and @immarvin , what is the plan for this issue?

mattaezell:
We found out today that we may need to modify `/etc/cumulus/datapath/traffic.conf` on our Cumulus switches.

immarvin:
hi @mattaezell , what do you expect to do with the config file `/etc/cumulus/datapath/traffic.conf` you mentioned? 

`updatenode -F` or thru ZTP?

immarvin:
hi @cxhong , would you please contact @mattaezell and take a look at this? thanks

mattaezell:
@immarvin @cxhong 
ZTP only works for initial deployment, but all of our switches are already deployed.  If implemented as a ZTP extension, we would consider reinstalling all of our switches.  That's our current plan for when we need to upgrade (since apt-get upgrade is potentially difficult).
`updatenode -F` is also acceptable (and the quickest way to get in on our switches), but that's not implemented today.

immarvin:
@mattaezell , ok, thanks for the info. we will first take a look at support sync files per node for `updatenode -F` and then ZTP.

cxhong:
[...]

PR #4841, opened 2018-02-22
Support synclist for cumulus switch

cxhong:
for issue #2962 

Pre-requirement for this PR:  
rsync command needs to be available on the cumulus switch.  Currently, we are request Mellanox/edgecore to pull into rsync package to the newer cumulus OS.  Meanwhile, user can get rsync package from cumulus repo:
http://repo3.cumulusnetworks.com/repo/pool/upstream/r/rsync/rsync_3.1.1-3_armel.deb

After rsync command installed on the cumulus switch,   user needs to set syncfiles to the **profile** attribute for the cumulus switch definition
```
#chdef mid08tor10 profile=/install/custom/switch/onie.syslist
# updatenode mid08tor10 -F
````

### DIFF
--- a/perl-xCAT/xCAT/CFMUtils.pm
+++ b/perl-xCAT/xCAT/CFMUtils.pm
@@ -340,6 +340,11 @@ sub updateCFMSynclistFile {
 
     foreach my $osimg (@osimgs)
     {
+
+        # this is for Cumulus switch 
+        if ($osimg eq "onie") {
+            next;
+        }
         my $cfmdir;
         $cfmdir = xCAT::CFMUtils->setCFMSynclistFile($osimg);
         if ($cfmdir)    # check for /install/osiamges/$osimg/cfmdir
--- a/perl-xCAT/xCAT/TableUtils.pm
+++ b/perl-xCAT/xCAT/TableUtils.pm
@@ -2000,10 +2000,18 @@ sub getimagenames()
     my @nodelist = @$nodes;
     my $nodetab  = xCAT::Table->new('nodetype');
     my $images =
-      $nodetab->getNodesAttribs(\@nodelist, [ 'node', 'provmethod', 'profile' ]);
+      $nodetab->getNodesAttribs(\@nodelist, [ 'node', 'provmethod', 'profile', 'nodetype' ]);
     my @imagenames;
     foreach my $node (@nodelist)
     {
+        #set imagename as "onie" if it is switch
+        #it used for cumulus switch
+        if ($images->{$node}->[0]->{nodetype} eq "switch" )
+        {
+            push  @imagenames, "onie";
+            next;
+        }
+        
         my $imgname;
         if ($images->{$node}->[0]->{provmethod})
         {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by zet809** [comment]
Hi, @cxhong , it seems we need to move this PR to 2.14

**Comment by whowutwut** [comment]
@gurevichmark @immarvin Are there any objection for these changes? 

@zet809  This would satisfy a customer requirement and I think we should consider this for 2.13.11...  maybe we just didn't have a chance to review?

**Comment by immarvin** [comment]
hi @whowutwut , I asked the cumulus support guys, their reply is:
```
Hi, best would be to contact your SE for that and ask him for this. Your could also setup a local mirror to solve this.
```
so I think maybe the installation binary is shipped by the switch manufacturer

On this PR, maybe the `profile` is not a proper  attribute to hold the synclist, we might need to add a new `synclist` attribute for node object

robin2008:
@cxhong 
why we use `profile` attribute here?  it will cause confusing as xcat is not using this attribute.

Is the requirement for per node sync list or just to support synclist for cumulus switch? why we continue to use `provmethod`, and use a special osimage for this kinds of onie switch.

cxhong:
@robin2008 ,  `profile` attribute is defined in the node definition but is not meaningful for switch, that's why I picked.  `provemethod` is used for cumulus OS bin file to install the cumulus switches.  I don't want to create a new attribute for synclist  since `synclists` is a attribute defined for the osimage.   it will causes more confusing to define synclists per node base (for compute node ).   
This is requirement to support synclist for cumulus switch only.

robin2008:
Yes, a new type of `osimage` is a good choice,  a new plugin might be required for `netboot=onie` to do the things for `nodeset` or `copycd`,  @immarvin will summarize the details and break down the tasks.

Do we need this before 20th April,  if not, maybe just consider the new design.

immarvin:
hi @whowutwut , the mini-design on this is created on https://github.com/xcat2/xcat-core/wiki/The-mini-design-of-syncfiles-for-cumulus , it will cost about 4 working days to finish, is it acceptable?

I will create a PR on item 3+4 for the following features in the mini-design by tomorrow

> add an option -r [the full path to scp or rsync] for updatenode to select scp/rsync, refine the code logic in updatenode and xdcp to enable this
> ```
> # updatenode mid05tor12 -F -r /usr/bin/scp
> # xdcp mid05tor12 -F -r /usr/bin/scp
> ```
> refine the code logic in startsyncfiles.awk , add an option -r to select the scp or rsync, the compute node can select the scp/rsync according to the existence of the 2 executables
> ```
> #xcat command: syncfiles -r   /usr/bin/scp
> ```


@cxhong , would you please take a look at the features mentioned in step #1 and #2 in the mini-design if convenient?

immarvin:
hi @whowutwut , for the question of @cxhong ,   `copycds` just copy the bin file to the relevant `pkgdir`directory with the path `/install/<osver>/<arch>/`, will not extract anything

immarvin:
covered by the task https://github.com/xcat2/xcat2-task-management/issues/59, close this ticket

PR #4843, opened 2018-02-23
Add installation steps for xcat-openbmc-py

zet809:
Add readme file to install xcat-openbmc-py


zet809:
No need this doc, so close.

PR #4845, opened 2018-02-25
fixes to default stanza handling

aaronknister:
At present if one has the word "default" in a stanza object name xCAT assumes it's a default stanza even if it's not. 

The cause that turned this up is storing the kernel name in the xCAT image name which with SLES includes "default" (e.g. 3.0.101-0.47.105-default). Cloning images or creating new ones using mkdef became problematic because xCAT would assume it to be a default stanza and produce the error: `Error: No object names were provided`. 

This PR makes the default object name more explicit so as not to catch object names that contain the word default but aren't default stanzas.

-Aaron

### DIFF
--- a/perl-xCAT/xCAT/DBobjUtils.pm
+++ b/perl-xCAT/xCAT/DBobjUtils.pm
@@ -1676,13 +1676,10 @@ sub readFileInput
 
             #  could have different default stanzas for different object types
 
-            if ($objectname =~ /default/) {
+            if ($objectname =~ /^default-([^-]+)$/) {
 
-                ($junk1, $objtype) = split(/-/, $objectname);
-
-                if ($objtype) {
-                    $objectname = 'default';
-                }
+                $objtype = $1;
+                $objectname = 'default';
 
                 next;
             }


perl-xCAT/xCAT/DBobjUtils.pm
```diff
@@ -1676,13 +1676,10 @@ sub readFileInput
 
             #  could have different default stanzas for different object types
 
-            if ($objectname =~ /default/) {
+            if ($objectname =~ /^([^-]+)-default$/) {
```
chenglch: I don't know the reason to change the format from `<default-type>` to `<type-default>`. Does `/^default-([^-]+)$/` work here?

aaronknister: That’s a good question. My testing suggested that the format of <default-type> didn’t work but <type-default> did. Looking at the code I’m confused about why my testing showed what it did. I’ll re-visit that and report back.

aaronknister: Oops, github stripped out my use of < > tags.

aaronknister: Yeah, I messed up. Not sure what I was thinking. I'll re-push the PR.

chenglch:
Has talked with @hu-weihua , if the error like below happens again, test case should be modified.
```
'RUN:echo -e "default-node:\\n    groups=all,compute\\ntestnode:\\n    objtype=node\\n    groups=all,rhels5.5\\n	xcatmaster= MS02.ppd.pok.com\\n       nfsserver=IS227.ppd.pok.com" > testfile [Thu Mar  1 18:42:23 2018]',
'ElapsedTime:0 sec',
'RETURN rc = 0',
'OUTPUT:',
'CHECK:rc == 0	[Pass]',
' ',
'RUN:cat testfile|mkdef -z [Thu Mar  1 18:42:23 2018]',
'ElapsedTime:1 sec',
'RETURN rc = 1',
'OUTPUT:',
'Error: The object name \'-e default-node\' is invalid, please refer to "man xcatdb" for the valid "xCAT Object Name Format"',
'CHECK:rc == 0	[Failed]',
```

Issue #4855, opened 2018-02-27
[OpenBMC] Python error message for bmcstate needs to match Perl - blocker.


whowutwut:
@zet809 We cannot GA 2.13.11 without this fixed. 

`rpower <node> bmcstate` needs to match the output of Perl when running in Python.  This was modified at the end of 2017 when I found problems identifying the root cause of the BMC not being Ready in a large cluster.   xCAT knows the return code and error but yet we masked it with a generic "Not Ready" state and required to add debug to really see what is going on with regards to the Rest API server.  

Here's the output in Perl code: 
```
[root@briggs01 ~]# lsdef -t site clustersite -i xcatdebugmode
Object name: clustersite
    xcatdebugmode=0
[root@briggs01 ~]# rpower mid05tor12cn05 bmcreboot
mid05tor12cn05: BMC reboot
[root@briggs01 ~]# while true; do  date; rpower mid05tor12cn05 bmcstate; sleep 5; done
Tue Feb 27 08:17:37 EST 2018
mid05tor12cn05: Error: [500] Login to BMC failed: 500 Can't connect to 172.11.139.5:443.
mid05tor12cn05: Error: BMC did not respond. Validate BMC configuration and retry the command.
Tue Feb 27 08:17:47 EST 2018
mid05tor12cn05: Error: [500] Login to BMC failed: 500 Can't connect to 172.11.139.5:443.
mid05tor12cn05: Error: BMC did not respond. Validate BMC configuration and retry the command.
Tue Feb 27 08:17:54 EST 2018
mid05tor12cn05: Error: [500] Login to BMC failed: 500 Can't connect to 172.11.139.5:443 (Connection refused).
mid05tor12cn05: Error: BMC did not respond. Validate BMC configuration and retry the command.
Tue Feb 27 08:17:59 EST 2018
mid05tor12cn05: Error: [500] Login to BMC failed: 500 Can't connect to 172.11.139.5:443 (timeout).
mid05tor12cn05: Error: BMC did not respond. Validate BMC configuration and retry the command.
Tue Feb 27 08:18:24 EST 2018
mid05tor12cn05: Error: [500] Login to BMC failed: 500 Can't connect to 172.11.139.5:443 (No route to host).
mid05tor12cn05: Error: BMC did not respond. Validate BMC configuration and retry the command.
Tue Feb 27 08:18:40 EST 2018
[...]

PR #4858, opened 2018-02-27
Add test case to compare the output of perl vs python for bmcstate when bmc reboots 

whowutwut:
Addresses the testcase_requested for #4855. 

After @gurevichmark completes the code change, will need to test out this again to see if it successfully runs for both Perl and Python, before we merge it in.

Here's a sample of running the test: [output.txt](https://github.com/xcat2/xcat-core/files/1764852/output.txt)

It finds some issues with the OpenBMC Python, return code...

### DIFF
--- a/xCAT-test/autotest/testcase/UT_openbmc/rpower_cases0
+++ b/xCAT-test/autotest/testcase/UT_openbmc/rpower_cases0
@@ -25,6 +25,10 @@ description: Record the output file into the xCAT test framework
 cmd: cat /tmp/xcattest.rpower.bmcstate.perl.out
 end
 
+start:rpower_bmcreboot_remove_tmpfile_perl 
+description: Remove the temporary file, output is captures above 
+cmd: rm -f /tmp/xcattest.rpower.bmcstate.perl.out
+end
 
 start:rpower_bmcreboot_record_python_output
 description: Start the test for python code
@@ -51,5 +55,10 @@ end
 start:rpower_bmcreboot_record_logs_python
 description: Record the output file into the xCAT test framework
 cmd: cat /tmp/xcattest.rpower.bmcstate.python.out
+end
+
+start:rpower_bmcreboot_remove_tmpfile_python
+description: Remove the temporary file, output is captures above 
+cmd: rm -f /tmp/xcattest.rpower.bmcstate.python.out
 end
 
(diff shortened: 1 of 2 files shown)


xCAT-test/autotest/testcase/UT_openbmc/rpower_cases0
```diff
@@ -0,0 +1,55 @@
...
+check:output=~1
+cmd: grep "BMC Ready" /tmp/xcattest.rpower.bmcstate.python.out | sort | uniq | wc -l
+check:output=~1
+end
+
+start:rpower_bmcreboot_record_logs_python
+description: Record the output file into the xCAT test framework
+cmd: cat /tmp/xcattest.rpower.bmcstate.python.out
+end
+
```
hu-weihua: @whowutwut, 
1  It seems above cases are not 6 test cases, they should be one case. 
2  It seems these cases will leave a temporary files under ``/tmp`` after running. Is it better to delete them ?

gurevichmark:
Code change for `rpower bmcstate` completed under #4866

xCAT-test/autotest/testcase/UT_openbmc/rpower_cases0
whowutwut: @hu-weihua I pushed a change for removing the file out /tmp.  Thank you.

What do you mean by the first comment?

hu-weihua: For FVT, one case should be complete independence. Setup environment, verify check point, restore environment should covered by one cases. Cases can be run in any sequence. So I think it is better to combine these 6 cases to one.  User need to run these 6 cases in specific sequence, otherwise will trigger error.

tingtli: @whowutwut @hu-weihua  I try to run this case in my environment. I found that the output of the bmcstate is always "Error: Login to BMC failed: Can't connect to 172.21.226.1 (Connection reset by peer". Seems the bmc is not up. So, there is no state change and the case failed.

tingtli: @whowutwut @hu-weihua  Since it's better that we put the 6 sections in one test case, I write a single case according to victor's original cases. To victor, would you like to open another PR or just past here and you modify your PR#. Please let me know. Thanks!

whowutwut: @tingtli  for the above comment, yes, if there is no state change, the testcase SHOULD fail.  

And to the 2nd comment above, if you can create a single test case, please go ahead.  Thank you!

tingtli:
Have a new one to address the test case for issue 4855. Close this one.

PR #4862, opened 2018-02-28
Fix the git short commit id to 7-byte long

neo954:
Since different git version generate different output format for `git rev-parse --short HEAD`. In this patch, fix the COMMIT_ID to 7-byte long.

### DIFF
--- a/build-ubunturepo
+++ b/build-ubunturepo
@@ -183,8 +183,8 @@ then
         short_short_ver=`cat Version|cut -d. -f 1`
         build_time=`date`
         build_machine=`hostname`
-        commit_id=`git rev-parse --short HEAD`
         commit_id_long=`git rev-parse HEAD`
+        commit_id="${commit_id_long:0:7}"
 
         package_dir_name=debs$REL
         #TODO: define the core path and tarball name
--- a/buildcore.sh
+++ b/buildcore.sh
@@ -188,8 +188,8 @@ function setversionvars {
     SHORTSHORTVER=`echo $VER|cut -d. -f 1`
     BUILD_TIME=`date`
     BUILD_MACHINE=`hostname`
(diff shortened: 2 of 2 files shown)


zet809:
Let's merge it and take a look at any improvement for the automation test result.

whowutwut:
@neo954 @zet809 What is issue that we saw that this fixes? 
Where do we reference the short ID?

neo954:
@vhu,

Since github.com abandoned TLSv1 support. We had to upgrade one of our build server to git version 2.14.2. After that, we observed run `git rev-parse --short HEAD` on this build server always generate 9-byte long hash instead of 7-byte long.

The short commit id hash is used in the `buildinfo` file and will be referred in the xCAT Jenkins Test Report. In the xCAT Jenkins Test Report generated on Wed, 28 Feb 2018 16:05:00 +0800, see screenshot below, xCAT Build version 2.13.11 commit `5cf796f` and commit `5cf796f6b` are listed as two builds. But actually, these two hash are for the same commit id. That is due to our Debian/Ubuntu build server has git version 1.9.1, which generate 7-byte long short commit id; and our Red Hat build server has git version 2.14.2, which generate 9-byte long short commit id.

screen shot 2018-03-02 at 14 35 12

Maybe, update the xCAT Jenkins Test Report to use the 40-byte long full-length commit id is better. But that will impact the performance of the `SELECT DISTINCT BuildCommitId FROM ...` operation in MariaDB when the mail report is generated. Clearly, it will turn slow. I did not estimate how much slower it will be, yet. Yes, there is index against the `BuildCommitId` column. So maybe the impact of performance is acceptable.

Anyway, I created this patch as a quick-and-dirty fix.

Thanks.

whowutwut:
Thanks for the explanation. Very helpful.

PR #4881, opened 2018-03-05
report clear error message when lacking python dependency.

robin2008:
UT:
```
rpower mid05tor12cn16 stat
Error: xCAT mgt=openbmc is using a Python based framework and there are some dependencies that are not met.
Error: python agent exited unexpectedly. See /var/log/xcat/agent.log for more details.
# pip install requests
Collecting requests
  Using cached requests-2.18.4-py2.py3-none-any.whl
Requirement already satisfied: certifi>=2017.4.17 in /usr/lib/python2.7/site-packages (from requests)
Requirement already satisfied: chardet<3.1.0,>=3.0.2 in /usr/lib/python2.7/site-packages (from requests)
Requirement already satisfied: idna<2.7,>=2.5 in /usr/lib/python2.7/site-packages (from requests)
Requirement already satisfied: urllib3<1.23,>=1.21.1 in /usr/lib/python2.7/site-packages (from requests)
Installing collected packages: requests
Successfully installed requests-2.18.4
# rpower mid05tor12cn16 stat
mid05tor12cn16: on
```

PR #4883, opened 2018-03-05
Fix the CI problem in github issue #4881

neo954:
Fix the problem by run the command line in question with `bash`.


hu-weihua:
thanks @neo954

PR #4905, opened 2018-03-08
[NOT MERGE before xcat-dep updated] Change the requires in spec file when we packaing some python module

robin2008:
- Import paramiko only when run rspconfig
- add the python-openbmc-dep requres in SPEC for ppc64le


zet809:
Let's close this PR since we don't plan to install xCAT-openbmc-py by default.

PR #4907, opened 2018-03-08
Handle the Python dependencies to xCAT-openbmc-py using EPEL/EXTRA/FC28 RPMs

whowutwut:
Having a difficult time resolving the issue of the Python dependencies.  

The list that we need is the following: 
`yum install python-gevent python2-docopt python-requests python-paramiko python-scp` 

Which can be satisfied by the following repos:
1. Local OS repo
2. EPEL repo 
3. RHEL 7 Extra packages REPO

But the version provided in those repos are OK except for  **gevent** and **greenlet**, we need higher level packages (python-gevent >= 1.2.2-2, python-greenlet >= 0.4.13-2)
 
Since we are support to GA 2.13.11 tomorrow, suggest that we push this action to the customer to download those two packages that are already there for Fedora28 rather than xCAT building a custom RPM dependency. 

This PR also removes `xCAT-openbmc-py` from automatically installing with xCAT 2.13.11.  So it will NOT default to Python, but the user who wants to run this would have to resolve these Python dependency packages to leverage the code. 

Will provide documentation to describe this in more details.


whowutwut:
2nd commit adds in some documentation for the install, URL provided for internal review http://c910loginx03.pok.stglabs.ibm.com/~vhu/python/references/coral/cluster_mgmt/scalability/index.html

@gurevichmark can you take a look at this doc?

whowutwut:
@robin2008 removed Mark as approver, we can still make doc updates at a later date, no problem.

zet809:
Hi, @whowutwut , as the document is for CORAL, do you think we shall add another page about how to install xCAT-openbmc-py and its dependencies in diskless SN?
Maybe after genimage for SN, chroot, and exec the manual step? Then, packimage?

whowutwut:
@zet809 What is the manual step?  

For the diskless image, they would add `xCAT-openbmc-py` to the package list, and the Requires from xCAT-openbmc-py should pull in the other packages.  

The user would have to add the epel, extras, and fedora28 custom repo to satisfy the requirements.  genimage will break without resolving the dependencies and this is what we want to happen to ensure we have the correct packages installed so xCAT Python code can work.

PR #4910, opened 2018-03-09
remove the dependency, Load SOAP module dynamically

robin2008:
UT:
```
chdef fake1 mgt=vbox
rpower fake1 stat
Error: SOAP::Lite perl module missing. Install perl-SOAP-Lite before running commands on Virtual Box nodes.
```


zet809:

PR #4914, opened 2018-03-09
add usercase for xcat-inventory

immarvin:
DOC preview http://10.3.5.21/xcat-doc/html/advanced/xcat-inventory/index.html


zet809:
Let's merge in case there is requirement to use this function and rethink how do we maintain the document about xcat-inventory

zet809:
Let's merge to have a document about xcat-inventory.
Maybe we shall move it out of xCAT since it is an independent tool.

PR #4925, opened 2018-03-12
conserver message change

gurevichmark:
Improve message when `makeconservercf` is issued while goconserver service is running, giving the user a hint that perhaps whey wanted to issue `makegocons` instead.

Before:
```
[root@briggs01 xcat]# makeconservercf mid05tor12cn02
Error: goconserver is started, please stop it at first.
[root@briggs01 xcat]#
```

After:
```
[root@briggs01 xcat]# makeconservercf mid05tor12cn02
Error: goconserver is being used as the console service, did you mean: makegocons <node>? If not, stop goconserver and retry.
[root@briggs01 xcat]#
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/conserver.pm
+++ b/xCAT-server/lib/xcat/plugins/conserver.pm
@@ -205,7 +205,7 @@ sub process_request {
         if (-x "/usr/bin/goconserver") {
             require xCAT::Goconserver;
             if (xCAT::Goconserver::is_goconserver_running()) {
-                my $rsp->{data}->[0] = "goconserver is started, please stop it at first.";
+                my $rsp->{data}->[0] = "goconserver is being used as the console service, did you mean: makegocons <noderange>? If not, stop goconserver and retry.";
                 xCAT::MsgUtils->message("E", $rsp, $cb);
                 return;
             }


whowutwut:
@gurevichmark Can you provide some more information in the description about the purpose of this, we had discussed but the community doesn't know 😄 

The main problem here is that as we switch to goconserver, and goconserver is running, if an admin that is familiar with conserver and their commands, we don't want them to blindly stop the goconserver service..

xCAT-server/lib/xcat/plugins/conserver.pm
```diff
@@ -205,7 +205,7 @@ sub process_request {
         if (-x "/usr/bin/goconserver") {
             require xCAT::Goconserver;
             if (xCAT::Goconserver::is_goconserver_running()) {
-                my $rsp->{data}->[0] = "goconserver is started, please stop it at first.";
+                my $rsp->{data}->[0] = "goconserver is running, did you mean: makegocons? If not, stop goconserver.";
```
whowutwut: Maybe change to:  `goconserver service is running, did you mean: makegocons?  If not, first stop goconserver and retry."

gurevichmark:
@whowutwut Message and PR description updated.

xCAT-server/lib/xcat/plugins/conserver.pm
```diff
@@ -205,7 +205,7 @@ sub process_request {
         if (-x "/usr/bin/goconserver") {
             require xCAT::Goconserver;
             if (xCAT::Goconserver::is_goconserver_running()) {
-                my $rsp->{data}->[0] = "goconserver is started, please stop it at first.";
+                my $rsp->{data}->[0] = "goconserver service is running, did you mean: makegocons? If not, stop goconserver and retry.";
```
whowutwut: Sorry for the flip flop here...   still think it's not too clear from the perspective of the user. We want to essentially say that goconserver is being used as the service and maybe from habit, they typed .`makeconservercf`...  

Should we say: 

```
Error: goconserver is being used as the console service, did you mean: makegocons <node>?  If not, stop goconserver and retry.
```

gurevichmark:
No problem. Message updated.

whowutwut:
@gurevichmark Are you able to specify the node as the input? 
And if the user runs `makegocons` generically, it should just show `makegocons` as the corresponding help usage...

gurevichmark:
Yes, it works when a node is specified:
```
[root@briggs01 xcat]# makeconservercf mid05tor12cn03
[root@briggs01 xcat]# echo $?
0
[root@briggs01 xcat]#
```
Also by default, if no noderange specified `makeconservercf` will create configuration for all nodes defined to xCAT. From man page:
```
EXAMPLES
       1. To create conserver configuration for all the nodes.
          makeconservercf

       2. To create conserver configuration for nodes node01-node10.
          makeconservercf node01-node10

       3. To remove conserver configuration for node01.
          makeconservercf -d node01

```

Similarly `makegocons`:
```
By default makegocons will register the session for all of the nodes in xcat.
```

gurevichmark:
At the time this message is issued, the <noderange> has been expanded, so if we display it, the list could be quite large.
Updated message to make it more generic.

Issue #3688, opened 2017-08-15
If multiple IP addresses used in xcatmaster field, updatenode will not work leaving node in bad state


whowutwut:
```
[root@briggs01 ~]# chdef mid05tor12cn11 xcatmaster="172.12.254.2,172.12.254.5" servicenode=sn02,sn05
1 object definitions have been created or modified.
[root@briggs01 ~]# updatenode mid05tor12cn11
xdsj There were no syncfiles defined to process. File synchronization has completed.                       h There were no syncfiles defined to process. File synchronization has completed.
Performing software maintenance operations. This could take a while, if there are packages to install.
```

It will continue to hang here....

If I remove the 2nd IP address and rerun update node, seems like postscripts are all gone. 

```
mid05tor12cn11: xcatdsklspost: downloaded postscripts successfully
mid05tor12cn11: find: '/xcatpost/': No such file or directory
mid05tor12cn11: chmod: missing operand after '+x'
mid05tor12cn11: Try 'chmod --help' for more information.
mid05tor12cn11: /tmp/fileKnD2rP.dsh: line 575: cd: //xcatpost: No such file or directory
mid05tor12cn11: xcatdsklspost: failed to download mypostscript.<nodename>, trying to call getpostscript.awk...
mid05tor12cn11: /tmp/fileKnD2rP.dsh: line 606: //xcatpost/getpostscript.awk: No such file or directory
mid05tor12cn11: /tmp/fileKnD2rP.dsh: line 606: //xcatpost/mypostscript: No such file or directory
mid05tor12cn11: grep: //xcatpost/mypostscript: No such file or directory
mid05tor12cn11: /tmp/fileKnD2rP.dsh: line 640: //xcatpost/getpostscript.awk: No such file or directory
mid05tor12cn11: /tmp/fileKnD2rP.dsh: line 640: //xcatpost/mypostscript: No such file or directory
mid05tor12cn11: grep: //xcatpost/mypostscript: No such file or directory
mid05tor12cn11: /tmp/fileKnD2rP.dsh: line 640: //xcatpost/getpostscript.awk: No such file or directory
mid05tor12cn11: /tmp/fileKnD2rP.dsh: line 640: //xcatpost/mypostscript: No such file or directory
^[[A^[[A^[[A^[[A^[[A^[[Amid05tor12cn11: grep: //xcatpost/mypostscript: No such file or directory
[...]

PR #4932, opened 2018-03-13
fix issue If multiple IP addresses used in xcatmaster field, updatenode will not work leaving node in bad state #3688

immarvin:
fix issue:
https://github.com/xcat2/xcat-core/issues/3688

UT:
```
[root@c910f03c05k21 xcat-core]# lsdef c910f03c17k42 -i xcatmaster
Object name: c910f03c17k42
    xcatmaster=c910f03c05k21,c910f03c05k27

[root@c910f03c05k21 xcat-core]# updatenode c910f03c17k42,c910f03c05k27
Error: Could not determine or resolve xcatmaster for c910f03c17k42. Will skip this node.
c910f03c05k27: bash: /usr/bin/rsync: No such file or directory
c910f03c05k27: rsync: connection unexpectedly closed (0 bytes received so far) [sender]
c910f03c05k27: rsync error: remote command not found (code 127) at io.c(605) [sender=3.0.9]
Performing software maintenance operations. This could take a while, if there are packages to install.

c910f03c05k27: vpdupdate is not supported on the pSeries KVM Guest platform
c910f03c05k27: updating VPD database
c910f03c05k27: trying to download postscripts...
c910f03c05k27: postscripts downloaded successfully
c910f03c05k27: trying to get mypostscript from 10.3.5.21...
c910f03c05k27: Tue Mar 13 04:40:09 EDT 2018 Running postscr
...
[root@c910f03c05k21 xcat-core]# echo $?
1
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/updatenode.pm
+++ b/xCAT-server/lib/xcat/plugins/updatenode.pm
@@ -306,6 +306,37 @@ sub preprocess_updatenode
         return;
     }
 
+
+    # get server names as known by the nodes
+    my %servernodes =
+      %{ xCAT::InstUtils->get_server_nodes($callback, $request->{node}) };
+
+    # it's possible that the nodes could have diff server names
+    # do all the nodes for a particular server at once
+
+    my @invalidnodes;
+    if($servernodes{undef}){
+        push @invalidnodes,@{$servernodes{undef}};
+    }
+
+    if ($servernodes{""}){
+        push @invalidnodes,@{$servernodes{""}};
+    }
+
+    if (@invalidnodes){
+        my %allnodes=map {$_,1} @{$request->{node}};
+        foreach my $node (@invalidnodes){
+           xCAT::MsgUtils->report_node_error($callback,$node,"Could not determine or resolve xcatmaster for $node. Will skip this node.")
+           delete $allnodes{$node};
+        }
+        $request->{node}=[];
+        push $request->{node}, map  $_ ,keys %allnodes;
+    }
+
+    unless (scalar @{$request->{node}}){
+        return;
+    }
+
     # preprocess generate mypostscripts files (-g) for hierarchy
     # if no sharedtftp then we need to broadcast this updatenode -g to all service nodes inorder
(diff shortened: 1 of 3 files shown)


xCAT-server/lib/xcat/plugins/updatenode.pm
```diff
@@ -306,6 +306,40 @@ sub preprocess_updatenode
         return;
     }
 
+
+    # get server names as known by the nodes
+    my %servernodes =
+      %{ xCAT::InstUtils->get_server_nodes($callback, $request->{node}) };
```
robin2008: We may need to ignore the nodes when xcatmaster is not defined as you run this on pre_,  it only be executed on MN, so for nodes under SN, they might not be has facing IP.

immarvin: updated

xCAT-server/lib/xcat/plugins/updatenode.pm
```diff
@@ -306,6 +306,40 @@ sub preprocess_updatenode
...
+        push @invalidnodes,@{$servernodes{""}};
+    }
+
+    if (@invalidnodes){
+        my %allnodes=map {$_,1} @{$request->{node}};
+        foreach my $node (@invalidnodes){
+           my $rsp = {};
+           $rsp->{errorcode}->[0]=1;
+           $rsp->{error}->[0] = "Could not determine or resolve xcatmaster for $node. Will skip this node. ";
+           $callback->($rsp);
```
robin2008: Could we use 'report_node_error' in MsgUtils.pm?  I think it is better to set the error to this node.

immarvin: updated

immarvin:
```
[root@c910f03c05k21 xcat-core]# updatenode c910f03c17k42
c910f03c17k42: Error: Could not determine or resolve xcatmaster for c910f03c17k42. Will skip this node.
[root@c910f03c05k21 xcat-core]# echo $?
1
```

PR #4941, opened 2018-03-13
Adding more OpenBMC Unit Testing test cases to quickly test success case

whowutwut:
The idea here is to quickly be able to run some sanity test against all the OpenBMC commands that xCAT supports leveraging xCAT-test but not getting hung up on the complexity of setting up the bundle and configuration file 

Using the shared scripts: 
```
[root@fs2vm112 autotest]# ./create_bundle_file  UT_openbmc supported
Filter of supported specified....
Creating bundle file for UT_openbmc...
cat: /opt/xcat/share/xcat/tools/autotest/testcase/UT_openbmc/scripts: Is a directory
supported_cmds_rpower
supported_cmds_rinv
-rw-r--r-- 1 root root 42 Mar 13 11:30 ./UT_openbmc.tmp.bundle
Run the test using:
   xcattest -f <config_file> -b ./UT_openbmc.tmp.bundle
   XCATTEST_CN=mid05tor12cn05  xcattest -b ./UT_openbmc.tmp.bundle
[root@fs2vm112 autotest]#
```

### DIFF
--- a/xCAT-test/autotest/testcase/UT_openbmc/rflash_cases0
+++ b/xCAT-test/autotest/testcase/UT_openbmc/rflash_cases0
@@ -70,3 +70,14 @@ cmd: rflash $$CN --delete 221d9020 221d9020
 check:rc==1 
 check:output=~Error: More than one firmware specified is not supported
 end
+
+start:rflash_delete_active_bmc
+description: Attempt to delete the active BMC firmware
+os:Linux
+hcp:openbmc
+cmd: rflash $$CN -l | grep \* | grep BMC | awk '{print $2}' | xargs -i{} rflash $$CN --delete {}
+check:rc==1
+check:output=~
+
+
+./xCAT-test/autotest/testcase/UT_openbmc/rflash_cases0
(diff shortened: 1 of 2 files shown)


gurevichmark:
Tested on briggs01. All tests passed.

whowutwut:
@gurevichmark Thanks for pushing that additional commits.  @hu-weihua The idea for this is to quickly be able to run a sanity test to make sure the OpenBMC functions are working. And if we encounter some issues in our internal cluster, we can leverage the xCAT-test framework to quickly run and see what commands are broken.

xCAT-test/autotest/testcase/UT_openbmc/rflash_cases0
```diff
@@ -70,3 +70,14 @@ cmd: rflash $$CN --delete 221d9020 221d9020
...
+start:rflash_delete_active_bmc
+description: Attempt to delete the active BMC firmware
+os:Linux
+hcp:openbmc
+cmd: rflash $$CN -l | grep \* | grep BMC | awk '{print $2}' | xargs -i{} rflash $$CN --delete {}
+check:rc==1
+check:output=~
+
+
+./xCAT-test/autotest/testcase/UT_openbmc/rflash_cases0
```
hu-weihua: It seems this case has not been finished? right?

whowutwut: @hu-weihua nice catch.. @gurevichmark could you finish this one?  
I think we only want to test BMC, does it prevent it automatically?  

If not, we could commit on top and get rid of this one...

hu-weihua:
Thanks @gurevichmark, I am ok for this pull request.

PR #4963, opened 2018-03-18
Refine test case  xcatconfig_c

hu-weihua:
``xcatconfig_c`` could not offer useful information when it failed in automation test system, so refine this case to offer more valid debug information.


bybai:
I am ok to merge it. Thanks @hu-weihua

Issue #4951, opened 2018-03-15
goconserver doesn't work in hierarchical mode


kcgthb:
`sh-hn01` is the MN, `sh-hn02` is a service node.

```
sh-hn01 # makegocons -q sh-101-61

NODE                             SERVER                           STATE
sh-101-61                        sh-hn02.SUNet                    avaiable
```
```
# rcons sh-101-61
Fatal error: EOF{"file":"github.com/xcat2/goconserver/console/client.go (296)","level":"error","msg":"Fatal error: Could not connect to sh-101-61\n","node":"sh-101-61","time":"2018-03-15T09:30:49-07:00"}
The connection is disconnected
[Enter `^Ec.' to exit]
Session is teminated unexpectedly, retrying....
Disconnected
```

`conserver` works fine in the exact same environment.

chenglch:
@kcgthb , I could not address where the error is in my environment. 
Is there any log about `sh-101-61` from `goconserver` on `sh-hn02.SUNet`?, by default it is in `/var/log/goconserver/server.log`.

neo954:
The problem was verified. I will close this one.

chenglch:
@kcgthb , the fix for this issue has been fixed in [xCAT 2.14](https://github.com/xcat2/xcat-core/wiki/XCAT_2.14_Release_Notes), if any problem, feel free to reopen this issue.

PR #4966, opened 2018-03-19
Fix command error in gocons caused by the env variable

chenglch:
In hierarchical mode, xCAT insert variables in the command line,
but goconserver do not support this format. This patch put the
definition for environment variable to `{params}->{env}` field so
that the first argument of  `{params}->{cmd}` is executable.

fix-bug: #4951


zet809:

PR #4990, opened 2018-03-22
NVIDIA CUDA driver installation document updating

neo954:
Update the CUDA installation document to CUDA 9.2 and RHEL 7.5


zet809:
Looks good, I agree to merge, thx @neo954

PR #4991, opened 2018-03-22
Accumulated go-xcat updates

neo954:
* Fix `vim` modelines
  * Fix `--long-help` outputs
  * Use `curl` when it is available. Otherwise fall back to use `wget`
  * Improved tarball file extension handling
  * Disable `xCAT-core.repo` and `xCAT-dep.repo` if they exist


neo954:
@immarvin, It is great if you have time and can review this.

Issue #4989, opened 2018-03-22
the error message of *def command is disordered when running internally


hu-weihua:
xcat version: 

```
[root@c910f02c01p13 ~]# lsdef -v
lsdef - Version 2.14 (git commit b0ea8ec79232994f200c387af2ba9b4d4dbfd687, built Wed Mar 21 06:15:49 EDT 2018)
```
The error message of ``bmcdiscover`` is disordered when it hit database problem.

```
[root@c910f02c01p13 ~]# bmcdiscover --range  10.4.30.254 -u USERID -p PASSW0RD -w
Writing  (10.4.30.254) to database...
Error: Command failed: chdef. Error message: No object names were provided.
Usage: chdef - Change xCAT data object definitions.
  chdef [-h | --help ] [-t object-types]
  chdef [-t object-types] [-o object-names] [-n new-name] [node]
  chdef [-V | --verbose] [-t object-types] [-o object-names] [-d | --dynamic]    [-z | --stanza] [-m | --minus] [-p | --plus]    [-w attr==val [-w attr=~val] ... ] [noderange] [attr=val [attr=val...]]

The following data object types are supported by xCAT.
auditlogboottargeteventlogfirmwaregroupkitkitcomponentkitrepomonitoringnetworknodenotificationosdistroosdistroupdateosimagepdupolicyrackroutesitetaskstatezonezvmivp
Use the '-h' option together with the '-t' option toget a list of valid attribute names for each object type..

Error: create or modify node is failed.

```
The correct format of chdef error msg looks like this:
```
# chdef -t node -o '' bmc=10.4.30.254 groups=all
Error: No object names were provided.

Usage: chdef - Change xCAT data object definitions.

  chdef [-h | --help ] [-t object-types]

  chdef [-t object-types] [-o object-names] [-n new-name] [node]

  chdef [-V | --verbose] [-t object-types] [-o object-names] [-d | --dynamic]
    [-z | --stanza] [-m | --minus] [-p | --plus]
    [-w attr==val [-w attr=~val] ... ] [noderange] [attr=val [attr=val...]]


The following data object types are supported by xCAT.

auditlog
boottarget
eventlog
firmware
group
kit
kitcomponent
kitrepo
monitoring
network
node
notification
osdistro
osdistroupdate
osimage
pdu
policy
rack
route
site
taskstate
zone
zvmivp

Use the '-h' option together with the '-t' option to
get a list of valid attribute names for each object type.
```

[...]

PR #4992, opened 2018-03-22
fix error 4989: The error message of bmcdiscover is disordered

zet809:
Fix issue #4989 

The root cause is failed to deal with this error message:
```
root@c910f03c05k10:~# bmcdiscover --range  10.4.30.254 -u USERID -p PASSW0RD -z
:
	objtype=node
	groups=all
	bmc=10.4.30.254
	cons=ipmi
	mgt=ipmi
```
With the fix, the output looks like:
```
root@c910f03c05k10:~# bmcdiscover --range  10.4.30.254 -u USERID -p PASSW0RD -z
Warning: Unable to establish IPMI v2 / RMCP+ session: insufficient resources for session for 10.4.30.254
```


xuweibj:
Agree to merge.

PR #4995, opened 2018-03-22
Change xCAT default log settings

neo954:
Avoid log compute node logs to other log files


whowutwut:
@neo954 Can you show the effects of this PR? What is the unwanted logs being generated today and what happens when that change is made to discard?

neo954:
How `rsyslogd` works is match each log entry with rules listed in its configuration file, by sequence, one by one. This change will stop the rule matching for remains rules if that log entry were decided to goes to `/var/log/xcat/computes.log`. That means, all log entries come from service nodes, and compute nodes will not goes to `/var/log/messages` or other log files any more.

Please also see my comments in https://github.ibm.com/vhu/c910env/issues/242

robin2008:
'~' will be replaced with `stop` in rsyslog new version, we can consider use it when not supporting for rhel6/sles11.

Now it is okay.

Issue #4999, opened 2018-03-23
[OpenBMC] No output for rpower bmcstate when username is wrong


xuweibj:
```
# chdef f6u03 bmcusername=roota
1 object definitions have been created or modified.
# rpower f6u03 bmcstate
Thu Mar 22 21:45:08 2018 f6u03: [openbmc_debug] login curl -k -c cjar -b cjar -X POST -H "Content-Type: application/json" https://10.6.3.100/login -d '{"data": ["roota", "xxxxxx"]}'
Thu Mar 22 21:45:08 2018 f6u03: [openbmc_debug] login [401] Invalid username or password
```

PR #5000, opened 2018-03-23
show error msg for rpower bmcstate when username is wrong

xuweibj:
#4999 

```
# rpower f6u03 bmcstate
f6u03: Error: [401] Invalid username or password
```


zet809:
Looks good, agree to merge, thx!

Issue #4961, opened 2018-03-16
Switching to goconvers doesn't modify conserver startup configuration. goconserver doesn't start on reboot. 


whowutwut:
When we switch over to `goconserver` the `sysVinit` still contains the configuration for conserver asking it to auto start when the server boot up...
```
[root@boston02 consoles]# chkconfig --list | grep conserver

Note: This output shows SysV services only and does not include native
      systemd services. SysV configuration data might be overridden by native
      systemd configuration.

      If you want to list systemd services use 'systemctl list-unit-files'.
      To see services enabled on particular target use
      'systemctl list-dependencies [target]'.

conserver      	0:off	1:off	2:on	3:on	4:on	5:on	6:off
```

So in the case that we switch to goconserver, it doesn't modify any start up scripts

```
[root@boston02 ~]# cd /var/log/consoles
[root@boston02 consoles]# rm -rf *
[root@boston02 consoles]# service conserver stop
Stopping conserver (via systemctl):                        [  OK  ]
[root@boston02 consoles]# service goconserver start
Redirecting to /bin/systemctl start goconserver.service
[root@boston02 consoles]#
[root@boston02 consoles]# ls -ltr
total 12
-rw------- 1 root root 55 Mar 17 14:55 sn02.log
-rw------- 1 root root 55 Mar 17 14:55 mid08tor03cn01.log
-rw------- 1 root root 55 Mar 17 14:55 mid08tor03cn10.log
```

Now let's reboot this server.... 

After reboot, you can tell right away the conserver has taken over again, because of the log files. 

```
[root@boston02 consoles]# ls -ltr
total 24
-rw------- 1 root root   55 Mar 17 14:55 sn02.log
-rw------- 1 root root   55 Mar 17 14:55 mid08tor03cn01.log
-rw------- 1 root root   55 Mar 17 14:55 mid08tor03cn10.log
-rw-r--r-- 1 root root 1239 Mar 17 15:01 sn02
-rw-r--r-- 1 root root 1182 Mar 17 15:01 mid08tor03cn01
-rw-r--r-- 1 root root 1182 Mar 17 15:01 mid08tor03cn10
[root@boston02 consoles]#
```

whowutwut:
Some options that I can think about.. @chenglch  I think we need to consider some changes here to make this work on reboot.. 

1. As part of the `makegocons` we should call `chkconfig conserver off` ... 
but then when user runs `makeconservercf` we have to put it back

[...]

PR #5001, opened 2018-03-23
Modify startup configuration for goconserver and conserver

chenglch:
After running makegocons, startup configuration would be changed
to make sure goconserver could start after rebooting. Changes
are made in makeconservercf too.

Fix-issue: #4961


zet809:
Looks good to me, I agree to merge, thx @chenglch

PR #5002, opened 2018-03-23
Add a test case for rvitals outputs format checking

neo954:
Make sure each line of the outputs of command `rvitals` matches regular expression `^[^:]+:[^:]+:[^:]+$`. Otherwise, this test case will fail.


hu-weihua:

Issue #4918, opened 2018-03-09
Create all lower case repo files within xCAT product 


whowutwut:
`go-xcat` started created lowercase repo config files for xCAT:
  * xcat-core.repo 
  * xcat-dep.repo 

But when you download the tar.bz2 files , they create with camelcase... 
   * xCAT-core.repo 
   * xCAT-dep.repo 

I run `go-xcat` to set up the xcat-core.repo and xcat-dep.repo to point at the devel version, but I don't install it: 
```
[root@fs2vm111 ~]#        /tmp/go-xcat --xcat-version devel install
Operating system:   linux
Architecture:       ppc64le
Linux Distribution: rhel
Version:            7.4

Reading repositories ...... done

xCAT is going to be installed.
Continue? [y/n] n
```

So I get these files: 
```
[root@fs2vm111 ~]# ls -ltr /etc/yum.repos.d/
total 24
-rw-r--r--. 1 root root 214 Mar  2 10:24 local-repository-0.repo
-rw-r--r--  1 root root 358 Mar  2 10:26 redhat.repo
-rw-r--r--  1 root root 135 Mar  2 10:26 xCAT-rhels7.4-snap6-path0.repo
-rw-r--r--  1 root root 150 Mar  2 10:26 xCAT-otherpkgs0.repo
-rw-------  1 root root 204 Mar  2 12:50 xcat-core.repo
-rw-------  1 root root 118 Mar  2 12:50 xcat-dep.repo
```

Then taking a build and extracting and making `./mklocalrepo.sh` I get the following files: 
```
[root@fs2vm111 xcat-core]# ls -ltr /etc/yum.repos.d/
total 28
-rw-r--r--. 1 root root 214 Mar  2 10:24 local-repository-0.repo
-rw-r--r--  1 root root 358 Mar  2 10:26 redhat.repo
-rw-r--r--  1 root root 135 Mar  2 10:26 xCAT-rhels7.4-snap6-path0.repo
-rw-r--r--  1 root root 150 Mar  2 10:26 xCAT-otherpkgs0.repo
-rw-------  1 root root 204 Mar  2 12:50 xcat-core.repo
-rw-------  1 root root 118 Mar  2 12:50 xcat-dep.repo
-rw-r--r--  1 root root 146 Mar  2 12:51 xCAT-core.repo <-------
```

This is inconsistent... then we also have 2 locations for xCAT that rpms could be pulling from

robin2008:
[...]

PR #5010, opened 2018-03-26
Use lowercase file name xcat-core.repo instead of xCAT-core.repo

neo954:
This pull request is for issue #4918

### DIFF
--- a/buildcore.sh
+++ b/buildcore.sh
@@ -500,8 +500,8 @@ if [ "$OSNAME" != "AIX" ]; then
     # Always recreate it, in case the whole dir was copied from devel to 2.x
     if [ -n "$1" ]; then embed="$1/"
     else embed=""; fi
-    cat >xCAT-core.repo << EOF
-[xcat-2-core]
+    cat >xcat-core.repo << EOF
+[xcat-core]
 name=xCAT 2 Core packages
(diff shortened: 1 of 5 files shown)


robin2008:
@neo954 
I have below comments:
- where to handle the repo file for xcat-dep ?
- For `go-xcat`, we need to disable `xCAT-*.repo` if it found those old repo file, for example, rename it to `xCAT-*.repo.disable`
- In `mklocalrepo.sh`, it is required to do the same thing.

BTW:  Better for you  to have an self-build core and dep tarball, and use it for UT on below senario
1, fresh installaion
2, upgrade

neo954:
@robin2008,

The patch for `go-xcat` is in pull request #4991.

whowutwut:
@neo954 I think we also need to modify in our GSA directory for the xcat-dep package builds, because I believe those scripts are creating xCAT-dep.repo files...   Not sure if anyone changed those..  and re-spin a development version of the xcat-dep package.

neo954:
The xcat-dep GSA directory was updated manually.
  * All file named `xCAT-dep.repo` were renamed to `xcat-dep.repo`
  * File `mklocalrepo.sh` was tweaked to rename `xCAT-dep.repo` if it exists.

robin2008:
remember to modify xcat.org accordingly.

PR #4484, opened 2017-12-08
updatenode -F not work in hierachy env as the user name is FQDN of MN

robin2008:
updatenode -F not work in hierachy env (#4455)
when run `updatenode -F`, we will get the user name in request and set it to `xdcp` via `DSH_FROM_USERID`.

UT:
1, updatenode -F from MN
```
# updatenode c910f03c09k05 -F -V
Running command on c910f03c09k03.pok.stglabs.ibm.com: ip -4 --oneline addr show |awk -F ' ' '{print $4}'|awk -F '/' '{print $1}' 2>&1

Running command on c910f03c09k03.pok.stglabs.ibm.com: chmod -R a+r /install/postscripts 2>&1

  c910f03c09k03.pok.stglabs.ibm.com: Internal call command: xdcp c910f03c09k05 --nodestatus -F /tmp/sync.list
Running internal xCAT command: xdcp ...

Running command on c910f03c09k03.pok.stglabs.ibm.com: ip -4 --oneline addr show |awk -F ' ' '{print $4}'|awk -F '/' '{print $1}' 2>&1

Running command on c910f03c09k03.pok.stglabs.ibm.com: rm /tmp/xdcpsynclist.30297 2>&1
File synchronization has completed for nodes.
```
2, updatenode -F from SN
```
# updatenode c910f03c09k05 -F -V
Running command on c910f03c09k04.pok.stglabs.ibm.com: ip -4 --oneline addr show |awk -F ' ' '{print $4}'|awk -F '/' '{print $1}' 2>&1

Running command on c910f03c09k04.pok.stglabs.ibm.com: chmod -R a+r /install/postscripts 2>&1

  c910f03c09k04.pok.stglabs.ibm.com: Internal call command: xdcp c910f03c09k05 --nodestatus -F /tmp/sync.list
Running internal xCAT command: xdcp ...

Running command on c910f03c09k04.pok.stglabs.ibm.com: ip -4 --oneline addr show |awk -F ' ' '{print $4}'|awk -F '/' '{print $1}' 2>&1

File synchronization has completed for nodes.
```
3, updatenode -F from MN for the node not behind SN
```
# chdef c910f03c09k05 servicenode=
1 object definitions have been created or modified.

# updatenode c910f03c09k05 -F -V
Running command on c910f03c09k03.pok.stglabs.ibm.com: ip -4 --oneline addr show |awk -F ' ' '{print $4}'|awk -F '/' '{print $1}' 2>&1

[...]

PR #5013, opened 2018-03-26
Properly fix updatenode syncfiles

jjohnson42:
https://github.com/xcat2/xcat-core/pull/4484 went to fix updatenode, but in the process introduced a security exposure.

This peels back 'forceroot' and replaces it with using the trusted approach in xcatd sub validate function.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/updatenode.pm
+++ b/xCAT-server/lib/xcat/plugins/updatenode.pm
@@ -1833,6 +1833,7 @@ sub updatenodesyncfiles
                 {
                     command => ["xdcp"],
                     node    => $syncfile_node{$synclist},
+                    username => $request->{username},
                     arg     => $args,
                     env     => $env
                 },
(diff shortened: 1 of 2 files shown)


robin2008:
@jjohnson42 
we have tried that, but it is not working for normal user
```
Running command on c910f03c05k20: rm /tmp/xdcpsynclist.4195 2>&1
c910f03c05k24: /bin/mkdir: cannot create directory '/var/xcat/syncfiles/opt': Permission denied
c910f03c05k24: /bin/mkdir: cannot create directory '/var/xcat/syncfiles/etc': Permission denied
c910f03c05k24: /bin/mkdir: cannot create directory '/var/xcat/syncfiles/tmp': Permission denied
```
Now updatenode only has a global working directory for storing the temp files on SN or CN, maybe we need it per user.

jjohnson42:
Ok, I see that, but is it not the case that any user can make root-owned files using xdcp the way it is at the moment?

robin2008:
Yes, it will be an security issue. I will create another PR for using user owned directory for normal user using `updatenode -F`

PR #5017, opened 2018-03-27
(#4973) add journal logs in xcatsnap for xcatd

### LINKED ISSUE #4973: [DEV]Add systemd journal for xcatd to xcatsnap
Now, most of the main OS has systemd,  and When rsyslog is down,  we could not get some valuable  information from log files.

As xcatd is started by systemd, It is better to retrieve the information from Query the systemd journal

robin2008:
To fix #4973 
UT: just run xcatsnap and check if the content in tarball

```
-rw-r--r-- root/root        23 2018-03-27 04:19 ./commands_output/cat__etc_issue.out
-rw-r--r-- root/root       238 2018-03-27 04:19 ./commands_output/lsxcatd_a.out
-rw-r--r-- root/root      1012 2018-03-27 04:19 ./commands_output/cat__proc_meminfo.out
-rw-r--r-- root/root      1007 2018-03-27 04:19 ./commands_output/cat__proc_cpuinfo.out
-rw-r--r-- root/root    104446 2018-03-27 04:19 ./commands_output/journalctl_b_nopager_u_xcatd.out         <<  =====  here it is
```

### DIFF
--- a/xCAT-server/sbin/xcatsnap
+++ b/xCAT-server/sbin/xcatsnap
@@ -239,7 +239,7 @@ sub snap_it {
             "ls $installdir",            "/usr/bin/crontab -l",
             "find /tftpboot -size -32k", "ls -lR $xcatroot",
             "arp -a", "ps -edlf", "ps -aux", "ulimit -a", "df -k",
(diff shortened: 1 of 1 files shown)


bybai:
I am OK to merge it.

Issue #4926, opened 2018-03-12
Do not use PNOR in docs or messages. Use "host" instead


gurevichmark:
PNOR refers to the technology used for flash memory used to store firmware on the host.
PNOR is not the same as "host".

bybai:
Hi @gurevichmark , "PNOR" appeared in openbmc_common.rst, you can check this doc too. Thanks.

PR #5018, opened 2018-03-27
Replace usage of PNOR with Host

gurevichmark:
#4926 

1. Replace usage of PNOR with Host
2. Sync up `.rst` files with `.pod`


docs/source/references/coral/cluster_mgmt/firmware/ipmi.rst
```diff
@@ -10,7 +10,7 @@ Collect the required files
 Collect the following files and put them into a directory on the Management Node. 
 
    * pUpdate
-   * pnor
```
whowutwut: Is this understandable from the supermicro point of view, what do they name the pnor image file?

gurevichmark:
Updated to make it a little more clear.

whowutwut:
@zet809 how do you feel about this change?

zet809:
Good to me, thx very much @gurevichmark and @whowutwut , let me merge it.

Issue #5023, opened 2018-03-28
exlist has minor typo error and redundant entries


marksmit-ibm:
Provided exlist: /opt/xcat/share/xcat/netboot/rh/compute.rhels7.ppc64.exlist
1. ./usr/share/man*   this line appears twice
2. ./lib/kdb/keymaps/include*  this line seems to have a typo error:  'kdb' should be 'kbd'
3. the following first line, should make the 2nd line not required:
./usr/share/doc*
./usr/share/doc/packages/cyrus-sasl/doc*

Reference doc:  http://xcat-docs.readthedocs.io/en/latest/guides/admin-guides/manage_clusters/ppc64le/diskless/customize_image/trim_diskless_rootimg.html?highlight=exlist

lsxcatd -a
Version 2.14  built Tue Mar 27 06:15:43 EDT 2018

marksmit-ibm:
It occurs to me that issue #3 may be there on purpose to help if the user wants to easily add back in, like this (just a guess on my part)
+ ./usr/share/doc/packages/cyrus-sasl/doc*

immarvin:
hi @bybai ,would you please take a look at this?

bybai:
Hi @marksmit-ibm , thanks your suggestions, will look into it.

immarvin:
hi @marksmit-ibm , will the duplicate entries break anything?

marksmit-ibm:
As long as a user sees both entries and chooses to remove, or comment out, both entries when they _do_ want man pages, then they would get the expected outcome.  I'm just thinking we should remove one for usability.  I have looked at the list many times and only when I did a grep/compare of line by line with another exlist, did I see the duplicate.

PR #5036, opened 2018-03-30
enhance exlist without typo and duplicate lines

bybai:
For #5023 

UT:
```
1, use updated exlist in osimage rhels7.5-ppc64le-netboot-compute

2, genimage rhels7.5-ppc64le-netboot-compute

3, packimage rhels7.5-ppc64le-netboot-compute

4, gunzip rootimg.cpio.gz; cpio -idmv < rootimg.cpio

5, check results:
tmp]# ls usr/share/man
ls: cannot access usr/share/man: No such file or directory
tmp]# ls usr/share/doc
ls: cannot access usr/share/doc: No such file or directory
tmp]# ls lib/kbd/keymaps/include
ls: cannot access lib/kbd/keymaps/include: No such file or directory
 ```


immarvin:
agree to merge

PR #5046, opened 2018-04-03
fix a bug for go-xcat test cases

tingtli:
fix a bug for go-xcat test cases


hu-weihua:

Issue #5060, opened 2018-04-04
xcatprobe switch_macmap should report an error when can not access the switch


gurevichmark:
When `xcatprobe switch_macmap` can not access the switch (for example, when file is missing from the switch), it displays:
```
[root@briggs01 openbmc]# xcatprobe switch_macmap mid05tor12
Switch      Port(MTU)        MAC address(VLAN)           Node
--------------------------------------------------------------------------------------
mid05tor12  swp1(1500)       N/A                         mid05tor12cn01
mid05tor12  swp2(1500)       N/A                         mid05tor12cn02
mid05tor12  swp3(1500)       N/A                         mid05tor12cn03
mid05tor12  swp4(1500)       N/A                         mid05tor12cn04
mid05tor12  swp5(1500)       N/A                         mid05tor12cn05
mid05tor12  swp6(1500)       N/A                         mid05tor12cn06
mid05tor12  swp7(1500)       N/A                         mid05tor12cn07
mid05tor12  swp8(1500)       N/A                         mid05tor12cn08
mid05tor12  swp9(1500)       N/A                         mid05tor12cn09,frame5u20
mid05tor12  swp10(1500)      N/A                         mid05tor12cn10
mid05tor12  swp11(1500)      N/A                         mid05tor12cn11
mid05tor12  swp12(1500)      N/A                         mid05tor12cn12
mid05tor12  swp13(1500)      N/A
mid05tor12  swp14(1500)      N/A                         mid05tor12cn14
mid05tor12  swp15(1500)      N/A                         mid05tor12cn15
mid05tor12  swp16(1500)      N/A                         mid05tor12cn16
mid05tor12  swp17(1500)      N/A                         mid05tor12cn17
mid05tor12  swp18(1500)      N/A                         mid05tor12cn18
```

The `xcatprobe switch_macmap` should report an error when can not access the switch.
```
[root@briggs01 openbmc]# /opt/xcat/share/xcat/scripts/configonie --switches mid05tor12 --ssh
Host 172.11.205.12 not found in /root/.ssh/known_hosts
ssh configured for mid05tor12

[root@briggs01 openbmc]# xcatprobe switch_macmap mid05tor12
Switch      Port(MTU)        MAC address(VLAN)           Node
--------------------------------------------------------------------------------------
mid05tor12  swp1(1500)       N/A                         mid05tor12cn01
mid05tor12  swp2(1500)       N/A                         mid05tor12cn02
[...]

PR #5064, opened 2018-04-05
Print out error msg for xcatprobe switch_macmap command if switch passwordless is not enabled

cxhong:
For issue #5060 

For the cumulus switch, xCAT get Mac table from ssh command instead of SNMP commands.  If switch passwordless is not enabled.  the command will failed and mac address fields from `xcatprobe switch_macmap` will be empty.   Instead of no mac address return to user,  we will log a error message.

```
# xcatprobe switch_macmap mid08tor03
mid08tor03 - Failed to get MAC table. Make sure passwordless SSH to the switch is enabled.                        [FAIL]
```

### DIFF
--- a/perl-xCAT/xCAT/MacMap.pm
+++ b/perl-xCAT/xCAT/MacMap.pm
@@ -356,7 +356,7 @@ sub dump_mac_info {
                 $ret{$switch}->{ErrorStr} = $self->{macinfo}->{$switch}->{ErrorStr};
 
                 # To show the error message that the username/password related error is for SNMP only
-                if ($ret{$switch}->{ErrorStr} =~ /user\s*name|password/i) {
+                if ($ret{$switch}->{ErrorStr} =~ /user\s*name|password$/i) {
                     $ret{$switch}->{ErrorStr} .= " through SNMP";
                 }
             } else {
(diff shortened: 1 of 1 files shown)


whowutwut:
@immarvin can you review this one?

perl-xCAT/xCAT/MacMap.pm
```diff
@@ -728,7 +728,12 @@ sub refresh_switch {
 
         my @res=xCAT::Utils->runcmd("ssh -o StrictHostKeyChecking=no -o PasswordAuthentication=no $switch 'bridge fdb show|grep -i -v permanent|tr A-Z a-z  2>/dev/null' 2>/dev/null",-1);
         if ($::RUNCMD_RC) {
-            xCAT::MsgUtils->message("S", "Failed to get mac table with ssh to $switch, fall back to snmp! To obtain mac table with ssh, please make sure the passwordless root ssh to $switch is available");
+            my $errmsg = "Failed to get MAC table from $switch. Make sure passwordless SSH to the switch is enabled.";
+            xCAT::MsgUtils->message("S", $errmsg);
```
gurevichmark: Looking at the description of the `MsgUtils->message()` it seems the "S" just logs to syslog. It seems "SE" is suggested there if log to syslog and error display is needed. Maybe "S" by itself does not set RC=1 ? OR maybe `MsgUtils->error_message()` is even better ? 
```
S - Message will be logged to syslog ( severe error)
					 syslog  facily (local4) and priority (err) will be used.
					 See /etc/syslog.conf file for the destination of the
					 messages.
                     Note S can be combined with other flags for example
					 SE logs message to syslog to also display the
					 message by daemon/ Client.pm.
```
```

**Comment by immarvin** [comment]
hi @cxhong ,  I remember the policy for cumulus switch here is:  **try to obtain mac table via `ssh` first.  if success, return result; otherwise, fall back to the snmp method.**  , so we may need to find out whether the code logic fall back to snmp branch? why the output of `xcatprobe switch_macmap` is blank?  I do not think we should exit with error

**Comment by cxhong** [comment]
@immarvin , the Cumulus switch didn't provide mac address table from it's mibs file.  that's why we didn't use snmp command instead of we used ssh to switches to run the command to get certain Mac address output.   In this case, snmp is configured.

**Comment by cxhong** [comment]
Actually,  we can get mac address using different mib OID via SNMP.   it's the vlan information we can't get via snmp.  The OID for vlan only showed PVID (default is 1)

File: perl-xCAT/xCAT/MacMap.pm (hunk 1 above)
Comment (cxhong) [reply perl-xCAT/xCAT/MacMap.pm]: tried `MsgUtils->error_message() `, didn't get message from command line

**Comment by immarvin** [comment]
ok.  the vlan information is mandatory for `xcatprobe switch_macmap` but not needed for refreshing macmap hash during switch based discovery.
So we'd better keep the code logic for refreshing macmap during switch discovery in `refresh_switch()` untouched, only prompt error for `xcatprobe switch_macmap` case.

File: perl-xCAT/xCAT/MacMap.pm [hunk 2]
```diff
@@ -728,7 +728,12 @@ sub refresh_switch {
...
+            if ($self->{collect_mac_info}) {
```
immarvin: all the modification should be placed in this branch. please leave the message and code logic untouched outside the conditional branch

cxhong:
@immarvin thanks, code modified, please review it again.

immarvin:
ok to merge

Issue #4766, opened 2018-01-31
[FVT]:rflash <noderange> <tar_file_path> -a could ONLY upload the tar file to the node but could not activate the firmware in python version.


tingtli:
rflash <noderange> <tar_file_path> -a could ONLY upload the tar file to the node but could not active the firmware in python version.

Here is the output of python version
```
[root@briggs01 firm]# rflash mid05tor12cn05 witherspoon.pnor.squashfs.tar -a -V
[briggs01]: Attempting to upload /home/ltt/firm/witherspoon.pnor.squashfs.tar, please wait...
[briggs01]: mid05tor12cn05: Uploading /home/ltt/firm/witherspoon.pnor.squashfs.tar ...
[root@briggs01 firm]# rflash mid05tor12cn05 -l
mid05tor12cn05: ID       Purpose State      Version
mid05tor12cn05: -------------------------------------------------------
mid05tor12cn05: f57b7b1b BMC     Active(*)  ibm-v2.0-0-r26-0-g7285b52
mid05tor12cn05: a419ff24 BMC     Active(+)  ibm-v2.0-0-r26.1-0-gfb7714a
mid05tor12cn05: 1829bcbd Host    Ready      IBM-witherspoon-ibm-OP9_v1.19_1.119
mid05tor12cn05: df4b8673 Host    Active(*)  IBM-witherspoon-ibm-OP9_v1.19_1.90
```

Here is the output of perl version
```
[root@briggs01 firm]# rflash mid05tor12cn05 witherspoon.pnor.squashfs.tar -a -V
[briggs01]: Attempting to upload /home/ltt/firm/witherspoon.pnor.squashfs.tar, please wait...
mid05tor12cn05: [briggs01]: Uploading /home/ltt/firm/witherspoon.pnor.squashfs.tar ...
mid05tor12cn05: [briggs01]: Could not find ID for firmware IBM-witherspoon-ibm-OP9_v1.19_1.119 to activate, waiting 10 seconds and retry...
mid05tor12cn05: [briggs01]: Firmware upload successful. Attempting to activate firmware: IBM-witherspoon-ibm-OP9_v1.19_1.119 (ID: 1829bcbd)
mid05tor12cn05: [briggs01]: rflash IBM-witherspoon-ibm-OP9_v1.19_1.119 started, please wait...
mid05tor12cn05: [briggs01]: Activating Firmware IBM-witherspoon-ibm-OP9_v1.19_1.119...  10%
[...]

PR #5068, opened 2018-04-09
modify rflash args for docopt parse

xuweibj:
1. #4766  Modify args format in openbmc2.pm, make that the args received by python is "-x [xxx] [-V|--verbose]"
2. Get "cwd" as task's member variable
3. use ``os.path.join`` to get full path
4. output for rspconfig -c
before modify:
```
TAR BMC Firmware Product Version: ibm-v2.0-0-r41-0-gd0c319e
```
after modify:
```
TAR xyz.openbmc_project.Software.Version.VersionPurpose.BMC Firmware Product Version: ibm-v2.0-0-r41-0-gd0c319e
```
5. catch "DocoptExit" error, and print out msg.

### DIFF
--- a/xCAT-openbmc-py/lib/python/agent/common/task.py
+++ b/xCAT-openbmc-py/lib/python/agent/common/task.py
@@ -54,6 +54,7 @@ class ParallelNodesCommand(BaseCommand):
         """
         self.inventory = inventory
         self.callback = callback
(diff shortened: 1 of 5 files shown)


zet809:
Looks good to me, I agree to merge. thx @xuweibj

PR #5074, opened 2018-04-10
Change CUDA 9.2 installation document and example setup script for POWER 9

neo954:
* Document updating, typeset tweaks
* Example script updating

### DIFF
--- a/docs/source/advanced/gpu/nvidia/osimage/rhels.rst
+++ b/docs/source/advanced/gpu/nvidia/osimage/rhels.rst
@@ -189,24 +189,24 @@ xCAT includes a script, ``cuda_power9_setup`` as example, to help user handle th
 Diskful osimage
 ^^^^^^^^^^^^^^^
 
-For diskful deployment, there is no need to change the osimage definition. Instead, add this postscript to your compute node postbootscrtips list.
+For diskful deployment, there is no need to change the osimage definition. Instead, add this postscript to your compute node postbootscrtips list. ::
 
     chdef p9compute -p postbootscripts=cuda_power9_setup
 
 Disless osimage
 ^^^^^^^^^^^^^^^
 
-For diskless deployment, the script need to add to the postinstall script of the osimage. And it should be run in the chroot environment. Please refer the following commands as an example.
+For diskless deployment, the script need to add to the postinstall script of the osimage. And it should be run in the chroot environment. Please refer the following commands as an example. ::
 
-    mkdir -p /install/custom/netboot
-    cp /opt/xcat/share/xcat/netboot/rh/compute.rhels7.ppc64le.postinstall /opt/xcat/share/xcat/netboot/rh/compute.rhels7.ppc64le.postinstall
+    mkdir -p /install/custom/netboot/rh
+    cp /opt/xcat/share/xcat/netboot/rh/compute.rhels7.ppc64le.postinstall /install/custom/netboot/rh/cudafull.rhels7.ppc64le.postinstall
 
-    cat >>/install/custom/netboot/rh/cudafull.rhels7.ppc64le.postinstall <-EOF
+    cat >>/install/custom/netboot/rh/cudafull.rhels7.ppc64le.postinstall <<-EOF
 
-    cp /install/postscripts/cuda_power9_setup /install/netboot/rhels7.5/ppc64le/compute/rootimg/tmp/cuda_power9_setup"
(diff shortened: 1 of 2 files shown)


xCAT/postscripts/cuda_power9_setup
```diff
@@ -40,3 +40,33 @@ chmod 0644 /usr/lib/systemd/system/nvidia-persistenced.service
...
+depends() {
+    return 0
+}
+
+installkernel() {
+    return 0
+}
+
+install() {
+    [ -d \$initdir/etc/modprobe.d/ ] || mkdir \$initdir/etc/modprobe.d
```
cxhong: where is $initdir from?

cxhong:
The nvidia.conf will not able to create from current cuda_power9_setup scripts.
I removed `$initdir` from cuda_power9_setup postscripts.  Followed your documented and verified on diskless image
````
[root@mid08tor03cn10 ~]# cat /etc/modprobe.d/nvidia.conf
options nvidia NVreg_EnableStreamMemOPs=1 NVreg_RegistryDwords="PeerMappingOverride=1"
[root@mid08tor03cn10 ~]# ls -ltr /usr/lib/dracut/modules.d/95nvidia/
total 64
-rwxr-xr-x 1 root root 355 Apr 10 14:18 module-setup.sh
[root@mid08tor03cn10 ~]# cat /etc/dracut.conf.d/nvidia_patch.conf
dracutmodules+="nvidia"
[root@mid08tor03cn10 ~]# grep PeerMapping /proc/driver/nvidia/params
[root@mid08tor03cn10 ~]# grep NVreg /proc/driver/nvidia/params
````
I didn't see new params updated.

**Comment by cxhong** [comment]
diskfull installation didn't work neither.

````
]# lsdef mid08tor03cn01 -i postbootscripts
Object name: mid08tor03cn01
    postbootscripts=otherpkgs,cuda_power9_setup
````

````
[root@mid08tor03cn01 ~]# ls -ltr /usr/lib/dracut/modules.d/95nvidia/
total 4
-rwxr-xr-x 1 root root 355 Apr 10 15:03 module-setup.sh
[root@mid08tor03cn01 ~]# cat /etc/dracut.conf.d/nvidia_patch.conf
dracutmodules+="nvidia"
[root@mid08tor03cn01 ~]# ls -ltr /etc/modprobe.d/nvidia.conf
ls: cannot access /etc/modprobe.d/nvidia.conf: No such file or directory
````

File: xCAT/postscripts/cuda_power9_setup (hunk 1 above)
Comment (whowutwut) [reply xCAT/postscripts/cuda_power9_setup]: This one is part of dracut...   but I don't think we need the `\` before this, it's using heredoc

Comment (robin2008) [reply xCAT/postscripts/cuda_power9_setup]: `$initdir` is required, it is dractu variable,  And yes,  `\` should be removed.

Comment (neo954) [reply xCAT/postscripts/cuda_power9_setup]: These backslashes are needed. Since this section is included in here document, between line 48 and 69. Without backslashes, the bash variable `$initdir` will be parsed in the here document context, and before the content will be written to file `module-setup.sh`.

Thus, backslashes before `$` are needed. Otherwise it is not working

**Comment by neo954** [comment]
@cxhong, Do not remove those backlashes, otherwise it is not working.

**Comment by neo954** [comment]
Hello @vhu, @cxhong and @robin2008,

I add a new patch to cover the problem I met on diskfull installation.  Since xCAT recommend user to put CUDA repository in `ospkgdir` instead of `otherpkgdir`, the `nvidia.ko` is packed into the initrd image in diskfull compute node installation as well.

A new hack was added, it will redo `mkinitrd` for diskfull installation. And it works.

File: docs/source/advanced/gpu/nvidia/osimage/rhels.rst [hunk 2]
```diff
@@ -189,24 +189,24 @@ xCAT includes a script, ``cuda_power9_setup`` as example, to help user handle th
 Diskful osimage
 ^^^^^^^^^^^^^^^
 
-For diskful deployment, there is no need to change the osimage definition. Instead, add this postscript to your compute node postbootscrtips list.
+For diskful deployment, there is no need to change the osimage definition. Instead, add this postscript to your compute node postbootscrtips list. ::
 
-    chdef p9compute -p postbootscripts=cuda_power9_setup
+    chdef p9compute -p postscripts=cuda_power9_setup
 
 Disless osimage
```
cxhong: should be `Diskless osimage`

Issue #4929, opened 2018-03-13
[fvt] xcatd failed to start in rhels7.5-snap2 diskless service node 


hu-weihua:
xCAT build : ``Version 2.13.11 (git commit d0ed517ee20e271e05ba6ffd06f4e3deb0c240c0, built Fri Mar  9 07:15:44 EST 2018)``

MN OS:   ``Red Hat Enterprise Linux Server release 7.4 (Maipo)``
SN OS:    ``rhels7.5-snap2-ppc64le``

For xCAT 20180309 build, after installing diskless SN, the xcatd in SN failed to start.

This issue also happened on ``rhels7.4-alt-rc1`` xCAT 20180308 build, no matter with ``xCAT-openbmc-py`` or without ``xCAT-openbmc-py``.

```
[root@sn02 ~]# xcatd -f
Subroutine unsupported redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 102.
Subroutine handled_commands redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 120.
Subroutine preprocess_request redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 721.
Subroutine retry_after redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 837.
Subroutine retry_check_times redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 853.
Subroutine process_request redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 895.
Subroutine parse_args redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 1120.
Subroutine parse_command_status redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 1422.
Subroutine fork_process_login redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 2117.
Subroutine get_functional_software_ids redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 2148.
Subroutine parse_node_info redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 2175.
Subroutine gen_send_request redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 2256.
Subroutine deal_with_response redefined at /opt/xcat/lib/perl/xCAT_plugin/openbmc.pm line 2329.
[...]

PR #5086, opened 2018-04-11
Fix the issue for stateless node in debug mode, /dev/log is not available

robin2008:
(#4929) - use imjournal to make rsyslogd work well with journald

UT:
1, patch xcat
2, genimage for a stateless osimage
3, set xcatdebugmode=1 in site
4, rinstall or nodeset the node with new osimage

After reprovisioned,  we can see that /dev/log is ready to use.
```
# cat /proc/cmdline
imgurl=http://172.20.254.2:80//install/ansible/custom/rhels7.5-alt-rc2-ppc64le-ANSIBLE-netboot-compute/rootimgdir/rootimg.cpio.gz XCAT=172.20.254.2:3001 NODE=mid08tor03cn10 FC=0  LOGSERVER=172.20.254.2  syslog.server=172.20.254.2 syslog.type=rsyslogd syslog.filter=*.*  xcatdebugmode=1 BOOTIF=**:**:**:**:**:**  selinux=0

#python -c "import logging;from logging.handlers import SysLogHandler;SysLogHandler(address='/dev/log', facility=SysLogHandler.LOG_LOCAL4)"
# echo $?
0
```

### DIFF
--- a/xCAT-server/share/xcat/netboot/rh/dracut_033/patch/syslog/module-setup.sh
+++ b/xCAT-server/share/xcat/netboot/rh/dracut_033/patch/syslog/module-setup.sh
@@ -16,7 +16,7 @@ install() {
     local _installs
     if type -P rsyslogd >/dev/null; then
         _installs="rsyslogd"
-        inst_libdir_file rsyslog/lmnet.so rsyslog/imklog.so rsyslog/imuxsock.so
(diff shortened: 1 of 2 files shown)


robin2008:
Note:  this PR is only for Redhat.  If we found the same issue on SLES12 and Ubuntu, need such modification too.

PR #5090, opened 2018-04-11
fix linux_diskless_kdump failure

bybai:
For https://github.com/xcat2/xcat2-task-management/issues/47
Fix 2 issues:
1, linux_diskless_kdump test case failure
2, deduce `sleep 900` to `sleep 300` in all diskless and statelite test cases.


hu-weihua:
It is ok for me, thanks @bybai

Issue #5080, opened 2018-04-10
Changing user password is ignored by REST API


gurevichmark:
Once the non-root user password is changed, REST API does not behave as expected:

1. Create a non-root user as described by http://xcat-docs.readthedocs.io/en/latest/advanced/restapi/restapi_setup/restapi_setup.html#use-non-root-account
2. Verify REST API works with userid/pw and token authentication
3. Change the user password with `passwd <non-root user>`

**Problems:**
1. Can still obtain an authentication token with **old** password and this token can be used for REST API calls.
2. Can not obtain an authentication token with **new** password, returns `Authentication failure` error.
3. Using **old** or **new** password returns `Authentication failure` error when using userid and password REST API format.

robin2008:
@gurevichmark 
It seems you have wrong understanding about the restapi user password, I think xCAT restapi only use the password in table `passwd`.
```
# add password to passwd table
tabch key=xcat,username=<wsuser> passwd.password=<wspw>
```

@bybai  Could you confirm it,   if the document is not clear.  we can update it,  I did not see how to change password there.

bybai:
Hi @gurevichmark and @gurevichmark 
The xCAT REST API used username and password from `passwd` table. After you change the password for the user, you should update it in `passwd` table, it is the same with `root` user. 
The  xCAT REST API token has no relationship to user's password. After the token is generated, it can be used until expire time. 

Here is the doc about how to "Configure passwords in xCAT MN" : http://xcat-docs.readthedocs.io/en/latest/guides/admin-guides/manage_clusters/ppc64le/configure/password.html?highlight=passwd

daniceexi:
@robin2008 The username is valid for Restapi access, see the usage example: 
~~~
curl -X POST -k 'https://127.0.0.1/xcatws/tokens?userName=root&userPW=cluster&pretty=1'
~~~

bybai:
User may be easy to confuse "os system user/passwd" with "xCAT REST API user/passwd", here are sections in doc for "add root/non-root account", we do not emphasize how to change REST API password. So we can add simple section "Change Password for REST API account" behind "Use non-root Account" section in doc [Set Up an Account for Web Service Access](http://xcat-docs.readthedocs.io/en/latest/advanced/restapi/restapi_setup/restapi_setup.html#use-non-root-account). The content is to emphasize if user want to change the password for REST API, he should update the new password in `passwd` table.

gurevichmark:
Changed documentation with #5095

PR #5095, opened 2018-04-11
Password change for xCAT REST API documentation update

gurevichmark:
For #5080 
Update the doc on what to do when password is changed.

### DIFF
--- a/docs/source/advanced/restapi/restapi_setup/restapi_setup.rst
+++ b/docs/source/advanced/restapi/restapi_setup/restapi_setup.rst
@@ -154,3 +154,5 @@ You should see some output that includes your list of nodes.
 
 If errors returned, check `/var/log/httpd/ssl_error_log` on xCAT MN.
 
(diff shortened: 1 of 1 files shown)


docs/source/advanced/restapi/restapi_setup/restapi_setup.rst
```diff
@@ -154,3 +154,5 @@ You should see some output that includes your list of nodes.
 
 If errors returned, check `/var/log/httpd/ssl_error_log` on xCAT MN.
 
+``Note:`` if passwords need to be changed in the future, make sure to update the xCAT passwd table. xCAT REST API uses passwords stored in that table to authenticate users.
```
bybai: I think it is better to add a new section, that will be clearer for user to find the method,  for example:

Change Password
------------------
If passwords need to be changed in the future, update the xCAT `passwd` table. xCAT REST API uses passwords stored in this table to authenticate users. 
```
     tabch key=xcat,username=<wsuser> passwd.password=<new_wspw>
```

bybai:
I am ok to merge it. Thanks @gurevichmark

PR #5096, opened 2018-04-11
ddns: log more infornation to show the DNS entry when an error occurs

kcgthb:
Display the entry that caused the error to happen, for easier debugging.


bybai:
Will merge this pull request next week, thanks @kcgthb

PR #5104, opened 2018-04-13
update a light status in test case rbeacon_stat

tingtli:
update a light status in test case rbeacon_stat


hu-weihua:

PR #5106, opened 2018-04-13
Adding new options overrides for nvidia

whowutwut:
Update the sample postscript with the current overrides that are being used.


neo954:
The patch looks good. I will merge it.

Issue #5082, opened 2018-04-10
Add attribute to site table to allow non-default REST API authentication token expiration


gurevichmark:
Currently REST API authentication token expires in 24 hours.
Add a new attribute to `site` table to allow the user to specify how may days or `never` for token expiration.
Default will still stay at 24 hours.

gurevichmark:
Code for managing tokens is in xCAT-server/lib/perl/xCAT/xcatd.pm gettoken().

robin2008:
I am not sure it is good to given a `never` expire token for all user,  it would be potential security risk.

I would suggest there are two steps to get an long-lived token.
1, get a short lived token first  (this is for all user)
2, Provide a new API to change its to expire time, and this API could be controlled by `policy` table, to assign it for only `service` account.

For normal user, they will still get a short-lived token.

bybai:
I think after we change the expire time format as human readable format, it is easy for user to customize the expire time to a long time as they expect. It is no need to support `never` expire token in our code considering security risk. 
Now, user can change expire time in `token` table using `tabedit` or `tabch` command, if we can provide REST API interface to change expire time, user can use or integrate the REST API to update the token expire time as they expect.

gurevichmark:
Had a meeting with Fernando and Victor today. They felt for CORAL and CSM use, it would be useful to have a ability to modify token expiration time by just changing `site` table entry. This way it can be changed for the whole cluster and all tokens created after that point will have new expiration time. They also wanted ability to have a token that `never` expires, similar to how GitHub manages tokens.

gurevichmark:
This issue will also be used to add "creation" time of the token and "last" access time of the token to the `token` table.
All times will be stored in human readable format.

PR #5116, opened 2018-04-17
Token expiration setting

gurevichmark:
#5082 
1. Allow token expiration setting in `site` table
2. Add token creation and token access times to the `token` table
3. Store token expiration, creation and access times in human readable format in `token` table
4. If "old" style tokens with unix DataTime format encountered, remove if expired, convert to human readable format if not expired

```
[root@briggs01 xcat]# tabdump token
#tokenid,username,expire,created,access,comments,disable
"4b3786e7-745a-41c1-ad27-0b4c4779fa84","wsuser","2018-4-23 14:24:33","2018-4-18 14:24:33",,,
"46a29de5-2dc3-4a12-b837-8098e64af50a","wsuser","never","2018-4-18 14:29:37",,,
"89a647ad-d428-441a-85d3-ed649e0229ef","wsuser","2018-4-19 14:30:11","2018-4-18 14:30:11",,,
[root@briggs01 xcat]#
```


docs/source/advanced/restapi/restapi_usage/restapi_usage.rst
```diff
@@ -58,7 +58,15 @@ Then in the subsequent REST API access, the token can be used to replace the use
 
     curl -X GET -k -H X-Auth-Token:5cabd675-bc2e-4318-b1d6-831fd1f32f97 'https://<FQDN of xCAT MN>/xcatws/<resource>?<parameters>
 
-The validity of token is 24 hours. If an old token has expired, you will get a 'Authentication failure' error. Then you need reacquire a token with your account.
+The default validity of a token is 1 day. This default can be changed by the setting of `expiretokendays` attribute in `site` table. :: 
+
+    chdef -t site clustersite expiretokendays=<days>
```
whowutwut: Would `tokenexpiredays` be more fitting? seems like my first thought is to have `token` be the start of the word...  this way if we have additional variables, it would start with `token` to group them together with a common starting word.....

docs/source/guides/admin-guides/references/man5/site.5.rst
```diff
@@ -429,6 +429,8 @@ site Attributes:
    --------------------
   XCAT DAEMON ATTRIBUTES
    --------------------
+   expiretokendays: Number of days before REST API token will expire. The default is 1.
+                    use 'never' if you want your token to never expire.
```
whowutwut: What's the ordering this variable? Should we do alphabetic?

perl-xCAT/xCAT/Schema.pm
```diff
@@ -1802,12 +1804,14 @@ zvmivp => {
...
     token => {
-        cols         => [qw(tokenid username expire comments disable)],
+        cols         => [qw(tokenid username created access expire comments disable)],
         keys         => [qw(tokenid)],
         table_desc   => 'The token of users for authentication.',
         descriptions => {
             tokenid  => 'It is a UUID as an unified identify for the user.',
             username => 'The user name.',
+            created  => 'Creation time for this token.',
+            access   => 'Last access time for this token.',
```
whowutwut: What's the implication of adding a variable in the middle of this, does this cause issues in backwards compatibility?  

If we have an existing token table with values and update xCAT, what happens to the data? 

@immarvin  thoughts?

xCAT-server/lib/perl/xCAT/xcatd.pm
```diff
@@ -351,18 +353,21 @@ sub validate {
     return 0;
 }
 
-my $tokentimeout = 86400;    # one day
+#MG
```
whowutwut: This comment should be removed

xCAT-server/lib/perl/xCAT/xcatd.pm
```diff
@@ -371,16 +376,34 @@ sub gettoken {
     my $tokens = $tokentb->getAllEntries;
     foreach my $token (@{$tokens}) {
 
-        #clean the expired token
-        if ($token->{'expire'} < time()) {
+        # Clean the expired tokens
```
whowutwut: I think token clean up should be the responsibility of the user or admin, we should not do this automatically.  Perhaps the design of the 1 day token expiration is not really useful, and during testing we generated a lot of tokens.  In real world, I don't think tokens will be generated very often,  so I think token management should be users responsibility.  (unless feedback indicates otherwise)

xCAT-server/lib/perl/xCAT/xcatd.pm
```diff
@@ -371,16 +376,34 @@ sub gettoken {
...
         }
     }
 
-    # create a new token for this request
+    # create a new token id
     my $uuid       = xCAT::Utils->genUUID();
-    my $expiretime = time() + $tokentimeout;
-    $tokentb->setAttribs({ tokenid => $uuid, username => $user }, { expire => $expiretime });
+    # extract site table setting for number of days before token expires
+    my @entries = xCAT::TableUtils->get_site_attribute("expiretokendays");
```
whowutwut: Is this just how the calls work? Store the result into an array? Would we ever get more than 1 response?

xCAT-server/lib/perl/xCAT/xcatd.pm
```diff
@@ -391,23 +414,38 @@ sub verifytoken {
...
             return $token->{'username'};
         }
     } else {
+        # Token entry was not found
         return undef;
     }
 }
 
+# Return passed in time as a string in YYYY/MM/DD HH:MM:SS format
+sub timeToString() {
```
whowutwut: Is this useful to add as a xCAT Utility function?   Do we have one already?

perl-xCAT/xCAT/Schema.pm
bybai: I think 'created' and 'access' should be put behind 'expire', in case user has script to get token from this table, should keep 'tokenid','user','expire' order. 'comments' and 'disable' are usually seldom used.

xCAT-server/lib/perl/xCAT/xcatd.pm
bybai: We do not have one, it can be added in Utils.pm

gurevichmark:
@bybai @whowutwut Thank you for reviewing this.
New commit implementing all your suggestions.

gurevichmark:
Added a new commit to modify test script and give examples of REST API automated testcases

bybai:
I am OK to merge it

bybai:
@hu-weihua , Could you pick up latest restapit test to have a try? Thanks.

PR #5120, opened 2018-04-18
Update the CUDA installation test script

neo954:
Use `$installroot` in the test script `rhel-cuda-diskless-install.sh`.


hu-weihua:
It is ok for me. thanks @neo954

PR #5133, opened 2018-04-23
Missing closing bracket

gurevichmark:
Build error caused by #5116 
Missing closing bracket

### DIFF
--- a/xCAT-server/lib/perl/xCAT/xcatd.pm
+++ b/xCAT-server/lib/perl/xCAT/xcatd.pm
@@ -476,4 +476,5 @@ sub verifytoken {
         # Token entry was not found
         return undef;
     }
+}
 1;


cxhong:
Good!

whowutwut:
Was this caught by Travis and how do we prevent this from happening again or could we? Did sandbox build catch it?

gurevichmark:
This happened when I moved one of the functions from `xcatd.pm` to `Utils.pm`. While removing the function from the end of `xcatd.pm`, I deleted one line too many. Thinking that no functional changes were done, I did not retest.

whowutwut:
Does a sandbox build catch it? On commit push the Travis should do this... did that fail?

gurevichmark:
The build was successful, this mistake prevents `xcatd` from starting.

PR #5135, opened 2018-04-24
Slight tweak the hack_uname() hack

neo954:
This patch update the `hack_uname()` function to only check directories whose name begin with a number.


neo954:
Will close this pull request

Issue #3902, opened 2017-09-12
Mismatch between nodelist.status and actual node status after discovery.


samveen:
xCATd version:
```
[samveen@xcatmn ~]$ lsxcatd -t
This is a Management Node
[samveen@xcatmn ~]$ lsxcatd -v
Version 2.13.6 (git commit 0ee33b536bc3c95e7537e8fb097c9ee43e34d0f9, built Tue Aug  8 07:15:43 EDT 2017)
```

Node `chain.chain` attribute:
```
[samveen@xcatmn ~]$ nodels vndc-srv-01110,vndc-srv-01111,vndc-srv-01107,vndc-srv-01108,vndc-srv-01105,vndc-srv-01102,vndc-srv-01112,vndc-srv-01101,vndc-srv-01114,vndc-srv-01100,vndc-srv-01095,vndc-srv-01097,vndc-srv-01103,vndc-srv-01104,vndc-srv-01109,vndc-srv-01113,vndc-srv-01098,vndc-srv-01115,vndc-srv-01099,vndc-srv-01106,vndc-srv-01096 chain.chain |xcoll -n
====================================
vndc-srv-01095,vndc-srv-01096,vndc-srv-01097,vndc-srv-01098,vndc-srv-01099,vndc-srv-01100,vndc-srv-01101,vndc-srv-01102,vndc-srv-01103,vndc-srv-01104,vndc-srv-01105,vndc-srv-01106,vndc-srv-01107,vndc-srv-01108,vndc-srv-01109,vndc-srv-01110,vndc-srv-01111,vndc-srv-01112,vndc-srv-01113,vndc-srv-01114,vndc-srv-01115
====================================
runcmd=bmcsetup,runimage=http://10.209.110.110/install/custom/runimage/d42_s_level_Discovered.tgz,runimage=http://10.209.110.110/install/custom/runimage/glpi_update.tgz,standby,shell
```

Actual node status:
```
[samveen@xcatmn ~]$ sudo PATH=$PATH bash -c 'xdsh all  "tail -n1  screenlog*|cut -f1 -d,"' |xcoll -n
====================================
[...]

Issue #4248, opened 2017-11-06
When booting into xCAT genesis shell, the status should be updated to be consistent


whowutwut:
Looking at this node: 
```
[root@briggs01 .ssh]# lsdef mid05tor12cn11 -i provmethod,status
Object name: mid05tor12cn11
    currchain=boot
    currstate=shell
    provmethod=rhels7.4-alternate-ppc64le-install-compute
    status=powering-on
```

There's no update on the status here,  we should update something to get it into a recognized state, perhaps it should be `status=shell`   Because leaving it in powering-on seems to be an error....

robin2008:
Agree, and I think it might be a long story for us to get a clear state transition.

we have some discussion when implementing 'xcatprobe clusterstatus' feature, and plan to enhance the state transition for different kinds of operation if we hit.

immarvin:
agree that the status update should be enhanced for procedures besides provision

whowutwut:
I was talking to @cxhong today and it seems like we should have some system status (powering-on, netbooting/installing,booted) and then it should go into system status where we run the post script and after all xcat postscript processing is done. Maybe end up with ready state

I know some folks in the past have asked how do we know when all of xcat postscript is complete. It might be good to update some file on the machine to contain the status so applications can check this "ready" or final state. Just some ideas. Maybe not the right issue to discuss in

zet809:
Seems related to issue #3902

xuweibj:
Debug record:

1. Add change status module for shell/standby state in "setdestiny" function in destiny.pm.

Scenario 1:
2. Run ``nodeset shell/standby`` against CN.
3. Check status of CN, it's "shell/standingby"
4. To let CN enter shell/standingby, run ``rpower boot``.
5. Check status, is "powering-on"

Scenario 2:
2. Run ``nodeset shell,standby`` against CN. It's same with the scenario 1 until exit shell.
3. After exit shell, status is changed to "standingby"

To do:
1. CN send status when enter shell/standby.
2. MN should receive it and change status.

PR #5136, opened 2018-04-24
update node status when enter shell/standby

xuweibj:
#4248 #3902 

When CN enter "shell/standby", run "getdestiny" to update node's status in getdestiny function destiny.pm.

UT steps:

1. ``nodeset shell,standby``
2. ``lsdef`` get that  status is "shell"
3. ``rpower boot``, check status is "powering-on"
4. When CN enter genesis, check status is "shell"
5. exit shell, check status is "standingby"

The reason that modify like this:

After run "nodeset", will create petitboot file for node including "destiny=*".
When node boot, will load petitboot file, get destiny and then enter corresponding status.

To update status after step 3, need to send status back to MN to update status. If add new command to update, we need to modify several parts code and modify policy table. But if we run "getdestiny" again, we do not need modify other code, and have tested have no other influence.


zet809:
Looks good, I agree to merge. Thx!

zet809:
This pull request may not fix #3902, so I remove it and merge this PR.

PR #5166, opened 2018-05-03
TravisCI wording and JSON format check

gurevichmark:
Fix some TravisCI grammar and wording.
Display contents of JSON string for PR comments.

### DIFF
--- a/travis.pl
+++ b/travis.pl
@@ -27,8 +27,8 @@ my $last_func_start = timelocal(localtime());
 #--------------------------------------------------------
 # Fuction name: runcmd
 # Description:  run a command after 'cmd' label in one case
-# Atrributes:
-# Retrun code:
+# Attributes:
+# Return code:
 #      $::RUNCMD_RC : the return code of command
 #      @$outref  : the output of command
 #--------------------------------------------------------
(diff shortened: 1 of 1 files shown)


travis.pl
```diff
@@ -143,19 +147,28 @@ sub check_pr_format{
...
+    my $comment_len = 0;
+    my $comment_content;
+    if ($comment_url_resp) {
+        print "\n\n>>>>>Dumper comment_url_resp:\n";
+        print Dumper $comment_url_resp;
+
+        $comment_content = $json->decode($comment_url_resp);
+        $comment_len = @$comment_content;
+    } else {
+        print "[send_back_comment] Not able to get response from $comment_url \n";
```
hu-weihua: Should return here if can not get response?

hu-weihua:
The main process is ok for me. Let's merge it first to develop iteratively quickly. Thanks @gurevichmark

travis.pl
gurevichmark: Yes, you are right. It would be better to `return 0` here, so that it would not stop Trevis execution, but no reason to continue in this function since we can not get the data necessary to update the PR. We can make this change next time this file is updated.

Issue #5158, opened 2018-05-02
[OpenBMC] Improve rinv performance for commonly used options


gurevichmark:
Currently `rinv <node> <option>` calls REST to extract all the inventory data from BMC `/inventory/enumerate` and then displays only the information for a specified `<option>`. That could take around 10 seconds. 
We can improve the performance by calling REST specifically for the data requested, for example `/inventory/serial`.

Related to #3889

PR #5185, opened 2018-05-09
OpenBMC rinv more specific REST calls

gurevichmark:
#5158 
Depending on a `rinv` option specified, call the specific REST API to get just that data, instead of getting all inventory data and filtering our output later.
No external input or output changes.


xuweibj:
Looks good to me. Agree to merge. Thanks @gurevichmark .

Issue #922, opened 2016-04-13
[CUSTOMER] Ping not working for non root users on Diskless PowerKVM guests (rhels7.2)


mew2057:
Hi, 
I've been playing with provisioning Diskless VM guests on PowerKVM and I ran into a bit of a problem. When I ssh into the guest and attempt to ping as a user my user is denied access to the ping command. This issue is resolved when I change permissions for /bin/ping, but it seems to represent an underlying permissions issue that I can't quite source. As far as I can tell SELinux is disabled on both the xcat master and guest nodes.

Thanks,
John

> [root@localhost ~]# ping 10.4.4.24
> PING 10.4.4.24 (10.4.4.24) 56(84) bytes of data.
> 64 bytes from 10.4.4.24: icmp_seq=1 ttl=64 time=0.152 ms
> 64 bytes from 10.4.4.24: icmp_seq=2 ttl=64 time=0.106 ms
> 64 bytes from 10.4.4.24: icmp_seq=3 ttl=64 time=0.060 ms
> ^C
> --- 10.4.4.24 ping statistics ---
> 3 packets transmitted, 3 received, 0% packet loss, time 1998ms
> rtt min/avg/max/mdev = 0.060/0.106/0.152/0.037 ms
>[root@localhost ~]# su - jdunham
> [jdunham@localhost ~]# ping 10.4.4.24
> ping: icmp open socket: Operation not permitted
> [jdunham@localhost ~]# exit
> [root@localhost ~]# chmod u+s /bin/ping
> [root@localhost ~]# su - jdunham
> Last login: Wed Apr 13 14:58:49 UTC 2016 on pts/0
> [jdunham@localhost ~]# ping 10.4.4.24
> PING 10.4.4.24 (10.4.4.24) 56(84) bytes of data.
> 64 bytes from 10.4.4.24: icmp_seq=1 ttl=64 time=0.447 ms
> 64 bytes from 10.4.4.24: icmp_seq=2 ttl=64 time=0.097 ms
> ^C
> --- 10.4.4.24 ping statistics ---
> 2 packets transmitted, 2 received, 0% packet loss, time 999ms
> rtt min/avg/max/mdev = 0.097/0.272/0.447/0.175 ms

whowutwut:
@mew2057  Just doing some debug here... 

On rhels7.2 ppc64le, I do see this problem.  

sles12 ppc64le, it does not exist.. 
```
fs2vm112:~ # ping 127.0.0.1
[...]

PR #5187, opened 2018-05-10
use tar as the default archive method instead of cpio

immarvin:
for issue https://github.com/xcat2/xcat-core/issues/922:

use tar as the default archive method instead of cpio 

UT:

```
[root@boston02 xCAT_plugin]# packimage rhels7.5-alt-rc2-ppc64le-netboot-compute
Packing contents of /install/netboot/rhels7.5-alternate/ppc64le/service/rootimg
archive method:tar
compress method:gzip

```

### DIFF
--- a/docs/source/guides/admin-guides/references/man1/packimage.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/packimage.1.rst
@@ -53,7 +53,7 @@ OPTIONS
 
 \ **-v**\           Command Version.
 
-\ **-m| -**\ **-method**\           Archive Method (cpio,tar,squashfs, default is cpio)
+\ **-m| -**\ **-method**\           Archive Method (cpio,tar,squashfs, default is tar)
 
 \ **-c| -**\ **-compress**\           Compress Method (pigz,gzip,xz, default is pigz/gzip)
 
--- a/xCAT-client/pods/man1/packimage.1.pod
+++ b/xCAT-client/pods/man1/packimage.1.pod
@@ -27,7 +27,7 @@ B<-h>          Display usage message.
 
 B<-v>          Command Version.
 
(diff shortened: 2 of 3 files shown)


robin2008:
You still need to update the `packimg` manpage or doc to say `tar` is default now.

immarvin:
hi @robin2008 , manpage and doc updated

robin2008:
close/reopen it for a new CI check

robin2008:
use tar as default, and leverage automation to test on all platform.
if we found it does not work on some platform, special handing that platform.

robin2008:
I just though the changes will not bring any new actions on user and diskless installation will cover everything.

It is not good user have to run `nodeset` when I just using another method to pack the image.

For #922 issue,  the key is that some commands (like, `ping`, `arping`) need special Linux capability(`man capabilities`),  there does not have many of those commands. You can run below to get:
```
[root@briggs01 ~]# getcap /usr/bin/*
/usr/bin/ping = cap_net_admin,cap_net_raw+p
```
For `root`, it default has such capabilities, but for normal user, need to be set via `setcap`

When using `cpio`, it will remove the capability from the file, so for normal user, it cannot work. But using `tar` could persist those capabilities.  So here is the fix in #922, but it only introduce the pack method, user has to explicitly specify it in CLI option '-m tar', and default is still 'cpio'

This PR is just change the default packing method to use `tar`, but it looks like using `tar` still introduce some other issues (For `var/cache`, it is better to put in into default exlist).

Actually we did not to get the complaints on losing the CAP for normal user, there are still another workaround,  to make a postscript to run `setcap` for those special commands.

immarvin:
this PR will be reverted due to backward compatibility issue

After https://github.com/xcat2/xcat2-task-management/issues/143 is completed, will reconsider this PR

Issue #5153, opened 2018-05-01
Several SSL listener processes running


gurevichmark:
Maybe this is related to #4996 , but on some of our machines there are several `SSL listener` processes running:

```
[root@stratton01 xcat]# ps -ef | grep "SSL li" | grep -v color
root       4615      1  0 Apr10 ?        00:38:33 xcatd: SSL listener
root       6698      1  0 09:38 ?        00:00:03 xcatd: SSL listener
[root@stratton01 xcat]#
```

```
[root@briggs01 ~]# ps -ef | grep "SSL li" | grep -v color
root      10167      1  0 Apr13 ?        00:00:00 xcatd: SSL listener
root      11168      1  0 Apr17 ?        00:00:00 xcatd: SSL listener
root      44539      1  0 Apr16 ?        00:00:00 xcatd: SSL listener
root     102160      1  0 09:35 ?        00:00:00 xcatd: SSL listener
[root@briggs01 ~]#
```

I am not able to recreate this, but it also appears to happen during the Travis testing, #5152

immarvin:
hi @robin2008 , any idea on this? seems there is no dangling request process under xcatd

robin2008:
```
# strace -p 44539
strace: Process 44539 attached
waitpid(-1,
```
SSL listener just wait for all children quit, and the children is pending.

```
# strace -p 44541
strace: Process 44541 attached
recv(12,


^Cstrace: Process 44541 detached
 <detached ...>
# strace -p 44550
strace: Process 44550 attached
recv(12, ^Cstrace: Process 44550 detached
 <detached ...>
# strace -p 44540
strace: Process 44540 attached
_newselect(16, [8], NULL, NULL, NULL
```

robin2008:
After `kill 44540`, we can see two DB connection process is quit,  but SSL listener (44539) is 100% CPU, using strace to attach it and see below:
```
waitpid(-1, 0x3fffcd2917d4, 0)          = -1 ECHILD (No child processes)
waitpid(-1, 0x3fffcd2917d4, 0)          = -1 ECHILD (No child processes)
waitpid(-1, 0x3fffcd2917d4, 0)          = -1 ECHILD (No child processes)
waitpid(-1, 0x3fffcd2917d4, 0)          = -1 ECHILD (No child processes)
waitpid(-1, 0x3fffcd2917d4, 0)          = -1 ECHILD (No child processes)
[...]

PR #5215, opened 2018-05-16
there are several xcatd SSL process remaining

robin2008:
Fix the issue #5153, to reduce the possibility that  the SSL process cannot quit successfully.
    ~~- using non-block waitpid when SSL quit, and add a bit more logs~~
    - Disable the signal handler after the SSL will quit.
    - refine the `generic_reaper` and `ssl_reaper` to better cover $pid_UDP
    - clean the $pid_UDP for child process, so that they don't need to consider UPD process and remove the code from  `plugin_reaper` and `dispatch_reaper`

UT: 
1, run xcattest for the regression.
```
xcattest -t xcatd_start
xCAT automated test started at Wed May 16 02:14:21 2018
******************************
To detect current test environment
******************************
Detecting: OS = linux
Warning: No compute node defined, can't get ARCH of compute node
Warning: No compute node defined, can't get HCP TYPE of compute node
******************************
loading test cases
******************************
To run:
xcatd_start
******************************
Start to run test cases
******************************
------START::xcatd_start::Time:Wed May 16 02:14:21 2018------

RUN:service xcatd status [Wed May 16 02:14:21 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
xcatd service is running
CHECK:rc == 0	[Pass]
CHECK:output =~ xcatd service is running	[Pass]

RUN:service xcatd stop [Wed May 16 02:14:21 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
Stopping xcatd (via systemctl):  [  OK  ]
CHECK:rc == 0	[Pass]

RUN:sleep 3 [Wed May 16 02:14:21 2018]
ElapsedTime:3 sec
RETURN rc = 0
OUTPUT:

RUN:service xcatd start [Wed May 16 02:14:24 2018]
ElapsedTime:3 sec
RETURN rc = 0
OUTPUT:
Starting xcatd (via systemctl):  [  OK  ]
CHECK:rc == 0	[Pass]
CHECK:output =~ Starting xcatd	[Pass]
CHECK:output =~ [  OK  ]	[Pass]

RUN:service xcatd status [Wed May 16 02:14:27 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
xcatd service is running
CHECK:rc == 0	[Pass]
CHECK:output =~ xcatd service is running	[Pass]

RUN:result=`ps -ef | grep "xcatd: SSL listener" | grep -v grep | wc -l`; if [ $result -eq 1 ]; then echo "only one process"; fi [Wed May 16 02:14:27 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
only one process
CHECK:output =~ only one process	[Pass]

RUN:result=`ps -ef | grep "xcatd: DB Access" | grep -v grep | wc -l`; if [ $result -eq 1 ]; then echo "only one process"; fi [Wed May 16 02:14:27 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
only one process
CHECK:output =~ only one process	[Pass]

RUN:result=`ps -ef | grep "xcatd: UDP listener" | grep -v grep | wc -l`; if [ $result -eq 1 ]; then echo "only one process"; fi [Wed May 16 02:14:27 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
only one process
CHECK:output =~ only one process	[Pass]

RUN:result=`ps -ef | grep "xcatd: install monitor" | grep -v grep | wc -l`; if [ $result -eq 1 ]; then echo "only one process"; fi [Wed May 16 02:14:27 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
only one process
CHECK:output =~ only one process	[Pass]

RUN:result=`ps -ef | grep "xcatd: Discovery worke" | grep -v grep | wc -l`; if [ $result -eq 1 ]; then echo "only one process"; fi [Wed May 16 02:14:27 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
only one process
CHECK:output =~ only one process	[Pass]

RUN:result=`ps -ef | grep "xcatd: Command log writer" | grep -v grep | wc -l`; if [ $result -eq 1 ]; then echo "only one process"; fi [Wed May 16 02:14:27 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
only one process
CHECK:output =~ only one process	[Pass]

------END::xcatd_start::Passed::Time:Wed May 16 02:14:27 2018 ::Duration::6 sec------
------Total: 1 , Failed: 0------

xCAT automated test finished at Wed May 16 02:14:27 2018
Please check results in the /opt/xcat/bin/../share/xcat/tools/autotest/result/,
and see /opt/xcat/bin/../share/xcat/tools/autotest/result//failedcases.20180516021421 file for failed cases.
see /opt/xcat/bin/../share/xcat/tools/autotest/result//performance.report.20180516021421 file for time consumption
```
2, test the `restartxcatd`. 
on Terminal 1, run `xdsh <> "sleep 60"`
on terminal 2, run `restartxcatd`

```
restartxcatd
restartxcatd invoked by root.

Restarting xCATd                                           [  OK  ]
# ps -ef|grep xcatd
root     13796     1  0 02:17 ?        00:00:00 xcatd: SSL listener
root     13797 13796  0 02:17 ?        00:00:00 xcatd: DB Access
postgres 13798  7072  0 02:17 ?        00:00:00 postgres: xcatadm xcatdb 10.3.5.20(33870) idle
root     13818 13796  0 02:18 ?        00:00:00 xcatd SSL: xdsh to c910f03c05k24 for root@localhost
root     13819 13818  0 02:18 ?        00:00:00 xcatd SSL: xdsh to c910f03c05k24 for root@localhost: xdsh instance
root     13866     1  1 02:18 ?        00:00:00 xcatd: SSL listener
root     13867 13866  1 02:18 ?        00:00:00 xcatd: DB Access
postgres 13868  7072  1 02:18 ?        00:00:00 postgres: xcatadm xcatdb 10.3.5.20(33886) idle
root     13869 13866  0 02:18 ?        00:00:00 xcatd: UDP listener
root     13870 13869  0 02:18 ?        00:00:00 xcatd: Discovery worker
root     13871 13866  0 02:18 ?        00:00:00 xcatd: install monitor
root     13872 13866  0 02:18 ?        00:00:00 xcatd: Command log writer
postgres 13877  7072  0 02:18 ?        00:00:00 postgres: xcatadm xcatdb 10.3.5.20(33888) idle
```
OK, fastreboot is still working, and we can see the new logs
```
May 16 02:18:14 c910f03c05k20 xcat[13796]:  xcatd main service 13796 quiescing
May 16 02:18:14 c910f03c05k20 xcat[13796]:  INFO xcatd is going to stop, waiting for 1 plugins to quit...
```

### DIFF
--- a/xCAT-server/sbin/xcatd
+++ b/xCAT-server/sbin/xcatd
@@ -1003,11 +1003,16 @@ my %immediatechildren;
 
 sub generic_reaper {
     local ($!);
+    #print "generic_reaper in $$...";
     while (($CHILDPID = waitpid(-1, WNOHANG)) > 0) {
-        if (($CHILDPID == $pid_UDP) && ($udpctl)) {
-
-            # got here because UDP child is gone
-            close($udpctl); $udpctl = 0;
+        #print "reaper for $CHILDPID...\n";
+        if ($CHILDPID == $pid_UDP) {
+            if ($udpctl) {
+                # got here because UDP child is gone
+                close($udpctl);
+                $udpctl = 0;
+            }
+            $pid_UDP = 0;
         }
         yield;
     }
@@ -1017,16 +1022,20 @@ sub generic_reaper {
 sub ssl_reaper {
     local ($!);
     my $numdone = 0;
+    #print "ssl_reaper in $$...";
     while (($CHILDPID = waitpid(-1, WNOHANG)) > 0) {
+        #print "reaper for $CHILDPID...\n";
         if ($immediatechildren{$CHILDPID}) {
             delete $immediatechildren{$CHILDPID};
             $sslclients--;
             $numdone--;
         }
-        if (($CHILDPID == $pid_UDP) && ($udpctl)) {
-
-            # got here because UDP child is gone
-            close($udpctl); $udpctl = 0;
+        if ($CHILDPID == $pid_UDP) {
+            if ($udpctl) {
+                # got here because UDP child is gone
+                close($udpctl);
+                $udpctl = 0;
+            }
             $pid_UDP = 0;
         }
         if ($CHILDPID == $cmdlog_svrpid) {
@@ -1041,32 +1050,36 @@ sub ssl_reaper {
 
 sub dispatch_reaper {
     local ($!);
+    #print "dispatch_reaper in $$...";
     while (($CHILDPID = waitpid(-1, WNOHANG)) > 0) {
+        #print "reaper for $CHILDPID...\n";
         if ($dispatched_children{$CHILDPID}) {
             delete $dispatched_children{$CHILDPID};
             $dispatch_children--;
         }
-        if (($CHILDPID == $pid_UDP) && ($udpctl)) {
-
-            # got here because UDP child is gone
-            close($udpctl); $udpctl = 0;
-        }
+        #if (($CHILDPID == $pid_UDP) && ($udpctl)) {
+        #
+        #    # got here because UDP child is gone
+        #    close($udpctl); $udpctl = 0;
+        #}
     }
     $SIG{CHLD} = \&dispatch_reaper;
 }
 
 sub plugin_reaper {
     local ($!);
+    #print "plugin_reaper in $$...";
     while (($CHILDPID = waitpid(-1, WNOHANG)) > 0) {
+        #print "reaper for $CHILDPID...\n";
         if ($plugin_children{$CHILDPID}) {
             delete $plugin_children{$CHILDPID};
             $plugin_numchildren--;
         }
-        if (($CHILDPID == $pid_UDP) && ($udpctl)) {
-
-            # got here because UDP child is gone
-            close($udpctl); $udpctl = 0;
-        }
+        #if (($CHILDPID == $pid_UDP) && ($udpctl)) {
+        #
+        #    # got here because UDP child is gone
+        #    close($udpctl); $udpctl = 0;
+        #}
     }
     $SIG{CHLD} = \&plugin_reaper;
 }
@@ -1162,6 +1175,7 @@ if (!defined $pid_MON) {
 }
 unless ($pid_MON) {
     $$progname = "xcatd: install monitor";
+    $pid_UDP = 0;
     close($udpctl); $udpctl = 0;
     do_installm_service;
     xexit(0);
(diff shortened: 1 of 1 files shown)


xCAT-server/sbin/xcatd
```diff
@@ -1614,18 +1628,32 @@ if (open($mainpidfile, "<", "/var/run/xcat/mainservice.pid")) {
...
-    } elsif ($lastpid == $pid_UDP) {
-        $pid_UDP = 0;
-    } elsif ($lastpid == $cmdlog_svrpid) {
-        $cmdlog_svrpid = 0;
-    } elsif ($lastpid == $pid_MON) {
-        $pid_MON = 0;
+my $remains = keys %immediatechildren;
+xCAT::MsgUtils->trace(0, 'I', "xcatd is going to stop, waiting for $remains plugins to quit...");
+while ($remains || $pid_UDP || $cmdlog_svrpid || $pid_MON) {
+    my $lastpid = waitpid(-1, WNOHANG);
```
chenglch: I hope to know the reason to replace wait() with waitpid(WNOHANG).

robin2008: The found issue is that SSL and DB is remaining, but no other child process.  It is not recreated normally, One potential cause is that $pid_UDP is not set to 0 sometimes when UDP process is quit, and it cause here just SSL block at `wait`.

To use `waitpid` here then we can have a chance in future to check if only DB process is alive (in case %immediatechildren contains PID which does not alive due to some unkown issue),  but I think this might be rare, so not do it this time.

chenglch: I don't think the explanation is reasonable.
 
See the original code around `$lastpid = wait();`, all of the global variables are reset, how could the code reach here but the `while` condition is true?   Have you checked the `child` signal handler  or the other `wait` calls that may influence the result of `while` condition here?

robin2008: It was observed the case that it block at `wait()` and all other child process are quit except for DB process.  So it is definitely some child handler not working as expected to set the global variables or SIG_CHILD is not passed rightly in some unexpected case (but we cannot recreate it). 

So the fix in child handler it to make the `while` work (`$pid_UDP` will always set to zero when it quit), but the changes  from `wait()` to `waitpid()` here is not for it, using `waitpid` and `wait` is doing the same thing to block here waiting for children done (except DB).

Without the `waitpid` change, it still works. But using `waitpid` just let we have chance to peek into the loop and do more (TODO) when things are wrong.

PR #5216, opened 2018-05-16
update case for issue 4855,  to compare the output of perl vs python for bmcstate when bmc reboots

tingtli:
Case Name: rpower_bmcreboot_perl_python_output
description:record the output for rpower bmcreboot and compare the output for perl and python version.


hu-weihua:
it is ok for me,thanks @tingtli

PR #5231, opened 2018-05-21
Thinker xCAT/postscripts/cuda_power9_setup,

neo954:
* Fix the 40-redhat.rules udev rule
  * Remake initrd image for diskfull compute node


neo954:
Have to tinker `cuda_power9_setup` after merge pull request #5137

whowutwut:
@markprez can you take a look at the changes here?

whowutwut:
@markprez, if you could also pull down these changes and incorporate it into the next image you launch, that would be appreciated.  Thanks

PR #5253, opened 2018-05-26
Update the testcase to cover the changes in #5119

robin2008:
Update the test cases to cover the changes in https://github.com/xcat2/xcat-core/pull/5119
using regular expression to cover the error/warn message which contains the server name

Here just use `(\[.*?\]: )?` to math the both output case for:
`Error: this is an example error message` and `Error: [sn01]: this is an example error message`

As this is general, just use one case for UT
UT:
```
XCATTEST_CN=c910f03c05k30 xcattest -t nodeset_check_warninginfo
xCAT automated test started at Fri May 25 23:47:56 2018
******************************
To detect current test environment
******************************
Detecting: OS = linux
Detecting: ARCH = ppc64le
Detecting: HCP = kvm
******************************
loading test cases
******************************
To run:
nodeset_check_warninginfo
******************************
Start to run test cases
******************************
------START::nodeset_check_warninginfo::Time:Fri May 25 23:47:58 2018------

RUN:if [[ "__GETNODEATTR(c910f03c05k30,arch)__" =~ "el" || "__GETNODEATTR(c910f03c05k30,arch)__" =~ "le" ]]; then bootloader=xnba; else bootloader=petitboot; fi; mkdef -t node -o testnode1 arch=ppc64el cons=ipmi groups=pbmc mgt=ipmi ip=10.1.1.200 mac=e6:d4:d2:3a:ad:06  monserver=10.1.1.1 nameservers=10.1.1.1 nodetype=ppc,osi profile=compute tftpserver=10.1.1.1 xcatmaster=10.1.1.1 netboot=$bootloader [Fri May 25 23:47:58 2018]
ElapsedTime:1 sec
RETURN rc = 0
OUTPUT:
1 object definitions have been created or modified.
CHECK:rc == 0	[Pass]

RUN:lsdef testnode1 [Fri May 25 23:47:59 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
Object name: testnode1
    arch=ppc64el
    cons=ipmi
    groups=pbmc
    ip=10.1.1.200
    mac=e6:d4:d2:3a:ad:06
    mgt=ipmi
    monserver=10.1.1.1
    nameservers=10.1.1.1
    netboot=xnba
    nodetype=ppc,osi
    postbootscripts=otherpkgs
    postscripts=syslog,remoteshell,syncfiles
    profile=compute
    tftpserver=10.1.1.1
    xcatmaster=10.1.1.1

[...]

PR #5119, opened 2018-04-18
show server name by default for Error and Warn message

robin2008:
For #4717 
Now xcat cli could show where the message is coming from (say mn01, or sn02, etc) when enable `-V`,  but in many case, user may forget to use it, especially for error/warn message, it will confuse user.

This PR is to show server name by default for Error and Warn message.
Before fix:
```
nodeset fake1 boot
fake1: boot
Error: dhcp server is not running.  please start the dhcp server.
fake1: boot
Error: dhcp server is not running.  please start the dhcp server.
```
After fix:
```
nodeset fake1 boot
fake1: boot
Error: [c910f03c05k20]: dhcp server is not running.  please start the dhcp server.
fake1: boot
Error: [c910f03c05k24]: dhcp server is not running.  please start the dhcp server.
```

For warnig message,  as below:
```
nodeset fake1 boot
fake1: boot
Warning: [c910f03c05k20]: Unable to find mac address for fake1
fake1: boot
Error: [c910f03c05k24]: dhcp server is not running.  please start the dhcp server.
```

PR #5259, opened 2018-05-28
Update the testcase to cover the changes for showing server name in warn/error message

robin2008:
Update the testcase to cover the changes in #5119
- missing some test cases

This is supplement fix for #5253


hu-weihua:
Related task xcat2/xcat2-task-management#132

hu-weihua:
It is ok for me, thanks @robin2008

PR #5261, opened 2018-05-29
check ip whether static for networkmanager config

xuweibj:
https://github.com/xcat2/xcat2-task-management/issues/154

If ipv4.method is 'manual' by `nmcli con show $nic 2>&1 | grep -E 'ipv4.method'`, it is static ip.
If not, it is dhcp.
If run cmd failed, check ip config file.

### DIFF
--- a/xCAT-probe/lib/perl/probe_utils.pm
+++ b/xCAT-probe/lib/perl/probe_utils.pm
@@ -179,12 +179,22 @@ sub is_static_ip {
     my $os  = get_os();
     my $rst = 0;
 
-    my $ipv4_method = `nmcli con show $nic 2>&1 | grep -E 'ipv4.method'`;
+    `which nmcli > /dev/null 2>&1`;
     unless ($?) {
-        if ($ipv4_method =~ /manual/) {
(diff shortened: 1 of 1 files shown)


xCAT-probe/lib/perl/probe_utils.pm
```diff
@@ -179,6 +179,14 @@ sub is_static_ip {
     my $os  = get_os();
     my $rst = 0;
 
+    my $ipv4_method = `nmcli con show $nic 2>&1 | grep -E 'ipv4.method'`;
```
robin2008: I think here `$nic` is the device name, it might not be the same as the connection name. I just got below on one env.
```
nmcli con show --active
NAME         UUID                                  TYPE            DEVICE
System eth0  8fb0a4ef-fd62-4c3c-8b4f-f34c0ba12302  802-3-ethernet  eth0
docker0      4dfabb05-4958-48c7-b570-89f02e518a83  bridge          docker0

nmcli con show eth0
Error: eth0 - no such connection profile.
```

Issue #5260, opened 2018-05-28
chtab try to  access /etc/xcat/*.sqlite by its own, but this does not work for non-root users


dhilst:
Hi everybody,

I was testing xCAT with non-root users. I followed the [docs](https://sourceforge.net/p/xcat/wiki/Granting_Users_xCAT_privileges/) for setting up user certificates, etc.. So I fired a `chtab key=... ` and got a timeout.

After some digging, I found (with strace) that `chtab` was trying to open `/etc/xcat/passwd.sqlite` by it's own and facing denied access. Bellow is the relevant strace lines... It try about 3 times, at each second, then it times out.

As work around is possible to use `sudo /opt/xcat/sbin/chtab ...` or `tabedit` that works fine with non-root users. 

```
getegid()                               = 1000
stat("/etc/xcat/site.sqlite", {st_mode=S_IFREG|0600, st_size=3072, ...}) = 0
open("/etc/xcat/site.sqlite", O_RDWR|O_CREAT|O_CLOEXEC, 0644) = -1 EACCES (Permission denied)
open("/etc/xcat/site.sqlite", O_RDONLY|O_CLOEXEC) = -1 EACCES (Permission denied)
rt_sigprocmask(SIG_BLOCK, [CHLD], [], 8) = 0
rt_sigaction(SIGCHLD, NULL, {SIG_DFL, [], 0}, 8) = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
nanosleep({1, 0}, 0x7ffd45cfa1f0)       = 0
nanosleep({0, 0}, NULL)                 = 0
stat("/etc/xcat/site.sqlite", {st_mode=S_IFREG|0600, st_size=3072, ...}) = 0
open("/etc/xcat/site.sqlite", O_RDWR|O_CREAT|O_CLOEXEC, 0644) = -1 EACCES (Permission denied)
open("/etc/xcat/site.sqlite", O_RDONLY|O_CLOEXEC) = -1 EACCES (Permission denied)
rt_sigprocmask(SIG_BLOCK, [CHLD], [], 8) = 0
rt_sigaction(SIGCHLD, NULL, {SIG_DFL, [], 0}, 8) = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
nanosleep({2, 0}, 0x7ffd45cfa1f0)       = 0
nanosleep({0, 0}, NULL)                 = 0
[...]

PR #5265, opened 2018-05-30
Add WARNING message if non-root user runs chtab command

cxhong:
For issue #5260

If user is not root,  running `chtab` command will log a warning message.
````
[cxhong@boston02 home]$ chtab
WARNING: chtab worked without passing xcatd, using tabch command instead if user is controled by policy mechanism
````

### DIFF
--- a/xCAT-server/sbin/chtab
+++ b/xCAT-server/sbin/chtab
@@ -66,7 +66,7 @@ unless ($target) {
 my $current_userid = getpwuid($>);
 if ($current_userid ne "root")
 {
-    print "WARNING: chtab bypasses xcatd and does not enforce xCAT policy tables. Using tabch instead.\n";
+    print "WARNING: chtab bypasses xcatd and does not enforce xCAT policy tables. Consider using tabch instead.\n";
 }
 
 my %keyhash = ();
--- a/xCAT-server/sbin/chtab
+++ b/xCAT-server/sbin/chtab
@@ -66,7 +66,7 @@ unless ($target) {
 my $current_userid = getpwuid($>);
 if ($current_userid ne "root")
 {
-    print "WARNING: chtab worked without passing xcatd, using tabch command instead if user is controled by policy mechanism\n";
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: xCAT-server/sbin/chtab [hunk 1]
```diff
@@ -63,6 +63,12 @@ unless ($target) {
     exit(1);
 }
 
+my $current_userid = getpwuid($>);
+if ($current_userid ne "root")
+{
+    print "WARNING: chtab worked without passing xcatd, using tabch command instead if user is controled by policy mechanism\n";
```
whowutwut: Should we change this to: 

WARNING: chtab bypasses xcatd and does not enforce xCAT policy tables.  Consider using tabch instead.

xCAT-server/sbin/chtab
```diff
@@ -63,6 +63,12 @@ unless ($target) {
     exit(1);
 }
 
+my $current_userid = getpwuid($>);
+if ($current_userid ne "root")
+{
+    print "WARNING: chtab bypasses xcatd and does not enforce xCAT policy tables. Using tabch instead.\n";
```
whowutwut: Did you mean to say `Use tabch instead` ?  Or is the code using tabch?

cxhong: the code will not use `tabch`.    
 I am not sure if just exit here if it is non-root user.  maybe some user has root permission, that's why I logged as `WARNING` message,  the command will still process, but it will hit timeout error.
```
Failed to connect to switches table after retrying 3 times: DBI connect('/etc/xcat/switches.sqlite','',...) failed: unable to open database file at /opt/xcat/lib/perl/xCAT/Table.pm line 972.

Can't call method "prepare" on an undefined value at /opt/xcat/lib/perl/xCAT/Table.pm line 998.
```

whowutwut: Can we change that last sentence to `Consider using tabch instead.`

cxhong: thanks, modified.

PR #5269, opened 2018-05-31
Refactor openbmc for better supporting redfish

xuweibj:
https://github.com/xcat2/xcat2-task-management/issues/157

move agent related code to AGENT.pm

```
# rpower f6u03 state
Mon Jun  4 21:37:55 2018 f6u03: [openbmc_debug] login curl -k -c cjar -b cjar -X POST -H "Content-Type: application/json" https://10.6.3.100/login -d '{"data": ["root", "xxxxxx"]}'
Mon Jun  4 21:37:55 2018 f6u03: [openbmc_debug] login 200 OK
Mon Jun  4 21:37:55 2018 f6u03: [openbmc_debug] list_power_states curl -k -c cjar -b cjar -X GET -H "Content-Type: application/json" https://10.6.3.100/xyz/openbmc_project/state/enumerate
Mon Jun  4 21:37:56 2018 f6u03: [openbmc_debug] list_power_states 200 OK
f6u03: on
```


whowutwut:
I don't agree with this PR.   And I think it should not be merged to master, but to a feature branch that should be tested in parallel and merged to master if we think it is OK,

This is a huge change or rename of many files, why is this necessary?  What is the effect of backwards compatibility or coexistence?  If we break this, then we should put this into a feature branch to possible merge in 2.15.  Maybe need to really think about the timeline for RedFish support within xCAT.   Should not break coexistence in minor release...

xuweibj:
@whowutwut , I think you are right. I will delete the changes related with rename. But I think if we are ready for release redfish support, modify the name "xCAT-openbmc-py" is better. 

And I think we could merge perl part to master branch first, we could check whether the changes affect "openbmc" support so that we can fix issue earlier.

robin2008:
@xuweibj  could we only leave the refactor part, but remove the redfish parts?  For example, `schema.pm` and `redfish.pm` as now we cannot support Redfish fully.

xCAT-server/lib/xcat/plugins/openbmc2.pm
```diff
@@ -134,8 +135,8 @@ sub process_request {
     my $request = shift;
     $callback = shift;
 
-    if (!xCAT::OPENBMC::exists_python_agent()) {
-        xCAT::MsgUtils->message("E", { data => ["The xCAT Python agent does not exist. Check if xCAT-openbmc-py package is installed on management node and service nodes."] }, $callback);
+    if (!xCAT::AGENT::exists_python_agent()) {
+        xCAT::MsgUtils->message("E", { data => ["The xCAT Python agent does not exist. Check if xCAT-hwctrl-py package is installed on management node and service nodes."] }, $callback);
```
robin2008: Let's still keep the package name now.

xCAT-server/lib/perl/xCAT/AGENT.pm
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/perl
+## IBM(c) 2107 EPL license http://www.eclipse.org/legal/epl-v10.html
```
robin2008: This would be 2018 here

xCAT-server/lib/xcat/plugins/openbmc2.pm
```diff
@@ -149,14 +150,14 @@ sub process_request {
...
-    my $pid = xCAT::OPENBMC::start_python_agent($$);
+    my $pid = xCAT::AGENT::start_python_agent($$);
     if (!defined($pid)) {
         xCAT::MsgUtils->message("E", { data => ["Failed to start the xCAT Python agent. Check /var/log/xcat/cluster.log for more information."] }, $callback);
         return;
     }
 
-    xCAT::OPENBMC::submit_agent_request($pid, $request, \%node_info, $callback);
-    xCAT::OPENBMC::wait_agent($pid, $callback);
+    xCAT::AGENT::submit_agent_request($pid, $request, "openbmc", \%node_info, $callback);
```
robin2008: Could we use 'OpenBMC' here? as I see it will be printed on message

robin2008:
Trigger a CI again

whowutwut:
@robin2008 not understanding why this was merged?? 
I thought @xuweibj agree that we should NOT merge this one

whowutwut:
@xuweibj please do not amend the commit after review has started. It’s misleading that we have 1 commit but changes were actually pushed to this pull request

xuweibj:
hi @whowutwut , we have discussed and decide to merge some code may affect openbmc first, so that we could test these changes earlier. 
I will not amend commit next time.

Issue #5176, opened 2018-05-08
rflash operation cause 100% CPU


robin2008:
When doing an rflash to switch back to a previously firmware, Observed 'top' on one service node and the rflash main process is using 100% cpu.  And using `strace -p`, got below:
```
waitpid(-1, 0x7fffdedff3b4, WNOHANG)    = 0
waitpid(-1, 0x7fffdedff3b4, WNOHANG)    = 0
waitpid(-1, 0x7fffdedff3b4, WNOHANG)    = 0
waitpid(-1, 0x7fffdedff3b4, WNOHANG)    = 0
waitpid(-1, 0x7fffdedff3b4, WNOHANG)    = 0
waitpid(-1, 0x7fffdedff3b4, WNOHANG)    = 0
waitpid(-1, 0x7fffdedff3b4, WNOHANG)    = 0
waitpid(-1, 0x7fffdedff3b4, WNOHANG)    = 0
waitpid(-1, 0x7fffdedff3b4, WNOHANG)    = 0
waitpid(-1, 0x7fffdedff3b4, WNOHANG)    = 0
waitpid(-1, 0x7fffdedff3b4, WNOHANG)    = 0
```

robin2008:
I think this is caused by 
`while ((my $cpid = waitpid(-1, WNOHANG)) > 0) {` in openbmc.pm plugin

we need to fix the wrong using of NOHANG waitpid.

Pay attention to python version too.

zet809:
Hi, @xuweibj , will you pls help take a look at this issue? Thx!

robin2008:
Now we make some enhancement in #5208, need to verify on the problem environment to see if it is working.

robin2008:
Merged, and not found new issue on rflash, just close it.

whowutwut:
Using May 30 build, this still happens....  

```
# lsxcatd -a
Version 2.14.1 (git commit f10d33f66c449b06104ce734f79af62d55c54564, built Wed May 30 06:15:48 EDT 2018)
This is a Management Node
cfgloc=Pg:dbname=xcatdb;host=192.168.95.1|xcatadm
dbengine=Pg
dbname=xcatdb
dbhost=192.168.95.1
dbadmin=xcatadm
```

`strace -p` on this pid shows: 
```
waitpid(-1, 0x7fffce209364, WNOHANG)    = 0
waitpid(-1, 0x7fffce209364, WNOHANG)    = 0
[...]

PR #5274, opened 2018-06-01
if not get valid pid when waitpid, sleep

xuweibj:
#5176 

When receive invalid ``cpid`` for waitpid, sleep 0.01s.

CPU less than 5% for one node.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -1088,8 +1088,10 @@ rmdir \"/tmp/\$userid\" \n";
         }
 
         if (%child_node_map) {
+            my $pid_flag = 0;
             while ((my $cpid = waitpid(-1, WNOHANG)) > 0) {
                 if ($child_node_map{$cpid}) {
+                    $pid_flag = 1;
                     my $node = $child_node_map{$cpid};
                     my $rc = $? >> 8;
                     if ($rc != 0) {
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1088,7 +1088,9 @@ rmdir \"/tmp/\$userid\" \n";
         }
 
         if (%child_node_map) {
+            my $null_flag = 0;
```
robin2008: Suggest to rename it to `child_end_flag`

xCAT-server/lib/xcat/plugins/openbmc.pm
```diff
@@ -1088,7 +1088,9 @@ rmdir \"/tmp/\$userid\" \n";
...
             while ((my $cpid = waitpid(-1, WNOHANG)) > 0) {
+                $null_flag = 1;
```
robin2008: Need to put this line into the `if` clause

whowutwut:
@robin2008 @xuweibj This pull request should contain more information and how it was tested to prove that it does help.  If customers search in the future, really hard to tell what is going on..

For example, why does the sleep help in this case?  what is going on that causes 100% CPU if this sleep is not there?

robin2008:
CPU 100% is straightforward symptom,  rflash main process will fork child process for uploading image, it will monitor if some of its children finish and do the following things.
But in our former codes, it execute systemcall `waipid` without anyy sleep,  then it consume all the CPU time.

With a tiny sleep, Linux could free the time cycle for other application, then no 100% CPU.

Multiple rflash is run in different process,  there is no relationship. for #5276, I've appended the investigation there.

PR #5298, opened 2018-06-18
Add security notification for 2018-06-12 OpenSSL reported issue - CVE-2018-0732

whowutwut:
Just a FYI to our customers informing about OpenSSL reported vulnerabilities: CVE-2018-0732 (Low)


whowutwut:
Doc only issue, will merge.

PR #5304, opened 2018-06-20
Support Ubuntu 18.04. Add code name bionic to the build script

neo954:
Code name `bionic` is added


neo954:
For task https://github.com/xcat2/xcat2-task-management/issues/207

PR #5308, opened 2018-06-21
fix error of xcatmn

xuweibj:
before modify:
```
[MN]: Checking on MN...                                                                   [FAIL]
   Checking xCAT daemon attributes configuration...                                      [FAIL]
       Attribute xcatmaxbatchconnections must be less than xcatmaxconnections.
```
after modify no this error msg.

### DIFF
--- a/xCAT-probe/subcmds/xcatmn
+++ b/xCAT-probe/subcmds/xcatmn
@@ -1146,7 +1146,9 @@ sub check_daemon_attributes {
     my $node_num = `nodels 2>&1 | wc -l`;
     chomp($node_num);
     my $xcatmaxconnections = 64;
+    my $xcatmaxconnections_site = 64;
     my $xcatmaxbatchconnections = 50;
+    my $xcatmaxbatchconnections_site = 50;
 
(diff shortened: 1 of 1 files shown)


hu-weihua:

whowutwut:
I dont agree with these Pull Request..  

I think we should be more transparent on the attributes that we are checking.., so that the user or xCAT team knows what it is that we care about.. If it's the fact that `xcatmaxbatchconnections must be less than xcatmaxconnections`, then we ONLY check if those are defined... 

The externals should look something like... 

```
...  
[MN]: Checking on MN...                                                   
   Checking xCAT daemon attributes configuration...                                      
       Checking  xcatmaxbatchconnections < xcatmaxconnections ...           [OK]
...
```

If Undefined....  no problem here and just say OK.   If it's wrongly defined, then we should flag it... 

But I don't think we should hard code a default value but instead not even check the values if they are not defined in the MN node, because that means the defaults should apply.

xuweibj:
Hi @whowutwut , I think we could modify the msg in the future. 
But for hard code a default value, if user just set one of the attributes, we should compare them too. At that time, code logic is more complex. And if user does not set them, in xCAT code read default value. So I think use default value here is better.

PR #5330, opened 2018-06-26
modify debug output function of pr 5326

xuweibj:
Use ``probe_utils->send_msg`` instead ``print_check_result``.


xuweibj:
Invalid pr, close.

PR #5335, opened 2018-06-26
create HA doc

bybai:
for https://github.ibm.com/vhu/c910env/issues/304 and https://github.com/xcat2/xcat2-task-management/issues/160

UT link: http://10.5.106.1/install/coral_ha/xcat-core/docs/build/html/references/coral/ha/index.html

### DIFF
--- a/docs/source/references/coral/ha/ha_mgmt_node_with_shared_data.rst
+++ b/docs/source/references/coral/ha/ha_mgmt_node_with_shared_data.rst
@@ -6,7 +6,10 @@ HA Solution Overview
 
 While a xCAT management node ``xcatmn1`` is running as a primary management node, another node - ``xcatmn2`` can be configured to act as primary management node in case ``xcatmn1`` becomes unavailable. The process is manual and requires disabling primary ``xcatmn1`` and activating backup ``xcatmn2``. Both nodes require access to shared storage described below. Use of Virtual IP is also requred.
 
-An interactive sample script ``xcatha`` is availabe to guide through the steps of disabling and activation of xCAT management nodes. ``Dryrun`` option in that scrip allows viewing the actions without executing them.
+An interactive sample script `xcatha.py <https://github.com/xcat2/xcat-extensions/blob/master/HA/xcatha.py>`_ is availabe to guide through the steps of disabling and activation of xCAT management nodes. ``Dryrun`` option in that scrip allows viewing the actions without executing them.
+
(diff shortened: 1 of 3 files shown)


docs/source/references/coral/ha/ha_mgmt_node_with_shared_data.rst
```diff
@@ -163,114 +178,12 @@ Use ``xcatha.py -a`` to start all related services: ::
...
-    #. Make diff::
-
-        diff xcatmn1.yaml xcatmn2.yaml 
-
-#. Run ``xcatprobe xcatmn -i eth0:0`` to find no fatal error
-
-#. Provision an existed compute node like ``cn1`` successfully.
+Unplanned failover: primary xCAT management node is not accessible
+``````````````````````````````````````````````````````````````````
+If primary xCAT management node becomes not accessible before being deactivated and backup xCAT management node is activated, it is recommended that the primary node is disconnected from the network before being rebooted. This will ensure that when services are started on reboot, they do not interfere with the same services running on the backup xCAT management node.
```
whowutwut: @bybai @gurevichmark  The whole purpose of this exercise is to find out and disable the services so they **do not** start up at boot.   Writing this snip means we do not know 100% that we have covered everything and do not know what will happen upon reboot of the old MN node in HA, we need to re-visit this.

docs/source/references/coral/ha/ha_mgmt_node_with_shared_data.rst
```diff
@@ -163,114 +178,12 @@ Use ``xcatha.py -a`` to start all related services: ::
     2018-06-24 22:13:19,353 - DEBUG - makeconservercf [Passed]
     2018-06-24 22:13:19,449 - DEBUG - systemctl start conserver [Passed]
```
whowutwut: Should these messages  contain `DEBUG` ?

bybai:
Hi @gurevichmark,
I modified some steps. 
1. The first part is how to configure primary HA MN, so I added the section title. After it is configured, admin can use this MN, so I added activate step.
2. The second part is how to configure and activate standby HA MN. In this part, user can use an existed MN or install a new standby MN, so install xcat without VIP on ``xcatmn2`` local disk and switch DB are first steps, and these steps are more flexible, the standby node can be installed any time and no effect to active primary MN node . When we want activate standby MN, it can be configured as HA MN using VIP and shared data, and then activate.

PR #5342, opened 2018-06-27
Fix $kernelpath for Ubuntu 18.04 on ppc64el

neo954:
Fix a typo


neo954:
See https://github.com/xcat2/xcat2-task-management/issues/212

Issue #5294, opened 2018-06-15
Statelite provision hung at ``netbooting`` for rhels 6.10 in PPC64


hu-weihua:
For support rhels 6.10 in PPC64, statelite provision hung at ``netbooting``.

xcat version: 
```
# lsdef -v
lsdef - Version 2.14.2 (git commit 61cbe03b939a44555566894fc0b2e11528da466b, built Wed Jun 13 06:15:51 EDT 2018)
```

Got below information from console.
```
.......
scsi 0:0:1:0: Direct-Access     AIX      VDASD            0001 PQ: 0 ANSI: 3
sd 0:0:1:0: [sda] 113246208 512-byte logical blocks: (57.9 GB/54.0 GiB)
sd 0:0:1:0: [sda] Write Protect is off
sd 0:0:1:0: [sda] Cache data unavailable
sd 0:0:1:0: [sda] Assuming drive cache: write through
sd 0:0:1:0: [sda] Cache data unavailable
sd 0:0:1:0: [sda] Assuming drive cache: write through
 sda: sda1 sda2 sda3
sd 0:0:1:0: [sda] Cache data unavailable
sd 0:0:1:0: [sda] Assuming drive cache: write through
sd 0:0:1:0: [sda] Attached SCSI disk
/sbin/rsyslogd-start: line 22: /tmp/syslog.server: No such file or directory
/sbin/rsyslogd-start: line 23: /tmp/syslog.filters: No such file or directory
/sbin/rsyslogd-start: line 25: /tmp/syslog.conf: No such file or directory
NET: Registered protocol family 10
lo: Disabled Privacy Extensions
dracut: Mounted root filesystem 10.2.1.13:/install/netboot/rhels6.10/ppc64/compute/rootimg
dracut: Loading SELinux policy
dracut: SELinux: Could not open policy file <= /etc/selinux/targeted/policy/policy.24: No such file or directory
No such file or directory
Setting up Statelite
Get to enable localdisk
```

immarvin:
hi @bybai ,would you please take a look at this statelite issue?

thx

bybai:
Hi @hu-weihua, could you give me your environment?

bybai:
It is verified by auto test.

chuckbrazie:
I am out of the office until 07/16/2018.

I am out of the office until  Monday July 16.


Note: This is an automated response to your message  "Re: [xcat2/xcat-core]
[...]

PR #5352, opened 2018-06-28
fix Statelite provision hung at ``netbooting`` for rhels 6.10 in PPC64

bybai:
For https://github.com/xcat2/xcat-core/issues/5294

statelite.pm select wrong rc.statelite.ppc.redhat


immarvin:
i am ok with the change

PR #5396, opened 2018-07-12
Add virtio_net to default network driver list for Ubuntu on ppc64el

neo954:
This will make the default diskless osimage work on KVM environment.


immarvin:
HI @neo954 , please provide the UT status

immarvin:
i am ok with the change

PR #5402, opened 2018-07-13
Draft setup chrony script

neo954:
This is for https://github.com/xcat2/xcat2-task-management/issues/218


zet809:
Let's merge and check the automation result.

PR #5404, opened 2018-07-16
Tweak makentp for chronyd

neo954:
This is for xcat2/xcat2-task-management#218.


neo954:
This pull request rely on pull request #5402.

zet809:
Let's merge and check the automation result.

PR #5405, opened 2018-07-16
Support for chrony in setupntp

neo954:
This is for xcat2/xcat2-task-management#218


neo954:
This pull request rely on pull request #5402.

robin2008:
@neo954   
In fact,  `setupntp` is xCAT interface for NTP service, if we cannot put all codes in this scripts,  it is better use this script to contains the new codes for `chronyd`, and `ntpd` codes in another scripts.

BTW:  I see there is special handling for cumulus switch or statelite nodes,  but I did not see this in your codes for `chronyd`

PR #5435, opened 2018-07-26
Add chrony to default Ubuntu 18.04 package lists

neo954:
I think it is better to add `chrony` to the package list explicitly.


immarvin:
i am ok with the change

Issue #5411, opened 2018-07-18
Unable to execute configonie as non-root user


kkvtran:
/opt/xcat/share/xcat/scripts/configonie --switches $SWITCH --ntp
Failed to connect to nodelist table after retrying 3 times: DBI connect('/etc/xcat/nodelist.sqlite','',...) failed: unable to open database file at /opt/xcat/lib/perl/xCAT/Table.pm line 972.

Can't call method "prepare" on an undefined value at /opt/xcat/lib/perl/xCAT/Table.pm line 998.

I could do 
updatenode $SWITCH -P enablesnmp     
$SWITCH: trying to download postscripts...
$SWITCH: postscripts downloaded successfully
$SWITCH: trying to get mypostscript from 192.168.224.6...
$SWITCH: Wed Jul 18 16:02:13 UTC 2018 Running postscript: enablesnmp
$SWITCH: postscript: enablesnmp exited with code 0
$SWITCH: Running of postscripts has completed.

I chatted with Casandra.  She asked me to open the issue for her to look in to the issue.  Thanks Casandra.

robin2008:
If it wants to operate DB and other privilege commands,  it is better to use it as a plugin.   
Here, you can try to use `lsdef` and parse the data for using as a workaround.

immarvin:
hi @kkvtran , any update on this?

cxhong:
I can recreate this on our test system.  I will add error message if ran by non-root user.  
In the near future, we should move `configonie` code to onie plugin.

cxhong:
if switch has passwordlesss enabled for root user.  the updatenode command to the switch will work too for non-root users.

kkvtran:
Yes, updatenode works for non-root user, i.e. tran58.

immarvin:
hi @kkvtran , if the issue is resolved and you are ok with @cxhong 's fix, please close this ticket, thanks

PR #5438, opened 2018-07-26
Return and log error message  if run by non-root users for config switch commands

cxhong:
for issue #5411
the three config switch command : configonie, configBNT and configMellanox , needs to access xCAT DB to get valid information.  non-root user will hit DB error:
````
[cxhong@briggs01 ~]$ /opt/xcat/share/xcat/scripts/configBNT --switches mid05tor26 --all
Failed to connect to site table after retrying 3 times: DBI connect('/etc/xcat/site.sqlite','',...) failed: unable to open database file at /opt/xcat/lib/perl/xCAT/Table.pm line 972.

Can't call method "prepare" on an undefined value at /opt/xcat/lib/perl/xCAT/Table.pm line 998.
````
For this PR, we will log the error message and exit the command if ran by non-root users:
````
]$ /opt/xcat/share/xcat/scripts/configonie --switches mid05tor26 --ntp
Can't run this command for non-root user
`````

### DIFF
--- a/xCAT-server/share/xcat/scripts/configBNT
+++ b/xCAT-server/share/xcat/scripts/configBNT
@@ -67,6 +67,13 @@ if ($::HELP)
     exit(0);
 }
 
+my $current_usr = getpwuid($>);
+if ($current_usr ne "root")
+{
+    print "Can't run this command for non-root user\n";
+    exit(1);
+}
+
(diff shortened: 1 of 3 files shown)


immarvin:
i am ok with the change

Issue #5381, opened 2018-07-09
 xdsh instance' with error 'Unimplemented: POSIX::tmpnam(): use File::Temp instead at /opt/xcat/lib/perl/xCAT/DSHCLI.pm line 1283  on ubuntu 18.04


hu-weihua:
xcat version 
```
Version 2.14.2 (git commit 4c5a2bbd770cac12f4854cbe0e4bf471daa711b4, built Fri Jul  6 06:17:27 EDT 2018)
```

Got error ``xdsh instance' with error 'Unimplemented: POSIX::tmpnam(): use File::Temp instead at /opt/xcat/lib/perl/xCAT/DSHCLI.pm line 1283``  in ubuntu18.04, this  errror showed up in below 2 csaes.

For ``xdsh`` with option ``-e``
```
RUN:echo "echo 'xdsh_e_filename test' > /tmp/xdsh_e_filename.txt" > /tmp/xdsh_script.sh [Mon Jul  9 01:49:47 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
CHECK:rc == 0   [Pass]

RUN:chmod +x /tmp/xdsh_script.sh [Mon Jul  9 01:49:47 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
CHECK:rc == 0   [Pass]

RUN:xdsh c910f03c09k05 -e /tmp/xdsh_script.sh [Mon Jul  9 01:49:47 2018]
ElapsedTime:0 sec
RETURN rc = 1
OUTPUT:
Error: [c910f03c09k03]: xdsh plugin bug, pid 5006, process description: 'xcatd SSL: xdsh to c910f03c09k05 for root@localhost: xdsh instance' with error 'Unimplemented: POSIX::tmpnam(): use File::Temp instead at /opt/xcat/lib/perl/xCAT/DSHCLI.pm line 1283.
' while trying to fulfill request for the following nodes: c910f03c09k05
CHECK:rc == 0   [Failed]
```

For ``xdsh`` with option ``-E``
```
RUN:echo 'export DSH_FANOUT=8' > /tmp/xdsh.test [Mon Jul  9 01:49:50 2018]
ElapsedTime:0 sec
RETURN rc = 0
OUTPUT:
CHECK:rc == 0   [Pass]

[...]

PR #5440, opened 2018-07-27
Use File::Temp::tmpnam() instead of POSIX::tmpnam()

neo954:
Fix github issue #5381.

Based on POSIX.1-2008 [1], `POSIX::tmpnam` was obsoleted. And in Perl 5.26, it was removed [2].

* [1] http://pubs.opengroup.org/onlinepubs/9699919799/
> Applications should use the `tmpfile()`, `mkstemp()`, or `mkdtemp()` functions instead of the obsolescent `tmpnam()` function.
* [2] https://perldoc.perl.org/perl5260delta.html#POSIX%3a%3atmpnam()-has-been-removed
> The fundamentally unsafe `tmpnam()` interface was deprecated in Perl 5.22 and has now been removed. In its place, you can use, for example, the `File::Temp` interfaces.

### DIFF
--- a/perl-xCAT/xCAT/DSHCLI.pm
+++ b/perl-xCAT/xCAT/DSHCLI.pm
@@ -14,6 +14,7 @@ use File::Basename;
 use locale;
 use strict;
 use File::Path;
+use File::Temp;
 use POSIX;
 use Socket;
 use Getopt::Long;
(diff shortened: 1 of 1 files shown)


neo954:
Here is unit test result.

```
# cat <<EOF >/tmp/foo
> #!/bin/bash
> date
> EOF
# chmod 0755 /tmp/foo
# /tmp/foo
Fri Jul 27 01:52:53 EDT 2018
```
**Before the fix**
```
# xdsh c910f03c17k04 -e /tmp/foo
Error: [c910f03c17k03]: xdsh plugin bug, pid 12428, process description: 'xcatd SSL: xdsh to c910f03c17k04 for root@localhost: xdsh instance' with error 'Unimplemented: POSIX::tmpnam(): use File::Temp instead at /opt/xcat/lib/perl/xCAT/DSHCLI.pm line 1283.
' while trying to fulfill request for the following nodes: c910f03c17k04
```
**After the fix**
```
root@c910f03c17k03:~# xdsh c910f03c17k04 -e /tmp/foo
c910f03c17k04: Fri Jul 27 01:58:38 EDT 2018
```

immarvin:
i am ok with the change

PR #5444, opened 2018-07-27
hardeths: avoid duplicate GATEWAY entries in /etc/sysconfig/network

kcgthb:
When `/etc/sysconfig/network` already contains a `GATEWAY=` line, the `hardeths` postscript will add a second `GATEWAY` entry, which could lead to issues.

This patch avoid this, by replacing the `GATEWAY=` line if it already exists in `/etc/sysconfig/network`, or adding it if it doesn't.

### DIFF
--- a/xCAT/postscripts/hardeths
+++ b/xCAT/postscripts/hardeths
@@ -73,7 +73,9 @@ else
         sed -i "s/HOSTNAME.*/HOSTNAME=`hostname`/" /etc/sysconfig/network
     fi
     if [ ! -z "$defgw" ]; then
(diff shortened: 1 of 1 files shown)


kcgthb:
Add an additional fix to make sure that `$defgw` only contains a single line.

When multiple default gateways are set before running `hardeths`, the `$defgw` variable can contain multiple lines, which generates errors when using it to build configuration files.

zet809:
hi, @bybai , will you pls help to take a look at? Thx!

bybai:
I am ok to merge it. Thanks @kcgthb

PR #5445, opened 2018-07-27
routeop: inconditionally replace the default gateway

kcgthb:
When `/etc/sysconfig/network` already contains a `GATEWAY=` line set with a specific destination, the `routeop replace ` command will *only* replace it if the new default gateway is exactly the same, and will end up adding an extra `GATEWAY=` line if it's different.

That's because the `egrep` command on line 306 matches the full gateway line (`GATEWAY=$gw`), instead of just the part that indicates that it's the default route (`GATEWAY=`)
https://github.com/xcat2/xcat-core/blob/738e5c8814398225949760471bb28982cf2627a5/xCAT/postscripts/routeop#L300-L309

This patch makes sure to:
1. _replace_ whatever default gateway is set, and to only _add_ a new entry if non exists in `/etc/sysconfig/network`.
2. replace an existing default route with the actual new value. `$route1` and `$routedest1` are not defined in the default-route case for Redhat.
3. fix a broken `sed` regex where the bash `$routedest1` was fixed up with the following `.*`

It's only been tested on Redhat and doesn't address the SuSE or Debian cases by lack of testing platforms on my end.

### DIFF
--- a/xCAT/postscripts/routeop
+++ b/xCAT/postscripts/routeop
@@ -258,7 +258,7 @@ replace_persistent_route()
                     echo "Persistent route \"$route\" has been added in $filename."
                 else 
                     # replace it
-                    sed -i -e "s/$routedest1.*/$route1/g" $filename
+                    sed -i -e "s/${routedest1}.*/$route1/g" $filename
                     echo "Persistent route \"$route\" has been replaced in $filename."
                 fi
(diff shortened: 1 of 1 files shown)


bybai:
I am ok to merge it, thanks @kcgthb

PR #5449, opened 2018-08-01
Avoid use command route in configeth, since it is obsoleted

neo954:
This solve some `confignetwork` problems on Ubuntu 18.04. Since `confignetwork` calls `configeth`. And in Ubuntu 18.04, command `route` does not exist by default.


neo954:
See xcat2/xcat2-task-management#247.

PR #5463, opened 2018-08-03
specify build pkg dir when build rpm/deb

xuweibj:
Add env "BUILDDESTDIR", if not null, will copy core tarball to this folder.

### DIFF
--- a/build-ubunturepo
+++ b/build-ubunturepo
@@ -26,7 +26,7 @@
 #                            
 #               LOG=<filename> - provide an LOG file option to redirect some output into log file
 #
-#               PKGDIR=<build_pkg_dir> - Copy build core tarball to PKGDIR if defined
+#               BUILDDESTDIR=<build_tarball_dir> - Copy build core tarball to BUILDDESTDIR if defined
 # For the dependency packages 1. All the xcat dependency deb packages should be uploaded to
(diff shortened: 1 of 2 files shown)


build-ubunturepo
```diff
@@ -400,6 +401,16 @@ __EOF__
     chgrp root $tar_name
     chmod g+w $tar_name
 
+    if [ "$PKGDIR" ]; then
```
neo954: I think it may need to validation user input before use it. In this way, `$PKGDIR` can be anything. It is not safe.

neo954: Say, pass `PKGDIR=/dev/mem` will ruin the system.

build-ubunturepo
```diff
@@ -26,6 +26,7 @@
 #                            
 #               LOG=<filename> - provide an LOG file option to redirect some output into log file
 #
+#               PKGDIR=<build_pkg_dir> - Copy build core tarball to PKGDIR if defined
```
immarvin: this variable should better be `DESTDIR`

build-ubunturepo
immarvin: do we need to remove any existing `$PKGDIR` ?

immarvin:
do you need to put xcat dependencies to the specified path?

immarvin:
i am ok with the change

PR #5470, opened 2018-08-06
SuSE Linux 15 support

jjohnson42:
Diskful install of SLE 15 support


xCAT-server/lib/xcat/plugins/sles.pm
```diff
@@ -1740,16 +1742,16 @@ sub copycd
...
-    my $dsc = <$dinfo>;
-    if ($dsc =~ /x86_64/) {
-        $darch = "x86_64";
-    }
-    if ($dsc =~ /Installer/ and $dsc =~ /SLE-/) {
+        my $dsc = <$dinfo>;
+        if ($dsc =~ /x86_64/) {
+            $darch = "x86_64";
+        }
+        if ($dsc =~ /Installer/ and $dsc =~ /SLE-15/) {
```
bybai: From line 1749 to line 1754, you change "sle" into "sle15", I think this code logic should work for other sles version, if change "sle" into "sle15", it becomes only support "sle15", does not cause problem for old sle version?

jjohnson42: The problem with the sles12 was caused by discnumber scope problem...  The most recent commit fixes the discnumber scope mistake.

In SuSE 15, all enterprise editions share common media, so user is no longer copying 'sles' specifically in.

In fact, the compute template here actually installs SLE HPC, not SLES.

xCAT-server/lib/xcat/plugins/anaconda.pm
```diff
@@ -335,9 +335,15 @@ sub mknetboot
...
-            next; 
+            my @ents = xCAT::TableUtils->get_site_attribute("disablenodesetwarning");
+            my $site_ent = $ents[0];
+            if (!defined($site_ent) || ($site_ent =~ /no/i) || ($site_ent =~ /0/))
+            {
+                if (!defined($::DISABLENODESETWARNING)) {    # set by AAsn.pm
+                    xCAT::MsgUtils->report_node_error($callback, $node, "OS image name must be specified in nodetype.provmethod");
+                    next;
+		}
+            }
```
robin2008: @jjohnson42 the `next` should be here,  xcat is not supporting the legacy node. With the `next`, the following codes will not be executed.

xCAT-server/lib/xcat/plugins/anaconda.pm
```diff
@@ -1064,9 +1070,16 @@ sub mkinstall
...
+
```
robin2008: `next` should be moved out to short stop the below codes.

xCAT-server/lib/xcat/plugins/sles.pm
```diff
@@ -203,8 +203,15 @@ sub mknetboot
...
-            next;
+            my @ents = xCAT::TableUtils->get_site_attribute("disablenodesetwarning");
+            my $site_ent = $ents[0];
+            if (!defined($site_ent) || ($site_ent =~ /no/i) || ($site_ent =~ /0/))
+            {
+                if (!defined($::DISABLENODESETWARNING)) {    # set by AAsn.pm
+                    xCAT::MsgUtils->report_node_error($callback, $node, "OS image name must be specified in nodetype.provmethod");
+                    next;
+                }
+            }
```
robin2008: next should be moved out to short stop the below deprecated codes.

xCAT-server/lib/xcat/plugins/sles.pm
```diff
@@ -849,15 +856,22 @@ sub mkinstall
...
-
+            my @ents = xCAT::TableUtils->get_site_attribute("disablenodesetwarning");
+            my $site_ent = $ents[0];
+            if (!defined($site_ent) || ($site_ent =~ /no/i) || ($site_ent =~ /0/))
+            {
+                if (!defined($::DISABLENODESETWARNING)) {    # set by AAsn.pm
+                    xCAT::MsgUtils->report_node_error($callback, $node, "OS image name must be specified in nodetype.provmethod");
+                    next;
+                }
+            }
```
robin2008: `next` should be moved out to short stop the below deprecated codes.

xCAT-server/share/xcat/install/sles/compute.sle15.tmpl
```diff
@@ -0,0 +1,102 @@
...
+      <!-- XCAT-PARTITION-END -->
+    </partitioning>
+    <add-on>
+        <add_on_products config:type="list">
+            <listentry><media_url>http://XCATNEXTSERVERHOOK/install/sles15/x86_64/2</media_url><product>sle-module-basesystem</product><product_dir>/Module-Basesystem</product_dir></listentry>
+            <listentry><media_url>http://XCATNEXTSERVERHOOK/install/sles15/x86_64/2</media_url><product>sle-module-hpc</product><product_dir>/Module-HPC</product_dir></listentry>
+            <listentry><media_url>http://XCATNEXTSERVERHOOK/install/sles15/x86_64/2</media_url><product>sle-module-server-applications</product><product_dir>/Module-Server-Applications</product_dir></listentry>
+            <listentry><media_url>http://XCATNEXTSERVERHOOK/install/sles15/x86_64/2</media_url><product>sle-module-server-applications</product><product_dir>/Module-Server-Applications</product_dir></listentry>
+            <listentry><media_url>http://XCATNEXTSERVERHOOK/install/sles15/x86_64/2</media_url><product>SLE_HPC</product><product_dir>/Product-HPC</product_dir></listentry>
+        </add_on_products>
```
robin2008: Is the `add-on` section here is mandatory?  And it is not good to hard code `x86_64` in the general template.

xCAT-server/lib/xcat/plugins/anaconda.pm
robin2008: Okay, I see your commit comments. It seems you really want bring the legacy mode back to work with `disablenodesetwarning`

robin2008:
@bybai  Let's merge it first after run a UT for sles12.3 ppc64le copycds and provision.

bybai:
@robin2008, our autotest sles12sp3 diskful installation test case is passed when it is using this "sles.pm", so we can let auto test to test it first.

bybai:
I am OK to merge it.

PR #5474, opened 2018-08-08
set buildall to default when build ubuntu

xuweibj:
If no "BUILDALL" input, set it to 1.


robin2008:
This will make it consistent between building rpm and building deb.

PR #5475, opened 2018-08-08
Fix test case go_xcat_noinput. Ignore the exit code of gpg installation

neo954:
This patch solve the test case problem on Ubuntu 16.04.


neo954:
Thanks @hu-weihua

PR #5478, opened 2018-08-08
Fixes xcat2/xcat-core issue 5476

pawsey-kbuckley:
get_site_attribute() call uses variable where it should
     use the bare string name of the variable

### DIFF
--- a/xCAT-server/lib/xcat/plugins/geninitrd.pm
+++ b/xCAT-server/lib/xcat/plugins/geninitrd.pm
@@ -180,7 +180,7 @@ sub geninitrd {
     my $initrdpath;
     my $kernelpath;
     my $tftpdir = "/tftpboot";
-    my @entries = xCAT::TableUtils->get_site_attribute("$tftpdir");
+    my @entries = xCAT::TableUtils->get_site_attribute("tftpdir");
     my $t_entry = $entries[0];
     if (defined($t_entry)) {
         $tftpdir = $t_entry;


pawsey-kbuckley:
This is for Issue 5476, not 146 as the title suggests.

whowutwut:
@pawsey-kbuckley  Thanks for your interest in the xCAT project! We really appreciate it.  I see this is your first contribution. 

As part of our contribution guidelines, before we are able to review/accept the contribution to the project, can you confirm whether or not you have signed a CLA?  Info on the process is here: http://xcat-docs.readthedocs.io/en/stable/developers/license/index.html

If you think you are covered under a CCLA (Corporate CLA) let me know the corporation and i'll cross reference with our records.  Thanks!

pawsey-kbuckley:
Sadly, I have had to "make enquiries" as to which CLA I need to sign, 
so there will probably be the inevitable delay that often accompanies
being the first to admit to doing something!

Whilst we are waiting for clarification on that, I am interested in the
fact that the Travis CI failed - is that because I'm not listed as someone
who has signed a CLA?

The change itself doesn't prevent the software from running, indeed,
it actually make it work in the situation where one moves the tftpdir 
and installdir out of the root  directory.

Here's hoping you'll be able to apply the commit soon,
Kevin

whowutwut:
No sometimes the CI fails. And we close and reopen to trigger. We are working to fix those CI issues  @hu-weihua do you know the cause of the failure?

zet809:
Used #5531 to replace this PR, so close it, thx very much for the effort @pawsey-kbuckley

PR #5479, opened 2018-08-08
Fix broken xcat-probing for static IP on ubuntu

thilo-maurer:
When running ubuntu and using `xcatprobe xcatmn -i <if>`, it is checked, among other things, that the management node IP address is defined to be static in the file `/etc/network/interfaces`.

This PR fixes a critical typo, that results in a wrong result for this check.


whowutwut:
@thilo-maurer Thanks for your interest in the xCAT project!  We appreciate it. I see you are an IBMer, so you are covered in the CCLA that IBM has signed for this project, excellent.   

@hu-weihua Could you take a look at this PR?

hu-weihua:
@thilo-maurer Thanks for your contribution.

PR #5485, opened 2018-08-09

neo954:
Use contemporary RSA key for signing. And use SHA-256, instead of SHA-1. RHEL5 is archaic.


neo954:
This pull request is something similar to #5316, but for Red Hat.

Issue #3890, opened 2017-09-10
Configfpc fails with "configfpc plugin bug"


ryanbowen:
We are attempting to provision our FPCs in our Lenovo enclosures using the configfpc plugin. Whenever it is ran against our MN interface, it fails partway through the first FPC with the below output:

`Sep 09 14:45:17 xcat13.cluster xcat[4943]: Bad arg length for Socket::inet_ntoa, length is 0, should be 4 at /opt/xcat/lib/perl/xCAT_plugin/ipmi.pm line 935, <$parfd> line 4.`
`Sep 09 14:45:17 xcat13.cluster xcat[4943]: xcatd: configfpc plugin bug, pid 4943, process description: 'xcatd SSL: configfpc for root@localhost: configfpc instance: ipmi instance' with error 'Died at /opt/xcat/sbin/xcatd line 1999, <$parfd> line 4.`

It seems to successfully see the first FPC but fails to configure it, and then outputs "could not change netmask" without continuing further.

immarvin:
hi @zet809 ,would you please take a look at this?

zet809:
hi, @ryanbowen , can you show me the xCAT version you are using and the node definition of the CN? Thx!

zet809:
Move to next release.

immarvin:
no update

PR #5486, opened 2018-08-09
fix configfpc with no gateway

jjohnson42:
Probable cause of:
https://github.com/xcat2/xcat-core/issues/3890


whowutwut:
@zet809  Can you review this one?

zet809:
hi, @jjohnson42 , in #3890, it mentioned `It seems to successfully see the first FPC but fails to configure it, and then outputs "could not change netmask" without continuing further.`, but your fix here is to check option `gateway`, I don't think the #3890 issue is fixed.

zet809:
From xCAT 2.13.6, the error show in #3890 shall be fixed by this PR, so merge this PR.

PR #5492, opened 2018-08-10
MacMap: skip obtaining macmap table from switches without any port connect to switchs and pdus while discovering switch

immarvin:
skip obtaining macmap table from switches without any port connect to switchs and pdus while discovering switch

UT finished on boston02

### DIFF
--- a/perl-xCAT/xCAT/MacMap.pm
+++ b/perl-xCAT/xCAT/MacMap.pm
@@ -529,17 +529,14 @@ sub refresh_table {
             xCAT::MsgUtils->message("S", "xCAT Table error:" . $entry->{node} . "Has missing or invalid switch.switch and/or switch.port fields");
         }
     }
+
     my $children = 0;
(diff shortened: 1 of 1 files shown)


immarvin:
hi @cxhong , would you please take a look at this PR.

I think this PR might has some issue, we need to filter out the switches that:
1) defined in `switch` column of `switch` table
2) without any switch  connected to its ports 

seems this PR might filter  the switches with any port connected to a non-switch-or-pdu node, would you please review this PR and modify the logic if any error?

thanks

immarvin:
oh, seems the logic is ok, confusing....

cxhong:
I run switch-based node discovery and switch discovery, looks fine.   
I will approve this PR

zet809:
Good to me, I agree to merge, thx!

immarvin:
fix https://github.com/xcat2/xcat-core/issues/5527

PR #5496, opened 2018-08-10
bash [[ notation requires bash

thilo-maurer:
When running ubuntu, and using the xCAT postscript `mkresolvconf`, one receives the error message:

```
Fri Aug 10 15:53:44 CEST 2018 [info]: xcat.mypostscript: Running postscript: mkresolvconf
./mkresolvconf: 28: ./mkresolvconf: [[: not found
Fri Aug 10 15:53:44 CEST 2018 [info]: xcat.mypostscript: postscript mkresolvconf return with 0
```

This is because in ubuntu `#!/bin/sh` links to `dash`
```
root@node1:/xcatpost# ll /bin/sh
lrwxrwxrwx 1 root root 4 Aug 10 15:47 /bin/sh -> dash*
```
which does not support the `[[ ... ]]` notation, since this is a bash notation. See https://stackoverflow.com/questions/3427872/whats-the-difference-between-and-in-bash

(On other distros where `sh -> bash` this is not an issue.)

This PR replaces bash brackets by sh brackets, (and leaves the now necessary quotes).

Unfortunately the return code stayed at 0 even with this error, which explains (somewhat) why it has not been discovered yet.

PS: I wonder how many xCAT deployments will break once this is merged. Nevertheless: it is a bug.

### DIFF
--- a/xCAT/postscripts/mkresolvconf
+++ b/xCAT/postscripts/mkresolvconf
@@ -25,7 +25,7 @@ if [ -n "$master" ] && [ -n "$domain" ]; then
 	#logger -t xcat "Created /etc/resolv.conf file on $node."
 	cp $conf_file $conf_file_bak > /dev/null 2>&1
 	echo "search $domain" >$conf_file
-    if [[ "$nameservers" != "" ]]; then
+    if [ "$nameservers" != "" ]; then
         for ns in ${nameservers//,/ }; do
     	    grep -q $ns $conf_file || \
                 echo "nameserver ${ns/<xcatmaster>/$master}" >>$conf_file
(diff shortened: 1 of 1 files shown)


thilo-maurer:
Won't work right now.
```bash
29        for ns in ${nameservers//,/ }; do
30            grep -q $ns $conf_file || \
31               echo "nameserver ${ns/<xcatmaster>/$master}" >>$conf_file
32        done
```
Error:
```
./mkresolvconf: 29: ./mkresolvconf: Bad substitution
```

immarvin:
hi @thilo-maurer , thanks for your PR. Before we can merge you PR, you need to sigh the CLA http://xcat-docs.readthedocs.io/en/stable/developers/license/contributors.html

immarvin:
on this issue, I think maybe changing `/bin/sh` to `/bin/bash` is ok

whowutwut:
@immarvin , @thilo-maurer  Is an IBMer so he is covered under the CCLA...  see https://github.com/xcat2/xcat-core/pull/5479#issuecomment-411426468

immarvin:
i am ok with the change

PR #5512, opened 2018-08-15
Fix xCAT migration test cases on SLES

neo954:
This patch fix the SLES migration test problem.


neo954:
Enclosed please find the unit test result.
[xcattest.output.txt](https://github.com/xcat2/xcat-core/files/2289602/xcattest.output.txt)

hu-weihua:

PR #5513, opened 2018-08-15
Enhance the performance tuning doc

robin2008:
1, add the content for httpd tuning (based on the contents in https://sourceforge.net/p/xcat/wiki/Hints_and_Tips_for_Large_Scale_Clusters/#httpd-servers)
2, adjust the parameters based on field feedback

You can check the UT content on http://10.3.5.10/abc/html/advanced/performance_tuning/index.html


immarvin:
i am ok with the change

PR #5521, opened 2018-08-16
getAllNodeAttribs will query site table too much (#5505)

### LINKED ISSUE #5505: Starting xcatd on service node will cause the database high CPU utilization on a large size cluster
Starting xcatd on service nodes is different with starting xcatd on MN.
To keep the consistent, here it will re-generate the service configuration for its role. For example, dhcp lease, pxe files, and console configuration.

And we found that with a 4K+ nodes cluster, and a service nodes with only 200+ compute nodes, the generation of console settings take about 1 minutes, and the CPU of database will keep higher for a while.

robin2008:
To fix #5505 
 - move `get_site_attribute` out from loop

UT:  
1, make postgresql could log statement, and clean the log file
2, remove /etc/conserver.cf,  and Restart xcatd on SN
3, /etc/conserver.cf could be generated successfully.
4, observer the postgreql daemon on MN, and no CPU high again.
And made the comparison for the query logs:
```
[root@c910f03c05k20 pg_log]# grep execute postgresql-Thu.log.good|wc
    122    1626   13884
[root@c910f03c05k20 pg_log]# grep execute postgresql-Thu.log.bad|wc
  28033  346027 3037395
```
Most of them are query site table.
```
[root@c910f03c05k20 pg_log]# grep execute postgresql-Thu.log.bad|grep site|wc
   9326  214490 1637637
```

### DIFF
--- a/perl-xCAT/xCAT/Table.pm
+++ b/perl-xCAT/xCAT/Table.pm
@@ -3417,6 +3417,12 @@ sub getAllNodeAttribs
     $self->{nodelist}->_build_cache([ 'node', 'groups' ]);
     $self->{_use_cache} = 1;
     $self->{nodelist}->{_use_cache} = 1;
+
+    my @hierarchy_attrs = ();
+    my $hierarchy_field = xCAT::TableUtils->get_site_attribute("hierarchicalattrs");
+    if ($hierarchy_field) {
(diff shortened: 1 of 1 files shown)


immarvin:
i am ok with the change

PR #5130, opened 2018-04-20
Update version_control_inventory.rst

MatthiasMerk:
fixed minor typo in 5.

PR #5530, opened 2018-08-17
To replace PR 5130: Update version_control_inventory.rst

zet809:
To replace PR #5130 

xcat-inventory document mini update.


immarvin:
agree to merge

Issue #5534, opened 2018-08-20
"hierarchicalattrs" cause bad performance when querying nodes attributes


robin2008:
hierarchicalattrs is a global site attributes to declare some nodes attributes aggregated from the node parent groups (the default behavior is to **override**)
This is introduced in https://github.com/xcat2/xcat-core/pull/961

While this also introduces an performance issue because it will query the site table for `hierarchicalattrs` arge amount of times although xCAT had already supported plugin-level site cache (process scope).

Note:  There are two major type of cache for DB data, 
- **node attributes cache**:  it is cached in `xCAT DB Access` process for different tables, it will age in seconds. `xCAT DB Access` process to determine when to cache.
- **site cache**: it is cached in plugin process,  it is initialized for all site attributes in `::XCATSITEVALS` at beginning of  handling an request.

In the expectation,  with **site cache**, there should only be one site table query for each request. But it is not we expected, there are lots of site query, we have already below PR to improve the performance:
https://github.com/xcat2/xcat-core/pull/3593
https://github.com/xcat2/xcat-core/pull/5521

The above improve the performance greatly, but still not hit the real root cause. Still not got the expectation result for site cache.

robin2008:
**Root Cause Analysis**
With lots of UT and the code level analysis of `Table.pm`,  the cause is that those methods (`getNodeAttribs/getNodesAttribs/getAllNodesAttribs`) will use `get_site_attribute` to query site attribute,  and `get_site_attribute` try to see `%::XCATSITEVALS` first, and then query the DB.

As we know, when plugin running, `%::XCATSITEVALS` is cached in plugin process,  but not exists in **xCAT DB Access** process. Unfortunately, the codes to get site attribute for `hierarchicalattrs` is invoked  from  **xCAT DB Access** process, in that no cache at all. 

**Actions**
**- Move the codes to get site attribute for `hierarchicalattrs` from DB process out to plugin process.**

robin2008:
There are still special cases for the `site` cache:
- 1, Service node booting: 
- 2, `XCATBYPASS` mode to run plugin
- 3, Daemons: `SSL Listener`, `UDP listener`, `installer monitor`, etc

In this ticket,  we will cover the item 1, and leave the item 2 and 3 as previous.

For Service node booting,  the things are done in `init_plugin`, instead of a normal request.  Then currently there is no site cache for it. To improve the performance, it is planned to do below:

**Actions:**
- cache site before `init_plugins` and clear it up when things done.

[...]

PR #5535, opened 2018-08-20
Enhancement for using site cache in plugin

robin2008:
To fix the issue #5534 
- using cache from plugin when getNodesAttribs/getNodeAttribs (pass it into DB process from plugin process)
- Site cache is a whole hash, so to use cache when by the hash is there, instead of the specified key is there. So that if no key defined in `site` table, with the cache,  it does not query DB again.
-  When xcatd restarted on service node,  cache site hash before init plugins.
-  For XCATBYPASS mode,  cache site hash before init plugins.

### DIFF
--- a/perl-xCAT/xCAT/Client.pm
+++ b/perl-xCAT/xCAT/Client.pm
@@ -174,6 +174,7 @@ sub submit_request {
 
         # Load plugins from either specified or default dir
         require xCAT::Table;
+        populate_site_hash();
         my %cmd_handlers;
         my @plugins_dirs = split('\:', $ENV{XCATBYPASS});
         if (-d $plugins_dirs[0]) {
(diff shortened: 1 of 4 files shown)


robin2008:
UT 

case 1, to verify if the `hierarchicalattrs` still work.
```
# mkdef -t group test111 postscripts=test111

# chdef fake1 -p groups=test111

# chdef fake1 postscripts=myscripts

lsdef sierra1 -i postscripts,groups
Object name: sierra1
    groups=compute,aw20,service1,test111
    postscripts=syslog,remoteshell,syncfiles,myscripts,confignetwork -s,test111
```
Passed,  the `postscripts` could be added together.

2, Verify if run plugin only query site one time

```
Before run pluign:
# grep execute postgresql-Tue.log |grep site|wc
   4976  114224  865531
# nodels sierra4354
sierra4354

After run pluign:
# grep execute postgresql-Tue.log |grep site|wc
   4977  114243  865687
```
so there is only 1 query (4977 -4976)

immarvin:
agree to merge

PR #5542, opened 2018-08-21
To fix issue that lose activation success log

xuweibj:
when run ``rflash -d``
before modify:
only one activation success log for BMC/PNOR firmware.
after modified:
```
f5u14: Firmware IBM-witherspoon-ibm-OP9_v1.19_1.189 activation successful.
f5u14: Firmware ibm-v2.0-0-r46-0-gbed584c activation successful.
```


bybai:
I am OK to merge it.

PR #5545, opened 2018-08-21
update goconserver quickstart document

zet809:
Update goconserver quickstart document


hu-weihua:
It is ok for me, thanks @zet809

PR #5553, opened 2018-08-24
Make disjoint mode working when service node up (inittime=1)

robin2008:
In disjoint mode,  service node will only handle those nodes managed by this SN.  Here has bugs for this feature when service nodes booting.

And this PR is to make disjoint mode working when service node up (inittime=1), and in the time not run in parallel to avoid access DB too much.


robin2008:
UT:
Define a 4k nodes cluster with service nodes and set the netboot with `petitboot`
```
site.xcatdebugmode=1
site.disjointdhcps=1
site.sharedtftp=0
```

now test for grub2,  
1, make 268 nodes netboot=petitboot, and managed by a SN
2, restart xcatd on this SN
3, check the /var/log/xcat/cluster.log on MN
```
grep total /var/log/xcat/cluster.log
Aug 24 05:32:26 sn01 xcat[20140]: DEBUG  petitboot: [total=268] nodes are ...
```

now test for grub2,  
1, make 100 nodes from `petitboot` to `grub2`
2, restart xcatd on this SN
3, check the /var/log/xcat/cluster.log on MN
```
Aug 24 05:47:49 sn01 xcat[20327]: DEBUG  petitboot: [total=168] nodes are  ...
Aug 24 05:48:01 sn01 xcat[20327]: DEBUG  grub2: [total=100] nodes are  ...
```

OK, we can see that the disjoint mode is working when restart xcatd on service node.

Set `site.disjointdhcps=1`, and restart xcatd on service node again,  we can see all the DB CPU utilization will be about 20% for a longer time. Check the logs:
```
Aug 24 06:02:10 sn01 xcat[20593]: DEBUG  petitboot: [total=4159] nodes are  ...
Aug 24 06:02:33 sn01 xcat[20593]: DEBUG  grub2: [total=100] nodes are  ...
```
Now, it will generate all petitboot for all compute nodes (**total=4159**)

daniceexi:
@robin2008 Xu Bin, could you add some code comments for the design and code logic, otherwise it's hard for later developer to understand.

robin2008:
Added the comments for the flow of `nodeset`,  absolute all branches are handled in `preprocess_request`, and the complex one is when `site.sharedtftp=0` and `site.disjointdhcps=1`

1,   when `sharedtftp=0`, it means all SNs share the same location as MN, so all boot files are generated on MN only.  No actions required SN to be involved.

2,  If `sharedtftp=1`  then All SNs need to generate those boot files too.
- 2.1 check if has service nodes,  if not, run on MN only,  and no need to consider AAsn.pm(inittime)
- 2.2 if not disjoint mode (disjointdhcps=0),  need to run on local and dispatch to All other MN and SNs
         But when run in AAsn.pm,  no needs to dispatch the request to others.  And to avoid accessing DB too much,  not fork multiple process to run nodeset too.

- 2.2 if disjoint mode (disjointdhcps=1), need to run on local and dispatch to parent SNs only and dhcp server which serving the dynamic range.
         And there are still a key difference that non-disjoint,  that is the node range is limited to managed CNs.

xCAT-server/lib/xcat/plugins/AAsn.pm
```diff
@@ -1260,8 +1260,9 @@ sub setup_TFTP
             my $plugins_dir = $::XCATROOT . '/lib/perl/xCAT_plugin';
             foreach my $modname (keys %netmethods)
             {
-                $cmdref->{node} = $netmethods{$modname};
-                $doreq->($cmdref, \&xCAT::Client::handle_response);
+                my $reqcopy = {%$cmdref};
```
zet809: What is the benefit and purpose of using a copy of `cmdref`?

xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -487,11 +492,29 @@ sub preprocess_request {
...
+            if ($mynodeonly == 0 || $ALLFLAG) {
+                if ($inittime) {
+                    $req->{_xcatpreprocessed}->[0] = 1;
+                    return [$req];
+                }
                 return xCAT::Scope->get_broadcast_scope_with_parallel($req, \@sn);
             }
 
+            # 3, Disjoint mode, run on local for owned CNs only and
+            # dispatch to parent SNs of the requesting nodes and `dhcpserver` which serving dynamic range in `networks` table.
```
zet809: For Disjoint mode, I think the command will run on 2 kinds of nodes: the SN/MN that servers the specified CN, the SN/MN that provide dynamic range.

xCAT-server/lib/xcat/plugins/AAsn.pm
robin2008: As each plugin will modify the `req` attribute.

xCAT-server/lib/xcat/plugins/grub2.pm
robin2008: Yes,  but current SN,  it need special handling,  so the logic is to get my owned CNs firsts,  and then dispatch to others (Related parent SNs or MN/SN which has dynamic range).

Issue #5558, opened 2018-08-27
statelite provision failed in rhels7.6-alternate beta ppc64le P9


bybai:
1. xcat version:
```
]# lsxcatd -v
Version 2.14.4 (git commit ec5b18fe13ed46ddb7ee94d39d5a9014cc0bde3a, built Fri Aug 24 06:15:50 EDT 2018)
```
2. debug from console:
```
pre-pivot:/sysroot# cd ../
pre-pivot:/# cat /proc/cmdline
root=nfs:10.6.13.10:/install/netboot/rhels7.6-alternate/ppc64le/compute/rootimg:ro STATEMNT=f6u13k11:/nodedata XCAT=10.6.13.10:3001 NODE=f5u14  LOGSERVER=10.6.13.10  syslog.server=10.6.13.10 syslog.type=rsyslogd syslog.filter=*.*  xcatdebugmode=1 BOOTIF=70:e2:84:14:6e:fb  selinux=0  MNTOPTS= rd.break=pre-pivot
pre-pivot:/# ls
bin  dracut-state.sh  imjournal.state  lib    proc  run   sys	   tmp	var
dev  etc	      init	       lib64  root  sbin  sysroot  usr
pre-pivot:/# exit
logout
[  OK  ] Started Dracut Emergency Shell.
[  248.019361] dracut-pre-pivot[1663]: mount.nfs: Protocol not supported
```

PR #5566, opened 2018-08-29
enhance statelite persistent dir mount options for rh7.6

bybai:
Fix https://github.com/xcat2/xcat-core/issues/5558

After this fix:
This nfsvers options is for  statelite persistent files in `statelite` table.
1. If there are no mount options in `statelite` table, the persistent files default mount options are "nolock,rsize=32768,tcp,timeo=14";
2. If there are mount options in `statelite` table, use the specific value.
3. If there is no mount option in `statelite` table, the nfserver version is different with the nfsclient version, we recommend add correct nfserver nfsvers for `mntopts` in `statelite` table.


immarvin:
agree to merge

immarvin:
hi @bybai , would you please take care the autotest result tomorrow and make sure this PR will not break some older redhat releases? thanks

robin2008:
@bybai   Need a FAQ in RTD.

PR #5582, opened 2018-08-31
Matching for bmcdiscovery

zet809:
The PR is used to enhance bmcdiscover in case a node with same mtm,serial,bmcip found.

The enhance include:
1. build global hash with "$mtm*$serial-$bmcip"==> node
2. when find an bmc/openbmc with mtm and serial, check the hash above, if exists a node, output and return directly. If no, create a tmp node with name "node-$mtm-$serial" or "node-$mac"

The UT result:
```
# bmcdiscover --range 172.11.139.3,172.11.139.5 -z
Warning: [briggs01]: Node bogusnode1 and bogusnode have the same mtms-ip keys: mtm*serial-bmc
Match node mid05tor12cn03 with bmc ip address: 172.11.139.3
Match node mid05tor12cn05 with bmc ip address: 172.11.139.5
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -51,7 +51,7 @@ my $openbmc_pass;
 my $done_num = 0;
 $::P9_WITHERSPOON_MFG_ID     = "42817";
 $::P9_WITHERSPOON_PRODUCT_ID = "16975";
-
+%::VPDHASH = ();
 my %node_in_list = ();
 
 #-------------------------------------------------------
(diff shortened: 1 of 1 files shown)


cxhong:
@zet809 so, if user has ip/mtm/serial defined in the compute node,  xCAT will not allow to run bmcdiscover?    how can they find mac address in this case?

zet809:
Oh, @cxhong , not ip of host, but the bmc ip. 
And it doesn't mean not allow to run bmcdiscover, but bmcdiscover won't generate a new temp node like `node-$mtm-$serial`, it will match to a predefined node directly. Then, the admin can operate the pre-defined node to do hardware discovery.

cxhong:
@zet809 as we discussed, we can update our discovery document to mention how to deal with the scenario that have pre-defined node matched.   
But I think we should still follow xcat-doc, predefine compute node, run bmcdiscover,  create  temp bmc node,  start bmc discover,  if we find there is matched bmcip/mtm/serial, we don't need to send bmc to set up ip address, only add mac address to compute node.

cxhong:
Thanks, @zet809

PR #5586, opened 2018-09-03
move test cases of kit from ubuntu to rhels7 test case bundle

hu-weihua:
Due to the running environment of kit is rhels, not ubuntu, so this PR is used to move test cases of kit from ubuntu to rhels7 test case bundle.


bybai:
I am OK to merge it.

PR #5587, opened 2018-09-03
support nics type Omnipath in confignetwork

bybai:
### The modification include
support nictypes as Omnipath  in confignetwork
We do not have Omnipath device, so let user to verify it.

`confignics` is deprecated, so this PR does not change `confignics`.


robin2008:
Note:  here no special handling for `Omnipath`,  xcat just treat it as a normal IB device.

PR #5602, opened 2018-09-05
Clarify doc for REST API over http

gurevichmark:
### The modification include

* Clarify documentation for REST API over http

* Fix links to a doc file

### DIFF
--- a/docs/source/advanced/restapi/restapi_setup/restapi_setup.rst
+++ b/docs/source/advanced/restapi/restapi_setup/restapi_setup.rst
@@ -37,7 +37,7 @@ To improve the security between the REST API clients and server, enabling the se
     sudo apache2ctl -t -D DUMP_MODULES | grep ssl
     apt-get install libjson-perl
 
-.. note:: If use of non-secure HTTP protocol is required, edit ``/etc/httpd/conf.d/xcat-ws.conf`` for RHEL or ``/etc/apache2/conf.d/xcat-ws.conf`` for others and change `RewriteEngine On` to `RewriteEngine Off`, then restart httpd or apache.
+.. note:: If use of non-secure HTTP protocol is required, edit ``/etc/httpd/conf.d/xcat-ws.conf`` for RHEL or ``/etc/apache2/conf.d/xcat-ws.conf`` for others and change ``RewriteEngine On`` to ``RewriteEngine Off``, then restart httpd or apache.
 
 Enable the Certificate of HTTPs Server (Optional)
 =================================================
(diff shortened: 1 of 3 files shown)


bybai:
I agree with the changes.

docs/source/advanced/restapi/restapi_setup/restapi_setup.rst
```diff
@@ -37,6 +37,8 @@ To improve the security between the REST API client and server, enabling the HTT
     sudo apache2ctl -t -D DUMP_MODULES | grep ssl
     apt-get install libjson-perl
 
+.. note:: If use of HTTP service is needed, edit `/etc/httpd/conf.d/xcat-ws.conf` for RHEL or `/etc/apache2/conf.d/xcat-ws.conf` for others and change `RewriteEngine On` to `RewriteEngine Off`, then restart httpd or apache.
```
whowutwut: Suggest:  "If use of non secure HTTP protocol is required.... " 

Did you mean for those files to be formatted as code? If so, this requires double back ticks instead of single back tick

docs/source/advanced/restapi/restapi_setup/restapi_setup.rst
```diff
@@ -3,7 +3,7 @@ Enable the HTTPS service for REST API
 
 To improve the security between the REST API client and server, enabling the HTTPS service on the xCAT MN is recommended. And the REST API client should use the 'https' to access web server instead of the 'http'.
```
whowutwut: I think we should change this line...

> To improve the security between the REST API clients and server, enabling the secure transfer protocol (https) is the default configuration.

whowutwut: The above heading of this section, it's not really "service" but protocol? 

> Enable the HTTPS protocol for Rest API

docs/source/advanced/restapi/restapi_setup/restapi_setup.rst
```diff
@@ -37,6 +37,8 @@ To improve the security between the REST API client and server, enabling the HTT
     sudo apache2ctl -t -D DUMP_MODULES | grep ssl
     apt-get install libjson-perl
 
+.. note:: If use of HTTP service is needed, edit `/etc/httpd/conf.d/xcat-ws.conf` for RHEL or `/etc/apache2/conf.d/xcat-ws.conf` for others and change `RewriteEngine On` to `RewriteEngine Off`, then restart httpd or apache.
+
```
whowutwut: Down below at line 63, the format of those steps is inconsistent with the rest of the doc, can we fix it?

whowutwut:
Another comment: https://xcat-docs.readthedocs.io/en/stable/advanced/restapi/restapi_setup/restapi_setup.html#extend-the-timeout-of-web-server 

This is really a trouble shooting, we could simplify the doc to a simple attention line...  

.. attention::  Some operations like ``create osimage`` (i.e copycds) may require a longer time to complete and may result in a "504 Gateway Timeout" Error.  To avoid this, modify the ``httpd.conf`` file and extend the timeout to a larger value: ``Timeout: 600``

gurevichmark:
@whowutwut New commit for suggested changes. Did not make suggested changes to the section "The steps to configure the certificate for https server:".

PR #5607, opened 2018-09-06
fix issue 5604, delete extra / of rootimgdir

xuweibj:
### The modification include

delete extra '/' in rootimgdir.

### The UT result
```
# cat test.sh
rootimg="/aaa/bbb/cccc//"
while (echo $rootimg | grep "/$")
do
    rootimg=${rootimg%/*}
done
echo $rootimg

# sh test.sh
/aaa/bbb/cccc
```


hu-weihua:

PR #5608, opened 2018-09-06
reuse MN_basic.bundle as CI test scope to fast test

hu-weihua:
After discussion, we need more faster CI test and remove some known issues test case out, so adopt ``MN_basic.bundle`` as ci test scope again.


zet809:
Thx @hu-weihua

PR #5621, opened 2018-09-10
SLE15 stateless support

jjohnson42:
For review and comment, here are commits that comprise SLE15 support.  It carries over more than it needs to in all likelihood, but represents a functional set of changes.


xCAT-server/share/xcat/netboot/sle/service.sle15.postinstall
```diff
@@ -0,0 +1,68 @@
+#!/bin/sh
+#-- Do not remove following line if you want to make use of CVS version tracking
```
bybai: I found you create new directory `/opt/xcat/share/xcat/netboot/sle` for sle15 stateless, and most files contents under it are the same with the old ones in  `/opt/xcat/share/xcat/netboot/sles/`, for example, 
`/opt/xcat/share/xcat/netboot/sle/service.sle15.postinstall` content is the same with `/opt/xcat/share/xcat/netboot/sles/service.sles12.postinstall`. I know sle15 osname is sle, the directory `/opt/xcat/share/xcat/netboot/sle` is needed.
Why do you create a symbolic link for the files with the same content? 
Or why do you create a symbolic link for directory `sles` and `sle`? 
I think the same file have 2 copies in different directory are hard to maintain.

cxhong:
we decided to break symbolic link between `sle->sles` to move sle 15 and above support forward.
so, we will merge this PR, thanks, @jjohnson42

PR #5645, opened 2018-09-19
otherpkgs: pipe results to logger to avoid "Argument list too long" errors

kcgthb:
### The modification include

Piping `$results` to `logger` instead of passing it as an argument avoids that limitation.

### DIFF
--- a/xCAT/postscripts/otherpkgs
+++ b/xCAT/postscripts/otherpkgs
@@ -818,7 +818,7 @@ EOF`
         if [ $R -ne 0 ]; then
               RETURNVAL=$R 
         fi
-        logger -p local4.info -t xcat "$result"
+        echo "$result" | logger -p local4.info -t xcat
         if [ $VERBOSE  ]; then
           echo "$result"
         fi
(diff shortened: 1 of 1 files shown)


robin2008:
@kcgthb Thanks for that.  And as there are lots of scripts to use `logger` in the existing method, do you know what kinds of `$result` will cause the `Argument list too long`.  
If the case is common, xcat  need to refine other places too.

kcgthb:
Hi @robin2008, 
Ah that's a good point indeed. I noticed the error in `otherpkgs`
https://github.com/xcat2/xcat-core/blob/4164dd2dc9501fa0b17f6f812bbec7b6118e707a/xCAT/postscripts/otherpkgs#L931

This is the output of `yum -y install $repo_pkg`. We install quite a few packages in `otherpkgs`, with many dependencies, so the output of the `yum install` command could be quite long.

kcgthb:
Hi @robin2008 

That's a good point indeed.

I noticed the error in `otherpks`: https://github.com/xcat2/xcat-core/blob/4164dd2dc9501fa0b17f6f812bbec7b6118e707a/xCAT/postscripts/otherpkgs#L926-L931

That's the output of `yum -y install $repo_pkgs`. We install quite a few packages in `otherpkgs`, and they come with many dependencies, so the resulting yum output can be quite substantial.

bybai:
Hi @kcgthb,

Since I am working on log tag enhancement support for xCAT, I enhanced `otherpkgs` at the same time in pull request https://github.com/xcat2/xcat-core/pull/5743. It should resolve your current issue, so I will close this PR, thanks.

PR #5654, opened 2018-09-21
Doc enhancements to localdisk instructions

gurevichmark:
localdisk instructions enhancements


robin2008:
@gurevichmark  could you set a live new documet environment, then we can check the final result.

gurevichmark:
@robin2008 The final result can be viewed at http://10.6.29.1/5654/localdisk.html at the bottom of the page.

PR #5659, opened 2018-09-24
Update cuda_power9_setup for CUDA 10.0

Obihoernchen:
CUDA 10.0 ships with the following file:
```
# cat /usr/lib/modprobe.d/nvidia.conf
blacklist nouveau

# Uncomment the following line to enable kernel modesetting support.
# There is NO graphical framebuffer (like OSS drivers) at the moment; this is
# only for Wayland. For Gnome, you also require an EGLStream build of Mutter.
#options nvidia-drm modeset=1
```
Previously (CUDA 9) this file was called: `/usr/lib/modprobe.d/blacklist-nouveau.conf`

We don't want to overwrite `blacklist nouveau`, so use another file for our custom module options.

### DIFF
--- a/xCAT/postscripts/cuda_power9_setup
+++ b/xCAT/postscripts/cuda_power9_setup
@@ -27,7 +27,7 @@ then
 	sed -i /SUBSYSTEM==\"memory\"/d ${IMG_ROOTIMGDIR}/lib/udev/rules.d/40-redhat.rules
 fi
 
-echo 'options nvidia NVreg_EnableStreamMemOPs=1 NVreg_RegistryDwords="PeerMappingOverride=1"' >${IMG_ROOTIMGDIR}/usr/lib/modprobe.d/nvidia.conf
+echo 'options nvidia NVreg_EnableStreamMemOPs=1 NVreg_RegistryDwords="PeerMappingOverride=1"' >>${IMG_ROOTIMGDIR}/usr/lib/modprobe.d/nvidia.conf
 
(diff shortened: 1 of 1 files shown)


xCAT/postscripts/cuda_power9_setup
```diff
@@ -27,7 +27,7 @@ then
 	sed -i /SUBSYSTEM==\"memory\"/d ${IMG_ROOTIMGDIR}/lib/udev/rules.d/40-redhat.rules
 fi
 
-echo 'options nvidia NVreg_EnableStreamMemOPs=1 NVreg_RegistryDwords="PeerMappingOverride=1"' >${IMG_ROOTIMGDIR}/usr/lib/modprobe.d/nvidia.conf
+echo 'options nvidia NVreg_EnableStreamMemOPs=1 NVreg_RegistryDwords="PeerMappingOverride=1"' >>${IMG_ROOTIMGDIR}/usr/lib/modprobe.d/nvidia.conf
```
whowutwut: What happens if we run this multiple times against the same image?

Obihoernchen: yeah it piles up. Might be better to use a seperate file like `/usr/lib/modprobe.d/nvidia_options.conf`?

neo954:
I think this patch will affect CUDA 9 support as well.

Obihoernchen:
This is now fixed with 03bd6f3

Issue #5661, opened 2018-09-24
Updatenode -f didn't work if servicenode attribute is ip address for compute node definition


cxhong:
I defined servicenode attributes as ip address instead of hostname
````
# lsdef mid08tor03cn01 -i servicenode
Object name: mid08tor03cn01
    servicenode=172.20.254.2
````
the `updatenode` command failed: 
````
# updatenode p9 -f
172.20.254.2 is not responding. No command will be issued to this host.
File synchronization failed for service nodes: "172.20.254.2"
````
looking the code,  `updatenode -f` calls `pping` and `pping` seems only accept the hostname not ip address
````
# pping 172.20.254.2
sn02: ping
172.20.254.2: noping
````

gurevichmark:
Yes, I saw something like this before with `lstree` command. I think we only support SN defined with names, not IP addresses: https://github.com/xcat2/xcat-core/issues/3230#issuecomment-306967476

cxhong:
from pping man page,  it only take noderange, not ip address. so that's working as design.

cxhong:
The definition of `servicenode`  
````
servicenode:	A comma separated list of node names (as known by the management node) that provides most services for this node. 
````
but xCAT will not block if user set servicenode as ip address

robin2008:
@cxhong   Any special consideration for setting an IP address?  if not,  better to follow the document to set the name instead.

cxhong:
I don't have special consideration,  just in case if user use ip address instead of hostname to define servicenode, that will cause issue and error message didn't mention ip address is invalid

PR #5664, opened 2018-09-25
Modify the output message if failed pping_hostname() 

cxhong:
on issue #5661, if servicenode attribute is ip address, `updatenode -f` will not sync files to service node. That will cause syncfile list failed during the node deployment.  
`updatenode -f` will call `pping_hostname()` sub routine and this routine only takes hostname,  so always convert host to the hostname before call pping_hostname(). 

unit test:
```
[root@boston02 xCAT]# lsdef mid08tor03cn01 -i groups,servicenode
Object name: mid08tor03cn01
    groups=all,p9,cn_regex,p9up,witherspoon,dd20
    servicenode=172.20.254.2
[root@boston02 xCAT]# updatenode dd20 -f
File synchronization has completed for service nodes: "172.20.254.2"
````

### DIFF
--- a/perl-xCAT/xCAT/DSHCLI.pm
+++ b/perl-xCAT/xCAT/DSHCLI.pm
@@ -3420,6 +3420,8 @@ sub verify_targets
     {
 
         my $hostname = $$resolved_targets{$user_target}{'hostname'};
+        #use hostname in case the host is ip address
+        my $hostname = xCAT::NetworkUtils->gethostname($host);
         push @ping_list, $hostname;
     }
 
--- a/perl-xCAT/xCAT/DSHCLI.pm
+++ b/perl-xCAT/xCAT/DSHCLI.pm
@@ -3419,9 +3419,7 @@ sub verify_targets
     foreach my $user_target (keys(%$resolved_targets))
     {
 
-        my $host = $$resolved_targets{$user_target}{'hostname'};
-        #use hostname in case the host is ip address
-        my $hostname = xCAT::NetworkUtils->gethostname($host);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by robin2008** [comment]
Do we really need to support this attribute as an IP address?  I'm just wondering if there are still any other issues.

```
       servicenode
                 A comma separated list of node names (as known by the management node) that provides most services for this node. The first service node
                 on the list that is accessible will be used.  The 2nd node on the list is generally considered to be the backup service node for this
                 node when running commands like snmove.
```

**Comment by zet809** [comment]
I agree with @robin2008 , servicenode is always used to specify which service node a command shall be dispatched to, so it shall be an xCAT node object. 
Why they specify the servicenode to be an IP address?

**Comment by cxhong** [comment]
yes, user should follow the doc, but if someone defined servicenode as ip address, xCAT didn't catch/block this,  and the error message didn't indicated  this also.
```
# updatenode p9 -f
172.20.254.2 is not responding. No command will be issued to this host.
File synchronization failed for service nodes: "172.20.254.2"
```

**Comment by robin2008** [comment]
`pping` can only support the xcat node range, so not all resolvable hostname could be used.
```
[root@briggs01 ~]# pping mid05tor12cn05-bmc
Warning: Invalid nodes in noderange:mid05tor12cn05-bmc
[root@briggs01 ~]# nslookup mid05tor12cn05-bmc
Server:		172.11.253.27
Address:	172.11.253.27#53

Name:	mid05tor12cn05-bmc.pok.stglabs.ibm.com
Address: 172.11.139.5

[root@briggs01 ~]# nodels mid05tor12cn05-bmc
Error: Invalid nodes and/or groups in noderange: mid05tor12cn05-bmc
```

PR #5666, opened 2018-09-26
Replace 5623: Correct typo in xCAT-vlan spec file comment

zet809:
### The modification include

xCAT-vlan spec file

### The UT result
`##The UT output##`


hu-weihua:
It is ok for me, thanks @zet809

PR #5672, opened 2018-09-27
[Docu] Remove "Install NVIDIA Management Library"

Obihoernchen:
You can't install NVML anymore. NVML has been part of CUDA since CUDA version 8. See: https://developer.nvidia.com/gpu-deployment-kit

> The GPU Deployment Kit is part of the CUDA Toolkit, in versions 8 and later, and is no longer available as a separate download.

The package is called: `cuda-nvml-dev-...`


whowutwut:
@Obihoernchen There's a reference to `nvml.rst` in `source/advanced/gpu/nvidia/repo/index.rst` that needs to also be removed so that when we build the doc, it doesn't find an empty reference.

whowutwut:
@Obihoernchen If  you  get a chance,  need to remove that last  line in file: https://raw.githubusercontent.com/xcat2/xcat-core/master/docs/source/advanced/gpu/nvidia/repo/index.rst 

That references the  file you deleted in the PR.  Thanks!

Issue #5252, opened 2018-05-25
syncfiles fails during stateful deployment but updatenode -F works


kcgthb:
Hi, 

I'm seeing some resurgence of #4455 / #4484 when deploying a node in hierarchical mode. The weird thing is that the `synfiles` postscript (silently) fails during deployment, but `updatenode <node> -F` works perfectly fine once the node has been deployed.

Here are some logs. I'm deploying `sh-ln01` from a service node (`sh-hn02`, `10.10.0.2`). The management node is `sh-hn01` (`10.10.0.1`). xCAT version is `xCAT-2.14-snap201804190715`

During deployment, here's what is logged from `syncfiles`:
```
+ logger -t xcat -p local4.info 'Performing syncfiles postscript'
++ uname
+ osname=Linux
+ xcatpostdir=/xcatpost
+ logger -t xcat -p local4.info './syncfiles: the OS name = Linux'
+ quit=no
+ count=5
+ returncode=0
+ '[' no = no ']'
+ '[' Linux = Linux ']'
++ /xcatpost/startsyncfiles.awk
+ returncode=0
+ '[' 0 -eq 0 ']'
+ quit=yes
+ let count=count-1
+ '[' yes = no ']'
+ '[' 0 -eq 0 ']'
+ logger -t xcat -p local4.info './syncfiles: Perform Syncing File action successfully'
```
so it appears to have worked, but none of the files in the synclist is actually copied over, and the `sshd` daemon on the deployed node logs this:
```
May 25 09:26:24 sh-ln01 xcat: Performing syncfiles postscript
May 25 09:26:24 sh-ln01 xcatpostinit1: ++ uname
May 25 09:26:24 sh-ln01 xcatpostinit1: + osname=Linux
May 25 09:26:24 sh-ln01 xcatpostinit1: + xcatpostdir=/xcatpost
May 25 09:26:24 sh-ln01 xcatpostinit1: + logger -t xcat -p local4.info './syncfiles: the OS name = Linux'
May 25 09:26:24 sh-ln01 xcat: ./syncfiles: the OS name = Linux
May 25 09:26:24 sh-ln01 xcatpostinit1: + quit=no
May 25 09:26:24 sh-ln01 xcatpostinit1: + count=5
May 25 09:26:24 sh-ln01 xcatpostinit1: + returncode=0
May 25 09:26:24 sh-ln01 xcatpostinit1: + '[' no = no ']'
[...]

Issue #4455, opened 2017-12-06
updatenode -F not working with hierarchy


mattaezell:
Updatenode works fine to nodes managed by the management node, but it's failing on nodes managed by service nodes.  It looks like it's trying to use the hostname of the management node as the username

```
# updatenode a02n01 -F
a02n01: Received disconnect from 10.10.0.119 port 22:2: Too many authentication failures
a02n01: Authentication failed.
a02n01: Received disconnect from 10.10.0.119 port 22:2: Too many authentication failures
a02n01: Authentication failed.
a02n01: rsync: connection unexpectedly closed (0 bytes received so far) [sender]
a02n01: rsync error: unexplained error (code 255) at io.c(605) [sender=3.0.9]
a02n01: Received disconnect from 10.10.0.119 port 22:2: Too many authentication failures
a02n01: Authentication failed.
a02n01: rsync: connection unexpectedly closed (0 bytes received so far) [sender]
a02n01: rsync error: unexplained error (code 255) at io.c(605) [sender=3.0.9]
File synchronization has completed for nodes.
```

Looking on that compute node:
```
Dec 06 21:30:35 a02n01 sshd[103855]: Invalid user mgmt1.myfqdn from 10.10.0.11 port 50912
Dec 06 21:30:35 a02n01 sshd[103855]: input_userauth_request: invalid user mgmt1.myfqdn [preauth]
Dec 06 21:30:35 a02n01 sshd[103855]: Postponed keyboard-interactive for invalid user mgmt1.myfqdn from 10.10.0.11 port 50912 ssh2 [preauth]
Dec 06 21:30:35 a02n01 sshd[103857]: pam_unix(sshd:auth): check pass; user unknown
Dec 06 21:30:35 a02n01 sshd[103857]: pam_unix(sshd:auth): authentication failure; logname= uid=0 euid=0 tty=ssh ruser= rhost=sn1.myfqdn
Dec 06 21:30:38 a02n01 sshd[103855]: error: PAM: Authentication failure for illegal user mgmt1.myfqdn from sn1.myfqdn
Dec 06 21:30:38 a02n01 sshd[103855]: Failed keyboard-interactive/pam for invalid user mgmt1.myfqdn from 10.10.0.11 port 50912 ssh2
Dec 06 21:30:38 a02n01 sshd[103855]: error: maximum authentication attempts exceeded for invalid user mgmt1.myfqdn from 10.10.0.11 port 50912 ssh2 [preauth]
[...]

PR #5680, opened 2018-10-04
syncfiles: fix for sync lists generated with the wrong username

kcgthb:
### The modification include

_##Check `dest-user` existence_

Since user accounts used to synchronize files or to run xCAT commands need to exist on both the MN and the SNs, this PR just adds a check to make sure that the username passed as `$$config{'dest-user'}` is actually a valid username, and not the MN's hostname, as described in #4455 and #5252.

### The UT result
`syncfiles` works


kcgthb:
Great, thanks!

PR #5681, opened 2018-10-05
Add two disc ID for rhels8.0

cxhong:
### The PR is to fix issue _no issue opened
copycds couldn't find discname for RHEL-8.0-20180920.2-ppc64le-dvd1.iso
````
# copycds RHEL-8.0-20180920.2-ppc64le-dvd1.iso
Copying media to /install//ppc64le
Media copy operation successful
Error when updating the osimage tables: Unable to find /opt/xcat/share/xcat/install directory for
# copycds -i RHEL-8.0-20180920.2-ppc64le-dvd1.iso
OS Image:/root/RHEL-8.0-20180920.2-ppc64le-dvd1.iso
DISTNAME:
ARCH:ppc64le
````
The output from debug messages:
```
xCAT: Allowing copycds -i RHEL-8.0-20180920.2-ppc64le-dvd1.iso for root from localhost
DESCRIPTS --- 8.0
INFO - Could not find ID=1537463022.540933 in the discinfo database for OS=8.0 ARCH=ppc64le NUM=
````
The other one worked fine:
````
# copycds -i RHEL-8.0-20180531.2-BaseOS-ppc64le-dvd1.iso
OS Image:/root/RHEL-8.0-20180531.2-BaseOS-ppc64le-dvd1.iso
DISTNAME:rhels8.0
ARCH:ppc64le
DISCNO:1

xCAT: Allowing copycds -i RHEL-8.0-20180531.2-BaseOS-ppc64le-dvd1.iso for root from localhost
DESCRIPTS --- Red Hat Enterprise Linux 8.0
INFO - Could not find ID=1527763142.409770 in the discinfo database for OS=Red Hat Enterprise Linux 8.0 ARCH=ppc64le NUM=1
````

### The modification include
Add two disc ID for rhels8.0 iso file

### The UT result
`##The UT output##`
```
# copycds -i RHEL-8.0-20180920.2-ppc64le-dvd1.iso
OS Image:/root/RHEL-8.0-20180920.2-ppc64le-dvd1.iso
DISTNAME:rhels8.0
ARCH:ppc64le
```
```
xCAT: Allowing copycds -n rhels8.0 RHEL-8.0-20180920.2-ppc64le-dvd1.iso for root from localhost
DESCRIPTS --- 8.0
INFO - detected distname=rhels8.0, arch=ppc64le
```

PR #5684, opened 2018-10-09
Revert "Add two disc ID for rhels8.0"

zet809:
Reverts xcat2/xcat-core#5681

The discinfo added in PR #5681 is useless for us, our code can deal with the .discinfo correctly if it is in the good format, so revert it with this PR.

For more information, pls reference: 
https://github.com/xcat2/xcat-core/pull/5681#issuecomment-428064764


zet809:
Close this one as it is OK to add disc id to DB.

PR #5685, opened 2018-10-09
delete all switchdiscover test cases

hu-weihua:
The PR is to delete all ``switchdiscover`` test cases to resolve switch access attempts  in lab.


zet809:
OK to me, we shall make a label for this PR in case we want to recovery switchdiscover cases.

PR #5686, opened 2018-10-09
fix xcatprobe xcatmn issues 5677&5678

xuweibj:
### The modification include

_##allow chrony for ntp_
If ``chronyc`` is installed on MN/SN, use ``chronyc`` to check whether ntp service is ready.
If not, still use ``ntpq`` to check.

_##allow rsync dir_
Based on doc https://xcat-docs.readthedocs.io/en/stable/advanced/hierarchy/define_service_nodes.html, tftpdir & installdir could be mounted and rsynced.
If ``sharedtftp=1``, check tftpdir whether mounted.
If ``installloc="/install"``, check installdir whether mounted.
If ``sharedtftp`` not 1, check tftpdir on SN whether same with MN's.
If ``installloc=``, check installdir on SN whether same with MN's.

_##allow name of SN contain other character_
grep ``\w\S+`` as node name.

### The UT result
1. If ``ntpd`` is down, but ``chronyd`` is running:
```
....
[mn]: Checking NTP service is configured...                                                                       [ OK ]
[mn]: Checking rsyslog service is configured...                                                                   [ OK ]
[mn]: Checking firewall is disabled...                                                                            [ OK ]
```

2. When ``installloc=``, and there are differences between MN and SN installdir.
```
[c910f03c17-k17]: Checking 'passwd' table is configured...                                                        [ OK ]
[c910f03c17-k17]: Checking important directories(installdir,tftpdir) are configured...                            [WARN]
[c910f03c17-k17]: installdir '/install' is not the same with MN's, may be error
[c910f03c17-k17]: Checking SELinux is disabled...                                                                 [ OK ]
```

### DIFF
--- a/xCAT-probe/lib/perl/probe_utils.pm
+++ b/xCAT-probe/lib/perl/probe_utils.pm
@@ -524,8 +524,29 @@ sub parse_node_range {
 
 #------------------------------------------
 sub is_ntp_ready{
+    my $check_cmd = shift;
+    $check_cmd = shift if (($check_cmd) && ($check_cmd =~ /probe_utils/));
     my $errormsg_ref = shift;
-    $errormsg_ref= shift if (($errormsg_ref) && ($errormsg_ref =~ /probe_utils/));
+
+    if ($check_cmd eq "chronyc") {
+        my $chronycoutput = 'chronyc tracking 2>&1';
+        if ($?) {
+            if ($chronycoutput =~ /Cannot talk to daemon/) {
+                $$errormsg_ref = "chronyd service is not running! Please setup ntp in current node";
+                return 0;
+            }
+            $$errormsg_ref = "command 'chronyc tracking' failed, could not get status of ntp service";
+            return 0;
+        }
+        if ($chronycoutput =~ /Leap status     : (.+)/) {
+            my $status = $1;
+            if ($status eq "Not synchronised") {
+                $$errormsg_ref = "chronyd did not synchronize.";
+                return 0;
+            }
+        }
+        return 1;
+    }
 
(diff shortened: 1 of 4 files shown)


robin2008:
@xuweibj   One PR is only for one issue.

xCAT-probe/lib/perl/probe_utils.pm
```diff
@@ -524,8 +524,29 @@ sub parse_node_range {
...
+        }
+        if ($chronycoutput =~ /Leap status     : (.+)/) {
+            my $status = $1;
+            if ($status eq "Not synchronised") {
+                $$errormsg_ref = "chronyd did not synchronize.";
+                return 0;
+            }
+        }
+        return 1;
+    }
```
hu-weihua: Could you refine the structure of this function to let it better to read and maintain? thanks

xCAT-probe/subcmds/xcatmn
```diff
@@ -1281,23 +1316,23 @@ while ($hierarchy_instance->read_reply(\%reply_cache)) {
                 #print ">>>$reply_cache{$servers}->[$_]<<<\n";
                 #For cases like below:
                 #c910f02c04p04: [ok]     :All xCAT daemons are running
-                if ($reply_cache{$servers}->[$_] =~ /^(\w+)\s*:\s*(\[\w+\]\s*):\s*(.*)/) {
+                if ($reply_cache{$servers}->[$_] =~ /^(\w\S+)\s*:\s*(\[\w+\]\s*):\s*(.*)/) {
```
hu-weihua: The node name  can not include specific character, adding ``\S+`` is not correct.

hu-weihua:

PR #5699, opened 2018-10-12
Using port 22 for unprivileged node ping 

robin2008:
### The modification include

when doing pping,  we will force to use port 22 for unprivileged node ping (TCP only), it could work when the node is under firewall as 22 is normally opened by firewall.

Note:  default is 80/443

### The UT result
1, start firewalld on sn02
`xdsh sn02 systemctl start firewalld`

Before apply the patch:
```
[root@boston02 ~]# pping sn02
sn02: noping
```
After apply the patch:
```
[root@boston02 xcat]# pping sn02
sn02: ping
[root@boston02 xcat]#nodestat sn02
sn02: sshd
```


immarvin:
i am ok with the change

Issue #5500, opened 2018-08-11
postscript mlnxofed_ib_install failes for ubuntu16.04 + OFED-4.4-1.0.00


thilo-maurer:
Installing Ubuntu 16.04 (diskful) and using xCAT-provided postscript `mlnxofed_ib_install.v2` to install Mellanox OFED fails. Using method as described at https://xcat-docs.readthedocs.io/en/stable/advanced/networks/infiniband/index.html.

Ubuntu installs successfully, also OFED installs successfully.
However the last lines executed by postscript `mlnxofed_ib_install`
```
#force openibd load all modules in need, restart again
sleep 1
service openibd restart
```
fail with
```
Unloading HCA driver:                                      [  OK  ]
Module mlx4_core does not exist, skipping...               [FAILED]
Module mlx4_ib does not exist, skipping...                 [FAILED]
Module mlx4_core does not exist, skipping...               [FAILED]
Module mlx4_en does not exist, skipping...                 [FAILED]
Module mlx5_core does not exist, skipping...               [FAILED]
Module mlx5_ib does not exist, skipping...                 [FAILED]
Module mlx5_fpga_tools does not exist, skipping...         [FAILED]
Module ib_umad does not exist, skipping...                 [FAILED]
Module ib_uverbs does not exist, skipping...               [FAILED]
Module ib_ipoib does not exist, skipping...                [FAILED]
Loading HCA driver and Access Layer:                       [  OK  ]
Module rdma_cm does not exist, skipping...                 [FAILED]
Module ib_ucm does not exist, skipping...                  [FAILED]
Module rdma_ucm does not exist, skipping...                [FAILED]
```

This happens still inside the ubuntu installer:
```
root       1235  0.0  0.0   3392  2496 hvc0     Ss+  17:55   0:00 /bin/sh /sbin/debian-installer
root       1259  0.0  0.0  27328 25280 hvc0     S+   17:55   0:09  \_ debconf -o d-i /usr/bin/main-menu
root       1265  0.0  0.0   3456  2944 hvc0     S+   17:55   0:00      \_ /usr/bin/main-menu
root      49227  0.0  0.0   3328  2112 hvc0     S+   18:57   0:00          \_ udpkg --configure --force-configure finish-install
root      49228  0.0  0.0   3392  2688 hvc0     S+   18:57   0:00              \_ /bin/sh /var/lib/dpkg/info/finish-install.postinst configure
[...]

PR #5723, opened 2018-10-22
Do not restart service openibd in the installation stage

neo954:
The PR is to fix issue *#5500*

Do not restart service openibd in the installation stage

### DIFF
--- a/xCAT-server/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2
+++ b/xCAT-server/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2
@@ -331,11 +331,13 @@ EOF
         env -i "PATH=${PATH}" /tmp/ofed/mountpoint/mlnxofedinstall "${MLNXOFED_OPTS[@]}"
 
         #force openibd load all modules in need, restart again
-        sleep 1
-        service openibd restart
-        if [ "$?" != "0" ] ;then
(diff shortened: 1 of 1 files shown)


xCAT-server/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install.v2
```diff
@@ -331,11 +331,13 @@ EOF
         env -i "PATH=${PATH}" /tmp/ofed/mountpoint/mlnxofedinstall "${MLNXOFED_OPTS[@]}"
 
         #force openibd load all modules in need, restart again
-        sleep 1
-        service openibd restart
-        if [ "$?" != "0" ] ;then
-            echo "[Error] service openibd restart failed."
-            exit 1
+        if [ "$NODESETSTATE" != "install" ]; then
```
robin2008: It is better to use equal for the specified case which need to start `openibd`

neo954: The purpose is to avoid start `openibd` when the script is in the `install` stage. Which is inside operating system installer.

With the `!=`, it allow the script to be used with `updatenode`.

hu-weihua:
I have no more comments, @robin2008 feel free to merge if you are ok  for these change.

neo954:
Hello @robin2008, the code is revised as you suggested.

PR #5732, opened 2018-10-25
Add reventlog to hw management list

gurevichmark:
This PR:

* Add `reventlog` to the list of hardware management commands at https://xcat-docs.readthedocs.io/en/latest/guides/admin-guides/manage_clusters/ppc64le/management/index.html

* Remove OpenBMC section at https://xcat-docs.readthedocs.io/en/stable/references/coral/known_issues/index.html. That section seems to be for the older version of xCAT and FW levels that would no longer apply to CORAL installation.


zet809:
Good to me, thx @gurevichmark

PR #5744, opened 2018-10-30
Have client fallback to IPv4 if IPv6 doesn't work

jjohnson42:
If IPv6 support is added to a system after xcat is running,
then trying to reach 'localhost' may end up being '::1' instead
of '127.0.0.1'.  This will be fine after xCAT restarts from before
IPv6 support being available, but until that time the daemon is
IPv4 only.  Have Client try IPv4 if IPv6 fails to accommodate this
situation.


immarvin:
hi @jjohnson42 , why there are 12 commits in this PR? is it possible to refine the timeline?

jjohnson42:
I don't quite understand the weird merge commits, I'll try to cherry pick the one commit with changes

PR #5756, opened 2018-11-01
add labels for all cases without labels

tingtli:
### The PR is to task 341#
add labels for all cases without labels


tingtli:
Close this one for code conflict.

PR #5758, opened 2018-11-02
Use postscript file efibootmgr to check http service instead of syslog

xuweibj:
### The modification include

_##Use efibootmgr file_

### The UT result
```
[mn]: Checking important directories(installdir,tftpdir) are configured...                                        [ OK ]
[mn]: Checking SELinux is disabled...                                                                             [ OK ]
[mn]: Checking HTTP service is configured...                                                                      [ OK ]
[mn]: Checking TFTP service is configured...                                                                      [ OK ]
[mn]: Checking DNS service is configured...                                                                       [ OK ]
```


robin2008:
@xuweibj What's the problem this PR for?  I think you need to mention it in your description what's the problem and why you fix it like this.

PR #5760, opened 2018-11-02
A more real world partitioning scheme for RHEL8

neo954:
### The modification include

_##A more real world partitioning scheme_

* A separate `/boot/efi` partition, size 256MiB, if needed.
* A separate `/boot` partition, size 512MiB.
* All the remain disk space will be used as LVM physical volume. Inside which, one volume group will be created.
* Root file system will be inside the LVM volume group, minimal 4096MiB, maximum 8192MiB.
* `/var` file system will be inside the LVM volume group, minimal 2048MiB, maximum 8192MiB.
* `/tmp` file system will be inside the LVM volume group, minimal 1024MiB, maximum 4096MiB.
* `/home` file system will be inside the LVM volume group, with 10% of the remain available space of the volume group, maximum 10240MiB.
* A swap partition will be created inside the LVM volume group with the Red Hat recommended size.

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/pre.rhels8
+++ b/xCAT-server/share/xcat/install/scripts/pre.rhels8
@@ -171,14 +171,17 @@ case "$(uname -m)" in
 esac
 if [ -d /sys/firmware/efi ]
 then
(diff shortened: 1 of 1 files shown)


immarvin:
agree to merge to cover by test

PR #5770, opened 2018-11-07
Make travis script go ahead when it failed to get PR information after max retries

robin2008:
### The modification include

- Make travis continue to finish the real test when it failed to get the PR information from github

- and redo the non-finished PR format checking at the end if all testing is passed.


hu-weihua:
It is ok for me, thanks  @robin2008 .

hu-weihua:
It is ok for me, thanks  @robin2008 .

gurevichmark:
@robin2008 The comment on line 116 still says "return error".

robin2008:
Yes, it still return error, all non-zero code is error,  but here I doesn't comment it very clearly as it is an special error.

PR #5777, opened 2018-11-08
several issues on site.httpport in otherpks and ospkgs

immarvin:
### The PR is to fix issues found in autotest on site.httpport


robin2008:
The codes are fine, merge it.

PR #5778, opened 2018-11-08
RHEL 8 Beta diskfull compute node support

neo954:
### The PR is for task _#xxx_ or to implement feature xcat2/xcat2-task-management#378

_##Feature description##_

### The content of the PR:
* [x] _##Revise the kickstart template based on RHEL8 default settings
* [x] _##Remove the dirty hack for RHEL8 alpha

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/post.rhels8
+++ b/xCAT-server/share/xcat/install/scripts/post.rhels8
@@ -18,6 +18,3 @@ do
         sed -i 's/ONBOOT=no/ONBOOT=yes/' "$i"
     fi
 done
-
-# Dirty hack for RHEL8 Alpha
-chkconfig network on


xCAT-server/share/xcat/install/rh/compute.rhels8.tmpl
```diff
@@ -16,10 +12,14 @@ lang en_US.UTF-8
...
+# Not run the Setup Agent on first boot
+firstboot --disable
+# Do not configure the X Window System
+skipx
 # System services
-services --disabled="NetworkManager"
+services --enabled="chronyd"
 # System timezone
-timezone #TABLE:site:key=timezone:value# --isUtc
+timezone #TABLE:site:key=timezone:value# --isUtc --ntpservers=#TABLE:noderes:$NODE:xcatmaster#
```
robin2008: Why we set ntp here at this time?  
`TABLE:noderes` will cause the performance issue, it is better leave it in `setupntp` postscripts.

Issue #5785, opened 2018-11-09
Failed to download network interface to Cumulus Switches


cxhong:
IST test system got few cumulus switches from WSC cluster  with static ip address already set up.  when adm ran `updatenode f1tor configinterface`, it return 0 but `/etc/network/interface.d/xCAT.intf file was not updated.

````
trying to download postscripts...
postscripts downloaded successfully
trying to get mypostscript from 10.66.3.2...
Thu Feb  8 15:09:39 EST 2001 [info]: xcat.mypostscript: Running postscript: configinterface
Thu Feb  8 15:09:40 EST 2001 [info]: xcat.mypostscript: postscript configinterface return with 0
````

cxhong:
The PR https://github.com/xcat2/xcat-core/pull/4834  changed to get  xCAT Master ip address from dhcpd leases file instead of `mypostscript`.  for this issue,  since the  ip address is manually changed to static and there are no dhcpd process involved.  The dhcpd lease file on the switch still have the old information, with old xCAT Master ip address, which cause the switch can't reach to old Master and can't able to download the interface files.
Also, need to better handle the error message and return code for the configinterface postscripts

cxhong:
code merged, will close the issue.

PR #5786, opened 2018-11-09
Modify configinterface postscripts to handle the MASTER IP address

cxhong:
The PR is to fix issue  #5785 

The modification include
1) get Master from mypostscript first, if not , get from dhcpd lease file.  ping the ip address first before call issue `curl` command
2) better handle the error message and return code

### The UT result
````
Fri Feb  9 09:18:15 EST 2001 [info]: xcat.mypostscript: Running postscript: configinterface
ERROR: The xCAT Master ip address 10.5.100.250 is not reachable
Fri Feb  9 09:18:26 EST 2001 [info]: xcat.mypostscript: postscript configinterface return with 1

````

### DIFF
--- a/xCAT/postscripts/configinterface
+++ b/xCAT/postscripts/configinterface
@@ -11,13 +11,14 @@ xcat_intf="/etc/network/interfaces.d/xCAT.intf"
 
 if [ -f /xcatpost/mypostscript ]; then
     MASTER=`grep '^MASTER_IP=' /xcatpost/mypostscript |cut -d= -f2|sed s/\'//g`
-else
-    MASTER=$(cat /var/lib/dhcp/dhclient.eth0.leases|sed -n 's/.*cumulus-provision-url.*http:\/\+\([^\/]\+\)\/.*/\1/p'|tail -1)
 fi
 
 if [ -z "$MASTER" ]; then
-    echo "xCAT Master unset! Cannot download interface description"
-    exit 2
+    MASTER=$(cat /var/lib/dhcp/dhclient.eth0.leases|sed -n 's/.*cumulus-provision-url.*http:\/\+\([^\/]\+\)\/.*/\1/p'|tail -1)
+    if [ -z "$MASTER" ]; then
+        echo "xCAT Master unset! Cannot download interface description"
+        exit 2
+    fi
 fi
 
 #Validate if this IP is reachable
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: xCAT/postscripts/configinterface [hunk 1]
```diff
@@ -9,29 +9,55 @@ fi
 
 xcat_intf="/etc/network/interfaces.d/xCAT.intf"
 
-MASTER=$(cat /var/lib/dhcp/dhclient.eth0.leases|sed -n 's/.*cumulus-provision-url.*http:\/\+\([^\/]\+\)\/.*/\1/p'|tail -1)
+if [ -f /xcatpost/mypostscript ]; then
```
immarvin: should the logic be 
```
if [ -f /xcatpost/mypostscript ]; then
    MASTER=`grep '^MASTER_IP=' /xcatpost/mypostscript |cut -d= -f2|sed s/\'//g 2>/dev/null`
fi

if [ -z "$MASTER" ]; then
    MASTER=$(cat /var/lib/dhcp/dhclient.eth0.leases|sed -n 's/.*cumulus-provision-url.*http:\/\+\([^\/]\+\)\/.*/\1/p'|tail -1 2>/dev/null)
fi
```

xCAT/postscripts/configinterface
```diff
@@ -9,29 +9,55 @@ fi
...
 UPDATED=0
+DOWNLOADED=0
 
 for name in $NODE ${GROUP//,/ } default; do
     curl -s -o $TMPINT -f http://${MASTER}/install/custom/sw_os/cumulus/interface/$name
     if [ -f $TMPINT ]; then
+        DOWNLOADED=1
         if ! diff $TMPINT $xcat_intf > /dev/null; then
+            cp $xcat_intf $ORIGFILE
             mv $TMPINT $xcat_intf
```
immarvin: should `mv -f` used here to avoid confirmation before overwriting?

immarvin: should any existing `$ORIGFILE` file be removed before `cp`?

xCAT/postscripts/configinterface
cxhong: changed

xCAT/postscripts/configinterface
cxhong: modified

immarvin:
ok with the change

PR #5793, opened 2018-11-12
do task396,  update xcat-inventory for the template and patten match part one.

tingtli:


tingtli:
code conflict , close this pr

PR #5798, opened 2018-11-13
For the newly installed xCAT-core, set site.xcatsslversion to TLSv1_2 as the default setting

neo954:
### The modification include

_##set `site.xcatsslversion` to `TLSv1_2` as the default setting_


xCAT-server/sbin/xcatconfig
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+
```
robin2008: why to remove this shebang?

robin2008:
This change will only be in 2.15,  and @neo954   could you append the UT result:
for fresh installation, we can leverage CI,
but for upgrade case,  need a UT result for it.

neo954:
The base branch was changed to `xcat2/rhels8`.

immarvin:
agree to merge

PR #5808, opened 2018-11-16
when no ipmitool-xcat tool, report error before go ahead the real discovery

robin2008:
### The PR is to fix issue about bmcdiscover from mail list
1, remove /opt/xcat/bin/ipmitool-xcat by accident
2, run `bmcdiscover`

### The modification include
Check if the tool is there, quit with error as other missing tool (`nmap`).

### The UT result
```
mv /opt/xcat/bin/ipmitool-xcat /opt/xcat/bin/ipmitool-xcat.bak
 bmcdiscover -s nmap --range 1.1.1.1
Error: 	There is no ipmitool-xcat in /opt/xcat/bin/, make sure that package ipmitool-xcat is installed successfully.
mv /opt/xcat/bin/ipmitool-xcat.bak /opt/xcat/bin/ipmitool-xcat
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
+++ b/xCAT-server/lib/xcat/plugins/bmcdiscover.pm
@@ -322,18 +322,25 @@ sub bmcdiscovery_processargs {
         ######################################
         # check if there is nmap or not
         ######################################
-        if (-x '/usr/bin/nmap')
-        {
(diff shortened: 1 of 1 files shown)


robin2008:
Not required test case for this As it is only for a very rare case which the tool deleted by accident.

xuweibj:
Looks good to me.

PR #5812, opened 2018-11-19
Update --allow instead of --force-yes for CI

zet809:
### The PR is to fix CI building issue

```
          'W: --force-yes is deprecated, use one of the options starting with --allow instead.',
          'W: --force-yes is deprecated, use one of the options starting with --allow instead.',
          'E: Sub-process /usr/bin/dpkg returned an error code (1)'
```

### DIFF
--- a/travis.pl
+++ b/travis.pl
@@ -339,7 +339,7 @@ sub install_xcat{
         }
     }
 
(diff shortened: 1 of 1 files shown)


zet809:
Seems not related to the dpkg error, so close this PR.

PR #5814, opened 2018-11-19
modify all callback data type to array in plugins [Do Not Merge]

xuweibj:
### The modification include

_##item1_

_##item2_

### The UT result
`TBD`


robin2008:
Not required this,  response->{data} could support both ARRAY and STRING.

PR #5823, opened 2018-11-20
upgrade to travis xenial env

robin2008:
### The PR is to resolve https://github.com/xcat2/xcat2-task-management/issues/402 upgrade travis environment to ubuntu 16.04 

Travis already support `xenial`, and xCAT's MN has drop off 14.04.  So it is better to migrate to 16.04 for our CI.

#### Including:
- refine the travis.yml to use `xenial` distribution, and update dependency
- modify travis.pl accordingly to catch the changes, also catch the changes in build scripts
- refine codes which does not pass the syntax check
- remove conserver related cases as it does not do good itself and impact CI result
- hack the 32 bit netmask to make dhcp testcase working.

UT:  
Just included in this Travis testing itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 os: linux
-dist: trusty
+dist: xenial
 sudo: required
 before_install:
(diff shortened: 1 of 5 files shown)


bybai:
I am ok to merge it.

Issue #5795, opened 2018-11-12
Spelling errors of `invoke_provison`, should be `invoke_provision`


neo954:
The spelling of `invoke_provison`should be `invoke_provision`.

Please correct the spelling error.

```
[root@c910f03c01p09 ~]# xcattest -l labelinfo | grep invoke
Full_installation_flat_docker                                             os=linux,others,docker,invoke_provison
import_osimage_with_environments_in_yaml                                  os=linux,others,xcat_inventory,invoke_provison
reg_linux_diskfull_installation_flat_postscripts_failed                   os=linux,others,postscripts,invoke_provison
reg_linux_diskless_installation_flat_postscripts_failed                   os=linux,others,postscripts,invoke_provison
invoke_provison          4
```

tingtli:
https://github.com/xcat2/xcat-core/pull/5842

Issue #5804, opened 2018-11-15
Some known issue about label of test case


hu-weihua:
There are below known issues about label of test case:

*  typo label name ``invoke_provison``
related issue : https://github.com/xcat2/xcat-core/issues/5795

```
[root@c910f02c02p15 testcase]# XCATTEST_CASEDIR="/xcat-core/xCAT-test/autotest/testcase" /tmp/hwh/xcattest -l  -s "invoke_provison"
set_xcatdebugmode_off_ubuntule_diskful
set_xcatdebugmode_value_two_rhels_diskful_flat_p8_le
set_xcatdebugmode_value_one_ubuntule_diskful
set_xcatdebugmode_value_one_rhels_diskful_flat_p8_le
reg_set_xcatdebugmode_value_one_rhels_diskful_flat
set_xcatdebugmode_off_rhels_diskful_flat_p8_le
set_xcatdebugmode_value_one_sles_diskful_flat_p8_le
reg_set_xcatdebugmode_off_rhels_diskful_flat
set_xcatdebugmode_value_one_ubuntux_vm_diskful
reg_set_xcatdebugmode_off_sles_diskful_flat
reg_set_xcatdebugmode_value_two_rhels_diskful_flat
set_xcatdebugmode_off_ubuntux_vm_diskful
set_xcatdebugmode_off_sles_diskful_flat_p8_le
reg_set_xcatdebugmode_value_one_sles_diskful_flat
reg_linux_diskless_installation_flat_postscripts_failed
Full_installation_flat_docker
import_osimage_with_environments_in_yaml
reg_linux_diskfull_installation_flat_postscripts_failed
```

* The directory name ``xcat-inventory`` is no a valid label name, due to we have added directory name as default label, so  ``xcat-inventory`` will result in error. Should change to ``xcat_inventory``. ``go-xcat`` is the same issue.

* There are 2 value of arch have same meaning. ``arch=x86``  and  ``arch=x86_64``, should just use one.

* The directory name ``snpool`` and label ``sn_pool`` just need use one.

hu-weihua:
Have been verified. close this issue.

PR #5842, opened 2018-11-28
update the error label

tingtli:
### The PR is to fix issue https://github.com/xcat2/xcat-core/issues/5795 and https://github.com/xcat2/xcat-core/issues/5804


hu-weihua:

PR #5844, opened 2018-11-28
update arch for x86_64

tingtli:
### The PR is to fix issue https://github.com/xcat2/xcat-core/issues/5804


hu-weihua:

PR #5846, opened 2018-11-28
Fix issue 5150: tabch and tabdump commands can not deal with invalid attribute names

zet809:
### The modification include

checking the specified column name before doing SQL query.

### The UT result
Before fix:
```
root@c910f03c05k10:~# tabch networkxyz='abcnetwork' networks.comments="ABC"
root@c910f03c05k10:~# echo $?
0
root@c910f03c05k10:~# tabdump -w 'netmaskxyz==255.0.0.0' networks
Error: [c910f03c05k10]: tabutils plugin bug, pid 6433, process description: 'xcatd SSL: tabdump for root@localhost: tabutils instance' with error 'Can't use string ("*XCATBUGDETECTED*:Can't call met"...) as a HASH ref while "strict refs" in use at /opt/xcat/lib/perl/xCAT_plugin/tabutils.pm line 1301.
'
root@c910f03c05k10:~# echo $?
1
```
With fix
```
root@c910f03c05k10:~# tabch networkxyz='abcnetwork' networks.comments="ABC"
Error: [c910f03c05k10]: Table "networks" have no column called "networkxyz"
root@c910f03c05k10:~# echo $?
1
root@c910f03c05k10:~# tabdump -w 'netmaskxyz==255.0.0.0' networks
Error: [c910f03c05k10]: Table "networks" have no column called "netmaskxyz"
root@c910f03c05k10:~# echo $?
1
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/tabutils.pm
+++ b/xCAT-server/lib/xcat/plugins/tabutils.pm
@@ -730,6 +730,11 @@ sub tabdump
             $recs = $tabh->getAllEntries("all");
         } else {           # filter entries
             foreach my $w (@{$OPTW}) {    # get each attr=val
+                my ($k, $v) = split('=', $w, 2);
+                unless (grep /$k/, @{ $xCAT::Schema::tabspec{$table}->{cols} }) {
+                    $cb->({ error => "Table \"$table\" have no column called \"$k\"", errorcode => [1] });
+                    return;
+                }
                 push @attrarray, $w;
             }
             @ents = $tabh->getAllAttribsWhere(\@attrarray, 'ALL');
@@ -2369,6 +2374,16 @@ sub tabch {
                     return 1;
 
                 }
+            }
+            my $err_found = 0;
+            for my $k (keys %keyhash) {
+                unless (grep /$k/, @{ $xCAT::Schema::tabspec{$table}->{cols} }) {
+                    $callback->({ error => "Table \"$table\" have no column called \"$k\"", errorcode => [1] });
+                    $err_found = 1;
+                }
+            }
+            if ($err_found) {
+                return 1;
             }
 
             #splice assignment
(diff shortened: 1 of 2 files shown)


xCAT-server/lib/xcat/plugins/tabutils.pm
```diff
@@ -730,6 +730,11 @@ sub tabdump
             $recs = $tabh->getAllEntries("all");
         } else {           # filter entries
             foreach my $w (@{$OPTW}) {    # get each attr=val
+                my ($k, $v) = split('=', $w, 2);
+                unless (grep /$k/, @{ $xCAT::Schema::tabspec{$table}->{cols} }) {
+                    $cb->({ error => "Table \"$table\" have no column called \"$k\"", errorcode => [1] });
```
gurevichmark: I think a better message is "No column xxx in table yyy"

xCAT-server/lib/xcat/plugins/tabutils.pm
```diff
@@ -2370,6 +2375,16 @@ sub tabch {
 
                 }
             }
+            my $err_found = 0;
+            for my $k (keys %keyhash) {
+                unless (grep /$k/, @{ $xCAT::Schema::tabspec{$table}->{cols} }) {
+                    $callback->({ error => "Table \"$table\" have no column called \"$k\"", errorcode => [1] });
```
gurevichmark: Same message comment as above.

xCAT-server/lib/xcat/plugins/tabutils.pm
```diff
@@ -562,6 +562,32 @@ sub tabrestore
     }
 }
 
+sub get_key_from_optw
```
robin2008: Do we have some method to make `buildWhereClause` in Table.pm invoke a common parsing the where conditions,  now it seem we need to modify two totally different codes if we want to add more support operates.

zet809:
hi, @robin2008  and @gurevichmark , modified as your comments. Pls help to review. Thx!

xCAT-server/lib/xcat/plugins/tabutils.pm
zet809: Good point, maybe a solution is add another parameter for this subroutine if only key is needed.

PR #5853, opened 2018-11-30
modify xcat-inventory cases with ssh [Do not merge]

xuweibj:
For ``ssh`` replace test

### The modification include

_##item1_

_##item2_

### The UT result
`##The UT output##`


xuweibj:
Will fix xcat-inventory cases in another pr, not change ``ssh`` cmd.

PR #5879, opened 2018-12-07
update xcat inventory cases for template and format change

tingtli:
### The PR is to do task https://github.com/xcat2/xcat2-task-management/issues/409
The change are following
```
1. change cases label from xcat_inventory to inventory_ci
2. some attributes key words format changes in command lines
3.  With the format change and attributes change, the check output return code changes.
4.  change case  import_validation_node_network_info_primarynic_switchport to  import_validation_node_network_info_connections_switchport for attribute primarynic changes to connections
5. site table change for case xcat_inventory_try_to_export_all_type_is_site_default_format

6. template changes from schema version 1 to schema version 2 for yaml and json files.

```


zet809:
Good to me, I agree to merge, thx @tingtli

Issue #5894, opened 2018-12-14
[FVT]:update lsdef usage.


tingtli:
In manpage, we have following description for lsdef
```
2.  To get a list of all the objects that have been defined.

            lsdef
              OR
            lsdef -a
```
But with our code, lsdef could only display the node definition, lsdef -a could display all the objects.

Please update the code or if the  code is correct , please update the document.

zet809:
hi, @tingtli , @gurevichmark have modified lsdef manpage, pls verify and close this issue. Thx!

tingtli:
verified , thanks!
```
2.  To get a list of all the objects that have been defined.

            lsdef -a
```

PR #5907, opened 2018-12-18
lsdef man page improvements

gurevichmark:
### The PR is to fix issue #5894


whowutwut:
@gurevichmark The Issue number should be not in the header and in the description to allow github to make the cross reference

PR #5925, opened 2018-12-29
add test cases for xcat-inventory backend

immarvin:


immarvin:
UT result:

https://github.com/xcat2/xcat-inventory/files/2716247/xcattest.log.20181229094600.txt

PR #5927, opened 2019-01-04
update case lsvmkvmhost to check the correct output

tingtli:
### The PR is to fix case lsvm_kvmhost
for following discussion

```
c910f03c09k11: Could not get any information about specified object
CHECK:output =~ c910f03c09k11: \s*\w.	[Pass]
Based on the habit of our vm naming, i think we probably could rewrite the case as follows

check:output=~$$CN:\s*$$CN\w.
```


gurevichmark:
@tingtli 
I do not think `check:output=~$$CN:\s*$$CN\w.` will work, because the two `$$CN` nodenames should be different, first one is the kvmhost name the second is the VM name.

Perhaps this expression will work `$$CN:\s*\w+(\r\n|\n|\r)`

Maybe you already know it, but I found this cool site where you can test regular expressions: https://regex101.com/r/cU5lC2/1

tingtli:
@gurevichmark hi Mark, sorry for the late response. I've updated the case, please take a look. Thanks!

PR #5930, opened 2019-01-08
Make `makedhcp` working on RHEL 8 xCAT management node

neo954:
### The PR is to make `makedhcp` working on RHEL 8 xCAT management node.

_##Feature description##_

### The content of the PR:
* [x] _##Patch `dhcp.pm` for the operating system version detection_
* [x] _##Patch `makenetworks` to make it ignore the `lo` interface_

### The UT result
```
# lsdef -t network
Could not find any object definitions to display.
# makenetworks
# lsdef -t network
10_0_0_0-255_0_0_0  (network)
```
```
# makedhcp -n
Renamed existing dhcp configuration file to  /etc/dhcp/dhcpd.conf.xcatbak

Warning: [c910f03c01p19]: No dynamic range specified for 10.0.0.0. If hardware discovery is being used, a dynamic range is required.
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -1757,7 +1757,7 @@ sub process_request
                 my $os_ver = $os;
                 $os_ver =~ s/[^0-9.^0-9]//g;
                 if (($os =~ /sles/i && $os_ver >= 11) ||
-                    ($os =~ /rhels/i && $os_ver >= 7)) {
+                    ($os =~ /rhels?/i && $os_ver >= 7)) {
 
                     $dhcpd_key = "DHCPD_INTERFACE";
                     if ($usingipv6 and $dhcpver eq "dhcpd6") {
@@ -1825,7 +1825,10 @@ sub process_request
         if ($usingipv6) {
 
             # sles11.3 and rhels7 has dhcpd and dhcpd6 config in the dhcp file
-            if ($os =~ /sles/i || $os =~ /rhels7/i) {
+            my $os_ver = $os;
+            $os_ver =~ s/[^0-9.^0-9]//g;
+            if (($os =~ /sles/i && $os_ver >= 11) ||
+                ($os =~ /rhels?/i && $os_ver >= 7)) {
                 if ($missingfiles{dhcpd}) {
                     $callback->({ error => ["The file /etc/sysconfig/dhcpd doesn't exist, check the dhcp server"] });
                 }
--- a/xCAT-server/lib/xcat/plugins/networks.pm
+++ b/xCAT-server/lib/xcat/plugins/networks.pm
@@ -391,7 +391,8 @@ sub donets
         foreach (@ip6table)
         {
             my @ent = split /\s+/, $_;
-            if ($ent[0] eq 'fe80::/64' or $ent[0] eq 'unreachable' or $ent[1] eq 'via') {
+            if ($ent[0] eq 'fe80::/64' or $ent[0] eq 'unreachable' or
+                $ent[1] eq 'via' or $ent[2] eq 'lo') {
 
                 #Do not contemplate link-local, unreachable, or gatewayed networks further
                 #DHCPv6 relay will be manually entered into networks as was the case for IPv4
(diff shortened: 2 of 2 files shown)


xCAT-server/lib/xcat/plugins/dhcp.pm
```diff
@@ -1757,7 +1757,7 @@ sub process_request
                 my $os_ver = $os;
                 $os_ver =~ s/[^0-9.^0-9]//g;
                 if (($os =~ /sles/i && $os_ver >= 11) ||
-                    ($os =~ /rhels/i && $os_ver >= 7)) {
+                    ($os =~ /rhels?/i && $os_ver >= 7)) {
```
robin2008: Would you explain why you need the change for the reviewer to understanding it?
For example,  for RHEL8, the `$os` is xxx, while for RHEL7, its foo,  this change is caused by xxx

And why you did not change `$os` to make OS name detection consistent, but modify the match logic here?

xCAT-server/lib/xcat/plugins/networks.pm
```diff
@@ -391,7 +391,8 @@ sub donets
         foreach (@ip6table)
         {
             my @ent = split /\s+/, $_;
-            if ($ent[0] eq 'fe80::/64' or $ent[0] eq 'unreachable' or $ent[1] eq 'via') {
+            if ($ent[0] eq 'fe80::/64' or $ent[0] eq 'unreachable' or
+                $ent[1] eq 'via' or $ent[2] eq 'lo') {
 
                 #Do not contemplate link-local, unreachable, or gatewayed networks further
                 #DHCPv6 relay will be manually entered into networks as was the case for IPv4
```
robin2008: Could you append the output of `/sbin/ip -6 route` here for reference?
And you should update the comment message accordingly to reflect your change.

xCAT-server/lib/xcat/plugins/dhcp.pm
neo954: @robin2008,

It is a good question. To explain it, I have a long story.

The `$os` variable is getting from `xCAT::Utils->osver()`. And in function `xCAT::Utils->osver()`, it read file `/etc/os-release` to check the name of the operating system. See related codes below.

https://github.com/xcat2/xcat-core/blob/b53100a59f7e40966fa4cce922547619ac2f23fe/perl-xCAT/xCAT/Utils.pm#L2283-L2285

https://github.com/xcat2/xcat-core/blob/b53100a59f7e40966fa4cce922547619ac2f23fe/perl-xCAT/xCAT/Utils.pm#L2315-L2317

And see the related part of content in file `/etc/os-release` below.

> On RHEL 7.5
> ```
> # grep ^NAME= /etc/os-release
> NAME="Red Hat Enterprise Linux Server"
> ```

> On RHEL 8 Snapshot 2
> ```
> # grep ^NAME= /etc/os-release
> NAME="Red Hat Enterprise Linux"
> ```

In short, on RHEL 7.5, `$os = xCAT::Utils->osver();` gets `rhels7.5`; and on RHEL 8 Snapshot 2, `$os = xCAT::Utils->osver();` gets `rhel8`. That is due to there is no `Server` substring in the `NAME` variable. I believe the `Server` substring will come back when RHEL 8 GA. But for now, I have to handle this change.

Thus, I choose to fix this minor problem by change the matching pattern here. Instead of change the codes in file `Utils.pm`. I think the code logic of `rhel` and `rhels` matching part is reasonable.

xCAT-server/lib/xcat/plugins/networks.pm
neo954: On RHEL 8, there is a route entry of subnet `::1` for loopback interface. It is reasonable. But I have to filter it out here.

> On RHEL 7.5
> ```
> # ip -6 route
> fe80::/64 dev eth0 proto kernel metric 256 pref medium
> fe80::/64 dev eth1 proto kernel metric 256 pref medium
> ```

> On RHEL 8 Snapshot 2
> ```
> # ip -6 route
> ::1 dev lo proto kernel metric 256 pref medium
> fe80::/64 dev env3 proto kernel metric 100 pref medium
> fe80::/64 dev env4 proto kernel metric 101 pref medium
> ```

xCAT-server/lib/xcat/plugins/dhcp.pm
robin2008: ok, it's fine.

PR #5934, opened 2019-01-09
Fix openbmcpy build issue

zet809:
### The modification include

xCAT-openbmc-py.spec

debian/dirs

### The UT result
`##The UT output##`


zet809:
Close this PR since it include some invalid commit.

PR #5936, opened 2019-01-09
Fix remoteshell compatibility with custom Match

jjohnson42:
If a user has a custom Match directive, remoteshell would
create an invalid configuration.  Fix by ensuring we are
outside of a match context by doing Match all explicitly at the end.

### The modification include

_##item1_

_##item2_

### The UT result
`##The UT output##`

_##Feature description##_

### The content of the PR:

### The UT result
`##The UT output##`

### DIFF
--- a/xCAT/postscripts/remoteshell
+++ b/xCAT/postscripts/remoteshell
@@ -59,6 +59,7 @@ then
 	logger -t $log_label -p local4.info "remoteshell:  setup /etc/ssh/sshd_config and ssh_config"
 	cp /etc/ssh/sshd_config /etc/ssh/sshd_config.ORIG
         #delete all occurance of the attribute and then add xCAT settings
+        echo "Match all" >>/etc/ssh/sshd_config
         sed -i '/X11Forwarding /'d /etc/ssh/sshd_config
         echo "X11Forwarding yes" >>/etc/ssh/sshd_config
         sed -i '/MaxStartups /'d /etc/ssh/sshd_config


immarvin:
i am ok with the change

immarvin:
this PR will cause the issue https://github.com/xcat2/xcat-core/issues/5944

the root cause is  `Match all` is not supported in `openssh-server-5.3p1-123.el6_9.ppc64` on `Red Hat Enterprise Linux Server release 6.10 (Santiago)`.

```
     Match   Introduces a conditional block.  If all of the criteria on the
             Match line are satisfied, the keywords on the following lines
             override those set in the global section of the config file,
             until either another Match line or the end of the file.  If a
             keyword appears in multiple Match blocks that are satisfied, only
             the first instance of the keyword is applied.

             The arguments to Match are one or more criteria-pattern pairs.
             The available criteria are User, Group, Host, LocalAddress,
             LocalPort, and Address.  The match patterns may consist of sing
single
             entries or comma-separated lists and may use the wildcard and
             negation operators described in the PATTERNS section of
             ssh_config(5).

             The patterns in an Address criteria may additionally contain
             addresses to match in CIDR address/masklen format, e.g.
             “192.0.2.0/24” or “3ffe:ffff::/32”.  Note that the mask length
             provided must be consistent with the address - it is an error to
             specify a mask length that is too long for the address or one
             with bits set in this host portion of the address.  For example,
             “192.0.2.0/33” and “192.0.2.0/8” respectively.

             Only a subset of keywords may be used on the lines following a
             Match keyword.  Available keywords are AllowAgentForwarding,
             AllowTcpForwarding, Banner, ChrootDirectory, ForceCommand,
             GatewayPorts, GSSAPIAuthentication, HostbasedAuthentication,
             KbdInteractiveAuthentication, KerberosAuthentication,
             KerberosUseKuserok, MaxAuthTries, MaxSessions,
             PubkeyAuthentication, AuthorizedKeysCommand,
             AuthorizedKeysCommandRunAs, PasswordAuthentication,
             PermitEmptyPasswords, PermitOpen, PermitRootLogin,
             RequiredAuthentications1, RequiredAuthentications2,
             RhostsRSAAuthentication, RSAAuthentication, X11DisplayOffset,
             X11Forwarding and X11UseLocalHost.
```

created a PR https://github.com/xcat2/xcat-core/pull/5945 to fix this 

@jjohnson42

immarvin:
the rh7.6 and rh6.10 autotest still fails, we would like to revert this PR to avoid broken development build. 

would you please provide a PR with backward compatibility? @jjohnson42

PR #5947, opened 2019-01-15
Improve the cuda_power9_setup to work with both RHEL 7.5 and 7.6

### LINKED ISSUE #5852: 40-redhat.rules file change causes postscript cuda_power9_setup to fail
With RHEL7.6-alt a change in the systemd rpm in file, $chroot//lib/udev/rules.d/40-redhat.rules
results in postscript cuda_power9_setup
failing to remove the hot memory add line in both the diskful and diskless builds.

> 	sed -i /SUBSYSTEM==\"memory\"/d ${IMG_ROOTIMGDIR}/lib/udev/rules.d/40-redhat.rules

The postscript command above finds the line in 7.5 but does not match the new 7.6-alt
SUBSYSTEM=="memory"   changed to
SUBSYSTEM!="memory"

Additionally I am unsure of the GOTO and LABEL purpose, and thus am not sure just deleting the one row would have the desired effect.

7.5-alt:
```
# Memory hotadd request
SUBSYSTEM=="memory", ACTION=="add", PROGRAM="/bin/uname -p", RESULT!="s390*", ATTR{state}=="offline", ATTR{state}="online"
```

7.6-alt:
```
# Memory hotadd request
SUBSYSTEM!="memory", ACTION!="add", GOTO="memory_hotplug_end"
PROGRAM="/bin/uname -p", RESULT=="s390*", GOTO="memory_hotplug_end"

ENV{.state}="online"
PROGRAM="/bin/systemd-detect-virt", RESULT=="none", ENV{.state}="online_movable"
ATTR{state}=="offline", ATTR{state}="$env{.state}"

LABEL="memory_hotplug_end"
```

- [x] A minor error exists in the documentation.

https://xcat-docs.readthedocs.io/en/latest/advanced/gpu/nvidia/osimage/rhels.html#power9-setup
for Diskful seems to have an error.  In the documentation text, 
" Instead, add this postscript to your compute node postbootscrtips list."
In order to match the example: "chdef p9compute -p postscripts=cuda_power9_setup"
I am guessing the diskfull text should read postscripts in place of postbootscripts.

https://xcat-docs.readthedocs.io/en/latest/guides/admin-guides/references/man5/postscripts.5.html
seems to indicate postscripts would be better than postbootscripts.

whowutwut:
Resolves #5852

### The modification include

_Create the overrides in /etc/udev  instead of modifying /lib/udev_
_Support for both RHEL 7.5 and RHEL 7.6 differences_

### DIFF
--- a/xCAT/postscripts/cuda_power9_setup
+++ b/xCAT/postscripts/cuda_power9_setup
@@ -15,21 +15,51 @@
 
 umask 0022
 
-mkdir ${IMG_ROOTIMGDIR}/etc/systemd/system/nvidia-persistenced.service.d
+echo "==> Override settings in /etc/systemd/system/nvidia-persistenced.service.d..."
+mkdir -p ${IMG_ROOTIMGDIR}/etc/systemd/system/nvidia-persistenced.service.d
 printf '[Service]\nPIDFile=/var/run/nvidia-persistenced/nvidia-persistenced.pid\n' > ${IMG_ROOTIMGDIR}/etc/systemd/system/nvidia-persistenced.service.d/pidfile.conf
 printf '[Service]\nRestart=always\n' > ${IMG_ROOTIMGDIR}/etc/systemd/system/nvidia-persistenced.service.d/restartalways.conf
 
 [ ! -z "${IMG_ROOTIMGDIR}" ] && CHROOTCMD="chroot ${IMG_ROOTIMGDIR}"
 $CHROOTCMD /bin/bash -c "systemctl enable nvidia-persistenced"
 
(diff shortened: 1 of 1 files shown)


zet809:
hi, @whowutwut , if this script is verified, shall we merge it into master branch?

whowutwut:
@zet809  Would have to double check with the version we have  in the  devops repo....  Unless @cxhong  has some cycles to check and then push to this branch if needed.

zet809:
The PR cover the cuda_power9_setup  script for both rh7.5 and rh7.6, so close this PR.

PR #5976, opened 2019-01-24
Have dhcp.pm recognize OPA/IB hwaddr

jjohnson42:
### The modification include

_##item1_

_##item2_

### The UT result
`##The UT output##`

_##Feature description##_

### The content of the PR:

### The UT result
`##The UT output##`

### DIFF
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -764,7 +764,9 @@ sub addnode
                         $hostname = $1 . "-hf" . $count;
                     }
                 }
-            } elsif (length($mac) == 23) {
+            } elsif (length($mac) == 23) { # 8 bytes of mac address
+		# Currently the only thing that has 8 bytes is an infiniband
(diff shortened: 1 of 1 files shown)


zet809:
hi,. @jjohnson42 , UT result means the result and recording of how do your test your fix. 
For example, what happens before your fix.
Then, with correct input, and the expect output is returned with the fix. 
Or incorrect input, expected fails.

xCAT-server/lib/xcat/plugins/dhcp.pm
```diff
@@ -764,6 +764,8 @@ sub addnode
                         $hostname = $1 . "-hf" . $count;
                     }
                 }
+            } elsif (length($mac) == 23) {
```
zet809: The omni path MAC address will be `00:11:75:01:01:0d:cd:c7`, the length is 23.
Maybe here need a comment.

jjohnson42: I have added comment to explain

Issue #5974, opened 2019-01-24
The logic of test case runcmdinstaller_command was changed incorrectly by PR 5941  


hu-weihua:
The logic of test case ``runcmdinstaller_command`` was changed incorrectly by PR #5941, need to change it back. 

```
------START::runcmdinstaller_command::Time:Wed Jan 23 10:40:30 2019------

FILENAME:/opt/xcat/bin/../share/xcat/tools/autotest/testcase//runcmdinstaller/cases0

RUN:chtab key=xcatdebugmode site.value="2" [Wed Jan 23 10:40:30 2019]
ElapsedTime:1 sec
RETURN rc = 0
OUTPUT:
CHECK:rc == 0   [Pass]

RUN:/opt/xcat/share/xcat/tools/autotest/testcase/commoncmd/retry_install.sh  f6u13k15 rhels7.6-alternate-ppc64le-install-compute [Wed Jan 23 10:40:31 2019]
ElapsedTime:555 sec
RETURN rc = 0
OUTPUT:
Try to retry rinstall 5 times ......
Try to install f6u13k15 on the 1 time...
rinstall f6u13k15 osimage=rhels7.6-alternate-ppc64le-install-compute
Provision node(s): f6u13k15
The status is not booted...
The status is not booted...
......
The status is not booted...
The status is not booted...
    status=booted
The tobooted is 0
PING f6u13k15 (10.6.13.15) 56(84) bytes of data.
64 bytes from f6u13k15 (10.6.13.15): icmp_seq=1 ttl=64 time=0.299 ms
64 bytes from f6u13k15 (10.6.13.15): icmp_seq=2 ttl=64 time=0.240 ms

--- f6u13k15 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1001ms
rtt min/avg/max/mdev = 0.240/0.269/0.299/0.033 ms
The pingable is 0
f6u13k15: Wed Jan 23 10:49:46 EST 2019
The canruncmd is 0
The provision succeed on the 1 time.....
The provision succeed......
CHECK:rc == 0   [Pass]
CHECK:output =~ Provision node\(s\)\: f6u13k15  [Pass]

RUN:a=0;while ! `lsdef -l f6u13k15|grep status|grep installing >/dev/null`; do sleep 20;((a++));if [ $a -gt 30 ];then break;fi done [Wed Jan 23 10:49:46 2019]
ElapsedTime:635 sec
RETURN rc = 0
OUTPUT:

[...]

Issue #5975, opened 2019-01-24
A typo in test case reg_linux_statelite_installation_flat


hu-weihua:
There is a typo of test case ``reg_linux_statelite_installation_flat``  in PR  https://github.com/xcat2/xcat-core/pull/5941.  The second argument should be ``rhels7.6-alternate-ppc64le-statelite-compute``, not ``osimage=rhels7.6-alternate-ppc64le-statelite-compute``. Need to fix.

```
RUN:/opt/xcat/share/xcat/tools/autotest/testcase/commoncmd/retry_install.sh  f6u13k15 osimage=rhels7.6-alternate-ppc64le-statelite-compute [Wed Jan 23 11:27:00 2019]
ElapsedTime:2 sec
RETURN rc = 1
OUTPUT:
Try to retry rinstall 5 times ......
Try to install f6u13k15 on the 1 time...
rinstall f6u13k15 osimage=osimage=rhels7.6-alternate-ppc64le-statelite-compute
Error: [f6u13k13]: Cannot find the OS image osimage=rhels7.6-alternate-ppc64le-statelite-compute in the osimage table.
rinstall failed, double check xcat command rinstall to see if it is a bug...
CHECK:rc == 0   [Failed]
```

PR #5977, opened 2019-01-24
refine cases to fix bug 5974,5975

tingtli:
### The PR is to fix issue https://github.com/xcat2/xcat-core/issues/5974 and  https://github.com/xcat2/xcat-core/issues/5975


hu-weihua:

PR #5993, opened 2019-02-06
Python example suggestions

NickyDaB:
# Python example suggestions

## Purpose
Suggested changes to python example script to make it more literal to new users. 

## Approach
I suggest a change in a variable name and an accessed attribute in the script. 

In the original script, the response object was called `all_nodes`. I suggest changing the name of this var to `response`.

Reasoning:
- `all_nodes` can imply that this is a simple python list of node names. 
- it is actually a response object with additional attributes that can be accessed such as `.content`, `.text`, and `.json`. 

In the original script, we access the `.content` attribute of the response object. I suggest we use the `.text` attribute instead.

Reasoning:
- `content` is "Content of the response, in bytes." It prints nicely in the console because python is smart enough when we call print. 
- Might be clearer to use the `.text` attribute for someone new looking at this example. This may make it clearer that we are looking at the response and printing the text of it in the console. 

## Contents
This pull request includes :
- A var name change to the example file. 
- A change in which attribute is accessed of the response object.

### DIFF
--- a/xCAT-server/xCAT-wsapi/xcatws-test.py
+++ b/xCAT-server/xCAT-wsapi/xcatws-test.py
@@ -77,20 +77,20 @@ except Exception as e:
 #
 # Send a request to get all nodes, passing in user and password
 #
-all_nodes = requests.get(get_all_nodes + "?userName=" + username + "&userPW=" + password, verify=False)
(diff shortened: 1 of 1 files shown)


NickyDaB:
I don't have access to fill out reviewers or labels.

PR #6034, opened 2019-02-25
doc: use pkill instead of ps | grep | awk | xargs

kcgthb:
pkill is provided by the same procps package as ps, so should be present
wherever ps is.


immarvin:
agree to merge, thanks @kcgthb

PR #6041, opened 2019-02-26
[RH8] RHEL 8 Service node automatic installation

neo954:
### The modification include

_##Package list for service node_
_##Patch for PostgreSQL document_
_##Revise RPM dependency for xCATsn_


immarvin:
ok to merge this

PR #6047, opened 2019-03-01
remove case nodestat_noderange

tingtli:
### The PR is remove case nodestat_noderange


hu-weihua:

PR #6073, opened 2019-03-12
Fix issue miss sitemaster

zet809:
### The modification include

1. In get_ServiceNode, add WARN message if no noderes.servicenode and site.master are not set
2. In xcatd, if preprocess_request return null, checking site.master and report error if not set.

### The UT result
Without fix, nothing returned:
```
~# rpower xcatcn state
~#
```
With fix
```
~# rpower xcatcn state
Error: [xcatmn]: No site.master set, please check
```


zet809:
This PR is disturbed by other commits, replace with  https://github.com/xcat2/xcat-core/pull/6074

PR #6074, opened 2019-03-12
Fix issue that site.master is not set

zet809:
### The modification include

1. In get_ServiceNode, add WARN message if no noderes.servicenode and site.master are not set
2. In xcatd, if preprocess_request return null, checking site.master and report error if not set.

### The UT result
Without fix, nothing returned:
```
~# rpower xcatcn state
~#
```
With fix
```
~# rpower xcatcn state
Warning: [xcatmn]: The 'master' attribute is not set in the site table and may cause  unexpected behavior.
```

### DIFF
--- a/perl-xCAT/xCAT/ServiceNodeUtils.pm
+++ b/perl-xCAT/xCAT/ServiceNodeUtils.pm
@@ -555,6 +555,10 @@ sub get_ServiceNode
                 {
                     push @{ $snhash{$master} }, $node;
                 }
+                else
+                {
+                    xCAT::MsgUtils->message('SW', "Unknown master for node: $node, neither noderes.servicenode nor site.master is set\n");
+                }
             }
         }
 
--- a/xCAT-server/sbin/xcatd
+++ b/xCAT-server/sbin/xcatd
@@ -2283,6 +2283,10 @@ sub dispatch_request {
         $SIG{CHLD} = 'DEFAULT';
         xCAT::MsgUtils->trace(0, "D", "xcatd: handle request '$req->{command}->[0]' by plugin '$modname''s preprocess_request");
         $reqs = ${ "xCAT_plugin::" . $modname . "::" }{preprocess_request}->($req, $dispatch_cb, \&do_request);
+        if (not(scalar @$reqs) and not(defined xCAT::TableUtils->get_site_attribute('master'))) {
+            $dispatch_cb->({ error => ['No site.master set, please check'], errorcode => [1] });
+            return;
+        }
     } else {    # otherwise, pass it in without hierarchy support
         $reqs = [$req];
     }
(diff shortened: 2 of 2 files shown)


xCAT-server/sbin/xcatd
```diff
@@ -2283,6 +2283,10 @@ sub dispatch_request {
         $SIG{CHLD} = 'DEFAULT';
         xCAT::MsgUtils->trace(0, "D", "xcatd: handle request '$req->{command}->[0]' by plugin '$modname''s preprocess_request");
         $reqs = ${ "xCAT_plugin::" . $modname . "::" }{preprocess_request}->($req, $dispatch_cb, \&do_request);
+        if (not(scalar @$reqs) and not(defined xCAT::TableUtils->get_site_attribute('master'))) {
+            $dispatch_cb->({ error => ['No site.master set, please check'], errorcode => [1] });
```
whowutwut: Don't like this message, would recommend something like: 

```
"WARN: The 'master' attribute is not set in the site table and may cause  unexpected behavior."
```

perl-xCAT/xCAT/ServiceNodeUtils.pm
```diff
@@ -555,6 +555,10 @@ sub get_ServiceNode
                 {
                     push @{ $snhash{$master} }, $node;
                 }
+                else
+                {
+                    xCAT::MsgUtils->message('SW', "Unknown master for node: $node, neither noderes.servicenode nor site.master is set\n");
```
whowutwut: What does this mean, unknown master?

For `, neither.... is set` would suggest `, the noderes.servicenode or site.master attribute is not set.`

whowutwut:
@zet809 One concern here, would this have any impact to the performance of any commands? Or is this minimal enough?  The other thing... is there a way that we could identify key attributes that must be set in order to have general function of xCAT and all we have to do is add those required attributes into some array....

I think `passwd` and `domain` is also something that xCAT will complain about users install and try to manage nodes...

xCAT-server/sbin/xcatd
zet809: Updated.

perl-xCAT/xCAT/ServiceNodeUtils.pm
zet809: In fact, if go into this section, it means both noderes.servicenode and site.master are not set. If either of them is set, there won't be such issue.

zet809:
hi, @whowutwut , thanks for your comments:
1. I think it is minimal modification since the checking only occurs if there is error.
2. There are 20+ plugins using get_ServiceNode subroutine, minimal effort to modify code.
3. You are right that we shall have some key attributes need to be checked. But it depends, for example, hardware control don't need passwd and domain attribute, so shall we checking them when triggering rpower or rinv? Maybe we can add some restriction that not all chdef to delete content of such key attributes in site table?

zet809:
Hi, @whowutwut , will you pls help to review my modification?

immarvin:
agree to merge first

PR #6078, opened 2019-03-12
Change the default_md in openssl.cnf.tmpl to sha384

neo954:
### The modification include

_##Change the default_md in `openssl.cnf.tmpl` to sha384_


immarvin:
i am ok to merge

PR #6081, opened 2019-03-12
Handle the exit code properly when `copycds` failed, and pass through outputs of `cpio`

neo954:
### The modification include

_##Handle the exit code of child process (`cpio`) properly_
_##Handle the output of `cpio` properly_
_##Handle the exit code properly when `copycds` failed_

### The unit test result ###

```
# copycds /mnt/iso/redhat/8/RHEL-8.0.0-RC-1.0/RHEL-8.0.0-20190228.1-ppc64le-dvd1.is
Copying media to /install/rhels8.0.0/ppc64le
cpio: write error: No space left on device
Error: [c910f03c05k21]: Media copy operation failed, status 1
# echo $?
1
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/anaconda.pm
+++ b/xCAT-server/lib/xcat/plugins/anaconda.pm
@@ -2250,7 +2250,7 @@ sub copycd
             print $KID $_ . "\n";
         }
         close($KID);
-        $rc = $?;
+        $rc = $? >> 8;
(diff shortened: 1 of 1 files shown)


neo954:
This is the same pull request as #6075. As the git history was rewritten, these patches were lost.

immarvin:
agree to merge

Issue #6062, opened 2019-03-08
Does Syncfile transfer whole directory trees? 


henkela:
Hi, 
we couldn't find anything about it in the documentation about whether it's possible to sync whole directory trees? Is this an undocumented feature?
Best,

gurevichmark:
Have you tried a statement like `/home/mikev/* -> /home/mikev/` in synclist file ?

immarvin:
hi @henkela , yes. You can sync directories with the sync list like:
```
# cat /tmp/synclist
/tmp/test  -> /tmp/
# ll /tmp/test/
total 0
-rw-r--r-- 1 root root 0 Mar  8 18:32 aa
-rw-r--r-- 1 root root 0 Mar  8 18:32 bbb
-rw-r--r-- 1 root root 0 Mar  8 18:32 cc
# updatenode mid21tor24cn03 -F
File synchronization has completed for nodes: "mid21tor24cn03"

# xdsh mid21tor24cn03 ls /tmp/test/
mid21tor24cn03: aa
mid21tor24cn03: bbb
mid21tor24cn03: cc

```

@gurevichmark , may be we should add this in the doc https://xcat-docs.readthedocs.io/en/stable/guides/admin-guides/manage_clusters/common/deployment/syncfile/syncfile_synclist_file.html#support-nodes-in-synclist-file

gurevichmark:
@immarvin What if there was also a directory `/tmp/test/dd` with a file `/tmp/test/dd/eee`? Would that directory and its file also be copied if `/tmp/test  -> /tmp/` or `/tmp/test/*  -> /tmp/` is used ?

immarvin:
hi @gurevichmark , yes, I think so

gurevichmark:
Made this and other doc improvements under #6084.

gurevichmark:
Seems that both `/tmp/test -> /tmp/` and `/tmp/test/* -> /tmp/` do the same thing. I can not see the difference.

immarvin:
hi @henkela , please close the ticket if the issue has been resolved, thanks

PR #6084, opened 2019-03-12
Some more doc fixes

gurevichmark:
Various doc fixes


docs/source/guides/admin-guides/manage_clusters/common/deployment/syncfile/syncfile_synclist_file.rst
```diff
@@ -45,7 +45,7 @@ sync file **/etc/file2** to the file **/etc/file2** on the node   ::
 
        /etc/file2 -> /etc/
 
-sync all files in **/home/mikev** to directory **/home/mikev** on the node  ::
```
immarvin: hi @gurevichmark , suggest also mentioning the synclist entry for directory like ` /home/mikev -> /tmp/mikev/` explicitly, so that customer won't be confused on the issue in https://github.com/xcat2/xcat-core/issues/6062

PR #6085, opened 2019-03-12
Remove --no-host-reboot option from rflash

gurevichmark:
### The modification include

* Remove `--no-host-reboot` option from `rflash`. Perl and Python versions.

* Update documentation


zet809:
Hi, @gurevichmark , so remove `--no-host-reboot` means host shall be off and then on whenever doing rflash, right? Then, if the host can not boot up, is there any useful message return?

gurevichmark:
@zet809 Good point. I added a new commit to display an extra message if host does not boot up.
I only did this for Perl version, the Python version needs a little more work because it does not seem to verify the host boots up. But since the `rflash` command is still disabled for Python we might not want to deal with that now. (#6093)

zet809:
Hi, @gurevichmark , it seems we'd better just remove the option from document, so replace this PR with https://github.com/xcat2/xcat-core/pull/6156

Issue #6096, opened 2019-03-14
syntax error `/opt/xcat/xcatdsklspost: line 498: break: only meaningful in a `for', `while', or `until' loop`


immarvin:
```
trying to download postscripts...
/opt/xcat/xcatdsklspost: line 498: break: only meaningful in a `for', `while', or `until' loop
postscripts downloaded successfully
trying to get mypostscript from c910f02c01p07...
```

immarvin:
fixed in https://github.com/xcat2/xcat-core/pull/6097

xuweibj:
Verified on ``Version 2.14.6 (git commit e97e5fa8f9f139cc04fc4e8ee5d794d5e159c17a, built Thu Mar 21 06:18:57 EDT 2019)``

PR #6097, opened 2019-03-14
fix issue syntax error in xcatdsklspost

immarvin:
### The PR is to fix issue https://github.com/xcat2/xcat-core/issues/6096

UT:
```
updating VPD database
trying to download postscripts...
postscripts downloaded successfully
trying to get mypostscript from c910f02c01p07...

```


neo954:

PR #6098, opened 2019-03-14
Do not show error message when command failure

neo954:
### The modification include

_##Do not show error message when command failure_

### The UT result

```
# xcatconfig -u
# echo $?
0
```


immarvin:
agree to merge

PR #6100, opened 2019-03-14
[RH8] Merge rhels8 branch to master

neo954:
This pull request is for xcat2/xcat2-task-management#443

### The modification include

_##Add xCAT support for RHEL 8_


immarvin:
agree to merge

PR #6101, opened 2019-03-14
[RH8] Revise Schema.pm based on the new default setting of xcatsslversion

neo954:
This pull request reflect the change of #6061


immarvin:
agree to merge

gurevichmark:
@neo954 You will also need to rebuild docs, since `Schema.pm` is used to generate some man pages.

PR #6102, opened 2019-03-14
Enlarge the default size of the root filesystem for RHEL 8

neo954:
This pull request enlarge the default size of the root filesystem for RHEL 8


immarvin:
agree to merge

neo954:
This pull request partly solve xcat2/xcat2-task-management#643.

PR #6095, opened 2019-03-14
[RH8] Revise xcatconfig, show warning message if `TLSv1' is used

neo954:
### The modification include

_##Revise xcatconfig, show warning message if `TLSv1' is used_

<details>
<summary>The UT result</summary>

```
# xcatconfig -c
__      ___   ___ _  _ ___ _  _  ___
\ \    / /_\ | _ \ \| |_ _| \| |/ __|                      _     +-+-+-+-+-+-+-+
 \ \/\/ / _ \|   / .` || || .` | (_ |                     oo\    |W|A|R|N|I|N|G|
  \_/\_/_/ \_\_|_\_|\_|___|_|\_|\___|                    (__)\   +-+-+-+-+-+-+-+
+--------------------------------------------------------------+ +-+-+-+-+-+-+-+
| Using `TSLv1' for `site.xcatsslversion' was depracated.      |:|W|A|R|N|I|N|G|
| Run `chdef -t site xcatsslversion=' to update your system to |:+-+-+-+-+-+-+-+
| the new default value. See `man site' for more details.      |:+-+-+-+-+-+-+-+
+--------------------------------------------------------------+:|W|A|R|N|I|N|G|
 ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::+-+-+-+-+-+-+-+

Setting up basic certificates.  Respond with a 'y' when prompted.

Existing xCAT certificate authority detected at /etc/xcat/ca, delete? (y/n):# NOTE use "-newkey rsa:2048" if running OpenSSL 0.9.8a or higher
Generating RSA private key, 2048 bit long modulus (2 primes)
.........................+++++
......................................................+++++
e is 65537 (0x010001)
Using configuration from openssl.cnf
Check that the request matches the signature
Signature ok
Certificate Details:
        Serial Number: 1 (0x1)
        Validity
            Not Before: Jan  1 01:01:01 1970 GMT
            Not After : Mar 14 06:00:42 2039 GMT
        Subject:
            commonName                = xCAT CA
        X509v3 extensions:
            X509v3 Subject Key Identifier:
                C9:6D:65:7F:58:E7:88:A7:D1:4D:16:F4:75:96:EA:F6:11:65:75:B5
            X509v3 Authority Key Identifier:
                keyid:C9:6D:65:7F:58:E7:88:A7:D1:4D:16:F4:75:96:EA:F6:11:65:75:B5

            X509v3 Basic Constraints: critical
                CA:TRUE
            X509v3 Key Usage:
[...]

PR #6105, opened 2019-03-14
Revert "[RH8] Revise xcatconfig, show warning message if `TLSv1' is used"

neo954:
Reverts xcat2/xcat-core#6095


neo954:
It seems pull request #6095 break brand new xCAT installation.

PR #6117, opened 2019-03-18
Add python3 to the default pkglist for RHEL 8

neo954:
### The modification include

_##Add python36 to the default compute node pkglist for RHEL 8_


immarvin:
agree to merge

Issue #6118, opened 2019-03-19
otherpkgdir won't work with localdir and http repo


zet809:
For my user case, for otherpkglist, some package are online, and others are in local directory. My osimage looks like below.
```
# cat my_osimage.yaml
osimage:
  my_osimage:
    basic_attributes:
      arch: ppc64le
      distribution: rhels7.6
      osdistro: rhels7.6-ppc64le
      osname: Linux
    imagetype: linux
    package_selection:
      otherpkgdir:
      - /install/REPO/software
      - http://ftp.unicamp.br/pub/linuxpatch/toolchain/at/redhat/RHEL7/
      otherpkglist:
      - '/install/custom/my_osimage/local_pkgs.otherpkglist'
      - '/install/custom/my_osimage/advance-toolchain.otherpkglist'
...
# cat /install/custom/my_osimage/local_pkgs.otherpkglist
aaa
bbb
ccc
# cat /install/custom/my_osimage/advance-toolchain.otherpkglist
advance-toolchain-at12.0-perf
advance-toolchain-at12.0-runtime
advance-toolchain-at12.0-devel
```
When deploy it to CN, the pkgs list in /install/custom/my_osimage/advance-toolchain.otherpkglist which can be get on-line can not be installed with otherpkgs script, the log below show more details. 
```
+ logger -p local4.info -t xcat.updatenode.postscript ' aaa bbb ccc installed.'
+ '[' ']'
+ '[' ' advance-toolchain-at12.0-perf* advance-toolchain-at12.0-runtime* advance-toolchain-at12.0-devel*' '!=' '' -a -n 172.16.253.100:80/install/REPO/software ']'
+ echo 'Warning: the packages  advance-toolchain-at12.0-perf* advance-toolchain-at12.0-runtime* advance-toolchain-at12.0-devel* could not be found in the yum/apt repository, falling back to rpm/dpkg command. If you want your packages to be installed with yum/apt, verify yum/apt is installed and createrepo/dpkg-scanpackages has been run.'
Warning: the packages  advance-toolchain-at12.0-perf* advance-toolchain-at12.0-runtime* advance-toolchain-at12.0-devel* could not be found in the yum/apt repository, falling back to rpm/dpkg command. If you want your packages to be installed with yum/apt, verify yum/apt is installed and createrepo/dpkg-scanpackages has been run.
[...]

PR #6126, opened 2019-03-19
fix issue otherpkgdir won't work with localdir and http repo #6118

immarvin:
### The PR is to fix issue https://github.com/xcat2/xcat-core/issues/6118

* add the http url repo in to repo first
* refine the package availability detection  logic
* add skip_if_unavailable option into repo conf to avoid yum operation abort due to broken repo


immarvin:
verified

zet809:
Looks good to me, @immarvin , pls modify the indent

zet809:
OK to me, let's refine the format in the next release.

PR #6137, opened 2019-03-20
Fix python error, TypeError: cannot use a string pattern on a bytes-like object

neo954:
### The modification include

_##Fix python error, TypeError: cannot use a string pattern on a bytes-like object_

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/pre.rhels8
+++ b/xCAT-server/share/xcat/install/scripts/pre.rhels8
@@ -77,7 +77,7 @@ try:
 			received = newSocket.recv(200)
 			if not received:
 				break
-			command = re.split('\s+',received)
+			command = re.split(b'\s+',received)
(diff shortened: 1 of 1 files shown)


neo954:
See unit test result below.

Without the patch.
```
# ./foo.py
Traceback (most recent call last):
  File "./foo.py", line 33, in <module>
    command = re.split('\s+',received)
  File "/usr/lib64/python3.6/re.py", line 212, in split
    return _compile(pattern, flags).split(string, maxsplit)
TypeError: cannot use a string pattern on a bytes-like object
```

With the patch
```
# ./foo.py
^CTraceback (most recent call last):
  File "./foo.py", line 28, in <module>
    newSocket, address = sock.accept()
  File "/usr/lib64/python3.6/socket.py", line 205, in accept
    fd, addr = self._accept()
```

immarvin:
agree to merge

PR #5967, opened 2019-01-22
Revise openssl settings for RHEL 8

neo954:
### The modification include

_##Remove the RANDFILE from openssl.cnf_
_##Update openssl.cnf.tmpl to the latest one from openssl 1.1.1a source_
_##Remove `-no_ssl2' command line argument for openssl client, totally_
e3e534d_

### The UT result
```
# xcatconfig -c

Setting up basic certificates.  Respond with a 'y' when prompted.

Existing xCAT certificate authority detected at /etc/xcat/ca, delete? (y/n):# NOTE use "-newkey rsa:2048" if running OpenSSL 0.9.8a or higher
Generating RSA private key, 2048 bit long modulus (2 primes)
............................................................................................+++++
.....+++++
e is 65537 (0x010001)
Using configuration from openssl.cnf
Check that the request matches the signature
Signature ok
Certificate Details:
        Serial Number: 1 (0x1)
        Validity
            Not Before: Jan  1 01:01:01 1970 GMT
            Not After : Jan 22 02:27:39 2039 GMT
        Subject:
            commonName                = xCAT CA
        X509v3 extensions:
            X509v3 Subject Key Identifier:
                06:CC:A5:1B:75:1F:B4:06:24:1B:C4:EE:D5:92:2E:D6:8D:DA:15:9C
            X509v3 Authority Key Identifier:
                keyid:06:CC:A5:1B:75:1F:B4:06:24:1B:C4:EE:D5:92:2E:D6:8D:DA:15:9C

            X509v3 Basic Constraints: critical
                CA:TRUE
            X509v3 Key Usage:
                Certificate Sign, CRL Sign
            Netscape Cert Type:
                SSL CA, S/MIME CA
Certificate is to be certified until Jan 22 02:27:39 2039 GMT (7305 days)
Sign the certificate? [y/n]:

1 out of 1 certificate requests certified, commit? [y/n]Write out database with 1 new entries
Data Base Updated
/root
Created xCAT certificate.
/etc/xcat/cert already exists, delete and start over (y/n)?Generating RSA private key, 2048 bit long modulus (2 primes)
....+++++
........................................................+++++
e is 65537 (0x010001)
/root
Using configuration from openssl.cnf
Check that the request matches the signature
Signature ok
Certificate Details:
        Serial Number: 2 (0x2)
        Validity
            Not Before: Jan  1 01:01:01 1960 GMT
            Not After : Jan 17 02:27:39 2039 GMT
        Subject:
            commonName                = c910f03c01p19
[...]

PR #6144, opened 2019-03-21
Partly revert "Revise openssl settings for RHEL 8 (#5967)"

neo954:
### The modification include

* Put "openssl s_client" command line argument "-no_ssl2" back.


immarvin:
agree to merge

PR #6148, opened 2019-03-21
add test case linux_diskless_kdump to weekly bundles

hu-weihua:
### The modification include

* add test case ``linux_diskless_kdump`` to weekly bundles


immarvin:
ok to merge

PR #6149, opened 2019-03-21
Fix python error, TabError: inconsistent use of tabs and spaces in indentation

neo954:
### The modification include

_##Fix #UNCOMMENTOENABLEDEBUGPORT# replacement_

_##Make variable output byte instead of string_


neo954:
This patch is going to fix the problem below.
```
[anaconda root@c910f03c01p10 tmp]# ./foo.py
  File "./foo.py", line 68
    if(command[0] == b"sh"): #DEBUG purposes only, wide open root priv command here.
                                                                                   ^
TabError: inconsistent use of tabs and spaces in indentation
```

immarvin:
agree to merge

PR #6150, opened 2019-03-21
Avoid noisy warning message on rh8 for makegocons 

robin2008:
Ignore the disable when not installed, not using package checking as we use a fake package there

### The modification include

Add a checking before run `systemctl disable xxx`

### The UT result
```
# mv /usr/sbin/conserver /usr/sbin/conserver.bak
# mv /etc/init.d/conserver /etc/init.d/conserver.bak
# systemctl daemon-reload
# systemctl stop goconserver

#makegocons c910f04x27v10
Warning: [c910f04x27v22]: Failed to execute command: systemctl disable conserver.
c910f04x27v10: Created
#makegocons -d c910f04x27v10
c910f04x27v10: Deleted

Apply the patch and restart xcatd

# systemctl stop goconserver
# makegocons c910f04x27v10
Starting goconserver service ...
c910f04x27v10: Created
```

### DIFF
--- a/xCAT-server/lib/perl/xCAT/Goconserver.pm
+++ b/xCAT-server/lib/perl/xCAT/Goconserver.pm
@@ -503,11 +503,13 @@ sub is_goconserver_running {
 sub switch_goconserver {
     my $callback = shift;
     # ignore SN as it is handled by AAsn
-    if ((-x "/usr/bin/systemctl" || -x "-x /bin/systemctl") && !$isSN) {
+    if ((-x "/usr/bin/systemctl" || -x "/bin/systemctl") && !$isSN) {
         my $cmd = "systemctl disable conserver";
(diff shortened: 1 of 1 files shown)


zet809:
Good to me, I agree to merge.

xCAT-server/lib/perl/xCAT/Goconserver.pm
```diff
@@ -537,9 +539,11 @@ sub switch_conserver {
     # ignore SN as it is handled by AAsn
     if ((-x "/usr/bin/systemctl" || -x "-x /bin/systemctl") && !$isSN) {
```
neo954: May need to remove the "-x" in the second constant string variable.

robin2008: Just give a new PR for it.

PR #6151, opened 2019-03-21
Fix a code error

robin2008:
### The modification include
Fix the code error to judge if `/bin/systemctl` is executed.


neo954:

PR #6155, opened 2019-03-22
Cumulative go-xcat updates

neo954:
### The modification include

_##Makes `go-xcat` debugging easier, when failed during "reading repository"_

_##Revise command line arguments handling for `completely uninstall`_


neo954:
Thanks @robin2008.

Issue #6065, opened 2019-03-08
rflash of BMC fw needs to power off Host first


gurevichmark:
The new version of BMC firmware fixes some security issues. As a result, running new BMC fw with older Host fw is no longer permitted.

The recommendation from openbmc team is to power off host before activating BMC fw.

For xCAT `rflash` command we should:

1. Power off host before `rflash -a <bmc fw id>` command is issued.
2. Power off host before `rflash -a <bmc fw .tar>` command is issued.
3. Power off host before `rflash -d` command is issued. 
4. Option `--no-host-reboot` should be removed.
5. Usage and man page should be updated.

gurevichmark:
Flashed the latest BMC FW:
```
[root@briggs01 ~]# rflash mid05tor12cn03 -l
mid05tor12cn03: ID       Purpose State      Version
mid05tor12cn03: -------------------------------------------------------
mid05tor12cn03: 376af621 BMC     Active     ibm-v2.1-438-g0030304-r15-0-g19832d3
mid05tor12cn03: 13ff6e72 Host    Active(*)  IBM-witherspoon-ibm-OP9-v2.0.8-2.2
mid05tor12cn03: ff727722 BMC     Active(*)  ibm-v2.3-476-g2d622cb-r26-0-g7fc4a06
mid05tor12cn03:
[root@briggs01 ~]# ping mid05tor12cn03
PING mid05tor12cn03 (172.12.139.3) 56(84) bytes of data.
64 bytes from mid05tor12cn03 (172.12.139.3): icmp_seq=1 ttl=64 time=0.222 ms
^C
--- mid05tor12cn03 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 0.222/0.222/0.222/0.000 ms

[root@briggs01 ~]# ssh mid05tor12cn03
Last login: Tue Mar 12 10:47:43 2019 from gateway

[root@mid05tor12cn03 ~]#
```

gurevichmark:
Flashed the latest Host FW. Host reboots with new BMC FW and new Host FW:
```
[root@briggs01 consoles]# rflash mid05tor12cn03 -l
mid05tor12cn03: ID       Purpose State      Version
mid05tor12cn03: -------------------------------------------------------
mid05tor12cn03: 376af621 BMC     Active     ibm-v2.1-438-g0030304-r15-0-g19832d3
[...]

PR #6156, opened 2019-03-22
Remove option -no-host-reboot for rflash

zet809:
### The PR is to fix issue  #6065

The modification include
* Remove --no-host-reboot option from rflash usage.
* Update documentation

The UT result


bybai:
I am ok to merge it.

PR #6166, opened 2019-03-25
Remove hardcodes password for sudoer postscript

cxhong:
### The modification include

_##Remove hardcodes username and password  for sudoer_
-##User needs to pass in arguments for sudoer username and password 

### The UT result
`##The UT output##`
````
# updatenode mid08tor03cn01 -P sudoer
mid08tor03cn01: =============updatenode starting====================
mid08tor03cn01: trying to download postscripts...
mid08tor03cn01: postscripts downloaded successfully
mid08tor03cn01: trying to get mypostscript from 172.20.254.2...
mid08tor03cn01: postscript start..: sudoer
mid08tor03cn01:
mid08tor03cn01: Usage: ./sudoer -u username -p password
mid08tor03cn01:         -u sudoer user name
mid08tor03cn01:         -p sudoer password
mid08tor03cn01: postscript end....: sudoer exited with code 1
mid08tor03cn01: Running of postscripts has completed.
mid08tor03cn01: =============updatenode ending====================


# updatenode mid08tor03cn01 -P "sudoer -u xcat2 -p rootpw"
mid08tor03cn01: =============updatenode starting====================
mid08tor03cn01: trying to download postscripts...
mid08tor03cn01: postscripts downloaded successfully
mid08tor03cn01: trying to get mypostscript from 172.20.254.2...
mid08tor03cn01: postscript start..: sudoer
mid08tor03cn01: postscript end....: sudoer exited with code 0
mid08tor03cn01: Running of postscripts has completed.
mid08tor03cn01: =============updatenode ending====================

[root@boston02 ~]# ssh xcat2@mid08tor03cn01
xcat2@mid08tor03cn01's password:
[xcat2@mid08tor03cn01 ~]$ pwd
/home/xcat2

````

### DIFF
--- a/xCAT/postscripts/sudoer
+++ b/xCAT/postscripts/sudoer
@@ -7,6 +7,28 @@
 #
 #------------------------------------------------------------------------------
 
+function usage() {
+  echo ""
+  echo "Usage: $0 -u username -p password"
+  echo -e "\t-u sudoer user name"
+  echo -e "\t-p sudoer password"
+  exit 1
+}
+
+while getopts "u:p:" opt;
+do
+  case $opt in
+    u) SUDOER="$OPTARG";;
+    p) SUDOERPW="$OPTARG";;
+  esac
+done
+
+if [ -z "$SUDOER" ] || [ -z $SUDOERPW ]
+then
+  usage;
+fi
+
+
 if [ -n "$LOGLABEL" ]; then
     log_label=$LOGLABEL
 else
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: xCAT/postscripts/sudoer [hunk 1]
```diff
@@ -7,6 +7,28 @@
...
+
+while getopts "u:p:" opt;
+do
+  case $opt in
+    u) SUDOER="$OPTARG";;
+    p) SUDOERPW="$OPTARG";;
+  esac
+done
+
+if [ -z "$SUDOER" ] || [ -z $SUDOERPW ]
```
immarvin: suggest `$SUDOER` and `$SUDOERPW` be set to default value if not specified to keep backward compatibility

neo954 (changes requested):
I don't like the idea of using command line argument to pass the password. This idea is even worse than hard code the password. I list some of the reasons below.

* User might need to run something like `chdef mid08tor03cn01 -p 'postscripts=sudoer -u xcat2 -p rootpw'` to make it works. It is complex.
* It use to have a single place to change the default password. Now it is scattered to node attribute across different compute nodes.
* Passing password with command line argument is insecure. The command line argument can be read by `ps ax`.

cxhong:
@neo954 , any suggestion which method will be better?   I think we should support different sudoer so can't define in the site table, or passwd table, or node attributes.    can we prompt the command and ask for  user input?

viniciusferrao:
This would be a heavy rewrite to make it correct. Probably for xCAT 2.19 (maybe 3.0)?

viniciusferrao:
This PR was superseeded by #7816 and it's now merged. It took a while but it landed.

Thanks @cxhong!

PR #6172, opened 2019-03-26
update nfs service setup in statelte installation test cases

hu-weihua:
### The modification include

* update nfs service setup in statelte installation test cases

### The UT result
```

RUN:if cat /etc/*release |grep SUSE >/dev/null;then service nfsserver restart; elif cat /etc/*release |grep "Red Hat" >/dev/null;then if [ -f /usr/lib/systemd/system/nfs-server.service ]; then service nfs-server restart; else service nfs start;fi; fi [Tue Mar 26 10:01:27 2019]
ElapsedTime:1 sec
RETURN rc = 0
OUTPUT:
Redirecting to /bin/systemctl restart nfs-server.service
CHECK:rc == 0	[Pass]
```


immarvin:
agree to merge

PR #6178, opened 2019-03-27
Add package NetworkManager to default RHEL 8 diskless package list

neo954:
Since `confignetwork` needs `NetworkManager`. Just add this package in the package list.


bybai:
"NetworkManager" is in RH8 minimum os, so it should be installed in RH8. I am OK to merge it.

PR #6190, opened 2019-03-28
update daily bundle to remove makeconservercf cases

tingtli:
### The PR is to remove makeconservercf cases from daily bundle files.


hu-weihua:

PR #6210, opened 2019-04-08
Remove some AIX references in docs

gurevichmark:
AIX Support only exists in xCAT version 2.9.x


whowutwut:
I added a comment into the description for AIX support.

PR #6211, opened 2019-04-09
Revise the pkglist files for RHEL 8 diskless compute node and service node

neo954:
### The modification include

_##Revise the pkglist for diskless compute node_

_##Revise the pkglist for diskless service node_


immarvin:
hi @neo954 , is `tuned` installed and enabled in diskless node?

immarvin:
agree to merge

PR #6216, opened 2019-04-10
Verify packaging commands can be executed during genimage processing

gurevichmark:
1. Today, when trying to `genimage` Ubuntu image on RH MN, a confusing message is displayed:
```
 Error: Can not create bootstraps for rootimage. Make sure you specified full http mirror path.
```

But the real reason is that `bootstrap` command is not there.
This PR adds a check to verify `bootstrap` is executable and displays a message if it is not:
```
Error: Can not execute debootstrap command. Make sure you are running on a Debian based system.
```

2. Today, when trying to `genimage` SLES image on RH MN, a confusing message is displayed:
```
 zypper invocation failed with rc: 72057594037927935
```

But the real reason is that `zypper` command is not there.
This PR adds a check to verify `zypper` is executable and displays a message if it is not:
```
Can not execute zypper command. Make sure you are running on SUSE based system.
```

### DIFF
--- a/xCAT-server/share/xcat/netboot/sles/genimage
+++ b/xCAT-server/share/xcat/netboot/sles/genimage
@@ -255,6 +255,12 @@ unless ($onlyinitrd) {
     my $non_interactive;
     if (!$prompt) { $non_interactive = "--non-interactive --no-gpg-checks --gpg-auto-import-keys"; }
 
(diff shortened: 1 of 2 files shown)


immarvin:
ok to merge, thanks @gurevichmark

PR #6220, opened 2019-04-11
Update the two README files in the xcat-core project

whowutwut:
To see the proposed changes rendered check out my forked copy: 
* https://github.com/whowutwut/xcat-core/blob/doc_badges/README.md
* https://github.com/whowutwut/xcat-core/blob/doc_badges/docs/README.md

I think the markdown is easier to maintain than RST,  so switching the main landing README to a markdown file.  At some point I think our RTD badges disappeared, so putting them  back, using only latest and stable to avoid the need to continue to update these pages on each release of xCAT. 

Simplified  some of the text in the README and updated links to the latest https


whowutwut:
@zet809 and @gurevichmark can you take a look at this and let me know if you have any issue to merge in?  You can see the rendering by  looking at my forked copy

zet809:
Looks great to me, thx @whowutwut

PR #6233, opened 2019-04-17
Add the shell function base64decode() to pre scripts.

neo954:
### The modification include

* Add function base64decode() into file pre.pkvm3
* Add function base64decode() into file pre.rh
* Add function base64decode() into file pre.sle15
* Add function base64decode() into file pre.sles


immarvin:
ok with the change

neo954:
Thanks @immarvin.

PR #6237, opened 2019-04-18
fix issue 6229 based on customer suggestion

xuweibj:
### The modification include

_##use <dict>.get replace read value directly_

_##modify msg to avoid invalid var "node"_

### The UT result
```
# rspconfig f5u14 ntpservers
f5u14: BMC NTP Servers: None
```


zet809:
Good to me, I agree to merge, thx @xuweibj

Obihoernchen:
Please configure more than one NTP server I think the output is missing a `,` between the IP addresses then.

PR #6238, opened 2019-04-18
Update the cuda setup script to account for difference in rhels 7.5 and rhels 7.6

### LINKED ISSUE #6212: Support CUDA 10.1 in xCAT
For cuda10.1, the "Memory hotadd request" section shall be commented out in /etc/udev/rules.d/40-redhat.rules.

```
## Memory hotadd request
#SUBSYSTEM!="memory", ACTION!="add", GOTO="memory_hotplug_end"
#PROGRAM="/bin/uname -p", RESULT=="s390*", GOTO="memory_hotplug_end"
#
#ENV{.state}="online"
#PROGRAM="/bin/systemd-detect-virt", RESULT=="none", ENV{.state}="online_movable"
#ATTR{state}=="offline", ATTR{state}="$env{.state}"
#
#LABEL="memory_hotplug_end"
```

whowutwut:
In RHELS 7.6 the memory online section has differences so this  script detects and comments  out the memory  section 

Instead of modifying the `/lib/udev` rules, we follow the linux  udev man page and copy the override  file  to /etc/... 
> However, files with identical filenames replace each other. Files in /etc have the highest priority, 

Resolves #6212 

This script has been in use in our internal repo where  we are testing CUDA 10.x

### DIFF
--- a/xCAT/postscripts/cuda_power9_setup
+++ b/xCAT/postscripts/cuda_power9_setup
@@ -15,19 +15,50 @@
 
 umask 0022
 
-mkdir ${IMG_ROOTIMGDIR}/etc/systemd/system/nvidia-persistenced.service.d
+echo "==> Override settings in /etc/systemd/system/nvidia-persistenced.service.d..."
+mkdir -p ${IMG_ROOTIMGDIR}/etc/systemd/system/nvidia-persistenced.service.d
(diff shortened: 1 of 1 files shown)


whowutwut:
Something is wrong with the Travis ☝️ , this script is not tested in the CI.

zet809:
Thanks, @whowutwut

PR #6241, opened 2019-04-18
add tar to compute.rhels8.pkglist

hu-weihua:
### The PR is to add tar to compute.rhels8.pkglist

Due to ``tar`` is a very useful and common command in Linux, should install it by default when provision a rhels8.0 node. For rhels7 and earlier version, the ``tar`` is installed by default.

### DIFF
--- a/xCAT-server/share/xcat/install/rh/compute.rhels8.pkglist
+++ b/xCAT-server/share/xcat/install/rh/compute.rhels8.pkglist
@@ -7,3 +7,4 @@ rsync
 util-linux
 wget
 python3
+tar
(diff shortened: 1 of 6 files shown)


xCAT-server/share/xcat/install/rh/compute.rhels8.pkglist
```diff
@@ -7,3 +7,4 @@ rsync
 util-linux
 wget
 python3
+tar
```
robin2008: Per the team decision, we will keep the same basic package list for diskless and diskful.
So you need to modify the other package list as well,  and if service node is not a symbol link, you need to change it too.

PR #6242, opened 2019-04-18
Remove conserver-xcat from all xCAT shipped RPM requirements

neo954:
* Remove `conserver-xcat` from `xCAT.spec`
* Remove `conserver-xcat` from `xCATsn.spec`
* `xCAT-server` requires `grub2-xcat >= 2.02-0.76`


zet809:
Good to me, I agree to merge, thx @neo954

PR #6243, opened 2019-04-18
Remove conserver-xcat from go-xcat package list

neo954:
Remove `conserver-xcat` from `$GO_XCAT_INSTALL_LIST`


zet809:
Good to me, I agree to merge, thx @neo954

PR #6245, opened 2019-04-18
[DO NOT MERGE] Add interface identifier for multple ip message

gurevichmark:
When displaying a message about detecting multiple IPs for the same interface, it would be helpful to know which interface.

Before the fix:
```
[root@briggs01 xcat]# rspconfig mid05tor12cn05 ntpservers
mid05tor12cn05: Interfaces with multiple IP addresses are not supported
[root@briggs01 xcat]#
```
After the fix:
```
[root@briggs01 xcat]# rspconfig mid05tor12cn05 ntpservers
mid05tor12cn05: Interfaces with multiple IP addresses are not supported: /xyz/openbmc_project/network/eth0_11/ipv4/ac98f1aa
[root@briggs01 xcat]#
```

### DIFF
--- a/xCAT-openbmc-py/lib/python/agent/hwctl/openbmc_client.py
+++ b/xCAT-openbmc-py/lib/python/agent/hwctl/openbmc_client.py
@@ -917,9 +917,9 @@ class OpenBMCRest(object):
                     if 'ip' in netinfo[nicid]:
                         msg = "Another valid ip %s found." % (v["Address"])
                         self._print_record_log(msg, 'get_netinfo')
-                        del netinfo[nicid]
                         netinfo['error'] = 'Interfaces with multiple IP addresses are not supported: %s' % (k)
-                        break
+                        # Store error, but keep going in case info is needed later
+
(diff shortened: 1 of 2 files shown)


xCAT-openbmc-py/lib/python/agent/hwctl/openbmc_client.py
```diff
@@ -918,7 +918,7 @@ def get_netinfo(self):
                         msg = "Another valid ip %s found." % (v["Address"])
                         self._print_record_log(msg, 'get_netinfo')
                         del netinfo[nicid]
-                        netinfo['error'] = 'Interfaces with multiple IP addresses are not supported'
+                        netinfo['error'] = 'Interfaces with multiple IP addresses are not supported: %s' % (k)
```
robin2008: This is fine to add the information, but obviously it need to support 169.x and a valid IP.
If this is the right behavior from BMC side, we may need to handle it, instead of break.

And the `break` will cause no  NTPServers be parsing anymore.

Issue #6019, opened 2019-02-21
go-xcat failed to install xCAT with error 'xcat-dep' is signed with an unknown key on sles12.3


bybai:
go-xcat failed to install xCAT with error 'xcat-dep' is signed with an unknown key on sles12.3

Reproduce steps:
1.  use `go-xcat` to install xCAT on sles12.3
```
]# /tmp/go-xcat install
Operating system:   linux
Architecture:       x86_64
Linux Distribution: sles
Version:            12.3
go-xcat Version:    1.0.34


Reading repositories ...... failed

Repository 'xcat-core' not found by alias, number or URI.
Repository 'xcat-dep' not found by alias, number or URI.
Repository 'SLES12-SP3-12.3-0' is up to date.
Repository 'SuSE-Linux-pkg0' is up to date.
Retrieving repository 'xcat-core' metadata [..done]
Building repository 'xcat-core' cache [....done]
Retrieving repository 'xcat-dep' metadata [.
File 'repomd.xml' from repository 'xcat-dep' is signed with an unknown key '60A3E9ACC6565BC9'. Continue? [yes/no] (no): no
error]
Repository 'xcat-dep' is invalid.
[xcat-dep|http://xcat.org/files/xcat/repos/yum/latest/xcat-dep/sles12/x86_64] Valid metadata not found at specified URL
Please check if the URIs defined for this repository are pointing to a valid repository.
Skipping repository 'xcat-dep' because of the above error.
Some of the repositories have not been refreshed because of an error.
Removing repository 'xcat-dep' [...done]
Repository 'xcat-dep' has been removed.
Removing repository 'xcat-core' [...done]
Repository 'xcat-core' has been removed.
```
2.  OS version
```
 more /etc/*release
::::::::::::::
/etc/os-release
::::::::::::::
NAME="SLES"
VERSION="12-SP3"
VERSION_ID="12.3"
PRETTY_NAME="SUSE Linux Enterprise Server 12 SP3"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:12:sp3"
::::::::::::::
/etc/SuSE-release
::::::::::::::
SUSE Linux Enterprise Server 12 (x86_64)
VERSION = 12
PATCHLEVEL = 3
# This file is deprecated and will be removed in a future service pack or release.
# Please check /etc/os-release for details about this release.
```

neo954:
The release xcat-dep tarball was signed with the wrong GPG key.

[...]

Issue #5306, opened 2018-06-21
Update xCAT Security Key to use RSA and stronger SHA-2 family hash: SHA512, SHA384, SHA256, and SHA224


neo954:
The issue is against the latest `xcat-core` build `xcat-dep-ubuntu-201806200437.tar.bz2`, and `xcat-dep` build `20180620.0615-core-debs-snap.tar.bz2`.

Command `apt-get` complained the GPG signatures were invalid: `F75B1BF678B644FDF3AACFC860A3E9ACC6565BC9`. See details below

```
# apt-get update
Get:1 file:/install/xcat/xcat-core bionic InRelease [1,928 B]
Get:2 file:/install/xcat/xcat-dep bionic InRelease [1,935 B]
Get:1 file:/install/xcat/xcat-core bionic InRelease [1,928 B]
Get:2 file:/install/xcat/xcat-dep bionic InRelease [1,935 B]
Err:1 file:/install/xcat/xcat-core bionic InRelease
  The following signatures were invalid: F75B1BF678B644FDF3AACFC860A3E9ACC6565BC9
Hit:3 http://ports.ubuntu.com/ubuntu-ports bionic InRelease
Hit:4 http://ports.ubuntu.com/ubuntu-ports bionic-updates InRelease
Hit:5 http://archive.ubuntu.com/ubuntu bionic InRelease
Err:2 file:/install/xcat/xcat-dep bionic InRelease
  The following signatures were invalid: F75B1BF678B644FDF3AACFC860A3E9ACC6565BC9
Hit:6 http://archive.ubuntu.com/ubuntu bionic-updates InRelease
Hit:7 http://ports.ubuntu.com/ubuntu-ports bionic-backports InRelease
Get:8 http://archive.ubuntu.com/ubuntu bionic-backports InRelease [74.6 kB]
Hit:9 http://ports.ubuntu.com/ubuntu-ports bionic-security InRelease
Reading package lists... Done
W: GPG error: file:/install/xcat/xcat-core bionic InRelease: The following signatures were invalid: F75B1BF678B644FDF3AACFC860A3E9ACC6565BC9
E: The repository 'file:/install/xcat/xcat-core bionic InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
[...]

PR #6250, opened 2019-04-19

neo954:
Use contemporary RSA key for signing.

This pull request will solve #6019.


neo954:
See #5306, #5316, #5318, #5485.

zet809:
Good to me, thx @neo954

PR #6251, opened 2019-04-19

neo954:
Fix typo in #6250


neo954:
See #5306, #5316, #5318, #5485, #6250.

PR #6255, opened 2019-04-22
Remove package conserver-xcat from service node otherpkgs package lists

neo954:
### The modification include

_##Remove package conserver-xcat from service node otherpkgs package lists_


hu-weihua:

neo954:
Thanks @hu-weihua

PR #6257, opened 2019-04-22
Let makentp fall back to use ntpd when systemctl is not available.

neo954:
### The modification include

_##Let `makentp` fall back to use `ntpd` when `systemctl` is not available._
_##Let postscript `setupntp` fall back to use `ntpd` when `systemctl `is not available._

### The UT result
```
# makentp
configuring management node: c910f03c01p06.
# ps ax | grep -v grep | grep ntp
24426 ?        Ss     0:00 ntpd -u ntp:ntp -p /var/run/ntpd.pid -g
```


immarvin:
ok with the change

Issue #6253, opened 2019-04-22
Dropped conserver in new xcat dep, but did not clear up related pkglist completely 


hu-weihua:
For new xcat dep ``xcat-dep-201904190243.tar.bz2``, xCAT dropped all ``conserver-xcat`` package, but did not clear it up in all related ``otherpkglist`` completely. That resulted in the SN installation will return failure.

At least, xCAT still support rhels7 now.

```
[root@ xcat-dep]# pwd
/xcat-dep

[root@ xcat-dep]# find  ./ -name *conserver*
./rh8/ppc64le/goconserver-0.3.2-snap201812130351.ppc64le.rpm
./rh8/x86_64/goconserver-0.3.2-snap201812130350.x86_64.rpm
./sles12/ppc64le/goconserver-0.3.2-snap201904180606.ppc64le.rpm
./sles12/x86_64/goconserver-0.3.2-snap201904180828.x86_64.rpm
./rh7/ppc64le/goconserver-0.3.2-snap201811080416.ppc64le.rpm
./rh7/x86_64/goconserver-0.3.2-snap201811080416.x86_64.rpm
./rh7/ppc64/goconserver-0.3.2-snap201904180332.ppc64.rpm

[root@c910f03c17k25 rh]# pwd
..../xcat-core/xCAT-server/share/xcat/install/rh

[root@ rh]#  grep "conserver" ./ -rn
./service.rhels5.ppc64.otherpkgs.pkglist:2:xcat/xcat-dep/rh5/ppc64/conserver-xcat
./service.rhels5.x86.otherpkgs.pkglist:2:xcat/xcat-dep/rh5/x86/conserver-xcat
./service.rhels5.x86_64.otherpkgs.pkglist:2:xcat/xcat-dep/rh5/x86_64/conserver-xcat
./service.rhels6.ppc64.otherpkgs.pkglist:2:xcat/xcat-dep/rh6/ppc64/conserver-xcat
./service.rhels6.x86_64.otherpkgs.pkglist:2:xcat/xcat-dep/rh6/x86_64/conserver-xcat
./service.rhels7.ppc64.otherpkgs.pkglist:2:xcat/xcat-dep/rh7/ppc64/conserver-xcat
./service.rhels7.ppc64le.otherpkgs.pkglist:2:xcat/xcat-dep/rh7/ppc64le/conserver-xcat
[...]

PR #6259, opened 2019-04-23
Add goconserver to the SLES12 otherpkgs.pkglist. In order to pull the xcat-dep directory into the zypper repository

neo954:
### The modification include

_##Add goconserver to the SLES12 otherpkgs.pkglist. In order to pull the xcat-dep directory into the zypper repository_


neo954:
See #6253, #6255.

neo954:
By default, xCAT did not install all the OS loaders on all platform. Only OS loaders working on the same platform as the xCAT manage node or service node are installed.

This pull request will makes xCAT install all the OS loaders on all platform. Just like what `go-xcat` already did.

immarvin:
i am ok with the change, thanks  @neo954

PR #6265, opened 2019-04-25
Add notice of nfs setting in xcat docker container document

hu-weihua:
### The PR is to add notice of nfs setting in xcat docker container document

Please refer http://10.3.17.25/advanced/docker/dockerized_xcat/run_xcat_in_docker_host.html to review. thanks

### DIFF
--- a/docs/source/advanced/docker/dockerized_xcat/run_xcat_in_docker_host.rst
+++ b/docs/source/advanced/docker/dockerized_xcat/run_xcat_in_docker_host.rst
@@ -112,5 +112,5 @@ For example ::
     [xcatmn]# lsxcatd -a
 
 .. Attention::
-    Recommend strongly to use external NFS if needed. If perfer to use xCAT container internal NFS service, the shared directory must be mount to Docker host by ``-v`` option when start xCAT container.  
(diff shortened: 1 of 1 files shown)


docs/source/advanced/docker/dockerized_xcat/run_xcat_in_docker_host.rst
```diff
@@ -110,3 +110,7 @@ Now container ``xcatmn`` will work as a normal xCAT management node, can run xCA
 For example ::
 
     [xcatmn]# lsxcatd -a
+
+.. Attention::
+    Recommend strongly to use external NFS if needed. If perfer to use xCAT container internal NFS service, the shared directory must be mount to Docker host by ``-v`` option when start xCAT container.
```
gurevichmark: Better to say:
"Use of external NFS is recommended. For internal NFS service, mount the shared directory with ``-v`` option when starting xCAT container."
Also, maybe we should explain what we mean by "external NFS" and "internal NFS" services ?

hu-weihua: @gurevichmark, thanks for your comments. I got comment from xubin, he said it is better to say ``NFS outside of xCAT Docker container`` and ``NFS inside of xCAT Docker container`` to replace  "external NFS" and "internal NFS".  May be it is easy for customer to understand. 

I have updated the document, could you help to review again? thanks

PR #6271, opened 2019-04-29
Filter out the http port from cumulus-provision-url

cxhong:
### The PR is to fix issue _#6268_  , 
after add support for site.httpport, the cumulus-provision-url changed to `http://$DHCPIP:$httpport/install/postscripts/cumulusztp`,   The postscripts `cumulusztp`, `documulusdiscovery`, and `configinterface` get MASTER dhcp ip address from cumulus-provision-url,  so needs to filter out the httpport from `$DHCPIP:$httpport`.


immarvin:
i am ok with the change

PR #6272, opened 2019-04-29
Reconstruction of the osimage resources search policy

neo954:
### The modification include

_##Reconstruction of the osimage resources search policy_
_##Will check configuration files named `LATEST`_

xcat-template 001


neo954:
@robin2008 @immarvin @zet809 It is great if we can merge this pull request before May 31.

PR #6273, opened 2019-04-30
Update mariadb related document for RHEL 8

neo954:
### The modification include

_##Update mariadb related document for RHEL 8_


zet809:
OK to me, I agree to merge. Thx @neo954

PR #6274, opened 2019-04-30
Fix site.httpport support for grub2-http

neo954:
### The modification include

_##Fix site.httpport support for grub2-http_


immarvin:
agree to merge, thx @neo954

PR #6275, opened 2019-04-30
Fix dhcp.pm for grub2-http and grub2-tftp

neo954:
### The modification include

_##Fix dhcp.pm for grub2-http and grub2-tftp_


immarvin:
i am ok with the change

PR #6280, opened 2019-05-01
xcatprobe xcatmn:  do not check DNS status on mn if site.externaldns=1

khm:
This pull request is to fix an issue with xcatprobe.

Currently, xcatprobe checks for DNS service on the management node unconditionally.  If site.externaldns is set to 1, there should not be a DNS service running on the management node, but `xcatprobe xcatmn` reports failure anyway.  This patch corrects this behavior, skipping themanagement node DNS check iff site.externaldns=1.

alter xcatmn to check the site table:
- if not on a service node 
- AND if externaldns is set to 1
  - skip the checkdns block
- otherwise check dns

### DIFF
--- a/xCAT-probe/subcmds/xcatmn
+++ b/xCAT-probe/subcmds/xcatmn
@@ -946,13 +946,24 @@ sub check_dns_service {
     my $nodename = `hostname -s 2>&1`;
     chomp($nodename);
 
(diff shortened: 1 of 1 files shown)


xuweibj:
Looks good to me. Thanks @khm .

PR #6281, opened 2019-05-01
xcatprobe xcatmn: Allow selinux permissive

khm:
This pull request is to fix an issue with xcatprobe.

Currently, xcatprobe checks to see if selinux is enabled, and reports failure if selinux is enabled.  However, selinux has three possible modes:  "disabled," "permissive," and "enforcing."  Of the three, only "enforcing" means that selinux is actively working on the system.  "Permissive" mode loads policies but does not interfere with operation, but merely logs potential violations.

This change adds a second test to `probe_utils`, which explicitly checks whether selinux is in "enforcing" mode.  In addition, the `check_selinux` test in xcatmn is modified:  if selinux is enabled and enforcing, the test reports failure.  If selinux is enabled but NOT enforcing (this is "permissive" mode), the test reports a warning.  If selinux is "disabled," the test passes with no warnings.

This allows sites who use selinux in "permissive" mode to monitor their systems to pass the `xcatprobe xcatmn` testing.

### DIFF
--- a/xCAT-probe/lib/perl/probe_utils.pm
+++ b/xCAT-probe/lib/perl/probe_utils.pm
@@ -242,6 +242,33 @@ sub is_selinux_enable {
 
 =head3
     Description:
+        Test if SELinux is enforcing in current operating system
+    Arguments:
+         None
+    Returns:
+        1 : yes
+        0 : no
+=cut
+
+#------------------------------------------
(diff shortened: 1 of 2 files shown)


xuweibj:
Looks good to me. Thanks @khm .

PR #6282, opened 2019-05-01
Support new PowerSupplyRedundancy endpoint

gurevichmark:
### The PR is to fix issue _#6270_

This PR deals with a chanded REST API path for Power Supply Redundancy in latest BMC FW

Previous path: `/xyz/openbmc_project/sensors/chassis/PowerSupplyRedundancy/action/getValue`
New path: `/xyz/openbmc_project/control/power_supply_redundancy`

Since we have to deal with older versions of BMC FW, both paths have to be supported.
The new version of BMC FW also makes this attribute read-only. Users can not longer set it.
This fix is for Perl version only. A separate PR will fix Python.

Logic:
* Attempt to use the old REST API path. If successful, code works as before.
* If old REST API path fails with 404 code and the request was to _set_ this attribute, fail with error message as before.
* (new path) If old REST API path fails with 404 code and the request was to _query_ this attribute, call the new REST API path. When response is received, display it in the same format as the old response.

```
[root@briggs01 xcat]# rflash mid05tor12cn03 -l
mid05tor12cn03: ID       Purpose State      Version
mid05tor12cn03: -------------------------------------------------------
mid05tor12cn03: 53105392 Host    Active(*)  IBM-witherspoon-OP9-v2.2-3.2
mid05tor12cn03: 33029bc1 Host    Active     IBM-witherspoon-ibm-OP9_v1.19_1.192
mid05tor12cn03: cf421794 BMC     Active(*)  ibm-v2.3-476-g2d622cb-r32-0-g9973ab0
mid05tor12cn03:
[root@briggs01 xcat]#


[root@briggs01 xcat]# rspconfig mid05tor12cn03 powersupplyredundancy=enabled -V
[briggs01]: Running command in Perl
mid05tor12cn03: [briggs01]: Error: 404 Not Found - Requested endpoint does not exist or may indicate function is not supported on this OpenBMC firmware.
[root@briggs01 xcat]#


[root@briggs01 xcat]# rspconfig mid05tor12cn03 powersupplyredundancy -V
[briggs01]: Running command in Perl
mid05tor12cn03: [briggs01]: BMC PowerSupplyRedundancy: Enabled
[root@briggs01 xcat]#
```

### DIFF
--- a/xCAT-server/lib/xcat/plugins/openbmc.pm
+++ b/xCAT-server/lib/xcat/plugins/openbmc.pm
@@ -401,6 +401,13 @@ my %status_info = (
     RSPCONFIG_GET_RESPONSE => {
         process        => \&rspconfig_response,
     },
+    RSPCONFIG_GET_PSR_REQUEST => {
+        method         => "GET",
+        init_url       => "$openbmc_project_url/control/power_supply_redundancy",
+    },
+    RSPCONFIG_GET_PSR_RESPONSE => {
+        process        => \&rspconfig_response,
+    },
     RSPCONFIG_GET_NIC_REQUEST => {
         method         => "GET",
         init_url       => "$openbmc_project_url/network/enumerate",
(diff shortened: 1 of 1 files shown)


zet809:
HI, @gurevichmark , look good to me, I agree to merge, thx!

PR #6288, opened 2019-05-05
remove kit cases from bundle since it is not supported in the new os version

tingtli:
remove kit cases from bundle since it is not supported in the new os version


hu-weihua:

PR #6297, opened 2019-05-07
probe-utils: Refactor switch statement

khm:
### The modification removes the switch() statement from the selinux check. 

 The switch() statement is only available in perl when `use v5.10` or `use switch` is declared.  Instead of introducing dependencies, we can use a for block to process the value.  After this change there is no longer a syntax error.

### output
```
$ perl -c xCAT-probe/lib/perl/probe_utils.pm 
xCAT-probe/lib/perl/probe_utils.pm syntax OK
```


xCAT-probe/lib/perl/probe_utils.pm
```diff
@@ -252,17 +252,20 @@ sub is_selinux_enable {
...
+    my $retval = 0;
     if (-e "/usr/sbin/getenforce") {
         my $enforce_mode = `/usr/sbin/getenforce`;
         chomp $enforce_mode;
-        switch ($enforce_mode) {
-          case "Disabled" { return 0; }
-          case "Permissive" { return 0; }
-          case "Enforcing" { return 1; }
-          else { return 0; }
+        for ($enforce_mode) {
```
cxhong: why we need for loop here?  there should be only one output, right?

xCAT-probe/lib/perl/probe_utils.pm
```diff
@@ -252,17 +252,20 @@ sub is_selinux_enable {
...
+          if    (/Disabled/)   { $retval = 0; }
+          elsif (/Permissive/) { $retval = 0; }
+          elsif (/Enforcing/)  { $retval = 1; }
+          else { $retval = 0; }
+        }
     } else {
-        return 0;
+        $retval = 0;
```
cxhong: don't think we need else block.

khm: fixed in 59f8845d3

xCAT-probe/lib/perl/probe_utils.pm
khm: It's not really a loop; the for() construct is here used to set $_ in the block for pattern matching.   This is the syntax used in [perlfaq](https://perldoc.perl.org/perlfaq7.html#How-do-I-create-a-switch-or-case-statement%3F) for "pure-perl" switch alternatives that support all versions.

PR #6300, opened 2019-05-08
update bundle since rhels8 is initial supported from 2.15

tingtli:
update bundle since rhels8 is initial supported from 2.15
This is for task https://github.ibm.com/xcat2/task_management/issues/240


hu-weihua:
It is ok for me. @tingtli , Could you update relevant test projects at same time? thanks.

PR #6307, opened 2019-05-10
Add multiple kernel support to mlnxofed_ib_install script, remove v2 version

cxhong:
---Remove version 1 of mlnxofed_ib_install since xCAT never support it any more.
---Change version 2 to mlnofed_ib_install, only support for one version
---Add multiple kernel support for the mlnofed_ib_install scripts


xCAT-server/share/xcat/ib/scripts/Mellanox/mlnxofed_ib_install
```diff
@@ -1,289 +1,432 @@
-#!/bin/sh
+#!/bin/bash
+#
+# This is the second version of Mellanox IB driver installation sample script
+# This is also the version we recommend to use right now.
```
whowutwut: Should we get rid of these lines 4-8. I don’t think we are consistent with adding change logs so are these comments even relevant anymore?

cxhong: removed

PR #6310, opened 2019-05-14
Add kernel module tg3 to diskless initrd image for ppc64el

neo954:
### The modification include

_##Add kernel module `tg3` to diskless initrd image for ppc64el_


immarvin:
agree to merge

PR #6315, opened 2019-05-15
Revised package list file for Ubuntu16.04.5

neo954:
### The modification include

_##Remove compute.ubuntu16.04.pkglist_
_##Diskless package list file for Ubuntu 16.04.5_


immarvin:
agree to merge

PR #6319, opened 2019-05-16
Update grub2-xcat requirement

neo954:
### The modification include

_##Update grub2-xcat requirement_


zet809:
Good to me, i agree to merge, thx @neo954

PR #6320, opened 2019-05-16
Link switchprobe file when building xcatprobe on Ubuntu

gurevichmark:
### The PR is to fix issue _#6313_

This PR adds a link to previously missing `switchprobe` file when building `xcatprobe` on Ubuntu.
The presence of `switchprobe` file is required for running `switch_macmap` subcommand of `xcatprobe`.


hu-weihua:
Let's try it. thanks @gurevichmark

PR #6222, opened 2019-04-12
Add documentation in RTD for how to use an xCAT docker image

hu-weihua:
### The PR is to describe how to use xcat docker image

Please refer  http://10.3.17.25/advanced/docker/index.html to review

PR #6324, opened 2019-05-20
Remove docker man pages

gurevichmark:
In a discussion related to #6222 it was decided not to support docker lifecycle management anymore.

This PR does not change any code, just fixes the documentation:

* Removes man page commands dealing with docker lifecycle management - `mkdocker`, `lsdocker` and `rmdocker`
* Removes man page section of `rpower` command dealing with docker nodes
* Syncs up `.rst` files that probably were forgotten to be update when `.pod` file was changed - `rflash` and `site`


zet809:
Looks good to me, will merge after CI passed

PR #6327, opened 2019-05-21
Remove xinetd from Depends

neo954:
### The modification include

_##Remove xinetd from Depends section of package xcat_


immarvin:
agree to merge

PR #6336, opened 2019-05-23
For Ubuntu 18.04.2, change back to use the original linux-image-generic kernel

neo954:
### The modification include

_##For Ubuntu 18.04.2, change back to use the original linux-image-generic kernel_

PR #6337, opened 2019-05-23
Revert "For Ubuntu 18.04.2, change back to use the original linux-image-generic kernel"

neo954:
Reverts xcat2/xcat-core#6336

Intend to revert this one. Since issue #6305 only affected the KVM guest run POWER9. But for xCAT user who run bare-metal POWER 9 machine, the HWE kernel is important.


immarvin:
@neo954 , would you please create a ticket to mention the different kernel package in ubuntu 18.04.2 diskless pkglists on P9 bare metal and P9 KVM geuests?  Then we can take it as a known issue so that customers encounter the issue can work around


thanks

PR #6345, opened 2019-05-27
Cumulative go-xcat updates

neo954:
### The modification include

_##Fix get_package_list_apt() problem when multiple list files exist_
_##Enhance curl exit code parsing_


immarvin:
agree to merge, thanks @neo954 's contribution for xCAT

PR #6352, opened 2019-06-06
Change the build script for xcat-dep tar ball, due to internal environment changes 

whowutwut:
Internally, we've moved build servers to docker containers for xcat-core, so the build server that we used  to run build of deps package is no longer there. 

It would be far easier to be able to build the package from our git clone of `xcat-core` since it's just a key signing and tarup process.. 

This pull request makes changes to the build script  to allow for this script to be able  to run in our own home directories, given we have access to the files in the shared filesystem. 

Also taking this opportunity to clean up the build script and check non zero return code so it doesn't  run rogue.

### DIFF
--- a/builddep.sh
+++ b/builddep.sh
@@ -88,12 +88,13 @@ else
 	YUMDIR=htdocs
 fi
 
-cd `dirname $0`
+SCRIPT=$(readlink -f "$0")
+SCRIPTPATH=$(dirname "$SCRIPT")
+echo "INFO: Running script from here: $SCRIPTPATH ..."
+
+cd $SCRIPTPATH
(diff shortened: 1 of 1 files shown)


whowutwut:
Since we're running  in our home directory, have a hard stop if the sign keys are there... 

```
WARNING: The gnupg key dir: xxxx/vhu/.gnupg exists, it will be overwitten. Stop.
WARNING: To continue, remove it and rerun the script.
[vhu@c910loginx04 xcat-core]$
```

whowutwut:
This script was used to create a new build http://xcat.org/files/xcat/xcat-dep/2.x_Linux/xcat-dep-201906061433.tar.bz2 , merging the changes as it has no effect on the code base

Issue #4610, opened 2018-01-08
[Travis CI] use "ShellCheck" to validate the shell scripts in xCAT commits


immarvin:
`ShellCheck` is a GPLv3 tool that gives warnings and suggestions for bash/sh shell scripts, we can leverage it to impose a strict check on the shell scripts in xCAT

From https://github.com/koalaman/shellcheck: 
The goals of ShellCheck are

* To point out and clarify typical beginner's syntax issues that cause a shell to give cryptic error messages.

* To point out and clarify typical intermediate level semantic problems that cause a shell to behave strangely and counter-intuitively.

* To point out subtle caveats, corner cases and pitfalls that may cause an advanced user's otherwise working script to fail under future circumstances.
Travis CI Setup

If you want to use ShellCheck in Travis CI, you can most easily install it via apt:
````
language: bash
addons:
  apt:
    sources:
    - debian-sid    # Grab ShellCheck from the Debian repo
    packages:
    - shellcheck
````

daniceexi:
👍 Good to see the action that we are looking for the ways to improve ourselves.

hu-weihua:
@immarvin, thanks for your information. It is very helpful for test.

samveen:
Packages for shellcheck are also available for `RH`/`Centos` from `EPEL` (named `ShellCheck`).

samveen:
@immarvin @hu-weihua A lot of the scripts are expected to run on very diverse systems. For those systems that provide bash, it may be good to create a matrix of expected bash versions against the base systems as a lot of shell features are introduced by newer major Bash versions. 

In this particular case, I see a script that expects to work on both AIX and linux. However I am unsure of what bash version is available on AIX, which means that I cannot make changes to it because I might be introducing gotchas depending on the version of bash available on AIX.

immarvin:
hi @samveen , thanks for the explanation. 
I remember that AIX does not ship bash by default, postscripts for AIX are supposed to be run with `ksh`(or `sh` linked to `ksh`).  And now we do not cover AIX verification in each xcat release.

So I think you do not need to consider AIX support  while modifying the postscripts unless:
1) the postscript for AIX only, i.e, the postscripts with the name `aix*` or with she-bang `#!/bin/ksh`
2) the code slices for AIX only, i.e, the code in conditional branches like `if [ $OS == "AIX" ]; then`

[...]

PR #4923, opened 2018-03-12
Shellcheck improvements for genesis-scripts

samveen:
This PR is a partial fix for issue #4610 (the low hanging fruit so to say).

PR #6366, opened 2019-07-04
[genesis-scripts]Minor logic and piping improvements

samveen:
### The PR is to partially fix issue #4610

The code here was previously part of Pull request #4923 . For easier checks, this contains the easily-checked portions of the above PR. The code with logic flow changes from the previous PR will be added as another PR once this is merged.

### DIFF
--- a/xCAT-genesis-scripts/usr/bin/bmcsetup
+++ b/xCAT-genesis-scripts/usr/bin/bmcsetup
@@ -1,3 +1,4 @@
+#!/bin/bash
 # IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
 # Lenovo (c) 2016
 #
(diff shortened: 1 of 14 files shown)


gurevichmark:
@samveen So are the changes to the 14 files in this PR due to the suggestions from running ShellCheck ?

samveen:
@gurevichmark No it isn't just ShellCheck improvements, there are also some minor improvements that I had made prior to running ShellCheck. However all the changes here are a subset of pull request #4923 which I'd submitted previously. However, as the previous pull request also had some changes that didn't work properly on Power machines, therefore I've had to split that into 2, one this, and another one to be created (after this is merged), with the breaking changes, so that those can be reviewed and fixed more easily.

samveen:
@neo954 This PR is a breakdown of PR #4923 to remove all those sections that are not Power architecture specific (which is what held up the previous PR). Would you take a look and review?

PR #6375, opened 2019-07-16
Reduce maximum socket buffer size so select works

watson6282:
The maximum socket receive buffer size (SO_RCVBUF) is set from the value
in /proc/sys/net/core/rmem_max.  When this value is close to INT_MAX,
the select() system call used by IO::Select->can_read() never returns
true.  This causes packets to never be read.  Therefore, reduce the
maximum socket receive buffer size to a value which works.

### DIFF
--- a/perl-xCAT/xCAT/SLP.pm
+++ b/perl-xCAT/xCAT/SLP.pm
@@ -64,6 +64,11 @@ sub dodiscover {
             my $sysctl;
             open($sysctl, "<", "/proc/sys/net/core/rmem_max");
             my $maxrcvbuf = <$sysctl>;
+            # select() on a socket will never succeed if the buffer is too large (i.e. near INT_MAX)
+            my $cap_maxrcvbuf = 2047*1024*1024;
+            if ($maxrcvbuf > $cap_maxrcvbuf) {
+                $maxrcvbuf = $cap_maxrcvbuf;
+            }
             my $rcvbuf    = $args{'socket'}->sockopt(SO_RCVBUF);
             if ($maxrcvbuf > $rcvbuf) {
                 $args{'socket'}->sockopt(SO_RCVBUF, $maxrcvbuf / 2);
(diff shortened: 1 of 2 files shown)


cxhong:
@  what is the buffer size in the /proc/sys/net/core/rmem_max?

watson6282:
# grep rmem_max /etc/sysctl.conf
net.core.rmem_max = 2147483647

cxhong:
what kind of issue did u see when this occurred?   currently, we are debugging a issue on ESS cluster, all the xCAT r command (like rpower, etc) timeout, but ipmitool to BMC worked fine.

watson6282:
That is the exact same issue.  I'm one of the systems administrators from LLNL that reported this issue originally.  After much debugging, I found that this line never returned true and so never read the response packets coming from the BMCs.
 ```
/opt/xcat/lib/perl/xCAT/IPMI.pm line 541
    if ($select->can_read($timeout)) {
```
Eventually, I found that this code, starting at line 208, lead to this issue.
```
        if (-r "/proc/sys/net/core/rmem_max") { # we can detect the maximum allowed socket, read it.
            my $sysctl;
            open($sysctl, "<", "/proc/sys/net/core/rmem_max");
            my $maxrcvbuf = <$sysctl>;
            my $rcvbuf    = $socket->sockopt(SO_RCVBUF);
            if ($maxrcvbuf > $rcvbuf) {
                $socket->sockopt(SO_RCVBUF, $maxrcvbuf / 2);
            }
            $maxpending = $maxrcvbuf / 1500; #probably could have maxpending go higher, but just go with typical MTU as a guess
```
Setting the socket receive buffer to something close to INT_MAX (2147483647) causes the select() call in the can_read() function to always fail.  Therefore, I suggested this pull request to ensure the buffer never gets too big.

**Comment by cxhong** [comment]
That's the great NEWS!!!  Thanks so much..

Anyway,  I tried to change the rmem_max value on our test system,   didn't recreate it..  
```
[root@stratton01 ~]# grep rmem_max /etc/sysctl.conf
net.core.rmem_max = 2147483647
[root@stratton01 ~]# cat /proc/sys/net/core/rmem_max
2147483647
[root@stratton01 ~]# rpower boston38 state
boston38: on
```

cxhong:
This fixes resolved issue TS002251433,  will merge to get into 2.15 release

PR #6382, opened 2019-07-22
Remove kit links

gurevichmark:
### The modification include:

* Remove kit links from Ubuntu build
* Remove kit links from Linux build
* Remove kit related man page files (`*.pod` and `*.rst`)

PR #6385, opened 2019-07-24
Remove kits from Ubuntu and go-xcat

gurevichmark:
The release regression test did its job and caught a failure installing xCAT on Ubuntu. It appears there is still some dependency on the kits:
```
Package xcat-buildkit is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
```

However, the Travis CI passed in the PR #6382 

Removing some more references to `xcat-buildkit`


whowutwut:
@gurevich , as part of changes  to `go-xcat` there's a changelog at  the top as well as a Version number.  So we should update those accordingly.  

Also would need to upload this resulting  file to xcat.org as there's no automation there.

gurevichmark:
New commit added to update `go-xcat` version and change log.

PR #6391, opened 2019-07-26
Implement HMAC-SHA256 in xCAT IPMI

jjohnson42:
This will be compatible with some current implementations
that do not have SHA-1.  HMAC-SHA1 may be still secure, but
SHA1 in any context *looks* bad even if it isn't.


cxhong:
@jjohnson42 , how did u run unit test for this?

MBlaschek:
Not sure why it did not make into the 2.16.2 release?
I use 2.16.5 and it is not there. 
I manually changed your additions! It works. Thanks.

Obihoernchen (approved):
Works flawlessly with various BMCs with and without SHA256.

Issue #1848, opened 2016-09-16
[Feature Request] Policy table entries for groups


samveen:
Is it possible to implement policy checks on a `user group` level instead of on a `per user` level? An example is given below:

## Existing Situation
```
[samveen@xcatmn1 ~]$ echo toan samveen | xargs -n1 id
uid=1003(toan) gid=1001(SysEng) groups=1001(SysEng),1008(xCATEng)
uid=1004(samveen) gid=1001(SysEng) groups=1001(SysEng),1008(xCATEng)
[samveen@xcatmn1 ~]$ echo toan samveen | xargs -n1 sudo bash /opt/xcat/share/xcat/scripts/setup-local-client.sh
...
[samveen@xcatmn1 ~]$ echo toan samveen | xargs mkdef ...
...
[samveen@xcatmn1 ~]$ lsdef -t policy 7.1,7.2
Object name: 7.1
    name=toan
    rule=allow
Object name: 7.2
    name=samveen
    rule=allow
``` 

## What I want
```
[samveen@xcatmn1 ~]$ echo toan samveen | xargs -n1 id
uid=1003(toan) gid=1001(SysEng) groups=1001(SysEng),1008(xCATEng)
uid=1004(samveen) gid=1001(SysEng) groups=1001(SysEng),1008(xCATEng)
[samveen@xcatmn1 ~]$ echo toan samveen | xargs -n1 sudo bash /opt/xcat/share/xcat/scripts/setup-local-client.sh
...
[samveen@xcatmn1 ~]$  mkdef -t policy  -o 7.1 name=xCATEng rule=allow
...
[samveen@xcatmn1 ~]$  mkdef -t policy  -o 7.2 name=SysEng rule=allow commands=lsdef,nodels
[samveen@xcatmn1 ~]$ lsdef -t policy 7.1,7.2
Object name: 7.1
    name=xCATEng
    rule=allow
Object name: 7.2
    name=SysEng
    rule=allow
    commands=lsdef,nodels
``` 
I create policy entries for groups instead of users, which allows me to reduce my policy rule complexity to groups of users instead of all users. Now, if there are new members of the `xCATEng` or `SysEng` groups, I will just need to create their client access keys using `/opt/xcat/share/xcat/scripts/setup-local-client.sh` and that all: they will have access according to their groups:
```
[samveen@xcatmn1 ~]$ echo jill jane mary | xargs -n1 id
[...]

PR #6398, opened 2019-08-07
feat: add ability to create usergroup based policy rules

samveen:
### The PR is to implement feature #1848 

_##Feature description##_
In the case of lots of users (or user churn) of a cluster, it is easier to add a group based policy instead of creating individual policies for users.

Code logic:
* Given the request's `peername`, get the groups for the user.
* If the group list contains the rule name, it's a match, so carry forward as previously, or check for username and carry forward as previously.

### The UT result
As I no longer have access to an xCAT environment, I am unable to unit-test this.

### DIFF
--- a/perl-xCAT/xCAT/Utils.pm
+++ b/perl-xCAT/xCAT/Utils.pm
@@ -4987,6 +4987,26 @@ sub natural_sort_cmp($$) {
 
 #--------------------------------------------------------------------------------
 
+=head3  groups
+      Pure perl implementation of /bin/groups
+=cut
+
+#--------------------------------------------------------------------------------
+sub groups($) {
+    my ($name)=@_;
+    my @list;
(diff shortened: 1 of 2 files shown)


samveen:
@gurevichmark @whowutwut  would you please review this?

viniciusferrao:
I'm on it. Let's see if it will make the cut to 2.18.

viniciusferrao:
Superseded by #7786, rebased onto current master with @samveen credited as co-author. The replacement includes the corrected implementation, regression tests, and documentation. Thanks for the original contribution.

samveen:
@viniciusferrao Thank you for this. Makes life so much easier for cluster management :1st_place_medal: 

For reference: Fixed via https://github.com/xcat2/xcat-core/pull/7786

PR #6405, opened 2019-08-16
postscripts:  make remoteshell work with Match directives

khm:
currently remoteshell uses sed to delete lines from the sshd_config
file and appends new lines.  if there are Match blocks in the
sshd_config file, sshd will fail to start after remoteshell runs.

this change moves from 'delete then append' to using s/// commands
so sed directly edits the files in situ.  this way the configuration
directives remain in their original order, and sshd_config files
containing Match blocks do not break.

### DIFF
--- a/xCAT/postscripts/remoteshell
+++ b/xCAT/postscripts/remoteshell
@@ -59,15 +59,12 @@ then
 	logger -t $log_label -p local4.info "remoteshell:  setup /etc/ssh/sshd_config and ssh_config"
 	cp /etc/ssh/sshd_config /etc/ssh/sshd_config.ORIG
         #delete all occurance of the attribute and then add xCAT settings
-        sed -i '/X11Forwarding /'d /etc/ssh/sshd_config
-        echo "X11Forwarding yes" >>/etc/ssh/sshd_config
-        sed -i '/MaxStartups /'d /etc/ssh/sshd_config
-        echo "MaxStartups 1024" >>/etc/ssh/sshd_config
+        sed -i 's/X11Forwarding .*/X11Forwarding yes/' /etc/ssh/sshd_config
+        sed -i 's/MaxStartups .*/MaxStartups 1024/' /etc/ssh/sshd_config
 
     if [ "$SETUPFORPCM" = "1" ]; then
         if [[ $OSVER == sle* ]];then
-            sed -i '/PasswordAuthentication /'d /etc/ssh/sshd_config
-            echo "PasswordAuthentication yes" >>/etc/ssh/sshd_config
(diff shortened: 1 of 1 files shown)


samveen:
There is a problem with this, as the script will fail (spectacularly) in the following cases: 
- these patterns exist and have a preceding comment marker (`#`), which means that post the change the items are still commented.
- The config file is missing these directives, where the default in-built value of the directive is being utilized, in which case no directive will be added. 

The case to maintain the correct order of directives in the SSH config only applies if the directives are present to being with. How does that affect the config order in case the directives do not exist previously?

khm:
I wouldn't describe the failure as spectacular, as this same editing method is used elsewhere in xCAT, and nobody is selling tickets to witness the results.

Regardless, I've updated the editing logic to cover the following situations:
 - The config string is present and uncommented
 - The config string is present and commented out
 - The config string is absent and Match blocks are present in the config
 - The config string is absent and there are no Match blocks in the config

cxhong:
I just ran some testing on this,  if `#` on the line, the script didn't remove it
```
[root@f6u13k14 ssh]# grep X11Forwarding /etc/ssh/sshd_config
#       X11Forwarding no
[root@f6u13k14 ssh]# grep -q X11Forwarding /etc/ssh/sshd_config
[root@f6u13k14 ssh]# echo $?
0
[root@f6u13k14 ssh]# sed -i 's/[# ]\+\?X11Forwarding .*/X11Forwarding yes/' /etc/ssh/sshd_config
[root@f6u13k14 ssh]# echo $?
0
[root@f6u13k14 ssh]# grep X11Forwarding /etc/ssh/sshd_config
#       X11Forwarding yes             <---- it replaced `no` to `yes`, but `#` still here
````

Also, can u retest this line..  `sed -i '1,/Match/i MaxStartups 1024' /etc/ssh/sshd_config`
it inserted `MaxStartups 1024` before each line  
```
[root@f6u13k14 ssh]# grep X11Forwarding /etc/ssh/sshd_config
[root@f6u13k14 ssh]# grep Match /etc/ssh/sshd_config
#Match User anoncvs
[root@f6u13k14 ssh]# sed -i '1,/Match/i X11Forwarding yes' /etc/ssh/sshd_config
[root@f6u13k14 ssh]# grep Match /etc/ssh/sshd_config
#Match User anoncvs
[root@f6u13k14 ssh]# grep X11Forwarding /etc/ssh/sshd_config |wc
    144     288    2592                               <----- 144 of `X11Forwarding`
```

khm:
I'll open a new pull request with a more comprehensive solution in the future.

PR #6404, opened 2019-08-15
Remove conserver-xcat during uninstall

gurevichmark:
### The modification include

Add `conserver-xcat` to uninstall list, so that it gets removed during `go-xcat uninstall`

PR #6402, opened 2019-08-12
Add xCAT-buildkit to go-xcat uninstall list

gurevichmark:
### The modification include

Add `xCAT-buildkit` to the list of packages to uninstall, so that if uninstalling xCAT versions prior to 2.15, `xCAT-buildkit` will be removed.

PR #6437, opened 2019-10-03
Revert "Remove kits from Ubuntu and go-xcat"

gurevichmark:
This reverts 4 commits from PR #6382 #6385 
Used `git revert <commit id command>`

### DIFF
--- a/docs/source/guides/admin-guides/references/man1/addkit.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/addkit.1.rst
@@ -1,0 +1,137 @@
+
+########
+addkit.1
+########
+
+.. highlight:: perl
+
+
+****
+NAME
+****
+
+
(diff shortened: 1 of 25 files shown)


besawn:
@gurevichmark It seems like we should also revert this 5th commit (https://github.com/xcat2/xcat-core/pull/6385/commits/2eb57b1db5fabe836860e1c18afeeb0c5d03f207) contained in PR #6385:
image

Can you think of any reason why we would not want to revert this one too, so the changelog matches the current state of the code?

gurevichmark:
I did not revert that one because it would revert the "Version" number back to `1.0.40` from `1.0.41`
But there were other changes to `go-xcat` after kits were removed to bump it to current version of `1.0.43`: #6404 and #6402. But now thinking about it I realize I need to remove the comment from `go-xcat` saying that kits were removed.

gurevichmark:
Verified in today's build that kits are back.

PR #6394, opened 2019-08-01
Modify pkgdir for ubuntu16.04.5 x86 testcase

gurevichmark:
Fixing daily regression test failure for ubuntu16.04.5 x86 testcase.
Item `1` from https://github.ibm.com/xcat2/team_process/issues/156

PR #6449, opened 2019-10-16
Undo ubuntu testcases

gurevichmark:
Revert two PR #6394 and #6396 
They might be the reason for recent Ubuntu 16 daily regression test failures.


cxhong:
Thanks, @gurevichmark ,  hopefully tomorrow's regression test will pass.

PR #6456, opened 2019-10-23
Add regression test tabdump debug for service node not removed

gurevichmark:
The PR is to debug issue of failing daily regression test of left over service node which in turn causes `makedhcp` command to fail.
Failing environment: `rhels7.6-P8LPARs-Mysql-master-daily`
Success environment: `rhels7.6-X86VMs-Mysql-master-daily`

Adding `tabdump servicenode` command in different places to the regression bundle so we can see where the service node gets removed in the success environment and where is does not in failing environment.


cxhong:
add more `tabdump_servicenode` on each group of test cases
````
SN_setup_case
tabdump_servicenode
reg_linux_diskless_installation_hierarchy
tabdump_servicenode
reg_linux_diskfull_installation_hierarchy
tabdump_servicenode
updatenode........ (there are few cases here, want to add one after that)
tabdump_servicenode
```
````

PR #6477, opened 2019-11-08
Port Generic CentOS check from Lenovo tree

jjohnson42:
This will do a more generic check if a discid is not found.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/anaconda.pm
+++ b/xCAT-server/lib/xcat/plugins/anaconda.pm
@@ -2064,6 +2064,8 @@ sub copycd
         {
             $distname = $xCAT::data::discinfo::distnames{$did};
         }
+    } elsif ($desc and $desc =~ /CentOS Linux (.*)/) {
+          $distname = "centos" . $1;
     }
 
     unless ($dno) {


cxhong:
@jjohnson42 , `copycds` still failed with this fix
```
# diff anaconda.pm anaconda.pm.orig
2067,2068d2066
<     } elsif ($desc and $desc =~ /CentOS Linux (.*)/) {
<           $distname = "centos" . $1;

]# copycds CentOS-7-power9-Minimal-1810.iso
Error: [f6u13k13]: copycds could not identify the ISO supplied, you may wish to try -n <osver>
```

cxhong:
The `CentOS-7-power9-Minimal-1810.iso`  doesn't contain `.discinfo`,  that's why failed.

with @jjohnson42  fixes, and if iso file contains .discinfo,  the `copycds` works
```
# ls -ltr .discinfo
-r--r--r-- 1 root root 44 Aug 15 17:51 .discinfo
# cat .discinfo
1565905914.062055
CentOS Linux 8
x86_64
ALL
````

jjohnson42:
What does discinfo look like on that CentOS-7-power9-Minimal?

Does it work on x86 centos 8?

cxhong:
somehow, `CentOS-7-power9-Minimal-1810.iso` doesn't contain `.discinfo` file.  not sure where we downloaded before.  
`CentOS-8-x86_64-1905-dvd1.iso` has `.discinfo` file,  copycds works with your fixes.

bybai:
I think CentOS-7-power9-Minimal-1810.iso does not have wget in its repo, which can cause diskfull installation failure, how do you handle this problem?

PR #6488, opened 2019-11-21
Remove /install/autoinst config files as part of nodepurge

gurevichmark:
### The modification include

* For each node in the noderange, check if there are `/install/autoinst/<node>` or  `/install/autoinst/<node>.pre` or  `/install/autoinst/<node>.post` files. If there are, remove them during `nodepurge` command processing.

* Update man page for `nodepurge` command.


gurevichmark:
Good point. I do not think we do. I'll add one in a different PR.

PR #6530, opened 2020-01-15
Fix service.sle15.tmpl

Obihoernchen:
product was missing and script paths were wrong


cxhong:
Thanks, @Obihoernchen

PR #6538, opened 2020-01-17
SLE15: genimage fixes

Obihoernchen:
Please check carefully!!!

**1) Fix $osver_host detection**
It looks like this never worked correctly.

The version strings on SLE(S) 11, 12 and 15 look like this:
```
VERSION="11.4"
VERSION="12-SP3"
VERSION="15-SP1"
```
The old regex `/VERSION = (\d+)/` didn't match at all.
New regex: `/VERSION="(\d+)/` should work for all versions.

**2) Fix special mounts**
- Mount /sys and /proc
- needed for newer versions of the SLE permissions RPM. See: openSUSE/permissions@2d0ef5d Otherwise some RPMs fail to run post install RPM scripts
- Create /sys to have /sys/kernel/fscaps. Fixes: "Warning: running kernel does not support fscaps"
- Fix permissions of /dev devices
- Make sure /sys and /proc are unmounted everytime

**3) Fix genimage with existing passwd**
- from https://github.com/jjohnson42/xcat-core/commit/3110dc8fd781c704c53052c4d5ff1f426fdb22c4

**4) Combine compute.sle15.x86_64.pkglist  and compute.sle15.pkglist**

**5) Actually set up ifcfg correctly in SLE15 (fix dhclient errors)**

**6) Fix /etc/resolv.conf overwrite**
In SLE15 /etc/resolv.conf is a symlink to /run/netconfig/resolv.conf.
The package sysconfig-netconfig creates the symlink to /run/netconfig/resolv.conf on the host OS.
It does not stay in the chroot environment.
Therefore, genimage overwrites /etc/resolv.conf of the host.
Let's not create a dummy file if /etc/resolv.conf is a symlink.


Obihoernchen:
This was tested with SLE15-SP1 and latest updates.

cxhong:
I tested this on the SLE15 ppc64 system, didn't see any issues.

cxhong:
Thanks, @Obihoernchen

PR #6539, opened 2020-01-17
dhcp.pm fix

po1vo:
dhcp.pm claimed to not add ip address of a node, overlapping with dhcp dynamic range, but added anyway

### DIFF
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -816,6 +816,8 @@ sub addnode
                                   ]
                             }
                         );
+                        print $omshell "close\n";
+                        next;
                     }
                 }
                 if ($lstatements)


gurevichmark:
@po1vo Thank you for the contribution.
Can you please post how this problem can be recreated ?

po1vo:
1. Pick an IP from a network dynamic dhcp range. 
1. Pick a node and set the node's `hosts.ip` to the IP from the first step. 
1. Do `makedhcp node` and the a message:
> "The ip address $ip of node $node overlaps with the DHCP dynamic range specified in networks table, will not add this ip address into dhcpd.leases file."

But the address is added to dhcpd.leases anyway

samveen:
This may break existing installations of xCAT. A release note/warning in the next version would be a good idea if this is merged.

gurevichmark:
@samveen Can you elaborate under which conditions this PR will be a problem ?

samveen:
The potential break comes in the differing expectations:
- The user expects that the assigned IP is statically assigned to the node, as the node has the value set in `hosts.ip` (thanks to lease record in dhcpd.leases).
- The cluster expects that the node IP should not be in the dynamic range.

I prefer that the node configuration should override the cluster expectation, as it is viewable and thus expected configuration. 

Without such a static lease, in the case of a node reboot/restart, the node may not get the same IP, which is potentially disastrous from the cluster's perspective.

viniciusferrao:
To be honest, the right interface is that we cannot reserve addresses inside the dynamic range. It exists for a reason, and we should not overlap it. Yes, ISC DHCP allows that by default, and yes, it works. However, it is not good practice.

Also, we are now shipping Kea as a DHCP backend, and Kea does not allow that by default, which leaves us in an inconsistent state where different DHCP implementations result in different outcomes.

The right thing here is to error out with a message saying it isn't allowed.

PR #6540, opened 2020-01-17
SLE15: Merge some stuff from Lenovo branch

Obihoernchen:
From https://github.com/jjohnson42/xcat-core/tree/lenovobuild2.14

### DIFF
--- a/xCAT-server/share/xcat/install/sles/compute.sle15.tmpl
+++ b/xCAT-server/share/xcat/install/sles/compute.sle15.tmpl
@@ -12,7 +12,7 @@
     </bootloader>
     <general>
       <clock>
-        <hwclock>GMT</hwclock>
+        <hwclock>UTC</hwclock>
         <timezone>#TABLE:site:key=timezone:value#</timezone>
       </clock>
       <keyboard>
--- a/xCAT-server/lib/xcat/plugins/sles.pm
+++ b/xCAT-server/lib/xcat/plugins/sles.pm
@@ -1753,10 +1753,22 @@ sub copycd
         }
         if ($dsc =~ /Installer/ and $dsc =~ /SLE-15/) {
             $discnumber = 1;
-            unless ($distname) { $distname = "sle15" };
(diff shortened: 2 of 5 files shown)


xCAT-server/share/xcat/install/sles/compute.sle15.tmpl
```diff
@@ -45,7 +45,9 @@
     </partitioning>
     <add-on>
         <add_on_products config:type="list">
-            #INSTALL_SOURCES# 
+            <listentry><media_url>http://XCATNEXTSERVERHOOK/install/#TABLE:nodetype:THISNODE:os#/x86_64/2</media_url><product>sle-module-basesystem</product><product_dir>/Module-Basesystem</product_dir></listentry>
+            <listentry><media_url>http://XCATNEXTSERVERHOOK/install/#TABLE:nodetype:THISNODE:os#/x86_64/2</media_url><product>sle-module-hpc</product><product_dir>/Module-HPC</product_dir></listentry>
+            <listentry><media_url>http://XCATNEXTSERVERHOOK/install/#TABLE:nodetype:THISNODE:os#/x86_64/2</media_url><product>sle-module-server-applications</product><product_dir>/Module-Server-Applications</product_dir></listentry>
         </add_on_products>
```
cxhong: is there reason why you only wants those three?

Obihoernchen: No not really. I just copied this from Lenovo branch. We can include all, but usually this should be sufficient for most compute nodes images I guess.

cxhong: I think we like to include all.  Can u put back `#INSTALL_SOURCES#` ?

Obihoernchen: done

cxhong: Thanks

PR #6557, opened 2020-01-31
configib: pass NMCLI_USED from confignetwork instead of checking again

kcgthb:
The `configib` postscript currently uses its own method to determine if NetworkManager should be used or not. The other network configuration scripts (like `configeth`) get a `NMCLI_USED` variable insted, that is set in `confignetwork` in a more robust fashion.

This PR let `configib` use the same mechanism, by calling it from `confignetwork` with:
```
NMCLI_USED=$networkmanager_active [...] configib
```


cxhong:
Thanks, @kcgthb

Issue #6534, opened 2020-01-15
SLE15 detection is broken


Obihoernchen:
SLE15 replaced `/etc/SuSE-release` with `/etc/os-release`.

Example: https://github.com/xcat2/xcat-core/pull/6533

cxhong:
Thanks, @Obihoernchen ,  there are few more place/commands we need to check too.

Obihoernchen:
confignetwork: https://github.com/xcat2/xcat-core/blob/master/xCAT/postscripts/confignetwork#L42

cxhong:
@Obihoernchen , are u working on hierarchy support for sle15 too?

PR #6570, opened 2020-02-17
Modify the test cases to support SLE15

cxhong:
Failed to set up service node during the SLE15 regression test.  it due to `otherpkgs` didn't run because osimage for service node has wrong path of otherpkglist
```
# lsdef -t osimage sle15-ppc64le-install-service
Object name: sle15-ppc64le-install-service
    imagetype=linux
    osarch=ppc64le
    osdistroname=sle15-ppc64le
    osname=Linux
    osvers=sle15
    otherpkgdir=/install/post/otherpkgs/sle15/ppc64le
    otherpkglist=/opt/xcat/share/xcat/install//service.sle15.ppc64le.otherpkgs.pkglist  <<<----
    pkgdir=/install/sle15/ppc64le
    pkglist=/opt/xcat/share/xcat/install/sle/service.sle15.pkglist
    postscripts=servicenode
    profile=service
    provmethod=install
    template=/opt/xcat/share/xcat/install/sle/service.sle15.tmpl
```
The correct the `otherpkglist` should be 
```
otherpkglist=/opt/xcat/share/xcat/install/sle/service.sle15.ppc64le.otherpkgs.pkglist
```

### DIFF
--- a/xCAT-test/autotest/testcase/installation/SN_setup_case
+++ b/xCAT-test/autotest/testcase/installation/SN_setup_case
@@ -46,13 +46,13 @@ check:rc==0
 cmd:cd /install/post/otherpkgs/__GETNODEATTR($$SN,os)__/__GETNODEATTR($$SN,arch)__/xcat/xcat-core && createrepo .
 check:rc==0
 
-cmd:if [[ "__GETNODEATTR($$SN,os)__" =~ "rh" ]]; then path="rh";elif [[ "__GETNODEATTR($$SN,os)__" =~ "sles" ]];then path="sles";fi; ver="__GETNODEATTR($$SN,os)__"; tmp=${ver%%.*};ver=`echo "$tmp"|sed 's:[a-zA-Z]::g'`;cd /install/post/otherpkgs/__GETNODEATTR($$SN,os)__/__GETNODEATTR($$SN,arch)__/xcat/xcat-dep/$path$ver/__GETNODEATTR($$SN,arch)__ && createrepo .;
+cmd:if [[ "__GETNODEATTR($$SN,os)__" =~ "rh" ]]; then path="rh";elif [[ "__GETNODEATTR($$SN,os)__" =~ "sle" ]];then path="sles";fi; ver="__GETNODEATTR($$SN,os)__"; tmp=${ver%%.*};ver=`echo "$tmp"|sed 's:[a-zA-Z]::g'`;cd /install/post/otherpkgs/__GETNODEATTR($$SN,os)__/__GETNODEATTR($$SN,arch)__/xcat/xcat-dep/$path$ver/__GETNODEATTR($$SN,arch)__ && createrepo .;
 check:rc==0
 
 cmd:chdef -t osimage __GETNODEATTR($$SN,os)__-__GETNODEATTR($$SN,arch)__-install-service otherpkgdir=/install/post/otherpkgs/__GETNODEATTR($$SN,os)__/__GETNODEATTR($$SN,arch)__
 check:rc==0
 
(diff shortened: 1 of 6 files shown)


cxhong:
This PR should fix most cases for the P9.
will fix some of cases for the x86 if didn't involve hierarchy support.  SLE15 x86_64 failed on `mysqlsetup` and service node didn't start up.

cxhong:
for the SLE15, the `dhcpd.conf` change to `/etc/dhcpd.conf` and has `/etc/dhcp` dir.  the logical of code for the `makedhcp`
1) if there is `/etc/dhcpd.conf`, it will use that dhcp conf file
2) if there is no `/etc/hdcpd.conf` and has `/etc/dhcp` dir,  it will assume dhcp conf file will be `/etc/dhcp/dhcpd.conf`

for sles12, it located at the `/etc/dhcp/dhcpd.conf`
for redhat, it located at the `/etc/dhcpd.conf`, but there is no `/etc/dhcp` dir.

In the `makedhcp` test cases, it removes `/etc/dhcpd.conf` to `/etc/dhcpd.conf.bak`,  will cause `makedhcp` command to choose dhcp conf file be `/etc/dhcp/dhcpd.conf`,  then dhcpd will not able to start.

cxhong:
for the begin of test cases, we move dhcpd.conf to dhcpd.conf.bak,  so the end of test cases, we always wants to restore the old files back.  that's why  I switched `dhdcpd.conf.back` to `dhcpd.conf`

peterwywong:
I still have many questions.

(1) The conditional statement of makedhcp_n_linux is as follows:
      if there is /etc/dhcp/dhcpd.conf, then rename dhcpd.conf to dhcpd.conf.bak (proposed, mv is changed to cp);
      else if there is /etc/dhcpd.conf, then rename the same (proposed: mv is changed to cp).

I could read your description above incorrectly. It seems your sequence of checking is the reverse.

The above logic does not check /etc/dhcp alone, it checks whether /etc/dhcp/dhcpd.conf is there as well. It would fail in the case of SLES 15 and then it goes to the else statement to see whether /etc/dhcpd.conf is there or not. It should be there.

Since Red Hat has /etc/dhcpd.conf (the same as SLES 15) and it runs fine, can you help us understand why it fails for SLES 15? Definitely, I miss something about /etc/dhcp.

(2) The reason for changing mv to cp is to keep dhcpd.conf. Why is it OK for other distros NOT to have dhcp.conf for this test? They work fine with dhcp.conf.bak.

(3) Is dhcpd.conf.bak modified during testing? If not, why do we need to "mv" it back to dhcpd.conf. dhcpd.conf is still there based on the use of "cp".

The code is new to me and I want to learn the details of the code. Hope you don't mind my many questions :) .

cxhong:
please check the `./xCAT-server/lib/xcat/plugins/dhcp.pm`  
```
my $dhcpconffile = $^O eq 'aix' ? '/etc/dhcpsd.cnf' : '/etc/dhcpd.conf';
my %dynamicranges; #track dynamic ranges defined to see if a host that resolves is actually a dynamic address
my %netcfgs;
my $distro = xCAT::Utils->osver();
my $checkdomain=0;

# dhcp 4.x will use /etc/dhcp/dhcpd.conf as the config file
my $dhcp6conffile;
if ($^O ne 'aix' and not -e $dhcpconffile and -d "/etc/dhcp") {
    $dhcpconffile  = '/etc/dhcp/dhcpd.conf';
    $dhcp6conffile = '/etc/dhcp/dhcpd6.conf';
}
````
the difference between redhat and sle15 is redhat doesn't have `/etc/dhcp/` directory,  sle15 has `/etc/dhcp/` directory,  if `/etc/dhcpd.conf` file is not available,  from the above the code,  `/etc/dhcp/dhcpd.conf` will be create for the sle15

3)  begin of test case,  we always keeps copy of origin files,  `makedhcp` will modify the `dhcpd.conf`,  that's why at the end of test case,  we want to origin file back

cxhong:
part of fixes for postscripts are for issue #6534

PR #6574, opened 2020-02-19
requirement of "aufs" added to "squashfs" method

sventor:
As some mainstream distributions don't provide "aufs" for their kernels, at least a hint to that essential requirement is prudent, I think.


gurevichmark:
Resolved by #6574

Issue #6576, opened 2020-02-20
copycds doesn't recognize CentOS 8.x ISO images


conxuro:
Trying to generate an image from Centos 8.0 and/or CentOS 8.1 ISO results in error.

```
$ copycds CentOS-8-x86_64-1905-dvd1.iso 
Error: [hpchead]: copycds could not identify the ISO supplied, you may wish to try -n <osver>

$ copycds CentOS-8.1.1911-x86_64-dvd1.iso 
Error: [hpchead]: copycds could not identify the ISO supplied, you may wish to try -n <osver>
```

It works with -n centos8.0 (or centos8.1) -a x86_64 but doesn't recognize ISO automatically like with CentOS 7.x ISO images.

---

Software used:
```
$ cat /etc/redhat-release 
CentOS Linux release 7.7.1908 (Core)

$ xcatconfig -v
Version 2.15 (git commit 218c6d3acc8bdbd7f72115e48cda2b1a3613d18a, built Mon Nov  4 15:17:59 EST 2019)
```

gurevichmark:
@conxuro 
CentOS 8.0 appears to be working with the latest development version of xCAT

```
[root@stratton01 ~]# lsxcatd -a
Version 2.15.1 (git commit 4dd25923817d6f4303fbddab3344f7d118815c98, built Thu Feb 20 00:34:44 EST 2020)
This is a Management Node
dbengine=SQLite
```
```
[root@stratton01 ~]# copycds /mnt/xcat/iso/centos/8.0/CentOS-8-x86_64-1905-dvd1.iso
Copying media to /install/centos8/x86_64
Media copy operation successful

[root@stratton01 ~]# lsdef -t osimage centos8-x86_64-install-compute
Object name: centos8-x86_64-install-compute
    imagetype=linux
    osarch=x86_64
    osdistroname=centos8-x86_64
    osname=Linux
    osvers=centos8
    otherpkgdir=/install/post/otherpkgs/centos8/x86_64
    pkgdir=/install/centos8/x86_64
    pkglist=/opt/xcat/share/xcat/install/centos/compute.centos8.pkglist
    profile=compute
    provmethod=install
    template=/opt/xcat/share/xcat/install/centos/compute.centos8.tmpl
[root@stratton01 ~]#
```

But for CentOS 8.1, `copycds` recognizes the osname and arch, but creates the osimage definition with the same name as CentOS8.0
```
[...]

PR #6581, opened 2020-02-20
Add iso id for CentOS 8.1

gurevichmark:
### The PR is to fix issue _#6576_

Add CentOS 8.1 identifier to distinguish between 8.0 and 8.1 packages:

```
[root@stratton01 xcat]#copycds /mnt/xcat/iso/centos/8.1/CentOS-8.1.1911-x86_64-dvd1.iso
Copying media to /install/centos8.1/x86_64
Media copy operation successful

Object name: centos8.1-x86_64-install-compute
    imagetype=linux
    osarch=x86_64
    osdistroname=centos8.1-x86_64
    osname=Linux
    osvers=centos8.1
    otherpkgdir=/install/post/otherpkgs/centos8.1/x86_64
    pkgdir=/install/centos8.1/x86_64
    pkglist=/opt/xcat/share/xcat/install/centos/compute.centos8.pkglist
    profile=compute
    provmethod=install
    template=/opt/xcat/share/xcat/install/centos/compute.centos8.tmpl

[root@stratton01 xcat]# ls -l /install/centos8.1/x86_64/BaseOS/Packages/centos-release*
-r--r--r-- 1 root root 21048 Jan  2 14:34 /install/centos8.1/x86_64/BaseOS/Packages/centos-release-8.1-1.1911.0.8.el8.x86_64.rpm
[root@stratton01 xcat]#

### CODE REVIEW

File: perl-xCAT/xCAT/data/discinfo.pm [hunk 1]
```diff
@@ -14,6 +14,7 @@ require Exporter;
 use strict;
 
 our %distnames = (
+    "1578087695.538488" => "centos8.1",     #x86_64
```
besawn: Please add the identifier for the ppc64le centos8.1 iso to this PR, too.

gurevichmark:
Adding entry for `ppc64le`
```
[root@stratton01 8.1]# mount -o loop CentOS-8.1.1911-ppc64le-dvd1.iso /tmp/gurevich/redhat_iso/
mount: /dev/loop1 is write-protected, mounting read-only

[root@stratton01 8.1]# cat /tmp/gurevich/redhat_iso/.discinfo                   
1578087695.537220
CentOS Linux 8
ppc64le
ALL
[root@stratton01 8.1]#
```

peterwywong:
I see centos6, centos7.0, rhels7 are in the distnames table.
Why don't we have centos8 there? I can see centos8.1 is needed.

gurevichmark:
@peterwywong 
Most of the time we can just read a string description like `CentOS Linux 8` below
```
[root@stratton01 8.0]# cat /tmp/gurevich/redhat_iso/.discinfo                   
1578087695.537220
CentOS Linux 8
ppc64le
ALL
[root@stratton01 8.0]#
```
But sometimes those files do not have the correct info to determine the OS version. For those cases we need to use the numeric identifier. See referenced #6576 for an example of when the string description was not correct.

peterwywong:
Could you point me to the code which does the parsing of discinfo?

gurevichmark:
There is a function `copycd()` in file https://github.com/xcat2/xcat-core/blob/master/xCAT-server/lib/xcat/plugins/anaconda.pm starting on line 1974.
On line 2038 is mounts the `.iso` file and opens the `.discinfo` from that `.iso`
Lines 2040-2047 read 4 lines from  `.discinfo` file into variables. First one ($did) is the disk identifier.
Then on line 2061 that $did is used to lookup an entry in the map `distnames` stored in https://github.com/xcat2/xcat-core/blob/master/perl-xCAT/xCAT/data/discinfo.pm

Issue #6618, opened 2020-03-04
mysqlsetup failed on rhels7.7


cxhong:
for the rhels7.7 release regression test,  It failed on the mysqlsetup
```
# mysqlsetup -i -V
Running command on c910f03c09k09: rpm -qa | grep -i mariadb 2>&1

Running command on c910f03c09k09: rpm -qa | grep -i perl-DBD-mysql 2>&1

Running command on c910f03c09k09: ps -ef | grep mysqld 2>&1

Running command on c910f03c09k09: pidof mysqld 2>&1

Input the alpha-numberic  password for xcatadmin in the MySQL database:
Input the password for root in the MySQL database:
Running command on c910f03c09k09: sed 's/^bind/#&/' /etc/my.cnf > /tmp/my.cnf; mv -f /tmp/my.cnf /etc/my.cnf;chmod 644 /etc/my.cnf 2>&1

Running command on c910f03c09k09: /usr/bin/mysql_install_db --user=mysql 2>&1

Command failed: /usr/bin/mysql_install_db --user=mysql 2>&1. Error message: Neither host 'c910f03c09k09' nor 'localhost' could be looked up with
'/usr/libexec/resolveip'
Please configure the 'hostname' command to return a correct
hostname.
If you want to solve this at a later stage, restart this script
with the --force option

The latest information about mysql_install_db is available at
https://mariadb.com/kb/en/installing-system-tables-mysql_install_db.

 /usr/bin/mysql_install_db --user=mysql failed.
```
File `/usr/libexec/resolveip` is not available on the rhels7.7
```
# ls -ltr /usr/libexec/resolveip
ls: cannot access /usr/libexec/resolveip: No such file or directory
````
`mysqlsetup` worked fine on the rhels7.6 and rhels8.0.  It looks like `/usr/bin/mysql_install_db` file on rhels7.7 is different than others.
```
  dirname0=`dirname $0 2>/dev/null`                     <-----dirname=/usr/bin
dirname0=`dirname $dirname0 2>/dev/null`          <------dirname=/usr

# relative from where the script was run for a relocatable install
elif test -n "$dirname0" -a -x "$dirname0/libexec/mysqld"
then
  basedir="$dirname0"                          <----basedir=/usr
  bindir="$basedir/libexec"                    <-----binddir=/usr/libexec
  resolveip="$bindir/resolveip"               <----resolvip=/usr/libexec/resolveip
  mysqld="$basedir/libexec/mysqld"
[...]

PR #6620, opened 2020-03-04
Modify mysqlsetup scripts on rhels7.7

cxhong:
The PR is to fix issue  #6618 

Add `--force` option to `/usr/bin/mysql_install_db` for rhels7.7 only.
```
 --force              Causes mysql_install_db to run even if DNS does not
                       work.  In that case, grant table entries that
                       normally use hostnames will use IP addresses.
 ```
If `--force` is set, it will not call `/usr/libexec/resolveip`,  and this function only used here to check if ip address can resolved the hostname.  
```
[root@c910f03c09k09 ~]# nslookup 10.3.9.9
9.9.3.10.in-addr.arpa   name = c910f03c09k09.pok.stglabs.ibm.com.

[root@c910f03c09k09 ~]# /usr/bin/resolveip 10.3.9.9
Host name of 10.3.9.9 is c910f03c09k09, c910f03c09k09.pok.stglabs.ibm.com
```

### DIFF
--- a/xCAT-client/bin/mysqlsetup
+++ b/xCAT-client/bin/mysqlsetup
@@ -930,9 +930,12 @@ sub initmysqldb
         $cmd = "$sqlcmd --user=mysql";
         #on rhels7.7, /usr/bin/mysql_install_db requires /usr/libexec/resolveip, 
         #but it's available at the /usr/bin/resolveip
-        #use --force to grant table entries use ipaddress instead of hostname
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by cxhong** [comment]
change to use symbol link as per rhels7.7 release notes
```
]# ls -ltr /usr/libexec/resolveip
lrwxrwxrwx 1 root root 18 Mar  4 13:27 /usr/libexec/resolveip -> /usr/bin/resolveip
```

Issue #6642, opened 2020-03-20
Package perl-IO-Stty is not automatically installed by xCAT 2.15.1


gurevichmark:
The package `perl-IO-Stty` does not automatically get installed when xCAT 2.15.1 is installed (incorrectly removed by #6504), which results in the following error:
```
[root@sn02 Downloads]# rscan -z hmc02
Error: [sn02]: hmc plugin bug, pid 112435, process description: 'xcatd SSL: rscan to hmc02 for root@localhost: hmc instance' with error 'Can't locate object method "stty" via package "IO::Tty" at /opt/xcat/lib/perl/xCAT/PPCcli.pm line 215.
' while trying to fulfill request for the following nodes: hmc02
[root@sn02 Downloads]#
```

The workaround is to manually install `perl-IO-Stty`, which is available in `xcat-deps` and restart xCAT.

cxhong:
revert PR #6504 

for sle15,  download `perl-IO-Stty-0.04-26.1.noarch.rpm` from https://download.opensuse.org/repositories/devel:/languages:/perl/SLE_15/noarch/   and will put to xcat-dep for sle15

PR #6643, opened 2020-03-20
Revert "Remove perl-IO-Stty requirement from xCAT spec"

cxhong:
Reverts xcat2/xcat-core#6504  for issue #6642


cxhong:
for sle15, download perl-IO-Stty-0.04-26.1.noarch.rpm from https://download.opensuse.org/repositories/devel:/languages:/perl/SLE_15/noarch/ and will put to xcat-dep for sle15

PR #6675, opened 2020-04-20
Testcase improvements

gurevichmark:
* Display better informational message from `nodeset_runimg` testcase.
* Run `retry_install` only 3 times instead of default 5. Some testcases could take more than 10 hours if retrying 5 times.

### DIFF
--- a/xCAT-test/autotest/testcase/commoncmd/retry_install.sh
+++ b/xCAT-test/autotest/testcase/commoncmd/retry_install.sh
@@ -12,7 +12,7 @@ then
     times=$3+1
     echo "Try to rinstall for $3 times ......"
 else
-    times=6
+    times=4
     echo "Try to rinstall for 5 times ......" 
 fi
 
--- a/xCAT-test/autotest/testcase/commoncmd/retry_install.sh
+++ b/xCAT-test/autotest/testcase/commoncmd/retry_install.sh
@@ -6,18 +6,16 @@ declare -i tryreinstall=1
 node=$1
 osimage=$2
 vmhost=`lsdef $node -i vmhost -c | cut -d '=' -f 2`
+times=3
 
 if [ $# -eq 3 ];
 then
-    times=$3+1
-    echo "Try to rinstall for $3 times ......"
-else
-    times=4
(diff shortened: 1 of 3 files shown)


xCAT-test/autotest/testcase/packimg/cases0
```diff
@@ -76,7 +76,7 @@ check:output=~archive method:cpio
 check:output=~compress method:gzip
 cmd:ls -l /install/netboot/__GETNODEATTR($$CN,os)__/__GETNODEATTR($$CN,arch)__/compute/rootimg.cpio.gz
 check:rc==0
-cmd:/opt/xcat/share/xcat/tools/autotest/testcase/commoncmd/retry_install.sh $$CN __GETNODEATTR($$CN,os)__-__GETNODEATTR($$CN,arch)__-netboot-compute
+cmd:/opt/xcat/share/xcat/tools/autotest/testcase/commoncmd/retry_install.sh $$CN __GETNODEATTR($$CN,os)__-__GETNODEATTR($$CN,arch)__-netboot-compute 3
```
cxhong, line 79: instead of change each test case, can we make `retry_install.sh` default to 3 times?

gurevichmark: Yea, that is even better idea. I will make the change.

xCAT-test/autotest/testcase/commoncmd/retry_install.sh
```diff
@@ -12,7 +12,7 @@ then
     times=$3+1
     echo "Try to rinstall for $3 times ......"
 else
-    times=6
+    times=4
     echo "Try to rinstall for 5 times ......"
```
cxhong, line 16: do u want to change this to 3 times?

gurevichmark: It will run 3 times, the code just uses the $times variable for counting:
`for (( tryreinstall = 1 ; tryreinstall < $times ; ++tryreinstall ))`

gurevichmark: But maybe I should change it so it is more clear.

PR #6685, opened 2020-04-29
Make sure squashfs installed on SLES for squashfs testcase

gurevichmark:
On SLES `squashfs` is not installed by default.
It is needed to run `squashfs` testcases.


cxhong:
maybe we can add `squashfs` rpm to compute pkglist.

PR #6505, opened 2019-12-09
Fix insecure processing of XML input

jjohnson42:
XML::Simple is not up to the task.  Switch to
XML::LibXML::Simple and coax it to not do as much
XML to avoid XXE attacks.

### The modification include

_##item1_

_##item2_

### The UT result
`##The UT output##`

# Please remove this line and below if fix issue

# Please remove this line and above for tasks or features

### The PR is for task _#xxx_ or to implement feature #xxx

_##Feature description##_

### The content of the PR:
* [ ] _##The mini-design link_
* [ ] _##The basic code logic_

### The UT result
`##The UT output##`

PR #6688, opened 2020-05-01
Use XML::LibXML::Simple instead of XML::Simple

gurevichmark:
Try to understand and recreate work in #6505 

Use some guidance for moving from `XML::Simple` to `XML::LibXML::Simple` from
https://www.perlmonks.org/?node_id=490846
https://perlmaven.com/how-to-preserve-xml-simple-element-order


gurevichmark:
This PR will replace #6505

The problem was simple. Need to `use XML::LibXML` instead of `use XML::LibXML::Simple`

PR #6700, opened 2020-05-10
xcattest: fix get_current_os and improve evironment logs

chenhanxiao:
s/rhels/rhel

also mention what may cause ‘Unsuitable current environment’

### DIFF
--- a/xCAT-test/xcattest
+++ b/xCAT-test/xcattest
@@ -1036,7 +1036,7 @@ sub load_case {
                     my @newvalidoslist = ();
                     foreach my $validos (@validoslist) {
                         if ($validos =~ /linux/i) {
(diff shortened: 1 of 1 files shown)


cxhong:
@chenhanxiao 
which test case did u hit the issue?  Can u show us test case and failed result?

chenhanxiao:
> @chenhanxiao
> which test case did u hit the issue? Can u show us test case and failed result?
Pls check:
https://github.com/xcat2/xcat-core/pull/6700/commits/8e12f3c9df9452a48cb52e8624e47d6360bdb926
 The commit msg of this PR didn't updated.
1) xcattest should provide the same list as get_current_os  did (NOT a bug)
2) improve logs as hints for newbie(e.g me)

gurevichmark:
Replaced by #6869

PR #6721, opened 2020-06-01
add namedincludes in site table documentation

kcgthb:
This PR adds the `namedincludes` directive from PR #4497 in the `site` table documentation.

PR #6723, opened 2020-06-01
Add namedincludes to site table documentation

gurevichmark:
Replaces #6721

Add `namedincludes` to site table documentation and man page.


kcgthb:
Ah my bad! Thanks for fixing this.

PR #6731, opened 2020-06-10
test: remove imgimport -v test case

chenhanxiao:
we don't support show version of imgimport.
-v means verbose for imgimport.
So delete this testcase.

Issue:
# imgexport -v
Error: [xcat]: Cannot find /root/-v

Reported-by: Yang Jian <fun_yang@foxmail.com>
Signed-off-by: Chen Hanxiao <chen_han_xiao@126.com>


gurevichmark:
Testcase in invalid, but also revealed other issues with `imgimport` and `imgexport` commands, which will be handled by #6745

PR #6732, opened 2020-06-12
Flush out deps repomd.xml

gurevichmark:
Sometimes `go_xcat_devel_from_repo` testcase fails on sles.
The error:
```
c910f04x12v07: Retrieving repository 'xcat-dep' metadata [..
c910f04x12v07: Signature verification failed for file 'repomd.xml' from repository 'xcat-dep'.
c910f04x12v07: Warning: This might be caused by a malicious change in the file!
c910f04x12v07: Continuing might be risky. Continue anyway? [yes/no] (no): no
c910f04x12v07: error]
c910f04x12v07: Repository 'xcat-dep' is invalid.
```

is probably because there is a cached version of `repomd.xml` somewhere on the network.

This PR calls `wget repomd.xml` before running `go-xcat` to try to flush out cached version of deps `repomd.xml` file.

### DIFF
--- a/xCAT-test/autotest/testcase/go_xcat/case3
+++ b/xCAT-test/autotest/testcase/go_xcat/case3
@@ -21,8 +21,9 @@ check:rc==0
 #Install additional packages on Red Hat
 cmd:if xdsh $$CN "grep \"Red Hat\" /etc/*release >/dev/null"; then xdsh $$CN "yum install -y yum-utils dnf-utils bzip2"; fi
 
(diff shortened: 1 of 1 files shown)


besawn:
Should this be done in `go-xcat` itself rather than as part of the test case?

gurevichmark:
Since it only appears to affect our lab machines behind some kind of proxy server, I did not want to pollute `go-xcat` script with it.

PR #6736, opened 2020-06-17
Add Cuda11 setup scripts

cxhong:
This PR is a workaround for cuda11 setup on the power 9 system. During  diskfull cuda11 installation,  nvidia-driver failed to install on the kickstart:
```
pre_msg: Preparing transaction from installation source
msg: Payload error - DNF installation has ended up abruptly: /bin/sh is needed by cuda-drivers-445.43-1.ppc64leTraceback (most recent call last):
  File "/usr/lib64/python3.6/site-packages/pyanaconda/payload/dnfpayload.py", line 283, in do_transaction
    base.do_transaction(display=display)
  File "/usr/lib/python3.6/site-packages/dnf/base.py", line 842, in do_transaction
    raise dnf.exceptions.TransactionCheckError(msg)
dnf.exceptions.TransactionCheckError: /bin/sh is needed by cuda-drivers-445.43-1.ppc64le
```
so, instead install cuda from pkglist,  xCAT provides `cuda11_power9_setup` which include cuda installation and `cuda_power9_setup`


xCAT-server/share/xcat/samples/cuda11/cuda_power9_setup
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
```
besawn, line 1: Why do we need to include `cuda_power9_setup` here?

cxhong: there are differences with the one we shipped from postscripts.  I indicated those two scripts are verified for the HPC software stack.

besawn:
I want to make a few updates to the included documentation before we merge this...

Issue #6748, opened 2020-06-24
enablekdump trashes the install NIC ifcfg file


mattaezell:
We use `confignetworks -s` to set our boot interface as static. It seems that later `enablekdump` re-enables DHCP on that interface. Is there any reason for that?

I think this was actually the root cause for #1972

cxhong:
We couldn't think any reason why `enablekdump` postscripts re-enables to DHCP .
this code is wrote since  2012.    I think i will like to make changes and keep boot interface as is.

bybai:
@mattaezell , please check if NetworkManager is enabled or not somewhere.

mattaezell:
> @mattaezell , please check if NetworkManager is enabled or not somewhere.

No, we don't have NetworkManager even installed in our compute image:

```
[root@a01n01 ~]# rpm -qa|grep -i NetworkManager
[root@a01n01 ~]#
```

gurevichmark:
@bybai Thanks for your input.
But I think the problem is in ` xcat-core/xCAT/postscripts/enablekdump`
It seems if network adapter does not start with `hf`, we set `/etc/sysconfig/network-scripts/ifcfg-` to dhcp.
![image](https://user-images.githubusercontent.com/16106630/86145699-3a708e00-bac5-11ea-906c-d0e4865bafc6.png)

We do not understand why we need to do that and are planning to remove this whole section of code so that `/etc/sysconfig/network-scripts/ifcfg-` will not be changed.

PR #6754, opened 2020-06-30
Do not modify installnic's interface in enablekdump postscripts

cxhong:
This PR is fix issue #6748 .

According the redhat8 documention guide:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/system_design_guide/installing-and-configuring-kdump_system-design-guide

I didn't find it's necessary to change interface `ifcfg` file to dhcp for kdump process,  so will keep interface as-is.


gurevichmark:
@cxhong Since you are updating the doc file, it seems this link is no longer valid, but if you replace `5` with `6`, it will work:

image

cxhong:
@gurevichmark , thanks, modified

PR #6780, opened 2020-07-24
Change the regex in post.xcat.ng to accommodate long file names

peterwywong:
By doing so, postscirpts with long file names can still be downloaded from MN/SN to CN.

"\1" is replaced by ".*" in the regular expression.

Due to the limited width (50-55) for href, the file name after > may not be completely captured. When the file name is not identical before and after ">", the file is not downloaded.

It is not clear to me why the file name appears before and after ">" by HTTPD.

".*" would match anything. As long as the file name is captured before ">" in href, the file is downloaded.

### DIFF
--- a/xCAT-server/share/xcat/install/scripts/post.xcat.ng
+++ b/xCAT-server/share/xcat/install/scripts/post.xcat.ng
@@ -134,7 +134,7 @@ function download_recursive()
             [ "$?" -ne "0" ] && return 1
             ;;
         esac
-    done < <(curl --fail --retry 20 --max-time 60 "${url}/" | grep -o '<a href="\([^"]*\)">\1</a>' | cut -d '"' -f 2)
+    done < <(curl --fail --retry 20 --max-time 60 "${url}/" | grep -o '<a href="\([^"]*\)">.*</a>' | cut -d '"' -f 2)
     return 0
 }
 
--- a/xCAT-server/share/xcat/install/scripts/post.xcat.ng
+++ b/xCAT-server/share/xcat/install/scripts/post.xcat.ng
@@ -134,7 +134,7 @@ function download_recursive()
             [ "$?" -ne "0" ] && return 1
             ;;
         esac
-    done < <(curl --fail --retry 20 --max-time 60 "${url}/" | grep -o '<a href="\([^"]*\)">.*</a>' | cut -d '"' -f 2)
+    done < <(curl --fail --retry 20 --max-time 60 "${url}/" | grep -o '<a href="\([^"]*\)">.*</a>' | egrep -v "O=D|Directory" | cut -d '"' -f 2)
     return 0
 }
 


peterwywong:
**Why RHEL 8.x jobs miss postscripts with long names, but others don't?**

RHEL 7.6 has the following template value: /opt/xcat/share/xcat/install/rh/compute.rhels7.tmpl
And /opt/xcat/share/xcat/install/rh/compute.rhels7.tmpl has the following INCLUDE statement:
#INCLUDE:#ENV:XCATROOT#/share/xcat/install/scripts/post.xcat#

SLES 12.3 has the following template value: /opt/xcat/share/xcat/install/sles/compute.sles12.ppc64le.tmpl
And /opt/xcat/share/xcat/install/sles/compute.sles12.ppc64le.tmpl has the following INCLUDE statement:
#INCLUDE:#ENV:XCATROOT#/share/xcat/install/scripts/post.sles11#
Which in turn has the following INCLUDE statement:
#INCLUDE:#ENV:XCATROOT#/share/xcat/install/scripts/post.xcat#

It means both of them point to install/scripts/post.xcat.

However, RHEL 8.x images are different.

RHEL 8.0 has the following template value: /opt/xcat/share/xcat/install/rh/compute.rhels8.tmpl
And /opt/xcat/share/xcat/install/rh/compute.rhels8.tmpl has the following INCLUDE statement:
#INCLUDE:#ENV:XCATROOT#/share/xcat/install/scripts/post.xcat.ng#

That is, RHEL 8.x run post.xcat.ng. Other OSes run post.xcat.

"post.xcat" uses the following wget command to copy /install/postscripts/* in MN to /xcatpost in CN:

wget -l inf -N -r --waitretry=10 --random-wait --retry-connrefused  -e robots=off -nH --cut-dirs=2 --reject "index.html*" --no-parent -t 20 -T 60 http://${MASTER_IP}:${HTTPPORT}${INSTALLDIR}/postscripts/ -P /xcatpost 2> /tmp/wget.log

"post.xcat.ng" uses Function download_recursive to download files, including a filtering regular expression:

function download_recursive()
{
    local url="$1"
    local dest_dir="$2"
    local f    [ "${url: -1}" = "/" ] && url="${url:0:-1}"
    [ "${dest_dir: -1}" = "/" ] && dest_dir="${dest_dir:0:-1}"
    mkdir -p "${dest_dir}"    while read -r f
    do
        case "$f" in
        *"/")
            download_recursive "${url}/${f}" "${dest_dir}/${f:0:-1}"
            [ "$?" -ne "0" ] && return 1
            ;;
        *)
            curl --fail --retry 20 --max-time 60 "${url}/${f}" -o "${dest_dir}/${f}"
            [ "$?" -ne "0" ] && return 1
            ;;
        esac
    done < <(curl --fail --retry 20 --max-time 60 "${url}/" | grep -o '<a href="\([^"]*\)">\1</a>' | cut -d '"' -f 2)
    return 0
}

curl is used to get all directories and files under /install/postscripts in MN and then a regular expression is used to select the directory and files.

The regular expression filters out files with long names. That is the problem.

peterwywong:
**Improvement over my original regex change which can copy postscripts of long names from MN to CN**

"curl" returns HTTP records for the to-be-copied directories and files. Here are two records: one has a shorter name and the other a longer one

<tr><td valign="top">[TXT]</td><td><a href="aixswupdate">aixswupdate</a></td><td align="right">2020-06-16 16:07  </td><td align="right"> 14K</td><td>&nbsp;</td></tr>
<tr><td valign="top">[TXT]</td><td><a href="xcatpostinit1.service.yast2">xcatpostinit1.servic..&gt;</a></td><td align="right">2020-06-16 16:07  </td><td align="right">346 </td><td>&nbsp;</td></tr>

For the long name, &gt is there to indicate truncation.

Original: curl --fail --retry 20 --max-time 60 "${url}/" | grep -o '<a href="\([^"]*\)">\1</a>' | cut -d '"' -f 2
My change: curl --fail --retry 20 --max-time 60 "${url}/" | grep -o '<a href="\([^"]*\)">.*</a>' | cut -d '"' -f 2

Replace "\1" by ".*". Basically, the second "repeated" name is ignored and the one before ">" is captured.

It turns out that there are two lines by "curl", at the beginning of the output:

   <tr><th valign="top">[ICO]</th><th><a href="?C=N;O=D">Name</a></th><th><a href="?C=M;O=A">Last modified</a></th><th><a href="?C=S;O=A">Size</a></th><th><a href="?C=D;O=A">Description</a></th></tr>  <tr><th colspan="5"><hr></th></tr>
<tr><td valign="top">[PARENTDIR]</td><td><a href="/install/">Parent Directory</a>       </td><td>&nbsp;</td><td align="right">  - </td><td>&nbsp;</td></tr>

The first record indicates how files are arranged in the directories. O=D indicates sorted in the descending order.

The use of ".*" returns "?C=N:O=D" for the first record, and "/install" (Parent Directory) for the second. But they are not relevant and should not be there. 

That is the reason "?C=N:O=D" and extra /install appear in /xcatpost of CN.

The added egrep fixes the problem and /xcatpost of CN has all the postscripts.

curl --fail --retry 20 --max-time 60 "${url}/" | grep -o '<a href="\([^"]*\)">.*</a>' | **egrep -v "O=D|Directory"** | cut -d '"' -f 2

To increase the width length of href by HTTP has a global impact to the HTTP behavior to the system.

This change in post.xcat.ng is more preferred.

PR #6781, opened 2020-07-24
Add IndexOptions NameWidth=* to xCAT/xcat.conf and xCATsn/xcat.conf.

peterwywong:
The changed files are supposedly to be put in /etc/httpd/conf.d on MN/SN.

These files are under xcat-core/xCAT and xCATsn in the xcat-core repository and may not look like they go to the above directory. However, their contents match, so I think I make the right changes.

The idea is to get the maximum width for href.

#6780 is one solution and this the other. Let's decide the better one.


besawn:
@peterwywong I think we decided to use the solution #6780, correct? Can we close this PR now without merging?

PR #6785, opened 2020-07-29
Add a separate test case to compare postscripts on MN and CN.

peterwywong:
It is added after diskfull installations in 10 bundles. Note that the Ubuntu bundles do not include diskfull installations.

Also note that diskless installations would not have this problem because all postscripts are copied and included in the install image without involvement of HTTP.

### DIFF
--- a/xCAT-test/autotest/testcase/installation/compare_postscripts
+++ b/xCAT-test/autotest/testcase/installation/compare_postscripts
@@ -3,7 +3,8 @@ os:Linux
 label:provision
 cmd:cd /install/postscripts; tar cvf /tmp/mn.tar *
 cmd:xdsh $$CN "cd /xcatpost; tar cvf /tmp/cn.tar *"
-cmd:scp $$CN:/tmp/cn.tar /tmp; rm $$CN:/tmp/cn.tar
+cmd:scp $$CN:/tmp/cn.tar /tmp
+cmd:xdsh $$CN "rm /tmp/cn.tar"
 cmd:mkdir -p /tmp/mn; tar xvf /tmp/mn.tar -C /tmp/mn
 cmd:mkdir -p /tmp/cn; tar xvf /tmp/cn.tar -C /tmp/cn; rm /tmp/cn/mypost*
 cmd:diff -r /tmp/mn /tmp/cn > /tmp/diff.list
@@ -11,6 +12,6 @@ check:rc==0
 cmd:cat /tmp/diff.list
 check:rc==0
 
-cmd:rm -fr /tmp/mn; rm -fr /tmp/cn; rm /tmp/diff.list"
+cmd:rm -fr /tmp/mn; rm -fr /tmp/cn; rm /tmp/mn.tar; rm /tmp/diff.list"
 check:rc==0
 end
--- a/xCAT-test/autotest/testcase/installation/compare_postscripts
+++ b/xCAT-test/autotest/testcase/installation/compare_postscripts
@@ -1,0 +1,17 @@
+start:compare_postscripts
+os:Linux
+label:provision
+cmd:cd /install/postscripts; tar cvf /tmp/sn.tar *
+cmd:scp /tmp/sn.tar $$CN:/tmp
+cmd:xdsh $$CN "cd /xcatpost; tar cvf /tmp/cn.tar *"
+cmd:xdsh $$CN "mkdir -p /tmp/sn; tar xvf /tmp/sn.tar -C /tmp/sn"
+cmd:xdsh $$CN "mkdir -p /tmp/cn; tar xvf /tmp/cn.tar -C /tmp/cn; rm -f /tmp/cn/mypost*"
+cmd:xdsh $$CN "diff -r /tmp/sn /tmp/cn > /tmp/diff.list"
+check:rc==0
+cmd:xdsh $$CN "cat /tmp/diff.list"
+check:rc==0
+
+cmd:rm /tmp/sn.tar
+cmd:xdsh $$CN "rm /tmp/cn.tar; rm -fr /tmp/sn; rm -fr /tmp/cn; rm /tmp/diff.list"
+check:rc==0
+end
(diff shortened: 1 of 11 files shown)


gurevichmark:
@peterwywong Ubuntu bundles do include diskfull installation:
```
[gurevich@c910loginx02 xcat-core]$ grep "reg_linux_diskfull_installation_flat" ./xCAT-test/autotest/bundle/ubuntu_*.bundle
./xCAT-test/autotest/bundle/ubuntu_ppcle_daily.bundle:reg_linux_diskfull_installation_flat
./xCAT-test/autotest/bundle/ubuntu_x86_daily.bundle:reg_linux_diskfull_installation_flat
[gurevich@c910loginx02 xcat-core]$
```

xCAT-test/autotest/testcase/installation/compare_postscripts
```diff
@@ -0,0 +1,17 @@
+start:compare_postscripts
+os:Linux
+label:provision
+cmd:cd /install/postscripts; tar cvf /tmp/sn.tar *
```
gurevichmark, line 4: should use `mn.tar` instead of `sn.tar` since tarring files on MN

xCAT-test/autotest/testcase/installation/compare_postscripts
```diff
@@ -0,0 +1,17 @@
...
+cmd:scp /tmp/sn.tar $$CN:/tmp
```
gurevichmark, line 5: Instead of copying tar from MN to the CN and running the untarring and diffing there using `xdsh`, I think it is better to copy the cn.tar to the MN and do untarring and diffing locally on the MN.

peterwywong:
"compare_postscripts" has been added to ubuntu_ppcle_daily.bundle and ubuntu_x86_daily.bundle already.

ubuntu_ppcle_weekly.bundle and ubuntu_x86_weekly.bundle do not have diskfull installations. There are 6 daily bundles and 6 weekly bundles, so we expect 12 file changes, but I only made 10. I meant to say "Note that there are **two**", but I wrote "Note that there are **the**".

mn.tar should be used. I started the work in a hierarchical cluster, so sn.tar was used initially.

What is the advantage of runing diff on MN over CN?

gurevichmark:
Ok, `the` instead of `two` is what made me think the no ubunty bundles were modified.
The advantage or running untarring and comparison on the MN vs CN is that, since testcase runs on MN, there is no need to do remote `xdsh` calls to the CN. Also MN most likely will have more disk space than CN to expand the `.tar` files.

xCAT-test/autotest/testcase/installation/compare_postscripts
```diff
@@ -1,17 +1,16 @@
...
 os:Linux
 label:provision
-cmd:cd /install/postscripts; tar cvf /tmp/sn.tar *
-cmd:scp /tmp/sn.tar $$CN:/tmp
+cmd:cd /install/postscripts; tar cvf /tmp/mn.tar *
 cmd:xdsh $$CN "cd /xcatpost; tar cvf /tmp/cn.tar *"
-cmd:xdsh $$CN "mkdir -p /tmp/sn; tar xvf /tmp/sn.tar -C /tmp/sn"
-cmd:xdsh $$CN "mkdir -p /tmp/cn; tar xvf /tmp/cn.tar -C /tmp/cn; rm -f /tmp/cn/mypost*"
-cmd:xdsh $$CN "diff -r /tmp/sn /tmp/cn > /tmp/diff.list"
+cmd:scp $$CN:/tmp/cn.tar /tmp; rm $$CN:/tmp/cn.tar
```
gurevichmark, line 6: I think instead of `rm $$CN:/tmp/cn.tar` you want `xdsh $CN rm /tmp/cn.tar`

xCAT-test/autotest/testcase/installation/compare_postscripts
```diff
@@ -1,17 +1,16 @@
...
+cmd:mkdir -p /tmp/cn; tar xvf /tmp/cn.tar -C /tmp/cn; rm /tmp/cn/mypost*
+cmd:diff -r /tmp/mn /tmp/cn > /tmp/diff.list
 check:rc==0
-cmd:xdsh $$CN "cat /tmp/diff.list"
+cmd:cat /tmp/diff.list
 check:rc==0
 
-cmd:rm /tmp/sn.tar
-cmd:xdsh $$CN "rm /tmp/cn.tar; rm -fr /tmp/sn; rm -fr /tmp/cn; rm /tmp/diff.list"
+cmd:rm -fr /tmp/mn; rm -fr /tmp/cn; rm /tmp/diff.list"
```
gurevichmark, line 14: Should `/tmp/mn.tar` also be removed ?

peterwywong:
Thanks Mark.

cxhong:
@peterwywong , thanks for create new test case for this.  
I will like to suggest to create a new file/dir `xCAT-test/autotest/testcase/postscripts/case0` , then put `compare_postscripts` testcase to file `case0`,  we can use to add more postscripts testcases in near future.

cxhong:
If I want to run this test case in any other cluster,  I have to run those two test cases first, right?
```
SN_setup_case
reg_linux_diskfull_installation_hierarchy
``` 
I think each test cases should not depends other test cases.   so, from my point view,  I think `compare_postscripts` should be independent and user can run it by itself.  like `confignetwork_s_installnic_diskful`  test case.

also, besides `tar` and `untar` the `/install/postscripts` dir,  have u able to try to create a postscripts with long name on the MN?  after provision the CN,   we can check if this postscripts downloaded to the CN `/xcatpost`.

peterwywong:
Casandra, if a CN is already a diskfull node, compare_postscripts can be run without any re-installation.

I support the idea of grouping all postscripts test cases together.

I continue working on my modified regex to retrieve postscripts with long names. I thought my solution is done, but I see something new today.

peterwywong:
[This is a test.](https://trello.com/c/MKA4QAQ6/9-this-is-a-test)

peterwywong:
[This is a test.](https://trello.com/c/MKA4QAQ6/9-this-is-a-test)

peterwywong:
[compare_postscripts - 1](https://trello.com/c/ZU7LfyXx/14-comparepostscripts-1)

peterwywong:
[compare_postscripts - 1](https://trello.com/c/ZU7LfyXx/14-comparepostscripts-1)

PR #6789, opened 2020-07-31
Add getadapter to xCAT readdoc

cxhong:
The `getadpater.rst` file is available, but didn't add to xCAT readdoc.


gurevichmark:
@cxhong Need to remove or comment out `getadapter.rst`  from the `exclude_patterns` list in `docs/source/conf.py`. Otherwise `.html` file will not be generated for it. 

Also, since we are going to be displaying this file to the user, can you fix these two typos there:

image

PR #6791, opened 2020-07-31
Remove the RC checking for the file removal command.

peterwywong:
Will investigate which rm command causes failure. It is likely that RC=1 when diff.list is empty.

Note: SUSE, Ubuntu and RHEL 7.x do not have missing postscripts with long names. Their HTTP setup could be different.


peterwywong:
[compare_postscripts - 2](https://trello.com/c/yh3BC9Ku/6-comparepostscripts-2)

PR #6793, opened 2020-08-03
feat: Secured OF prompt

### LINKED ISSUE #6637: xCAT support for Secured OF (Openfirmware) prompt
Hi

Background
I use xCAT to build and deploy VIO Servers and Novalink for a large customer using Power8 and Power9 systems which are HMC connected and Novalink managed once built.

I have come across an issue where we are unable to rinstall with netboot as when an LPAR is powered on, in firmware 940 on P9 systems it uses a "Secured OF" prompt rather than the firmware prompt which it had on firmware level 930 and below. This is preventing the use of the xCAT capabiities to netboot and install VIO Servers and RHEL operating systems on a LPAR using HMCs. 

Is there any plans to support these?. If no what would be the alternative we could use?. I need Urgent help at least to understand where we are going with xCAT with support for "Secured OF" prompt. 

Thanks in advance and appreciate any comments on this urgently.

gurevichmark:
### The PR is to fix issue _#6637_

Code contributed by @hgunasin and @wabe1968 in #6637

### DIFF
--- a/perl-xCAT/xCAT/LparNetbootExp.pm
+++ b/perl-xCAT/xCAT/LparNetbootExp.pm
@@ -4,6 +4,7 @@
 
 package xCAT::LparNetbootExp;
 
+
 #-----------------------------------------------------------------------------
 
 =head1   LparNetbootExp
@@ -33,6 +34,7 @@ Usage: Return macaddress
                         3: Set disk as 1st boot device, network as 2nd boot device
                         4: set disk as boot device
         -M      Discovery ethernet adapter mac address and location code
+        -a      Use Secure OF
         --help  Prints this help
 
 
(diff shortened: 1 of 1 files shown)


viniciusferrao:
@Obihoernchen I would vote for merging this one as-is.

I've read the code multiple times and it seems it won't cause any harm. I tried my best to work around the Secured OF thing to see if I could run it on our AC922 POWER9, but that's only available on PowerVM models with LPAR, which is a hardware thing that also can't be emulated. So... I don't have the means to test it.

I'm also not aware of anyone who has those machines and would let us borrow them to test this PR, so that's wishful thinking.

That said, after reading it multiple times, it seems like it already works and just hasn't been merged. We should merge it. It won't affect any other systems, and if something breaks because of this, we can revert later. Don't want to throw this one away.

I'm not sure if @gurevichmark could just help us here with some memory of why it didn't go through... any information would be very helpful.

viniciusferrao (approved):

viniciusferrao:
@gurevichmark can you sign the CLA so we can merge this PR?

Thanks!

viniciusferrao:
Retrigger CI/CD.

viniciusferrao:
Thanks @gurevichmark

Issue #6832, opened 2020-09-21
Syncfiles not working during diskfull install


nealep:
Howdy! We're having some trouble with file synchronization during a diskfull installation of CentOS 7.8. The server is running XCat version 2.16. When we run `rinstall node` the installation process runs to completion but after the reboot, the files aren't synced and we get the following in the /var/log/xcat/xcat.log file:

```
RCP=
'[' '!' -e /usr/bin/rsync ']'
logger -t xcat.deployment.postscript -p local4.info 'Performing syncfiles postscript'
uname
osname=Linux
xcatpostdir=/xcatpost
logger -t xcat.deployment.postscript -p local4.info 'syncfiles: the OS name = Linux'
quit=no
count=5
returncode=0
'[' no = no ']'
cat /etc/os-release
greip -i cumulus
'[' Linux = Linux ']'
/xcatpost/startsyncfiles.awk -v RCP=
returncode=1
```

Running `updatenode node -F` after the installation does distribute the files correctly though. 

The `syncfiles` file contains: 

```
/etc/hosts -> /etc/hosts
/etc/passwd -> /etc/passwd
```
I thought it might be related to https://github.com/xcat2/xcat-core/issues/5252 but `/var/xcat/syncfiles` and `/var/xcat/nodes/syncfiles` both exist on the master node and this setup is not hierarchical. Furthermore, the `synclists` value for the osimage is set to a correctly formatted file that exists. 

Any thoughts would be appreciated!

nealep:
Looks like the root cause of the issue was that SSH was not starting due to permissions on private keys. That prevented rsync from doing its thing for whatever reason. Fixing those permissions appears to have fixed the problem.

besawn:
@nealep Different versions of SSH appear to require different permissions for the private keys. We originally had to change from `0600` to `0640` due to this earlier issue: https://github.com/xcat2/xcat-core/issues/2617 

Which version of SSH did you have installed when you encountered this issue?

PR #6833, opened 2020-09-21
Change private key permissions to 600

nealep:
### The PR is to fix issue https://github.com/xcat2/xcat-core/issues/6832

Some versions of OpenSSH will not load private keys with permissive permissions. Setting permission to 600 allows OpenSSH to start and load the created private keys and avoid some downstream errors during postbootscript sequences.

### The modification include
* postscripts/remoteshell - changes the permissions of private keys from 640 to 600

### DIFF
--- a/xCAT/postscripts/remoteshell
+++ b/xCAT/postscripts/remoteshell
@@ -152,7 +152,7 @@ if [ $? -ne 0 ]; then
                 getcredentials.awk ssh_dsa_hostkey | grep -v '<'|sed -e 's/&lt;/</' -e 's/&gt;/>/' -e 's/&amp;/&/' -e 's/&quot/"/' -e "s/&apos;/'/" > /etc/ssh/ssh_host_dsa_key
                 MYCONT=`cat /etc/ssh/ssh_host_dsa_key`
    done
-   egrep  -i "^ssh_keys:" /etc/group >/dev/null 2>&1 && chown root:ssh_keys /etc/ssh/ssh_host_dsa_key 2>/dev/null && chmod 640 /etc/ssh/ssh_host_dsa_key
+   egrep  -i "^ssh_keys:" /etc/group >/dev/null 2>&1 && chown root:ssh_keys /etc/ssh/ssh_host_dsa_key 2>/dev/null && chmod 600 /etc/ssh/ssh_host_dsa_key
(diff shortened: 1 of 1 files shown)


besawn:
@nealep Thank you again for the pull request. The pull requests you have opened recently have been very helpful; I think it would greatly benefit the xCAT project if you could formally join the project as a contributor. Are you interested in joining the project as a contributor by submitting a [Contributor License Agreement](https://xcat-docs.readthedocs.io/en/latest/developers/license/contributors.html)?

nealep:
Sure, that'd be great! I'll have to get the CCLA put through legal, though, which will take some time.

besawn:
@nealep I understand that obtaining legal approvals can be time consuming, so for these existing pull requests, the xCAT core team may create equivalent pull requests to address these issues so the changes can be included in the next release. In the long term, we do appreciate your support of the project in whatever manner is most efficient for you. If you are able and willing to pursue getting approval to submit a CLA and CCLA, we would be happy to have you become a formal xCAT contributor. If that process becomes too much of a burden, we still welcome your participation, even if we are not able to directly accept pull requests. Thanks again!

nealep:
I'm going to go ahead and submit the CCLA to the legal folks. No idea what the turn around time might be so in the meantime it's probably worth the xCAT maintainers making a duplicate PR for this and https://github.com/xcat2/goconserver/pull/64 then pulling in the changes that way. I'm sure you'll hear from me when the form gets signed!

cxhong:
I couldn't recreate issue #6832  with centos7.8 images and rhels8.2 images.
But according the man page:
```
/etc/ssh/ssh_host_ecdsa_key
     /etc/ssh/ssh_host_ed25519_key
     /etc/ssh/ssh_host_rsa_key
             These files contain the private parts of the host keys.  These files should only be
             owned by root, readable only by root, and not accessible to others.  Note that sshd
             does not start if these files are group/world-accessible.
````
we should change the permission to 600. only allow root to read/write.   
I verified this PR on centos7.8/x86_64  and rhels8.2/ppc64le,  didn't hit any issues.  I think should be ready to merge

viniciusferrao:
We should close this one @Obihoernchen

I've read this PR and #6842 multiple times, and this does not apply. @cxhong also found this isn't an issue at all.

We need SSH Hostbased Authentication working, and this change effectively breaks that. It was long discussed in #2617.

So this is a close for me, and also on #6842.

PR #6834, opened 2020-09-22
Add REST test cases, a new bundle to the ppc64le daily bundle

peterwywong:
This PR contains 21 REST test cases, a new REST bundle and its inclusion to the RHEL ppcle daily bundle.

Note that this RHEL ppcle daily bundle has been included by its corresponding weekly bundle. Only RHEL 8.1 has been tested so far (on a private 2-node cluster and a private 3-node cluster), so this set of changes will NOT be tested on SLES and Ubuntu in our daily regression runs at this time. Will decide to expand the OS coverage after the initial regression results are examined.

The test cases were demonstrated to the team in a working session many weeks ago. A few more tests and curl features have been added based on suggestions. From that point on, a lot of work has been put into to run tests successfully in a hierarchical environment.

Highlights of the changes:
(1) mod_ssl is required to REST test cases.
(2) Minimal changes are made to set up the REST run environment:
    - Installation of mod_ssl
    - Minimal changes to /etc/httpd/conf.d/ssl.conf (this file is provided by the mod_ssl RPM). In the documentation, the changes are in the optional section, but they are not optional.
    - Copy /install/postscripts/ca/ca-cert.pm from MN to /root of CN, so that the CN can make REST request with a legitimate certificate.
    - "xcatconfig -c" and "updatenode -k" are NOT used at all. The existing /etc/xcat/cert/server-pred.pem, /etc/xcat/cert/ ca.pem and /install/postscripts/ca/ca-cert.pm are used.
    - Therefore, the cleanup is much easier.
(3) A REST bundle is created to run 24 tests.
(4) All curl commands have "-s" to suppress unnecessary output.
(5) A few tests have "-k" as suggested by Mark. The use "-k" is to ignore the legitimacy of the server. So even though the server certificate is not valid, the request is still being processed.
(6) The curl features: debug=1, pretty=1 and xcoll=1 are tested.
(7) Many error cases are tested too, such as asking for a wrong OS image.
(8) node range is tested.


gurevichmark:
@peterwywong 

* Why does the bundle file name needs to identify OS and architecture (`rhels_ppcle_restapi.bundle `). Why is it not just `restapi.bundle` ?
* Above you said _(8) node range is tested._ Which testcase is that ?
* The `restapi_setup_on_MN_SN_CN` testcase hardcodes the userid and password. Better to use `__GETTABLEVALUE(...)`, just like the rest of the testcases.
* Maybe add a few more testcases to see what happens when 
   * non-existent path is called, something like `https://$$MN/xcatws/abc?userName=$username&userPW=$password`
   * invalid option is passes, like `https://$$MN/xcatws/nodes?userName=$username&userPW=$password&xyz=123`

peterwywong:
Why does the bundle file name needs to identify OS and architecture (rhels_ppcle_restapi.bundle ). Why is it not just restapi.bundle ?
I might change it to rhels_restapi.bundle since x86 might run it fine too. 
The mod_ssl installation is the first test case of the REST bundle. Its installation methods are different on SLES and Ubuntu. For example, Ubuntu uses apt-get.

Above you said (8) node range is tested. Which testcase is that ?
restapi_delete_temp_CN_2 deletes 2 nodes with temp_node[1-2]. My bundle intentionally creates two nodes with PUT and POST and then deletes them.

The restapi_setup_on_MN_SN_CN testcase hardcodes the userid and password. Better to use __GETTABLEVALUE(...), just like the rest of the testcases.
Will change them.

Will add two more tested as suggested.

viniciusferrao:
May I suggest a thing? ppcle is not an architecture, it's ppc64le.

peterwywong:
ppc64le is indeed the architecture name. Currently, the file name for the ppc64le daily regression is rhels_ppcle_daily.bundle. The new bundle rhels_ppcle_restapi.bundle is named accordingly.

However, it can be named restapi.bundle as suggested by Mark. 

restapi_setup_on_MN_SN_CN and restapi_cleanup_on_MN_SN_CN will be rewritten to set up and clean up differently based on the distros and architectures.

In fact, SN will be removed too because SN need not be modified based on the improved minimal changes.

gurevichmark:
> 
> 
> Why does the bundle file name needs to identify OS and architecture (rhels_ppcle_restapi.bundle ). Why is it not just restapi.bundle ?
> I might change it to rhels_restapi.bundle since x86 might run it fine too.
> The mod_ssl installation is the first test case of the REST bundle. Its installation methods are different on SLES and Ubuntu. For example, Ubuntu uses apt-get.

I think it is best to just name it `restapi.bindle` and for one line in setup testcase that would be different because of the installation method, to check what OS testcase is running on. Similar to the way installation is done in `go_xcat_devel_from_repo` testcase
```
#Install additional packages on Red Hat
cmd:if "grep \"Red Hat\" /etc/*release >/dev/null"; then "yum install -y yum-utils dnf-utils bzip2"; fi

#Install additional packages on Ubuntu
cmd:if  "grep \"Ubuntu\" /etc/*release >/dev/null"; then "DEBIAN_FRONTEND=noninteractive apt-get install -y gnupg"; fi

```
> 
> Above you said (8) node range is tested. Which testcase is that ?
> restapi_delete_temp_CN_2 deletes 2 nodes with temp_node[1-2]. My bundle intentionally creates two nodes with PUT and POST and then deletes them.
> 

Ok, I see. Then maybe add a few testcases to test the `/nodes/{noderange}/nodels` to get back a list of nodes matching a noderange.  Try it when a noderange is a group name, like `/nodes/all/nodels`, incremental range, like `/nodes/node1+2/nodels`
and an attribute match, like `/nodes/nodehm.mgt==kvm/nodels`.

> The restapi_setup_on_MN_SN_CN testcase hardcodes the userid and password. Better to use __GETTABLEVALUE(...), just like the rest of the testcases.
> Will change them.
> 
> Will add two more tests as suggested.

peterwywong:
The test cases are enhanced and added based on Mark's suggestions. And he has done a code review.

Since restapi_setup_on_MN_CN and restapi_cleanup_on_MN_CN have conditional statements to take care of various OSes, rhels_ppcle_restapi.bundle is now replaced by restapi.bundle.

In restapi_setup_on_MN_CN, "service httpd restart" was done with the mod_ssl installation, with ONE OS check.

However, "service httpd restart" should be done AFTER the ssl.conf modification. So, "service httpd restart" is now done separately.

xCAT-test/autotest/bundle/rhels_ppcle_daily.bundle
```diff
@@ -307,3 +307,4 @@ xcatstanzafile_objtype
 xcatstanzafile_specificvalue
 xcatstanzafile_tab
 go_xcat_devel_from_repo
+#INCLUDE:restapi.bundle#
```
gurevichmark, line 310: Might need to move the running of `restapi.bundle` before `go_xcat_devel_from_repo` testcase. That testcase installs xCAT on the CN and makes it a MN. So it might not behave as a CN and respond to `xdsh` command correctly.

xCAT-test/autotest/testcase/restapi/cases0
```diff
@@ -0,0 +1,227 @@
...
+check:rc==0
+check:output=~__GETNODEATTR($$CN,os)__-__GETNODEATTR($$CN,arch)__
+end
+
+start:restapi_list_networks
+description: List networks on the management node with "curl -X GET"
+label:restapi
+cmd:username=__GETTABLEVALUE(key,system,username,passwd)__;password=__GETTABLEVALUE(key,system,password,passwd)__;xdsh $$CN "curl -X GET -s -k --cacert /root/ca-cert.pem 'https://$$MN/xcatws/networks?userName=$username&userPW=$password'"
+check:rc==0
+check:output!~""
```
gurevichmark, line 74: I do not think checking for non-empty output is enough here. Perhaps there is a way to verify that some networks type of info is returned using `__GETTABLEVALUE__`?

xCAT-test/autotest/testcase/restapi/cases0
```diff
@@ -0,0 +1,227 @@
...
+description: Delete a temporary compute node with "curl -X DELETE"
+label:restapi
+cmd:username=__GETTABLEVALUE(key,system,username,passwd)__;password=__GETTABLEVALUE(key,system,password,passwd)__;xdsh $$CN "curl -X DELETE -s --cacert /root/ca-cert.pem 'https://$$MN/xcatws/nodes/temp_node[1-2]?userName=$username&userPW=$password'"
+check:rc==0
+cmd:lsdef
+check:output!~temp_node1
+end
+
+start:restapi_list_temp_CN_2
+description: Delete a temporary compute node with "curl -X DELETE"
```
gurevichmark, line 120: Is this a description typo ? This is not a "delete" testcase.

xCAT-test/autotest/testcase/restapi/cases0
```diff
@@ -0,0 +1,227 @@
...
+check:rc==0
+check:output=~__GETTABLEVALUE(node,$$CN,groups,nodelist)__
+end
+
+start:restapi_list_nodes
+description: List nodes on the management node with "curl -X GET"
+label:restapi
+cmd:username=__GETTABLEVALUE(key,system,username,passwd)__;password=__GETTABLEVALUE(key,system,password,passwd)__;xdsh $$CN "curl -X GET -s --cacert /root/ca-cert.pem 'https://$$MN/xcatws/nodes?&userName=$username&userPW=$password'"
+check:rc==0
+check:output=~$$CN
```
gurevichmark, line 58: I do not think that checking for the output to contain $$CN is enough. Since command is being executed using `xdsh $$CN` any output from CN will contain $$CN as prefix.

peterwywong:
Can you elaborate more on restapi_list_temp_CN_2? This is a DELETE test case to delete 2 temp nodes. The two nodes are created by PUT and POST early on.

I should modify the description from "Delete a temporary compute node" to "Delete two temporary compute nodes."

peterwywong:
Mark, I addressed your comments above and submitted new changes. I appreciate your very detailed code review.

gurevichmark (approved):

besawn (approved):
Looks good!

PR #6840, opened 2020-09-29
Remove experimental warnings of using given-when and smartmatch.

peterwywong:
For given-when, lsvm was used to verify the correctness. The changed file is kvm.pm.

For smartmatch, there are two files: openbmc.pm and OPENBMC.pm. 

For the change in openbmc.pm, "reventlog mid05tor12c15 resolved=" and "reventlog mid05tor12c15 resolved=48" were tested on briggs01 for correctness.

For the change in OPENBMC.pm, "rflash f5u14 -l -V" was used on stratton01 for correctness.


peterwywong:
The meaning of "my @temp = grep ({$command =~ $_ } @ARGV);":

Grep the "exact" $command ($_) from the array of ARGV. The result is a list of zero or more $command.

temp[0] gets the first one item for comparison purposes. temp[0] can be empty.

gurevichmark (approved):

PR #6842, opened 2020-09-30
Change Host Private Key Permissions to 600

cxhong:
This PR is duplicated with #6833

@nealep is processing the CCLA.  In the meantime, I am making a duplicate PR to check in his fixes.

### DIFF
--- a/xCAT/postscripts/remoteshell
+++ b/xCAT/postscripts/remoteshell
@@ -152,7 +152,7 @@ if [ $? -ne 0 ]; then
                 getcredentials.awk ssh_dsa_hostkey | grep -v '<'|sed -e 's/&lt;/</' -e 's/&gt;/>/' -e 's/&amp;/&/' -e 's/&quot/"/' -e "s/&apos;/'/" > /etc/ssh/ssh_host_dsa_key
                 MYCONT=`cat /etc/ssh/ssh_host_dsa_key`
    done
-   egrep  -i "^ssh_keys:" /etc/group >/dev/null 2>&1 && chown root:ssh_keys /etc/ssh/ssh_host_dsa_key 2>/dev/null && chmod 640 /etc/ssh/ssh_host_dsa_key
+   egrep  -i "^ssh_keys:" /etc/group >/dev/null 2>&1 && chown root:ssh_keys /etc/ssh/ssh_host_dsa_key 2>/dev/null && chmod 600 /etc/ssh/ssh_host_dsa_key
    if ! grep "PRIVATE KEY" /etc/ssh/ssh_host_dsa_key > /dev/null 2>&1 ; then
         rm /etc/ssh/ssh_host_dsa_key
 	else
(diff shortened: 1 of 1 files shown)


besawn:
Moving this PR to the next release until we can better understand which versions of OpenSSH require private keys with 0640 vs 0600 permissions.

nealep:
From `man.openbsd.org/sshd` for OpenSSH Version 8.4: 

/etc/ssh/ssh_host_ecdsa_key
/etc/ssh/ssh_host_ed25519_key
/etc/ssh/ssh_host_rsa_key
These files contain the private parts of the host keys. These files should only be owned by root, readable only by root, and not accessible to others. Note that sshd does not start if these files are group/world-accessible.

Looks like similar language is in the manual going back to version BSD version 3.0 which I think corresponds to OpenSSH Version ~2. I haven't been able to find when exactly they introduced the error behavior, however. It looks like group permissions were just strongly encouraged up until relatively recently or it's a whacky RedHat patch.

nealep:
Some more interesting information: 

https://bugzilla.redhat.com/show_bug.cgi?id=1801459

Sounds like 0640 _will_ be acceptable if the group is ssh-keys. I don't have a way to verify this behavior at present. Does someone have a RHEL 7 image they can experiment with? Otherwise, I can try to do it in the next week or two?

besawn:
@nealep The current behavior was due to this issue: https://github.com/xcat2/xcat-core/issues/2617
which references this issue:
https://bugzilla.redhat.com/show_bug.cgi?id=819896

Given that different versions of OpenSSH require different ownership and permissions, a more reliable approach might be to generate a temporary host private key on the compute node to determine the group and permissions that are expected by the installed version of OpenSSH, then copy the correct keys over and set the ownership and permissions to match those of the temporary key. This is the approach I would like to explore next, when time permits.

peterwywong:
An xCAT management node has two sets of host keys, one under /etc/ssh and the other /etc/xcat/hostkeys. The SECOND set of host keys is distributed to /etc/ssh of each compute node by /install/postscripts/remoteshell.

In March 2017, **PR 2724 fix issue Wrong permissions for ssh host private keys on CentOS 7 #2617** modified the permission of the host keys from 600 to 640 and replaced their group ownership from root to ssh_keys on compute nodes where Group ssh_keys is defined. Only CentOS and RHEL have ssh_keys defined.

Since the releases CentOS 7.0 and RHEL 7.0 in 2014, the host keys have been set to have Permission 640 and Group ssh_keys by **sshd-keygen**.

sshd-keygen is a script only available on CentOS and RHEL and packaged in openssh-server. It calls ssh-keygen to create keys and then sets the above permission and group ownership.

Here is /usr/libexec/openssh/sshd-keygen of RHEL 8.2, with openssh-server-8.0p1-4.el8_1.ppc64le.

#!/bin/bash

# Create the host keys for the OpenSSH server.
KEYTYPE=$1
case $KEYTYPE in
        "dsa") ;& # disabled in FIPS
        "ed25519")
                FIPS=/proc/sys/crypto/fips_enabled
                if [[ -r "$FIPS" && $(cat $FIPS) == "1" ]]; then
                        exit 0
                fi ;;
        "rsa") ;; # always ok
        "ecdsa") ;;
        *) # wrong argument
                exit 12 ;;
esac
KEY=/etc/ssh/ssh_host_${KEYTYPE}_key

**KEYGEN=/usr/bin/ssh-keygen**
if [[ ! -x $KEYGEN ]]; then
        exit 13
fi

# remove old keys
rm -f $KEY{,.pub}

# create new keys
if ! $KEYGEN -q -t $KEYTYPE -f $KEY -C '' -N '' >&/dev/null; then
        exit 1
fi

# sanitize permissions
**/usr/bin/chgrp ssh_keys $KEY
/usr/bin/chmod 640 $KEY**
/usr/bin/chmod 644 $KEY.pub
if [[ -x /usr/sbin/restorecon ]]; then
        /usr/sbin/restorecon $KEY{,.pub}
fi

exit 0

The information given by nealep on sshd for OpenSSH Version 8.4 is likely for ssh-keygen. sshd-keygen of CentOS and RHEL makes use of ssh-keygen with added requirements as described above.

We know xCAT distributes a common set of host keys to all compute nodes; CentOS and RHEL require Permission 640 and Group ssh_keys for them; the fixes in PR 2724 was to comply with those requirements. We still need this compliance today.

viniciusferrao:
@Obihoernchen This should be closed, just like #6833.

PR #6861, opened 2020-10-21
Only install bridge-utils on RH7.x

gurevichmark:
Only install `brigde-utils` for testcases running on RH7.x


xCAT-test/autotest/testcase/confignetwork/cases0
```diff
@@ -607,7 +607,7 @@ cmd:xdsh $$CN "rm -rf /tmp/backupnet/"
 cmd:xdsh $$CN "mkdir -p /tmp/backupnet/"
 check:rc==0
 cmd:if grep SUSE /etc/*release;then xdsh $$CN "cp -f /etc/sysconfig/network/ifcfg-* /tmp/backupnet/"; elif grep -E "Red Hat|CentOS" /etc/*release;then xdsh $$CN "cp -rf /etc/sysconfig/network-scripts /tmp/backupnet/"; elif grep Ubuntu /etc/*release;then xdsh $$CN "cp -f /etc/network/interfaces.d/* /tmp/backupnet/";else echo "Sorry,this is not supported os"; fi
-cmd:if grep -E "Red Hat|CentOS" /etc/*release;then xdsh $$CN "yum -y install bridge-utils";fi
+cmd:if grep -E "Red Hat|CentOS" /etc/*release | grep "release 7";then xdsh $$CN "yum -y install bridge-utils";fi
```
besawn, line 610: I have a suggestion, how about if instead of tying this logic to a specific release, we tie it to whether the package exists or not. Something like this:
```
cmd:xdsh $$CN "if yum -q list bridge-utils ; then yum -y install bridge-utils; fi"
```
So in this case it will install on RHEL 7 and skip the install on RHEL 8, but then it has the advantage of also installing on any other OS that has a package named `bridge-utils` available.

gurevichmark: Yes, that is even better. New commit added.

besawn (approved):

PR #6865, opened 2020-10-28
Adding 2.16.1 release information

gurevichmark:
xCAT 2.16.1 release


besawn (changes requested):
@gurevichmark Can you change the order of the entries in 2.16.x.csv so that 2.16.1 is above 2.16.0? I would like to keep it consistent with how we did 2.15.x.csv, which added to the top of the table rather than the bottom for new releases.

gurevichmark:
Good catch. Order changed.

besawn (approved):
Looks good, thanks.

PR #6876, opened 2020-11-12
[doc only] Add vlan to networks documentation

krono:
Apparently the doc was contributed some time ago and is live (https://xcat-docs.readthedocs.io/en/stable/advanced/networks/vlan/index.html) but not referenced in the toc for networking (https://xcat-docs.readthedocs.io/en/stable/advanced/networks/index.html)


krono:
[Ok, this is strange: the build obviously passed: https://travis-ci.org/github/xcat2/xcat-core/builds/743197541 but was not reported here :/ ]

PR #6924, opened 2021-03-16
Add support for SLES15 SP2

gurevichmark:
### The PR is to fix issue _#6856_

To support SLES15 SP2, the following changes were needed.

* Skip over directories `/install/sle15.2/ppc64le/1/Product-SLES_SAP` and `/install/sle15.2/ppc64le/1/Product-SUSE-Manager-Server-4.1` when building a list of products to include when generating AutoYaST file.
Since AutoYaST specifies `<product>SLES</product>`, all other "products" included in the `<add_on_products>` section cause conflict when installing SLES15 SP2.
* Add `wget` to the pkglist file. Needed to get postscripts file to the provisioned node.
* Add extra subdirectory repositories so `genimage` can find packages when building diskless image.
* Remove soft link of `resolve.conf` file when building diskless image. The destination of the link is a local file, which does not exist when booting the node and therefore fails when writing contents to it.

### DIFF
--- a/xCAT-server/share/xcat/netboot/sles/genimage
+++ b/xCAT-server/share/xcat/netboot/sles/genimage
@@ -2006,7 +2006,13 @@ sub generic_post { # This function is meant to leave the image in a state approx
     print $cfgfile "NETWORKING=yes\n";
     close($cfgfile);
 
-    # SLE15 has a symlink to /run/netconfig/resolv.conf and does not need a dummy file
+    # SLE15.2 has a symlink to /run/netconfig/resolv.conf but still need a dummy file
+    if (-l "$rootimg_dir/etc/resolv.conf" && $osver =~ "15.2") {
+        # Remove resolv.conf, so that link is gone and code below will
+        # create a dummy file
+        unlink("$rootimg_dir/etc/resolv.conf");
+    }
+    # SLE15.0 has a symlink to /run/netconfig/resolv.conf and does not need a dummy file
     if (! -l "$rootimg_dir/etc/resolv.conf") {
         open($cfgfile, ">", "$rootimg_dir/etc/resolv.conf");
         print $cfgfile "#Dummy resolv.conf to make boot cleaner";
--- a/xCAT-server/share/xcat/netboot/sles/genimage
+++ b/xCAT-server/share/xcat/netboot/sles/genimage
@@ -299,12 +299,20 @@ unless ($onlyinitrd) {
             if (-d "$dir/1") {
                 $ddir .= "/1";
             }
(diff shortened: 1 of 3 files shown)


gurevichmark:
### UT

|OS|Arch|Provision|xCAT installs|
|---|-|-----|---|
|SLES15.2|Power|Yes|Yes|
|SLES15.2|x86| | |

xCAT-server/share/xcat/netboot/sles/genimage
```diff
@@ -299,12 +299,20 @@ unless ($onlyinitrd) {
             if (-d "$dir/1") {
                 $ddir .= "/1";
             }
-            system("zypper -R $rootimg_dir $non_interactive ar file:$ddir $osver-$i");
-            $i++;
-            if (-d "$dir/2") {
-                $ddir = $dir . "/2";
+            if ($osver =~ "^sle15.2") {
```
besawn, line 302: Can this be restructured to not depend explicitly on the OS version being sle15.2? Most likely sle15.3, sle15.4, etc will need the same behavior, so it would be better if this could check for the condition that needs to be adjusted for rather than the very specific OS version.

xCAT-server/share/xcat/netboot/sles/genimage
```diff
@@ -1998,7 +2006,13 @@ sub generic_post { # This function is meant to leave the image in a state approx
     print $cfgfile "NETWORKING=yes\n";
     close($cfgfile);
 
-    # SLE15 has a symlink to /run/netconfig/resolv.conf and does not need a dummy file
+    # SLE15.2 has a symlink to /run/netconfig/resolv.conf but still need a dummy file
```
besawn, line 2009: Similar comment here, can this be checked in a more generic way than 15.2?

gurevichmark: New commit added to eliminate checking for specific OS version.

besawn (approved):

xCAT-server/share/xcat/netboot/sles/genimage
```diff
@@ -299,12 +299,20 @@ unless ($onlyinitrd) {
             if (-d "$dir/1") {
                 $ddir .= "/1";
             }
-            system("zypper -R $rootimg_dir $non_interactive ar file:$ddir $osver-$i");
-            $i++;
-            if (-d "$dir/2") {
-                $ddir = $dir . "/2";
+            if (-d "$ddir/Product-SLES" && -d "$ddir/Module-Basesystem" && -d "$ddir/Module-Legacy") {
+                # If "Modile" and "Product" directories are there, use them for package repositories
```
besawn, line 303: Small typo here: "Modile" should be "Module"

PR #6930, opened 2021-03-25
xcatprobe fixes

gurevichmark:
### The PR is to fix issue _#6929_

Verified on RH, SLES and Ubuntu

### DIFF
--- a/xCAT-probe/subcmds/xcatmn
+++ b/xCAT-probe/subcmds/xcatmn
@@ -48,7 +48,7 @@ Description:
 Options:
     -h : Get usage information of $program_name
     -V : Output more information for debug
-    -i : Specify the network interface name of provision network on management node. if not specified, will guess the provision network from site table master attribute. Recommand to use -i option. If there is vlan in the network interface, provide the network interface with vlan infomation, such as '-i <nic>.<vlan>'.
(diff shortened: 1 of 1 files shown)


xCAT-probe/subcmds/xcatmn
```diff
@@ -173,7 +174,8 @@ sub do_main_job {
         $rst        = check_dhcp_leases(\$checkpoint, \@error);
         print_check_result($checkpoint, "w", $rst, \@error);
         $rc |= $rst;
-    } else {
+
```
besawn, line 177: This logic needs to be modified in order to run on both MN and SN.

xCAT-probe/subcmds/xcatmn
```diff
@@ -1005,13 +1007,30 @@ sub check_dns_service {
                     }
                 } else {
 
-                    # if there is no sn, nslookup mnip
+                    # if there is no sn, "nslookup mnip mnip"
                     my $nslkp = `nslookup -type=A $serverip $serverip 2>&1`;
                     chomp($nslkp);
                     my $tmp = grep { $_ =~ "Server:[\t\s]*$serverip" } split(/\n/, $nslkp);
                     if (!$tmp) {
                         $rc = 1;
```
besawn, line 1015: This section should return from here rather than setting an rc and then returning at the bottom of the function.

besawn (changes requested):
The team conducted a group review of this PR and identified two changes that are needed.

gurevichmark:
Another commit added to fix issues identified during the review and grammar and spelling fixes.

besawn (approved):

PR #6948, opened 2021-04-21
Support Oracle Linux on service nodes

gurevichmark:
### The PR extends _#6919_

* Links relates service node files to support Oracle Liniux on service nodes
* Uses `anaconda.pm` to dynamically detect Oracle Linix instead of hardcoded values in `discinfo.pm`

#### UT:
Oracle Linux 7.9 - diskful compute and service
Oracle Linux 8.3 - diskful compute and service
Oracle Linux 7.9 - diskless compute
Oracle Linux 8.3 - diskless compute


xCAT-server/lib/xcat/plugins/anaconda.pm
```diff
@@ -2109,6 +2109,24 @@ sub copycd
...
             close($dinfo);
         }
+        elsif ($desc =~ /Oracle Linux/)
+        {
+            #
+            # Attempt to auto-detect for OL8 OS, the last element has typically been the version
+            # ex: "Oracle Linux 8.3.0"
+            #
+            my @ol_version = split / /, $desc;
+            $distname = "ol" . $ol_version[-1];
```
besawn, line 2119: I learned something new here, clever use of `-1` index!

xCAT-server/lib/xcat/plugins/anaconda.pm
```diff
@@ -2109,6 +2109,24 @@ sub copycd
...
+        }
+        elsif ($desc =~ /OL-/)
+        {
+            #
+            # Attempt to auto-detect for OL7 OS, the last element has typically been the version
```
besawn, line 2124: I think this comment should be:
```
# Attempt to auto-detect for OL7 OS, the first element has typically been the version
```

besawn (changes requested):
Maybe fix the comment for future clarity?

besawn (approved):

perl-xCAT/xCAT/data/discinfo.pm
```diff
@@ -136,13 +136,6 @@ our %distnames = (
     "1390839789.062069" => "SL6.5",         #x86_64 DVD ISO Install
 
     "1529960248.065311" => "ol6.10",        # x86_64, Oracle Linux 6.10
```
besawn, line 138: @viniciusferrao FYI: For Oracle Linux, @gurevichmark made an update in this PR to try to automatically detect the OS version from the ISO description in `anaconda.pm` rather than requiring every individual ISO release being added individually. This is the approach we have been trying to take as we add new OS support to hopefully allow ol8.4, ol8.5, ol8.6 to work automatically without requiring future intervention.

Just wanted to make you aware of this change and the rationale for it.

viniciusferrao:
That's awesome @besawn. I was trying to understand how xCAT detects the OS. I've observed that on CentOS for example there's no need to put all the existing ISOs, but I wasn't able to figure it by myself, my programming background is C and I struggled a lot trying to understand the Perl code and creating the PR without breaking other things.

But at the end of the day it worked. I'll dig on this PR to better understand how this detection works. This is really good and thanks for pointing out.

There's other feature that I plan to do, for example, create two templates for the UEK kernel and the Red Hat's compatible one, but this is on my list of tasks to do. For now no one asked for it but I'm anticipating.

Thanks again @besawn and @gurevichmark for accepting the PR in first place and enhancing Oracle Linux support :)

PR #6951, opened 2021-04-21
Remove pidfile /dev/null for chronyd in setupntp on Ubuntu systems

peterwywong:
The test case nodeset_cmdline fails on Ubuntu 18.04.2. This test case wants to load the genesis kernel on a compute node.

It turns out that the TFTP server does not run at this point.

Further investigation indicates that "chronyd ... "pidfile /dev/null"" turns the character-oriented device /dev/null into a text file /dev/null. By having that, the TFTP server stops working. Since the genesis kernel cannot be passed to the compute node through the TFTP server, the compute node loads whatever on its own hard drive.

This change removes "pidfile /dev/null" only if the management node runs a Ubuntu distro.

### DIFF
--- a/xCAT/postscripts/setupntp
+++ b/xCAT/postscripts/setupntp
@@ -168,15 +168,13 @@ warn_if_bad "$?" "Failed to configure the system to maintain the RTC in universa
 # Synchronize and set the system clock once
 logger -t $log_label -p local4.info "Syncing the clock ..."
 
-OS_name=`cat /etc/os-release | grep "NAME=\"Ubuntu\""`
-
-if [ -z $OS_name ]
+if [ -f /etc/os-release ] && cat /etc/os-release |grep NAME|grep Ubuntu>/dev/null
 then
-       pidfile_option="pidfile /dev/null"
-else
        # Some versions of chronyd on Ubuntu distros have an issue
        # with the valid option "pidfile /dev/null".
        pidfile_option=""
+else
+       pidfile_option="pidfile /dev/null"
 fi
 
 chronyd -f /dev/null -q "$(
--- a/xCAT/postscripts/setupntp
+++ b/xCAT/postscripts/setupntp
@@ -145,7 +145,7 @@ fi
 unset MASTER
 unset NTPSERVERS
 
-check_exec_or_exit cp cat logger
+check_exec_or_exit cp cat logger grep
 check_exec_or_exit systemctl timedatectl hwclock
 
 systemctl stop ntp.service 2>/dev/null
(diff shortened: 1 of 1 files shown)


xCAT/postscripts/setupntp
```diff
@@ -167,14 +167,26 @@ warn_if_bad "$?" "Failed to configure the system to maintain the RTC in universa
 
 # Synchronize and set the system clock once
 logger -t $log_label -p local4.info "Syncing the clock ..."
+
+OS_name=`cat /etc/os-release | grep "NAME=\"Ubuntu\""`
```
gurevichmark, line 171: I think all postscript already get `OSVER` environment variable passed into them.
Perhaps that could be used instead of figuring it out again ?

peterwywong: Great point.

In /install/postscripts/hardeths, OSVER is set in the script:
```
if [ -f /etc/os-release ] && (cat /etc/os-release |grep -i  '^NAME=[ "]*Cumulus Linux[ "]*$' >/dev/null 2>&1); then
    OSVER="cumulus"
fi

if [ -f /etc/os-release ] && (cat /etc/os-release |grep NAME|grep -i SLE >/dev/null); then
   OSVER="SLE"
fi
```
In /install/postscripts/syslog, it is assumed that $OSVER is set:
```
      if ( pmatch $OSVER "ubuntu*" ) || ( is_lsb_ubuntu ) || ( pmatch $OSVER "debian*" ); then
```
In many scripts, we have $OSVER and $ARCH, but **export** does not include them. Where are they defined?

As an experiment, **echo $OSVER** gives out nothing in setupntp.

besawn (changes requested):
Please update line 148 from:
```
check_exec_or_exit cp cat logger
```
to
```
check_exec_or_exit cp cat logger grep
```
to account for the new dependency on grep.

besawn (approved):

Obihoernchen:
This is an issue on Rocky/Alma 8.5, too.

peterwywong:
Is it a permission denied problem on /dev/null?

Do you see this problem on Rocky 8.4?

Rocky 8.4 runs fine in the daily regression run going through "setupntp". We can add Rocky to the condition if needed.

Obihoernchen:
No chronyd or systemd will somehow remove the character special device and replace it with an empty file.
You can easily check it with file.

```
❯ file /dev/null
/dev/null: character special (1/3)
❯ # setupntp...
❯ file /dev/null
/dev/null: empty
```

Removing the pidfile option fixes this issue.

URZ-HD:
We have seen this issue also constantly on RHEL8.5. Removing the "pidfile /dev/null" from "pidfile_options" solves the problem.
So at least for RH8.5 there should be the same exclusion as for Ubuntu ...

Obihoernchen:
This is already the case https://github.com/xcat2/xcat-core/pull/7080

PR #6957, opened 2021-04-29
Remove erroneous = from man page

gurevichmark:
Pulled from `git cherry-pick e1996bcd13047fa500f69005d2dc16932f4937fd`

### DIFF
--- a/xCAT-client/pods/man8/copycds.8.pod
+++ b/xCAT-client/pods/man8/copycds.8.pod
@@ -4,7 +4,7 @@ B<copycds> - Copies Linux distributions and service levels from DVDs/ISOs to the
 
 =head1 SYNOPSIS
 
-B<copycds> [{B<-n|--name|--osver>}=I<distroname>] [{B<-a|--arch>}=I<architecture>] [{B<-p|--path>}=I<ospkgpath>] [B<-o>|B<--noosimage>] [B<-w>|B<--nonoverwrite>] {I<iso | device-path>} ...
+B<copycds> [{B<-n|--name|--osver>} I<distroname>] [{B<-a|--arch>} I<architecture>] [{B<-p|--path>} I<ospkgpath>] [B<-o>|B<--noosimage>] [B<-w>|B<--nonoverwrite>] {I<iso | device-path>} ...
 
 B<copycds> [B<-i>|B<--inspection>] {I<iso | device-path>}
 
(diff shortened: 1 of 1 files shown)


peterwywong (commented):
I ran copycds yesterday. The use of "=" was fine.
```
copycds <ISO file> --path=/install
```
I could be missing something.

gurevichmark:
Maybe it works for some options, but not for others?
```
[root@f6u13k04 ~]# copycds /tmp/automation_mountpoint/iso/OracleLinux-R8-U3-x86_64-dvd-stable.iso -a=x86_64
Error: [f6u13k04]: Requested distribution architecture =x86_64, but media is x86_64
[root@f6u13k04 ~]#
```

peterwywong:
fs2vm104 is running the RC version of RHEL 8.4. The following copycds command runs fine.
```
[root@fs2vm104 /]# copycds RHEL-8.4.0-20210409.0-ppc64le-dvd1.iso --path=/install/peter
Copying media to /install/peter
99.81%

Finally done.

[root@fs2vm104 /]# cd /install/peter
[root@fs2vm104 peter]# ls
AppStream  boot  extra_files.json  images      ppc                      RPM-GPG-KEY-redhat-release
BaseOS     EULA  GPL               media.repo  RPM-GPG-KEY-redhat-beta  TRANS.TBL
```
There is no complaint on "=". 

But "-p=/install/peter" does not work. "=" is considered part of the directory name.

peterwywong (commented):
fs2vm104 is running the RC version of RHEL 8.4. The following copycds commands run fine.
```
[root@fs2vm104 /]# copycds RHEL-8.4.0-20210409.0-ppc64le-dvd1.iso --path=/install/peter

[root@fs2vm104 /]# copycds RHEL-8.4.0-20210409.0-ppc64le-dvd1.iso --path /install/peter

However, "-p=/install/peter" does not work. "=" is considered part of the directory name as Mark said.
```
With this change of removing "=" in the man page, "--path=/install/peter" continues to work.

Any concern for user experience, especially from those who have been using "--path="?

Could the man page be considered not completely correct?

gurevichmark:
Interesting. It appears `GetOptions()` handles short and long flags differently.
```
root@c910f03c11k09:~# copycds -a=x86_64 /tmp/automation_mountpoint/iso/OracleLinux-R7-U9-Server-x86_64-dvd-stable.iso
Error: [c910f03c11k09]: Requested distribution architecture =x86_64, but media is x86_64


root@c910f03c11k09:~# copycds --arch=x86_64 /tmp/automation_mountpoint/iso/OracleLinux-R7-U9-Server-x86_64-dvd-stable.iso
Copying media to /install/ol7.9/x86_64
Media copy operation successful
root@c910f03c11k09:~#
```

I think changing man page is appropriate to show the proper use of the command, which will work for both short and long versions. Code is not changing, so if someone has been using `--path=`, it will still work for them.

besawn (approved):
I agree with Mark's rationale above. As short and long options are parsed differently, the updated man page seems more correct, but the code will continue to allow the long option= behavior if anyone is using it.

@peterwywong Any additional concerns or do you think we can merge this?

peterwywong:
I agree with you both. Let's merge it.

In this case, "--path=" becomes an unspecified option. Whoever know it can use it.

peterwywong (approved):

PR #6960, opened 2021-04-29
Modify test case makentp_extntpserver_value and add it to seven daily bundles

peterwywong:
Test case makentp_extntpserver_value is used to test an NTP server.

Since each weekly bundle includes its corresponding daily bundle, this test case is added to the daily bundles only.


gurevichmark (dismissed):

besawn (commented):
Does this line succeed in removing the file:
`cmd:rm /tmp/abc123`
?

If I run it manually I get prompted for confirmation, but this may not happen when the test framework is running:
```
[root@c910loginx04 ~]# rm /tmp/abc123
rm: remove regular file ‘/tmp/abc123’? y
```
I think including the `-f` flag might be more robust against errors:
```
[root@c910loginx04 ~]# rm -f /tmp/abc123 
[root@c910loginx04 ~]#
```

I think this combination of variable and temporary file could be renamed for better clarity and to better match the existing code in this file:
```
cmd:Time=`cat /tmp/abc123 | awk '{print $1}'`;date | grep $Time
```
compared to:
```
cmd:extntpserver=`cat /tmp/extntpserver | awk -F = '{print \$2}'`;chtab key=extntpservers site.value=$extntpserver
```
A few observations:
- There is no direct connection between `$Time` and `/tmp/abc123`.
- `$Time` is well named, but maybe too generic and at risk of conflicting with another variable or keyword?
- `$Time` is the only capitalized variable in this file, so it would be more consistent to keep it lowercase. 
So I would suggest some variation like this for clarity, you choose what sounds best:

```
curryear=`cat /tmp/curryear`;
ntpyear=`cat /tmp/ntpyear`;
dateyear=`cat /tmp/dateyear`;
timeyear=`cat /tmp/timeyear`;
myyear=`cat /tmp/myyear`;
etc
```
Is the awk needed here:
```
cmd:Time=`cat /tmp/abc123 | awk '{print $1}'`;date | grep $Time
 ```
 or could this be simplified to:
 ```
 cmd:Time=`cat /tmp/abc123`;date | grep $Time
 ```
 ?

peterwywong:
@besawn Your suggestions are great and have been addressed.

besawn (approved):
Looks good, thanks!

PR #6967, opened 2021-05-04
Improve verification of makentp_extntpserver_value

peterwywong:
This is an improvement on https://github.com/xcat2/xcat-core/pull/6960 .

For example, when this test case fails on RHEL 8.3 with a bad NTP server, it does not fail on **makentp** nor **makntp -V**, but fails on the next command to compare year 2000 to the current year. When this test case fails on Ubuntu 16.04.5, it fails on "makentp" right away with rc=1. That is, the failures occur at different points in the test case.

Therefore, unnecessary **output comparisons** are removed.

**check:output!~Error** is added for **makentp -V** mainly for older versions of distros, although it may not be needed since the error could have been captured by the previous command **makentp**. At any rate, this is an extra checking and does not hurt.

### DIFF
--- a/xCAT-test/autotest/testcase/makentp/cases0
+++ b/xCAT-test/autotest/testcase/makentp/cases0
@@ -54,12 +54,8 @@ cmd:if [[ "__GETNODEATTR($$CN,os)__" =~ "rhel" ]];then clock -w;else hwclock -w 
 check:rc==0
 cmd:makentp
 check:rc==0
-check:output=~configuring management node: $$MN
 cmd:makentp -V
-check:rc==0
-check:output=~configuring management node: $$MN
-check:output=~Calling ... /install/postscripts/setupntp
-check:output=~Daemon chronyd configured
+check:output!~Error
 cmd:curryear=`cat /tmp/currentyear`;date | grep $curryear
 check:rc==0
 check:output=~\d
--- a/xCAT-test/autotest/testcase/makentp/cases0
+++ b/xCAT-test/autotest/testcase/makentp/cases0
@@ -54,8 +54,6 @@ cmd:if [[ "__GETNODEATTR($$CN,os)__" =~ "rhel" ]];then clock -w;else hwclock -w 
 check:rc==0
 cmd:makentp
 check:rc==0
-cmd:makentp -V
-check:output!~Error
 cmd:curryear=`cat /tmp/currentyear`;date | grep $curryear
 check:rc==0
 check:output=~\d


xCAT-test/autotest/testcase/makentp/cases0
```diff
@@ -54,12 +54,8 @@ cmd:if [[ "__GETNODEATTR($$CN,os)__" =~ "rhel" ]];then clock -w;else hwclock -w
 check:rc==0
 cmd:makentp
 check:rc==0
-check:output=~configuring management node: $$MN
 cmd:makentp -V
-check:rc==0
-check:output=~configuring management node: $$MN
-check:output=~Calling ... /install/postscripts/setupntp
-check:output=~Daemon chronyd configured
+check:output!~Error
```
gurevichmark, line 58: It seems like `makentp.pm` can also exit with:
```
280     if (!-f "/usr/sbin/ntpd") {
281         send_msg(\%request, 1, "Please make sure ntpd is installed on $nodename.");
282         return 1;
283     }
```
and
```
362     my $rc = xCAT::Utils->stopservice($ntp_service);
363     if ($rc != 0) {
364         send_msg(\%request, 1, "Failed to stop nptd on $nodename.");
365         return 1;
366     }
```
```
468     my $rc = xCAT::Utils->startservice($ntp_service);
469     if ($rc != 0) {
470         send_msg(\%request, 1, "Failed to start nptd on $nodename.");
471         return 1;
472     }
```
So probably need to check for `Failed` and `Please make sure ntpd is installed` also ?

peterwywong: On a Ubuntu 16.04.5 node, /usr/sbin/ntpd is renamed, so cannot be found.

```
RUN:makentp [Sat Jan  1 00:00:00 2000]
ElapsedTime:0 sec
RETURN rc = 1
OUTPUT:
configuring management node: c910f03c11k12.
Please make sure ntpd is installed on c910f03c11k12.
CHECK:rc == 0   [Failed]
```

Therefore, the test case returns "Failed" and the output provides hints on what is missing. The output is given by makentp.pm.

gurevichmark: Oh, I see the testcase runs `makentp` and just checks RC, then runs `makentp -V` and just checks output. Is there a reason why it can not run `makentp -V` and check both RC and `output!~=Error`

peterwywong: The original test case runs both **makentp** and **makentp -V**. I am now thinking, probably only **makentp ** is needed.

For Ubuntu 16.04.5, consider two failures (1) a non-working NTP server, (2) /usr/sbin/nptd is not available. Here are the results of **makentp**:
```
(1) RUN:makentp [Sat Jan  1 00:00:01 2000]
ElapsedTime:21 sec
RETURN rc = 1
OUTPUT:
configuring management node: c910f03c11k12.
Error from command ntpdate -t5 10.0.0.102
     1 Jan 00:00:22 ntpdate[11191]: no server suitable for synchronization found.
Please check 10.0.0.102, make sure time is synced (can be validated by 'ntpq -p'), then rerun makentp command again
CHECK:rc == 0   [Failed]

(2) RUN:makentp [Sat Jan  1 00:00:01 2000]
ElapsedTime:0 sec
RETURN rc = 1
OUTPUT:
configuring management node: c910f03c11k12.
Please make sure ntpd is installed on c910f03c11k12.
CHECK:rc == 0   [Failed]

The check on**makentp -V** is not done once the failure is affirmed.
```
For RHEL 8.3, consider two failures (1) a non-working NTP server, (2) /usr/sbin/chronyd is not available.. Here are the results of **makentp**.
```
(1) RUN:makentp [Sat Jan  1 00:00:00 2000]
ElapsedTime:11 sec
RETURN rc = 0
OUTPUT:
configuring management node: f6u13k04.
Will configure chronyd instead.
Calling ... /install/postscripts/setupntp 10.0.0.102
Daemon chronyd configured.
CHECK:rc == 0   [Pass]

RUN:makentp -V [Sat Jan  1 00:00:11 2000]
ElapsedTime:11 sec
RETURN rc = 0
OUTPUT:
[f6u13k04]: configuring management node: f6u13k04.
[f6u13k04]: Will configure chronyd instead.
[f6u13k04]: Calling ... /install/postscripts/setupntp 10.0.0.102
[f6u13k04]: Daemon chronyd configured.
CHECK:output !~ Error   [Pass]

RUN:curryear=`cat /tmp/currentyear`;date | grep $curryear [Sat Jan  1 00:00:22 2000]
ElapsedTime:0 sec
RETURN rc = 1
OUTPUT:
CHECK:rc == 0   [Failed]

For RHEL 8.x, the time comparison captures the NTP problem. **makentp** and **makentp -V* do not.

(2) RUN:makentp [Sat Jan  1 00:00:00 2000]
ElapsedTime:0 sec
RETURN rc = 1
OUTPUT:
configuring management node: f6u13k04.
Please make sure ntpd is installed on f6u13k04.
CHECK:rc == 0   [Failed]

When chronyd is not running, the error message of **ntpd** is given by makentp.pm. This script eeds to be enhanced later for checking chronyd.
```
Without the above output, I think, for showing the failure, only **makentp** is needed. That is, **makentp -V** is not.

If we have two checks: (1) for RC (2) for output string. Once the RC shows failure, the output string is not tested further.

Based on the current changed code, I can remove **makentp -V**.  I think we have code to capture NTP errors.
(1) **cmd:makentp** is able to capture NTP errors for older distros.
(2) **cmd:curryear=`cat /tmp/currentyear`;date | grep $curryear** is able to capture errors for newer ones.

besawn: @peterwywong I think we expect the changes this PR to pass if we merge them, correct? If so, I would propose we stick with this version of the test case for now and potentially revisit this in a future effort to make additional improvements here.

besawn: After additional discussion, I think removing `makentp -V` from the testcase is a good plan and we can re-add it later in an updated form if we think it is useful.

besawn (approved):

PR #6975, opened 2021-05-13
Add initial support for deploying Ubuntu 20.04 nodes.

sjjf:
This configures automated installation using the subiquity installer, as documented at https://ubuntu.com/server/docs/install/autoinstall-quickstart and other locations on the Internet.

The new installer completely replaces the old debian-installer system with something that has lots in common with cloud-init, in particular it can be configured to use the nocloud-net datasource to get its automated install configuration (the equivalent of the old d-i preseed file).

Unfortunately, the new installer apears to only support two options for network installations: either NFS mounting the installation media (or its extracted contents), or by downloading and mounting locally an ISO image with the necessary parts of the installation media. Specifically, the installation mounts the media somehow, and then it looks for the `casper/` directory in the root of the media - this directory must contain the `filesystem.squashfs` and `installer.squashfs` files. The installer will then bring up the system using these images, and run the installer.

It's possible to create an ISO image with only the necessary files (the `casper/` directory and the `.disk` directory containing some meta-data about the installation media), but given the result is a 700MB ISO image (compared with the original 1.4GB ISO image) this is not much of an improvement.

The method that has worked best in my testing is NFS mounting the pkgdir directory (as created by copycds, e.g. /install/ubuntu20.04/x86_64/) - this bypasses any need to create ISO images, and seems to work reliably. The installer needs `netboot=nfs nfsroot=${instserver}:${pkgdir}` added to the command line to make this work.

The automated installation config is exported in a format suitable for the nocloud-net datasource. This takes the form of a directory containing a `user-data` file which is the actual configuration file (in a cloud-config like format), and a meta-data file that can be empty. This initial support creates a per-node directory and drops the processed template file into the user-data location, along with an empty meta-data file. The installer is then pointed at this config by adding `autoinstall ds=nocloud-net;s=http://${instserver}:${httpport}/install/autoinst/${node}/` to the command line. 

The rest of the kernel command line configuration is much simpler in this case, so I've made most of the debian-installer specific command line options conditional on the use of d-i.

There are a couple of other things I'd like to investigate - there are some errors in the existing Ubuntu prescript that I haven't figured out fixes for yet, the existing prescript also does some pre-configuring of partition configurations that would need to be reworked since they're d-i specific, and I'd like to create a meaningful default template for the new installer.

If this approach is generally acceptable to the xCAT developers then I'll proceed with finishing up the rest of that work.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/debian.pm
+++ b/xCAT-server/lib/xcat/plugins/debian.pm
@@ -164,6 +164,19 @@ sub using_dracut
     return 0;
 }
 
+# Check whether the new subiquity installer is used by this os
+sub using_subiquity
+{
+    my $os = shift;
+    if ($os =~ /ubuntu(\d+\.\d+)/) {
+        if ($1 >= 20.04) {
+            return 1;
+        }
+    }
+
+    return 0;
+}
+
 sub copyAndAddCustomizations {
     my $source = shift;
     my $dest   = shift;
@@ -567,6 +580,12 @@ sub mkinstall {
                         my $pltfrm = getplatform($ref->{'osvers'});
                         my $tmplfile = xCAT::SvrUtils::get_tmpl_file_name("$installroot/custom/install/$pltfrm",
                             $ref->{'profile'}, $ref->{'osvers'}, $ref->{'osarch'}, $ref->{'osvers'});
+                        if (using_subiquity($os)) {
+
+                            # in this context we use the genos parameter to search for the subiquity template
+                            $tmplfile = xCAT::SvrUtils::get_tmpl_file_name("$installroot/custom/install/$pltfrm",
+                                $ref->{'profile'}, $ref->{'osvers'}, $ref->{'osarch'}, $ref->{'osvers'}, "subiquity");
+                        }
                         if (!$tmplfile) {
                             $tmplfile = xCAT::SvrUtils::get_tmpl_file_name("$::XCATROOT/share/xcat/install/$pltfrm",
                                 $ref->{'profile'}, $ref->{'osvers'}, $ref->{'osarch'}, $ref->{'osvers'});
@@ -716,9 +735,27 @@ sub mkinstall {
             $tmperr = "Unable to find template in $installroot/custom/install/$platform or $::XCATROOT/share/xcat/install/$platform (for $profile/$os/$arch combination)";
         }
         if (-r "$tmplfile") {
+            my $autoinstfile = "$installroot/autoinst/" . $node;
+
+            # handle nocloud-net datasource for ubuntu 20.04+
+            if (using_subiquity($os)) {
+
+                # clean up existing files to make way for the new directory
+                if (-f $autoinstfile) {
+                    unlink($autoinstfile);
+                }
+                mkpath($autoinstfile);
+
+                # create empty meta-data file
+                open(my $fh, ">", $autoinstfile . "/meta-data");
+                close($fh);
+
+                # point the template output at the /user-data file
+                $autoinstfile = "$autoinstfile/user-data";
+            }
             $tmperr =
               xCAT::Template->subvars($tmplfile,
-                "$installroot/autoinst/" . $node,
+                $autoinstfile,
                 $node,
                 $pkglistfile,
                 $pkgdir,
@@ -728,7 +765,11 @@ sub mkinstall {
               );
         }
 
+        # maybe Debian will decide to use subiquity at some point?
         my $prescript = "$::XCATROOT/share/xcat/install/scripts/pre.$platform";
+        if (using_subiquity($os)) {
+            $prescript = $prescript . ".subiquity";
+        }
         my $postscript = "$::XCATROOT/share/xcat/install/scripts/post.$platform";
 
         # for powerkvm VM ubuntu LE#
@@ -801,6 +842,16 @@ sub mkinstall {
                             and $kernpath = "$pkgdir/install/netboot/vmlinuz"
                             and -r "$pkgdir/install/netboot/initrd.gz"
                             and $initrdpath = "$pkgdir/install/netboot/initrd.gz"
+                        ) or
+                        (-r "$pkgdir/casper/hwe-vmlinuz"
+                            and $kernpath = "$pkgdir/casper/hwe-vmlinuz"
+                            and -r "$pkgdir/casper/hwe-initrd"
+                            and $initrdpath = "$pkgdir/casper/hwe-initrd"
+                        ) or
+                        (-r "$pkgdir/casper/vmlinuz"
+                            and $kernpath = "$pkgdir/casper/vmlinuz"
+                            and -r "$pkgdir/casper/initrd"
+                            and $initrdpath = "$pkgdir/casper/initrd"
                         )
                     )
                 ) or (
@@ -844,7 +895,12 @@ sub mkinstall {
             if ($docopy) {
                 mkpath("$tftppath");
                 copy($kernpath, "$tftppath/vmlinuz");
-                copyAndAddCustomizations($initrdpath, "$tftppath/initrd.img");
+                # we don't want to customise the subiquity initrd
+                if (using_subiquity($os)) {
+                    copy($initrdpath, "$tftppath/initrd.img");
+                } else {
+                    copyAndAddCustomizations($initrdpath, "$tftppath/initrd.img");
+                }
             }
 
             # We have a shot...
@@ -869,10 +925,32 @@ sub mkinstall {
                 $instserver = $ent->{nfsserver};
             }
 
-            my $kcmdline = "nofb utf8 auto url=http://" . $instserver . ":$httpport/install/autoinst/" . $node;
+            my $kcmdline = "nofb utf8 auto xcatd=" . $instserver;
 
-            $kcmdline .= " xcatd=" . $instserver;
-            $kcmdline .= " mirror/http/hostname=" . $instserver.":$httpport";
+            if (using_subiquity($os)) {
+                $kcmdline .= " autoinstall ip=dhcp netboot=nfs nfsroot=${instserver}:${pkgdir}";
+                $kcmdline .= " ds=nocloud-net;s=http://${instserver}:${httpport}/install/autoinst/${node}/";
+            } else {
+                $kcmdline .= " url=http://${instserver}:$httpport/install/autoinst/$node";
+                $kcmdline .= " mirror/http/hostname=${instserver}:$httpport";
+
+                # default answers as much as possible, we don't want any interactiveness :)
+                $kcmdline .= " priority=critical";
+
+                # parse Mac table to get one mac address in case there are multiples.
+                my $net_params = xCAT::NetworkUtils->gen_net_boot_params($ent->{installnic}, $ent->{primarynic}, $mac);
+                if (exists($net_params->{nicname})) {
+                    $kcmdline .= " netcfg/choose_interface=" . $net_params->{nicname};
+                } elsif (exists($net_params->{mac})) {
+                    $kcmdline .= " netcfg/choose_interface=" . $net_params->{mac};
+                }
+
+                #from 12.10, the live install changed, so add the live-installer
+                if (-r "$pkgdir/install/filesystem.squashfs") {
+                    $kcmdline .= " live-installer/net-image=http://${instserver}:$httpport${pkgdir}/install/filesystem.squashfs";
+                }
+            }
+
             if ($maxmem) {
                 $kcmdline .= " mem=$maxmem";
             }
@@ -881,13 +959,6 @@ sub mkinstall {
             my $mac;
             if ($macent->{mac}) {
                 $mac = xCAT::Utils->parseMacTabEntry($macent->{mac}, $node);
-            }
-
-            my $net_params = xCAT::NetworkUtils->gen_net_boot_params($ent->{installnic}, $ent->{primarynic}, $mac);
-            if (exists($net_params->{nicname})) {
-                $kcmdline .= " netcfg/choose_interface=" . $net_params->{nicname};
-            } elsif (exists($net_params->{mac})) {
-                $kcmdline .= " netcfg/choose_interface=" . $net_params->{mac};
             }
 
             #TODO: dd=<url> for driver disks
@@ -913,9 +984,6 @@ sub mkinstall {
 
             #$kcmdline .= " netcfg/wireless_wep= netcfg/get_hostname= netcfg/get_domain=";
 
-            # default answers as much as possible, we don't want any interactiveness :)
-            $kcmdline .= " priority=critical";
-
             # Automatically detect all HDD
             # $kcmdline .= " all-generic-ide irqpoll";
 
@@ -930,11 +998,6 @@ sub mkinstall {
             # I don't need the timeout for ubuntu, but for debian there is a problem with getting dhcp in a timely manner
             # safer way to set hostname, avoid problems with nameservers
             $kcmdline .= " hostname=" . $node;
-
-            #from 12.10, the live install changed, so add the live-installer
-            if (-r "$pkgdir/install/filesystem.squashfs") {
-                $kcmdline .= " live-installer/net-image=http://${instserver}:$httpport${pkgdir}/install/filesystem.squashfs";
-            }
 
             xCAT::MsgUtils->trace($verbose_on_off, "d", "debian->mkinstall: kcmdline=$kcmdline kernal=$rtftppath/vmlinuz initrd=$rtftppath/initrd.img");
             $bootparams->{$node}->[0]->{kernel} = "$rtftppath/vmlinuz";
--- a/xCAT-server/lib/perl/xCAT/SvrUtils.pm
+++ b/xCAT-server/lib/perl/xCAT/SvrUtils.pm
@@ -718,8 +718,15 @@ sub update_tables_with_templates
     $genos =~ s/\..*//;
     if ($genos =~ /rh.*s(\d*)/) {
         $genos = "rhels$1";
+    } elsif ($osver =~ /ubuntu/) {
+        # for Focal and later Ubuntu we repurpose genos to indicate the use
+        # of the subiquity installer
+        if ($osver =~ /ubuntu(\d+\.\d+)/) {
+            if ($1 >= 20.04) {
+                $genos = "subiquity";
+            }
+        }
     }
-
 
     #print "osver=$osver, arch=$arch, osname=$osname, genos=$genos\n";
     my $installroot = xCAT::TableUtils->getInstallDir();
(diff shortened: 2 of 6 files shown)


gurevichmark:
@sjjf  Thank you for your contribution, it truly help us keep xCAT product up to date.
Your approach looks good to us. We hope you can continue finishing up the rest of this work.

karcaw:
What other work needs to be done to get this finished?

xCAT-server/lib/xcat/plugins/debian.pm
```diff
@@ -164,6 +164,19 @@ sub using_dracut
     return 0;
 }
 
+# Check whether the casper-based installer is used by this os
+sub using_subiquity
+{
+    my $os = shift;
+    if ($os =~ /ubuntu(\d+)/) {
+        if ($1 >= 20.04) {
```
karcaw, line 172: This comparison will not work with ubuntu20.04.02, as it only picks up the 20 part, which is always less than 20.04

karcaw:
Here is a sample Ubuntu Template file that i have working, https://gist.github.com/karcaw/61c4faa504b2d7eb7f724c3e99da9f57, i built it using the existing templates, package list, and post-code from the old d-i installer code.

No auto-paritioning, or other pre-code.

xCAT-server/lib/xcat/plugins/debian.pm
sjjf: Thanks, I've updated the regex to capture the full string.

sjjf:
Finishing this off needs (at least a preliminary list - I may find more when I've got back into the code):

- a working default template, with the same functionality or as close as possible to the old d-i version
- a fixed/adjusted pre.ubuntu script (the current version is out of date for modern Ubuntu, certainly for 20.04)
- at least trying to figure out how to support netboot (though I don't use that in my deployment so I'm not sure if I can get a working and tested implementation)
- documentation
- updates to tests (not sure if there are any unit/integration tests directly affected - any pointers would be appreciated)

I also need to figure out how to test it effectively in my own environment, but that shouldn't affect the implementation.

I'm hoping to get this all sorted in the next few weeks, since I'm now at the point where I _need_ it in my own deployment.

sjjf:
Okay, I think this is ready for serious review and potentially merging.

The change introduces support for deploying an OS that uses the subquity installer, which is (as far as I can tell) mostly an Ubuntu expansion/development of the Curtin installer for Debian based systems. This installer is built on top of cloud-init, using the nocloud datasource to get the automated installation config, which itself is an extension of the standard cloud-init config. The existing autoinstall template generation code has been extended to create a nocloud datasource directory for the  node being deployed, with the processed template being presented as the user-data file.

The vmlinuz and initrd paths on newer Ubuntu have moved under the `casper/` directory, and the mkinstall code has been modified to handle this - this change will affect Ubuntu 18.04 as well, since that version has also shifted over to using casper to provide the installation environment. However the remainder of the changes will only affect Ubuntu 20.04, and are specific to the use of the subiquity installer.

The installer is netbooted via NFS, making use of the `/install` export on the xCAT master. This directly accesses the `$pkgdir/casper/filesystem.squashfs` and `$pkgdir/casper/installer.squashfs` which provide the live system and the actual installer code. These are large files - 380MB and 250MB respectively on Ubuntu 20.04.3, and although they *can* be provided to the installation via other means like http or tftp, I found NFS was by far the simplest and most reliable approach.

I've attempted to provide a default template which closely follows the behaviour of the d-i template, and I think I've mostly succeeded with a few caveats. First, there doesn't appear to be a way to specify any kind of automated mirror selection for the apt config - it appears to be a simple manual process specifying a hostname. The selection of an appropriate default is really beyond the scope of this work, given how regional it is - if you'd like to weigh in on what to pick I'd be happy to adjust the current setting.

The second caveat is the storage configuration, and particularly the way that the `getinstdisk` script is integrated. I found I had to remove the d-i customisation to the initrd that was being used by the d-i template, and rely on the early-commands to do all the work. Because of the nature of the subiquity storage configuration I found I had to jump through some hoops to get the detected installation disk into the autoinstall configuration. I don't feel like the solution is *elegant*, but it does work.

Finally, I found I was running into issues with name resolution in the installation environment, which seemed to be related to a failure of the systemd-resolved in the environment to pick up the DNS config supplied by DHCP. This was easily resolved by removing the default symlink in `/etc/resolv.conf` with a file pointing at the xCAT master. This *may* be an issue specific to my testing environment, but it shouldn't break anything elsewhere even if it isn't strictly necessary.

Along with these changes I created a new subiquity specific pre-install script, which replaces the d-i storage config creation as well as updating the syslog service handling (running rsyslog with a config pointing it at the xCAT master).

In a number of locations I used the existing `genos` parameter to the file/template search functions in order to differentiate between the d-i template/scripts and the subiquity alternatives. This has previously been unused by anything other than RH, but I believe differentiating beween d-i and subiquity installers in the Debian context is conceptually consistent with the RH use, so I figured it was reasonable. There may be other cases where this was necessary or would be useful which I missed - if so, please point them out and I'll amend things.

I haven't even attempted to do anything with the netboot paths, as I don't actually use that configuration in my environment. I'm sure something *can* be done with Ubuntu 20.04, but I think it needs to be implemented by someone who uses that functionality, rather than a half-baked implementation by someone who doesn't use it. If you'd like some warnings or similar added to the netboot code paths to indicate that it's not supported with Ubuntu 20.04 please let me know what to add and where.

This has been tested reasonably well in my environment - it successfully deploys nodes using both the default template and my own custom templates, including running postscripts and post-boot scripts successfully. I haven't tested many of the scripts in the Ubuntu 20.04 context - there may be things that need updating, particularly since `/bin/sh` in this context is actually `dash`, which is more POSIX than `bash` (this forced some minor changes to the `getinstdisk` script, for example). However, the basics of deploying and running scripts all work, and I think the changes are sufficiently self-contained that there's no risk of breaking anything else.

If there are specific tests you'd like me to run please let me know. Otherwise, I believe this is probably safe to merge.

gurevichmark:
@sjjf Thank you for your contribution.
I tried using your changes in our x86 VM environment.
* I had to set `primarynic` node attribute to `ens3`, otherwise `rinstall` complained that it was not set
* I had to set `kcmdline` node attribute to something (just a dummy `modprobe.blacklist=nvidiafb`), otherwise `rinstall` complained that it was not set
* I had to set `nfsserver` to the IP of Management node, otherwise `nfsmount: can't parse IP address 'xxxxxx'` was displayed during boot.

My osimage definition:
```
Object name: ubuntu20.04.3-x86_64-install-compute
    imagetype=linux
    osarch=x86_64
    osname=Linux
    osvers=ubuntu20.04.3
    otherpkgdir=/install/post/otherpkgs/ubuntu20.04.3/x86_64
    pkgdir=/install/ubuntu20.04.3/x86_64
    pkglist=/opt/xcat/share/xcat/install/ubuntu/compute.ubuntu20.04.x86_64.pkglist
    profile=compute
    provmethod=install
    template=/opt/xcat/share/xcat/install/ubuntu/compute.subiquity.tmpl
```

Node starts booting, but eventually gets stuck here:
```
start: subiquity/Meta/status_GET
finish:     subiquity/Install/install/postinstall/install_openssh-server: installing openssh-server
start:     subiquity/Install/install/postinstall/install_bash: installing bash
finish:     subiquity/Install/install/postinstall/install_bash: installing bash
start:     subiquity/Install/install/postinstall/install_nfs-common: installing nfs-common
finish:     subiquity/Install/install/postinstall/install_nfs-common: Command '['systemd-cat', '--level-prefix=false', '--identifier=subiquity_log.2100', '/snap                                                            p/subiquity/2651/usr/bin/python3', '-m', 'curtin', 'system-install', '-t', '/targ                                                            get', '--', 'nfs-common']' returned non-zero exit status 100.
finish:   subiquity/Install/install/postinstall: Command '['systemd-cat', '--level-prefix=false', '--identifier=subiquity_log.2100', '/snap/subiquity/2651/usr/b                                                            bin/python3', '-m', 'curtin', 'system-install', '-t', '/target', '--', 'nfs-commo                                                            on']' returned non-zero exit status 100.
finish: subiquity/Install/install: Command '['systemd-cat', '--level-prefix=false', '--identifier=subiquity_log.2100', '/snap/subiquity/2651/usr/bin/python3', '                                                            '-m', 'curtin', 'system-install', '-t', '/target', '--', 'nfs-common']' returned                                                              non-zero exit status 100.
start:   subiquity/ErrorReporter/1634829157.309226990.install_fail/add_info
  running tar -c --transform='s/^/c910f04x12v07-logs\//' /var/crash /var/logs/installer |nc -l 8080
tar: Removing leading `/' from member names
tar: Removing leading `/' from hard link targets
finish:   subiquity/ErrorReporter/1634829157.309226990.install_fail/add_info: written to /var/crash/1634829157.309226990.install_fail.crash
start: subiquity/Error/run
start:   subiquity/Error/run/run
start:     subiquity/Error/run/run/command_0: tar -c --transform='s/^/c910f04x12v07-logs\//' /var/crash /var/logs/installer |nc -l 8080

```

Any advice on how to proceed ?

gurevichmark:
@sjjf Thank you for your contribution.
I tried using your changes in our x86 VM environment.
* I had to set `primarynic` node attribute to `ens3`, otherwise `rinstall` complained that it was not set
* I had to set `kcmdline` node attribute to something (just a dummy `modprobe.blacklist=nvidiafb`), otherwise `rinstall` complained that it was not set
* I had to set `nfsserver` to the IP of Management node, otherwise `nfsmount: can't parse IP address 'xxxxxx'` was displayed during boot.

My osimage definition:
```
Object name: ubuntu20.04.3-x86_64-install-compute
    imagetype=linux
    osarch=x86_64
    osname=Linux
    osvers=ubuntu20.04.3
    otherpkgdir=/install/post/otherpkgs/ubuntu20.04.3/x86_64
    pkgdir=/install/ubuntu20.04.3/x86_64
    pkglist=/opt/xcat/share/xcat/install/ubuntu/compute.ubuntu20.04.x86_64.pkglist
    profile=compute
    provmethod=install
    template=/opt/xcat/share/xcat/install/ubuntu/compute.subiquity.tmpl
```

Node starts booting, but eventually gets stuck here:
```
start: subiquity/Meta/status_GET
finish:     subiquity/Install/install/postinstall/install_openssh-server: installing openssh-server
start:     subiquity/Install/install/postinstall/install_bash: installing bash
finish:     subiquity/Install/install/postinstall/install_bash: installing bash
start:     subiquity/Install/install/postinstall/install_nfs-common: installing nfs-common
finish:     subiquity/Install/install/postinstall/install_nfs-common: Command '['systemd-cat', '--level-prefix=false', '--identifier=subiquity_log.2100', '/snap                                                            p/subiquity/2651/usr/bin/python3', '-m', 'curtin', 'system-install', '-t', '/targ                                                            get', '--', 'nfs-common']' returned non-zero exit status 100.
finish:   subiquity/Install/install/postinstall: Command '['systemd-cat', '--level-prefix=false', '--identifier=subiquity_log.2100', '/snap/subiquity/2651/usr/b                                                            bin/python3', '-m', 'curtin', 'system-install', '-t', '/target', '--', 'nfs-commo                                                            on']' returned non-zero exit status 100.
finish: subiquity/Install/install: Command '['systemd-cat', '--level-prefix=false', '--identifier=subiquity_log.2100', '/snap/subiquity/2651/usr/bin/python3', '                                                            '-m', 'curtin', 'system-install', '-t', '/target', '--', 'nfs-common']' returned                                                              non-zero exit status 100.
start:   subiquity/ErrorReporter/1634829157.309226990.install_fail/add_info
  running tar -c --transform='s/^/c910f04x12v07-logs\//' /var/crash /var/logs/installer |nc -l 8080
tar: Removing leading `/' from member names
tar: Removing leading `/' from hard link targets
finish:   subiquity/ErrorReporter/1634829157.309226990.install_fail/add_info: written to /var/crash/1634829157.309226990.install_fail.crash
start: subiquity/Error/run
start:   subiquity/Error/run/run
start:     subiquity/Error/run/run/command_0: tar -c --transform='s/^/c910f04x12v07-logs\//' /var/crash /var/logs/installer |nc -l 8080

```

Any advice on how to proceed ?

sjjf:
@gurevichmark Thanks for the feedback, and sorry for the delay responding.

I've fixed the use of primarynic instead of installnic.

Regarding the use of IP address versus hostname, is this something that requires changes in the code? I'm not sure if there's a viable way to get the hostnames to resolve that early in the bring-up process, so it may simply be that it can't work unless IP addresses are used.

Regarding the kcmdline reference, I note that it's documented as being deprecated and to use addkcmdline instead, but the addkcmdline documentation suggests it's not exactly equivalent - how should that be handled? Is there a mechanism for extracting the two sets of command line options (provisioning and provisioned) so that they can be used correctly, or would that need to be done in my code?

sjjf:
Just rebased against master, no additional changes as I'm waiting on further feedback.

gurevichmark:
@sjjf Thanks for the resent update and rebase.

* The `installnic` change is good. 

* For `kcmdline`, is there a need to have `GRUB_CMDLINE_LINUX=<kcmdline>` appear in the `/etc/default/grub`, after the initial install runs `late-commands`? It looks like contents of `addkcmdline` get appended to `imgargs kernel` line in `/tftpboot/xcat/xnba/nodes/<node>` file. I tried removing the `echo "GRUB_CMDLINE_LINUX=\"#TABLE:bootparams:$NODE:kcmdline#\"" >>/target/etc/default/grub;` from `xCAT-server/share/xcat/install/ubuntu/compute.subiquity.tmpl` and my VM booted ok with x86 Ubuntu20

* The list of packages in `xCAT-server/share/xcat/install/ubuntu/compute.subiquity.tmpl` is hardcoded and does not get updated from the `xCAT-server/share/xcat/install/ubuntu/compute.ubuntu20.04.x86_64.pkglist`

* If I use the list of packages hardcoded in `xCAT-server/share/xcat/install/ubuntu/compute.subiquity.tmpl`, when the installation runs, it complains about not finding `nfs-common` and `chrony`. I removed them from the list and the installation was successful. Perhaps they are not included in the iso I used `ubuntu-20.04.3-live-server-amd64.iso`

I think this PR is a great initial step for supporting Ubuntu20. We will merge it after the holidays (beginning of January 2022) and then use additional PRs to address some of the problems listed above.

besawn:
@sffj Thank you for the contribution. We are in the process of performing a final review before merging, and I just realized that this appears to be your first contribution to the xCAT project. The xCAT project require all contributors to submit a contributor license agreement (CLA) before we can accept any contributions. Would you be willing to become an official contributor to the project by completing the CLA here: https://xcat-docs.readthedocs.io/en/latest/developers/license/contributors.html?

sjjf:
@besawn I filled out the paperwork (both on my part, and from my employer) and sent the forms through to the xcat-legal list on 2021-06-18, with the message subject "xCAT Contributor License Agreement - Simon Fowler (resend)". I didn't subscribe to the list so I'm not entirely sure what happened afterwards, but as far as I can tell the message went through.

besawn:
@sjjf Sorry for the confusion, I have located your CLA and CCLA. It was an oversight on my part. Thank you again for your support of xCAT and your willingness to be a contributor. I will perform a final review on this pull request this week.

centredecalcul:
Thank you very much.

Custom partition file did not work because of the lack of the '-n' option in read dash builtin (used by base64decode function).

Using Perl instead of 'sh' to decode works in Ubuntu 20.04 installer :

```
diff /opt/xcat/lib/perl/xCAT/Template.pm*
467,468c467
<                         #$partcontent .= "base64decode</tmp/partscript.enc >/tmp/partscript\n";
<                         $partcontent .= "perl -ne 'use MIME::Base64; print decode_base64(\$_)' </tmp/partscript.enc > /tmp/partscript\n";
---
>                         $partcontent .= "base64decode</tmp/partscript.enc >/tmp/partscript\n";
```

conxuro:
> @sjjf Thanks for the resent update and rebase.
> 
>     * The `installnic` change is good.
> 
>     * For `kcmdline`, is there a need to have `GRUB_CMDLINE_LINUX=<kcmdline>` appear in the `/etc/default/grub`, after the initial install runs `late-commands`? It looks like contents of `addkcmdline` get appended to `imgargs kernel` line in `/tftpboot/xcat/xnba/nodes/<node>` file. I tried removing the `echo "GRUB_CMDLINE_LINUX=\"#TABLE:bootparams:$NODE:kcmdline#\"" >>/target/etc/default/grub;` from `xCAT-server/share/xcat/install/ubuntu/compute.subiquity.tmpl` and my VM booted ok with x86 Ubuntu20
> 
>     * The list of packages in `xCAT-server/share/xcat/install/ubuntu/compute.subiquity.tmpl` is hardcoded and does not get updated from the `xCAT-server/share/xcat/install/ubuntu/compute.ubuntu20.04.x86_64.pkglist`
> 
>     * If I use the list of packages hardcoded in `xCAT-server/share/xcat/install/ubuntu/compute.subiquity.tmpl`, when the installation runs, it complains about not finding `nfs-common` and `chrony`. I removed them from the list and the installation was successful. Perhaps they are not included in the iso I used `ubuntu-20.04.3-live-server-amd64.iso`
> 
> 
> I think this PR is a great initial step for supporting Ubuntu20. We will merge it after the holidays (beginning of January 2022) and then use additional PRs to address some of the problems listed above.

I have already tried the changes in xCAT 2.16.3, adding manually the files modified in commit [cf1dbe7](https://github.com/xcat2/xcat-core/commit/cf1dbe78bc54336bbd3a245ac6d8515cbcc775b7). I have used the ISO ubuntu-20.04.4-live-server-amd64.iso

I found the same issues and some others:

* The installer stops and throws the error "`Error: Unable to find requested field <kcmdline> from table <bootparams>, with key <node=node1>`", probably because this field is empty in the node's definition. Removing the echo "GRUB..." allows to continue the installation without errors. Maybe should be checked if it is defined before trying to use it.

* The package list hardcoded in the **template** doesn't work with the Ubuntu 20.04.4 live ISO (and I supose with 20.04.3 too), if there is no Internet connection during the installation because those packages are not in the ISO image. As defaults, only should be added the packages included in the ISO, and would be better to allow support for a separated pkglist file. I haven't tried to set up a local repo, not sure if it's supported yet.

* As said before, the **pkglist** file is not used in this new template. Adding #INCLUDE_DEFAULT_PKGLIST_PRESEED# doesn't work, because it adds the package list with spaces, but the template/subiquity uses another format. I didn't go deep into that INCLUDE function, but it needs to be adapted to the new Ubuntu installer.

* In a node with multiple disk, **getinstdisk** does not work properly. It selects as `install_disk` a random disk (sometimes sdg others sdf, etc.), when it should be the first disk (sda). I don't know if there is another option to define it explicitly in the template, but forcing install_disk="/dev/sda" at the end of getinsdisk, do the trick. Maybe needs a kind of sort in the getinstdisk when getting the devices or allowing to force one specific option in the template.

* **confignetwork** script doesn't work but mkresolvconf with systemd. In Ubuntu 20.04 NetworkManger is not used nor installed but Netplan, and this postscript is not adapted to it. Also I had problems getting IP via DHCP (the default config in netplan), not sure yet why, and if it's related to the postscript or not.

Thanks for the work, and I hope Ubuntu 20.04 and the future LTS will have full support soon!

c0nsaw:
> template=/opt/xcat/share/xcat/install/ubuntu/compute.subiquity.tmpl

Struggling with this, could you provide your   template=/opt/xcat/share/xcat/install/ubuntu/compute.subiquity.tmpl please

PR #6999, opened 2021-06-29
Doc updates for MariaDB and mypostscript.tmpl

gurevichmark:
Doc updates:
* MariaDB version 5 is not available on RHEL8, version 10 used for regression testing
* Explicitly state that tables with non `$NODE` key can not be used in `VARNAME=#TABLE:tablename:xxxxx:attribute#` in `mypostscript.tmpl` file.


peterwywong (commented):
We may not need "It does not support tables with key other than ``$NODE``." since it is stated that for Type 2: its key is ``$NODE``.

I would like to suggest the following for the second added sentence.
Some of the tables that do not use ``$NODE`` as the key are ``passwd``, ``rack``, ``token``::

gurevichmark:
I like that wording better, thanks for the suggestion @peterwywong

peterwywong (approved):

besawn (approved):

Issue #6868, opened 2020-10-29
go-xcat should display the latest version of xcat-dep package in repository


gurevichmark:
When running `go-xcat --xcat-version=devel update` the following is displayed:
```
xCAT Dependency Packages
========================

Package Name                Installed                      In Repository
------------                ---------                      -------------
elilo-xcat                  3.14-5                         3.14-4
fping                       (not installed)                3.10-4.el8
goconserver                 0.3.2-snap202005211628         0.3.2-snap201909161515
grub2-xcat                  2.02-0.76.el7.1.snap2019051602 2.02-0.76.el7.1.snap2019051602
ipmitool-xcat               1.8.18-3                       1.8.18-2
net-snmp-perl               5.8-10.el8                     5.8-10.el8
perl-AppConfig              (not installed)                1.71-13.el8
perl-Crypt-CBC              2.33-20.el8                    2.33-20.el8
perl-Crypt-Rijndael         1.13-10.el8                    1.13-10.el8
perl-Crypt-SSLeay           0.72-20.el8                    0.72-20.el8
perl-Digest-SHA1            2.13-23.el8                    2.13-23.el8
perl-Expect                 1.35-6.el8                     1.35-6.el8
perl-HTML-Form              6.03-19.el8                    6.03-19.el8
perl-HTTP-Async             0.30-2                         0.30-2
perl-IO-Stty                0.03-21.el8                    0.03-21.el8
perl-IO-Tty                 1.12-13.el8                    1.12-13.el8
perl-JSON                   2.97.001-4.el8                 2.97.001-2.el8
perl-Mail-Sender            0.903-7.el8                    0.903-7.el8
perl-Net-HTTPS-NB           0.14-2                         0.14-2
perl-Net-Telnet             3.04-9.el8                     3.04-9.el8
perl-SOAP-Lite              (not installed)                1.27-3.el8
pyodbc                      (not installed)                3.0.7-1.el8
syslinux-xcat               3.86-2                         3.86-2
xCAT-genesis-base-ppc64     2.14.5-snap201811160710        2.14.5-snap201811160710
xCAT-genesis-base-x86_64    2.14.5-snap201811190037        2.14.5-snap201811190037
xnba-undi                   1.0.3-131028                   1.0.3-131028
yaboot-xcat                 1.3.17-rc1                     1.3.17-rc1

xCAT is going to be updated.
Continue? [y/n]
```

Some packages in the `In Repository` column are older than the ones in `Installed` column. 
[...]

PR #7001, opened 2021-07-08
go-xcat provides correct versions of packages to be installed in the third column of display

peterwywong:
The PR is to fix Issue 6868 - https://github.com/xcat2/xcat-core/issues/6868.

Detailed comments are added to functions **check_repo_version_dnf** and **check_repo_version_yum**, so they are not repeated here.

There is only one case where the expected version of goconserver is not installed from xCAT 2.14.0 to xCAT 2.16.0 on a RHEL 7.6 system.

As xCAT 2.16.0 is to be installed, the expected version of goconserver 0.3.2-snap201909171016 is displayed in the 3rd column based on the "rpm -v update".
```
Package Name                Installed                      In Repository
------------                ---------                      -------------
elilo-xcat                  3.14-4                         3.14-4
fping                       (not installed)                2.4b2_to-2
goconserver                 0.2.2-snap201803292225         0.3.2-snap201909171016
```

After xCAT 2.16.0 is installed on that RHEL 7.6 system, we have:
```
[root@fs2vm102 tmp]# rpm -qa | grep xCAT-2.16
xCAT-2.16-snap202006161607.ppc64le
perl-xCAT-2.16-snap202006161607.noarch 

[root@fs2vm102 tmp]# rpm -qa | grep goconserver
goconserver-0.2.2-snap201803292225.ppc64le
```

goconserver is a required package for xCAT. **It is possible that the older version of goconserver is sufficient to support xCAT 2.16.** Therefore, the newer version is not applied.

This piece of information may not be available until the actual updating is being processed.

This does not seem to happen frequently and may not be easy to tackle. We could decide not to do more work on that.

### DIFF
--- a/xCAT-server/share/xcat/tools/go-xcat
+++ b/xCAT-server/share/xcat/tools/go-xcat
@@ -1,4 +1,4 @@
-#!/bin/bash
+
 #
 # go-xcat - Install xCAT automatically.
 #
@@ -37,7 +37,6 @@
 # 2021-05-07 Mark Gurevich <gurevich@us.ibm.com>
 #     - Add support for Oracle Linux
 #
-
 
 function usage()
 {
@@ -604,6 +603,11 @@ function check_package_version()
 #       $@      package names
 function check_repo_version_dnf()
 {
+	# Note that check_repo_version_dnf must be defined before check_repo_version_yum.
+	# "compgen -A function check_repo_version_" returns _dnf before _yum.
+	# Since RHEL 8.x distros can run yum as well. If _yum is defined before _dnf,
+	# the _dnf function will never be used.
+
 	type dnf >/dev/null 2>&1 || return 255
 	local -a name=()
 	local -a ver=()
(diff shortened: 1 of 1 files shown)


peterwywong:
This new code has been tested on:
(1) RHEL 7.6 on POWER8, xCAT 2.14.0 is installed initially, upgraded to 2.16.0 and then 2.16.3.
(2) RHEL 8.4 on POWER8, xCAT 2.16.0 is installed initially, upgraded to 2.16.2 and then 2.16.3.

An older but similar code has been tested on an x86 VM running RHEL 7.6.

They are all successful, except the goconsever issue on RHEL 7.6 as described above.

gurevichmark:
When the upgrade runs, are the entries in the log showing which version of `goconserver` is being attempted to be installed ?

gurevichmark (changes requested):
At the beginning of this file there is comment section.
It needs a bump to version number (1.0.49`), update to the copyright year (`2021`) and a new entry for the changes made.

peterwywong:
A POWER8 VM is running xCAT 2.14.0. Here is the display for **go-xcat --xcat-core=./xcat-core-2.16.0-linux.tar.bz2 --xcat-dep=./xcat-dep-2.16.0-linux.tar.bz2 install**:

```
xCAT Dependency Packages
========================

Package Name                Installed                      In Repository
------------                ---------                      -------------
elilo-xcat                  3.14-4                         3.14-4
fping                       (not installed)                2.4b2_to-2
goconserver                 0.2.2-snap201803292225         0.3.2-snap201909171016
grub2-xcat                  2.02-0.16.el7.snap201506090204 2.02-0.76.el7.1.snap2019051602
ipmitool-xcat               1.8.18-0                       1.8.18-3
lldpd                       (not installed)                0.7.15-6.1
net-snmp-perl               5.7.2-20.ael7b                 5.7.2-20.ael7b
```

As I mentioned above, this installation does not upgrade goconserver to the level displayed on the 3rd column. However, ipmitool-xcat is.

gurevichmark:
I think I know what is going on with `goconserver`.

`goconserver` is not part of `GO_XCAT_INSTALL_LIST` inside of `go-xcat`, it is installed because 2.16.0 version `xCAT.spec` file has:
`Requires: goconserver` 
As a result, when upgrading xCAT from 2.14.0 to 2.16.0, since `goconserver` version `0.2.2` is already installed, and `xCAT.spec` does not insist on a specific version of `goconserver`, it is not upgraded.
The 2.16.1 version `xCAT.spec` file has:
`Requires: goconserver >= 0.3.3`  (#6872)
With that change, upgrading to 2.16.1 or later will force the upgrade of the `goconserver`

peterwywong:
Mark, I looked at xCAT.spec and Nate explained how it is used for building the RPM.
There is no difference between **go-xcat install** and **go-xcat update** when upgrading the system from xCAT 2.14.0 to 2.16.0. The earlier go-xcat and the new go-xcat are tested.
In all four combinations, the installed goconserver 0.2.2 version is **NOT** replaced by the 0.3.2 version for xCAT 2.16.0.
Note that version 0.3.2 is displayed in the third column before actual installation.
I think Nate is fine with this and no further work is needed.

gurevichmark (approved):

besawn (approved):

Issue #7000, opened 2021-06-30
confignetworks fail to configure the correct network mask in a VLAN on top of a LACP bond.


viniciusferrao:
Hello, I think I've hit an issue when trying to configure VLANs on top of Bonds on EL8, an /16 network is being mistakenly configured as /24 after running `confignetworks`

xCAT Release and packages:
```
[root@headnode repos]# rpm -qa *xCAT*
xCAT-2.16.2-snap202105201708.x86_64
xCAT-genesis-base-ppc64-2.14.5-snap201811160710.noarch
xCAT-probe-2.16.2-snap202105201708.noarch
xCAT-genesis-scripts-ppc64-2.16.2-snap202105201708.noarch
perl-xCAT-2.16.2-snap202105201708.noarch
xCAT-server-2.16.2-snap202105201708.noarch
xCAT-client-2.16.2-snap202105201708.noarch
xCAT-genesis-scripts-x86_64-2.16.2-snap202105201708.noarch
xCAT-genesis-base-x86_64-2.14.5-snap201811190037.noarch
xCAT-buildkit-2.16.2-snap202105210049.noarch
xCAT-vlan-2.16.2-snap202105201708.noarch
```

Running OS:
```
[root@headnode repos]# cat /etc/oracle-release 
Oracle Linux Server release 8.4
```

`xcatprobe` is getting OK for everything:
```
[root@headnode repos]# xcatprobe xcatmn -i bond0
[mn]: Checking all xCAT daemons are running...                                                                    [ OK ]
[mn]: Checking xcatd can receive command request...                                                               [ OK ]
[mn]: Checking 'site' table is configured...                                                                      [ OK ]
[mn]: Checking provision network is configured...                                                                 [ OK ]
[mn]: Checking 'passwd' table is configured...                                                                    [ OK ]
[mn]: Checking important directories(installdir,tftpdir) are configured...                                        [ OK ]
[mn]: Checking SELinux is disabled...                                                                             [ OK ]
[mn]: Checking HTTP service is configured...                                                                      [ OK ]
[mn]: Checking TFTP service is configured...                                                                      [ OK ]
[mn]: Checking DNS service is configured...                                                                       [ OK ]
[mn]: Checking DHCP service is configured...                                                                      [ OK ]
[mn]: Checking NTP service is configured...                                                                       [ OK ]
[mn]: Checking rsyslog service is configured...                                                                   [ OK ]
[...]

PR #7005, opened 2021-07-16
Return the correct network mask by matching the netname exactly in nicutils.sh

peterwywong:
This PR is to fix https://github.com/xcat2/xcat-core/issues/7000 .

Mark is correct that the problem is with **grep -sq ".\*netname=$netname"** in **get_network_attr()** of **xCAT/postscripts/nicutils.sh**.
```
function get_network_attr {
   local netname=$1
   local attrname=$2

   local netline=""
   local index=1
   while [ $index -le $NETWORKS_LINES ]
   do
       eval netline=\$NETWORKS_LINE$index
       echo "$netline" | grep -sq ".*netname=$netname" && break;
       ((index+=1))
   done

   if [ $index -le $NETWORKS_LINES ]; then
       echo "$netline" | $sed -e 's/||/\n/g' | $awk -F'=' '$1 == "'$attrname'" {print $2}'   else
       return 1
   fi
}
```
The while loop examines records of Table networks. An example record is as follows:
```
<node name>: +++ netline='netname=subnet_10||net=102.6.0.0||mask=255.255.0.0||mgtifname=||gateway=||dhcpserver=||tftpserver=||nameservers=||ntpservers=||logservers=||dynamicrange=||staticrange=||staticrangeincrement=||nodehostname=||ddnsdomain=||vlanid=||domain=||mtu=||disable=||comments='
```
Assume that there is another record with netname=subnet and mask=255.255.255.0 and br0 is associated with it. And this record is inserted **after** the one above.

**grep -sq "\*.netname=subnet"** sees a match for netname=subnet_10 and does not search further. It returns mask=255.255.0.0, which is the wrong record.

By seeing **||** right after $netname, **grep -sq ".\*netname=$netname||"** does an exact match.

I did some string manipulation of the record earlier and achieved the same thing.

By noticing **||** right after $netname, Nate suggested a simpler and more elegant solution. Thanks Nate.

### DIFF
--- a/xCAT/postscripts/nicutils.sh
+++ b/xCAT/postscripts/nicutils.sh
@@ -397,7 +397,7 @@ function get_network_attr {
    while [ $index -le $NETWORKS_LINES ]
    do
        eval netline=\$NETWORKS_LINE$index
-       echo "$netline" | grep -sq ".*netname=$netname" && break;
+       echo "$netline" | grep -sq ".*netname=$netname||" && break;
        ((index+=1))
    done
 


gurevichmark (approved):

besawn (approved):

viniciusferrao:
That was fast. Thanks guys!

Issue #6934, opened 2021-04-01
Specifying VM machine type (patch included)


rbudden:
Hi,

I was looking to migrate our VM infrastructure to a modern virtual machine architecture for emulation support of ICH9. Below is a simple patch to allow the use of the Q35 (or any other) QEMU-KVM machine type by means of the 'vmothersetting' attribute of a node definition.

I'm sure this could be fancier, but suffices. Would be great if this support could be patched in.

Thanks!
-Robert

Add setting, verify w/chdef:
```
[root@xcat:~]# lsdef exp-b32-puppet | grep vmother
    vmothersetting=cpumode:host-passthrough;machine:q35
```

Make the VM, then view XML build on hypervisor:
```
[root@hypervisor~]# virsh dumpxml 1 | grep "type arch"
    <type arch='x86_64' machine='pc-q35-rhel7.6.0'>hvm</type>
```

Perl Diff/Patch:
```` perl
diff -u /opt/xcat/lib/perl/xCAT_plugin/kvm.pm.bak /opt/xcat/lib/perl/xCAT_plugin/kvm.pm|hldi
--- /opt/xcat/lib/perl/xCAT_plugin/kvm.pm.bak   2021-04-01 16:45:34.664878918 -0400
+++ /opt/xcat/lib/perl/xCAT_plugin/kvm.pm       2021-04-01 17:12:59.063725044 -0400
@@ -731,6 +731,7 @@
     my @passthrudevices;
     my $memnumanodes;
     my $cpumode;
+    my $machine_type;
     my $advsettings = undef;
     if (defined $confdata->{vm}->{$node}->[0]->{othersettings}) {
         $advsettings = $confdata->{vm}->{$node}->[0]->{othersettings};
@@ -742,6 +743,7 @@
     #pci passthrough:    "devpassthrough:<pci device name1>,<pci device name2>..."
     #memory binding:     "membind:<numa node set>"
     #cpu mode:           "cpumode:<host-model|host-passthrough>"
+    #machine type        "machine:<pc|q35>"
     if ($advsettings) {
         my @tmp_array = split ";", $advsettings;
         foreach (@tmp_array) {
@@ -767,9 +769,18 @@
             if (/cpumode:(.*)/) {
                 $cpumode = $1;
             }
+
+            if (/machine:(.*)/) {
+                $machine_type= $1;
+            }
         }
     }
 
+    # Support other machine types instead of 'pc', say 'q35'
+    if (defined $machine_type) {
[...]

PR #7021, opened 2021-08-17
Support VM machine types specified through vm.othersettings

peterwywong:
This PR is to address https://github.com/xcat2/xcat-core/issues/6934.

Thanks to rbudden for providing the code. The comments are slightly enhanced.

### DIFF
--- a/perl-xCAT/xCAT/Schema.pm
+++ b/perl-xCAT/xCAT/Schema.pm
@@ -226,7 +226,19 @@ qq{ link,ro - The file is readonly, and will be placed in tmpfs on the booted no
 'vncport' => 'Tracks the current VNC display port (currently not meant to be set',
 'textconsole' => 'Tracks the Psuedo-TTY that maps to the serial port or console of a VM',
 'powerstate' => "This flag is used by xCAT to track the last known power state of the VM.",
-'othersettings' => "This allows specifying a semicolon delimited list of key->value pairs to include in a vmx file of VMware or KVM. For partitioning on normal power machines, this option is used to specify the hugepage and/or bsr information, the value is like:'hugepage:1,bsr=2'. For KVM cpu mode use either:'cpumode:host-passthrough' or 'cpumode:host-model'. This improves performance on x86 VMs significantly. For KVM cpu pinning, this option is used to specify the physical cpu set on the host, the value is like:\"vcpupin:'0-15,^8'\",Its syntax is a comma separated list and a special markup using '-' and '^' (ex. '0-4', '0-3,^2') can also be allowed, the '-' denotes the range and the '^' denotes exclusive. For KVM memory binding, the value is like:'membind:0', restrict a guest to allocate memory from the specified set of NUMA nodes. For PCI passthrough, the value is like:'devpassthrough:pci_0001_01_00_0,pci_0000_03_00_0', the value for PCI device format also can be like:'devpassthrough:0001:01:00.1', the PCI devices are assigned to a virtual machine, and the virtual machine can use this I/O exclusively, the devices list are a list of PCI device names delimited with comma, the PCI device names can be obtained by running B<virsh nodedev-list> on the host.",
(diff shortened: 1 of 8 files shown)


peterwywong:
Note that Lines 714-718 are probably a piece of dead code:
```
    if (defined($hypcpumodel) and $hypcpumodel eq "ppc64") {
        $xtree{os}->{type}->{arch}    = "ppc64";
        $xtree{os}->{type}->{machine} = "pseries";
        delete $xtree{os}->{bios};
    }
```
The reason is that the latest POWER systems has the hypcpumodel as ppc64le, not ppc64.

In fact, {machine} is set by qemu if it is not set here.

It is definitely the case for x86_64.

gurevichmark:
> 
> 
> Note that Lines 714-718 are probably a piece of dead code:
> 
> ```
>     if (defined($hypcpumodel) and $hypcpumodel eq "ppc64") {
>         $xtree{os}->{type}->{arch}    = "ppc64";
>         $xtree{os}->{type}->{machine} = "pseries";
>         delete $xtree{os}->{bios};
>     }
> ```
> 
> The reason is that the latest POWER systems have the hypcpumodel as ppc64le, not ppc64.
> 
> In fact, {machine} is set by qemu if it is not set here.
> 
> It is definitely the case for ppc64le and x86_64.

The `ppc64` might be a valid `hypcpumodel` for Power7 hypervisors or earlier.

gurevichmark (changes requested):
Man page for the `node` object needs to be updates to indicate that `machine:` is a valid key for `vmothersetting` attribute:

image

peterwywong:
I have made significant changes to the othersettings section of Schema.pm.
In addition to changing its presentation format, I clarify the meaning of each option. Please verify their correctness, in particular the one for PCI passthrough.

peterwywong:
It turns out that Mark has a PR for the Rocky-related rst files which are waiting to be merged.

create_man_pages.py for my othersettings changes n Schema.pm also generates those Rocky-related rst files.

To avoid this PR conflict, I will close this PR. Once Mark's rst files are merged, I will then re-submit kvm.pm, Schema.pm and its associated rst files.

perl-xCAT/xCAT/Schema.pm
```diff
@@ -226,7 +226,19 @@ qq{ link,ro - The file is readonly, and will be placed in tmpfs on the booted no
...
+'othersettings' => "This is a semicolon-delimited list of key-value pairs to be included in a vmx file of VMware or KVM. DO NOT use 'chdef <node> -p vmothersetting=...' to add new options to it because chdef uses commas, not semicolons, to separate items.
+          Hugepage on POWER systems:
+             Specify the hugepage and/or bsr (Barrier Synchronization Register) values, e.g., 'hugepage:1,bsr=2'.
+          KVM CPU mode:
+             Specify how the host CPUs are utilized, e.g., 'cpumode:host-passthrough', 'cpumode:host-model'. With the passthrough mode, the performance of x86 VMs can be improved significantly.
+          KVM CPU pinning:
+             Specify which host CPUs are used, e.g., 'vcpupin:'0-15,^8', where '-' denotes the range and '^' denotes exclusion. This option allows a comma-delimited list.
```
gurevichmark, line 235: I think it should be `denotes exclusiveness` instead of `denotes exclusion`. `exclusion` means "being excluded", where as old test talks about being `exclusive`.

perl-xCAT/xCAT/Schema.pm
```diff
@@ -226,7 +226,19 @@ qq{ link,ro - The file is readonly, and will be placed in tmpfs on the booted no
...
+          KVM memory binding:
+             Specify which host memmory modules are used, e.g., 'membind:0'. This option allows a guest to access specified memory regions.
```
gurevichmark, line 237: Typo in `memmory`

gurevichmark: The old writeup says `specified set of NUMA nodes`. It is not applicable ?

PR #7043, opened 2021-10-05
Update xCAT installation docs

gurevichmark:
Update xCAT installation documentation to:
* Fix some formatting and wording
* Use latest/stable builds instead of development builds


besawn (approved):

gurevichmark:
`Schema.pm` is used both as an "executable" when xCAT is running and as a source to generate man pages. We run `create_man_pages.py` to read the `Schema.pm` and generate `.rst` files from it. This removes the need to store `.rst` files for those man pages in `git` and risk them getting out of sync with `Schema.pm`.
The 4 `.rst` files for this PR are just the files for our documentation, there is no "executable" part for them. They are their own source.

PR #7052, opened 2021-10-20
Change uefi to not do elilo

gskouson:
### This PR is to allow a genesis boot without elilo 

The particular Dell R640 systems we have don't seem to like something about elilo, but I couldn't see a way around xcat trying to use elilo.  I made this change to use boot the genesis image directly rather than having xnba boot elilo. 

I didn't see another way to avoid using elilo.


besawn:
This pull request was addressed by #7061.

PR #7062, opened 2021-11-04
Enable Oracle Linux and Rocky Linux to Run Statelite Test Cases Successfully

peterwywong:
/etc/*release contains "Red Hat" in the output on RHEL and Oracle Linux, but not Rocky Linux. The last one has "Rocky" in it. 

Rocky Linux has a problem with the following command in reg_linux_statelite_installation_hierarchy_by_nfs:
```
cmd:if cat /etc/*release |grep SUSE >/dev/null; then cp /opt/xcat/share/xcat/tools/autotest/testcase/installation/litefile_sles.csv /tmp/litefile.csv; tabrestore /tmp/litefile.csv; elif cat /etc/*release |grep "Red Hat" >/dev/null; then  tabrestore /opt/xcat/share/xcat/tools/autotest/testcase/installation/litefile.csv; fi
```
The litefile table is loaded on MN only if "/etc/*release" contains the string "Red Hat". If true, the files in litefile.csv will be copied from MN to SN. And later, CN will mount to /install on SN for provisioning.

Since Rocky Linux missed all these files, it failed on reg_linux_statelite_installation_hierarchy_by_nfs, reg_linux_statelite_installation_hierarchy_by_ramdisk and reg_linux_statelite_installation_flat.

The inclusion of Rocky "(\\|Rocky)" for OS checking in the above three statelite test cases fixed the problem.

However, Oracle Linux failed too, not due to these missing files. The console showed problems in dracut-pre-pivot.

It turns out that "ol" is not included in the subroutine "using_dracut" in anaconda.pm. Once included, reg_linux_statelite_installation_hierarchy_by_nfs ran to completion quickly.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/anaconda.pm
+++ b/xCAT-server/lib/xcat/plugins/anaconda.pm
@@ -145,7 +145,7 @@ sub process_request
 sub using_dracut
 {
     my $os = shift;
-    if ($os =~ /(rhels|rhel|centos|rocky)(\d+)/) {
+    if ($os =~ /(rhels|rhel|centos|rocky|ol)(\d+)/) {
         if ($2 >= 6) {
             return 1;
         }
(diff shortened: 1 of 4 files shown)


besawn (approved):

xmhpc:
🤙🤙🤙

PR #7100, opened 2022-01-29
Fix typos on help output for genimage and OpenStack

viniciusferrao:
There's a typo on genimage thats easy to fix, also the explanation for
Fedora is extremely outdated, on modern Fedora releases statelite is
generated automatically:

[root@headnode ~]# lsdef -t osimage
fedora35-x86_64-install-compute  (osimage)
fedora35-x86_64-install-service  (osimage)
fedora35-x86_64-netboot-compute  (osimage)
fedora35-x86_64-statelite-compute  (osimage)

Also theres a typo on OpenStack setup that grep caught, so we added
it too, for cosmetic reasons.


besawn (approved):
Thank you for the documentation improvements.

PR #7104, opened 2022-02-08
Update SLES15 service and compute template

gurevichmark:
Continue work of supporting SLES15 SP3 (#7099)

This PR:
* Adds a service node template in the new format. The compute and service node templates are identical for SLES12, so SLES15 service node template is the same as for compute node
* Remove `#` entry from the SLES15 pkglist file. On SLES, `#` in the pkglist never worked, but the error was not reported, since prior to SLES15, the AutoYaST file problems were ignored by the installer.
* Removes the extra reboot after SLES15 installation.
* Adds a check to existing testcases to verify that after installation compute node ends up in `boot` state


xCAT-test/autotest/testcase/installation/reg_linux_diskfull_installation_hierarchy
```diff
@@ -58,11 +58,7 @@ check:output=~Provision node\(s\)\: $$CN
 cmd:xdsh $$SN "if [[ -f /var/lib/dhcp/db/dhcpd.leases ]]; then cat /var/lib/dhcp/db/dhcpd.leases; elif [[ -f /var/lib/dhcpd/dhcpd.leases ]];then cat /var/lib/dhcpd/dhcpd.leases;elif [[ -f /var/lib/dhcp/dhcpd.leases ]];then cat /var/lib/dhcp/dhcpd.leases; fi"
 cmd:if [[ "__GETNODEATTR($$CN,arch)__" =~ "ppc" ]]; then sleep 120;elif [[ "__GETNODEATTR($$CN,arch)__" =~ "x86_64" ]];then sleep 200;else sleep 180;fi
 
-#cmd:lsdef -l $$CN | grep status
-#check:rc==0
-#check:output!~booted
-
-cmd:ping $$CN -c 3
+ping $$CN -c 3
```
besawn, line 61: Should this be:
`cmd:ping $$CN -c 3`
?

gurevichmark: Good catch. I do not know what happened here.

besawn (approved):

PR #7135, opened 2022-03-24
Allow SLES15 node on Power to reboot in standby mode

gurevichmark:
Because of two phase installation process on SLES15, the diskfull node winds up with `currstate=standby` after the initial install. On x86, `xCAT-server/lib/xcat/plugins/xnba.pm` removes `/tftpboot/xcat/xnba/nodes/<node>.pxelinux` which allows node to reboot from local disk. On Power `xCAT-server/lib/xcat/plugins/grub2.pm` does not remove `/tftpboot/boot/grub2/grub2-<node>`, which prevents node from rebooting from local disk.

This PR makes behavior on Power to be the same as on x86, by removing `/tftpboot/boot/grub2/grub2-<node>` when `currstate=standby`.
This RP also removes the testcase changes added by #7104 and instead modifies testcases to verify that diskfull node can be rebooted after initial installation.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/grub2.pm
+++ b/xCAT-server/lib/xcat/plugins/grub2.pm
@@ -289,7 +289,7 @@ sub setstate {
         my $os_version = $nodeos;
         $os_version =~ s/sles//i; # Strip sles if there
         $os_version =~ s/sle//i; # String sle if there
-        if ($os_version ge "15") {
+        if ($os_version >= "15") {
             # Make sure for SLES15 or higher can still boot 
             # from disk in "standby" state
             unlink("grub2-$node");
--- a/xCAT-server/lib/xcat/plugins/grub2.pm
+++ b/xCAT-server/lib/xcat/plugins/grub2.pm
@@ -100,7 +100,8 @@ sub setstate {
     my $tftpdir      = shift;
     my %nrhash       = %{ shift() };
     my $linuximghash = shift();
-    my $nodearch = shift;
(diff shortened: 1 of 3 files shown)


xCAT-server/lib/xcat/plugins/grub2.pm
```diff
@@ -284,6 +285,15 @@ sub setstate {
     chdir("$bootloader_root");
     if ($cref->{currstate} eq "offline" or $cref->{currstate} eq "boot") {
         unlink("grub2-$node");
+    } elsif ($cref->{currstate} eq "standby" and $nodeos =~ /^sle/i) {
+        my $os_version = $nodeos;
+        $os_version =~ s/sles//i; # Strip sles if there
+        $os_version =~ s/sle//i; # String sle if there
+        if ($os_version ge "15") {
```
besawn, line 292: I think this should be numeric comparison `>=` rather than string comparison `ge`.
For example, when setting `$os_version = "sle9"`
```
if ($os_version ge "15")
```
returns true (incorrect)
```
if ($os_version >= "15")
```
return false (correct)

Can you double check my logic here for me?

besawn (changes requested):

xCAT-server/lib/xcat/plugins/grub2.pm
gurevichmark: Yes, you are right `>=` works better.

besawn (approved):

PR #7136, opened 2022-03-25
Add support for ed25519 host keys

khm:
### The PR is to add support for ed25519 host keys.

This updates `plugins/credentials.pm`, `xcatconfig`, and the `remoteshell` and `aixremoteshell` postscripts to support ed25519 host key types.

Currently these keys are created as part of EL8 installs, but ignored by xcat.  If a user caches this host key type in their .known_hosts file, and a stateless node reboots, EL8 will regenerate keys and the user will get a scary warning!

With these updates xcat should generate and use these hostkeys alongside the others.

The logic in the postscripts was duplicated from the support for ECDSA key types.


besawn (approved):
Looks good, thank you for the contribution.

PR #7137, opened 2022-03-25
rh/genimage:  update systemd list to include Rocky Linux

khm:
### The modification include

_rh/genimage_: added an elsif clause to account for Rocky Linux.


besawn (approved):
Looks good, thank you.

Issue #6485, opened 2019-11-15
Console logs reset every Sunday at 3:26 logging appears to stop 


markprez:
Version 2.15 (git commit 86dd65c954220d62eb8d1765b989a63b016d0437, built Tue Nov  5 01:37:14 EST 2019)
We see our logs get reset every Sunday at 03:26am on our clusters running goconserver.
On c685mgt03 I increased gocons logging and see in 

{"file":"github.com/xcat2/goconserver/console/server.go (383)","level":"info","msg":"Handle signal: hangup, reload configuration file\n","time":"2019-11-03T03:36:01-05:00"}

After this all logs are restarted and empty but we believe we are missing data.

gurevichmark:
This appears to the be issue that added log rotation: https://github.com/xcat2/goconserver/issues/11

gurevichmark:
@markprez Do you see the rotated out logs, like `/var/log/consoles/<node>.log-<date>` in addition to `/var/log/consoles/<node>.log` ? Are both logs zero size ?

cxhong:
@markprez , looking the `/var/log/consoles`,   I saw the file rotated every 7 days.
```
[root@mgt03 nodes]# ls -ltr /var/log/consoles/f4n01*
-rw------- 1 root root    113 Oct 17 11:05 /var/log/consoles/f4n01.log-20191020
-rw------- 1 root root      0 Oct 20 03:33 /var/log/consoles/f4n01.log-20191027
-rw------- 1 root root 623925 Nov  2 18:16 /var/log/consoles/f4n01.log-20191103
-rw------- 1 root root 315105 Nov  8 12:06 /var/log/consoles/f4n01.log-20191110
-rw------- 1 root root 309432 Nov 13 21:34 /var/log/consoles/f4n01.log
[root@mgt03 nodes]# ls -ltr /var/log/consoles/c685f8ufm6*
-rw------- 1 root root 1022026 Oct 20 03:32 /var/log/consoles/c685f8ufm6.log-20191020
-rw------- 1 root root 3579206 Oct 27 03:46 /var/log/consoles/c685f8ufm6.log-20191027
-rw------- 1 root root 3587386 Nov  3 03:25 /var/log/consoles/c685f8ufm6.log-20191103
-rw------- 1 root root 3564458 Nov 10 03:19 /var/log/consoles/c685f8ufm6.log-20191110
[...]

PR #7222, opened 2022-07-27
logrotate daily rotation and compression

hillenbrkl:
As the xCAT log files might become very large, they should be rotated on a daily basis and compressed.


Obihoernchen:
You should use `compress` together with `delaycompress`, otherwise you might hit https://github.com/xcat2/xcat-core/issues/6485.

gurevichmark:
Replaced by #7277

PR #7240, opened 2022-08-23
Update xCAT-genesis-base.spec

kcgthb:
Automatically get xCAT version from installed package, rather than having to manually replace the `%%REPLACE_CURRENT_VERSION%%` string.

Without this change, the `xCAT-genesis-builder` RPMs provided at https://xcat.org/files/xcat/xcat-dep/2.x_Linux/beta/ generate RPMs named like this:
```
Wrote: /root/rpmbuild/SRPMS/xCAT-genesis-base-x86_64-%REPLACE_CURRENT_VERSION.src.rpm
Wrote: /root/rpmbuild/RPMS/noarch/xCAT-genesis-base-x86_64-%REPLACE_CURRENT_VERSION.noarch.rpm
```

### DIFF
--- a/xCAT-genesis-builder/xCAT-genesis-base.spec
+++ b/xCAT-genesis-builder/xCAT-genesis-base.spec
@@ -1,5 +1,5 @@
-%define version %(rpm -q xCAT --qf "%{VERSION}" 2>/dev/null)
-Version: %{?version:%{version}}%{!?version:%(cat Version)}
+%global version %(rpm -q xCAT --qf "%{VERSION}" 2>/dev/null | grep -Po '[0-9\.]+' || cat ../Version)
+Version: %{version}
 Release: %{?release:%{release}}%{!?release:snap%(date +"%Y%m%d%H%M")}
 %ifarch i386 i586 i686 x86
 %define tarch x86
--- a/xCAT-genesis-builder/xCAT-genesis-base.spec
+++ b/xCAT-genesis-builder/xCAT-genesis-base.spec
@@ -1,4 +1,4 @@
-%global version %(rpm -q xCAT --qf "%{VERSION}" 2>/dev/null | grep -Po %%'[0-9\.]+' || echo "2.16.5")
+%global version %(rpm -q xCAT --qf "%{VERSION}" 2>/dev/null | grep -Po '[0-9\.]+' || echo "2.16.5")
 Version: %{version}
 Release: %{?release:%{release}}%{!?release:snap%(date +"%Y%m%d%H%M")}
 %ifarch i386 i586 i686 x86
--- a/xCAT-genesis-builder/xCAT-genesis-base.spec
+++ b/xCAT-genesis-builder/xCAT-genesis-base.spec
@@ -1,4 +1,4 @@
-%define version     %%REPLACE_CURRENT_VERSION%%
+%define version %(rpm -q xCAT --qf "%{VERSION}" 2>/dev/null)
 Version: %{?version:%{version}}%{!?version:%(cat Version)}
 Release: %{?release:%{release}}%{!?release:snap%(date +"%Y%m%d%H%M")}
 %ifarch i386 i586 i686 x86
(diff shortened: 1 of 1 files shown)


gurevichmark:
@kcgthb Will this only work on the system where xCAT is installed ?

gurevichmark:
@kcgthb So do you want to change this PR?  Maybe, check for xCAT rpm, and if not there, default to `%%REPLACE_CURRENT_VERSION%%` ? Or just cancel this PR ?

kcgthb:
Yeah, I didn't think about the possibility to build the `xCAT-genesis-base` RPM on a system where xCAT is not installed. What's the use case for this, specifically? Given that any distribution can be used to generate the Genesis base, independently of the xCAT server distribution (according to the `buildrpm` [comments](https://github.com/xcat2/xcat-core/blob/9df17c8d00562ef5a8857eb5fbf3d6c04fdcd2ca/xCAT-genesis-builder/buildrpm#L4-L5)), I'm not sure to understand when building that RPM on a non-xCAT server could be useful.

Or, given that the `xCAT-genesis.base` SPEC file already contains a [check](https://github.com/xcat2/xcat-core/blob/9df17c8d00562ef5a8857eb5fbf3d6c04fdcd2ca/xCAT-genesis-builder/xCAT-genesis-base.spec#L2) on the value of `%{version}` and a fallback to the repo's `Version` file in case it's not defined, maybe just use that? The issue here is that I'm not sure this fallback is being used at this time, it doesn't look like it's looking for the `Version` file in the correct directory, nor that the check is working at all.

So what about this instead?
Try to get the xCAT version from the `xCAT` RPM if it's installed, and then rely on the repository's `Version` file if it's not.

gurevichmark:
When `rpmbuild` starts running, it runs from `/tmp/<process_id>`, so `cat ../Version` does not work and results in:
```
Building xCAT-genesis-base rpm from /root/rpmbuild/SOURCES/xCAT-genesis-base-ppc64.tar.bz2 and /tmp/xcat-core/xCAT-genesis-builder/xCAT-genesis-base.spec ...
cat: ../Version: No such file or directory
error: line 2: Empty tag: Version:
[root@f6u13k12 xCAT-genesis-builder]#
```

What do you think about something like
`%define version     %(rpm -q xCAT --qf "%{VERSION}" 2>/dev/null | grep -Po '[0-9\.]+' || echo "2.16.3")`

This way if `xCAT-genesis-base` is built on the system where xCAT server is installer or `xCAT-genesis-builder` is installed, the version will get picked up from one of the installed xCAT rpms, otherwise it will get the version of the latest version for which `xCAT-genesis-base` was updated. And if someone needs to make changes to `xCAT-genesis-base` and rebuild it, as described in https://github.com/xcat2/xcat-core/blob/master/xCAT-genesis-builder/README.md, they can manually update to the next version number.

kcgthb:
> When rpmbuild starts running, it runs from /tmp/<process_id>, so cat ../Version does not work and results in:

True, that would only work if the RPM is created directly with `rpmbuild -ba xCAT-genesis-base.spec` in the cloned `xcat-core` repository.

> This way if xCAT-genesis-base is built on the system where xCAT server is installer or xCAT-genesis-builder is installed, the version will get picked up from one of the installed xCAT rpms, otherwise it will get the version of the latest version for which xCAT-genesis-base was updated. And if someone needs to make changes to xCAT-genesis-base and rebuild it, as described in https://github.com/xcat2/xcat-core/blob/master/xCAT-genesis-builder/README.md, they can manually update to the next version number.

Yes, that's probably the best approach indeed! I changed the `.spec` file accordingly.

Thanks for the suggestion!

gurevichmark:
Is that `%%` in the `rpm -q xCAT --qf "%{VERSION}" 2>/dev/null | grep -Po %%'[0-9\.]+'` supposed to be there ?
Does not seem to work for me on RH8.6 with xCAT server installed.

kcgthb:
Ugh, sorry, the double `%` was not supposed to be there. Fixed now.

besawn (approved):

PR #7242, opened 2022-08-27
Fix non-local networks being skipped in makedhcp

lebonez:
If this logic check exists all non-local virtual networks are skipped if the site table has dhcpinterfaces defined or if the node is an enabled service node with dhcpinterfaces defined. Obviously this requires querynics to occur and to have at least one network mgtifname be set to '!remote!' I'd be interested if anyone knows the consequences if any of removing these lines? Our setup does not work since we use the site table to define interfaces on a per service/management node definition. Maybe someone also has a recommendation of best configuration for a combination of service nodes and remote (non-local) networks? Do we use the site table or service node table for dhcpinterfaces? Or do we just use the networks table by itself?

### DIFF
--- a/xCAT-server/lib/xcat/plugins/dhcp.pm
+++ b/xCAT-server/lib/xcat/plugins/dhcp.pm
@@ -1966,9 +1966,7 @@ sub process_request
         my $netif = $line[1];
         if ($netif =~ /!remote!\S+/) {
             $netif =~ s/!remote!\s*(.*)$/$1/;
-            if (!defined($activenics{"!remote!"})) {
-                next;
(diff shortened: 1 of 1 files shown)


lebonez:
Looking at this closer the only configuration that will work is to have servicenodes all defined with mgtifname set to something like "!remote!,eth0" and no dhcpinterfaces set in the site table or have no service nodes with dhcpinterfaces and have at least one network set with mgtifname = !remote!. This does not seem valid but the last one seems the most realistic.

Obihoernchen:
Thank you for your contribution. Was merged in #7567.

Issue #7246, opened 2022-09-05
Unsufficient check in credentials.pm


salben21:
Hi,
I recognize that one can request the ssh_root_key's of every zone from every client independently of its zone affiliation by a simple script using getcredentials.awk and allowcred.awk. This is a security issue if you use multiple zones with sshbetweenodes=no to avoid a privilege escalation to other nodes.

It seems that in credentials.pm of the xcatd following checks are missing:

If a ssh_root_key was requested check 
 - if the client is in the zone of the requested key's zone otherwise reject the request.
 - if the sshbetweennodes is set to yes in the requested zone, otherwise reject the request
 
best

salben21:
Workaround for xCAT-server/lib/xcat/plugins/credentials.pm from my side


        if ((($parm =~ /^ssh_root_key/) || ($parm =~ /^ssh_root_pub_key/)) && ($foundkeys == 0)) {
            my ($rootkeyparm, $zonename) = split(/:/, $parm);
            my $client_zonename=xCAT::Zone->getmyzonename($client);
            
            if ($zonename) {
                $parm = $rootkeyparm;           # take the zone off
                xCAT::MsgUtils->trace(0, 'I', "credentials: The node ($client) is asking for sshkeys of zone: $zonename.");
                if ($client_zonename eq $zonename) {
                  my $sshbetweenodes_allowed = xCAT::Zone->enableSSHbetweennodes($client);
                  if (($sshbetweenodes_allowed == 1) || ($parm =~ /^ssh_root_pub_key/)) { # check if sshbetweennodes is allowed or pub key is requested
                    $sshrootkeydir = xCAT::Zone->getzonekeydir($zonename);
                    if ($sshrootkeydir == 1) {      # error return
                        xCAT::MsgUtils->trace(0, 'W', "credentials: The zone: $zonename is not defined.");
                    } else {
                        $foundkeys = 1;    # don't want to read the zone data twice
                    }
                  } else {
                      xCAT::MsgUtils->trace(0, 'E', "credentials: Not allowed to read root's private ssh key because sshbetweennodes is disabled.");
                      $sshrootkeydir = 1;
                  }
                } else {
                    xCAT::MsgUtils->trace(0, 'E', "credentials: Not allowed to read root's private ssh key of different zone.");
                    $sshrootkeydir = 1;
                }
            } elsif ($client_zonename) { # check if no zonename is submitted but node is in a zone
[...]

PR #7247, opened 2022-09-09
Have server enforce SSH zone separation

jjohnson42:
When a client requests a zone's ssh credentials, make sure the client is consistent with the specified zone.

Per github user salben21

### The modification include

_##item1_

_##item2_

### The UT result
`##The UT output##`

_##Feature description##_

### The content of the PR:

### The UT result
`##The UT output##`


besawn (approved):
I tested the conditions described in the original issue (#7246) and the fix contained in this pull request.

The change in this pull request adds additional checking to enforce the following:
1.) Disallow distribution of root keys between zones.
2.) Disallow distribution of root keys within a zone when sshbetweennodes=no for that zone.

Thank you @salben21 and @jjohnson42 for the contributions.

PR #7256, opened 2022-09-30
Add ipmi_si module to be able to run bmcsetup manually

Obihoernchen:
Just a quick addition for easier debugging.

### DIFF
--- a/xCAT-genesis-scripts/usr/bin/bmcsetup
+++ b/xCAT-genesis-scripts/usr/bin/bmcsetup
@@ -110,6 +110,7 @@ fi
 
 # Add ipmi_devintf module to allow the ipmitool operation in-band
 modprobe ipmi_devintf
+modprobe ipmi_si
 
 for parm in `cat /proc/cmdline`; do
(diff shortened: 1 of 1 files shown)


gurevichmark:
@Obihoernchen 
Is the idea to be able to run `bmcsetup` from the command line when the node is running genesis or on OS provisioned compute node ?

Obihoernchen:
@gurevichmark I think my use case was to run bmcsetup via updatenode -P on a OS provisioned node manually which mitght not load ipmi_si by default.

gurevichmark:
@Obihoernchen 
Looking at `xCAT-genesis-scripts/usr/bin/doxcat` we do `modprobe ipmi_si` when booting with genesis, but only for `x86`.
For `ppc64` we do `modprobe ipmi_powernv` instead. Should your changes to `bmcsetup` script do the same ?

Obihoernchen:
@gurevichmark good point. Good idea I guess. I don't have a ppc64 node for testing though.
Changed.

gurevichmark (approved):

besawn (approved):

PR #7276, opened 2022-10-28
Make sure initd files are installed with xCAT-server

gurevichmark:
Put a dependency on `initscripts` (for RH) or `insserv-compat` (for SLES).
`xcatd` verifies that `/etc/init.d/functions` or `/lib/lsb/init-functions` files are there before xCAT can start.
Those files are provided by `initscripts` or `insserv-compat` RPMs. 
Usually that RPM is installed with OS install, but was missing on RH9 x86 on one of the test machines.


Obihoernchen:
yes initscripts is not installed by default on RHEL9 anymore.

Obihoernchen:
yes initscripts is not installed by default on RHEL9 anymore.

besawn (approved):

PR #7277, opened 2022-10-31
Update logrotate config

hillenbrkl:


Obihoernchen:
As @gurevichmark we don't want to force compression. Users should be able to enable/disable compression using `/etc/logrotate.conf`.

PR #7317, opened 2023-01-20
bmcsetup: revert previous meaning of bmcport for Dell servers

kcgthb:
This PR is a proposal to re-align the meaning of `ipmi.bmcport` for Dell servers.

### The modification include

Change `bmcsetup` to make sure that `bmcport=0` sets the iDRAC NIC on Dell servers as shared, and `bmcport=1` configures it to use the dedicated interface, as documented in the `ipmi` table doc.


kcgthb:
It would be great if this PR could be reviewed, and eventually merged.
Thanks!

samveen:
@gurevichmark @besawn for your kind attention.

kcgthb:
I also removed the dependency on `ipmitool delloem` and used IPMI raw commands instead, to make it more robust across Dell server generations, and less dependent on specific `ipmitool` versions.

Obihoernchen (approved):

PR #7319, opened 2023-01-24
Update docs for password encryption

gurevichmark:
### The PR is to address issue _#7312_

* Update documentation to only mention encryption when discussing `system` password. 
* Update documentation and testcases to use `sha512`.


Obihoernchen:
Please replace `openssl passwd -1` with `openssl passwd -6`, too.

besawn (approved):

PR #7320, opened 2023-01-27
Use xCAT::Utils:CheckVersion to compare BIND versions without using version.pm

peterwywong:
The PR is to improve https://github.com/xcat2/xcat-core/pull/7260.

Some older systems may not have version.pm installed, which can cause failure while doing xCAT upgrades. This PR makes use of xCAT::Utils:CheckVersion to achieve the same function.

And, there is a bug in the for loop of Checkversion:
    for ($index = 0 ; $index <= $max_index ; $index++)

The arrays start at index 0. "<=" should be "<".

When two versions are different, the loop is exited before hitting $index = $max_index. However, when they are the same, the comparison continues with $index = $max_index, which is one larger than the size of the arrays.

### DIFF
--- a/perl-xCAT/xCAT/Utils.pm
+++ b/perl-xCAT/xCAT/Utils.pm
@@ -2201,7 +2201,7 @@ sub CheckVersion
     my $index = 0;
     my $max_index = ($len_a > $len_b) ? $len_a : $len_b;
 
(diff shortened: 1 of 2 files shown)


peterwywong:
Good catch, Mark. I forgot to remove "use version".

gurevichmark (approved):

besawn (approved):

PR #7328, opened 2023-02-03
Fix EL9 chroot detection for syncfiles

Obihoernchen:
Syncfiles do not work during stateful deployment (kickstart process) on EL9 host because the chroot detection is broken on EL9. The return message changed:
    
EL9 shows the following message:

```
Running in chroot, ignoring command ...
```
    
Previously this was:

```
Running in chroot, ignoring request ...
```
    
Truncate regex to detect both.
Because of this, `sshd` was never started during kickstart and syncfiles `rsync` failed.

PS: I didn't include `command` as regex OR on purpose. Maybe they come up with yet another term in the future ;)

### DIFF
--- a/xCAT/postscripts/xcatlib.sh
+++ b/xCAT/postscripts/xcatlib.sh
@@ -430,7 +430,7 @@ function startservice {
    local retmsg
    retmsg=`$cmd 2>&1`
    retval=$?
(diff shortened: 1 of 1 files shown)


peterwywong (approved):

besawn (approved):
Looks good - thank you for the contribution!

Issue #6981, opened 2021-05-18
postscript syncfiles return with 1


aaronhcarr:
Running xCAT 2.16.1 on CentOS 8.3

Compute node installs, but postscript syncfiles return with 1.

Other postscripts complete correctly.  Mellanox OFED, etc.

Also, it's getting some files from somewhere.  /etc/passwd is current.  But the slurm.conf and munge.key files defined in synclists=/install/post/otherpkgs/slurm.synclist do not.

[root@demo~]# cat /install/post/otherpkgs/slurm.synclist
/etc/slurm/slurm.conf -> /etc/slurm/slurm.conf
/etc/munge/munge.key -> /etc/munge/munge.key

If I try to run it manually on the node, I get this:

[root@c001 ~]# /xcatpost/syncfiles
awk: /xcatpost/startsyncfiles.awk:25: fatal: can't open two way pipe `/inet/tcp/0/127.0.0.1/400' for input/output (Address family not supported by protocol)
awk: /xcatpost/startsyncfiles.awk:25: fatal: can't open two way pipe `/inet/tcp/0/127.0.0.1/400' for input/output (Address family not supported by protocol)

If I run this on the master, it works:
xdcp compute -F /install/post/otherpkgs/slurm.synclist

I never had this issue with CentOS 7.x

peterwywong:
Perhaps, it takes longer to get OpenSSL installed on CentOS 8.3.

Could you test whether it is helpful to add the following loop before **if (!system("test -f openssl"))**  in startsyncfiles.awk?
```
  count = 0
  while (system("test -f /usr/bin/openssl") && count < 10 ) {
        system("sleep 1")
        count += 1
  }
```
The number of iterations can be adjusted for testing purposes.

In addition,  **if (!system("test -f openssl"))** should be replaced by **if (!system("test -f /usr/bin/openssl"))**.

Without the full path of openssl, the former always returns 1 (not found). The negation of 1 is 0 (false), so the error message is never printed.

peterwywong:
@aaronhcarr and @buzh: Do you have CentOS 8.3 or RHEL 8.3 on both MN and CN? Do you see this problem with diskfull and/or diskless provisioning?

@buzh could you describe more about how runxcatpost is run? Is it part of the postscript processing during node provisioning? Or do you run it with updatenode after the node is booted?

I don't see /inet/tcp on my CentOS 8.3, RHEL 7.6 and RHEL8.4 nodes. Do you have it on yours? 
[...]

PR #7329, opened 2023-02-03
Make sure sshd is launched in anaconda chroot on EL8 and EL9-based distros

buzh:
### The PR is to fix issue #6981

The `remoteshell` script that is included in `xcatdefaults` intends to start sshd after anaconda goes into the %post section. On EL8 and later, this fails as there is no longer init.d compatability as there was in EL7.

This causes the `syncfiles` postscript to fail, either because `sshd` is simply not running, or if anaconda was launched with `inst.sshd` because the intended destination from the synclists file is not correct. From the sshd's context, the correct path would be `/mnt/sysroot/etc/hosts` instead of `/etc/hosts` because sshd is running outside of the chroot.

The fix detects OSVER being either rocky, rhels8 or rhels9 and starts sshd from the correct path, after the host keys are generated.

### DIFF
--- a/xCAT/postscripts/remoteshell
+++ b/xCAT/postscripts/remoteshell
@@ -640,6 +640,12 @@ else
     fi
     restartservice sshd
 fi
+
+if [[ $OSVER == rocky* || $OSVER == rhels8* || $OSVER == rhels9 ]]; then
+    pkill -9 sshd # Get rid of any previously started sshd, e.g. from anaconda inst.sshd
+    /sbin/sshd -f /etc/ssh/sshd_config || echo "ERR: Could not start sshd in chroot!"
+fi
+
 #if the service restart with "service/systemctl" failed
 #try to kill the process and start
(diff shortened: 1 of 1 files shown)


Obihoernchen:
I think this special case is not needed. Check: #7328

peterwywong:
Before the merging of https://github.com/xcat2/xcat-core/pull/7328, we have not seen failures of remoteshell and syncfiles on our RHEL 8.x or Rocky Linux 7.9 systems.

By merging the changes in https://github.com/xcat2/xcat-core/pull/7328, remoteshell and syncfiles also run successfully on our RHEL 9.0 system.

Therefore, buzh, please provide more information about the failures you observed on RHEL 8 and Rocky Linux systems, especially about the issue with host keys.

buzh:
> @buzh can you confirm that #7328 fixes your issue?

Unable to check at the moment, but happy to assume this fixes it. Thanks!

PR #7332, opened 2023-02-04
kvm: replace ac97 with ich6 for default audio support

thiell:
ac97 audio support has been removed from QEMU.
For example in RHEL 9.1 (qemu-kvm-7.0.0), creating new VMs fail to start with the following error:

	'AC97' is not a valid device model name.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1995819


gurevichmark:
@thiell Thank you for your contribution.
Any reason why you chose `ich6` and not `ich9` ?

gurevichmark (approved):

thiell:
@gurevichmark I tested the latest xcat snapshot with ich6 as a new default on EL9.1 but other than that, no particular reason, no. Thanks!

PR #7335, opened 2023-02-08
Fix mysql setup for RH family of OSes

gurevichmark:
Currently when `mysqlsetup` script is executed on RHEL, it enables `mariadb` service:
```
XCATMYSQLADMIN_PW=12345 XCATMYSQLROOT_PW=12345 /opt/xcat/bin/mysqlsetup -i -V 
:
:
Running command on c910f04x12v02: chkconfig mariadb on 2>&1

Backing up xCAT Database to /root/xcat-dbback.
:
:
```

But on Alma, Rocky and Oracle, this error is displayed:
```
XCATMYSQLADMIN_PW=12345 XCATMYSQLROOT_PW=12345 /opt/xcat/bin/mysqlsetup -i -V
:
:
Running command on c910f04x12v02: chkconfig mysql on 2>&1

Command failed: chkconfig mysql on 2>&1. Error message: error reading information on service mysql: No such file or directory.

 chkconfig mysql on failed. MySQL will not restart on reboot.
Backing up xCAT Database to /root/xcat-dbback.
:
:
```

This PR makes Alma, Rocky and Oracle behave similar to RH, and enables `mariadb` service instead of `mysql`.

### DIFF
--- a/xCAT-client/bin/mysqlsetup
+++ b/xCAT-client/bin/mysqlsetup
@@ -952,7 +952,7 @@ sub mysqlstart
     else
     {
         if ($::MariaDB == 1) {    # running MariaDB
-            if ( ($::linuxos =~ /rh.*/) || ($::linuxos =~ /sles15.*/) ) {
+            if ( $::linuxos =~ /rh.*|ol.*|rocky.*|alma.*|sles15.*/) {
                 $ret = xCAT::Utils->startservice("mariadb");
             } else {              # sles
                 $ret = xCAT::Utils->startservice("mysql");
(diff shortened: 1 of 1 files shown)


peterwywong:
Mark, I wonder why we did not see this problem on Rocky and Oracle Linux earlier. Is it because we don't use MySQL in our testing?

gurevichmark:
> Mark, I wonder why we did not see this problem on Rocky and Oracle Linux earlier. Is it because we don't use MySQL in our testing?

I think `chkconfig <service> on` sets up the service to automatically restart when the node reboots. Since we normally do not reboot the MN while regression test is running, we most likely do not see the problems because of this error.

besawn (approved):

PR #7349, opened 2023-03-01
Fix bind version comparison

gurevichmark:
### The PR is to fix PR _#7126_

That PR used incorrect version comparison to compare versions of `bind`
As a result on SLES15.4 version of bind is `9.16.20` which compared to be less that `9.16.6`.

This PR uses `sort --version-sort` to do a "natural" version comparison

### DIFF
--- a/xCAT-server/sbin/makenamed.conf
+++ b/xCAT-server/sbin/makenamed.conf
@@ -63,8 +63,19 @@ do
 	echo "		$i;"
 done >>$FILE
 echo "	};" >>$FILE
+
+# Natural version compare against version of bind.
+# If version 9.16.6 or higher, turn off DNSSEC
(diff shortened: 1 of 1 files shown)


besawn (commented):
For consistency, can we take the same approach here we used for https://github.com/xcat2/xcat-core/pull/7320?

besawn (approved):

Obihoernchen:
@gurevichmark Why not using the same approach as https://github.com/xcat2/xcat-core/pull/7320?

gurevichmark:
The `CheckVersion()` for #7320 is in Perl. Calling `sort` seemed easier than rewriting the function in shell script.

Obihoernchen:
You're right, I totally missed that. Sorry.

PR #7362, opened 2023-03-11
xcatws doesn't forward URI parameters with null values to xcatd

po1vo:
Let's say i want to zero a cell in a table, e.g. `nodech -t node -o host1 ip=` , which works perfectly fine. However doing the same via xcatws call like `PUT /xcatws/nodes/host1?ip=` does nothing: "No object definitions have been created or modified."

Please review the PR.


gurevichmark:
@po1vo 

Thanks for your contribution and support of the xCAT project. 

Before merging any pull requests, we require all users to first submit a Contributor License Agreement (CLA). The license agreement can be found here: https://xcat-docs.readthedocs.io/en/latest/developers/license/contributors.html

PR #1595, opened 2016-07-27
Add imgexport command to backup image information for developer to debug

caomengmeng:
Add imgexport command to backup image information for developer to debug

PR #7365, opened 2023-03-15
Remove imgexport command from testcase

gurevichmark:
Release testcase `reg_linux_statelite_installation_flat` on SLES15.4 fails when running `imgexport` command. The VM runs out of disk space.
Looking at the "blame" history, that command was added by #1595 for debugging purposes and appears to not be necessary. Most likely it was forgotten to be removed after debug was done.
This PR removes that command to allow the testcase to succeed.


gurevichmark:
Closing, adjusted the DB use in regression jobs. Using PostgreSQL in SLES15.4 daily run which does not run `reg_linux_statelite_installation_flat`. https://github.ibm.com/xcat2/xcat_automation_clusters/pull/167

PR #7371, opened 2023-03-31
Fix #7370 ospkgs issue

### LINKED ISSUE #7370: ospkgs not working properly on rocky8 when add osupdates to pkgdir list
The update for Rocky/RHEL 8 in the ospkgs postscript is not working properly when more directories than the default pkgdir are added, resulting in a void baseurl after adding the default repos from the main image dir, causing an error when executing the postscript (updatenode -S or updatenode -P ospkgs).

For example:
```
lsdef -t osimage rocky8.6-x86_64-install-compute -i pkgdir
Object name: rocky8.6-x86_64-install-compute
    pkgdir=/install/rocky8.6/x86_64,/install/osupdates/rocky8/x86_64/appstream,/install/osupdates/rocky8/x86_64/baseos,/install/osupdates/rocky8/x86_64/epel,/install/osupdates/rocky8/x86_64/extras,/install/osupdates/rocky8/x86_64/powertools
```

The installation set the repos OK but ospkg postscript creates 8 repositories instead of 7:
```
yum repolist
repo id                                                                                                                                  repo name
xCAT-rocky8.6-path0                                                                                                                      xCAT-rocky8.6-path0
xCAT-rocky8.6-path1                                                                                                                      xCAT-rocky8.6-path1
xCAT-rocky8.6-path2                                                                                                                      xCAT-rocky8.6-path2
xCAT-rocky8.6-path3                                                                                                                      xCAT-rocky8.6-path3
xCAT-rocky8.6-path4                                                                                                                      xCAT-rocky8.6-path4
xCAT-rocky8.6-path5                                                                                                                      xCAT-rocky8.6-path5
xCAT-rocky8.6-path6                                                                                                                      xCAT-rocky8.6-path6
xCAT-rocky8.6-path7                                                                                                                      xCAT-rocky8.6-path7
```

After the first two repos, which contain the defualt pkgdir (/install/rocky8.6/x86_64/AppStream/ /install/rocky8.6/x86_64/BaseOS/), the next one is always wrong:
```
cat /etc/yum.repos.d/xCAT-rocky8.6-path2.repo 
[xCAT-rocky8.6-path2]
name=xCAT-rocky8.6-path2
baseurl=http://
enabled=1
gpgcheck=0
skip_if_unavailable=True
```

After digging into the issue it seems a problem with a wrong control of the index for the repositories when it sets the two repos for the default pkgdir.

conxuro:
### The modification include

A fix in the **ospkgs** postscript that skips an extra increment of index variable by continuing to the next iteration of the loop that initializes the array "os_path" with all the valid repository paths.

In the specific case for RHEL 8/9 like OSes the script enters in a default_pkgdir if statement into another loop to add AppStreaam and BaseOS subdirectories and after each iteration it increases the index variable too, therefore when this loop ends continues until the end of the main loop which adds another increment to the index variable. For that reason, the proposed solution is to skip one of those increments to avoid including a void entry in the array.

Another solution could be not to make a loop for the AppStream and BaseOS subdirectories but to add two lines with the array_set_element and to increase the index once.
I.e.:
- changing this:
```
# for rhels8/9, centos8/9, ol8/9, alma8/9 and rocky8/9 the repodata is in ./BaseOS, ./AppStream, not in ./
for arg in "BaseOS" "AppStream"
do
    ospkgdir_ok="$ospkgdir/$arg"
    array_set_element os_path $index $ospkgdir_ok
    index=$(expr $index + 1)
done
continue
```
- to this:
```
# for rhels8/9, centos8/9, ol8/9, alma8/9 and rocky8/9 the repodata is in ./BaseOS, ./AppStream, not in ./
array_set_element os_path $index "$ospkgdir/AppStream"
index=$(expr $index + 1)
array_set_element os_path $index "$ospkgdir/BaseOS"
```

### DIFF
--- a/xCAT/postscripts/ospkgs
+++ b/xCAT/postscripts/ospkgs
@@ -357,6 +357,7 @@ do
                      array_set_element os_path $index $ospkgdir_ok
                      index=$(expr $index + 1)
                  done
+		 continue
             fi # if...elif..fi
          fi  # eq default_pkgdir
       fi    # match rhel*


samveen:
correct fix in #7380

viniciusferrao:
The underlying issue (#7370) was fixed in #7530, which also supersedes #7380. @Obihoernchen this can be closed. Thanks @conxuro for the original fix and testing.

PR #7376, opened 2023-04-21
The IP address of testnode1 in nodeset_disjointdhcps_petitboot, nodeset_grub2 and nodeset_petitboot must be in the subnet of MN and SN

peterwywong:
```
[root@f1n03k04 nodeset]# xdsh f1n03k05 ip route | grep src | awk '{print $10}'| cut -d '.' -f1-2
10.10

[root@f6u13k10 tmp]# xdsh f6u13k11 ip route | grep src | awk '{print $10}'| cut -d '.' -f1-2
10.6
```
This change in nodeset_disjointdhcps_petitboot was tested on f1n03k04/k05 and f6u13k10/k11. It pass on both.

### DIFF
--- a/xCAT-test/autotest/testcase/nodeset/cases0
+++ b/xCAT-test/autotest/testcase/nodeset/cases0
@@ -384,10 +384,10 @@ description: Verify the disjointdhcps feature when petitboot is used for OS load
 label:others
 cmd:lsdef testnode1 > /dev/null 2>&1;if [[ $? -eq 0 ]]; then lsdef testnode1 -z >/tmp/testnode1.stanza ;rmdef testnode1;fi
 cmd:rm -f /tftpboot/petitboot/testnode1
-cmd:mkdef -t node -o testnode1 arch=ppc64le cons=bmc groups=ipmi ip=10.1.1.200 mac=e6:d4:d2:3a:ad:06 mgt=ipmi profile=compute os=rhels7.99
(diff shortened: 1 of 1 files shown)


peterwywong:
Based on the feedback from Nate, I changed the way to extract the first two octets of IP address by getting them from "lsdef $$SN", instead of "xdsh $$SN".

This change applies to three test cases in this PR.

And one extra change is needed for nodeset_grub2.

The file under /tftpboot/boot/grub2 is removed, created, and grepped from in this test case. The file name contains the first two octets from the IP address in the hexadecimal form with a length of 2, and in upper case.

An example:
first_octet=`printf "%02x" $temp`;first_octet="${first_octet^^}";

(1) printf is used to get the hexadecimal number with the length of 2.
(2) ${<variable^^} converts 0a to 0A, ad to AD.

I tested the changes (except the removal of the grub2 file at the beginning of nodeset_grub2) on various distros and on different subnets (10.10 and 172.20). All were successful.

I later noticed that I need to make the similar change to the file removal as well.

This added change was tested on one distro and it passed too.

besawn (approved):
For future reference, this could be further simplified by replacing these sections:
`lsdef $$SN | grep ip=`
with
`lsdef $$SN -ci ip`

Example:
```
shell# lsdef sn02 -ci ip 
sn02: ip=172.20.254.2
```

Issue #6525, opened 2020-01-14
SLE 15 insserv-compat dependency missing


Obihoernchen:
Hi,

SLE 15 needs the `insserv-compat` package which provides the `/lib/lsb/init-functions` file needed for xcatd.
Otherwise xCAT fails to start with `Error, don't know how to start on this platform`. See: https://github.com/xcat2/xcat-core/blob/master/xCAT-server/etc/init.d/xcatd#L56-L74

Please add it as a RPM dependency for SLE 15.

cxhong:
the `insserv-compat` is in the Sle15
```
 # zypper info insserv-compat
Loading repository data...
Reading installed packages...


Information for package insserv-compat:
---------------------------------------
Repository     : Basesystem-Module 15-0
Name           : insserv-compat
Version        : 0.1-2.15
Arch           : noarch
Vendor         : SUSE LLC <https://www.suse.com/>
Support Level  : Level 3
Installed Size : 8.4 KiB
Installed      : Yes
Status         : up-to-date
Source package : insserv-compat-0.1-2.15.src
Summary        : Simple insserv replacement for systemd only systems
Description    :
    insserv is used enable System V init scripts. This package only
    contains a stub implemenation for use on systemd only systems.
```
for sle15, we need to two iso file
```
-rw-r--r--   1 root root  622336000 Jan  2 15:19 SLE-15-Installer-DVD-ppc64le-GM-DVD1.iso
-rw-r--r--   1 root root 5059958784 Jan  2 15:20 SLE-15-Packages-ppc64le-GM-DVD1.iso
```
did u run copycds for both iso?

cxhong:
just checking my branch again,  actually, I put `insserv` to my pkglist.  I will check in the files, should be available tomorrow's dev build.

Obihoernchen:
I know `insserv-compat` is in SLE-15 and I have no issue obtaining it.
I'm talking about adding this as a dependency to the `xcatd` RPM.

Obihoernchen:
`net-tools-deprecated` should be a dependency for SLE15 RPMs, too.

cxhong:
it merged in the PR https://github.com/xcat2/xcat-core/pull/6528

Obihoernchen:
@cxhong no I'm talking about the xCAT RPMs not the pkglists. The main xCAT RPM (for the installation on the master node) needs these dependencies for SLE15, too.

cxhong:
[...]

PR #7386, opened 2023-05-19
Add workaround for deprecated old commands in sles-15 into go-xcat

samveen:
The PR is to fix issue #6525

The go-xcat script is modified to install the packages that provide the deprecated commands before the xcat packages are installed on sles 15.


samveen:
@gurevichmark for you kind attention.

Obihoernchen (approved):

Issue #7181, opened 2022-05-23
Allow specifying path to omshell and OMAPI algoritum.


bviviano:
I am providing a patch against xCAT 2.16.3's `dhcp.pm` that will allow you to specify the path to `omshell` and the key algorithm used by OMAPI by setting values in the site table accordingly:
```
chdef -t site omshell="/opt/dhcp/bin/omshell"
chdef -t site omapi-algorithm="HMAC-SHA512"
```

If the site table entries do not exist, then the service will default to `/usr/bin/omshell` and HMAC-MD5, to maintain backwards compatibility.

This change is needed to support FIPS with xCAT when doing `makedhcp -a`.

[Per this solutions article from Red Hat](https://access.redhat.com/solutions/6443601), the `dhcp-server` RPM in RHEL8 does not support FIPS.  However, the most recent stock ISC DHCPd does support FIPS with levels up to SHA512.

With this patch, you can install the DHCPd from isc.org into an alternate location, like `/opt` and tell xCAT where to find omshell and what key algorithm to use when talking to omshell, so that `makedhcp -a` will work on a MN with FIPS enabled.

```
--- dhcp.pm-orig        2022-05-19 07:15:28.887655630 -0400
+++ dhcp.pm     2022-05-23 07:59:59.612842185 -0400
@@ -186,10 +186,24 @@
         }
     }
 
+    # Get path for omshell from site table, if set
+    my $omshellbin="/usr/bin/omshell";
+    my @oms=xCAT::TableUtils->get_site_attribute("omshell");
+    if ($oms[0]){
+        $omshellbin=$oms[0];
+    }
+    # Get HMAC algorithum from site table, if set
+    my $omapialgorithm = "HMAC-MD5";
+    my @omapia=xCAT::TableUtils->get_site_attribute("omapi-algorithm");
+    if ($omapia[0]){
+        $omapialgorithm=$omapia[0];
+    }
+
     # open ipv4 omshell file handles - $OMOUT will contain the response
-    open2($OMOUT, $OMIN, "/usr/bin/omshell ");
+    open2($OMOUT, $OMIN, "$omshellbin ");
 
     # setup omapi for the connection and check for the node requested
+    print $OMIN "key-algorithm $omapialgorithm\n";
     print $OMIN "key "
       . $omapiuser . " \""
       . $omapikey . "\"\n";
@@ -252,9 +266,10 @@
 
     # if using IPv6 addresses check using omshell IPv6 port
[...]

Issue #7173, opened 2022-05-16
Add support for HMAC-SHA256 to ddns.pm and dhcp.pm, for xcat_key.


bviviano:
I know xCAT doesn't officially support FIPS, but our site runs with FIPS and I wanted to provide a patch to ```ddns.pm``` I put together that will allow makedns and named/dhcp to work with an HMAC-SHA256 xcat_key, which is FIPS compliant, instead of the default HMAC-MD5 one.  MD5 is a pretty ancient method, and SHA256 support at this point (even for non FIPS sites) isn't unreasonable.

I'd like to request incorporating this into a future release of xCAT, as it's only a total of 4 lines of code changes and doesn't require any other updates to support "makedns" with HMAC-256 (at least that I've found).
```
--- ddns.pm-git	2022-05-13 05:45:43.859153040 -0400
+++ ddns.pm	2022-05-16 08:31:03.999754671 -0400
@@ -1358,7 +1358,7 @@
                 $ctx->{privkey} = encode_base64(genpassword(32));
                 chomp($ctx->{privkey});
             }
-            push @newnamed, "key xcat_key {\n", "\talgorithm hmac-md5;\n", "\tsecret \"" . $ctx->{privkey} . "\";\n", "};\n\n";
+            push @newnamed, "key xcat_key {\n", "\talgorithm hmac-sha256;\n", "\tsecret \"" . $ctx->{privkey} . "\";\n", "};\n\n";
             $ctx->{restartneeded} = 1;
         }
     }
@@ -1550,6 +1550,7 @@
             }
 
             my $resolver = Net::DNS::Resolver->new(nameservers => [$ip]);
+            my $keyrr = new Net::DNS::RR('xcat_key. IN KEY 512 3 163 '.$ctx->{privkey});
             my $entry;
             my $numreqs = 300; # limit to 300 updates in a payload, something broke at 644 on a certain sample, choosing 300 for now
             my $update = Net::DNS::Update->new($zone);
@@ -1564,7 +1565,7 @@
 
                     # sometimes even the xcat_key is correct, but named still replies NOTAUTH, so retry
                     for (1 .. 3) {
-                        $update->sign_tsig("xcat_key", $ctx->{privkey});
+                        $update->sign_tsig($keyrr);
                         $numreqs = 300;
                         my $reply = $resolver->send($update);
                         if ($reply) {
@@ -1586,7 +1587,7 @@
[...]

Issue #6757, opened 2020-07-02
makedhcp fails with no error RHEL7.7/7.8 with FIPS enabled


kjhee43:
RHEL7.6 with FIPS enabled:
makedhcp works as expected and dhcpd.conf is created/updated properly

RHEL7.7/7.8 with FIPS enabled: 
makedhcp executes but fails silently. Nothing is populated/updated in dhcpd.conf

cxhong:
@kjhee43 , how did u enable the FIPS?

kjhee43:
[https://access.redhat.com/solutions/137833](url)

cxhong:
the above link didn't work for me.  
same procedure documented here ?
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/chap-federal_standards_and_regulations

kjhee43:
I fixed the link. Try now.

cxhong:
Thanks.  
need to check if I have account to access it.

kjhee43:
```

How can I make RHEL 6/7/8 FIPS 140-2 compliant?
Solution Verified - Updated June 19 2020 at 9:48 AM - English
Environment

    Red Hat Enterprise Linux 6
    Red Hat Enterprise Linux 7
    Red Hat Enterprise Linux 8

Issue

    How to configure Red Hat Enterprise Linux 6 for FIPS compliance?
    How to configure Red Hat Enterprise Linux 7 for FIPS compliance?
    How to configure Red Hat Enterprise Linux 8 for FIPS compliance?
    To configure RHEL 6 or RHEL 7 to be compliant with the Federal Information Processing Standard (FIPS) Publication 140-2 (ref) several changes need to be made to ensure that accredited cryptographic modules are used. How?

Resolution

See also: How to disable FIPS mode in RHEL 6 or RHEL 7

WARNING: Ideally, when aiming for FIPS compliance, new machines should be installed from scratch with the installer booted using the fips=1 kernel argument. This will ensure the OS is running continuous tests on the RNG input used for cryptographic functions (e.g., user password hashing, LUKS key generation). It's also worth noting that RHEL often performs one-time actions post-install at firstboot that use crypto (e.g., ssh host key generation). The following practical instructions are meant for situations where it's desirable/necessary to convert an existing system to FIPS mode.

    Determine and install appropriate package versions
    While virtually every version of RHEL supports FIPS-enforcing mode, specific package versions are required if strict FIPS compliance is desired
[...]

PR #7389, opened 2023-05-26
Add other algorithms for DNS/DHCP (HMAC-SHA512)

gskouson:
### Change ddns.pm to allow for other signing algorithms and key names

Without changes to the site table, this should default to working as expected in the past.  The default is still HMAC-MD5

### Change dhcp.pm to allow for other signing algorithms and key names

Without changes to the site table, this should default to the past functionality.  

### Summary of changes

In both cases an optional site table parameter (omapi-algorithm)  is used to specify the signing algorithm for ddns and dhcp communication with the DNS server.  It adds support for HMAC-SHA1, HMAC-SHA256, HMAC-SHA384 and HMAC-SHA512 as possible signing options for DNS communications.  While additional options may also be possible the code expects the above options, otherwise HMAC-MD5 is used.

It also allows the site table parameter (omapi-username) to choose a different username for the DNS communication.  In some cases the DNS provider may require you to use the key name and secret they provide rather than allowing the use of the key name and secret provided by xCAT.

I have no specific unit tests for this change.  I've build/installed RPMs based on this change with no impacts to existing test installation.  I've successfully added the omapi-algorithm and omapi-username information to the site table and have updated the passwd table to reflect the username and secret.  It seems to build local DNS configurations and also allow for externaldns=1 and allow to point at an external DNS provider when the external provider uses the appropriate key and username.

### The UT result
No specific unit tests written for this change
`##The UT output##`

### DIFF
--- a/docs/source/advanced/domain_name_resolution/domain_name_resolution.rst
+++ b/docs/source/advanced/domain_name_resolution/domain_name_resolution.rst
@@ -90,6 +90,16 @@ For example: ::
 
 Edit **/etc/resolv.conf** to contain the cluster domain value you set in the site table's **domain**  attribute above, and to point to the same DNS server you will be using for your nodes (if you are using DNS).
 
+By default xCAT uses HMAC-MD5 as the DNS signing algorithm.  It's possible that a different algorithm is required or desired for your installation.  You can set the omapi-algorithm value in the site table to choose a different algorithm. ::
+
+      chdef -t site omapi-algorithm=HMAC-SHA256
+
+If you choose to update the algorithm, you will also need to supply a new secret for the omapi entry in the passwd table. This can be done using the dnssec-keygen command.  For example something like ::
+
+       dnssec-keygen -a HMAC-SHA256 -b 128 -n host xcat_key
+
+This should generate files with the needed key.  This key will need to be entered into the passwd table and the DNS and DHCP files will need to be recreated (makedns, makedhcp) to use the new method.
+
 Option #1: Running DNS on Your Management Node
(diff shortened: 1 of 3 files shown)


samveen:
Please review #7173 and #7181 as well (other HMAC related Issues/PRs) for related ideas

gskouson:
We reviewed both of these pull requests and this merge request should incorporate the changes from both of these with the addition of allowing a change in the key username.  

We also added info to the documentation.

I wasn't sure of the proper way to pull everything together, so I created this pull request.  I'm happy to do something differently.

gskouson:
This should also resolve the concern with #6757

sgroel:
I've implemented these changes with success, would appreciate this being merged for future release version.

samveen:
@gurevichmark @besawn for your kind attention.

samveen:
@gskouson I believe you will need to sign the xCAT CLA for individuals  and submit it for this PR to be accepted. (I'd had to do that for my first PR too.)

gskouson:
We've signed an organization agreement from Penn State.  I should be on the contributor list.

besawn:
@gskouson I have received your contributor license agreement, thank you.

sgroel:
A minor change that may be needed here:
The site table entries containing a dash in the attribute name will cause mypostscript to throw errors. This script is automatically generated and dumps the contents of the site table and exports them as variables. One of the offending entries is omapi-algorithm. The change is simple enough though, convert any dashes to underscores.

Obihoernchen:
superseded by #7563
Thank you for the contribution.

PR #7420, opened 2024-02-10
Add more disk infos for copycds

Obihoernchen:
Add more disk infos for copycds.


samveen:
disk ids for almalinux-8.x too please?

Obihoernchen:
Done

PR #7437, opened 2024-04-19
Do not rely on GPG key id

Obihoernchen:
- use the key name if possible
- instead of downloading repomd.xml.key, create it on the fly

@gurevichmark please check if this works for your environment, too.

### DIFF
--- a/build-ubunturepo
+++ b/build-ubunturepo
@@ -373,7 +373,8 @@ __EOF__
            #echo "GPGSIGN=$GPGSIGN specified, the repo will not be signed"
            echo "" >> conf/distributions
        else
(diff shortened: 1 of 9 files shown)


Obihoernchen:
Added some EL9 related fixes

gurevichmark:
@Obihoernchen 
Appears to work in our regression environment.
I generated `xcat-core` ubuntu and rpm builds using `consortium` branch.
The regression runs of x86 `rhel8.6` and `x86 `ubuntu18.04.02` were successful using that build.
I did not rebuild `xcat-dep`, the regression runs used existing `xcat-dep` package.

PR #7440, opened 2024-05-07
EL9: Adding parsing NM resolv.conf for dns entries

Obihoernchen:
Adding parsing NM resolv.conf for dns entries (if there is no lease file).

EL9 does not have `/var/lib/dhclient/*$str_inst_nic*` anymore.
Data is in `/var/run/NetworkManager/resolv.conf` now.

Try to get DNS config from this new file.


Obihoernchen:
Might fix https://github.com/xcat2/xcat-core/pull/7440

samveen:
> Might fix #7440

This is this PR itself. I expect you wanted this to be some other bug.

PR #7444, opened 2024-05-28
Use 'nmcli con modify' for nicextraparams on RHEL9

alexrichert:
### The modification include

 - update to xCAT/postscripts/nicutils.sh to include RHEL9-specific logic in `add_extra_params_nmcli()` that calls `nmcli connection modify` instead of writing to ifcfg files.
 -  updates configeth postscript to use the `nmcli con modify` approach for RHEL9 in two places.

### The UT result
`##The UT output##`

### DIFF
--- a/xCAT/postscripts/configeth
+++ b/xCAT/postscripts/configeth
@@ -68,21 +68,21 @@ function configipv4(){
             if [ "$str_nic_mtu" != "$str_default_token" ]; then
                 echo "MTU=${str_nic_mtu}" >> $str_conf_file
             fi
-	    #add extra params
-	    i=0
-	    while [ $i -lt ${#array_extra_param_names[@]} ]
-	    do
-	        name="${array_extra_param_names[$i]}"
-		value="${array_extra_param_values[$i]}"
-                echo "  $i: name=$name value=$value"
-                grep -i "${name}" $str_conf_file
-                if [ $? -eq 0 ];then
-                    sed -i "s/.*${name}.*/${name}=${value}/i" $str_conf_file
-                else
-		    echo "${name}=${value}" >> $str_conf_file
-                fi
-        	i=$((i+1))
-	    done		
+            #add extra params
+            i=0
+            while [ $i -lt ${#array_extra_param_names[@]} ]
+            do
+                name="${array_extra_param_names[$i]}"
+            value="${array_extra_param_values[$i]}"
+                    echo "  $i: name=$name value=$value"
+                    grep -i "${name}" $str_conf_file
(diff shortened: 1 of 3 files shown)


xCAT/postscripts/configeth
```diff
@@ -68,21 +68,21 @@ function configipv4(){
...
-		    echo "${name}=${value}" >> $str_conf_file
-                fi
-        	i=$((i+1))
-	    done		
+            #add extra params
+            i=0
+            while [ $i -lt ${#array_extra_param_names[@]} ]
+            do
+                name="${array_extra_param_names[$i]}"
+            value="${array_extra_param_values[$i]}"
```
samveen, line 76: please fix your code formatting settings to match the xCAT coding standards, as the indentation is no longer correct.

Obihoernchen (changes requested):
Please add matching strings for alma9* and rocky9*, too.

alexrichert:
@Obihoernchen to be clear, when you say similar logic, do you mean using nmcli calls instead of the ifcfg files?

Obihoernchen (changes requested):
There is some Space/Tab indentation mixup. Can you please fix this? Otherwise it looks good.

Obihoernchen (approved):

landcaster:
Not sure if this is the appropriate place to report this and if this is specific to our site, but this merge breaks infiniband configuration on our Rocky 9 cluster.

The error response we got in the script logs for configib is:
"nmcli Error: unknown connection" 

Digging into this, the IB looks to never have been connected via "nmcli con add " before calls  to "nmcli con modify" are made.

I modified the configib script after Line 491 to make the following call instead of trying to write an ifcfg-$nic file.   
```
elif [ $OS_name == 'redhat' ]
 then
   # First ip address
   if [ $ipindex -eq 1 ]
     ###
     if [[ "$OSVER" =~ (rhels9|alma9|rocky9) ]]; then
       prefix=$(convert_netmask_to_cidr $netmask)
       nmcli con add type infiniband con-name $nic ifname $nic ipv4.method manual  ipv4.addresses  $nicip/$prefix connection.autoconnect-priority 9
     else 
       ...
     fi
   ....
```

I can possibly work on creating a pull request to address this, but it might take me awhile as I am a bit short on free time.
As far as how to approach the problem, it is not clear to me if copying the configipv4 and configipv6 function in the configeth script with the appropriate changes would be a good approach to take as I would assume those functions have had extensive work already done to cover many of the cases one would encounter with xcat.

Obihoernchen:
Thank you for your report. I'll take a look at it.

PR #7459, opened 2024-07-25
configib: fix/clean up el9 support

alexrichert:
### The modifications include

- Remove EL9-specific logic from non-redhat OSes
- Add EL9 connections with nmcli before modifying them for nicextraparams
- Don't populate $cfgfile when using EL9

### The UT result
`##The UT output##`

### DIFF
--- a/xCAT/postscripts/configib
+++ b/xCAT/postscripts/configib
@@ -177,7 +177,7 @@ then
         exit
     fi
 	
-    if [ $OS_name != 'ubuntu' ]; then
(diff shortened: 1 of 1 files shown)


alexrichert:
@Obihoernchen I noticed a number of issues with my previous attempt, :crossed_fingers: this should work better...

alexrichert:
We've tested with EL8 and EL9. It doesn't break EL8, and in EL9, it creates a NM connection and does not try to remove or write an ifcfg file.

PR #7464, opened 2024-08-08
Fix for getting MAC table over SSH on recent ONIE switches

kcgthb:
### The modification include

_## fix remote PATH_

Run the remote `bridge` command in a login shell, to make sure PATH is properly defined

_## fix remote error propagation_

Add `set -o pipefail` to ensure that errors are properly propagated back through the remote SSH command


Obihoernchen (approved):

Obihoernchen (approved):
Thank you!

PR #7489, opened 2025-01-08
FIX: EL 9.5 installation

viniciusferrao:
This PR is aimed to fix the use of Enterprise Linux 9.5

Based on changes made by @dhilst


viniciusferrao:
I don't get why this failed. I've tested on a local server and it worked, even the failed test about reading a dynamic group.

Well that may be a "Work on my machine" case.

Obihoernchen:
I think it's a duplicate of https://github.com/xcat2/xcat-core/pull/7482
Can you please check if https://github.com/xcat2/xcat-core/pull/7482 fixes your issues, too?

viniciusferrao:
I'm usually test changes the files directly on a installed system. Yes this is bad, I know. However I wanted to compile xCAT properly to check if in fact it solves the issue, but I'm having a lot of difficulty with the `xcat-dep` repository that needs a non-existent `pokgsa` directory.

PR #7500, opened 2025-05-30
dodiscovery: when parsing DMI FRU data, only ignore the product name …

davidmonro:
…if it is entirely whitespace, rather than just containing some

### The modification include

Anchoring the regex to the beginning and end of the string


Obihoernchen (approved):

Obihoernchen:
Thank you for this fix :)

PR #7523, opened 2026-04-26
fix: harden makentp generated ntp.conf

### LINKED ISSUE #2435: makentp generates insecure configuration
`makentp` generates insecure default configuration, as it:
1. listen on all interfaces, and there doesn't seem to be a way in XCAT to specify that the MN NTP server should be restricted to specific interfaces (similar to `site.dnsinterfaces`)
2. no restriction of any type is included in the generated `ntp.conf`, especially a way to disable the monitoring facility that can be used in amplification attacks (CVE-2013-5211).

In comparison, the default RHEL/CentOS `ntp.conf` contains the following statements:
```
# For more information about this file, see the man pages
# ntp.conf(5), ntp_acc(5), ntp_auth(5), ntp_clock(5), ntp_misc(5), ntp_mon(5).

driftfile /var/lib/ntp/drift

# Permit time synchronization with our time source, but do not
# permit the source to query or modify the service on this system.
restrict default nomodify notrap nopeer noquery

# Permit all access over the loopback interface.  This could
# be tightened as well, but to do so would effect some of
# the administrative functions.
restrict 127.0.0.1
restrict ::1

# Hosts on local network are less restricted.
#restrict 192.168.1.0 mask 255.255.255.0 nomodify notrap

# Use public servers from the pool.ntp.org project.
# Please consider joining the pool (http://www.pool.ntp.org/join.html).
server 0.centos.pool.ntp.org iburst
server 1.centos.pool.ntp.org iburst
server 2.centos.pool.ntp.org iburst
server 3.centos.pool.ntp.org iburst

# Disable the monitoring facility to prevent amplification attacks using ntpdc
# monlist command when default restrict does not include the noquery flag. See
# CVE-2013-5211 for more details.
# Note: Monitoring will not be disabled with the limited restriction flag.
disable monitor
```

Generating a one-size-fits-all configuration maybe outside reach, but having a way to include an external, static configuration file in the generated `ntp.conf`, similar to what's described in #2423, would greatly improve the situation. 

I guess that would be a good thing for all the generated configuration files, such as:
* `/etc/named.conf`
* `/etc/hosts`
* `/etc/dhcp/dhcpd.conf`
* `/etc/ntp.conf`
* `/etc/conserver.cf`

viniciusferrao:
`ntpd` config generated by `makentp` had no access restrictions, explicitly disabled authentication, and was vulnerable to CVE-2013-5211 amplification attacks.

This PR adds:
* `restrict` lines to block unauthorized access
* `iburst` for faster initial sync
* replace "disable auth" with "disable monitor".

Security hardening applies to all platforms.

Fixes #2435

### DIFF
--- a/xCAT-server/lib/xcat/plugins/makentp.pm
+++ b/xCAT-server/lib/xcat/plugins/makentp.pm
@@ -320,14 +320,17 @@ sub process_request {
       or xCAT::MsgUtils->message('SE',
         "Cannot open $ntpcfg for NTP update. \n");
 
+    print CFGFILE "restrict default nomodify notrap nopeer noquery\n";
+    print CFGFILE "restrict 127.0.0.1\n";
+    print CFGFILE "restrict ::1\n\n";
+
     if (defined($ntp_servers) && $ntp_servers) {
         my @npt_server_array = split(',', $ntp_servers);
 
         # add ntp servers one by one
         foreach my $ntps (@npt_server_array) {
             if (!$ntp_master) { $ntp_master = $ntps; }
-            print CFGFILE "server ";
-            print CFGFILE "$ntps\n";
+            print CFGFILE "server $ntps iburst\n";
         }
     }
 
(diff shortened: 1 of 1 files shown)


viniciusferrao:
Closed that by mistake @Obihoernchen.

Good to go.

Obihoernchen:
Could adding restrict and removing disable auth break existing setups?

viniciusferrao:
> Could adding restrict and removing disable auth break existing setups?

Yes, it can break some custom setups that we probably didn't support, but it should not break the normal xCAT NTP flow.

`restrict default nomodify notrap nopeer noquery` line still allows clients to sync time. It does not use `ignore`, `noserve`, or `notrust`, so compute nodes, which is what we care, can still be able to use the headnode as their NTP source, and the headnode can still be able to sync from the configured upstream servers.

Those options are the default in any modern Linux distribution.

Removing `disable auth` should also be okay for the generated config. We are not adding or enforcing any authentication. It just stops explicitly disabling `ntpd` auth subsystem. Which should be default anyway.

So at the end of the day I can't see any issue with this one. Those who actually need this configuration, which I really doubt, should change it themselves. The ideia is to have sane defaults.

Obihoernchen (approved):

PR #7536, opened 2026-05-02
fix: invalidate NodeRange caches inherited across fork

viniciusferrao:
This PR aims to fix the annoying CI/CD error due to a race condition, while fixing a real bug that may occur in production.

- Fix non-deterministic CI failures in group-definition regression tests (`chdef_group`, `mkdef_group`, `rmdef_group`) caused by stale in-memory caches surviving across `fork()`.
- Track the PID at cache-build time in `NodeRange.pm` and force a fresh database read whenever a forked child detects it inherited caches from its parent.

## Root Cause

`xcatd` forks a child process for each plugin request. `NodeRange.pm` maintains module-level caches (`@allnodeset`, `%allgrphash`, `@grplist`) with a 5-second TTL. When a child is forked, it inherits these caches **and their timestamps** from the parent process.

If the parent populated caches within the past 5 seconds (from handling a prior request), the forked child treats them as valid and skips the database read. This means database changes committed between cache population and the fork—such as `mkdef -t group` updating the `nodelist.groups` column—are invisible to the child.

The `lsdef -s <groupname>` test step calls `noderange(groupname, 0)` inside the plugin child. With a stale `%allgrphash`, the group name cannot be resolved to its members. Since `verify=0`, `expandatom` returns the group name as a literal, the plugin looks for a **node** with that name, and fails with:

```
Error: Could not find an object named 'testgrp1' of type 'node'.
```

## Why It Is Non-Deterministic

- Tests pass when >5 seconds elapse between group creation and `lsdef` (cache expires naturally).
- Tests pass when the parent had no recent NodeRange activity (no inherited cache).
- Tests fail when the CI runner is fast and the parent served another request within 5 seconds, populating a cache that the child then inherits stale.

## Fix

Add a `$cachepid` variable that records `$$` at cache-build time. All three cache-validity checks now also require `$cachepid == $$`, ensuring forked children always rebuild from the database on first use.

## Notes

This one was extremely hard to find, and I had to ask Mr. Claude for help.

### DIFF
--- a/perl-xCAT/xCAT/NodeRange.pm
+++ b/perl-xCAT/xCAT/NodeRange.pm
@@ -41,8 +41,11 @@ my $glstamp         = 0;
 my $allnodesetstamp = 0;
 my $allgrphashstamp = 0;
 my %allgrphash;
-my $retaincache  = 0;
-my $recurselevel = 0;
+my $retaincache      = 0;
+my $recurselevel     = 0;
+my $nodeset_pid      = 0;
+my $grplist_pid      = 0;
+my $allgrphash_pid   = 0;
 
 my @cachedcolumns;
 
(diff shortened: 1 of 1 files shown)


Obihoernchen (approved):

Obihoernchen (approved):
Awesome <3 Thank you!

PR #7543, opened 2026-05-04
feat: Use systemd instead of legacy initscripts

### LINKED ISSUE #7343: /etc/init.d/xcatd missing the right path for RH9 (xCAT 2.16.5)
Hello,

This might be solved already or i missed something (because RH9 install now working nicely, with epel9), but last checked, the right path for rhel9 is still missing on **/etc/init.d/xcatd** 
A simple code like the one below should be added. 
Also, initscripts is needed for this functions file. 
```
elif [ -f /etc/rc.d/init.d/functions ]; then
  #echo RH9
  . /etc/rc.d/init.d/functions
  START_DAEMON=daemon
  STATUS=MStatus
  LOG_SUCCESS=RHSuccess
  LOG_FAILURE=RHFailure
  LOG_WARNING=passed
```

THX!

viniciusferrao:
This PR fixes a long-standing issue with xCAT that was a proper systemd setup:

- Use a native systemd `xcatd.service` start/stop path instead of delegating to the SysV init script.
- Keep SysV fallback support for non-systemd boots and old releases while dropping the EL7+ runtime dependency on `initscripts`/`chkconfig`.
- The same for Ubuntu 16.04+
- Make Debian maintainer scripts use systemd only when systemd is actually booted.
- Keep generated DEBs readable by old supported Ubuntu releases
- Update the `xcatd` autotest service commands to dispatch through systemd when available.

Fixes xcat2/xcat-core#7343.

### DIFF
--- a/perl-xCAT/debian/rules
+++ b/perl-xCAT/debian/rules
@@ -36,8 +36,8 @@ binary-arch: build install
 	dh_installdocs
 #	dh_installexamples
 	dh_install -X".svn"
-	chmod 644 `pwd`/debian/perl-xcat/opt/xcat/lib/perl/xCAT/*
-	chmod 755 `pwd`/debian/perl-xcat/opt/xcat/lib/perl/xCAT/data
+	find `pwd`/debian/perl-xcat/opt/xcat/lib/perl/xCAT -type d -exec chmod 755 {} +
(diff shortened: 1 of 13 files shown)


Obihoernchen (approved):
awesome 👍

PR #7546, opened 2026-05-05
fix: move kernel quiet flag from hardcoded to osimage default

### LINKED ISSUE #6916: [suggestion] remove `quiet` from kernel command line in /tftpboot files
Hi all, 

I noticed that pretty much all the boot configuration files generated in `/tftpboot` are set to enable quiet boot (ie. suppress the kernel messages during boot). It seems to be hard-coded in many of the plugins:
https://github.com/xcat2/xcat-core/blob/master/xCAT-server/lib/xcat/plugins/anaconda.pm#L1334
https://github.com/xcat2/xcat-core/blob/master/xCAT-server/lib/xcat/plugins/mknb.pm#L302
etc.

What's the rationale behind this? Wouldn't it be better to keep nodes booting in verbose mode, to facilitate diagnosing issues during discovery or deployment?

viniciusferrao:
This PR modified the `quiet` kernel parameter behaviour as discussed in #6916. The kernel parameter was hardcoded in `anaconda.pm` and `sles.pm`, making it impossible to get verbose boot output without editing the plugin source. The existing `addkcmdline` mechanism only appends to the kernel command line, so there was no clean way to remove `quiet`.

This moves `quiet` out of the plugin `kcmdline` construction and into the `linuximage.addkcmdline` default, set during `copycds` osimage creation.

## How it works

```bash
# New osimage gets quiet by default
$ copycds /iso/rhels9.iso
$ lsdef -t osimage rhels9-x86_64-install-compute -i addkcmdline
  addkcmdline=quiet

# Want verbose boot for debugging?
$ chdef -t osimage rhels9-x86_64-install-compute addkcmdline=""
$ nodeset cn-01 osimage=rhels9-x86_64-install-compute
# boots with verbose kernel output

# Want quiet back?
$ chdef -t osimage rhels9-x86_64-install-compute addkcmdline="quiet"
```

## Caveats
- Existing osimages are not modified on `copycds` re-run -- only new rows get the default

## Behaviour matrix

| Test | Before patch | After patch |
|------|-------------|-------------|
| PXE config with default `addkcmdline=""` | `quiet inst.repo=...` | `inst.repo=...` (no quiet) |
| PXE config with `addkcmdline=quiet` | n/a | `inst.repo=... quiet` (quiet from addkcmdline) |
| PXE config with `addkcmdline=""` | n/a | `inst.repo=...` (no quiet) |
| Toggle works via `chdef` | No | Yes -- per-osimage control |

Fixes #6916

### DIFF
--- a/xCAT-server/lib/perl/xCAT/SvrUtils.pm
+++ b/xCAT-server/lib/perl/xCAT/SvrUtils.pm
@@ -909,6 +909,11 @@ sub update_tables_with_templates
                             pkglist      => $pkglistfile,
                             otherpkglist => $otherpkgsfile,
                             otherpkgdir => "$installroot/post/otherpkgs/$osver/$arch");
+                        my ($existing) = $linuximagetab->getAttribs({ imagename => $imagename }, 'pkgdir');
+                        unless ($existing) {
+                            $tb_cols{addkcmdline} = "quiet";
+                        }
+
                         $linuximagetab->setAttribs(\%key_col, \%tb_cols);
 
                     } else {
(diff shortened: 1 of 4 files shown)


Obihoernchen (approved):
Yeah that's a good solution.

PR #7584, opened 2026-07-11
feat: add xCAT-release package for DNF systems

viniciusferrao:
Adds a `xcat-release` RPM with the xCAT repository definitions and signing key to ease the installation of xCAT in Enterprise Linux.


dhilst (dismissed):

buildrpms.pl
```diff
@@ -134,6 +134,7 @@ sub arch {
     xCAT-server
     xCAT-test
     xCAT-vlan
+    xcat-release
```
dhilst: Shouldn't this be xCAT-release?

dhilst (changes requested):

buildrpms.pl
viniciusferrao: You used my own spell against myself.

Yes you're right.

dhilst (approved):

PR #7585, opened 2026-07-11
fix(xcat-client): make profile probe safe under errexit

viniciusferrao:
The generated `/etc/profile.d/xcat.sh` ran the Perl path probe as a bare pipeline, so `grep` returning 1 triggered `set -e` before the following status check.

This moves the probe into a negated `if` condition and adds regression coverage for missing and present `/usr/local/share/perl5`, including `pipefail`.

Validation:
- `prove xCAT-test/unit/xcat_profile_errexit.t` under Bash 3.2 and 5.3
- `perlcritic xCAT-test/unit/xcat_profile_errexit.t`
- `rpmspec --target x86_64-redhat-linux -P xCAT-client/xCAT-client.spec`
- ShellCheck on the rendered profile

### DIFF
--- a/xCAT-client/xCAT-client.spec
+++ b/xCAT-client/xCAT-client.spec
@@ -318,8 +318,7 @@ MANPATH=\$XCATROOT/share/man:\$MANPATH
 export XCATROOT PATH MANPATH
 export PERL_BADLANG=0
 # If /usr/local/share/perl5 is not already in @INC, add it to PERL5LIB
-perl -e "print \"@INC\"" | grep -E "(^|\W)/usr/local/share/perl5($| )" > /dev/null
-if [ \$? = 1 ]; then
+if ! perl -e "print \"@INC\"" | grep -qE "(^|\W)/usr/local/share/perl5($| )"; then
     export PERL5LIB=/usr/local/share/perl5:\$PERL5LIB
 fi
 EOF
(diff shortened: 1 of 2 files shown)


dhilst:
The fix is fine, the problem is that the test will not ran automatically. Another complication is that this tests the .spec not the profile generated by the spec, so it will not be runnable by xcattest. The cleaner solution is to 1) refactor it to install the generated profile (without sourcing it) and 2) install the unit (.t files) somewhere and create a xcattest test case to run the as part of ci_test, this should make they run after install by the github_action_xcat_test.pl, so we're guarded against future regressions

dhilst:
> The fix is fine, the problem is that the test will not ran automatically. Another complication is that this tests the .spec not the profile generated by the spec, so it will not be runnable by xcattest. The cleaner solution is to 1) refactor it to install the generated profile (without sourcing it) and 2) install the unit (.t files) somewhere and create a xcattest test case to run the as part of ci_test, this should make they run after install by the github_action_xcat_test.pl, so we're guarded against future regressions

I think we can merge this, I can wire up the .t file in xcattest later

dhilst (approved):

viniciusferrao:
> The fix is fine, the problem is that the test will not ran automatically. Another complication is that this tests the .spec not the profile generated by the spec, so it will not be runnable by xcattest. The cleaner solution is to 1) refactor it to install the generated profile (without sourcing it) and 2) install the unit (.t files) somewhere and create a xcattest test case to run the as part of ci_test, this should make they run after install by the github_action_xcat_test.pl, so we're guarded against future regressions

This was an issue I've found while reviving the OpenHPC manuals for xCAT. The automated script always failed due to that.

Since there's other issue that is unrelated to the original goal, I think we should and can fix in another PR.

PR #7586, opened 2026-07-13
fix(xcat-server): retry suite 3 on zero RAKP2 HMAC

### LINKED ISSUE #7512: xCAT IPMI does not work anymore after server firmware update
Hello,

Very strange, after the firmware update the xCAT IPMI commands no longer work.

`node209`: **old** firmware
`node207`: **new** firmware
```
# rpower node209 status
node209: on

# rpower node207 status
node207: [ul-dcsm-x01]: Error: ERROR: Incorrect password provided
```
I swear, the password is correct! The xCAT tables haven't changed either. And after firmware roll-back, it worked again.

But the strange thing is that `ipmitool-xcat` has no problems with the new firmware:
```
# ipmitool-xcat -H 10.10.20.209 -I lanplus -U $USER -P $PASSWORD chassis power status
Chassis Power is on

# ipmitool-xcat -H 10.10.20.207 -I lanplus -U $USER -P $PASSWORD chassis power status
Chassis Power is on
```
Well, I saw in the code that `rpower` isn't  using `ipmitool-xcat`, but communicates directly with the IPMI endpoint. Therefore, I suspect the bug lies there. After all, `ipmitool` and `ipmitool-xcat` don't seem to be have any problems with the new firmware.

Do you perhaps have any ideas on how I could narrow it down further?

viniciusferrao:
Fixes #7512.

Reject defective suite-17 RAKP2 responses containing a 32-byte all-zero authentication code and retry once with suite 3. The invalid response is never accepted, and subsequent independent logins continue to start with SHA-256.

### DIFF
--- a/xCAT-test/autotest/testcase/ipmi_unit/cases0
+++ b/xCAT-test/autotest/testcase/ipmi_unit/cases0
@@ -1,0 +1,8 @@
+start:ipmi_rakp2_unit_tests
+description:Run the IPMI RAKP2 Perl unit tests through xcattest
+os:Linux
+label:mn_only,ci_test,ipmi,unit,ipmi_unit
+cmd:prove -I/opt/xcat/lib/perl -I/opt/xcat/lib/perl/xCAT /opt/xcat/share/xcat/tools/autotest/unit/ipmi_*.t
+check:rc==0
+check:output=~All tests successful
+end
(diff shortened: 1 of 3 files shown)


xCAT-test/autotest/testcase/ipmi_unit/cases0
dhilst, line 1: 🔝

dhilst (approved):
The user confirmed that this fixed his problem and the tests are running in the CI, LGTM!

PR #7588, opened 2026-07-14
feat: add configurable NFS root options for statelite images

### LINKED ISSUE #6495: Is it possible to change NFS mount options for the root directory of a statelite image?
I would like to change the attribute caching options on the root directory of one of our statelite images to attempt to find a workaround for some load-related slowdowns we're seeing.  

However, I can't find a way to append additional NFS mount options or change the existing ones that are generated for the node in the files in ```/tftpboot/xcat/xnba/nodes/```.

```
[root@j1x1 xcat]# grep nfs /tftpboot/xcat/xnba/nodes/jfe01
imgargs kernel root=nfs:10.187.1.150:/install/netboot/centos7-statelite/x86_64/login-20191015/rootimg:ro STATEMNT=10.187.1.150:/install/litetree/login XCAT=10.187.1.150:3001 NODE=jfe01 XCATHTTPPORT=80  console=tty0 console=ttyS0,115200n8r MNTOPTS= BOOTIF=01-${netX/machyp}
```

Is there a supported method to change or append additional parameters at the end of the ```root=``` argument?  I've seen the ```mntopts``` field in the ```statelite``` table, but that appears to be for persistent statelite file directories, not the image's root directory.

viniciusferrao:
Add `osimage.nfsrootopts` for NFS-root statelite images. These options are included in the `root=nfs:` kernel parameter used by current EL, Debian, Ubuntu, and SLES images.

The root remains read-only. xCAT always adds `ro` and rejects damaging options: `rw` and `defaults`.

Fixes #6495


dhilst (changes requested):
Strengthen reg_linux_statelite_installation_hierarchy_by_nfs test to cover the new functionality (no ci_test label for this one). Check that if unhappy path prevent the provision of the node or if invalid options are silently pruned, and that valid options survive the happy path.

Also add an integration test to assess the behavior of chdef -t osimage -o <osimage> nfsrootops=<invalid options> followed by an nodeset <node> osimage=<osimage> and include ci_test label on it. Note that this is not a unit test, it must run the commands and assess the output + exit code as other testcases usually do

I think the best UX for invalid options is to fail at chdef time and do not touch the site table

viniciusferrao:
@dhilst solved. Invalid nfsrootopts values now fail at chdef before any database write, preserving the existing osimage value.

dhilst (approved):

PR #7594, opened 2026-07-16
feat: support installation on FIPS-enabled systems

viniciusferrao:
Support xCAT installations on systems with FIPS mode enabled.

- Skip DSA SSH key generation and copy paths.
- Use HMAC-SHA256 for ISC DHCP OMAPI/DDNS under FIPS and require EL8 `dhcp-server >= 12:4.3.6-48`.
- Use P-256 discovery bootstrap keys under FIPS.


dhilst:
FYI: The xcat-dep packages are being rebuilt/tested with FIPS enabled on EL, if you have problems with xcat-dep packages try it out with the devel, instead of latest, and let me know

viniciusferrao:
> FYI: The xcat-dep packages are being rebuilt/tested with FIPS enabled on EL, if you have problems with xcat-dep packages try it out with the devel, instead of latest, and let me know

This one is still draft. I need to retest when packages land.

viniciusferrao:
I think we may need to defer this one to 2.20.

PR #7595, opened 2026-07-16
fix(xcat-core): BuildRequire pod2man/Pod::Man/Pod::Html on SUSE

dhilst:
%build in the pod-using packages runs xpod2man (client, vlan, confluent) or pod2man directly (test, buildkit, OpenStack-baremetal, SoftLayer) to render man pages. On RHEL those come from perl-podlators, pulled in transitively; the openSUSE Leap 15.6 minimal mock chroot has only perl-base, so the build failed with "Can't locate Pod/Man.pm" / "pod2man: command not found".

The earlier %suse_version guard dropped perl-Pod-Html on SUSE (Pod::Html is in core perl there) but nothing then provided Pod::Man / the pod2man binary in the build root. Add a SUSE-only BuildRequires perl(Pod::Man) + perl(Pod::Html) (the full perl package provides both) so xpod2man/pod2man resolve. EL is unaffected (%if 0%{?suse_version}).

Verified: full opensuse-leap-15.6-x86_64 build now produces all 13 signed rpms; repomd.xml signs and verifies against the xCAT Signing Key.


viniciusferrao (changes requested):
buildrpms.pl adds perl-generators to the perl-xCAT mock chroot, but Leap 15.6 does not provide it.

Maybe we should skip it on SUSE?

dhilst:
> buildrpms.pl adds perl-generators to the perl-xCAT mock chroot, but Leap 15.6 does not provide it.
> 
> Maybe we should skip it on SUSE?

This was wip, it inline the dependencies for suse now, since there is no perl-generators for suse

viniciusferrao:
I've converted to draft if it's still WIP, @dhilst.

PR #7596, opened 2026-07-16
fix: avoid sysvinit paths on systemd enabled systems 

### LINKED ISSUE #7593: xCAT still ships /etc/init.d legacy initscripts
We need to get rid of them in systemd based distributions, while not breaking older OS releases.

viniciusferrao:
This PR aims to fix #7593.

- Stop modern RPM and DEB packages from creating `/etc/init.d/xcatd`.
- Retain sysvinit support for legacy systems.
- Replace hard-coded init-script callers with init-agnostic handling.

### DIFF
--- a/perl-xCAT/debian/postrm
+++ b/perl-xCAT/debian/postrm
@@ -21,9 +21,9 @@ set -e
 
 case "$1" in
     upgrade)
-	if [ -x /etc/init.d/xcatd ] && [ -f "/proc/cmdline" ]; then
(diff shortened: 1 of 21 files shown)


dhilst (changes requested):
Create a test case that test that succeed iff `test -d /run/systemd/system && test ! -f /etc/init.d/xcatd` exit code is 0

dhilst (approved):

PR #7599, opened 2026-07-17
fix(xcatprobe): prefer ss for listener checks

### LINKED ISSUE #7093: xcatprobe xcatmn fails on SLE15SP3
'xcatprobe xcatmn' fails on sle15sp3 like:

```
test:~ # xcatprobe xcatmn
[mn]: Checking all xCAT daemons are running...                                                                    [ OK ]
[mn]: Checking xcatd can receive command request...                                                               [FAIL]
[mn]: Attribute 'xcatdport' in 'site' table is set to 3001, but xcatd isn't listening on 3001
[mn]: Attribute 'xcatiport' in 'site' table is set to 3002, but xcatd isn't listening on 3002
=================================== SUMMARY ====================================
[MN]: Checking on MN...                                                                                           [FAIL]
    Checking xcatd can receive command request...                                                                 [FAIL]
        Attribute 'xcatdport' in 'site' table is set to 3001, but xcatd isn't listening on 3001
        Attribute 'xcatiport' in 'site' table is set to 3002, but xcatd isn't listening on 3002
```

This is due to the command 'netstat' no longer being present on sle15 as of SP3.

https://github.com/xcat2/xcat-core/blob/30d107b9cc9e34cc893b832a5617e633c4a4c244/xCAT-probe/subcmds/xcatmn#L364

viniciusferrao:
`xcatprobe xcatmn` uses `netstat` to check whether the xcatd ports are listening. Since `netstat` is not installed by default on SLES 15 SP3, the probe reports ports 3001 and 3002 as unavailable even when xcatd is running normally.

Use `ss` when available and keep `netstat` as a fallback for older systems. Port matching is now exact, and the HTTP listener check uses the same helper.

Fixes #7093

### DIFF
--- a/xCAT-probe/debian/control
+++ b/xCAT-probe/debian/control
@@ -7,6 +7,6 @@ Standards-Version: 3.9.4
 
 Package: xcat-probe
 Architecture: all
-Depends: ${perl:Depends}
+Depends: ${perl:Depends}, iproute2 | net-tools
 Recommends: wget, dnsutils, tftp-hpa, tcpdump
 Description: Provides a toolkit to probe possible issues in xCAT
--- a/xCAT-probe/lib/perl/probe_utils.pm
+++ b/xCAT-probe/lib/perl/probe_utils.pm
@@ -183,6 +183,23 @@ sub _command_available {
     return 0;
 }
 
+sub _capture_command_output {
+    my @command = @_;
+    my $pid = open(my $fh, '-|');
+    return unless defined $pid;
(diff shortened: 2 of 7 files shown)


dhilst (changes requested):
There are some problems:

1. We need to make ss | netstat dependency explicit
2. Add a test case with ci_test label on it so it runs within the CI
3. The functions fail silently on error (for example when the commands are not available)

dhilst (approved):

PR #7601, opened 2026-07-17
docs(sles): clarify installer SSH password

### LINKED ISSUE #7235: 2.16.4: accessing installer via ssh don't work
I'm running 2.16.4 on the MN and tried to debug an installation problem for sles15.3 (#7232)  using debuglevel=2 and the ssh login as described here : https://xcat-docs.readthedocs.io/en/stable/troubleshooting/os_installation/ssh_enable.html?highlight=sles%20debug%20installation

The root entry in passwd table exist:
[root@vlxmn01 ~]# tabdump passwd
#key,username,password,cryptmethod,authdomain,comments,disable
"omapi","xcat_key","YYYYYYYYYYYYYYYYYYYYYYYYYYYYYY",,,,
"system","root","ZZZZZZZZZZZZZZZZZZZZZZ",,,,

and the installation stops as described in the documentation listed above, but the password doesn't work. 
The password inside the passwd - table has been generated with help of keepass, contains upper/lower, digits and special characters (lenght 15)
What is strange that the installation process indicates a BAD password (see attached png) and I can't login with the password stored in the passwd table.
sles-debugLevel2-installation
 
I also changed the password in the autoyast file /install/autoinstall/<nodeName> in the user_password - attribute manually using openssl passwd -1 for encryptions with
<encrypted config:type="boolean">true</encrypted>
and in cleartext with 
<encrypted config:type="boolean">false</encrypted>
but I still can't login to the node via ssh -X root@<nodeName>.mps.mpg.de
The node is pingable and ssh seems to be operational as I get an login prompt and an error message form sshd after 4 failed login attempts.

I don't have an idea which password is used for session. Any advice would extremely helpful, as I'm stuggeded with my planned update related to #7232. Many thanks in advance.

viniciusferrao:
Clarify that SSH access to the SLES installer uses the temporary password
`cluster`, not the installed system root password from the xCAT `passwd` table.

Also document how to set a different temporary password and warn that it will
be visible in the boot configuration and kernel command line.

Fixes #7235

### DIFF
--- a/docs/source/troubleshooting/os_installation/ssh_enable.rst
+++ b/docs/source/troubleshooting/os_installation/ssh_enable.rst
@@ -15,7 +15,24 @@ When ssh access to the installer is enabled, the admin can login into the instal
     ***  login using 'ssh -X root@<node>'  ***
     ***  run 'yast' to start the installation  ***
 
-   Just as the message above suggests, the admin can open 2 sessions and run ``ssh -X root@<node>`` with the configured system password in the ``passwd`` table to login into the installer, then run ``yast`` to continue installation in one session and inspect the installation process in the installer in the other session.
+   xCAT configures the SLES installer with the temporary root password ``cluster``.
+   This password is used only by the installer; it is not the installed system's
+   root password from the ``passwd`` table.
+
+   To use a different temporary installer password, add the SUSE SSH parameters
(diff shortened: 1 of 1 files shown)


viniciusferrao:
Close/Reopen to retrigger CI.

dhilst (approved):

PR #7623, opened 2026-07-21
fix(build): describe git version from lightweight tags too

viniciusferrao:
A `USEGITVER=1` build derives its version with `git describe`, which only considers annotated tags. xCAT tags releases with lightweight tags, so without a reachable annotated tag `git describe` fails and `VER` comes out empty. Use `git describe --tags`.

Recovered from the unmerged lenovobuild branch (25248d19). The commit's `.post`→`.POST` rename and its `makegenesisbuilderrpm` hunk are omitted (unrelated, and the latter targets a script no longer on master).

### DIFF
--- a/buildcore.sh
+++ b/buildcore.sh
@@ -195,7 +195,7 @@ fi
 
 function setversionvars {
     if [ ! -z "$USEGITVER" ]; then
-        VER=`git describe`
+        VER=`git describe --tags`
         VER=${VER/-/.post}
         VER=${VER/-/.}
     else


dhilst (changes requested):
This has to be double checked

Here are some bits I can remember about how git state affects the builds

1. We are using buildrpms.pl so changing builcore is inocuous
2. Git revision goes to Version.pm from Gitinfo. The timestamp was removed in order to make the build deterministic, i.e., the build CANNOT depend on the clocktime to be deterministic
3. The XXXX in snapXXXX comes from git commit timestamp, again, to be deterministic. There is a caveat here that our internal CI is appending the build number in order to make the update work, otherwise dnf update P would not update P

viniciusferrao:
> This has to be double checked
> 
> Here are some bits I can remember about how git state affects the builds
> 
> 1. We are using buildrpms.pl so changing builcore is inocuous
> 2. Git revision goes to Version.pm from Gitinfo. The timestamp was removed in order to make the build deterministic, i.e., the build CANNOT depend on the clocktime to be deterministic
> 3. The XXXX in snapXXXX comes from git commit timestamp, again, to be deterministic. There is a caveat here that our internal CI is appending the build number in order to make the update work, otherwise dnf update P would not update P

Hmm this may have superseeded? If there's nothing to salvage from here we can close this as rejected.

viniciusferrao:
@dhilst to be honest, `buildrpms.pl` is a separate deterministic path. This PR only repairs the explicit `buildcore.sh` when using it with USEGITVER=1 mode. Can you approve?

dhilst (approved):

PR #7625, opened 2026-07-22
fix(netboot): allow the rootfs image to be downloaded over HTTPS

viniciusferrao:
The diskless/statelite xcatroot scripts only recognize an `http://` rootfs image URL; an `https://` imgurl falls through to the "unsupported" path even though the download already uses wget/curl, which handle TLS. Accept `xhttps` alongside `xhttp` in the protocol test, across every current xcatroot variant (rh dracut/dracut_033/dracut_047/dracut_105 stateless and statelite, plus fedora, sles and ubuntu).

Recovered from the unmerged lenovobuild branch (751bce24, extended to the dracut generations and distros that postdate it).


viniciusferrao:
Should be considered as a feature this one.

angelocr (approved):

dhilst (approved):

PR #7627, opened 2026-07-22
refactor(ddns): centralize Net::DNS version policy

viniciusferrao:
Replace the four direct Net::DNS 1.36 comparisons with one shared policy backed by `xCAT::Utils->version_cmp`. This preserves behavior for released Net::DNS versions while making the signing, keyfile, and reconciliation paths use one dotted-version decision.

This should land before #7594, which can then be rebased onto the shared policy.


dhilst (approved):
Nice

PR #7629, opened 2026-07-22
fix(template): fall back to $::XCATROOT when XCATROOT is not in %ENV

viniciusferrao:
`envvar()` resolves a template variable purely from `%ENV`, so a template referencing `$XCATROOT` expands to empty whenever `XCATROOT` is not in the process environment. `$::XCATROOT` is set as a package global at module load and is reliably available, so use it for that one variable before falling back to `%ENV`.

Recovered from the unmerged lenovobuild branch (da72c6b8).

### DIFF
--- a/xCAT-server/lib/perl/xCAT/Template.pm
+++ b/xCAT-server/lib/perl/xCAT/Template.pm
@@ -1617,6 +1617,10 @@ sub envvar
         $envvar =~ s/^\$//;
     }
 
+    # $XCATROOT is reliably set as a package global even when it is absent from
+    # %ENV, so a template referencing it does not expand to undef.
+    if ($envvar eq 'XCATROOT') { return $::XCATROOT; }
+
     return ($ENV{$envvar});
 }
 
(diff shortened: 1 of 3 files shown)


xCAT-server/lib/perl/xCAT/Template.pm
```diff
@@ -1617,6 +1617,10 @@ sub envvar
         $envvar =~ s/^\$//;
     }
 
+    # $XCATROOT is reliably set as a package global even when it is absent from
+    # %ENV, so a template referencing it does not expand to undef.
+    if ($envvar eq 'XCATROOT') { return $::XCATROOT; }
```
dhilst: I think it would be better

return $::XCATROOT if $envvar eq 'XCATROOT' && ! exists $ENV{XCATROOT}

And a unit test for happy and sad path

dhilst (changes requested):

dhilst (approved):

PR #7630, opened 2026-07-22
fix(networks): make IPv6 route filtering in donets() robust

viniciusferrao:
When `donets()` scans the host route table to seed the networks table, it mis-classifies several route lines as networks: link-local routes with a prefix other than the exact `fe80::/64`, `ip -6 route` `nexthop` continuation lines, and the `default` route. Broaden the filter to exclude any `fe80::` prefix, nexthop lines, and default routes.

Recovered from the unmerged lenovobuild branch (8b79cf85 and 7094ba0c, bundled as both edit the same filter).


dhilst (changes requested):
Add unittest for dotnets

dhilst (approved):

PR #7632, opened 2026-07-22
fix(nodech): support comma-delimited values in the ,= and ^= operators

viniciusferrao:
`nodech`'s `,=` (append) and `^=` (remove) operators treated the right-hand side as a single opaque value, so a comma-delimited list was mishandled: `,=` appended the literal `"a,b"` (and compared duplicates against the combined string), and `^=` tried to strip the combined `"a,b"` and left the individual groups in place. Split the value on commas in both operators and add/remove each element individually.

Recovered from the unmerged lenovobuild branch (086b0c0d and 95c78b33).


dhilst (changes requested):
Add unit test for nodech asserting happy and sad paths

dhilst (approved):

PR #7640, opened 2026-07-23
refactor(openbmc): reuse retry scheduler for 503 responses

viniciusferrao:
Reuse the existing `retry_after` helper for OpenBMC HTTP 503 retries instead of maintaining a second local scheduler.

This preserves the current three retries, `_RESPONSE` to `_REQUEST` transition, three-second delay, and terminal 503 handling.

### DIFF
--- a/xCAT-test/autotest/testcase/openbmc_unit/cases0
+++ b/xCAT-test/autotest/testcase/openbmc_unit/cases0
@@ -1,0 +1,8 @@
+start:openbmc_503_retry_unit_tests
+description:Run the OpenBMC 503 retry Perl unit tests through xcattest
+os:Linux
+label:mn_only,ci_test,openbmc,unit,openbmc_unit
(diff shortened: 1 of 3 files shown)


xCAT-test/autotest/testcase/openbmc_unit/cases0
dhilst, line 1: We already have code for that in the github_action_xcat_test.pl, we can remove this one

dhilst (approved):

dhilst (changes requested):
Remove the autotest/testcase/openbmc_unit

xCAT-test/autotest/testcase/openbmc_unit/cases0
viniciusferrao: @dhilst, we should not remove this wrapper. It adds the OpenBMC test to the `ci_test` list; without it, the `.t` file is skipped. `dhcp_unit` and `ipmi_unit` work the same way.

dhilst (approved):

PR #7651, opened 2026-07-23
fix(ddns): retry makedns/nsupdate on transient SERVFAIL

dhilst:
ddns.pm's DNS update send has a 3-try retry loop, but it only retries on NOTAUTH -- any other non-success reply, including SERVFAIL, falls through to a hard failure with no retry.

named routinely replies SERVFAIL transiently right after a zone (re)load, before the zone is ready to accept dynamic updates; the update succeeds a second or two later. So makedns fails on a recoverable condition. It surfaces as "Failure encountered updating <zone> with entry '', SERVFAIL" (the empty \$entry is just the loop variable's state at the batch flush -- cosmetic). This reliably breaks real flat-cluster provisioning on el9/el10, where the test case's makedns -n runs right after named reloads the zone with the new compute node. The code already had a 'sleep 1' with a comment acknowledging this timing race, but the retry loop never covered SERVFAIL.

Retry SERVFAIL with a short pause (mirroring the NOTAUTH retry) in both copies of the send loop.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ddns.pm
+++ b/xCAT-server/lib/xcat/plugins/ddns.pm
@@ -1576,6 +1576,34 @@ sub update_namedconf {
     }
 }
 
+# Send a signed dynamic DNS update, retrying transient rejections. Right after a zone (re)load named
+# can reply NOTAUTH, or SERVFAIL before the zone is ready to accept dynamic updates; both are
+# recoverable, so retry a few times (pausing on SERVFAIL). Returns 0 only when the update was
+# accepted (NOERROR) and 1 on a real or persistent failure -- so callers must NOT report success when
+# this returns non-zero. (The previous inline loop fell through to success after exhausting its
+# retries on NOTAUTH/SERVFAIL, so makedns claimed success even when the update was never accepted.)
+sub send_ddns_update {
+    my ($ctx, $resolver, $update, $zone, $entry) = @_;
+
+    for my $attempt (1 .. 3) {
+        ddns_sign_update($ctx, $update);
+        my $reply = $resolver->send($update);
+        if (!$reply) {
+            xCAT::SvrUtils::sendmsg([ 1, "No reply received when sending DNS update to zone $zone" ], $callback);
+            return 1;
+        }
+        my $rcode = $reply->header->rcode;
+        return 0 if $rcode eq 'NOERROR';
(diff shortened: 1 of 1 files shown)


xCAT-server/lib/xcat/plugins/ddns.pm
```diff
@@ -1691,6 +1691,13 @@ sub add_or_delete_records {
                             if ($reply->header->rcode eq 'NOTAUTH') {
                                 next;
                             }
+                            # named can transiently reply SERVFAIL right after a zone (re)load, before it
+                            # is ready to accept dynamic updates -- pause briefly and retry instead of
+                            # failing the whole makedns run on a recoverable condition.
+                            if ($reply->header->rcode eq 'SERVFAIL') {
```
viniciusferrao: On attempt 3 this `next` exits the loop without calling or returning failure. The function then reports both "Completed updating DNS records." and "DNS setup is completed", although the update was never accepted.

A solution would be to loop on it and fail if it's actually a failure, like this:

```perl
sub send_ddns_update {
    my ($ctx, $resolver, $update, $zone, $entry) = @_;

    for my $attempt (1 .. 3) {
        ddns_sign_update($ctx, $update);
        my $reply = $resolver->send($update);

        if (!$reply) {
            xCAT::SvrUtils::sendmsg(
                [1, "No reply received when sending DNS update to zone $zone"],
                $callback
            );
            return 1;
        }

        my $rcode = $reply->header->rcode;
        return 0 if $rcode eq 'NOERROR';

        if (($rcode eq 'NOTAUTH' || $rcode eq 'SERVFAIL') && $attempt < 3) {
            sleep 2 if $rcode eq 'SERVFAIL';
            next;
        }

        xCAT::SvrUtils::sendmsg(
            [1, "Failure encountered updating $zone with entry '$entry', error was $rcode. See more details in system log."],
            $callback
        );
        return 1;
    }
}
```



Then use `return 1 if send_ddns_update(...);` at both batch-send sites.

viniciusferrao (changes requested):
The transient-SERVFAIL diagnosis is right, but the current retry path turns three consecutive SERVFAIL replies into a successful-looking makedns run. Please keep a terminal error path and add a mocked-resolver regression test for transient and persistent SERVFAIL.

viniciusferrao (approved):
The terminal error path now correctly prevents false completion after persistent DNS update failures. Approved.

PR #7652, opened 2026-07-23
fix(ddns): retry makedns/nsupdate on transient SERVFAIL (2.18)

dhilst:
ddns.pm's DNS update send has a 3-try retry loop, but it only retries on NOTAUTH -- any other non-success reply, including SERVFAIL, falls through to a hard failure with no retry.

named routinely replies SERVFAIL transiently right after a zone (re)load, before the zone is ready to accept dynamic updates; the update succeeds a second or two later. So makedns fails on a recoverable condition. It surfaces as "Failure encountered updating <zone> with entry '', SERVFAIL" (the empty $entry is just the loop variable's state at the batch flush -- cosmetic). This reliably breaks real flat-cluster provisioning on el9/el10, where the test case's makedns -n runs right after named reloads the zone with the new compute node. The code already had a 'sleep 1' with a comment acknowledging this timing race, but the retry loop never covered SERVFAIL.

Retry SERVFAIL with a short pause (mirroring the NOTAUTH retry) in both copies of the send loop.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ddns.pm
+++ b/xCAT-server/lib/xcat/plugins/ddns.pm
@@ -1576,6 +1576,34 @@ sub update_namedconf {
     }
 }
 
+# Send a signed dynamic DNS update, retrying transient rejections. Right after a zone (re)load named
+# can reply NOTAUTH, or SERVFAIL before the zone is ready to accept dynamic updates; both are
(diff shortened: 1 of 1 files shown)


dhilst:
Closed in favor of the auto-backport

PR #7656, opened 2026-07-23
fix(bmcsetup): enable BMC setup for Lenovo SR635/SR655

viniciusferrao:
The original Lenovo SR635/SR655 (IPMI manufacturer 19046, product 1287) use an AMI "TSM" BMC reached over a LAN-over-USB interface. It needs OEM raw commands to bring the interface up, select the LAN channel, and restart TSM networking once configured. This adds a dedicated vendor branch for it, plus a udev rule and helper that bring up the USB NIC (169.254.0.120) so in-band ipmitool can reach the TSM; the user-slot auto-scan is skipped for these machines (fixed slot).

Recovered from the unmerged lenovobuild branch (922dfe18, 42357f9d, 02da8a6b, 765e2ad3, 408cf2b3), adapted to master's bmcsetup structure.

### DIFF
--- a/xCAT-genesis-scripts/etc/udev/rules.d/99-tsm.rules
+++ b/xCAT-genesis-scripts/etc/udev/rules.d/99-tsm.rules
@@ -1,0 +1,1 @@
+SUBSYSTEM=="net", ATTRS{idVendor}=="046b", ATTRS{idProduct}=="ffb0", RUN+="/sbin/setuptsmnic"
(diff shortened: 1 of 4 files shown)


dhilst (approved):
I guess we want this in 2.18.2 too

viniciusferrao:
> I guess we want this in 2.18.2 too

I didn't considered to 2.18.2 since it's hardware enablement and not a bug fix.

PR #7665, opened 2026-07-23
fix(genesis): find the boot NIC over InfiniBand (IPoIB)

viniciusferrao:
When a node boots from an InfiniBand interface, doxcat could not match `BOOTIF` to the netdev, so boot-NIC detection failed. Load `ib_ipoib` and, if the normal Ethernet lookup finds nothing, fall back to matching the InfiniBand device by its (shorter) IPoIB address.

`modprobe ib_ipoib` is harmless where there is no IB hardware (the module simply loads no devices), and the IB lookup only runs when the Ethernet lookup fails — so non-IB nodes are unaffected.

Not lab-validated (no InfiniBand fabric available).

Recovered from the unmerged lenovobuild branch (aeeebd5f).

### DIFF
--- a/xCAT-genesis-builder/xCAT-genesis-base.spec
+++ b/xCAT-genesis-builder/xCAT-genesis-base.spec
@@ -51,6 +51,8 @@ BuildRequires: ipmitool
 BuildRequires: iproute
 BuildRequires: kexec-tools
 BuildRequires: kernel-core
+BuildRequires: kernel-modules
+BuildRequires: kernel-modules-extra
 BuildRequires: lldpad
 BuildRequires: lvm2
 BuildRequires: mdadm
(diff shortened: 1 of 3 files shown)


dhilst (changes requested):
The doxcat logic looks fine, but it's a no-op as-is: the genesis image doesn't ship the InfiniBand stack. xCAT-genesis-base.spec BuildRequires only kernel-core, and on EL the IB modules live in kernel-modules — so ib_ipoib/ib_core/mlx5_ib aren't in the image. I checked three built genesis RPMs (el8/9/10) and none contain any /infiniband/ modules. On a booted node modprobe ib_ipoib would fail and the fallback never matches.

Two ways to fix it:

(1) Add BuildRequires: kernel-modules (+ kernel-modules-extra) — the dracut installkernel() sweeps everything in the chroot's modules.dep, so all IB drivers land automatically.
- Pros: generic genesis image, covers any HCA out of the box.
- Cons: bloats the initramfs for all nodes.

(2) Handpick with instmods, e.g. instmods ib_ipoib mlx5_ib ib_umad — BuildRequire the package but pull only the needed drivers.
- Pros: saves space.
- Cons: less generic — new/other HCAs need explicit additions.

Either way, rebuild/republish for each target (el8/9/10, ppc64le). Since there's no IB fabric in CI, this'll need a manual check on real hardware.

dhilst (approved):

PR #7691, opened 2026-07-25
fix(xnba): warn instead of failing when pxelinux.0 is unavailable

viniciusferrao:
`nodeset` aborts with "Unable to find pxelinux.0 from syslinux" when neither the copy under the tftp directory nor the one under `$XCATROOT` exists. Only legacy BIOS clients chain pxelinux, and not every distribution still ships syslinux, so a UEFI-only environment cannot run `nodeset` at all.

Copy pxelinux.0 when it is available, and warn rather than fail when it is not.

Recovered from the unmerged lenovobuild branch (4f0a5106), which skipped silently; this warns so the missing loader is still reported.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/xnba.pm
+++ b/xCAT-server/lib/xcat/plugins/xnba.pm
@@ -762,16 +762,15 @@ sub process_request {
 
     #back to normal business
     if (!-r "$globaltftpdir/xcat/pxelinux.0") {
-        unless (-r $::XCATROOT . "/share/xcat/netboot/syslinux/pxelinux.0") {
-            $::XNBA_callback->({ error => [ "Unable to find pxelinux.0 at " . $::XCATROOT . "/share/xcat/netboot/syslinux/pxelinux.0" ], errorcode => [1] });
-            return;
+        if (-r $::XCATROOT . "/share/xcat/netboot/syslinux/pxelinux.0") {
+            copy($::XCATROOT . "/share/xcat/netboot/syslinux/pxelinux.0", "$globaltftpdir/xcat/pxelinux.0");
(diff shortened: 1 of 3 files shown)


xCAT-server/lib/xcat/plugins/xnba.pm
```diff
@@ -762,16 +762,15 @@ sub process_request {
...
         }
-        copy($::XCATROOT . "/share/xcat/netboot/syslinux/pxelinux.0", "$globaltftpdir/xcat/pxelinux.0");
-        chmod(0644, "$globaltftpdir/xcat/pxelinux.0");
     }
     unless (-r "$globaltftpdir/xcat/pxelinux.0") {
-        $::XNBA_callback->({ error => ["Unable to find pxelinux.0 from syslinux"], errorcode => [1] });
-        return;
+        # only legacy BIOS clients chain pxelinux, and not every distribution
+        # still ships syslinux, so warn instead of failing the whole request
+        $::XNBA_callback->({ warning => ["Unable to find pxelinux.0 from syslinux, legacy BIOS nodes that chain pxelinux will not boot until it is available"] });
```
dhilst: I guess this will only trigger when we remove syslinux-xcat from the dependencies!? I'm a bit afraid of this flooding over all nodeset call in that case, I'm approving for now because syslinux-xcat dep should guard it. Fell free to double check it or merge.

I spend some time trying to figure out if this would not trigger every nodeset, but I'm not convinced

dhilst (dismissed):

dhilst (dismissed):

dhilst (approved):

dhilst (approved):

PR #7701, opened 2026-07-30
build: Fixes to get 2.19 green on CI/CD 

dhilst:
- Improve buildrpms.pl
  - handle mock cancelation
  - init mock chroot before building to prevent corrupted build to break the next builds

### DIFF
--- a/buildrpms.pl
+++ b/buildrpms.pl
@@ -491,9 +491,10 @@ sub buildspkgs {
     # initialised root for both the srpm and the binary rebuild within this run. We reach
     # here only when actually building (past the diskcache skip), so flat-disk reuse across
     # runs is preserved -- we just guarantee a known-good starting point each time.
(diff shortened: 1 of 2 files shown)


viniciusferrao (changes requested):
@dhilst, I found the same issues here:

- Keep `$blk` alive until process exit; it currently leaves scope before workers start and releases the flock.
- Propagate `sh_retry()` and ForkManager child failures; the parent can currently sign a partial repository and exit 0.
- Limit `sweep_mock_mounts()` to tracked chroots; it currently unmounts every `/var/lib/mock` mount and can break unrelated builds.

viniciusferrao (changes requested):
@dhilst, this was totally on Codex side:

Check `$exit_signal` and `$core_dump` in `run_on_finish`; a worker killed by SIGKILL or OOM can report `$exit_code == 0`, so the parent can still sign a partial repository.

viniciusferrao (approved):
Looks good.

PR #7712, opened 2026-08-07
fix(build-ubunturepo): scope the build lock + scratch under an option…

dhilst:
### Problem

`build-ubunturepo` guarded builds with a single **host-global, fail-fast** lock at
`/var/lock/xcatbld.lock` (`flock -n`). Two builds on one host — e.g. the **devel** and
**stable** Ubuntu CD lanes on the same build host — therefore collided on that one lock,
and the loser exited `1` with `ERROR: Can't get lock ...`, **failing the whole pipeline**,
even though each lane builds from its **own checkout** into its **own `DEST`** and they
share no state.

### Root cause

The lock path encodes nothing about the build's identity and `flock -n` fails immediately
instead of waiting, so *any* second Ubuntu build on the host loses. But
`build-ubunturepo` builds its packages **in-place in its own source checkout** — it
rewrites `debian/changelog` and `debian/control`, drops `*.orig.tar.gz` at the checkout
root and runs `dpkg-buildpackage` inside the package dirs — so the resource two concurrent
builds actually contend for is **the checkout**, not the host.

### Fix

Key the lock on the **checkout path** (`$curdir`):

```sh
lock_id_for() { printf '%s' "$1" | md5sum | cut -c1-12; }
LOCKFILE="/var/lock/xcatbld-$(lock_id_for "$curdir").lock"
exec 8>"$LOCKFILE"
if ! flock -n 8; then
    echo "ERROR: Can't get lock $LOCKFILE for checkout $curdir.  Another build is already using this checkout.  Exiting...."
    exit 1
fi
```

- Builds of the same checkout still fail-fast (they would corrupt each other in-place).
- Builds of distinct checkouts get distinct locks and run in parallel — which is
exactly the devel + stable CD case.
- The lock stays on the local /var/lock (reliable flock, unlike the NFS/virtiofs checkout),
and the source tree is left byte-pristine.

### DIFF
--- a/build-ubunturepo
+++ b/build-ubunturepo
@@ -35,6 +35,15 @@
 #
 #               DEST=<directory> - provide a directory to contains the build result
 #
+# Running builds in parallel on one host:  the build lock is scoped to the source checkout (see
+#                                 the "build-lock" block below), so two builds from DIFFERENT
+#                                 checkouts -- e.g. the devel and stable Ubuntu CD lanes -- run
+#                                 concurrently, while two builds of the SAME checkout still fail-fast
+#                                 (they build in-place and would corrupt each other).  For parallel
+#                                 builds give each a separate checkout and a separate output tree
+#                                 (a distinct DEST).  GPG_HOME may be SHARED between parallel builds
+#                                 -- it is used read-only for signing.
+#
 # For the dependency packages 1. All the xcat dependency deb packages should be uploaded to
 #                                "pokgsa/projects/x/xcat/build/ubuntu/xcat-dep/debs/" on GSA
 #                             2. run ./build-ubunturepo -d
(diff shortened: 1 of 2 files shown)


viniciusferrao (dismissed):
The change is correct for the stated CI setup, where concurrent builds use separate source checkouts, `DEST`, `GPG_HOME`, and `BUILDDIR`. I found no blocking correctness issue.

### Non-blocking: test and document `BUILDDIR`

The current CI does not invoke `build-ubunturepo` with `BUILDDIR`, so the new locking path is not exercised.

Please add a small test covering:

- the same `BUILDDIR` rejects a second process;
- different `BUILDDIR`s can acquire their locks concurrently;
- an unset `BUILDDIR` retains the global lock.

Also add `BUILDDIR` to the script usage comments and document that parallel builds require separate source checkouts, output trees, and GPG homes.

dhilst:
> The change is correct for the stated CI setup, where concurrent builds use separate source checkouts, `DEST`, `GPG_HOME`, and `BUILDDIR`. I found no blocking correctness issue.
> ### Non-blocking: test and document `BUILDDIR`
> 
> The current CI does not invoke `build-ubunturepo` with `BUILDDIR`, so the new locking path is not exercised.
> 
> Please add a small test covering:
> 
>     * the same `BUILDDIR` rejects a second process;
> 
>     * different `BUILDDIR`s can acquire their locks concurrently;
> 
>     * an unset `BUILDDIR` retains the global lock.
> 
> 
> Also add `BUILDDIR` to the script usage comments and document that parallel builds require separate source checkouts, output trees, and GPG homes.

TBH BUILDDIR is unnecessary, the CI has 2 pipelines, each pipeline has its own checkout dir, the build calls dpkg-buildpackage which write files relative to the debian folders, so it is safe to run they in parallel as long as they run in distinct dirs (which CI does as noted previously). So I moved the lock file to be parametrized by the hash of the checked out dir.

viniciusferrao (approved):

PR #7716, opened 2026-08-09
fix(syncfiles): name root explicitly on the xdcp subrequest

viniciusferrao:
`syncfiles` builds an xdcp subrequest with no username, so the sync runs as whatever identity xcatd happens to have: xdcp takes its user from `-l`, then `DSH_FROM_USERID`, and otherwise leaves it unset.

2c9bdf848 removed the `forceroot` flag and replaced it by passing the real username through. `updatenode` was updated; `syncfiles` was not. The request comes from the node, so there is no calling user to carry, and root is what it runs as.

Recovered from lenovobuild (4cb5a894d + d46eec274); the first passed a string, which does not survive the `->[0]` the consumers use.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/syncfiles.pm
+++ b/xCAT-server/lib/xcat/plugins/syncfiles.pm
@@ -129,7 +129,7 @@ sub syncfiles {
             push @$args, "$::RCP";
         }
         my $env = ["DSH_RSYNC_FILE=$synclistfile"];
-        $subreq->({ command => ['xdcp'], node => [$node], arg => $args, env => $env }, $callback);
+        $subreq->({ command => ['xdcp'], username => ['root'], node => [$node], arg => $args, env => $env }, $callback);
     }
     return 1;
 }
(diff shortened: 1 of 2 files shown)


xCAT-server/lib/xcat/plugins/syncfiles.pm
```diff
@@ -129,7 +129,7 @@ sub syncfiles {
             push @$args, "$::RCP";
         }
         my $env = ["DSH_RSYNC_FILE=$synclistfile"];
-        $subreq->({ command => ['xdcp'], node => [$node], arg => $args, env => $env }, $callback);
+        $subreq->({ command => ['xdcp'], username => ['root'], node => [$node], arg => $args, env => $env }, $callback);
```
dhilst, line 132: Can't this cause problems with file permissions? I mean, all files end up owned by root in the computing nodes? Or this is already the case?

dhilst (approved):
I double checked it, it is correct, the ownership comes from the file not the login user

PR #7720, opened 2026-08-09
fix(ci): make package validation noninteractive

viniciusferrao:
- make Ubuntu package and dependency installs noninteractive
- keep source-layout unit tests out of the installed-tree syntax scan

The syntax scan was compiling unit tests from `/opt/xcat`, where their source fixtures do not exist. The unit suite still runs from the preserved checkout.

### DIFF
--- a/github_action_xcat_test.pl
+++ b/github_action_xcat_test.pl
@@ -475,6 +475,7 @@ sub check_syntax{
         foreach my $file (@files) {
             next if($file =~ /\/opt\/xcat\/share\/xcat\/netboot\/genesis\//);
             next if($file =~ /\/opt\/xcat\/probe\//);
+            next if($file =~ /\/opt\/xcat\/share\/xcat\/tools\/autotest\/unit\//);
 
             @output = runcmd("file $file");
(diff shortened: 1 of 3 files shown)


github_action_xcat_test.pl
```diff
@@ -475,6 +475,7 @@ sub check_syntax{
         foreach my $file (@files) {
             next if($file =~ /\/opt\/xcat\/share\/xcat\/netboot\/genesis\//);
             next if($file =~ /\/opt\/xcat\/probe\//);
+            next if($file =~ /\/opt\/xcat\/share\/xcat\/tools\/autotest\/unit\//);
```
dhilst, line 478: Why skip these?

dhilst (approved):

github_action_xcat_test.pl
viniciusferrao: These are source-tree tests. Their `FindBin` paths expect the repository layout, which is not present under `/opt/xcat`, so compiling the installed copies produces false syntax failures. `run_unit_tests()` already proves the complete directory from the preserved checkout before this installed-tree scan.

PR #7721, opened 2026-08-09
feat(genesis): build images with OpenEmbedded

viniciusferrao:
This PR adds the source and build path for an OpenEmbedded-based Genesis image. It replaces the per-distribution build model with one maintained platform layer.

The image provides:

- the existing xCAT registration, discovery, and action protocol
- systemd and NetworkManager across seven architectures
- Newt and plain consoles backed by the same status model
- signed system extensions for restricted vendor tools
- open hardware providers
- checksums, license records, SPDX SBOMs, and VEX reports

`x86_64` is the primary target and `ppc64le` is the second physical-hardware target. The remaining architectures are supported through QEMU until physical systems are available.

## Distribution

This PR does not publish binary Genesis images or replace the existing `xCAT-genesis-base-*` packages. Official RPM and DEB packaging and repository publication are handled separately through xcat-dep. The package consumer installs the exported image where `mknb` can validate and deploy it.

The existing distribution-based Genesis builder remains available during the transition.

## Compatibility

The new runtime reuses the existing `getdestiny` and `nextdestiny` clients. Their shared implementations were updated to support one-shot requests while preserving the retry and error behavior used by the existing Genesis image.

General DHCP, boot-policy, and server fixes were reviewed separately in #7720, #7722, #7723, and #7724. Service-node certificate delegation is provided by #7738.

## Testing

- 811 Genesis assertions
- clean x86_64 image and signed-extension builds from pinned sources
- BitBake parsing for x86, x86_64, ppc64, ppc64le, armv7hf, aarch64, and riscv64
- BIOS and non-secure UEFI boots
- direct and service-node registration and certificate enrollment
- registered, rejected, and sequential discovery paths
- successful runtime loading of a signed system extension
- fatal action failure without an infinite restart loop
- release export with checksums, licenses, SPDX SBOMs, and VEX reports

Secure Boot and the server-side work required for a complete IPv6-only boot path are outside this PR. Physical x86_64 and ppc64le tests remain release gates.


dhilst (changes requested):
The getdestiny / nextdestiny rewrite changes behavior for the old genesis image. Since the two will live side by side until we retire the old one, it's still calling these scripts in its original mode — and there the rewrite quietly changes what they do: the old scripts, when the management node had nothing new for a node, would return empty and let boot carry on, whereas the new ones just keep retrying instead. The problem this raises is that a legacy node can hang early in boot if the management node ever answers without a destiny and without an error — exactly the case the old code shrugged off. Could we keep the old behavior on that path — retry only on a genuine connection failure or an explicit error, and otherwise return empty — so the old genesis keeps booting the way it always has while both images coexist?

dhilst (approved):

PR #7724, opened 2026-08-11
feat(mknb): install exported Genesis images

viniciusferrao:
Genesis can now be delivered as a prebuilt kernel and compressed initramfs with a `SHA256SUMS` manifest. `mknb` detects this layout, validates the required files, and publishes them under the configured TFTP root.

Invalid or incomplete exports fail without replacing the current boot artifacts.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/mknb.pm
+++ b/xCAT-server/lib/xcat/plugins/mknb.pm
@@ -35,6 +35,144 @@ sub _select_network_addresses {
     return (\%legacy, \%selected);
 }
 
+sub _prebuilt_genesis_requested {
+    my ($directory) = @_;
+    return -e "$directory/initramfs.cpio.gz"
+      || -l "$directory/initramfs.cpio.gz"
+      || -e "$directory/SHA256SUMS"
+      || -l "$directory/SHA256SUMS";
+}
+
+sub _read_prebuilt_genesis_checksums {
+    my ($directory) = @_;
+    my $checksum_file = "$directory/SHA256SUMS";
+    return (undef, "Missing Genesis checksum file: $checksum_file")
+      unless -f $checksum_file && !-l $checksum_file;
+
+    open(my $checksum_fh, '<:raw', $checksum_file)
+      or return (undef, "Unable to read Genesis checksum file: $checksum_file");
+
+    my %expected;
+    while (my $line = <$checksum_fh>) {
+        chomp($line);
+        unless ($line =~ /^([0-9a-f]{64})  ([A-Za-z0-9][A-Za-z0-9._-]*)$/) {
+            close($checksum_fh);
+            return (undef, "Invalid Genesis checksum entry: $line");
+        }
+        my ($digest, $name) = ($1, $2);
+        if (exists($expected{$name})) {
+            close($checksum_fh);
+            return (undef, "Duplicate Genesis checksum entry: $name");
+        }
+        $expected{$name} = $digest;
+    }
+    close($checksum_fh);
+
+    return (\%expected, undef);
+}
+
+sub _sha256_file {
+    my ($path) = @_;
+    return (undef, "Digest::SHA is required to verify Genesis artifacts")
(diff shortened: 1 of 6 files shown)


xCAT-server/lib/xcat/plugins/mknb.pm
```diff
@@ -35,6 +35,144 @@ sub _select_network_addresses {
...
+    }
+    close($checksum_fh);
+
+    return (\%expected, undef);
+}
+
+sub _sha256_file {
+    my ($path) = @_;
+    return (undef, "Digest::SHA is required to verify Genesis artifacts")
+      unless eval { require Digest::SHA; 1 };
```
dhilst: I think this require can be safely hoisted to "use Digest::SHA", double check if it is distributed with perl in all our targets or if we need to update in the genesis package dependencies for all targets: arch {x86_64, ppc64le} * ubuntu-{20,22,24,26}.04, el{8,9,10}

xCAT-server/lib/xcat/plugins/mknb.pm
```diff
@@ -35,6 +35,144 @@ sub _select_network_addresses {
     return (\%legacy, \%selected);
 }
 
+sub _prebuilt_genesis_requested {
+    my ($directory) = @_;
+    return -e "$directory/initramfs.cpio.gz"
+      || -l "$directory/initramfs.cpio.gz"
+      || -e "$directory/SHA256SUMS"
+      || -l "$directory/SHA256SUMS";
+}
```
dhilst, line 104: This seems a little weak for a **pre-built genesis was requested** test, it return true if it is called with any directory with a initramfs.cpio.gz or SHA256SUMS inside it. Maybe we can generate a genesis manifest file and test for it in conjunction with the checksum and the initramfs. **pre-built genesis was requested** must return true if and only if this is directory containing a _"neo"_ genesis prebuilt image 

Also note that _prebuilt_genesis_requested($dir) is not strict about the presence of the checksum but _read_prebuilt_genesis_checksums is, the strictness of theses functions regarding the presence of these files must match

xCAT-test/unit/mknb_exported_genesis.t
```diff
@@ -0,0 +1,221 @@
...
+    xCAT_plugin::mknb::_prebuilt_genesis_requested($export),
+    'an exported initramfs selects the prebuilt path',
+);
+
+my ($installed_initrd, $install_error) =
+  xCAT_plugin::mknb::_install_prebuilt_genesis($export, $tftpdir, 'x86_64');
+is($install_error, undef, 'a valid export installs without an error');
+is(
+    $installed_initrd,
+    "$tftpdir/xcat/genesis.fs.x86_64.gz",
```
dhilst, line 120: I think this file is perfect for the "_prebuilt_genesis_requested" test in conjunction with initramfs and the checksum (instead of disjunction `||`)

dhilst (changes requested):

xCAT-server/lib/xcat/plugins/mknb.pm
viniciusferrao: Digest::SHA is now loaded at module scope. I added the dependency to both RPM branches, including s390x, and confirmed the Debian dependency on libdigest-sha-perl. The packaging test covers all three cases.

xCAT-server/lib/xcat/plugins/mknb.pm
viniciusferrao: Fixed. A prebuilt Genesis image is now selected only when xcat-genesis.manifest, kernel, initramfs.cpio.gz, and SHA256SUMS are all present.

xCAT-test/unit/mknb_exported_genesis.t
viniciusferrao: Fixed. The test now requires the manifest, kernel, initramfs, and checksum file together. It also covers unmarked directories, partial exports, and incomplete marked exports.

dhilst (approved):

Issue #7710, opened 2026-08-04
Ubuntu >= 22.04: xcat metapackage guarantees isc-dhcp-server while DHCP backend auto-selection defaults to Kea


viniciusferrao:
On Ubuntu 22.04+ the DHCP backend auto-selection introduced with the Kea backend (#7517) defaults to `kea`, but the `xcat` metapackage's dependency line still prefers ISC:

```
Depends: ..., isc-dhcp-server | kea, ...
Recommends: bind9, net-tools, nmap, kea, ...
```

apt satisfies an alternation with its first member, so the package dependency guarantees `isc-dhcp-server` — a backend that `makedhcp` will not select on these releases — while `kea` is only pulled in indirectly through `Recommends`.

## Environment

- Ubuntu 24.04 (noble), amd64
- `xcat` 2.18.0-snap202606250818 from `http://xcat.org/files/xcat/repos/apt/latest/xcat-core jammy main`

## Reproduction

`apt-get -s install xcat` in a clean Ubuntu 24.04 environment:

1. **Default apt configuration (`APT::Install-Recommends "true"`):** both DHCP servers are installed — `isc-dhcp-server` via the `Depends` alternation and `kea` (with `kea-dhcp4-server` 2.4.1) via `Recommends`. `makedhcp` works and configures Kea.

2. **`--no-install-recommends`:** only `isc-dhcp-server` is installed. `makedhcp` then fails hard:

   ```
   The selected DHCP backend 'kea' is not available on this system.
   ```

   `xCAT::DHCP::Backend->default_backend` selects `kea` for Ubuntu >= 22.04, and `dhcp.pm` calls `new_backend(check_available => 1)` with no fallback, so a minimal install of the metapackage produces a management node whose `makedhcp` is broken out of the box, with no hint that installing `kea` (or setting `site.dhcpbackend=isc`) would resolve it.

## Relevant code

- `xCAT/debian/control` — `Depends: ..., isc-dhcp-server | kea, ...`
- `perl-xCAT/xCAT/DHCP/Backend.pm` — `default_backend` returns `kea` for Ubuntu >= 22.04
- `xCAT-server/lib/xcat/plugins/dhcp.pm` — `new_backend(check_available => 1)`

[...]

Issue #7454, opened 2024-07-15
netplan configuration for Ubuntu in configeth script


abhishek-sa1:
Is there any plan for supporting Network configuration using netplan for Ubuntu in `configeth` script in next release of xCAT?

@Obihoernchen

samveen:
@abhishek-sa1  Netplan.io is a configuration generator that in turn generates Network Manager configuration (or for `systemd-networkd`), and doesn't actually do any network management itself. 

As xCAT already supports NetworkManager directly, adding an extra layer of abstraction is a waste of effort and possibly an additional point of failure, in case netplan's behaviour changes.

viniciusferrao:
@abhishek-sa1 the feature will land on 2.19 release which is due for 2 weeks from today.

PR #7725, opened 2026-08-13
fix(xcat-core): Ubuntu Subiquity diskful install fails when noderes.installnic is unset

dhilst:
Problem — The Ubuntu diskful (Subiquity) install fails on any node without noderes.installnic. compute.subiquity.tmpl resolves #TABLE:noderes:$NODE:installnic#, and Template.pm's tabdb raises a fatal "Unable to find requested field <installnic>" → "Failed to generate xnba configurations", so the CN never enters the installer. EL/SLES pass because their statefull templates never reference installnic (gen_net_boot_params defaults to the boot MAC).
Fix — Resolve installnic with the non-fatal #TABLEBLANKOKAY# token and treat empty installnic as match-by-MAC (EL-equivalent). Diskless path untouched; nodes that set installnic unaffected.


dhilst:
This is under test in the CI

viniciusferrao:
@dhilst ~~BIND9 on 26.04? It has been deprecated already and EL10 does not have it anymore. Any specific reason?~~

I need to sleep more and rest. I mixed DHCP with DNS in my head. Sorry.

viniciusferrao (changes requested):
Hi @dhilst 

Some issues:

1. The `installnic` fix itself looks good. Using `#TABLEBLANKOKAY#` and treating an empty value as match-by-MAC fixes the reported issue without changing configured nodes.

The problem is that this PR has grown far beyond that scope. It now mixes DHCP, NTP, packaging, Netplan, `xcatd`, and several other Ubuntu provisioning fixes. These should be split into separate PRs.

2. Please change the unconditional Magic SysRq reboot, this looks like a hack:

```sh
sleep 90
echo b >/proc/sysrq-trigger
```

This can reboot the machine without syncing or unmounting filesystems, and the timer may fire while a slow installation is still finishing. The `/dev/tcp` command also ignores failures and does not confirm that `nodeset` actually completed.

Since `toram` is already being added to address the NFS shutdown hang, the hard reboot should not be necessary.

Maybe we should remove it.

3. The new Netplan writer puts every interface under `ethernets:`. An interface such as `eth0.100` needs a proper `vlans:` stanza with `id` and `link`; otherwise it will not be recreated after reboot.

It's a better idea to reserve the existing VLAN path or add proper Netplan VLAN support and test it with `netplan generate`.


4. This pattern is not an exact node-name match:

```perl
\Q$node\E\b.*
```

For example, `compute` also matches `compute-01`. That could query or delete the wrong host block.

The goal shoule be to match the full marker format and add a regression test covering similar node names.

5. The new condition applies to every Ubuntu/Debian provisioning method, including netboot and statelite. It also moves `syncfiles` after the existing postbootscripts, which changes the execution order.

This should be limited to Subiquity diskful installs, and `syncfiles` should still run before postbootscripts that may depend on those files.

My suggestion is to keep this PR limited to the `installnic` fix and its regression test, then submit the other fixes separately.

Thank you.

dhilst:
Superseded by #7755, #7756, #7758, #7759, #7760, #7761, #7762.

| Replacement | What it covers |
|---|---|
| #7755 | `installnic` unset breaks Subiquity xnba generation — the fix you approved, on its own |
| #7756 | Ubuntu 26.04: `bind9` Depends (apt 3's solver drops the Recommends) + genimage codename |
| #7758 | `makedhcp`: backend auto-fallback (#7710) + the `-q` omshell hang + exact node matching |
| #7759 | xcatd never respawns its install monitor |
| #7760 | NTP on Ubuntu MNs: daemon selector, `setupntp` fixes, package dependencies |
| #7761 | The Ubuntu Subiquity diskful install chain |
| #7762 | netplan support in `configeth` (#7454) |

PR #7726, opened 2026-08-15
fix(remoteshell): stop rewriting the administrator's sshd_config

viniciusferrao:
On every deployment `remoteshell` strips the administrator's `X11Forwarding`
and `MaxStartups` lines from `/etc/ssh/sshd_config` and appends its own, so
local settings are gone after each install.

xCAT does not depend on the `MaxStartups` value, so the postscript now leaves
it alone. For `X11Forwarding`, where the running config has an active
`Include <dir>/*.conf` line it writes `<dir>/01-xcat.conf` and leaves
`sshd_config` untouched; otherwise it edits in place as before, so anything
older than OpenSSH 8.2 is unaffected. The directory is read from the file, and
anything it cannot resolve to a single directory falls back to the in-place
edit.

An administrator who sets `X11Forwarding` above the `Include` line still wins.
That takes a deliberate edit, and respecting it is the intent.

#5936 fixed this with a `Match all` block and was reverted in 3c95f2b79:
`MaxStartups` is illegal inside `Match`, so the file stopped parsing. Jarrod
Johnson's diagnosis was right; this takes a different route.

### DIFF
--- a/xCAT-test/unit/remoteshell_sshd_config.t
+++ b/xCAT-test/unit/remoteshell_sshd_config.t
@@ -1,0 +1,454 @@
+#!/usr/bin/env perl
+use strict;
+use warnings;
+
+use File::Copy qw(copy);
+use File::Path qw(make_path);
+use File::Slurper qw(read_text write_text);
+use File::Temp qw(tempdir);
+use FindBin;
+use lib "$FindBin::Bin/../lib";
+use Test::More;
+
+use XCAT::Test::File qw(repo_path);
+
+my $postscripts = repo_path('xCAT/postscripts');
+my $remoteshell = "$postscripts/remoteshell";
+my $sshd_helper = "$postscripts/remoteshell-sshd-config";
+plan skip_all => 'remoteshell postscript not found' unless -r $remoteshell;
+plan skip_all => 'remoteshell sshd helper not found' unless -x $sshd_helper;
+# The postscript is written against GNU sed; sed -i means something else on BSD.
+plan skip_all => 'postscript targets Linux nodes' unless $^O eq 'linux';
+is( system( 'sh', '-n', $sshd_helper ), 0,
+    'the sshd configuration helper has POSIX shell syntax' );
+
+sub run_sshd_helper {
+    my (%opt) = @_;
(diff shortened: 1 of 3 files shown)


xCAT-test/unit/remoteshell_sshd_config.t
```diff
@@ -0,0 +1,309 @@
...
+    return;
+}
+
+sub read_file {
+    my ($path) = @_;
+    return '' unless -e $path;
+    open my $fh, '<', $path or die "open $path: $!";
+    local $/;
+    return <$fh>;
+}
```
dhilst: We can use File::Slurper for read/write_file

xCAT/postscripts/remoteshell
```diff
@@ -54,22 +54,110 @@ if [ "$USEFLOWCONTROL" = "YES" ] || [ "$USEFLOWCONTROL" = "yes" ] || [ "$USEFLOW
...
+    mv -f "$xcat_tmp" "$xcat_dest" 2>/dev/null || { rm -f "$xcat_tmp"; return 1; }
+    return 0
+}
+
+# sshd keeps the first value it finds for a keyword, so a setting ahead of the
+# Include line or in an earlier fragment still wins over ours; log it.
+xcat_warn_if_overridden() {
+    xcat_kw=$1
+    xcat_own="$XCATSSHDDROPINDIR/$2"
+    if sed -n '/^[[:space:]]*[Ii][Nn][Cc][Ll][Uu][Dd][Ee][[:space:]]/q; p' /etc/ssh/sshd_config 2>/dev/null | \
```
dhilst: Using `/I` _ignore-case_ would make it more readable

dhilst (changes requested):

viniciusferrao:
@dhilst `/I` applied.

File::Slurper not added: it is not core, and xCAT-test/unit has no non-core dependencies today, so the suite would fail wherever it is not installed. Kept the small `read_file` helper.

Also added test coverage for the Include detection, which had none before.

Can you look again?

dhilst (approved):

PR #7729, opened 2026-08-17
fix(xcatd): stop a noderange from running a command through the ^ operator

viniciusferrao:
The `^` noderange operator reads node names from a file, and `NodeRange` opened that file with a two-argument `open`. Two-argument `open` honors shell metacharacters in the path, so a noderange like `^"id|"` would run a command. Because xcatd expands noderanges while handling a request, that command ran as root on the management node.

Switching to a three-argument `open` with an explicit read mode fixes it: the value is only ever treated as a filename. The operator still works, so `^/tmp/nodes` reads the file as before.

Recovered from the lenovobuild branch, where the original fix removed the `^` operator outright. This keeps the documented operator and closes the command path instead.


viniciusferrao:
@dhilst I've bypassed the merge rules for this one.

PR #7741, opened 2026-08-21
feat: add `genpdu` pdutype for generic SNMP PDUs (PDU2-MIB)

kcgthb:
Add a new pdutype (`genpdu`) for PDUs implementing the Raritan PDU2-MIB. A single MIB covers the Raritan PX2/PX3/PX4/PXC/SRC/PXO/BCM series, the Server Technology PRO3X/PRO4X series, the Legrand intelligent PDUs, and should work for all PDUs following the PDU2-MIB.

Supports rpower (whole-PDU and per-outlet), rinv and rvitals over SNMP v1, v2c or v3, with credentials read from the pdu table.

Unlike the existing types, sensor units and decimal precision are read from the MIB per sensor rather than hardcoded, so readings are correct across models that report differing precision for the same sensor. Outlet switching capability is probed at connect time, so metered-only models report a single "unsupported" message instead of a per-outlet error.

rspconfig is not supported for `genpdu`.

Tested on Raritan PX4-5851-E7V2 (fw 4.2.10.5-50400, switched), PX3-1901U-N1 and PX3-1901U-N1A6 (fw 4.0.20.5-49038, metered), and PX2-1901U-N1A6 (fw 4.0.20.5-49038, metered).

### DIFF
--- a/xCAT-server/lib/xcat/plugins/pdu.pm
+++ b/xCAT-server/lib/xcat/plugins/pdu.pm
@@ -57,8 +57,8 @@ my $pduhash;
 # pduId is the PDU's link ID and is 1 when linking is not used; BCM2/PMC use 0
 # for the main controller. Linked and BCM2/PMC deployments are out of scope for
 # genpdu, because pduoutlet.pdu is parsed as PDU:outlet and has no field for a
-# link ID. pduCount reports how many units are behind one agent and is 1 on a
-# standalone PDU.
+# link ID. pduCount reports how many units sit behind one agent, 1 when
+# standalone.
 #-------------------------------------------------------
 my $PDU2_PDUCOUNT    = ".1.3.6.1.4.1.13742.6.3.1.0";      #pduCount,             ro, scalar
 my $PDU2_SWITCHOP    = ".1.3.6.1.4.1.13742.6.4.1.2.1.2";  #switchingOperation,  rw, off(0) on(1) cycle(2)
@@ -66,15 +66,15 @@ my $PDU2_OUTLETSTATE = ".1.3.6.1.4.1.13742.6.4.1.2.1.3";  #outletSwitchingState,
 my $PDU2_OUTLETCOUNT = ".1.3.6.1.4.1.13742.6.3.2.2.1.4";  #outletCount,          ro, INDEX { pduId }
 my $PDU2_PDUID       = 1;
 
+#PDUs whose linked-unit warning has already been printed.
+my %genpdu_linked;
+
 #rinv / rvitals sources. The nameplate and unitConfiguration entries are
 #INDEX { pduId }; the measurement and sensor-configuration entries are
 #INDEX { pduId, <inlet|outlet>Id, sensorType }.
-#A sensor whose SensorSignedMinimum is below zero can read negative and must be
-#taken from the signed value column, per the SensorSignedMinimum DESCRIPTION.
-#The two columns are not interchangeable in either direction: the unsigned
-#column is undefined for such a sensor (a PX4 answers 0 for inlet reactive
-#power), and the signed column is undefined for sensors whose range exceeds
-#Integer32 (the same PX4 answers 0 for active and apparent energy).
+#SensorSignedMinimum selects the value column. On a PX4 the unsigned column
+#returns 0 for reactive power, and the signed column returns 0 for active
+#energy, whose range exceeds Integer32.
 my $PDU2_NAMEPLATE       = ".1.3.6.1.4.1.13742.6.3.2.1.1";    #nameplateEntry
 my $PDU2_UNITCONF        = ".1.3.6.1.4.1.13742.6.3.2.2.1";    #unitConfigurationEntry
 my $PDU2_INLETVAL        = ".1.3.6.1.4.1.13742.6.5.2.3.1.4";  #measurementsInletSensorValue
@@ -841,11 +841,34 @@ sub pdu2_session_args {
 
 #-------------------------------------------------------
 
+=head3  pdu2_get
+
+   Read one object. Returns its value and one of 'ok', 'absent' or 'failed'.
+   A missing object is exception text under v2c/v3 and an error under v1, so
+   both count as absent rather than as a failure.
+
+=cut
+
+#-------------------------------------------------------
+sub pdu2_get {
+    my ($session, $oid) = @_;
+
+    #Cleared first so the error below belongs to this GET.
+    $session->{ErrorStr} = '';
+    my $val = $session->get($oid);
+    my $err = $session->{ErrorStr} || '';
+
+    return ($val, 'ok') if (defined $val and $val ne '' and $val !~ /^No Such/i);
+    return (undef, 'failed') if ($err ne '' and $err !~ /no such/i);
+    return (undef, 'absent');
+}
+
+#-------------------------------------------------------
+
 =head3  pdu2_session_probe
 
-   First exchange with a generic SNMP PDU. Returns false when the PDU does not
-   answer at all, so callers report a connection failure, and records whether
-   its outlets can be switched.
+   Returns false when the PDU does not answer, so callers report a connection
+   failure, and records whether its outlets can be switched.
 
 =cut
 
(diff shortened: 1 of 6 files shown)


viniciusferrao:
Hi @kcgthb thanks for the contribution.

I'll take a look on it.

xCAT-server/lib/xcat/plugins/pdu.pm
```diff
@@ -614,6 +731,215 @@ sub connectTopdu {
...
+    #Metered-only PDU2 models (for example PX2-1901U) answer the nameplate and
+    #measurement tables normally but have no outletSwitchControlTable instances.
+    #Probe once here, the same way the irpdu path above probes a version OID,
+    #so callers can refuse switching with one clear message instead of leaking
+    #the agent's "No Such Instance" text from every outlet read.
+    my $probe = $session->get("$PDU2_OUTLETSTATE.$PDU2_PDUID.1");
+    $session->{genpdu_switchable} =
+        ((defined $probe) and ($probe =~ /^\d+$/ or $probe =~ /^(on|off)$/i)) ? 1 : 0;
+
+    return $session;
```
viniciusferrao: Could we query `pduCount` before this?

A timeout or bad credentials can also leave this probe undefined, so the session would be returned as a non-switchable PDU.

Verifying `pduCount == 1` would also keep linked PDUs out of the current `pduId = 1` path.

xCAT-server/lib/xcat/plugins/pdu.pm
```diff
@@ -614,6 +731,215 @@ sub connectTopdu {
...
+   this to one GET per entity per sensor rather than three.
+
+=cut
+
+#-------------------------------------------------------
+sub pdu2_sensor_fmt {
+    my ($session, $valoid, $unitoid, $decoid, $idx, $cache) = @_;
+
+    my $raw = $session->get("$valoid.$idx");
+    return undef unless (defined $raw and $raw =~ /^-?\d+$/);
```
viniciusferrao: Could this account for signed sensor values? The value OID depends on `SensorSignedMinimum`.
 
 If the minimum is negative, PDU2-MIB says `.4` is absent and the reading is in `.6`. Reactive power can hit this case and would otherwise be silently skipped.
 
 Something along these lines:

 ```perl
 my $min = $session->get("$signedminoid.$idx");
 return undef unless defined($min) && $min =~ /^-?\d+$/;
 
 my $oid = ($min < 0) ? $signedvaloid : $valoid;
 my $raw = $session->get("$oid.$idx");
 ```

xCAT-server/lib/xcat/plugins/pdu.pm
```diff
@@ -46,6 +46,63 @@ my $callback;
...
+#INDEX { pduId }; the measurement and sensor-configuration entries are
+#INDEX { pduId, <inlet|outlet>Id, sensorType }.
+my $PDU2_NAMEPLATE   = ".1.3.6.1.4.1.13742.6.3.2.1.1";    #nameplateEntry
+my $PDU2_UNITCONF    = ".1.3.6.1.4.1.13742.6.3.2.2.1";    #unitConfigurationEntry
+my $PDU2_INLETVAL    = ".1.3.6.1.4.1.13742.6.5.2.3.1.4";  #measurementsInletSensorValue
+my $PDU2_INLETUNITS  = ".1.3.6.1.4.1.13742.6.3.3.4.1.6";  #inletSensorUnits
+my $PDU2_INLETDEC    = ".1.3.6.1.4.1.13742.6.3.3.4.1.7";  #inletSensorDecimalDigits
+my $PDU2_OUTLETVAL   = ".1.3.6.1.4.1.13742.6.5.4.3.1.4";  #measurementsOutletSensorValue
+my $PDU2_OUTLETUNITS = ".1.3.6.1.4.1.13742.6.3.5.4.1.6";  #outletSensorUnits
+my $PDU2_OUTLETDEC   = ".1.3.6.1.4.1.13742.6.3.5.4.1.7";  #outletSensorDecimalDigits
```
viniciusferrao: Here you should probably add the signed sensor constants:

```perl
  my $PDU2_INLETVAL       = ".1.3.6.1.4.1.13742.6.5.2.3.1.4";  #measurementsInletSensorValue
  my $PDU2_INLETSIGNEDVAL = ".1.3.6.1.4.1.13742.6.5.2.3.1.6";  #measurementsInletSensorSignedValue
  my $PDU2_INLETSIGNEDMIN = ".1.3.6.1.4.1.13742.6.3.3.4.1.27"; #inletSensorSignedMinimum
  my $PDU2_INLETUNITS     = ".1.3.6.1.4.1.13742.6.3.3.4.1.6";  #inletSensorUnits
  my $PDU2_INLETDEC       = ".1.3.6.1.4.1.13742.6.3.3.4.1.7";  #inletSensorDecimalDigits

  my $PDU2_OUTLETVAL       = ".1.3.6.1.4.1.13742.6.5.4.3.1.4";  #measurementsOutletSensorValue
  my $PDU2_OUTLETSIGNEDVAL = ".1.3.6.1.4.1.13742.6.5.4.3.1.6";  #measurementsOutletSensorSignedValue
  my $PDU2_OUTLETSIGNEDMIN = ".1.3.6.1.4.1.13742.6.3.5.4.1.27"; #outletSensorSignedMinimum
  my $PDU2_OUTLETUNITS     = ".1.3.6.1.4.1.13742.6.3.5.4.1.6";  #outletSensorUnits
  my $PDU2_OUTLETDEC       = ".1.3.6.1.4.1.13742.6.3.5.4.1.7";  #outletSensorDecimalDigits
```

xCAT-server/lib/xcat/plugins/pdu.pm
```diff
@@ -46,6 +46,63 @@ my $callback;
...
+# pduId is the PDU's link ID and is 1 when linking is not used; BCM2/PMC use 0
+# for the main controller. Linked and BCM2/PMC deployments are out of scope for
+# genpdu, because pduoutlet.pdu is parsed as PDU:outlet and has no field for a
+# link ID. Read pduCount (.1.3.6.1.4.1.13742.6.3.1.0) to confirm a unit is
+# unlinked; it returns 1 when it is.
+#-------------------------------------------------------
+my $PDU2_SWITCHOP    = ".1.3.6.1.4.1.13742.6.4.1.2.1.2";  #switchingOperation,  rw, off(0) on(1) cycle(2)
+my $PDU2_OUTLETSTATE = ".1.3.6.1.4.1.13742.6.4.1.2.1.3";  #outletSwitchingState, ro, on(7) off(8)
+my $PDU2_OUTLETCOUNT = ".1.3.6.1.4.1.13742.6.3.2.2.1.4";  #outletCount,          ro, INDEX { pduId }
+my $PDU2_PDUID       = 1;
```
viniciusferrao, line 67: Could we add a named constant for `pduCount`? It would let us verify that the PDU is standalone before continuing with `pduId = 1`.

```perl
my $PDU2_PDUCOUNT    = ".1.3.6.1.4.1.13742.6.3.1.0";      #pduCount, scalar
```

xCAT-server/lib/xcat/plugins/pdu.pm
```diff
@@ -614,6 +731,215 @@ sub connectTopdu {
...
+    for (my $outlet = 1; $outlet <= $count; $outlet++) {
+        foreach my $st (@PDU2_OUTLET_SENSORS) {
+            my $fmt = pdu2_sensor_fmt($session, $PDU2_OUTLETVAL, $PDU2_OUTLETUNITS,
+                        $PDU2_OUTLETDEC, "$PDU2_PDUID.$outlet.$st", \%outlet_cache);
+            next unless (defined $fmt);
+            my $name = $PDU2_SENSOR_NAME{$st} || "sensor $st";
+            xCAT::SvrUtils::sendmsg("outlet $outlet $name: $fmt", $callback, $pdu);
+        }
+    }
+}
```
viniciusferrao, line 1093: The calls here will also need to pass the signed-value and signed-minimum OIDs to `pdu2_sensor_fmt` using the suggested constants at the start of the file. I can draft a suggestion if you want.

viniciusferrao (changes requested):
Those are the suggestions to make the code more durable and more generic.

I don't have the PDUs to test anything, so the suggested changes are based only on references and documentation.

Also, do you want to add some tests? If we want to keep the functionality working tests are welcome. I can try to write them.

xCAT-server/lib/xcat/plugins/pdu.pm
kcgthb: Done, with the names you suggested. I also grouped the five columns read by `pdu2_sensor_fmt` into one hashref per entity type (`$PDU2_INLET_SENSOR` and `$PDU2_OUTLET_SENSOR`), otherwise the sub ends up with seven positional OID arguments.

xCAT-server/lib/xcat/plugins/pdu.pm
kcgthb: Good catch, thank you! I checked on one of our PX4s (PX4-5851-E7V2, fw 4.2.10.5-50400) and it's actually a bit worse than a skipped sensor:

```
inletSensorSignedMinimum            .27.1.1.29 = -63000
measurementsInletSensorValue         .4.1.1.29 = 0
measurementsInletSensorSignedValue   .6.1.1.29 = -1360
inletSensorUnits                     .6.1.1.29 = 23 (var)
```

The MIB says the unsigned column is undefined for those sensors rather than absent, and the device does answer it, with 0. So `rvitals` was happily reporting `Reactive Power: 0 var` for a real -1360 var reading. The sample output in the docs page even had that wrong value baked into it, now fixed.

I also checked what happens if we always use the signed column, but that doesn't work either: active energy comes back as 93653612 in the unsigned column and 0 in the signed one, since its range exceeds Integer32. So yes, the signed minimum has to decide which column to use.

I changed two details from your snippet:

- I didn't make column 27 mandatory, since older firmware might not implement it. An absent minimum keeps the unsigned reading, and only a failed read (timeout, auth error) skips the sensor.
- I read the minimum per entity instead of caching it, since it selects the value column and a stale one would give a wrong reading.

xCAT-server/lib/xcat/plugins/pdu.pm
kcgthb: Done, both call sites now pass the column set for the entity type. The pre-loop outlet probe still reads the unsigned column directly, since the first outlet sensor is `rmsCurrent`.

xCAT-server/lib/xcat/plugins/pdu.pm
kcgthb: Added as `$PDU2_PDUCOUNT`, and used in the session probe as described in the other thread.

xCAT-server/lib/xcat/plugins/pdu.pm
kcgthb: You're right, and it turns out the problem was a bit wider than the probe: the session builder never checked `ErrorStr`, unlike the irpdu path in `connectTopdu`. So an unreachable PDU or a wrong credential was reported as "this PDU does not support outlet switching", while `rinv` and `rvitals` printed nothing at all.

`pduCount` is now the first exchange with the device, in a new `pdu2_session_probe`. The callers already turn a failed probe into "Couldn't connect to \<pdu\>", so the three commands report it properly now. I checked that `pduCount` returns 1 on the three generations we have here: PX2-1901U-N1A6 and PX3-1901U-N1A6 on fw 4.0.20.5-49038, and PX4-5851-E7V2 on 4.2.10.5-50400. Since I can only test Raritan units, implementations without a usable `pduCount` get a nameplate read as a fallback.

For `pduCount != 1` I made it a warning rather than an error. The count includes link units, and the primary is still `pduId` 1, so its outlets are controlled correctly. Switching is also gated by the outlet state probe on that same `pduId`, so BCM2 and PMC meters aren't exposed as switchable. The warning is printed once per PDU.

kcgthb:
Thanks for the review! I added both changes.

I also added a second commit for an error handling problem I found around them: a missing object comes back as exception text under v2c/v3, but as an error under v1, and I was treating both cases the same way. That's how `rinv` could print "No Such Instance currently exists at this OID" as the value of a nameplate field that the model doesn't populate.

I added tests in `xCAT-test/unit/pdu_genpdu.t`, in the same style as `ipmi_rakp2.t`, using a fake SNMP session and readings captured from one of our PX4s. Happy to add more if you see anything missing.

viniciusferrao:
> Thanks for the review! I added both changes.
> 
> I also added a second commit for an error handling problem I found around them: a missing object comes back as exception text under v2c/v3, but as an error under v1, and I was treating both cases the same way. That's how `rinv` could print "No Such Instance currently exists at this OID" as the value of a nameplate field that the model doesn't populate.
> 
> I added tests in `xCAT-test/unit/pdu_genpdu.t`, in the same style as `ipmi_rakp2.t`, using a fake SNMP session and readings captured from one of our PX4s. Happy to add more if you see anything missing.
> 

I will take a look on it until tomorrow. This will land on 2.19.

Thank you.

xCAT-server/lib/xcat/plugins/pdu.pm
```diff
@@ -903,40 +918,42 @@ sub pdu2_session_probe {
...
-        my $u = $session->get("$oids->{units}.$idx");
-        my $d = $session->get("$oids->{dec}.$idx");
+    my $st = (split /\./, $idx)[-1];
+    unless (exists $cache->{$st}) {
+        #A failed read here would cache the wrong scaling for every entity
+        #behind this one.
+        my ($u, $ustate) = pdu2_get($session, "$oids->{units}.$idx");
+        my ($d, $dstate) = pdu2_get($session, "$oids->{dec}.$idx");
+        return undef if ($ustate eq 'failed' or $dstate eq 'failed');
         $u = -1 unless (defined $u and $u =~ /^-?\d+$/);
```
viniciusferrao: `UseSprintValue` may return `SensorUnitsEnumeration` as its MIB label
(for example `amp`, `volt`, or `watt`) when PDU2-MIB is loaded. This
line would then convert it to `-1`, causing `rvitals` to lose the unit
suffix.

Could we support both the numeric and enum-label representations here,
as `outletstat()` already does for the outlet state? Please also add a
test where the units GET returns `amp` rather than `2`.

viniciusferrao (changes requested):
@kcgthb everything seems fine. There's still just one comment on `pdu.pm`.

I think after that we are good to go. If you need help on the last one let me know.

xCAT-server/lib/xcat/plugins/pdu.pm
kcgthb: Good catch, thank you! You're right, and I could reproduce it on a PX4 with the MIB in the search path. SNMP.pm does return the label:

```
MIBS=unset        units = [2]     state = [7]
MIBS=+PDU2-MIB    units = [amp]   state = [on]
```

And `rvitals` on the same PDU, before and after the fix, a few minutes apart:

```
before:  412    9.452    6046    6239    0.97  60.0     9
after:   412 V  9.463 A  6055 W  6246 VA  0.97  60.0 Hz  9 %
```

So I added `pdu2_enum()`, which takes the number, the label, or the `label(value)` form Net-SNMP produces when quick printing is off, and used it for the sensor units and for `outletSwitchingState`. The test you asked for is in, with `amp`, plus `amp(2)` and a mixed case label, since the MIB spells some of them `wattHour` and `degreeC`.

While I was there I found a related one in the switching probe: it only accepted `on` or `off`, but an outlet can be in any of the `SensorStateEnumeration` states, so a PDU whose outlet 1 happened to be in `standby` would have been reported as metered. `pdu2_get()` already tells an absent instance from a failed read, so the probe now just checks that we got an answer at all.

Numeric readings are unaffected either way, and the plugin is running on our staging MN against all three PDU generations.

viniciusferrao (approved):
Thanks @kcgthb 

This will land for 2.19.

kcgthb:
Excellent, thank you!

PR #7745, opened 2026-08-21
fix(rspconfig): accept one network value per BMC

viniciusferrao:
A node can carry more than one BMC, and rspconfig already opens a session per BMC. A setting such as `ip=` carried a single value, so every BMC received the same one. Two BMCs cannot share an address, so such a node could not be configured through rspconfig at all.

A comma separated value now gives one setting per BMC, in the order the sessions are numbered. Only `ip`, `netmask` and `gateway` read a list, because a comma belongs to the value itself in a free form SNMP community string. A value without a comma still reaches every BMC, so existing use is unchanged. An entry that is missing or empty reports the mismatch instead of reaching the address encoders.

Recovered from the lenovobuild branch.

### DIFF
--- a/xCAT-server/lib/xcat/plugins/ipmi.pm
+++ b/xCAT-server/lib/xcat/plugins/ipmi.pm
@@ -868,6 +868,25 @@ sub next_setnetinfo {
     &setnetinfo($sessdata);
 }
 
+# The settings that take one value per BMC. Every other subcommand keeps its
+# argument whole, because a comma can belong to the value itself.
+my %per_bmc_subcommand = map { $_ => 1 } qw(ip netmask gateway);
+
+#-------------------------------------------------------
+# A node can carry more than one BMC, and each session then runs with its own
+# bmcnum. A comma separated value gives one setting per BMC, in that order.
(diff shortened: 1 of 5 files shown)


dhilst (approved):

xCAT-server/lib/xcat/plugins/ipmi.pm
```diff
@@ -895,6 +914,19 @@ sub setnetinfo {
...
+        unless (defined $bmcvalue) {
+            $callback->({ errorcode => [1], error => ["The value $argument does not carry a setting for BMC " . $sessdata->{bmcnum} . ", give one value per BMC"] });
+            return;
+        }
+        $argument = $bmcvalue;
+
+        # The follow-up callbacks read the subcommand again, so leave this
+        # session holding the value of its own BMC
+        $sessdata->{subcommand} = "$subcommand=$argument";
+    }
```
dhilst: Extract this into a pure function and test it instead of grepping the code

dhilst (changes requested):

dhilst (approved):

PR #7746, opened 2026-08-22
fix(blade): let a discovery request reach the findme handler

viniciusferrao:
A booting node sends a findme request, which names no node — the node is what the request asks xCAT to find. blade.pm's preprocessor demands a noderange for every command, so it answered `Missing Noderange` with an error code and returned nothing. The daemon then had no request to dispatch, so the findme handler never ran. The plugin has been answering every discovery request with an error, and blade discovery has not happened. The noderange check is older than the handler.

Hand a findme request through, as `switch.pm` does for commands it doesn't preprocess. Every other command keeps the check.

The handler asked every row of the `mp` table for an inventory, but that table also holds Power BMCs, FSPs, BPAs, HMCs and management modules, none of which answer a blade inventory. Select only blades: a row whose type is `blade`, or a row with no type whose `mpa` names a different node that is a management module — the shipped `e1350` template leaves a blade's type empty, so testing the type alone loses them. Return before reading the arp table when no blades are found, so a site without a chassis does no work.

### DIFF
--- a/perl-xCAT/xCAT/DiscoveryUtils.pm
+++ b/perl-xCAT/xCAT/DiscoveryUtils.pm
@@ -9,6 +9,31 @@ $XML::Simple::PREFERRED_PARSER = 'XML::Parser';
 
 use xCAT::MsgUtils;
 
+sub is_findme_request {
+    my ($request) = @_;
+
+    return 0 unless ref $request eq 'HASH';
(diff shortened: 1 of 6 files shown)


dhilst (changes requested):
Same as #7745, extract the behavior into a pure function and test it instead of testing the source code

dhilst (approved):

PR #7748, opened 2026-08-23
fix(genimage): resolve Mellanox drivers from the target kernel

### LINKED ISSUE #7781: genimage rhels10 issue with mlx_en
During genimage, using /opt/xcat/share/xcat/netboot/rh/genimage, the dracut-install call is failing on RHEL10 using xCAT 2.18.
       Error output: "dracut-install: Failed to find module 'mlx_en'"

I believe Redhat 10 has removed support for connectx mlx driver
How was xCAT 2.18+RHEL10 tested and got past this issue with genimage?  Was the mlx_en driver brought in via an external repo, from something like ELRepo?  Trying to understand if there's something incorrect in my env or if this is actually an open issue with genimage support and RHEL10 at this time.

This is the code in rh/genimage that still refers to mlx_en
```
# Add the default driver list
if ($arch eq 'x86' or $arch eq 'x86_64') {
    push @ndrivers, qw/tg3 bnx2 bnx2x e1000 e1000e igb mlx_en virtio_net be2net/;
} elsif ($arch eq 'aarch64') {
    push @ndrivers, qw/tg3 bnx2 bnx2x e1000e igb mlx_en virtio_net/;
} elsif ($arch eq 'ppc64') {
    push @ndrivers, qw/e1000 e1000e igb ibmveth ehea/;
} elsif ($arch eq 's390x') {
    push @ndrivers, qw/qdio ccwgroup/;
}
```
Thanks

viniciusferrao:
`genimage` adds its default network-driver list independently of the hardware installed in the management node. That list still contained `mlx_en`, so EL10 target kernels without mlx4 emitted a nonfatal missing-module diagnostic even though image generation completed.

Resolve optional Mellanox defaults from the target kernel: keep a real `mlx_en`, otherwise use `mlx4_en` when available, add `mlx5_core` when available, and omit unavailable defaults. Available `mlx4_ib`, `mlx5_ib`, and `ib_ipoib` modules are included for automatic defaults, including modules installed through driver updates. Explicit-only `netdrivers` lists are not expanded.

Genesis now tries mlx4 and mlx5 independently and succeeds when either loads. Regression tests cover legacy/OFED, mlx4, EL10, compressed modules, duplicate entries, InfiniBand, and explicit requests.

Closes #7781.

### DIFF
--- a/docs/source/guides/admin-guides/manage_clusters/common/deployment/install_new_kernel.rst
+++ b/docs/source/guides/admin-guides/manage_clusters/common/deployment/install_new_kernel.rst
@@ -84,9 +84,13 @@ The default driver list: ::
          sels-ppc: tg3 e1000 e1000e igb ibmveth ehea be2net
          sles-ppcle: scsi_mod libata scsi_tgt jbd2 mbcache crc16 virtio virtio_ring libahci crc-t10dif scsi_transport_srp af_packet ext3 ext4 virtio_pci virtio_blk scsi_dh ahci megaraid_sas sd_mod ibmvscsi
 
-The Mellanox entries are conditional. ``genimage`` retains a real ``mlx_en``
-module from the target kernel, otherwise uses ``mlx4_en`` when available, and
-adds ``mlx5_core`` when available. Missing optional Mellanox modules are omitted.
+The automatically added Mellanox entries are conditional. ``genimage`` retains
+a real ``mlx_en`` module from the target kernel, otherwise uses ``mlx4_en`` when
+available, and adds ``mlx5_core`` when available. Available ``mlx4_ib`` and
+``mlx5_ib`` modules are also added. If either InfiniBand driver and ``ib_ipoib``
+are available, ``ib_ipoib`` is added for IP over InfiniBand. Missing optional
+Mellanox modules are omitted. Explicit ``netdrivers`` entries do not enable
+other optional Mellanox defaults.
 
 Note: With this approach, xCAT will search for the drivers in the rootimage. You need to make sure the drivers have been included in the rootimage before generating the initrd. You can install the drivers manually in an existing rootimage (using chroot) and run genimage again, or you can use a postinstall script to install drivers to the rootimage during your initial genimage run.
 
--- a/docs/source/guides/admin-guides/manage_clusters/common/deployment/install_new_kernel.rst
+++ b/docs/source/guides/admin-guides/manage_clusters/common/deployment/install_new_kernel.rst
@@ -77,12 +77,16 @@ Generally, the genimage command has a default driver list which will be added to
 
 The default driver list: ::
 
-         rh-x86:   tg3 bnx2 bnx2x e1000 e1000e igb mlx_en virtio_net be2net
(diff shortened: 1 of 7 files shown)


dhilst (changes requested):
On ppc64el the default list at xCAT-server/share/xcat/netboot/ubuntu/genimage:204 already names both mlx_en and mlx4_en, so the new rewrite block turns the placeholder into a second mlx4_en.ko whenever the mlx4 tree is present — which is the normal case on ppc64le. Nothing de-dupes afterwards: the guard just above only covers driver-update-disk names, and the modules.dep loop's `unless (grep { $_ eq $dep } @ndrivers)` only protects dependencies, not the seed list. The duplicate is harmless in the staging copy (same destination, overwritten), but the generated init emits [ -f /lib/mlx4_en.ko ] && insmod /lib/mlx4_en.ko twice, so every ppc64el stateless boot logs a spurious insmod: can't insert 'mlx4_en.ko': File exists. Simplest fix is to drop the now-redundant mlx_en from the ppc64el list, or de-dupe ndrivers after the substitution. Separately, xCAT-test/unit/mellanox_netdrivers.t asserts on the source text of the scripts rather than on behaviour, and its second loop passes for sles only because sles/genimage:955 still contains a dead s/mlx_en/mlx4_en/ — that script's list uses mlx4_en directly, so the regex matches nothing but the substitution's own source line. As written the test would keep passing if the rewrite were moved somewhere it never runs, and it would not have caught the duplication above; a check on the resulting driver list would be worth more than a grep for the code that builds it.

viniciusferrao:
> On ppc64el the default list at xCAT-server/share/xcat/netboot/ubuntu/genimage:204 already names both mlx_en and mlx4_en, so the new rewrite block turns the placeholder into a second mlx4_en.ko whenever the mlx4 tree is present — which is the normal case on ppc64le. Nothing de-dupes afterwards: the guard just above only covers driver-update-disk names, and the modules.dep loop's unless (grep { $_ eq $dep } @ndrivers) only protects dependencies, not the seed list. The duplicate is harmless in the staging copy (same destination, overwritten), but the generated init emits [ -f /lib/mlx4_en.ko ] && insmod /lib/mlx4_en.ko twice, so every ppc64el stateless boot logs a spurious insmod: can't insert 'mlx4_en.ko': File exists. Simplest fix is to drop the now-redundant mlx_en from the ppc64el list, or de-dupe @ndrivers after the substitution. Separately, xCAT-test/unit/mellanox_netdrivers.t asserts on the source text of the scripts rather than on behaviour, and its second loop passes for sles only because sles/genimage:955 still contains a dead s/mlx_en/mlx4_en/ — that script's list uses mlx4_en directly, so the regex matches nothing but the substitution's own source line. As written the test would keep passing if the rewrite were moved somewhere it never runs, and it would not have caught the duplication above; a check on the resulting driver list would be worth more than a grep for the code that builds it.

That's a bug. Let me check that.

viniciusferrao:
@dhilst both issues are fixed. Can you look again?

dhilst (approved):

PR #7755, opened 2026-08-24
fix(xcat-core): capture Subiquity install failing when installnic is…

dhilst:
The Ubuntu diskful (Subiquity) install fails on any node that does not set noderes.installnic. compute.subiquity.tmpl resolves #TABLE:noderes:$NODE:installnic#, and Template.pm's tabdb raises "Unable to find requested field <installnic> from table <noderes>" when the value is absent, aborting with "Failed to generate xnba configurations" so the compute node never enters the installer. EL and SLES pass the identical case because their statefull templates never reference installnic (gen_net_boot_params defaults to the boot MAC).

Assert that the template uses the non-fatal #TABLEBLANKOKAY# token for installnic and treats an empty installnic the same as "mac" (match by MAC, no NIC rename). These assertions fail on the current template, capturing the defect.

### DIFF
--- a/xCAT-server/lib/perl/xCAT/Template.pm
+++ b/xCAT-server/lib/perl/xCAT/Template.pm
@@ -372,6 +372,8 @@ sub subvars {
         $inc =~ s/#XCATVAR:([^#]+)#/envvar($1)/eg;
         $inc =~ s/#ENV:([^#]+)#/envvar($1)/eg;
         $inc =~ s/#UBUNTU_SUBIQUITY_APT_CONFIG#/ubuntu_subiquity_apt_config($media_dir)/eg;
+        $inc =~ s/#SUBIQUITYINSTALLNIC#/subiquity_install_nic()/eg;
+        $inc =~ s/#SUBIQUITYINSTALLMAC#/subiquity_install_mac()/eg;
         $inc =~ s/#MACHINEPASSWORD#/machinepassword()/eg;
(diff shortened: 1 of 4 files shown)


viniciusferrao (changes requested):
Hi @dhilst.

I think there is one behavior that I think needs to be corrected before merging.

### `primarynic` is skipped when `installnic` is empty

xCAT's fallback order is:

```text
installnic -> primarynic -> mac
```

The current change treats an empty `installnic` as `mac` immediately. Because of that, a node with `primarynic` configured but no `installnic` will ignore `primarynic` and generate the target netplan using `mac.mac`.

Could we preserve the existing fallback behavior here?

Ideally, the effective interface and MAC should be resolved in Perl, possibly by reusing `gen_net_boot_params()`, instead of duplicating part of its logic in the shell template. That would also handle cases where `primarynic` contains either an interface name or a MAC address.

I would also extend the test coverage to include:

- `installnic` explicitly set;
- `installnic` empty and `primarynic` set to an interface name;
- `installnic` empty and `primarynic` set to a MAC address;
- both attributes empty, which should finally fall back to `mac`.

The current test only checks literal strings in the template and assumes that an empty `installnic` always means `mac`. Rendering the template with controlled values and checking the generated netplan would provide better regression coverage.

viniciusferrao (approved):

viniciusferrao:
Approved.

PR #7756, opened 2026-08-24
fix(xcat-core): make xCAT work on Ubuntu 26.04 (resolute)

dhilst:
Two independent defects stop xCAT working on Ubuntu 26.04 (resolute).

`bind9` is only a `Recommends` on the `xcat` and `xcatsn` metapackages. `bind9` is required by `makedns`, so it must be a hard dependency regardless of the system's APT recommendation policy — as a `Recommends` it is installed only while that policy asks for it, and nothing guarantees `named` is on the node. On an Ubuntu 26.04 management node it was not there: no `/usr/sbin/named`, and `makedns` failed with "failed to start named".

The Ubuntu netboot `genimage` reduces an osimage's `osvers` to a debootstrap suite with a bare `s/\.\d+$//`. That also strips the minor from a two-part `osvers` — an initial release with no point-release ISO — so `ubuntu26.04` becomes `26` and debootstrap dies with `E: No such script: /usr/share/debootstrap/scripts/26`. It is not 26.04-specific: an initial-release 18.04/20.04/22.04 ISO hits it too.

Assert `bind9` is a hard `Depends` on both metapackages, and drive `genimage`'s real codename derivation over a table of `osvers`. Both fail today.

### DIFF
--- a/xCAT-server/share/xcat/netboot/ubuntu/genimage
+++ b/xCAT-server/share/xcat/netboot/ubuntu/genimage
@@ -270,7 +270,10 @@ unless ($onlyinitrd) {
         my $mirror = (defined $aptmirror[0] && length $aptmirror[0])
                    ? $aptmirror[0] : 'http://archive.ubuntu.com/ubuntu';
         (my $codename = $osver) =~ s/^ubuntu//;
-        $codename =~ s/\.\d+$//;                  # 24.04.4 -> 24.04
+        # Strip ONLY a trailing third component. A bare s/\.\d+$// also strips the minor from a
+        # two-part osvers (ubuntu26.04 -> "26"), which misses the %cn map below and reaches
+        # debootstrap as a bogus suite ("E: No such script: .../debootstrap/scripts/26").
+        $codename =~ s/^(\d+\.\d+)\.\d+$/$1/;   # 24.04.4 -> 24.04; 26.04 stays 26.04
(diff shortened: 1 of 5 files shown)


viniciusferrao (changes requested):
@dhilst This is more cosmetic, but I have one concern before approving.

The PR currently says that APT 3 / solver3 does not install `Recommends` by default. I could not reproduce that behavior. With APT 3.0.3, recommended packages were still installed unless `APT::Install-Recommends` was disabled or `--no-install-recommends` was used.

This does not invalidate the dependency change. The stronger and simpler justification is that `bind9` is required for xCAT DNS management, so it should be a hard dependency regardless of the system's APT policy.

I suggest using wording along these lines:

> `bind9` is required by `makedns`, so it must be a hard dependency regardless of the system's APT recommendation policy.

Could you update the PR description, commit messages, test comments, and test assertion text to avoid attributing this specifically to solver3?

- Please fix the indentation of the new comment and regex in `genimage`. They are inside the `unless` block but currently look like they are outside it.
- Since this PR includes production fixes, the title should probably use `fix(xcat-core):` rather than `test(xcat-core):`.

Or just remove the comments, they may sound uncessary to be honest.

viniciusferrao (approved):

viniciusferrao:
Merged @dhilst

PR #7758, opened 2026-08-24
fix(dhcp): makedhcp fails on a stock Ubuntu MN, and makedhcp -q hangs or answers wrongly

### LINKED ISSUE #7710: Ubuntu >= 22.04: xcat metapackage guarantees isc-dhcp-server while DHCP backend auto-selection defaults to Kea

dhilst:
Closes #7710.

On a stock Ubuntu management node `makedhcp` fails at startup, and `makedhcp -q` misreports.
Auto-selection picks kea where apt installed only isc-dhcp-server. The query spawns omshell on
the releases whose write paths avoid it, and hangs until the caller times out. It reports an
unreadable `dhcpd.conf` as a node with no reservation. It reports an InfiniBand node with no
hardware address.

All four faults are in `xCAT-server/lib/xcat/plugins/dhcp.pm`. Backend selection returns the
preferred backend without checking that it is installed. `listnode` calls `_omshell_query_host`
unconditionally. `_query_isc_static_host` tests `dhcpd.conf` with `-r`, then opens it unchecked.
Its parser matches `hardware ethernet` only, and the last declaration between the markers
overwrites the first.

Auto-selection now falls back to the installed backend and records `fallback_from`, so
`process_request` names the package to install; a backend forced through `site.dhcpbackend`
still fails hard. The query reads the static host block from `dhcpd.conf` and never spawns
omshell. `_read_isc_conf_lines` returns the read error, and `listnode` answers `error` with
`errorcode 1`. The parser accepts any hardware type and keeps the first declaration. The query
and the delete share `_isc_host_start_re` and `_isc_host_end_re`, and `_delete_isc_static_host`
accepts an explicit line list.

`dhcp_backend_fallback.t`, `dhcp_isc_static_host_query.t` and the additions to
`dhcp_static_host_markers.t` were pushed before their fixes. CI ran red on the test commits and
green on the fixes.

Part of #7725 split

### DIFF
--- a/perl-xCAT/xCAT/DHCP/Backend.pm
+++ b/perl-xCAT/xCAT/DHCP/Backend.pm
@@ -33,6 +33,20 @@ sub choose {
     my $selected = $normalized eq 'auto' ? $class->default_backend(%args) : $normalized;
 
     if ( $args{check_available} && !$class->available( $selected, %args ) ) {
+        # Ubuntu's metapackage guarantees isc-dhcp-server but only Recommends kea, so a
+        # --no-install-recommends install has auto preferring a kea that is not there. Fall
+        # back to whatever is installed; an explicitly forced backend still fails hard. #7710
+        if ( $normalized eq 'auto' ) {
+            for my $alt (qw(kea isc)) {
+                next if $alt eq $selected;
+                next unless $class->available( $alt, %args );
+                return {
+                    requested     => $normalized,
+                    name          => $alt,
+                    fallback_from => $selected,
+                };
+            }
+        }
         return {
             requested => $normalized,
             name      => $selected,
--- a/perl-xCAT/xCAT/DHCP/Backend/ISC.pm
+++ b/perl-xCAT/xCAT/DHCP/Backend/ISC.pm
@@ -12,6 +12,12 @@ sub name {
     return 'isc';
 }
 
+# The backend auto-selection preferred, and undef when it did not fall back. #7710
+sub fallback_from {
+    my $self = shift;
+    return ref($self) ? $self->{selection}{fallback_from} : undef;
+}
+
 sub implemented {
     return 1;
 }
--- a/perl-xCAT/xCAT/DHCP/Backend/Kea.pm
+++ b/perl-xCAT/xCAT/DHCP/Backend/Kea.pm
@@ -22,6 +22,12 @@ sub name {
     return 'kea';
 }
 
+# The backend auto-selection preferred, and undef when it did not fall back. #7710
+sub fallback_from {
+    my $self = shift;
+    return ref($self) ? $self->{selection}{fallback_from} : undef;
+}
+
 sub implemented {
     return 1;
 }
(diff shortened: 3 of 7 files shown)


viniciusferrao (dismissed):
Looks good.

Althrough there's a lot of verbose comments.

viniciusferrao:
@dhilst some points:

1. The PR now conflicts with `master`, and the conflicting static-host API was
   extended by #7777 to preserve multiple declarations and IPoIB twins. The
   rebase must retain the current `$config`/`$hostname` deletion semantics.

2. Failure to read `dhcpd.conf` is currently reported as a successful
   "reservation not found" result.

3. The query parser only understands `hardware ethernet`, while `master` can
   now generate primary and twin InfiniBand declarations.

viniciusferrao (changes requested):
I've commented without opening the request changes. Take a look at my last one.

viniciusferrao (approved):
LGTM

PR #7759, opened 2026-08-24
fix(xcat-core): a dead xcatd install monitor never comes back

dhilst:
### Problem

xcatd forks the install monitor once, at startup. The monitor listens on `xcatiport` for node install-status updates and the `next` boot-flip request. When it dies the SIGCHLD reaper clears `$pid_MON` and nothing forks it again. One death of that child, from a stray signal or a lost socket takeover during an xcatd restart, leaves the port dead for the life of the daemon: installing nodes cannot report booted or ask for the boot flip until xcatd is restarted.

### Fix

xcatd re-forks the monitor from the main service loop when `$pid_MON` is clear and `xcatiport` is set. The child closes the SSL listener, the UDP control socket, the rescanplugins channel and the undispatched client connections before entering `do_installm_service`.

Both reapers call `reap_install_monitor`, so the death is recorded whichever one is the handler.

`do_installm_service` dies when it cannot bind the port, so an unguarded re-fork would spin. `xCAT::RespawnUtils` doubles the delay from `XCATD_MON_RESPAWN_MIN_INTERVAL` (5s) to `XCATD_MON_RESPAWN_MAX_INTERVAL` (300s). Retrying never stops: with no monitor alive, nothing can reset a spent budget. A monitor that ran `XCATD_MON_RESPAWN_HEALTHY` seconds (60) resets the delay.

Part of #7725 split

### DIFF
--- a/perl-xCAT/xCAT/RespawnUtils.pm
+++ b/perl-xCAT/xCAT/RespawnUtils.pm
@@ -1,51 +1,38 @@
 # IBM(c) 2007 EPL license http://www.eclipse.org/legal/epl-v10.html
 package xCAT::RespawnUtils;
 
-# Pacing for a parent that has to keep a child alive; xcatd's install monitor is the caller.
+# Backoff for a parent that re-forks a child when it dies; xcatd's install monitor is the
+# caller. The delay doubles from min_interval to max_interval and holds there.
 #
-# The child is re-forked whenever it dies, but not as fast as fork() returns -- it may be
-# dying because a resource it needs is held by someone else, and retrying flat out burns CPU
-# and interferes with whatever handshake it performs to claim that resource. So attempts back
-# off, doubling from min_interval to max_interval and then holding there.
+# There is no attempt limit. A spent budget cannot be refilled: with no child alive nothing
+# resets it, and the resource stays dead until xcatd restarts. A child that ran `healthy`
+# seconds served, so its death resets the delay.
 #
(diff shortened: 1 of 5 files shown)


viniciusferrao (changes requested):
Hi again @dhilst.

I think the rate limiting makes sense here, however the hard retry cap leaves `xcatd` in the same state this PR is intended to fix.

After failed respawns, `$mon_respawn_attempts` is incremented past the maximum and the parent stops trying permanently. It's a softlock situation.

Since no monitor is running, nothing can later reset the counter. If the port becomes available afterward, `xcatiport` remains dead until `xcatd` is restarted.

Please keep retrying at a bounded interval or backoff, or reset the retry budget after a cooldown.

A behavioral test covering “fail several times, release the port, then recover” would also be more useful here than only checking that the respawn code exists.

Thanks.

viniciusferrao (changes requested):
@dhilst some points:

1. The initial monitor is started while `generic_reaper` is still the `SIGCHLD` handler. If it dies before `ssl_reaper` is installed, it can be reaped without clearing `$pid_MON`, so the respawn logic never starts.

2. A respawned monitor inherits sockets already stored in `@pendingconnections`. Those should probably be closed in the child before entering `do_installm_service()`.

3. The PR description still documents the old retry limit and environment variables. It should be updated to match the new indefinite backoff behavior.

viniciusferrao (approved):
LGTM

PR #7760, opened 2026-08-24
fix(xcat-core): makentp fails on a stock Ubuntu MN (timesyncd cannot serve time)

dhilst:
### Problem

An xCAT management node must **serve** time to its compute nodes — they are pointed at `ntpservers=<xcatmaster>`. Ubuntu ships only `systemd-timesyncd`, an SNTP *client* that disciplines the local clock and cannot answer queries, and `makentp` supports chronyd and ntpd only. On a stock Ubuntu MN it therefore fails outright:

```
RUN:makentp
configuring management node: ...
Please make sure ntpd is installed on ...
CHECK:rc == 0  [Failed]
```

Two further defects compound it:

* `setupntp` hard-requires `hwclock` through `check_exec_or_exit`, but Ubuntu 24.04 moved `hwclock` into `util-linux-extra`, absent from minimal images. So the whole NTP setup aborts — including the clock step, which does not use `hwclock` at all.
* `systemd-timesyncd` is part of systemd rather than a time-daemon package, so nothing displaces it and it keeps disciplining the clock against whichever daemon xCAT configures.

### Fix

Add `xCAT::NTP::Backend`, a selector for the NTP daemon, in the same spirit as `xCAT::DHCP::Backend`. It honours `site.ntpbackend`, defaults per distro family (EL7+/SLES15+ chrony, older ntpd, Ubuntu/Debian chrony) and downgrades to whichever of chrony/ntpd is actually installed rather than installing a second daemon. `makentp` selects through it instead of probing `/usr/sbin/chronyd` directly.

Make `hwclock` optional in `setupntp` — use it when present, log and continue when not — and disable `systemd-timesyncd` there. Pull `util-linux-extra` through `Recommends` and the diskless pkglist so `hwclock` is usually available anyway.

### On the hard dependency

This adds `chrony | ntp` to the `xcat` metapackage's `Depends` and `(chrony or ntp)` to the `xCAT` rpm's `Requires`. That is deliberate and worth calling out, because Ubuntu's default is timesyncd and this overrides it for anyone installing the metapackage.

The argument for it: this sits on the **metapackage**, whose stated purpose is "a common, default xCAT setup", and which already hard-depends on `isc-dhcp-server | kea`, `apache2` and `nfs-kernel-server`. Anyone installing it is standing up a management node, and a management node that cannot serve time cannot serve its compute nodes. It is an alternative (`chrony | ntp`), not a forced choice of chrony.

If the maintainers would rather not carry it, the selector and the `setupntp` fixes stand on their own and the dependency can drop to a `Recommends` — say so and I will respin.

### DIFF
--- a/docs/source/guides/admin-guides/references/man1/makentp.1.rst
+++ b/docs/source/guides/admin-guides/references/man1/makentp.1.rst
@@ -32,6 +32,8 @@ By default, it sets up the NTP server for xCAT management node. If \ **-a**\  fl
 
 \ *site.ntpservers*\  -- the NTP servers for the service node and compute node to sync with. The keyword <xcatmaster> means that the node's NTP server is the node that is managing it (either its service node or the management node).
 
+\ *site.ntpbackend*\  -- the NTP daemon to configure. Valid values are auto, chrony, and ntpd. The default is auto, which selects chrony on distributions that ship it and ntpd on older ones such as EL6 and SLES 12. If the selected daemon is not installed, \ **makentp**\  uses the other one and reports the change. The same value reaches the service nodes and the compute nodes, which the \ *setupntp*\  postscript also accepts as \ **-**\ **-backend chrony|ntpd**\ .
+
 To setup NTP on the compute node, add \ *setupntp*\  postscript to the \ *postscripts*\  table and run \ **updatenode node -P setupntp**\  command.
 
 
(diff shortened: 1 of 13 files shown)


viniciusferrao (changes requested):
@dhilst some points:

1. `setupntp` fallback is asymmetric. `chrony` can fall back to `ntpd`, but `--backend ntpd` goes directly to `setupntp.traditional` without checking if `ntpd` exists. A node with only chrony installed would therefore fail instead of falling back to chrony.

2. `xCAT::NTP::Backend` considers chrony available when `chronyd` exists, while `makentp` requires both `chronyd` and `systemctl`. These checks should probably be consistent.

3. Since `site.ntpbackend` is now user-facing, it should probably be documented in the site table/schema help alongside `site.dhcpbackend`.

Anyway, it looks good.

viniciusferrao (approved):
LGTM

PR #7761, opened 2026-08-24
fix(xcat-core): the Ubuntu Subiquity diskful install never completes

dhilst:
reg_linux_diskfull_installation_flat has never passed on Ubuntu. All the case ever reports is:

ssh: connect to host <cn> port 22: Connection refused

Behind that are five defects stacked on each other, which is why they are in one PR. Each only becomes visible once the previous one is fixed — until casper boots you cannot discover the installer has no DNS, and until apt works you never reach the reboot that hangs. Four cannot be reproduced alone.

casper does not boot. It panics with "Unable to find a medium containing a live file system" and PXE-loops: without boot=casper it never processes netboot=nfs and scans local disks instead. That gets you to the next one, nfsmount: can't parse IP address — casper mounts the live filesystem with klibc's nfsmount, which has no resolver, so nfsroot must be an address. The ds= URL stays a name; cloud-init fetches it later, where DNS works.

The node installs and never reboots. systemd-shutdown waits forever on an lvm/pvscan stuck in uninterruptible I/O against the still-mounted NFS root. toram has casper copy the squashfs to RAM and unmount the source, leaving nothing to wait on. casper parses only nfsroot= and takes the whole value as the path, so mount options cannot be appended — toram is its supported alternative.

The installer has no DNS, so the install hangs on archive.ubuntu.com. glibc discards a nameserver line naming a host, so the xcatmaster name left the installer — and the in-target apt inheriting the file — with nothing usable. Resolve it to an address first, while the live installer's DHCP resolv.conf still works.

In-target apt cannot find packages: E: Unable to locate package chrony. On classic-sources releases Subiquity renders the target's sources.list from the install media alone. sources_list: looks like the fix but is a curtin key Subiquity's schema ignores, so the archive goes in through sources:. Deb822 releases (24.04+) are excluded — their primary mirror already lands in ubuntu.sources and legacy .list files would duplicate the suites.

The node PXE-loops back into the installer. The boot flip is a next request to xcatd's install monitor on port 3002, and it was going unanswered. The flip now goes out from the live installer over bash's /dev/tcp, retried, with the exchange checked and failures written to the install log instead of silently starting another install. The monitor dying and never coming back is fixed separately in #7759

A perfect install reports status=failed. syncfiles runs in the in-target chroot and asks the MN to scp files into a node with no sshd yet. It is deferred to the postbootscripts, on the diskful path only — netboot and statelite already run there — and prepended, so it still precedes anything consuming the files.

Also: wait_for_provision in retry_install.sh is now overridable via WAIT_FOR_PROVISION. The default stays 20 minutes so a boot loop fails fast; Ubuntu diskful is simply slower on the happy path.


viniciusferrao (changes requested):
@dhilst some points:

1. The boot-flip command hardcodes port `3002`, but the install monitor uses `site.xcatiport`. This should probably use the configured port, with `3002` only as the default.

2. The boot-flip considers any replies a successful handshake. Could we explicitly require `ready` before sending `next`, and `done` afterward?

3. `subiquity_nfsroot_server()` uses `getipaddr()` without forcing IPv4, while this path is configured with `ip=dhcp`/`dhcp4`. On dual-stack systems it may return IPv6 and generate an invalid `nfsroot=<ipv6>:<path>`.

Also, if `getent ahostsv4` fails in the DNS setup, we fall back to writing the hostname as a `nameserver`, which brings back the original invalid `resolv.conf` case.

viniciusferrao:
@dhilst, I still see one question:

* `getipaddr($host, OnlyV4 => 1)` can still return a cached address from an earlier unrestricted lookup, because the `getipaddr()` cache bypass only checks `OnlyV6` and `GetAllAddresses`, not `OnlyV4`. If the cached address is IPv6, we can still generate the invalid `nfsroot=<ipv6>:<path>` case.

Could we either fix the cache behavior for `OnlyV4`, or otherwise make sure this lookup cannot reuse an IPv6 cached result?

viniciusferrao (approved):
LGTM

PR #7762, opened 2026-08-24
fix(xcat-core): configeth configures nothing on netplan-rendered Ubuntu nodes

### LINKED ISSUE #7454: netplan configuration for Ubuntu in configeth script

dhilst:
Closes #7454.

### Problem

Ubuntu 18.04 and later render the network with netplan. ifupdown is not installed and `/etc/network/interfaces.d/*` is ignored entirely — so `configeth`'s Debian branch, which writes exactly there, configures nothing at all on a modern Ubuntu node.

### Fix

Write `/etc/netplan/90-xcat-<nic>.yaml` and `netplan apply` when netplan is actually installed; the legacy ifupdown branch is kept for nodes that still run it.

The drop-in is **regenerated in full** on every change rather than edited in place. Its inputs are recorded in `# xcat-state:` comment lines in the same file, which netplan ignores, so the file stays the single source of truth and re-rendering is idempotent. That is what makes these hold — an in-place `sed` editor gets all three wrong:

* A NIC named `<parent>.<vid>` is emitted under `vlans:` with `id` and `link`, so netplan recreates the VLAN after a reboot. The ifupdown branch has always written `vlan-raw-device` for these; declaring them as plain ethernets would lose that. A dotted name whose suffix is not numeric stays an ethernet.
* Addresses keep the order they were added.
* A route is deduplicated on the whole `(to, via)` pair, so a second route sharing a gateway is still written.

`nicextraparams` are carried into the interface stanza, matching what the ifupdown branch does with them.

### DIFF
--- a/xCAT-test/unit/configeth_netplan.t
+++ b/xCAT-test/unit/configeth_netplan.t
@@ -11,19 +11,13 @@ use Test::More;
 # not installed and /etc/network/interfaces.d/* is ignored entirely, so configeth's Debian
 # branch configured nothing at all. It must write /etc/netplan/*.yaml and `netplan apply`.
 #
-# Three things the writer has to get right, all of which a naive in-place editor gets wrong:
+# These tests drive configipv4/configipv6/delete_nic_config_files -- the branch selection that
+# is the actual fix -- rather than the private write_netplan_* writers underneath them. Driving
+# only the writers passes even when the netplan branch is deleted from configipv4 outright.
 #
-#   * A VLAN interface (<parent>.<vid>) belongs under `vlans:` with `id` and `link`. Declared
-#     as a plain ethernet it is never recreated after a reboot.
-#   * Multiple addresses on one NIC must keep the order they were added.
-#   * Routes must be idempotent on the whole (to, via) pair -- not on either field alone, or a
-#     second route sharing a gateway is silently swallowed.
-#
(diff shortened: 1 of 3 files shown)


viniciusferrao (changes requested):
Hi @dhilst 

Some points:

1. On a dual-stack NIC where IPv6 appears before IPv4, `configipv6` writes the IPv6 address, but the first `configipv4` call resets the Netplan file again because `num_v4num == 0`. The final file loses IPv6. I think the reset needs to happen once per NIC, not once per address family.

2. During diskful provisioning, when `reboot_nic_bool=0`, the Netplan branch can still run `ip link set ... down`, but later skips `netplan_apply`. This may leave the interface down until reboot. The down operation should probably use the same guard, or be avoided entirely for Netplan.

3. VLAN parents are always declared under `ethernets`. A VLAN such as `bond0.100` or `br0.100` would redefine an existing bond/bridge as an Ethernet device, which Netplan rejects.

4. There is also one merge case worth testing: Netplan appends `addresses` and `routes` from multiple YAML files. An earlier cloud-init/installer address may therefore survive alongside the new xCAT address, even with the `90-xcat-*` drop-in.

Look if they make sense.

viniciusferrao (approved):
LGTM

PR #7765, opened 2026-08-24
refactor(genimage): reuse shared OS version parser

viniciusferrao:
Reuse `xCAT::SvrUtils::parseosver` when selecting the Enterprise Linux major version for RH genimage package-manager commands. Keep an explicit allowlist for supported RH-family names, including CentOS Stream and legacy RHEL Client and ComputeNode variants.

Preserve the existing yum fallback and EL8+ DNF, releasever, and module-platform behavior.

### DIFF
--- a/xCAT-test/unit/rh_genimage_package_manager.t
+++ b/xCAT-test/unit/rh_genimage_package_manager.t
@@ -3,19 +3,82 @@ use strict;
 use warnings;
 
 use FindBin;
-use lib "$FindBin::Bin/../lib";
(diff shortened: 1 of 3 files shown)


xCAT-test/unit/rh_genimage_package_manager.t
dhilst: We have helpers for that

dhilst (changes requested):

dhilst (approved):

PR #7766, opened 2026-08-24
refactor(routeop): reuse route classification helpers

viniciusferrao:
Reuse routeop's existing `is_ipv6_route` and `ipv4_route_prefix` helpers for route family classification and dotted-netmask conversion.

Preserve the existing IPv4, IPv6, default-route, gateway, interface, and command-selection behavior across route existence and add, delete, replace, and persistent-route paths.


dhilst (approved):
This is another case we would benefit of using BATS

PR #7767, opened 2026-08-25
refactor(utils): centralize executable lookup

viniciusferrao:
Add a dependency-light `xCAT::CommandUtils::find_executable` helper and reuse it across DHCP backend selection, Kea command discovery, and xCAT-probe command checks.

Preserve each caller's existing PATH order, system fallback policy, return shape, and shell-free executable checks while shipping the helper in the RPM, Debian, and xCAT-probe payloads.

### DIFF
--- a/xCAT-test/unit/command_utils.t
+++ b/xCAT-test/unit/command_utils.t
@@ -2,23 +2,20 @@ use strict;
 use warnings;
 
 use FindBin;
+use lib "$FindBin::Bin/../lib";
 use lib "$FindBin::Bin/../../perl-xCAT";
 
 use Cwd qw/getcwd/;
 use File::Path qw/make_path/;
+use File::Slurper qw/write_text/;
 use File::Temp qw/tempdir/;
 use Test::More;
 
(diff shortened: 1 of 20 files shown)


xCAT-test/unit/command_utils.t
```diff
@@ -0,0 +1,215 @@
...
+open( my $source_fh, '<', $command_utils_source )
+  or die "Unable to read $command_utils_source: $!";
+my $source = do { local $/; <$source_fh> };
+close($source_fh) or die "Unable to close $command_utils_source: $!";
+like(
+    $source,
+    qr/my \@SYSTEM_FALLBACK_DIRS = qw\(\s*\/usr\/sbin\s*\/usr\/bin\s*\/sbin\s*\/bin\s*\);/s,
+    'the built-in system fallback order matches the replaced callers'
+);
+
```
dhilst, line 21: We can use is_deeply here instead of grepping the source

dhilst (changes requested):

dhilst (approved):

PR #7768, opened 2026-08-25
test: centralize repository file reads

viniciusferrao:
Add shared source-tree test helpers for resolving repository paths and reading complete files with cwd-independent behavior and explicit I/O diagnostics.

Reuse them in the EL10 netboot DHCP client, Ubuntu live media guardrail, and RH genimage package manager tests while preserving all existing assertions and repository-relative paths.

### DIFF
--- a/xCAT-test/lib/XCAT/Test/File.pm
+++ b/xCAT-test/lib/XCAT/Test/File.pm
@@ -1,0 +1,73 @@
+package XCAT::Test::File;
+
+use strict;
+use warnings;
+
+use Cwd qw(abs_path);
+use Exporter qw(import);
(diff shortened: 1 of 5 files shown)


dhilst (approved):
Why not https://metacpan.org/pod/File::Slurper or even better https://metacpan.org/pod/Path::Tiny?

viniciusferrao:
> Why not https://metacpan.org/pod/File::Slurper or even better https://metacpan.org/pod/Path::Tiny?

These tests need to work from a clean checkout with the system Perl. `File::Slurper` and `Path::Tiny` aren't core modules or test dependencies, so they may not be installed. That's why the helper only uses core modules.

PR #7772, opened 2026-08-25
feat(postscripts): establish package utility loading

viniciusferrao:
`ospkgs` and `otherpkgs` will share package-specific shell helpers, but they can run from the source tree, `/install/postscripts`, or `/xcatpost`.

Add `xcatpkgutils.sh` as a POSIX `/bin/sh` sibling library and load it before package processing begins. Normal executions remain unchanged. If the postscript tree is incomplete or the library cannot be loaded, the scripts stop immediately with a clear error instead of continuing with an invalid runtime layout.

This establishes a portable shared-library boundary for follow-up deduplication without changing package selection or package-manager behavior.

### DIFF
--- a/xCAT/postscripts/ospkgs
+++ b/xCAT/postscripts/ospkgs
@@ -25,6 +25,22 @@
 
 #set -x
 
+XCATPKGUTILS_LOADER="$(dirname "$0")/xcatpkgutils-loader.sh"
+if [ ! -r "$XCATPKGUTILS_LOADER" ]; then
+    printf 'Error: package utility loader is not readable: %s\n' "$XCATPKGUTILS_LOADER" >&2
+    unset XCATPKGUTILS_LOADER
+    exit 1
+fi
+# shellcheck source-path=SCRIPTDIR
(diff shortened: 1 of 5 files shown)


xCAT/postscripts/ospkgs
```diff
@@ -25,6 +25,12 @@
 
 #set -x
 
+XCATPKGUTILS_LOADER="$(dirname "$0")/xcatpkgutils-loader.sh"
+if [ -r "$XCATPKGUTILS_LOADER" ]; then
+    . "$XCATPKGUTILS_LOADER" 2>/dev/null || :
+fi
+unset XCATPKGUTILS_LOADER
```
dhilst, line 32: Wouldn't `2>/dev/null || :` contradict "If the postscript tree is incomplete or the library cannot be loaded, the scripts stop immediately with a clear error instead of continuing with an invalid runtime layout."?

dhilst (dismissed):

dhilst (approved):

PR #7780, opened 2026-08-27
feat(confluent): export the switch topology of each node

viniciusferrao:
makeconfluentcfg now exports the switch and port of each node from the switch
table, keeping one entry per interface so a multi-NIC node does not lose all
but one port, and retracting topology that xCAT no longer holds.

Also fixes enclosure.manager and enclosure.bay, which were read from the wrong
table when the command runs with no node range and so were always empty.


dhilst (approved):

viniciusferrao:
I was hoping that @jjohnson42 would like to see that.

PR #7782, opened 2026-08-27
fix(template): honour installnic when the kickstart names the install device

viniciusferrao:
The kickstart network line named the install device from mac.mac alone, so a
node that sets noderes.installnic got a kernel command line that obeys it and a
kickstart that configures a different adapter.

Resolve the device through the order xCAT already uses for the netboot kernel
parameters, which the Ubuntu template was alone in sharing.

### DIFF
--- a/xCAT-test/unit/template_install_device.t
+++ b/xCAT-test/unit/template_install_device.t
@@ -1,0 +1,126 @@
+#!/usr/bin/env perl
+use strict;
+use warnings;
+
+use File::Spec;
+use FindBin;
+use Test::More;
+
+my $root = File::Spec->catdir( $FindBin::Bin, '..', '..' );
+BEGIN {
+    push @INC,
+      File::Spec->catdir( $FindBin::Bin, '..', '..', 'perl-xCAT' ),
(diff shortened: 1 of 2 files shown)


xCAT-test/unit/template_install_device.t
```diff
@@ -0,0 +1,126 @@
...
+# The source has to keep the unique local address built from the hardware
+# address, not from the device name.
+my $src = File::Spec->catfile( $root, 'xCAT-server', 'lib', 'perl', 'xCAT', 'Template.pm' );
+open( my $fh, '<', $src ) or die "Unable to read $src: $!";
+my $source = do { local $/; <$fh> };
+close($fh);
+
+my ($ksbody) = $source =~ /\nsub kickstartnetwork \{(.*?)\n\}\n/s;
+ok( defined($ksbody), 'the kickstartnetwork body was located' );
+
```
dhilst: This test is coupled to source code

viniciusferrao: Agreed and fixed.

dhilst (approved):

dhilst (approved):

PR #7789, opened 2026-08-30
refactor(postscripts): share RPM package manager discovery

viniciusferrao:
Move the repeated dnf/yum discovery in `ospkgs` and `otherpkgs` into the shared POSIX package utility library. The helper keeps the existing `/usr/bin/dnf`-before-`/usr/bin/yum` checks.

Each caller retains its current behavior when neither command is available: `ospkgs` reports the missing manager and exits, while `otherpkgs` continues through its zypper and raw RPM fallbacks.

### DIFF
--- a/xCAT/postscripts/xcatpkgutils.sh
+++ b/xCAT/postscripts/xcatpkgutils.sh
@@ -2,6 +2,18 @@
 # EPL license http://www.eclipse.org/legal/epl-v10.html
 
 # Shared POSIX shell helpers for the ospkgs and otherpkgs postscripts.
+
+xcat_find_rpm_package_manager()
+{
+    if [ -x "${1:-/usr/bin}/dnf" ]; then
+        printf '%s\n' dnf
+    elif [ -x "${1:-/usr/bin}/yum" ]; then
+        printf '%s\n' yum
+    else
+        return 1
+    fi
+}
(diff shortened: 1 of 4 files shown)


xCAT/postscripts/xcatpkgutils.sh
```diff
@@ -2,6 +2,18 @@
...
 
 # Shared POSIX shell helpers for the ospkgs and otherpkgs postscripts.
+
+xcat_find_rpm_package_manager()
+{
+    if [ -x "${1:-/usr/bin}/dnf" ]; then
+        printf '%s\n' dnf
+    elif [ -x "${1:-/usr/bin}/yum" ]; then
+        printf '%s\n' yum
+    else
```
dhilst, line 12: @viniciusferrao Are you sure we want these line breaks here? If these are being used as command we may see `dnf\n ...`

viniciusferrao: How? `\n` should be stripped when we do something like `cmd=$(function)`, as far as I know.

viniciusferrao: The `\n` is here so we use tools like `| while read`.

dhilst (approved):

PR #7800, opened 2026-08-31
refactor(dhcp): share OMAPI command runner

viniciusferrao:
Move the duplicated OMAPI command-file process handling into a shared DHCP module and use it from both `makedhcp` and `dhcpop`.


perl-xCAT/xCAT/DHCP/OmapiRunner.pm
```diff
@@ -0,0 +1,77 @@
...
+        sleep $class->_poll_interval();
+    }
+
+    kill 'KILL', $pid;
+    waitpid( $pid, 0 );
+    return 'killed';
+}
+
+sub _fork {
+    return fork();
```
dhilst: Shouldn't we use xcat's xfork here?

dhilst (approved):

PR #7802, opened 2026-08-31
refactor(networkutils): remove the legacy validate_ip helper

viniciusferrao:
`NetworkUtils` carried two IPv4 validators with different rules and no real IPv6 checking. `validate_ip` accepted any value containing a colon, and two callers tested its array-reference return instead of its content, so their checks could never fail.

There is now one validator per question: `isIpv4addr` (the former `isIpaddr`), `isIpv6addr`, and `isValidIp` for either family. All in-tree callers use them. `isIpaddr` and `validate_ip` remain as deprecated compatibility shims for one cycle.

### DIFF
--- a/perl-xCAT/xCAT/NetworkUtils.pm
+++ b/perl-xCAT/xCAT/NetworkUtils.pm
@@ -1687,7 +1687,7 @@ sub getNodeIPaddress
     }
 
     # Quick return if pass in an IP
-    return $nodetocheck if (xCAT::NetworkUtils->isIpaddr($nodetocheck));
+    return $nodetocheck if (xCAT::NetworkUtils->isIpv4addr($nodetocheck));
 
     my $nodeip = xCAT::NetworkUtils->getipaddr($nodetocheck);
     if (!$nodeip)
@@ -2132,7 +2132,7 @@ sub validate_ip
 
 #-------------------------------------------------------------------------------
 
-=head3    isIpaddr
+=head3    isIpv4addr
 
     returns 1 if parameter is has a valid IP address form.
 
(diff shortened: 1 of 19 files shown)


perl-xCAT/xCAT/NetworkUtils.pm
```diff
@@ -2183,6 +2138,136 @@ sub isIpaddr
...
+{
+    my ($class, @IPs) = @_;
+    foreach my $ip (@IPs) {
+        unless (isValidIp($ip)) {
+            return ([ 1, "Invalid IP address: $ip" ]);
+        }
+    }
+    return ([0]);
+}
+
```
dhilst: Shouldn't we remove this? Is there any caller of it in the source code?

xCAT-test/unit/networkutils_ip_validation.t
```diff
@@ -0,0 +1,163 @@
...
+        'xCAT-server/lib/xcat/plugins/DBobjectdefs.pm',
+        'xCAT-server/lib/xcat/plugins/hosts.pm',
+        'xCAT-server/lib/xcat/plugins/openbmc.pm',
+    );
+    foreach my $path (@renamed) {
+        unlike( slurp($path), qr/NetworkUtils(?:->|::)isIpaddr\(/,
+            "$path calls isIpv4addr instead of the deprecated alias" );
+    }
+};
+
```
dhilst: This test should be removed, together with the slurp helper

dhilst (changes requested):
I left the comments inline with the code

perl-xCAT/xCAT/NetworkUtils.pm
viniciusferrao: Yes. This was a leftover mistake.

viniciusferrao:
Done @dhilst

dhilst (approved):

PR #7806, opened 2026-09-01
build(xcat-core): Remove old build scripts and add --source-only to new scripts 

dhilst:
xcat-core has two build systems and only one runs. buildrpms.pl is what CI and CD call; makerpm, buildcore.sh and buildlocal.sh are the legacy build scripts. This PR removes the legacy scripts and refactors the Ubuntu package build to mirror what we already have for RPMs.

Changes

- buildrpms.pl --source-only. It already runs mock --buildsrpm; this skips the --rebuild. Rejected with --merge-core-repos.
- Source-only skips binary re-indexing and the .repo/buildinfo. Indexing an empty dir replaces good metadata with a repo that resolves nothing.
- Deletes makerpm, buildcore.sh, buildlocal.sh, #7774's helpers and their tests. Subsumes #7805 — its line is in the fixture that goes with makerpm.
- builddebs.pl replaces build-ubunturepo. BUILDALL=1→--force, GPGSIGN=1→--gpg-sign, GPG_HOME→--gpg-home, DEST→--dest, DISTS→--dist. GSA upload, PROMOTE/PREGA and -d dropped.
- Builds once, publishes to every codename. Packages are Perl; only xCAT, xCATsn, xCAT-genesis-scripts carry an arch. No sbuild, no chroots.
- BuildUtils.pm: version/release, probe staging, arch and dist tables, control pinning, changelog rewrite, reprepro conf, build lock. Pure functions, driven by build_utils.t.
- build-ubunturepo kept as the oracle, do-not-extend. Delete once the pipelines move.
- Updates github_action_xcat_test.pl, travis.pl, docs. Fixes the CI comment: build-ubunturepo was rm -rfing the checkout's own parent.
- build_ubunturepo_lock.t and ubuntu_2604_pkglist.t grepped source. Both run the code now.


viniciusferrao (changes requested):
Hi @dhilst some considerations:

1. This is easy to fix: `builddebs.pl` is not executable; `chmod +x` on it.

2. The `.orig.tar.gz` filename uses the wrong version

   This currently uses `$PKGVER`:

       orig_tarball_name($pkg, $PKGVER)

   For a version such as `2.19.0-snap...`, Debian expects the upstream tarball to be named with `2.19.0`, without the Debian revision. This should use `$VERSION`.

3. `--dest` can behave dangerously when the directory does not exist, i.e: `abs_path($opts{dest})` returns `undef` for a new path. This can make the output paths become `/debs` and `/xcat-core`.

   It can be fixed creating the destination first or using `File::Spec->rel2abs()`.

4. Genesis `debian/control` file is left behind: `xCAT-genesis-scripts` does not originally have a regular `debian/control`. The builder creates it from `control-amd64` or `control-ppc64el`, but the cleanup only restores files that already existed, making the checkout dirty and the next architecture may start with the previous architecture's control file.

viniciusferrao:
LGTM

viniciusferrao (approved):

PR #7807, opened 2026-09-01
test(httpd): stage the source-only helper in the RPM builder fixture

viniciusferrao:
`xCAT-test/unit/apache_config_sources.t` fails on master, so every pull request
that merges with it reports a red unit suite. `stage_makerpm_fixture` copies
`makerpm` into a partial tree but stages only
`build-utils/sync-xcat-apache-configs`. `makerpm` also sources
`build-utils/source-only.sh`, so the staged copy dies before it configures the
build mode and never reaches `rpmbuild`. The test arrived before `makerpm`
sourced that helper, and the helper arrived without the fixture that stages it,
so each change passed on its own branch.

Stage the helper alongside the one already staged. It is sourced rather than
executed, so it needs no mode change, and `makerpm` reads no other file from
`build-utils`.


viniciusferrao:
Duplicate of #7805, which makes the same change. Closing in favour of that one.

PR #7811, opened 2026-09-02
fix(xcatd): refuse an XML request that carries a document type declaration

viniciusferrao:
No request that xCAT sends carries a document type declaration, but the daemon
accepted one and expanded the entities it declared, so a short request could
grow into a large document and consume the memory and the time of the daemon.

Refuse the declaration. The parser option that stops entity expansion does not
cover an entity named inside an attribute, so it leaves the same growth
available by another route.

### DIFF
--- a/xCAT-test/unit/xml_doctype_refused.t
+++ b/xCAT-test/unit/xml_doctype_refused.t
@@ -1,0 +1,93 @@
+#!/usr/bin/env perl
+use strict;
+use warnings;
+
+use FindBin;
+use Test::More;
+
+my $libdir = "$FindBin::Bin/../../xCAT-server/lib/perl";
+my $xmlpm  = "$libdir/xCAT/XML.pm";
+plan skip_all => 'xCAT::XML not found' unless -r $xmlpm;
+eval { require XML::Simple; require XML::Parser; 1 }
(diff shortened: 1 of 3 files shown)


xCAT-test/unit/xml_doctype_refused.t
```diff
@@ -0,0 +1,112 @@
...
+ok(scalar(@doctype) >= 1, 'the module refuses a document type declaration');
+my @extern = $source =~ /(ExternEnt\s*=>\s*sub)/g;
+ok(scalar(@extern) >= 1, 'the module still refuses an external entity');
+
+# Both entry points must build their parser in the one place that carries the
+# refusals, so the two versions of XML::Simple this module serves cannot drift
+# apart.
+like($source, qr/sub _hardened_parser/, 'the module builds its parser in one routine');
+my @callers = $source =~ /(_hardened_parser\(\))/g;
+is(scalar(@callers), 2, 'both entry points call that routine');
```
dhilst: This is testing the source code, the above tests already check for the intended behavior, this test can be removed

dhilst (changes requested):

dhilst (approved):

PR #7814, opened 2026-09-02
refactor(confluent): share first-row attribute flattening

viniciusferrao:
Reuse one helper for the identical nodepos and mp result-shaping paths used by explicit-node requests. Keep nodehm missing-row recovery and switch multi-row handling separate because they have different contracts.

Replace the remaining source-based confluent test with real plugin execution. The table double now matches xCAT::Table’s missing-row return shape and covers explicit-node selection, conserver routing, and location and enclosure export.


xCAT-server/lib/xcat/plugins/confluent.pm
```diff
@@ -194,6 +194,17 @@ sub process_request {
...
 
+sub _first_node_attribute_rows {
+    my $entries = shift;
+    my @rows;
+    foreach my $entry (@{$entries}) {
+        foreach my $node (keys %{$entry}) {
+            push @rows, $entry->{$node}->[0];
+        }
+    }
+    return @rows;
```
dhilst, line 205: I would return an array reference here

dhilst (approved):

PR #7818, opened 2026-09-04
ci(xcat-core): Fixes to get CI running for all 2.19 targets 

dhilst:
Fixes for the problems regarding RC1 of 2.19 captured by the CI


viniciusferrao (changes requested):
Hi @dhilst, two points here:

1. The `ppc64el` naming cleanup makes sense given xCAT’s historical
   Power naming. Could we also add replacement/conflict metadata for
   the old `xcat-genesis-base-ppc64` package, as we do for the scripts
   package, so existing installations can upgrade cleanly? Keeping
   the legacy `genesis/ppc64/` directory is fine; this is only about
   the package-name transition.

2. In `verify-genesis-payload`, `grep -v '^[/-]'` also drops absolute
   command paths, so missing binaries such as `/usr/bin/awk` can pass
   verification. Could we check required absolute paths relative to
   the payload root as well, and add a test for a missing one?

Thanks!

viniciusferrao (approved):
LGTM

PR #7820, opened 2026-09-07
Make ospkgs and detect_dhcpd work on current Ubuntu releases

viniciusferrao:
The shared Ubuntu package lists named packages that current releases no
longer carry. ospkgs hands the whole list to one apt-get install, so one
unknown name failed the transaction and updatenode -S installed nothing from
the list. Both detect_dhcpd copies refused to run unless /usr/sbin/tcpdump
existed. Debian and Ubuntu install it as /usr/bin/tcpdump.

The shared lists keep ntp and ntpdate, so releases up to 24.04 keep the
daemon they had. Only names gone from every current release change, and
26.04, which dropped ntp, gets release lists of its own.

| List | Removed | Replacement |
| --- | --- | --- |
| service, service.ubuntu26.04 | libodbc1 | unixodbc |
| kvm, kvm.ubuntu26.04 | libvirt-bin | libvirt-daemon-system, libvirt-clients |
| kvm.x86_64, kvm.ubuntu26.04.x86_64 | qemu-kvm | qemu-system-x86 |
| kvm.ppc64el, kvm.ppc64le, and the 26.04 pair | qemu-kvm | qemu-system-ppc |
| kvm.ubuntu24.04.riscv64 | qemu-kvm | qemu-system-misc |
| kvm, kvm.ubuntu26.04 (no list for the architecture) | qemu-kvm | qemu-system |
| compute.ubuntu26.04, service.ubuntu26.04, kvm.ubuntu26.04, netboot compute.ubuntu26.04 | ntp, ntpdate | chrony |

The service lists also gain libdbd-pg-perl beside libdbd-mysql-perl, as a
service node may run the xCAT database on PostgreSQL. Every current kvm list
names qemu-utils: ospkgs installs without recommends, and libvirt needs
qemu-img for the qcow2 volumes kvm.pm creates. 12.04, 14.04 and 16.04 keep
kvm lists with the old names, and the ppc64le kvm lists are links to the
ppc64el ones, as the other lists do. On 26.04 riscv64 libvirt depends on
qemu-kvm or qemu-system and nothing provides qemu-kvm, so apt installs
qemu-system there whatever the list names. detect_dhcpd resolves tcpdump
through PATH and the standard directories and runs the resolved path.

On Subiquity releases the install list reaches a node only through ospkgs.
The autoinstall package block is fixed in the template.

### DIFF
--- a/xCAT-server/share/xcat/install/ubuntu/kvm.ppc64le.pkglist
+++ b/xCAT-server/share/xcat/install/ubuntu/kvm.ppc64le.pkglist
@@ -1,0 +1,1 @@
+kvm.ppc64el.pkglist
--- a/xCAT-server/share/xcat/install/ubuntu/service.ubuntu26.04.pkglist
+++ b/xCAT-server/share/xcat/install/ubuntu/service.ubuntu26.04.pkglist
@@ -1,5 +1,5 @@
 openssh-server
-ntp
+chrony
 gawk
 nfs-common
 snmpd
--- a/xCAT-probe/subcmds/detect_dhcpd
+++ b/xCAT-probe/subcmds/detect_dhcpd
@@ -5,6 +5,7 @@ BEGIN { $::XCATROOT = $ENV{'XCATROOT'} ? $ENV{'XCATROOT'} : -d '/opt/xcat' ? '/o
 
 use lib "$::XCATROOT/probe/lib/perl";
 use probe_utils;
+use xCAT::CommandUtils;
 use File::Basename;
 use IO::Socket::INET;
 use Time::HiRes qw(gettimeofday sleep);
@@ -63,7 +64,8 @@ if ($::TEST) {
     exit 0;
 }
 
-unless (-x "/usr/sbin/tcpdump") {
+my $tcpdump = xCAT::CommandUtils::find_executable('tcpdump');
+unless ($tcpdump) {
     probe_utils->send_msg("$output", "f", "Tool 'tcpdump' is installed on current server");
     probe_utils->send_msg("$output", "d", "$program_name needs to leverage 'tcpdump', please install 'tcpdump' first");
     exit 1;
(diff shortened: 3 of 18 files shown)


xCAT-server/share/xcat/install/ubuntu/compute.pkglist
```diff
@@ -1,6 +1,5 @@
 openssh-server
-ntp
-ntpdate
+chrony
```
viniciusferrao: @dhilst check if this does not broke you `makentp` work.

xCAT-server/share/xcat/install/ubuntu/kvm.ppc64le.pkglist
dhilst, line 1: Shouldn't we make this a symlink to the above? Or remove it since `le` is not the correct architecture for Ubuntu anyway

xCAT-server/share/xcat/install/ubuntu/service.ubuntu26.04.pkglist
```diff
@@ -0,0 +1,7 @@
+openssh-server
+chrony
+gawk
+nfs-common
+snmpd
+libdbd-mysql-perl
```
dhilst, line 6: Hmmm, I see a problem here, xCAT supports postgres for service nodes, add postgres drivers in this list too so we cover both

dhilst (changes requested):

xCAT-server/share/xcat/install/ubuntu/compute.pkglist
dhilst: I see this was reverted. It will not break makentp work but it will fail to install on old Ubuntus, as they have no chrony and this one should be used for fallback when there is no versioned template. `chrony` is used for Ubuntu >= 18.04

dhilst (approved):

PR #7773, opened 2026-08-25
feat(xcat-core): support riscv64 nodes on EL10

viniciusferrao:
This PR adds initial support for the RISC-V architecture.

riscv64 nodes boot through UEFI and grub2 only. Kea and ISC answer RISC-V UEFI clients (client architecture 27, and 28 for HTTP boot) with `boot/grub2/grub2.riscv64`, `mknb` writes one grub2 discovery configuration per network, and discovered riscv64 nodes default to `netboot=grub2`.

EL10 riscv64 install and netboot profiles are added, with the installer media paths for the family.

`copycds` publishes `EFI/BOOT/grubriscv64.efi` from riscv64 media as `/tftpboot/boot/grub2/grub2.riscv64`, so a node can net boot without a manual copy. Media of other architectures are untouched.

PR #7821, opened 2026-09-07
feat(ubuntu): support riscv64 management and provisioned nodes on 24.04 and 26.04

viniciusferrao:
This PR adds support for Ubuntu as management node and as provisioned node on the RISC-V architecture, on 24.04 and 26.04.

copycds recognizes riscv64 live-server media and builds the netboot grub2
loader from the package on the media. builddebs.pl builds the management node
packages for riscv64 and go-xcat installs them. Netboot images get the riscv64
network drivers and resolver libraries.

There are also fixes for other architectures:

- The apt mirror follows the image architecture.
- Kernel command lines are escaped for grub2.
- The x86 token maps to i386 for debootstrap.
- xcat-server declares grub-common.

Follows #7773
Depends on xcat2/xcat-dep#67


dhilst (changes requested):
The PR still treats riscv64 as unknown in debian.pm, so every diskful RISC-V install logs a false “Unknown arch” message. The network- driver test never runs normally: imgutils cannot find xCAT::SvrUtils, and the error becomes skip_all, leaving CI green without coverage. lookup_netboot_arch.t loads xCAT::Schema from /opt/xcat before selecting the checkout, making results depend on the installed xCAT version. The grub-mkimage dependency test greps Perl source instead of executing behavior, so it can pass against dead code. The commits lack Signed-off-by trailers.

dhilst (approved):

PR #7823, opened 2026-09-08
test(xcat-core): Introduce BATS & convert shell scripting tests to it

dhilst:
This PR introduces [BATS](https://github.com/bats-core/bats-core) for testing shell scripts and convert some tests to it. This should make shell tests easier to write and read. It follow the same spirit of the perl .t tests


viniciusferrao (changes requested):
Hi @dhilst, a few things to fix before merging:

1. **Some `! grep` assertions don't fail the test.**

   In `sles_pre_script.bats`, the checks for leftover `<location>` and `<lba_support>` elements can fail without failing the test, because they aren't the last command. For example, leaving `<lba_support>` in the XML still passes as long as `<linear>` is removed. The bare-signal check in `remoteshell_restart.bats` has the same problem.

   Could we use `run -1 grep ...` for these? That explicitly requires grep to return “no match,” rather than relying on `!` and `set -e`.

2. **The diskless download test still writes to the host.**

   `capture_xcatdsklspost_wget()` mocks `wget`, but `download_postscripts()` still redirects stderr to `/tmp/wget.log`. The shell opens and truncates that file before calling the mock, so a test run can overwrite an existing log or fail if another user owns it.

   Source-only mode can also load the installed `/xcatpost/xcatlib.sh` and read host configuration before reaching the return. Sourcing the script replaces the test's `echolog()` mock too.

   Please make the download log path overridable and put it under `BATS_TEST_TMPDIR`. Source-only loading should skip host initialization, and the logging mock should be defined after sourcing.

3. **Check that the SSH fallback actually waits before restarting.**

   `remoteshell_restart.bats` checks the kill and start in separate logs. Those assertions still pass if the call to `xcat_wait_for_processes_to_exit` is removed, or if SSH starts before the kill. Testing the wait helper separately doesn't catch either change in the caller.

   Could we record kill, polling, and start in one event log and assert their order? Having the mocked process remain alive for at least one poll would also check that the fallback really waits.


Thanks.

viniciusferrao (approved):
LGTM.

PR #7824, opened 2026-09-08
doc(xcat-core): specify distro versions in support matrix

dhilst:
The operating-system support table named distro families without defining which releases the claim covered, and it kept separate legacy CentOS and Windows rows that are outside the xCAT 2.19 CI target contract. This made the release matrix ambiguous and impossible to translate into a finite CI backlog.


docs/source/overview/support_matrix.rst
```diff
@@ -1,24 +1,29 @@
...
++--------+---------+-------+-------+-----+-------+--------+--------+--------+----------+
+| Ubuntu | 24.04   | no    | yes   | no  | yes   | yes    | yes    | yes    | no       |
++--------+---------+-------+-------+-----+-------+--------+--------+--------+----------+
+| Ubuntu | 26.04   | no    | yes   | no  | yes   | yes    | yes    | yes    | no       |
++--------+---------+-------+-------+-----+-------+--------+--------+--------+----------+
 
-riscv64 support covers EL10 compute nodes (Rocky Linux 10 and the RHEL 10 RISC-V developer preview) that boot through UEFI firmware and grub2. See :doc:`/guides/admin-guides/manage_clusters/riscv64/index` for details.
+.. note::
+
+   EL: Enterprise Linux, such as: RHEL, Rocky Linux, Alma Linux, CentOS and Oracle Linux.
```
viniciusferrao:
```suggestion
   EL stands for Enterprise Linux, such as: Red Hat Enterprise Linux (RHEL), Rocky Linux, AlmaLinux, CentOS and Oracle Linux.
```

docs/source/overview/support_matrix.rst
```diff
@@ -1,24 +1,29 @@
...
+| EL     | 9       | yes   | yes   | yes | yes   | yes    | yes    | yes    | no       |
++--------+---------+-------+-------+-----+-------+--------+--------+--------+----------+
+| EL     | 10      | yes   | yes   | yes | yes   | yes    | yes    | yes    | yes      |
++--------+---------+-------+-------+-----+-------+--------+--------+--------+----------+
+| SLES   | 12      | yes   | yes   | yes | yes   | yes    | yes    | yes    | no       |
++--------+---------+-------+-------+-----+-------+--------+--------+--------+----------+
+| SLES   | 15      | yes   | yes   | yes | yes   | yes    | yes    | yes    | no       |
++--------+---------+-------+-------+-----+-------+--------+--------+--------+----------+
+| Ubuntu | 20.04   | no    | yes   | no  | yes   | yes    | yes    | yes    | no       |
++--------+---------+-------+-------+-----+-------+--------+--------+--------+----------+
```
viniciusferrao:
```suggestion
```

20.04 is EoL we don't have to test.

viniciusferrao (changes requested):
There is also the Power mess, zVM and other details but I don't want to tackle that now.

We don't have `s390x` to test, which is zVM. Power is a mess because it's not well written: `ppc`, `ppc64` or `ppc64le`?

I think we can work on that for 2.20 and leave this as is for now.

dhilst:
> There is also the Power mess, zVM and other details but I don't want to tackle that now.
> 
> We don't have `s390x` to test, which is zVM. Power is a mess because it's not well written: `ppc`, `ppc64` or `ppc64le`?
> 
> I think we can work on that for 2.20 and leave this as is for now.

I suggest we keep it simple and left only the distro vs architecture. So x86_64, ppc64le & riscv64

PR #7829, opened 2026-09-11
feat(ubuntu): install the osimage pkglist during the Subiquity autoinstall

viniciusferrao:
This PR adds support for installing the osimage pkglist during a Subiquity
autoinstall. compute.subiquity.tmpl ends its packages list with
#INCLUDE_DEFAULT_PKGLIST_AUTOINSTALL#, and Template.pm replaces that line with
one item per pkglist package, includes followed. Plain names and tasks install
this way, without recommended packages. A version pin, a target release, an
architecture qualifier, a removal or group record, a #ENV: setting and an
osimage with environvar stay with ospkgs, which still applies the whole list
after the first boot.

The installer's apt sources also gain the osimage pkgdir mirrors, next to the
otherpkgs repository. A repeated Ubuntu archive mirror carries the signing key
the installer gives it, and a repository named twice is offered once. The
installer removes these sources from the target at the end of the install.

There are also fixes for the Subiquity releases:

- The otherpkgs source is a Deb822 stanza on 24.04 and later, since curtin
  drops trusted=yes when it converts a one-line source.
- Ubuntu 20.04, 22.04 and 24.04 get default compute pkglists with chrony,
  because ntp pulls ntpsec, which conflicts with the chrony the template
  installs.

### DIFF
--- a/xCAT-server/lib/perl/xCAT/Template.pm
+++ b/xCAT-server/lib/perl/xCAT/Template.pm
@@ -173,6 +173,11 @@ sub subvars {
         $inc =~ s/#INCLUDE_DEFAULT_RMPKGLIST_S#/#INCLUDE_RMPKGLIST:$pkglistfile#/g;
     }
 
+    # osimage environvar reaches apt-get through ospkgs alone, so such an image installs its list there,
+    # and its pkgdir mirrors, which may need those variables too, stay out of the installer's sources
+    my $environvar_set = ( $namedargs{environvar} // '' ) =~ /\S/;
+    my $installer_pkgdirs = $environvar_set ? undef : $namedargs{pkgdirs};
+    my @autoinstall;
     if (("ubuntu" eq $platform) || ("debian" eq $platform)) {
 
         # since debian/ubuntu uses a preseed file instead of a kickstart file, pkglist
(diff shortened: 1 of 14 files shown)


xCAT-server/lib/perl/xCAT/Template.pm
```diff
@@ -1786,13 +1847,30 @@ sub ubuntu_subiquity_apt_mirror
     return ($ent && defined($ent->{value}) && length($ent->{value})) ? $ent->{value} : $default;
 }
 
+# The key curtin names in the Deb822 sources it writes for the primary and the security mirror on
+# 24.04 and later, and the default mirrors it writes them for.
+my $UBUNTU_ARCHIVE_KEYRING = '/usr/share/keyrings/ubuntu-archive-keyring.gpg';
+my @UBUNTU_ARCHIVE_MIRRORS = qw(http://archive.ubuntu.com/ubuntu http://security.ubuntu.com/ubuntu http://ports.ubuntu.com/ubuntu-ports);
```
dhilst: I think there is an apt_mirror in the site table, or something like that, can you doublecheck plz?

xCAT-test/unit/debian_mkinstall_pkglist_handover.t
dhilst, line 1: Avoid BAIL_OUT as I stop all the test suite, use "die" instead. Also this test is grepping the source code, test the code for intended behavior

dhilst (changes requested):
I left inline comments

xCAT-server/lib/perl/xCAT/Template.pm
viniciusferrao: Fixed. Removed the hardcoded list; site.ubuntu_apt_mirror is used as the primary.

xCAT-test/unit/debian_mkinstall_pkglist_handover.t
viniciusferrao: Removed the test.

dhilst (approved):

PR #7830, opened 2026-09-11
test(xcat-core): Add client/server tests to xcat criticial points

dhilst:
Add client/server tests for validating critical points in the CI, including:

- DHCP
- DNS
- HTTP
- TFTP
- xCAT-server


viniciusferrao (changes requested):
Hi @dhilst two small points:

1. In `kea_node_reservations()`, could we preserve the node’s configured
   hostname? Replacing the reserved `hostname` with `host-name` in
   `option-data` lets a hostname or FQDN supplied by the client take
   precedence, so the lease and DNS updates can use the wrong name.
   A test with a client advertising a different name would catch this.

2. In `httpc.py`, `_decode()` reports success based only on the HTTP
   status, even when curl fails or times out while downloading the
   body. Could we also require curl to finish successfully and add
   a truncated-response test? Otherwise, checks for HTTP 200 and
   a nonempty body can pass on an incomplete download.

Thanks!

PR #7831, opened 2026-09-14
fix(xcat-core): nodeset cannot boot an Ubuntu POWER install from live media

dhilst:
`nodeset` stops the diskful install on every ppc64le cell of `xcat-core-devel-ubuntu-cd`, with
`The network boot initrd.gz is not found in /install/ubuntu24.04.4/ppc64el/install/netboot`. The
Ubuntu ppc64el live-server ISO carries no netboot tree, and the installer kernel and initrd sit
under `casper`. In `debian.pm`, `%INSTALL_BOOT_FILES` describes no casper layout for POWER, and
`mkinstall` applies a second POWER-only precondition before it calls the resolver. The POWER table
gains the casper entries, and `install_media_is_bootable` replaces the precondition.
`debian_install_boot_files.t` fails without the change.


viniciusferrao (approved):
LGTM

PR #7832, opened 2026-09-14
fix(xcat-core): the Ubuntu Genesis image ships without a DHCP client

dhilst:
`builddeb-genesis-base` names no `isc-dhcp-client` and no `ifenslave`, so `dracut_install` reports each missing command, returns 0, and the deb ships the holes — `dhclient` is absent from the payload on all four Ubuntu management nodes, and without it Genesis gets no lease. `hwclock` is different: focal and jammy have it in `util-linux` and no `util-linux-extra` to install, so naming that package unconditionally stops the build. The branch adds the two packages unconditionally and passes `util-linux-extra` through `optional_packages()`, which keeps a package only where apt has a candidate. `genesis_ubuntu_build_root.t` runs the module and the selector.

### DIFF
--- a/xCAT-genesis-builder/builddeb-genesis-base
+++ b/xCAT-genesis-builder/builddeb-genesis-base
@@ -30,6 +30,7 @@ REQUIRED_PACKAGES="
     nfs-common rpcbind pciutils usbutils parted
     dosfstools e2fsprogs lvm2 mdadm net-tools
     bc psmisc rsync wget cpio
+    isc-dhcp-client ifenslave util-linux-extra
     dpkg-dev debhelper fakeroot devscripts vim-tiny
 "
 if [ "$BUILDARCH" = "amd64" ]; then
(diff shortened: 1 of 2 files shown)


viniciusferrao (changes requested):
Thanks @dhilst, two things to fix before merging:

1. Could we make `util-linux-extra` conditional on the Ubuntu release?
   Ubuntu 22.04 doesn't provide that package; `hwclock` is still in
   `util-linux` there. Adding it unconditionally makes dependency
   installation fail on a supported build target. The test should
   cover both package layouts rather than always require
   `util-linux-extra`.

2. `verify-genesis-payload` comes from #7818 and isn't in this branch.
   Could we include it here, or make that dependency explicit and
   ensure #7818 lands first? Otherwise the new call stops the build
   before packaging. A test that actually invokes the verifier would
   catch this; the current assertion only checks that its name
   appears in the build script.

Thanks!

dhilst:
Both done. You're right on the first — and it's `Suggests`, not `Recommends`, so nothing pulls it in automatically. Measured on the four Ubuntu MNs: focal and jammy report `Candidate: (none)` and carry `hwclock` in `util-linux`; noble and resolute have it in `util-linux-extra`. So instead of a codename table, `optional_packages()` runs after `apt-get update` and keeps a package only where `apt-cache show` finds a candidate. On the second, the `verify-genesis-payload` call is gone rather than sequenced: **this PR no longer depends on #7818 or #7835 and can merge on its own.** The source-grep assertion went with it.

viniciusferrao (approved):
LGTM

PR #7833, opened 2026-09-14
fix(xcat-core): Remove legacy xCAT-genesis-builder

dhilst:
The `xCAT-genesis-builder` directory holds the Genesis image sources and, beside them, a builder
that no pipeline runs: `buildrpm`, `debuild-xcat-genesis-base`, `install`, `check`, the spec and
the root `makerpm-genesisbuilder`. `build-ubunturepo` has no caller either, and its `-r` flag has
set an unread variable since 2013. The documentation tells the reader to install an rpm no build
produces. The branch deletes the dead files and renames the directory `xCAT-genesis-base` for what
it holds. `build_utils.t` drives the probe helper staging that replaces `build-ubunturepo`.


viniciusferrao (changes requested):
Thanks @dhilst, but there are some questions:

1. Could we keep the OpenEmbedded section of the README under
   `xCAT-genesis-base/README.md`, with the paths updated? It documents
   the build, export, and signed-extension workflows that remain in
   the tree. Only the legacy builder instructions need removing.

2. Could we leave `xCAT-genesis-builder` in `GO_XCAT_UNINSTALL_LIST`?
   Older installations may still have that RPM installed, and removing
   it from this list means `go-xcat uninstall` no longer asks the
   package manager to remove it. Keeping the uninstall entry doesn't
   require keeping the old builder.

3. Should we remove all of that already? Does that affects the transtion to OpenEmbedded that we will be pushing?

Thanks!

dhilst:
> Thanks @dhilst, but there are some questions:
> 
>     1. Could we keep the OpenEmbedded section of the README under
>        `xCAT-genesis-base/README.md`, with the paths updated? It documents
>        the build, export, and signed-extension workflows that remain in
>        the tree. Only the legacy builder instructions need removing.
> 
>     2. Could we leave `xCAT-genesis-builder` in `GO_XCAT_UNINSTALL_LIST`?
>        Older installations may still have that RPM installed, and removing
>        it from this list means `go-xcat uninstall` no longer asks the
>        package manager to remove it. Keeping the uninstall entry doesn't
>        require keeping the old builder.
> 
>     3. Should we remove all of that already? Does that affects the transtion to OpenEmbedded that we will be pushing?
> 
> 
> Thanks!

Hi,

1 & 2. Yes, sure

3. The builder is not being used, the genesis base is built inside the .spec in el now for example so keeping two build methods are error prone

PR #7834, opened 2026-09-14
fix(xcat-core): the Ubuntu ppc64le diskful install stops without grub-ieee1275

dhilst:
A diskful install of Ubuntu on a ppc64el node runs to the bootloader step and stops. curtin reports
`dpkg-query: package 'grub-ieee1275' is not installed`, and the node never reaches the disk it
installed. `ubuntu_subiquity_apt_config` in `Template.pm` writes `APT::Install-Recommends "false"`
for every architecture, and on a PReP machine `grub-ieee1275` reaches the target only as a
recommended package of the kernel image. The apt block now keeps the recommended packages on
ppc64el. `ubuntu_subiquity_pkglist.t` fails both of its assertions without the change.

### DIFF
--- a/xCAT-server/lib/perl/xCAT/Template.pm
+++ b/xCAT-server/lib/perl/xCAT/Template.pm
@@ -1850,6 +1850,19 @@ sub ubuntu_subiquity_apt_mirror
 # The key curtin names in the Deb822 source it writes for the primary apt mirror on 24.04 and later.
 my $UBUNTU_ARCHIVE_KEYRING = '/usr/share/keyrings/ubuntu-archive-keyring.gpg';
 
+# The apt lines that keep the recommended packages out of the install. ospkgs installs without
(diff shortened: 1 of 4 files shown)


viniciusferrao (changes requested):
There's one issue with the new BATS thing.

Could we make the archive path overridable, like `XCAT_ERROR_CONSOLE`,
and point it at `BATS_TEST_TMPDIR`? Mocking `tar` doesn't prevent the
shell from opening `/run/testnode-logs.tar`. An unprivileged run passes
despite that write failing, while a root run can truncate an existing
file.

Having the tar mock emit a marker would also let the test check that
the output reaches the intended file.

Do you agree?

viniciusferrao:
@dhilst I would drop the final
`[ ! -e /run/testnode-logs.tar ]` assertion. That file might already
exist before the test, so its presence doesn't mean the test touched
it.

viniciusferrao (approved):
LGTM.

PR #7835, opened 2026-09-14
fix(xcat-core): the Ubuntu Genesis image is built from the build host kernel

dhilst:
dracut copies the kernel, the modules and every command out of the root it runs in. The Ubuntu
Genesis deb was built once on the build host, so every Ubuntu release received the build host
kernel, and the pipelines converted the EL rpm with `alien` instead. `builddebs.pl` gains
`--genesis`, `--genesis-only` and `--genesis-dist`, and builds the deb once per codename inside
that codename's sbuild chroot. `builddeb-genesis-base` takes `--expect-codename` and stops in the
wrong root. `genesis_log_errors` fails the build on an error line a zero exit status hides.

### DIFF
--- a/build-utils/lib/XCAT/BuildUtils.pm
+++ b/build-utils/lib/XCAT/BuildUtils.pm
@@ -37,7 +37,7 @@ our @EXPORT_OK = qw(
     buildinfo_text
     targetarch_from_target
     genesis_chroot_name genesis_target_arch genesis_build_plan
-    genesis_log_errors genesis_log_deny_rules
+    genesis_log_errors genesis_log_deny_rules deb_belongs_to_dist
 );
 
 # Both builders echo the commands they run under --verbose.  Set once, after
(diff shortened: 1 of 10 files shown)


viniciusferrao (changes requested):
Daniel, some details here:

1. Could we limit `deb_belongs_to_dist()` to the Genesis packages?
   It currently treats any trailing `~...` in a package version as
   a codename. With `--release 1~rc1`, ordinary xCAT packages are
   silently excluded from every suite, even without `--genesis`.
   A non-Genesis prerelease case would help cover this.

2. The default Genesis plan includes `focal`, but the package requires
   `debhelper-compat (= 13)` and a stock Focal chroot provides compat
   12. Could we either arrange the newer build dependency there or
   leave Focal out of the Genesis defaults? Otherwise a plain
   `--genesis` run stops on its first release.

viniciusferrao (approved):
LGTM.

PR #7836, opened 2026-09-14
fix(xcat-core): a unit test reads the installed xCAT instead of its own tree

dhilst:
Fifty-nine unit tests load xCAT modules from `/opt/xcat` instead of from the checkout, because
every xCAT module runs `use lib $::XCATROOT/lib/perl` while it compiles and `$::XCATROOT` falls
back to `/opt/xcat`. On a build agent that carries xCAT, the suite measures the installed packages.
`xCAT-test/lib/XCAT/Test/XcatRoot.pm` builds a scratch directory with the installed layout, filled
with symlinks into the checkout, and points `XCATROOT` at it. `unit_tests_read_the_checkout.t` runs
every other unit test against a decoy tree and fails without the fix.


dhilst:
This PR surfaces a structural problem in the tests:

- Some tests depend on uninstalled files (.spec, buildrpms.pl, etc)
- Most xCAT modules use `use lib $::XCATROOT/lib/perl` at compile time and $::XCATROOT falls back to /opt/xcat

As result running unit tests from the source tree exercise the files in /opt/xcat instead from the source tree, and we can't run all the tests from the /opt/xcat because some files are not installed (as stated at 1). I'm thinking about it

viniciusferrao (changes requested):
There is one issue in `XCAT::Test::Source`:

Could we avoid treating every initial `@INC` directory as trusted for
xCAT modules? That includes paths inherited through `PERL5LIB` or `-I`.
If an xCAT module is missing from the current checkout, Perl can load
it from an older checkout on that path, and the exit guard still
reports success.

Keeping those paths available for third-party dependencies is fine,
but xCAT modules should come from this checkout or an explicit test
stub. A regression test with a decoy xCAT module in an inherited
`PERL5LIB` directory would catch this.

That would make more sense.

dhilst:
I'm closing this until I find a better approach to this problem. I opened it because testing jumping to /opt bite me multipletimes but the current solution is just as bad or worse than the problem

PR #7837, opened 2026-09-14
fix(xcat-core): the Genesis deb version never advances past snap000000000000

dhilst:
Every Genesis deb the Ubuntu dependency pipeline publishes is versioned `2.19.0-snap000000000000`,
so apt never upgrades an installed `xcat-genesis-base` and the channel cannot ship a Genesis fix.
`Release` is a tracked placeholder that `buildrpms.pl` overwrites with the commit time, and the
dependency job clones xcat-core only to build the Genesis debs, so `buildrpms.pl` never runs there
and `builddebs.pl` reads the placeholder as authoritative. `builddebs.pl` now treats an all-zero
`Release` as unstamped and falls back to the commit time. `builddebs_release_placeholder.t` fails
on the unfixed tree.


viniciusferrao (approved):
LGTM

PR #7838, opened 2026-09-14
fix(xcat-core): the diskless flat case clears the machine type on riscv64

dhilst:
`reg_linux_diskless_installation_flat` fails on an EL 10 riscv64 compute node after the node
provisions correctly. The check `vmothersetting contains machine` fails and `lsdef` shows the
attribute empty. The case corrupts the KVM machine type, checks that the node does not boot, and
restores it, but the restore reads `str3` from a ladder that names ppc64 and x86_64 only. On
riscv64 `str3` is empty, `chdef` writes an empty `vmothersetting`, and the later comparison ends
with `unary operator expected`. Both ladders gain a riscv64 arm.
`diskless_flat_vmothersetting_machine.t` fails five of fifteen assertions without the change.


viniciusferrao (changes requested):
@dhilst do you think that it's a good idea the BATS test also check what
`chdef` receives after cleanup? With only `machine:virt`, it should
write an empty value; with other settings present, those should stay.
Currently it only checks for the absence of `unary operator expected`,
so a cleanup that leaves the machine setting behind still passes.

If don't just dismiss my review and merge it.

viniciusferrao (approved):
LGTM

PR #7839, opened 2026-09-14
fix(xcat-core): a compute node cannot resolve the management node by short name

dhilst:
On Ubuntu 26.04 the installed compute node keeps systemd-resolved, which takes the DNS server from
DHCP and no search domain, so `/etc/resolv.conf` reads `search .`. `mypostscript` then runs
`updateflag.awk $MASTER 3002` with the short management node name, the flag update fails eleven
times, and the node stays at postbooting. The management node does offer domain-search; the node
discards it. `compute.subiquity.tmpl` writes a netplan with `dhcp4: true` alone, so
systemd-networkd applies its `UseDomains` default of no. Both branches gain `dhcp4-overrides`.
`ubuntu_subiquity_installnic.t` fails without the change.


viniciusferrao (approved):
LGTM

PR #7840, opened 2026-09-14
fix(xcat-core): gettimezone returns an error sentence as a timezone name

dhilst:
The EL 10 riscv64 compute node installs no package. Its kickstart carries
`timezone Could not determine timezone checksum --utc`, anaconda answers
`One or zero arguments are expected for the timezone command`, and the node holds
`status=installing`. `gettimezone` in `Utils.pm` compares `/etc/localtime` against every file under
`/usr/share/zoneinfo` and returns the string `Could not determine timezone checksum` when that
pipeline exits non-zero. The Rocky 10.2 riscv64 cloud image ships no `/etc/localtime`.
`gettimezone` now reads the symlink first, then the scan, then `/etc/timezone`, and returns UTC.
`utils_gettimezone.t` asserts each source separately.


viniciusferrao (approved):
LGTM

PR #7841, opened 2026-09-14
fix(xcat-core): the Ubuntu POWER diskful install never leaves the installer

dhilst:
`reg_linux_diskfull_installation_flat` fails on `ubuntu-22-ppc64le-devel` and
`ubuntu-24-ppc64le-devel`. The Subiquity installer starts, errors in its early-commands and
reboots, nine times in 75 minutes, and the case ends on `Permission denied (publickey,password)`.
`mkinstall` in `debian.pm` selected `pre.ubuntu.subiquity` and then replaced it with
`pre.ubuntu.ppc64` for every ppc64 node, and that script writes a partman recipe Subiquity cannot
parse. `install_prescript` now decides from the installer and the architecture together, and
`pre.ubuntu.subiquity` gains a PReP branch. `debian_install_prescript.t` covers it. The branch
stacks on `fix/ci-ubuntu-ppc-boot-files`, so open it after that pull request merges.


viniciusferrao (dismissed):
LGTM

viniciusferrao (approved):

PR #7842, opened 2026-09-14
fix(xcat-core): the Ubuntu ppc64el netboot image has no dig

dhilst:
`genimage` returns 2 with `Failed to find usr/bin/dig` for ubuntu22.04 and ubuntu24.04 on ppc64el,
so `packimage` writes no `initrd.gz` and the diskless compute node never boots.
`xCAT-server/share/xcat/netboot/ubuntu` carried a ppc64el package list for 20.04 and 26.04 only,
and `imgutils::get_profile_def_filename` falls back to `compute.pkglist`, which installs no dig, no
cpio and no chrony. The branch adds the 22.04 and 24.04 ppc64el lists, with the ppc64le spelling
each release already carries. `ubuntu_ppc64el_pkglists.t` fails without the files.


viniciusferrao (approved):
LGTM

PR #7843, opened 2026-09-14
fix(xcat-core): grub2 cannot load the Ubuntu 26.04 ppc64el netboot initrd

dhilst:
grub2 fails to load the Ubuntu 26.04 ppc64el netboot initrd, and the compute node never reaches the kernel. `netboot/ubuntu/genimage` copied the whole `lib/firmware` tree into the initrd, which took the payload to 783 MB of kernel plus initrd; grub2 stops somewhere above 752 MB. The firmware step now copies only the firmware that `modinfo -F firmware` names for the modules already going into the initrd, resolving Ubuntu's compressed file names. `ubuntu_genimage_initrd_firmware.t` extracts that step and reads the initrd it writes. Two of its assertions fail against the unfixed genimage.


viniciusferrao (changes requested):
Daniel, I think this one is not ready. It's firmware related, so:


1. Could firmware discovery inspect the same module files the copy
   step uses? That step checks `$customdir` and `$pathtofiles`, but
   `initrd_firmware_files()` only checks `$rootimg_dir`. A custom
   driver can therefore reach the initrd without its required firmware,
   even when that firmware exists in the image. Reusing the resolved
   source, or inspecting the staged module, would avoid this.

2. Could we preserve the kernel’s firmware lookup directories?
   The helper misses `updates/`, `updates/<kernel>/`, and `<kernel>/`
   beneath `lib/firmware`. The previous copy retained those files;
   now an override can disappear, leaving either the base firmware
   or no matching firmware at all.

We aren't thinking about custom-driver case and firmware overrides.

They are rare, but happens constantly.

If you didn't understand what I mean let me know.

viniciusferrao (approved):
LGTM.

Issue #7848, opened 2026-09-14
otherpkgs: zypper branch refreshes/removes the wrong repo name when http OTHERPKGDIR entries are present


ChrisPJames:
Summary

When otherpkgs adds a local package directory as a repo, the repo id is built from localrepoindex, which is offset by the number of http repos already added in this pass:

sh
```
                index=$(array_get_size repo_path)
                localrepoindex=$[urlrepoindex+index]
                if [ $hasyum -eq 1 ] || [ $haszypper -eq 1 ] ; then
                    REPOFILE="$repo_base/xCAT-otherpkgs$localrepoindex.repo"
                    echo "[xcat-otherpkgs$localrepoindex]" > $REPOFILE
                    echo "name=xcat-otherpkgs$localrepoindex" >> $REPOFILE
```

But the zypper verification step that follows refers to the repo by index, not localrepoindex:

sh
```
                elif [ $haszypper -eq 1 ]; then
                    ...
                    result=`zypper --non-interactive refresh xcat-otherpkgs$index 2>&1`
                    if [ $? -eq 0 ]; then
                        rc=0
                        array_set_element repo_path $index $path
                    else
                        ...
                            result=`zypper sd xcat-otherpkgs$index`
```

With no http entries in OTHERPKGDIR, urlrepoindex is 0, the two are equal, and everything works. With one or more http entries, localrepoindex = urlrepoindex + index, so the repo file is written as xcat-otherpkgs1 (say) but the refresh is run against xcat-otherpkgs0 — which is the http repo, not the one just added. Consequences:

The refresh "succeeds" against the wrong repo, so rc=0 is set and the package is put on the repo_pkgs list even though the newly added local repo was never refreshed; zypper install then fails to find it, or pulls it from the wrong place.
On the failure path, zypper sd xcat-otherpkgs$index deletes the http repo instead of the one that failed, leaving the broken local repo in place.

The yum/dnf branch is not affected (it uses $yumcmd list $fn, which does not name the repo).

Environment
xCAT version: <rpm -q xCAT output> (management node: Rocky Linux 9)
SLES nodes (zypper branch) with an osimage whose otherpkgdir contains at least one http:// entry in addition to a local directory

Steps to reproduce
chdef -t osimage <sles-image> otherpkgdir=http://mirror.example.com/somerepo/,/install/post/otherpkgs/<os>/<arch>
Put a package that exists only in the local directory into the image's otherpkgs.pkglist.
[...]

Issue #7847, opened 2026-09-14
otherpkgs: http/https OTHERPKGDIR handling — trailing space in generated baseurl, and [ -n "OTHERPKGDIR_INTERNET" ] missing $


ChrisPJames:
Summary

Two small bugs in the code that splits OTHERPKGDIR into local and http entries and writes repo files for the http ones. Both are in the same few lines, so filing together.

1. Every http repo URL gets a trailing space.

When OTHERPKGDIR is parsed, http entries are accumulated with a space before the comma:

sh
```
        if [ $dirtype = 'http' ]; then
            OTHERPKGDIR_INTERNET="${OTHERPKGDIR_INTERNET}${dir} ,"
```

Later the list is split on , alone (IFS=','), so each URL retains the trailing space and it goes straight into the repo file:

sh
`                echo "baseurl=$url" >> $REPOFILE`

producing baseurl=http://mirror.example.com/repo/  — note the space before end-of-line. dnf/yum's ini parser happens to strip trailing whitespace so this works by accident on EL, but the apt branch writes deb http://mirror.example.com/repo/  into sources.list.d, and any future consumer of these files that does not trim will break.

2. if [ -n "OTHERPKGDIR_INTERNET" ] tests a literal string.

sh
```
    #add repo for url repos in otherpkgdir
    if [ -n "OTHERPKGDIR_INTERNET" ];then
```

The $ is missing, so this compares the 20-character string OTHERPKGDIR_INTERNET against empty and is always true. Today it is harmless only because the loop body iterates over an empty array when no http dirs are configured, but the guard is dead code and the intent is clearly -n "$OTHERPKGDIR_INTERNET".

Environment
xCAT version: <rpm -q xCAT output> (management node: Rocky Linux 9)
Affects any osimage whose otherpkgdir attribute contains an http:// or https:// entry

Steps to reproduce
chdef -t osimage <image> otherpkgdir=/install/post/otherpkgs/<os>/<arch>,http://mirror.example.com/somerepo/
updatenode <node> -P otherpkgs
cat -A /etc/yum.repos.d/xCAT-otherpkgs0.repo on the node

Actual result
```
[xcat-otherpkgs0]$
name=xcat-otherpkgs0$
baseurl=http://mirror.example.com/somerepo/ $
enabled=1$
```

(cat -A shows the trailing space before $.)

Expected result
baseurl=http://mirror.example.com/somerepo/$
Fix
diff
```
     for dir in ${dir_array[@]}
[...]

Issue #7846, opened 2026-09-14
otherpkgs: logs "<packages> installed." even when the install failed


ChrisPJames:
Summary

In the repository-install branches, otherpkgs records the failure correctly and then, on the very next line, unconditionally logs success for the same packages:

sh
```
            result=`eval $envlist $yumcmd -y install $repo_pkgs 2>&1`
            R=$?
            if [ $R -ne 0 ]; then
              RETURNVAL=$R
              logger -p local4.err -t $log_label "$envlist $yumcmd -y install $repo_pkgs failed."
            fi
            logger -p local4.info -t $log_label "$repo_pkgs installed."
```

The same pattern is in the zypper and apt-get install branches. The plain rpm -Uvh / dpkg -i fallback is worse — it logs "$plain_pkgs installed." with no check at all:

sh
```
        result=`eval $envlist $supdatecommand $plain_pkgs 2>&1`
        R=$?
        if [ $R -ne 0 ]; then
              RETURNVAL=$R
        fi
        logger -p local4.info -t $log_label "$plain_pkgs installed."
```

So a failed transaction produces this in syslog:

```
xcat: yum -y install foo bar baz failed.
xcat: foo bar baz installed.
```

and for the rpm fallback just:

`xcat: foo bar baz installed.`

with nothing at all indicating failure, because the actual rpm output is only echoed when VERBOSE is set. The script's exit code is right (RETURNVAL is set), but anyone reading the node log, or alerting on the installed. line, is told the packages went on when they did not.

The post-remove branches have the same shape ("$repo_pkgs_postremove removed." and "$plain_pkgs_postremove removed." logged regardless of $R).

Environment
xCAT version: <rpm -q xCAT output> (management node: Rocky Linux 9)
Any node OS; all package-manager branches are affected

Steps to reproduce
Add a package name to an otherpkgs.pkglist that exists in the repo metadata but whose dependencies cannot be satisfied (or, simpler, temporarily break the repo's baseurl after yum list has succeeded so yum install fails).
updatenode <node> -P otherpkgs
grep xcat /var/log/messages on the node.

Actual result
```
xcat:  yum -y install <pkgs> failed.
xcat: <pkgs> installed.
```

Expected result

Only one of the two — failed. on failure, installed. on success — and the rpm/dpkg fallback and the remove branches should follow the same rule.

Fix

Move the success message into an else. For the yum branch:

diff
```
[...]

Issue #7845, opened 2026-09-14
otherpkgs: leftover diagnostic echo lines print the entire pkglist on every run, regardless of VERBOSE


ChrisPJames:
Summary

Inside the #NEW_INSTALL_LIST# sublist loop, after the package list is split into an array, otherpkgs prints two diagnostic lines unconditionally:

sh
```
    oifs=$IFS
    IFS=$','
    pkgsarray=($pkglist)
    IFS=$oifs
    echo "pkgsarray: ${pkgsarray[@]}, ${#pkgsarray[@]}"
    echo "yum/dnf: $hasyum ($yumcmd), apt: $hasapt, zypper: $haszypper"
```

Everything else in the script that reports internal state is gated on if [ $VERBOSE ]; these two are not. They look like debugging left in from the dnf/yum detection change. The result is that every updatenode ... -P otherpkgs and every node install echoes the complete package list — dozens to hundreds of names, space-separated, on one line — followed by the package-manager flags, once per sublist, into the updatenode output and the node's /var/log/xcat/xcat.log.

On a node with several #NEW_INSTALL_LIST# sections that is several such lines per run, and across a noderange of any size the updatenode output becomes mostly pkglist dumps rather than the install results one actually wants to see.

Environment
xCAT version: <rpm -q xCAT output> (management node: Rocky Linux 9)
Any node OS; the lines are unconditional

Steps to reproduce
Any otherpkgs.pkglist with more than a handful of entries.
updatenode <node> -P otherpkgs

Actual result
```
<node>: pkgsarray: ipmitool mt-st mtx <... every package in the list ...>, 47
<node>: yum/dnf: 1 (yum), apt: 0, zypper: 0
<node>: ipmitool mt-st mtx ... installed.
```

Expected result

Nothing printed for these two lines unless VERBOSE is set, consistent with every other diagnostic in the script.

Fix

Either delete the two lines, or gate them the same way as the rest of the script:

```
diff
     oifs=$IFS
     IFS=$','
     pkgsarray=($pkglist)
     IFS=$oifs
-    echo "pkgsarray: ${pkgsarray[@]}, ${#pkgsarray[@]}"
-    echo "yum/dnf: $hasyum ($yumcmd), apt: $hasapt, zypper: $haszypper"
+    if [ $VERBOSE ]; then
+        echo "pkgsarray: ${pkgsarray[@]}, ${#pkgsarray[@]}"
[...]

PR #7850, opened 2026-09-14
fix(otherpkgs): the postscript truncates its syslog output and logs a failed install as installed

### LINKED ISSUE #7844: otherpkgs: package-manager output is logged to syslog as a single #012-escaped message and silently truncated
Summary

otherpkgs captures the full output of yum/dnf/zypper/apt-get into $result and hands it to logger as one argument:

`result=`eval $envlist $yumcmd -y upgrade 2>&1``
...
`logger -p local4.info -t $log_label "$result"`

logger sends the whole multi-line string as one syslog message. rsyslog (default $EscapeControlCharactersOnReceive on) rewrites every embedded newline as #012, so an entire yum transaction — dependency resolution, the package table, every Erasing/Installing line, the summary — lands in /var/log/messages (and on the xCAT management node via remote syslog) as a single line several kilobytes long.

Two consequences:

Unreadable. The transaction is one line with #012 separators; you cannot grep a package name and see its context, and log viewers/Observium/journal tooling render it as a wall of text.
Silently truncated. rsyslog's default $MaxMessageSize is 8 KiB. Any transaction whose output exceeds that is cut off at the end — which is exactly where Failed:, Error: and the final package summary appear. In the example below the Dependency Removed: list ends after 3 of 8 packages; the remainder never reached the log.

Environment
xCAT version: <rpm -q xCAT output> (management node: Rocky Linux 9)
Node in the example: RHEL 7.9 x86_64, yum
Also affects any node where otherpkgs runs (dnf, zypper, apt-get branches share the same pattern)

Steps to reproduce
Put a pkglist in place whose -package pre-remove (or upgrade) touches more than a handful of packages, e.g. -rhn-client-tools, -realmd, -initial-setup on an EL7 node.
updatenode <node> -P otherpkgs (or a normal install).
grep xcat /var/log/messages on the node (or on the MN's remote syslog).

Actual result

One line, ~8 KiB, #012-escaped, truncated (excerpt, trimmed in the middle):

```
Sep 14 17:23:43 bocs6 xcat.deployment.postscript: INFO  Loaded plugins: langpacks, product-id, search-disabled-repos#012No Match for argument: gnome-initial-setup*#012No Match for argument: fprintd-pam#012Resolving Dependencies#012--> Running transaction check#012---> Package cryptsetup.x86_64 0:2.0.3-6.el7 will be erased#012--> Processing Dependency: cryptsetup for package: 1:python-blivet-0.61.15.76-1.el7_9.noarch#012---> Package initial-setup.x86_64 0:0.3.9.45-1.el7 will be erased ... #012Removed:#012  cryptsetup.x86_64 0:2.0.3-6.el7  ... #012Dependency Removed:#012  anaconda-core.x86_64 0:21.48.22.159-1.el7                                     #012  anaconda-gui.x86_64 0:21.48.22.159-1.el7                                      #012  anaconda-tui.x86_64 0:21.48.22.159-1.el7
Sep 14 17:23:49 bocs6 xcat.deployment.postscript: INFO  ipmitool mt-st mtx installed.
```

(The transaction removed 16 packages; the log ends mid-list at the 8 KiB boundary.)

Expected result

One syslog message per line of package-manager output, the same way the output appears on the console — greppable, and never truncated by the message-size limit:

```
Sep 14 17:23:43 bocs6 xcat.deployment.postscript: INFO  Resolving Dependencies
Sep 14 17:23:43 bocs6 xcat.deployment.postscript: INFO  --> Running transaction check
Sep 14 17:23:43 bocs6 xcat.deployment.postscript: INFO  ---> Package cryptsetup.x86_64 0:2.0.3-6.el7 will be erased
...
Sep 14 17:23:43 bocs6 xcat.deployment.postscript: INFO  Dependency Removed:
Sep 14 17:23:43 bocs6 xcat.deployment.postscript: INFO    anaconda-core.x86_64 0:21.48.22.159-1.el7
...
Sep 14 17:23:49 bocs6 xcat.deployment.postscript: INFO  ipmitool mt-st mtx installed.
```
Fix

logger logs each line of standard input as a separate message when no message argument is given. Feeding $result on stdin instead of as an argument fixes both the escaping and the truncation, with no other behavioural change. The offending line is byte-identical in all seven places it occurs (the three upgrade branches, the three repository pre-remove branches, and the plain rpm -ev pre-remove), so:

`sed -i 's/logger -p local4.info -t \$log_label "\$result"/echo "$result" | logger -p local4.info -t $log_label/' /install/postscripts/otherpkgs`

Representative hunk:

```
diff
     result=`eval $envlist $yumcmd -y upgrade 2>&1`
     R=$?
     if [ $R -ne 0 ]; then
           RETURNVAL=$R
     fi
-    logger -p local4.info -t $log_label "$result"
+    echo "$result" | logger -p local4.info -t $log_label
     if [ $VERBOSE  ]; then
       echo "$result"
     fi
```

Feeding stdin also avoids the argument-length limit and means lines beginning with - (e.g. --> Running transaction check) can never be mistaken for logger options.

Tested on RHEL 7.9 (util-linux 2.23) and Rocky 9 (util-linux 2.37); logger has behaved this way on stdin since long before either.

dhilst:
`otherpkgs` passed the whole package-manager transaction to `logger` as one argument, so rsyslog escaped every newline to `#012` and cut the message at 8 KiB, which is where `Failed:` and the summary appear; it then logged "`<packages> installed.`" whether or not the install returned zero, at eight sites; `[ -n "OTHERPKGDIR_INTERNET" ]` omitted the `$` and always tested true; the list that guard walks left a space before each separator, which reached the generated `baseurl`; and the zypper path refreshed and removed `xcat-otherpkgs$index` while the file had been written as `xcat-otherpkgs$localrepoindex`, which differ once an http `OTHERPKGDIR` entry is present. Five commits, one per defect, each with its own assertions in the new `xCAT-test/bats/postscripts_otherpkgs.bats`, which fail against the parent commit.

Fix #7844 #7845 #7846 #7847 #7848


viniciusferrao (changes requested):
Thanks @dhilst, I would only suggest a SLES test:

could `run_zypper_local_repo()` execute
the whole refresh/cleanup branch rather than individual lines? It
currently runs `zypper sd` even in the success case, so it doesn’t
check that a successful refresh keeps the repository. The alias
assertions are useful; checking that distinction would complete them.

Otherwise... it's LGTM.

viniciusferrao (approved):
LGTM

PR #7851, opened 2026-09-14
fix(xcatdsklspost): updatenode leaves /xcatpost populated when site.cleanupdiskfullxcatpost is set

### LINKED ISSUE #7849: `updatenode` repopulates `/xcatpost` and leaves it behind despite `cleanupdiskfullxcatpost=yes`
With `cleanupdiskfullxcatpost=yes` in the `site` table, a successful diskful deployment cleans `/xcatpost` as expected. However, a subsequent `updatenode` execution downloads the postscripts again and leaves them in `/xcatpost`, undoing that cleanup.

We can observe this in a hierarchical deployment with service nodes.

Steps to reproduce:
1. Set `cleanupdiskfullxcatpost=yes`.
2. Complete a diskful node deployment, the log reports `cleanup of /xcatpost completed`.
3. Run a postscript on that node, for example:
   ```sh
   updatenode <node> -P setroute
   ```
4. After successful completion, `/xcatpost` is fully populated again

I would expect successful `updatenode` postscript execution to honor the cleanup preference, or a corresponding option to provide that behavior if this setting is intentionally limited to installation.

The main concern here is unnecessary information exposure on multiuser nodes: `/xcatpost` is mode `0755`, and many retained files are world-readable. Depending on site postscript contents, these could expose configuration details or embedded secrets. But in any case, enabling cleanup during deployment does not keep this material removed after subsequent maintenance operations, and could come as a surprise to the site admins.

Would there be a way to invoke cleanup again after successful `updatnode` executions?

dhilst:
An administrator sets `site.cleanupdiskfullxcatpost`, deploys a diskful node and finds `/xcatpost` empty, and then the next `updatenode` downloads the postscripts again and leaves them there at mode 0755, world-readable, on every maintenance run. `xCAT/postscripts/xcatdsklspost` is what repopulates `/xcatpost`, and it read only `CLEANUPXCATPOST`, which it already honours on that same path, so the two sibling attributes differed by nothing except which code path consulted them — install-only was never the design, the asymmetry is the defect, and a third control for one directory would only make an administrator choose between settings that do the same thing. It now reads `CLEANUPDISKFULLXCATPOST` as well, keeps `updateflag.awk` exactly as the install-time cleanup does, and leaves a failed run populated so the postscripts are still there to diagnose. `xCAT-test/bats/xcatdsklspost_xcatpost_cleanup.bats` captures it: three of its six tests fail on the test commit and pass on the fix.

Fix #7849


viniciusferrao (changes requested):
Daniel,

We should restrict the new `CLEANUPDISKFULLXCATPOST` cleanup to diskful
nodes. `xcatdsklspost` also handles diskless and statelite nodes, and
the site setting reaches those too. With `cleanupdiskfullxcatpost=yes`
and `cleanupxcatpost=no`, the new block still deletes their postscripts
after a successful run.

Thanks.

viniciusferrao (approved):
LGTM.

PR #7853, opened 2026-09-15
fix(xcat-core): one node's request to the install monitor blocks every other node

### LINKED ISSUE #7852: xcatd: install monitor runs nodeset next / updatenodestat synchronously — blocks all other installing nodes, and a plugin failure kills the monitor with no restart
This worked for our environment but may not be appropriate for all sites.  I have logged this anyway.

Summary

In do_installm_service, the handlers for the two messages every installing node sends — next (→ nodeset $node next) and installstatus (→ updatenodestat) — call plugin_command() inline in the install monitor process. The fork that used to wrap them is still in the source, commented out:

perl
```
                if ($text =~ /next/) {
                    my %request = (
                        command => ['nodeset'],
                        node    => [$node],
                        arg     => ['next'],
                    );

                    # node should be blocked, race condition may occur otherwise
                    #my $pid=xCAT::Utils->xfork();
                    #unless ($pid) { # fork off the nodeset and potential slowness
                    xCAT::MsgUtils->trace(0, "I", "xcatd: triggering \'nodeset $node next\'...");
                    plugin_command(\%request, undef, \&build_response);
                    #exit(0);
                    #}
                    close($conn);
```

and the same for installstatus. The comment says the node "should be blocked" to avoid a race. But the node is not blocked by this — see below — so the synchronous call buys nothing, while costing three things:

Every other installing node is stalled. The install monitor is a single process with a single accept loop. While nodeset next runs (on a management node with a few hundred nodes, nodeset → makedhcp/tftp/boot-config regeneration takes several seconds), no other node's installstatus, next, getpostscript or syncfiles is serviced; they queue in the listen backlog. With a batch of nodes installing in parallel the monitor spends most of its time inside plugin_command and nodes see the monitor as unresponsive.
A plugin failure takes the install monitor down. In the SSL path, plugins run in a forked child so an exit, an uncaught die, or an XS-library crash only kills that request. In the install monitor path they run in the monitor itself. nodeset fans out through destiny to several other plugins via do_request; any of them calling exit() or dying outside an eval terminates the install monitor process.
Nothing restarts it. When the monitor dies, ssl_reaper in the parent just records the fact:
perl
```
           if ($CHILDPID == $pid_MON) {
               $pid_MON = 0;
           }
```

There is no respawn. Port xcatiport (3002) stops listening, and from that moment every node that finishes installing sends next/installstatus into a closed port. nodeset next never runs, so the node's boot target is never switched from install to local disk, and the node sits at installing in nodelist.status (and re-installs on its next PXE) until someone notices and restarts xcatd.

Why the "node should be blocked" comment does not hold

The reply to the node is sent before the plugin call, not after:

```
`perl`
            while (my $text = <$conn>) {
                alarm(0);
                print $conn "done\n";          # <-- node released here
                ...
                if ($text =~ /next/) {
                    ...
                    plugin_command(\%request, undef, \&build_response);   # runs after "done" was already sent
                    close($conn);
```

$conn comes from $socket->accept, and IO::Socket objects have autoflush on, so done goes on the wire immediately. The node side (updateflag.awk) does:

```
`awk`
    if($0 == "done")
      break
```

and exits — the postscript continues and the node reboots. So whether nodeset next runs inline or in a forked child, the node has already been told done and is already on its way to reboot. The synchronous call does not close the window between "node reboots" and "boot config updated"; it only makes the install monitor unavailable to everyone else during that window. (If that window is a real concern the fix is to move print $conn "done\n" after the nodeset in the next handler — which is orthogonal to whether the work is forked.)

The same function already forks nodeset enact in the setiscsiparms handler ("fork off the nodeset and potential slowness"), so forking plugin_command from the install monitor is an established pattern in this code, not a new one.

Environment
xCAT version: <rpm -q xCAT output> (management node: Rocky Linux 9)
Observed with: <number> nodes installing concurrently, <OS> osimages
<paste from /var/log/messages: "possible BUG encountered by xCAT install monitor service" lines, or the last "xcatd: install monitor" trace line before it went quiet>
<paste: ss -ltnp | grep 3002 showing nothing listening, ps -ef | grep "install monitor" showing no process, while the main xcatd and SSL listener were still up>

Steps to reproduce
Kick off an install on a batch of nodes (rinstall <noderange>) large enough that several reach the end of the kickstart %post within a few seconds of each other.
On the MN, watch 'ss -tn state established "( sport = :3002 )"' — connections accumulate while one nodeset next is in flight.
To reproduce the monitor death directly: put a die "test" (or exit 1) outside any eval in a plugin reached by nodeset (e.g. at the top of process_request in destiny.pm), then let one node send next. Observe that xcatd: install monitor is gone from ps, port 3002 is closed, xcatd itself is still running and lsxcatd -v is happy, and every subsequent node stays at installing.
Actual result

Nodes stuck at installing, port 3002 not listening, no error visible from xcatd status; only systemctl restart xcatd recovers.

Expected result

A plugin failure while handling one node's next/installstatus affects only that node; other nodes' install-monitor requests are serviced concurrently; the install monitor stays up.

Fix

Restore the fork that is already there in comments, using xexit so the child reaps its own children before exiting (same as the setiscsiparms handler):

diff
```
                 if ($text =~ /next/) {
                     my %request = (
                         command => ['nodeset'],
                         node    => [$node],
                         arg     => ['next'],
                     );

-                    # node should be blocked, race condition may occur otherwise
-                    #my $pid=xCAT::Utils->xfork();
-                    #unless ($pid) { # fork off the nodeset and potential slowness
-                    xCAT::MsgUtils->trace(0, "I", "xcatd: triggering \'nodeset $node next\'...");
-                    plugin_command(\%request, undef, \&build_response);
-                    #exit(0);
-                    #}
+                    # "done" has already been sent to the node above, so running
+                    # this inline does not block the node - it only blocks the
+                    # install monitor for every other node.  Fork it, as the
+                    # setiscsiparms handler already does.
+                    my $pid = xCAT::Utils->xfork();
+                    unless ($pid) {
+                        xCAT::MsgUtils->trace(0, "I", "xcatd: triggering \'nodeset $node next\'...");
+                        plugin_command(\%request, undef, \&build_response);
+                        xexit(0);
+                    }
                     close($conn);
                 } elsif ($text =~ /installstatus/) {
                     my @tmpa = split(' ', $text);
                     for (my $i = 1 ; $i <= @tmpa - 1 ; $i++) {
                         my $newstat = $tmpa[$i];
                         my %request = (
                             command => ['updatenodestat'],
                             node    => [$node],
                             arg     => ["$newstat"],
                         );
                         # TODO, better to use constant here for the specified status
                         if ($newstat eq 'installing') {
                             xCAT::MsgUtils->trace(0, "I", "xcat.updatestatus - $node: provisioning detected...");
                         } elsif ($newstat eq 'netbooting') {
                             xCAT::MsgUtils->trace(0, "I", "xcat.updatestatus - $node: provisioning detected...");
                         }
-                        # node should be blocked, race condition may occur otherwise
-                        #my $pid=xCAT::Utils->xfork();
-                        #unless ($pid) { # fork off the nodeset and potential slowness
-                        plugin_command(\%request, undef, \&build_response);
-
-                        #exit(0);
-                        #}
+                        my $fpid = xCAT::Utils->xfork();
+                        unless ($fpid) {
+                            plugin_command(\%request, undef, \&build_response);
+                            xexit(0);
+                        }
                     }
                     close($conn);
```

Running with this patch on our management nodes since 2025: multiple nodes installed across many sites and vessels, no install-monitor deaths, no nodes stuck at installing, and no reinstall loops attributable to the reboot/boot-config window (which, per the analysis above, this change neither opens nor closes).

Separately from the fork, it would be worth having the parent respawn the install monitor when ssl_reaper sees $pid_MON exit, as it is the only xcatd subprocess whose loss is silent to xcatd status and to every client except installing nodes.

dhilst:
The xcatd install monitor serves every installing node from one accept loop. It accepts a
connection, resolves the peer to a node and dispatches the request in line, so nothing else is
accepted until that request returns. On a flat el9 cluster, with one node holding three
connections that send nothing, a second node waited 5.7 seconds for the monitor's greeting.
The wait scales with the load: 1.7 s for one held connection, 9.7 s for five.

do_installm_service in xCAT-server/sbin/xcatd calls plugin_command directly in every branch --
next, installstatus, getpostscript, syncfiles, hpcbootstatus, locktftpdir. The per-request fork
that once stood there is commented out with the note that the node must be blocked, because
'nodeset next' and 'installstatus' for one node write the same chain row.

The monitor now gives each connection its own child and keeps that ordering per node: the child
for a node reads a pipe left by the previous child for the same node, and starts when that pipe
reaches end of file. Live children are capped at 64 and the rest wait in the listen backlog. The
parent reaps them and releases the gate, and a child that dies no longer takes the monitor with
it.

xCAT-test/unit/xcatd_install_monitor_concurrency.t lifts do_installm_service out of the program
and drives it with real clients on a port of its own. Without this change a second node waits
5.7 seconds for its greeting and the monitor does not survive a request that kills its handler;
with it the second node waits 0.004 s. Removing the pipe gate turns the ordering assertion red,
and raising the limit turns the backlog assertion red. Measured on the lab management node with
the patched binary installed, a second node waits 0.008 s against 5.716 s on the stock binary,
and a real getpostscript from the compute node is served by a forked handler

fix #7852


viniciusferrao (changes requested):
Daniel,

1. Could we preserve per-node ordering when a waiting child dies?
   With A running and B and C queued for the same node, killing B
   closes its pipe and lets C run before A finishes. The fork-failure
   fallback also closes the predecessor without waiting. Both paths
   can reintroduce the chain-row race. Please keep later requests
   behind the earlier work, with regression tests for these cases.

2. Could the concurrency test use a scratch PID file instead of
   `/var/run/xcat/installservice.pid`? Running it as root overwrites
   the live monitor’s PID file. The capped-monitor case also leaves
   its dead PID behind, since the restore happens before that case.
   Please redirect every PID-file access in the lifted service,
   rather than saving and restoring the host’s file.

Thanks!

viniciusferrao (changes requested):
Thanks @dhilst, I've missed one more thing.

I found one remaining wakeup race: if the active child exits after
the queue check but before `accept()`, SIGCHLD is handled before
there is an accept to interrupt. The parent then blocks even though
a connection is already queued and ready to run.

Could we use a persistent wakeup, such as a self-pipe, or a bounded
wait that rechecks the queue? A regression that forces the child to
exit between the queue check and accept would cover this.

That's a suggestion. You should see of it applies.

Thanks!

viniciusferrao (approved):
LGTM.

PR #7854, opened 2026-09-15
fix(xcat-core): the genesis directory rename leaves the build scripts and tests on the old path

dhilst:
`xCAT-genesis-builder` becomes `xCAT-genesis-base` in #7833, and two other pull requests add
code that names the directory: #7818 adds the genesis deb build in `builddebs.pl` and eight
test files, #7835 adds `genesis_deb_per_codename.t` and ships `verify-genesis-payload`. Each
branch is correct alone. A tree that carries #7833 and either of the others copies a directory
that no longer exists, and its tests read absent paths and skip.

This change points `build_one_genesis_deb` in `builddebs.pl`, `buildsources_genesis_base` in
`buildrpms.pl` and ten test files at `xCAT-genesis-base`, and deletes
`genesis_base_deb_arch.bats`, whose subject `debuild-xcat-genesis-base` no longer exists under
any name.

**This pull request depends on #7833, #7818 and #7835, and carries them in its diff until they
merge.** The change itself is one commit of twenty lines; everything else shown here belongs to
those three. Merge them first and this diff collapses to that commit. It cannot be opened
against master alone: eight of the twelve files do not exist there, so there is nothing for the
change to edit.

`prove xCAT-test/unit/genesis_*.t` and the five genesis BATS files pass on this branch. On a tree
with #7833 and #7818 and without this change they skip or fail, because the paths they probe are
absent.

I will keep it as draft before we merge the dependencies. Again it is a single commit, but depend and carries several others, once the dependencies land into master the commits should disappear after rebase


viniciusferrao:
Is this one correct?

It looks too convoluted, @dhilst.

PR #7855, opened 2026-09-15
fix(xcat-core): makedns leaves records behind because named cannot verify its own TSIG key

dhilst:
`makedns -d <noderange>` fails on EL 8 and EL 9 with `Failure encountered updating <zone> with
entry '', error was FORMERR`, and the records it was asked to remove stay in the zone. `named`
logs `request has invalid signature: TSIG xcat_key: tsig verify failure (BADKEY)`.

`ddns.pm` writes the TSIG key into `named.conf` and then signs its own updates with a different
algorithm. When `Net::DNS` is older than 1.36 the key block is rewritten to `hmac-md5`, while the
update is signed with the algorithm from `site.dhcpomapialgorithm`, which defaults to
`hmac-sha256`. `named` cannot verify a signature made with an algorithm the key does not declare,
so it answers BADKEY and the zone is left unchanged. A second defect compounds it: the retry path
appends a TSIG record to a message that already carries one, and a message with two TSIG records
is malformed.

This change stops `ddns.pm` rewriting the algorithm it is about to sign against, and makes the
retry build a fresh message instead of adding a second TSIG record to the sent one.
`OmapiPolicy.pm` reports the algorithm it selected so the two agree.

`ddns_named_key_algorithm.t` captures the key block and the signing algorithm disagreeing, and
`ddns_update_retry.t` captures the retry carrying two TSIG records. Both fail without the fix.


viniciusferrao (approved):
LGTM, however, one suggestion: add a retry case with `rr_del()` and a
prerequisite, and check that both survive unchanged. The current
fixture only adds an A record, while the reported failure involves
`makedns -d`.

PR #7864, opened 2026-09-23
feat(os): support openEuler LTS management, service and compute nodes

viniciusferrao:
This PR adds support for openEuler 20.03 LTS SP4, 22.03 LTS SP4 and 24.03 LTS SP4 on x86_64. It covers management nodes, installed and diskless service nodes, and stateful and stateless compute nodes.

`copycds` imports native media, `go-xcat` installs signed packages, and Genesis uses native kernels and packages. Install and netboot profiles provide compute and service images.

Depends on xcat2/xcat-dep#77.


xCAT-server/share/xcat/netboot/rh/dracut_047/xcatroot
dhilst, line 1: I think you want to edit dracut_105 files, dracut_047 is old code

xCAT-test/integration/openeuler_postscript_repositories.t
dhilst, line 1: This test is mixing perl python and shell

PR #7865, opened 2026-09-23
fix(xcat-core): Version still names the released 2.19.0

dhilst:
2.20 version bump


viniciusferrao:
@dhilst it should be 2.20

Then we use the backport scheme on the 2.19 branch.

Version
```diff
@@ -1 +1 @@
-2.19.0
+2.19.1
```
viniciusferrao:
```suggestion
2.20.0
```

viniciusferrao (changes requested):

viniciusferrao (approved):

PR #7866, opened 2026-09-23
fix(xcat-core): repository indexing fails on every target on a shared build tree

dhilst:
createrepo_dir passed --database, which writes *.sqlite.bz2. Building those needs
SQLite, and SQLite needs POSIX locks, POSIX locks fails into chained NFS


viniciusferrao (approved):
Thanks @dhilst.

One non-blocking suggestion: use `--no-database` explicitly rather
than just removing `--database`. createrepo_c versions before 1.0
generate SQLite metadata by default, so the explicit flag would
keep the same problem from returning on older build hosts.

Otherwise, LGTM.

PR #7867, opened 2026-09-23
fix(xcat-core): build locks the shared tree refuses to grant

dhilst:
Update the build scripts to implement locking with mkdir instead of flock

flock fails into chained NFS


viniciusferrao (changes requested):
Thanks @dhilst.

Could we handle cancellation for the checkout lock used by
`builddebs.pl` too? The new lock relies on `DESTROY`, which doesn't
run when the process is terminated by an unhandled SIGTERM/SIGINT.
That leaves the directory behind, and the next build reports
“another build already holds” even after the owner has exited.

`buildrpms.pl` already releases its lock in the cancellation path.
Could we add equivalent cleanup for the Debian builder, releasing
only after its build subprocesses have stopped? A test that
terminates the lock holder and then reacquires the lock would
cover this.

viniciusferrao (approved):

